PR #757, opened 2018-11-15
Detect broken HTLC links at startup

pm47:
If we have stopped eclair while it was forwarding HTLCs, it is possible
that we are in a state were an incoming HTLC
was committed by both sides, but we didn't have time to send
and/or sign the corresponding HTLC to the downstream node.

In that case, if we do nothing, the incoming HTLC will
eventually expire and we won't lose money, but the channel
will get closed, which is a major inconvenience.

This check will detect this and will allow us
to fast-fail HTLCs and thus preserve channels.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
@@ -46,10 +46,13 @@ class Switchboard(nodeParams: NodeParams, authenticator: ActorRef, watcher: Acto
   private val initialPeers = {
     val channels = nodeParams.channelsDb.listChannels()
     val peers = nodeParams.peersDb.listPeers()
-    
-    val brokenHtlcs = checkBrokenHtlcsLink(channels, nodeParams.privateKey)
-    val brokenHtlcKiller = context.actorOf(Props[HtlcReaper], name = "htlc-reaper")
-    brokenHtlcKiller ! brokenHtlcs
+
+    checkBrokenHtlcsLink(channels, nodeParams.privateKey) match {
+      case Nil => ()
+      case brokenHtlcs =>
+        val brokenHtlcKiller = context.actorOf(Props[HtlcReaper], name = "htlc-reaper")
+        brokenHtlcKiller ! brokenHtlcs
+    }
 
     channels
       .groupBy(_.commitments.remoteParams.nodeId)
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -114,8 +126,64 @@ class Switchboard(nodeParams: NodeParams, authenticator: ActorRef, watcher: Acto
...
+      .filter(_.direction == OUT)
+      .map(_.add)
+
+    // Here we do it differently because we need the origin information.
+    val relayed_out = channels
+      .flatMap(_.commitments.originChannels.values)
+      .collect { case r: Relayed => r }
+      .toSet
+
+    val htlcs_broken = htlcs_in.filterNot(htlc_in => relayed_out.exists(r => r.originChannelId == htlc_in.channelId && r.originHtlcId == htlc_in.id))
```
sstone: What happens for direct payments that we've received ? there is no corresponding `Relayed` origin so they will be considered broken ?

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -22,22 +22,34 @@ import akka.actor.{Actor, ActorLogging, ActorRef, OneForOneStrategy, Props, Stat
...
 
   // we load peers and channels from database
   private val initialPeers = {
-    val channels = nodeParams.channelsDb.listChannels().groupBy(_.commitments.remoteParams.nodeId)
+    val channels = nodeParams.channelsDb.listChannels()
     val peers = nodeParams.peersDb.listPeers()
+    
+    val brokenHtlcs = checkBrokenHtlcsLink(channels)
+    val brokenHtlcKiller = context.actorOf(Props[HtlcReaper], name = "htlc-reaper")
+    brokenHtlcKiller ! brokenHtlcs
```
sstone: Since broken HTLCs are known when we create the reaper actor, why not pass them to its constructor ?

pm47: Because they would never be garbage collected.

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
pm47: oops, good catch

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
sstone: It would also be a bit more efficient not to create the reaper actor if the list is empty

pm47: I did it, but I don't really like that kind of branching that early in the code.

Also, the fact that the reaper's state is never cleaned up is not great.

PR #761, opened 2018-11-19
Disconnect peer after a timeout when waiting for revocation

### LINKED ISSUE #745: Stuck outgoing payment with ACINQ peer
A user reports two expired payments sent through ACINQ node, details are here: https://github.com/btcontract/lnwallet/issues/99#issuecomment-432511392

pm47:
If a peer doesn't quickly reply to a `commit_sig`, we assume that it is
experiencing technical issues, and we disconnect. This will make pending
(unsigned) `update_add_htlc` to be fast-failed and will hopefully limit
the number of htlc that time out in the network.

By default we wait 20 seconds, configurable with
`eclair.revocation-timeout`.

This fixes #745.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1896,6 +1930,13 @@ class Channel(val nodeParams: NodeParams, wallet: EclairWallet, remoteNodeId: Pu
       case _ => throw CommitmentSyncError(d.channelId)
     }
 
+    commitments1.remoteNextCommitInfo match {
+      case Left(_) =>
+        // we expect a quick response from our peer
```
sstone: It's the same comment as when we've just sent a signature but it's a different context and it should be different here

PR #762, opened 2018-11-19
Auto probing of local channels

pm47:
This is a simple mechanism to test our direct peers by sending fake
small payments to them. A probe is considered successful if the peer
replies with an `UnknownPaymentHash` error.


btcontract:
While at it, what do you think about https://github.com/ACINQ/eclair-wallet/issues/94?
This behavior may be specified in config, it has proven to be a good bug hunting tool in my case.

pm47:
But the point here is to close zombie channels right?

btcontract:
To spot them, yes. What happens next may differ depending on whether this happens on server side or mobile app. In a letter case user may be interested to keep the channel open and let peer fix the bug.

pm47:
I'm not sure this probing should be enabled on eclair-wallet, it makes more sense for always-on nodes

pm47:
I think this feature could be merged, but disabled by default because:
- nodes aren't very reliable on current mainnet and this would probably cause too many channels to get closed for our users
- there is currently no way to tell destination node it is being probed, so that could cause unnecessary support work

PR #763, opened 2018-11-19
Disable local channels below reserve

pm47:
The goal is to reduce attempts from other nodes in the network to use
channels that are unbalanced and can't be used to relay payments.

This leaks information about the current balance and is a privacy
tradeoff, particularly in this simplistic implementation. A better way
would be to add some kind of hysteresis in order to prevent trivial
probing of channels.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -620,6 +621,11 @@ class Channel(val nodeParams: NodeParams, wallet: EclairWallet, remoteNodeId: Pu
                   log.info(s"adding paymentHash=${u.paymentHash} cltvExpiry=${u.cltvExpiry} to htlcs db for commitNumber=$nextCommitNumber")
                   nodeParams.channelsDb.addOrUpdateHtlcInfo(d.channelId, nextCommitNumber, u.paymentHash, u.cltvExpiry)
               }
+              if (Helpers.aboveReserve(d.commitments) ^ Helpers.aboveReserve(commitments1)) {
```
sstone: `Helpers.aboveReserve(d.commitments1) && !Helpers.aboveReserve(commitments)` is what we're actually testing for and would be easier to read

pm47: I suppose you mean `!Helpers.aboveReserve(d.commitments) && Helpers.aboveReserve(commitments1)`

PR #778, opened 2018-12-12
Optional per-node features in conf

pm47:
Allow setting specific `globalfeatures` and `localfeatures` for given nodes.


eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -160,13 +162,21 @@ object NodeParams {
     val maxAcceptedHtlcs = config.getInt("max-accepted-htlcs")
     require(maxAcceptedHtlcs <= Channel.MAX_ACCEPTED_HTLCS, s"max-accepted-htlcs must be lower than ${Channel.MAX_ACCEPTED_HTLCS}")
 
+    val overrideFeatures: Map[PublicKey, (BinaryData, BinaryData)] = config.getConfigList("override-features").map {e =>
+        val p = PublicKey(e.getString("nodeid"))
+        val gf = BinaryData(e.getString("global-features"))
+        val lf = BinaryData(e.getString("local-features"))
+      (p -> (gf, lf))
```
dpad85: Code indentation is not consistent

PR #779, opened 2018-12-12
Routing: use custom implementation for the shortest path algorithm and the graph

araspitzu:
This pull request aims at replacing the currently used library jgraph with a custom implementation of a directed-unweighted-graph, and a custom implementation of dijkstra's shortest path algorithm


araspitzu:
The new implementation computes the route cost together with the path and taking into account the amount being routed, channels with not enough capacity and or htlc_minimum_msat bigger than the payment will not be considered (this should yield better routes for big payments). The dijkstra algorithm is used together with fibonacci heaps for fast computation of the route, benchmarks are in the branch origin/microbenchmark.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -0,0 +1,326 @@
...
+    * @param amountMsat the amount (in millisatoshis) we want to transmit
+    * @param ignoredEdges a list of edges we do not want to consider
+    * @param extraEdges a list of extra edges we want to consider but are not currently in the graph
+    * @return
+    */
+  def shortestPath(g: DirectedGraph, sourceNode: PublicKey, targetNode: PublicKey, amountMsat: Long, ignoredEdges: Seq[ChannelDesc], extraEdges: Set[GraphEdge]): Seq[Hop] = {
+    dijkstraShortestPath(g, sourceNode, targetNode, amountMsat, ignoredEdges, extraEdges).map(graphEdgeToHop)
+  }
+
+  def dijkstraShortestPath(g: DirectedGraph, sourceNode: PublicKey, targetNode: PublicKey, amountMsat: Long, ignoredEdges: Seq[ChannelDesc], extraEdges: Set[GraphEdge]): Seq[GraphEdge] = {
```
pm47: Maybe make `ignoredEdges` a `Set` as well?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -0,0 +1,326 @@
...
+
+    // optionally add the extra edges to the graph
+    val graphVerticesWithExtra = extraEdges.nonEmpty match {
+      case true => g.vertexSet() ++ extraEdges.map(_.desc.a).toSet ++ extraEdges.map(_.desc.b).toSet
```
pm47: `.toSet` seem superfluous

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -0,0 +1,220 @@
+package fr.acinq.eclair.router
+
+import fr.acinq.bitcoin.Crypto.PublicKey
+import org.scalatest.FunSuite
+import RouteCalculationSpec._
+import fr.acinq.eclair.ShortChannelId
+import fr.acinq.eclair.router.Graph.GraphStructure.{GraphEdge, DirectedGraph}
+import fr.acinq.eclair.wire.ChannelUpdate
+
+class GraphSpec extends FunSuite {
```
dpad85: Indentation in this file is not consistent with the rest of the project, 2 spaces should be used, not tabs.

PR #781, opened 2018-12-13
Add exclusive locks to channels sqlite db

pm47:
Note that there is no risk of corruption currently (concurrent writes would just fail with `SQLITE_BUSY`), but with this we make sure that there is only one instance of eclair using the same `eclair.sqlite` db file at any time.

From https://www.sqlite.org/pragma.html: 
> There are three reasons to set the locking-mode to EXCLUSIVE.
> 1. The application wants to prevent other processes from accessing the database file.
> 2. [...]
> 3. [...]

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/AuditDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/AuditDb.scala
@@ -41,6 +41,8 @@ trait AuditDb {
 
   def stats: Seq[Stats]
 
+  def close: Unit
+
 }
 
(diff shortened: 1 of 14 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/AuditDb.scala
```diff
@@ -41,6 +41,8 @@ trait AuditDb {
 
   def stats: Seq[Stats]
 
+  def close: Unit
```
sstone: Scala convention would be to use `close()` since this methods has side-effects

PR #782, opened 2018-12-13
Make ping less aggressive

pm47:
It currently leads to too many disconnections, in part because the initial delay is too short (peer may be under heavy load when syncing).


rorp:
Also it would be great to be able to configure whether eclair should close connections or not if it doesn't receive ` pong`s.

pm47:
@rorp agree, the spec does make it optional:


>A node sending a ping message:
> [...]
>    if it doesn't receive a corresponding pong: 
>[...]
> MAY terminate the network connection

eclair-core/src/main/resources/reference.conf
```diff
@@ -85,6 +85,8 @@ eclair {
   router-init-timeout = 5 minutes
 
   ping-interval = 30 seconds
+  ping-timeout = 10 seconds // will disconnect if peer takes longer than that to respond
+  ping-disconnect = true // disconnect if no answer to our pongs
```
```suggestion
  ping-disconnect = true // disconnect if no answer to our pings
```

PR #784, opened 2018-12-18
Relay to channel with lowest possible balance

pm47:
Our current channel selection is very simplistic: we relay to the
channel with the largest balance. As time goes by, this leads to all
channels having the same balance.

A better strategy is to relay to the channel which has the smallest
balance but has enough to process the payment. This way we save larger
channels for larger payments, and also on average channels get depleted
one after the other.

### DIFF
--- a/eclair-core/pom.xml
+++ b/eclair-core/pom.xml
@@ -213,6 +213,12 @@
         </dependency>
         <!-- TESTS -->
         <dependency>
+            <groupId>com.softwaremill.quicklens</groupId>
+            <artifactId>quicklens_${scala.version.short}</artifactId>
+            <version>1.4.11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>com.whisk</groupId>
             <artifactId>docker-testkit-scalatest_${scala.version.short}</artifactId>
             <version>0.9.8</version>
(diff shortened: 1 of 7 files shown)


sstone:
There is no specific test for this (but there was no specific test to check that we used the channel with the maximum capacity either).

pm47:
rebased

pm47:
Let's wait for more feedbacks on the lightning-rfc's pull requests before merging this.

btcontract:
Should comment here be "lowest balance"?
https://github.com/ACINQ/eclair/blob/c75082954bae6c5a8ab0a4d3d0ca8a12ea4a7e4f/eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala#L278

pm47:
@btcontract good catch, that's a left over from previous strategy.

btcontract:
Does `selectPreferredChannel` check that new channel it deviates to is not the same channel it got a payment from? (https://github.com/lightningnetwork/lnd/issues/2470)

pm47:
I'm not sure tbh, will write a test. If we don't I don't think that's necessarily a problem.

btcontract:
It's not strictly a bug but seems like it's almost never what a sender would want.

btcontract:
Actually I think I'm wrong here and allowing this is correct: https://github.com/lightningnetwork/lnd/issues/2470#issuecomment-454385813

pm47:
NB: still waiting for https://github.com/lightningnetwork/lightning-rfc/pull/538 before merging this.

PR #790, opened 2018-12-21
Use CLTV as tie-breaker for offered htlc output sorting

araspitzu:
Fixes https://github.com/lightningnetwork/lightning-rfc/issues/448 , by using the CLTV value from the htlc-timeout-tx when sorting the htlc_sigs in the commit_signature message.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -490,8 +490,8 @@ object Commitments {
...
     val localRevocationPubkey = Generators.revocationPubKey(remoteParams.revocationBasepoint, localPerCommitmentPoint)
-    val commitTx = Transactions.makeCommitTx(commitmentInput, commitTxNumber, keyManager.paymentPoint(localParams.channelKeyPath).publicKey, remoteParams.paymentBasepoint, localParams.isFunder, Satoshi(localParams.dustLimitSatoshis), localRevocationPubkey, remoteParams.toSelfDelay, localDelayedPaymentPubkey, remotePaymentPubkey, localHtlcPubkey, remoteHtlcPubkey, spec)
-    val (htlcTimeoutTxs, htlcSuccessTxs) = Transactions.makeHtlcTxs(commitTx.tx, Satoshi(localParams.dustLimitSatoshis), localRevocationPubkey, remoteParams.toSelfDelay, localDelayedPaymentPubkey, localHtlcPubkey, remoteHtlcPubkey, spec)
+    val (commitTx, htlcOutputInfo) = Transactions.makeCommitTx(commitmentInput, commitTxNumber, keyManager.paymentPoint(localParams.channelKeyPath).publicKey, remoteParams.paymentBasepoint, localParams.isFunder, Satoshi(localParams.dustLimitSatoshis), localRevocationPubkey, remoteParams.toSelfDelay, localDelayedPaymentPubkey, remotePaymentPubkey, localHtlcPubkey, remoteHtlcPubkey, spec)
```
sstone: `spec` should be the last parameter here

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -33,6 +33,10 @@ import scala.util.Try
 object Transactions {
 
   // @formatter:off
+
+  // type alias for the map that carries around the input index of the htlc output AND the cltv of the htlc-timeout transaction
+  type HtlcTimeoutTxInputInfo = Map[Int, Long]
```
sstone: Why not use a `Map[Int, DirectedHTlc]` here ? It could then be use to compute the HTLC success and timeout tx

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -197,27 +201,37 @@ object Transactions {
...
+      txOut = outputsWithHtlcCltvInfo.map(_._1),
       lockTime = locktime)
-    CommitTx(commitTxInput, LexicographicalOrdering.sort(tx))
+
+    val (sortedTx, htlcInputIndexToCltv) = TransactionUtils.sortByBIP69AndCLTV(tx, outputsWithHtlcCltvInfo)
+    (CommitTx(commitTxInput, sortedTx), htlcInputIndexToCltv)
   }
 
-  def makeHtlcTimeoutTx(commitTx: Transaction, outputsAlreadyUsed: Set[Int], localDustLimit: Satoshi, localRevocationPubkey: PublicKey, toLocalDelay: Int, localDelayedPaymentPubkey: PublicKey, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, feeratePerKw: Long, htlc: UpdateAddHtlc): HtlcTimeoutTx = {
+  def makeHtlcTimeoutTx(commitTx: Transaction, htlcOutputInfo: HtlcTimeoutTxInputInfo, outputsAlreadyUsed: Set[Int], localDustLimit: Satoshi, localRevocationPubkey: PublicKey, toLocalDelay: Int, localDelayedPaymentPubkey: PublicKey, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, feeratePerKw: Long, htlc: UpdateAddHtlc): HtlcTimeoutTx = {
```
sstone: See comment above: we should not need `outputsAlreadyUsed` anymore

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -442,6 +456,32 @@ object Transactions {
...
+    * Finds the index of a txOut in a transaction, given the pubKeyScript and some optional metadata
+    * @param tx
+    * @param pubkeyScript
+    * @param outputsAlreadyUsed
+    * @param amount_opt
+    * @param cltv_opt the cltv of the htlc-timeout-tx the is going to spend @param pubKeyScript
+    * @param htlcOutputInfo a map containing the correct index of the input for htlc output and it's 2nd stage cltv
+    * @return
+    */
+  def findPubKeyScriptIndex(tx: Transaction, pubkeyScript: BinaryData, outputsAlreadyUsed: Set[Int], amount_opt: Option[Satoshi], cltv_opt: Option[Long] = None, htlcOutputInfo: HtlcTimeoutTxInputInfo = Map.empty): Int = {
```
sstone: See above: we should not need  `outputsAlreadyUsed` anymore

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
araspitzu: Got it

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
araspitzu: Initially i thought of this as a special case only for offered htlcs but i like the suggestion, we can get rid of 'OutputsAlreadyUsed' and store all the information (input_index, htlc) within this map. Will do.

pm47:
Let's bring back this PR to life! :ghost:

sstone:
I still finds this too complex, and it does not fix the hack we introduced with the "outputs already used" set. I think instead we should address this in 2 steps:
- first, create a proper link between commitment specs and commit tx outputs, instead of matching pubkey scripts to find which commit tx output we should be spending
- then, update how outputs are sorted to check CLTV expiries when amounts and payment hashes are the same

To implement the first step, when we build the commit tx and HTLC txs:
- filter spec items and prune dust outputs etc...
- from each remaining spec item create a pair (item, tx output)
- sort the list: the tx outputs become our commit tx outputs, built from the corresponding spec item in the list. 

WDYT ?

araspitzu:
I agree it's too complex, i initially tried to link outputs with HTLCs with the type alias but it's clearly not readable enough. I like the suggestion of creating a stronger link between the `CommitmentSpec` and the transaction outputs probably with a Map or a list or tuples, this will be easier to carry around and to reason about (rather than matching with pubkey script).

sstone:
I have an idea for that, I'll open a PR so we can discuss it

araspitzu:
I'm curious to see your proposal :), in any case i just pushed mine in f646576. It removes both the type alias for `HtlcTimeoutTxInputInfo` and the infamous map for "outputsAlreadyUsed", instead the `makeCommitTx` method returns an additional list of tuples `(SpecItem, Int)` where spec item can be a placeholder for toLocal/toRemote OR a `DirectedHtlc`, the second element of the tuple is simply the output index of the item in the sorted commit transaction. I think we can do even better by placing this extra list in the `CommitTx` class but this triggers a codec update and db migration..

sstone:
I've updated my code to use some of your ideas and published it there https://github.com/ACINQ/eclair/tree/link-committx-outputs  
They're similar but how spec items are referenced when you build htlc txs is handled differently in `makeHtlcTxs`

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -197,39 +197,97 @@ object Transactions {
...
+
+    outputs.sortWith(CommitmentOutputLink.sort).toIndexedSeq
+  }
+
+  def makeCommitTx(commitTxInput: InputInfo,
+                   commitTxNumber: Long,
+                   localPaymentBasePoint: PublicKey,
+                   remotePaymentBasePoint: PublicKey,
+                   localIsFunder: Boolean,
+                   outputs: IndexedSeq[CommitmentOutputLink]): CommitTx = {
```
t-bast: nit: is it important this it's an `IndexedSeq` and not just a `Seq`? Maybe I'm missing it, but it seems we're not using random-access to the elements anywhere, we're always zipping with index.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -197,39 +197,97 @@ object Transactions {
...
 
-  def makeCommitTx(commitTxInput: InputInfo, commitTxNumber: Long, localPaymentBasePoint: PublicKey, remotePaymentBasePoint: PublicKey, localIsFunder: Boolean, localDustLimit: Satoshi, localRevocationPubkey: PublicKey, toLocalDelay: CltvExpiryDelta, localDelayedPaymentPubkey: PublicKey, remotePaymentPubkey: PublicKey, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, spec: CommitmentSpec): CommitTx = {
+
+  /**
+   * Represent a link between a commitment spec item (to-local, to-remote, htlc) and the actual output in the commit tx
+   * @param output transaction output
+   * @param redeemScript redeeem script that matches this output (most of them are p2wsh)
+   * @param link commitment spec item this output is built from
+   */
+  case class CommitmentOutputLink(output: TxOut, redeemScript: Seq[ScriptElt], link: CommitmentSpecLink)
```
t-bast: Same here, I think that the field `link` is really confusing.
The description of the param suggests a better name, maybe `specItem` or `item`?
I'm not a big fan of these suggestions, we need to bikeshed a bit more. This is a very critical part of the codebase, so I think we really need to spend time making it as clear as possible.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -118,7 +118,7 @@ trait StateTestsHelperMethods extends TestKitBase with fixture.TestSuite with Pa
     (payment_preimage, cmd)
   }
 
-  def addHtlc(amount: MilliSatoshi, s: TestFSMRef[State, Data, Channel], r: TestFSMRef[State, Data, Channel], s2r: TestProbe, r2s: TestProbe): (ByteVector32, UpdateAddHtlc) = {
+  def addHtlc(amount: MilliSatoshi, s: TestFSMRef[State, Data, Channel], r: TestFSMRef[State, Data, Channel], s2r: TestProbe, r2s: TestProbe, cltvExpiry: Long = 400144L): (ByteVector32, UpdateAddHtlc) = {
```
t-bast: You're not using that `cltvExpiry` anywhere :)

eclair-core/src/test/scala/fr/acinq/eclair/transactions/TransactionsSpec.scala
```diff
@@ -38,6 +40,20 @@ import scala.util.{Failure, Random, Success, Try}
 
 class TransactionsSpec extends FunSuite with Logging {
 
+  val localFundingPriv = PrivateKey(randomBytes32 :+ 1.toByte)
```
t-bast: Why the `:+ 1.toByte`?

araspitzu: I think it was a leftover from when we used to initialize private keys in compressed form but now it's useless.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -29,7 +29,10 @@ case object IN extends Direction { def opposite = OUT }
 case object OUT extends Direction { def opposite = IN }
 // @formatter:on
 
-case class DirectedHtlc(direction: Direction, add: UpdateAddHtlc)
+sealed trait CommitmentSpecLink
```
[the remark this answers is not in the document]
araspitzu: Great suggestion, i struggled to keep the naming consistent between the CommitmentSpec and this new class that represent the commitment output. I applied your suggestion in the latest commit.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -197,39 +197,97 @@ object Transactions {
...
+
+  object CommitmentOutputLink {
+    def sort(a: CommitmentOutputLink, b: CommitmentOutputLink): Boolean = (a.link, b.link) match {
+      case (DirectedHtlc(OUT, htlcA), DirectedHtlc(OUT, htlcB)) if htlcA.paymentHash == htlcB.paymentHash && htlcA.amountMsat == htlcB.amountMsat =>
+        htlcA.cltvExpiry <= htlcB.cltvExpiry
```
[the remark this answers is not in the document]
araspitzu: If `htlcA.cltvExpiry == htlcB.cltvExpiry` then the two HTLC are exactly the same and we don't care about their order in the transaction.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
araspitzu: It isn't, I've made is a simple `Seq` in the latest commit

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
araspitzu: I came up with the name `link` because this holds together the  output of the transaction with the output in our `CommitmentSpec`. Since this is very critical it's important that everyone undestands it and we should definitely bikeshed until we agree on the naming.

araspitzu:
I think we should re-do the E2E test once we agree on the changes to the code (IIRC i did some test after opening the PR but it has passed a year now), also there is https://github.com/lightningnetwork/lightning-rfc/pull/539 that can help with interop testing.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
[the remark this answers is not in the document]
t-bast: Oh you mean the outpoint will be exactly the same, righhht.
But then why do we need to explicitly sort on cltvExpiry?
Isn't the lexicographical ordering on the output already taking into account cltvExpiry and producing something deterministic?
I'm probably missing something important there :)

eclair-core/src/test/scala/fr/acinq/eclair/transactions/TransactionsSpec.scala
t-bast: Gotcha, good to remote it then, if we can make these tests slightly easier to read it's a great win!

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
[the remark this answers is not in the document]
t-bast: Ok I re-read the linked issue, I get it now. All good :)

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
araspitzu: Okay i have some proposals for the naming:
- Keep class name `CommitmentOutputLink` but field name `link` -> `specItem`
- Keep class name `CommitmentOutputLink` but field name `link` -> `commitmentSpecItem`
- Keep class name `CommitmentOutputLink` but field name `link` -> `specItem` and `output` -> `txOut`
- Rename `CommitmentOutputLink` to `CommitmentSpecOutput`
- Rename `CommitmentOutputLink` to `CommitmentSpecOutput` and field `link` -> `specItem`

t-bast: If we rename `CommitmentSpecLink` to `CommitmentOutput` (see my other comment), my preference would be to keep this class named `CommitmentOutputLink` and rename the field `link` -> `specOutput: CommitmentOutput` or `specItem: CommitmentOutput`

araspitzu: `CommitmentSpecLink` is already gone in dddac35 :). I agree with your suggestion to rename the `link` field to `specItem: CommitmentOutput`, let's see if there are other suggestions/ACK so we can move forward.

araspitzu:
Commit 453c7d0 applies the new naming, if there isn't other feedback we can go ahead with testing.

eclair-core/src/test/scala/fr/acinq/eclair/transactions/TransactionsSpec.scala
```diff
@@ -306,14 +317,68 @@ class TransactionsSpec extends FunSuite with Logging {
...
+
+    val htlcOut1 :: htlcOut2 :: htlcOut3 :: _ = commitTx.tx.txOut.toList
+
+    assert(htlcOut1.amount == 10000000.sat) // htlc1 first because of the smallest amount (BIP69)
+    assert(htlcOut2.amount == 20000000.sat) // htlc2 and htlc3 have the same amount
+    assert(htlcOut3.amount == 20000000.sat)
+
+    //htlc2 comes first because its pubKeyScript is lexicographically smaller than htlc3's
+    assert(htlcOut2.publicKeyScript.toHex < htlcOut3.publicKeyScript.toHex)
+    assert(htlcOut2.publicKeyScript.toHex == "002001ced9e8dad97b85eb0b7d101f7a79587fa890b79ffa7cf98cff1812444b8fe8")
```
t-bast: Instead of these hardcoded publicKeyScript, why not expose `outputs` from line 363 and use the following asserts:

```scala
assert(htlcOut2.publicKeyScript === outputs.filter(_.specItem == DirectedHtlc(IN, htlc2)).map(_.output.publicKeyScript).head)
assert(htlcOut3.publicKeyScript === outputs.filter(_.specItem == DirectedHtlc(IN, htlc3)).map(_.output.publicKeyScript).head)
```

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -23,6 +23,7 @@ import fr.acinq.bitcoin.Script._
 import fr.acinq.bitcoin.SigVersion._
 import fr.acinq.bitcoin._
 import fr.acinq.eclair._
+import fr.acinq.eclair.transactions.CommitmentOutput.ToLocal
```
t-bast: nit: remove

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -206,39 +207,97 @@ object Transactions {
...
+
+   * @param output transaction output
+   * @param redeemScript redeem script that matches this output (most of them are p2wsh)
+   * @param specItem commitment spec item this output is built from
+   */
+  case class CommitmentOutputLink(output: TxOut, redeemScript: Seq[ScriptElt], specItem: CommitmentOutput)
+
+  object CommitmentOutputLink {
+    def sort(a: CommitmentOutputLink, b: CommitmentOutputLink): Boolean = (a.specItem, b.specItem) match {
+      case (DirectedHtlc(OUT, htlcA), DirectedHtlc(OUT, htlcB)) if htlcA.paymentHash == htlcB.paymentHash && htlcA.amountMsat == htlcB.amountMsat =>
```
t-bast: Why do we only do this for outgoing HTLCs and not incoming as well?

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -206,39 +207,97 @@ object Transactions {
...
+
+    if (toRemoteAmount >= localDustLimit) outputs.append(
+      CommitmentOutputLink(
+        TxOut(toRemoteAmount, pay2wpkh(remotePaymentPubkey)),
+        pay2pkh(remotePaymentPubkey),
+        CommitmentOutput.ToRemote))
+
+    trimOfferedHtlcs(localDustLimit, spec).foreach(htlc =>
+      outputs.append(CommitmentOutputLink(
+        TxOut(htlc.add.amountMsat.truncateToSatoshi, pay2wsh(htlcOffered(localHtlcPubkey, remoteHtlcPubkey, localRevocationPubkey, ripemd160(htlc.add.paymentHash.bytes)))),
```
t-bast: Maybe worth removing the duplicated `htlcOffered()`?

```scala
val htlcScript = htlcOffered(localHtlcPubkey, remoteHtlcPubkey, localRevocationPubkey, ripemd160(htlc.add.paymentHash.bytes)))
TxOut(htlc.add.amountMsat.truncateToSatoshi, pay2wsh(htlcScript), htlcScript, htlc)
```

And same below for htlc received?

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -268,25 +334,32 @@ object Transactions {
...
     }
     (htlcTimeoutTxs, htlcSuccessTxs)
   }
 
-  def makeClaimHtlcSuccessTx(commitTx: Transaction, outputsAlreadyUsed: Set[Int], localDustLimit: Satoshi, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, remoteRevocationPubkey: PublicKey, localFinalScriptPubKey: ByteVector, htlc: UpdateAddHtlc, feeratePerKw: Long): ClaimHtlcSuccessTx = {
+  def makeClaimHtlcSuccessTx(commitTx: Transaction, outputs: Seq[CommitmentOutputLink], localDustLimit: Satoshi, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, remoteRevocationPubkey: PublicKey, localFinalScriptPubKey: ByteVector, htlc: UpdateAddHtlc, feeratePerKw: Long): ClaimHtlcSuccessTx = {
     val redeemScript = htlcOffered(remoteHtlcPubkey, localHtlcPubkey, remoteRevocationPubkey, ripemd160(htlc.paymentHash.bytes))
-    val pubkeyScript = write(pay2wsh(redeemScript))
-    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(htlc.amountMsat.truncateToSatoshi))
+    val Some(outputIndex) = outputs.zipWithIndex.find {
```
t-bast: There is a small regression here. If for some reason the htlc matching fails, we previously threw an `OutputNotFound` (in `findPubKeyScriptIndex`). Now it's going to be a pattern matching failure on the `Some`, and it's a bit less explicit.

It seems to me that an exception here would be really bad, as it would prevent us from reacting to a remote commitment spent on-chain. Why not make this function return an `Option[ClaimHtlcSuccessTx]` and log a big error message where it's called in `Helpers.scala` if it returns `None`? This way we would miss claiming the htlc success, but at least we'd be able to claim our main output. Maybe it's a bad idea and we'd prefer eclair to crash in that case to force the user to react, let me know your opinion.

Same for the `claimHtlcTimeoutTx` below.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -483,10 +557,10 @@ object Transactions {
     ClosingTx(commitTxInput, LexicographicalOrdering.sort(tx))
   }
 
-  def findPubKeyScriptIndex(tx: Transaction, pubkeyScript: ByteVector, outputsAlreadyUsed: Set[Int], amount_opt: Option[Satoshi]): Int = {
+  def findPubKeyScriptIndex(tx: Transaction, pubkeyScript: ByteVector, amount_opt: Option[Satoshi]): Int = {
     val outputIndex = tx.txOut
       .zipWithIndex
-      .indexWhere { case (txOut, index) => amount_opt.map(_ == txOut.amount).getOrElse(true) && txOut.publicKeyScript == pubkeyScript && !outputsAlreadyUsed.contains(index) } // it's not enough to only resolve on pubkeyScript because we may have duplicates
+      .indexWhere { case (txOut, index) => amount_opt.map(_ == txOut.amount).getOrElse(true) && txOut.publicKeyScript == pubkeyScript }
```
t-bast: nit:
```suggestion
      .indexWhere { case (txOut, _) => amount_opt.forall(_ == txOut.amount) && txOut.publicKeyScript == pubkeyScript }
```

eclair-core/src/test/scala/fr/acinq/eclair/transactions/TransactionsSpec.scala
```diff
@@ -234,20 +241,20 @@ class TransactionsSpec extends FunSuite with Logging {
 
     {
       // local spends delayed output of htlc1 timeout tx
-      val claimHtlcDelayed = makeClaimDelayedOutputTx(htlcTimeoutTxs(0).tx, localDustLimit, localRevocationPriv.publicKey, toLocalDelay, localDelayedPaymentPriv.publicKey, finalPubKeyScript, feeratePerKw)
+      val claimHtlcDelayed = makeClaimDelayedOutputTx(htlcTimeoutTxs(1).tx, localDustLimit, localRevocationPriv.publicKey, toLocalDelay, localDelayedPaymentPriv.publicKey, finalPubKeyScript, feeratePerKw)
```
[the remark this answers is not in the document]
araspitzu: Note that these are HTLC 2nd stage transactions not HTLC outputs. The order changed because we now generate the 2nd stage transactions sorted in the same way we sort the HTLC outputs in the commitment transaction, and this wasn't the case before (details at `Transactions.makeHtlcTxs`)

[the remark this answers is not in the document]
t-bast: Got it, nothing to worry about then

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
araspitzu: Yes throwing an exception would be pretty bad if it interrupts the execution flow, in fact we always "nest" the transaction generation within a `try/catch` in the [`generateTx` function](https://github.com/ACINQ/eclair/blob/33e9ac56c665573ac6f2c25aeadcd75a152fc6a3/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala#L618), this is used also when calling `makeClaimHtlcSuccessTx` to generate the HTLC-success tx.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
araspitzu: The extra CLTV tie-breaking needs to be applied only to outgoing HTLC because they're the  ones that can be spent *by us* via an HTLC-timeout transaction. The HTLC-timeout transactions have the `cltvExpiry` in the locktime and different CLTV values will result in different transactions, this forces us to agree upon an ordering even when the offered HTLC outputs are identical. I've had some trouble putting this together myself, i think it deserves a comment in the code :)

eclair-core/src/test/scala/fr/acinq/eclair/transactions/TransactionsSpec.scala
```diff
@@ -342,41 +339,51 @@ class TransactionsSpec extends FunSuite with Logging {
...
       val localSig = Transactions.sign(txinfo, localPaymentPriv)
       val remoteSig = Transactions.sign(txinfo, remotePaymentPriv)
-      Transactions.addSigs(txinfo, localFundingPriv.publicKey, remoteFundingPriv.publicKey, localSig, remoteSig)
+      (Transactions.addSigs(txinfo, localFundingPriv.publicKey, remoteFundingPriv.publicKey, localSig, remoteSig), outputs)
     }
 
-    val htlcOut1 :: htlcOut2 :: htlcOut3 :: _ = commitTx.tx.txOut.toList
+    // htlc1 comes before htlc2 because of the smaller amount (BIP69)
+    // htlc2 and htlc3 have the same amount but htlc2 comes first because its pubKeyScript is lexicographically smaller than htlc3's
+    // htlc5 comes before htlc3 and htlc4 because of the smaller CLTV
```
t-bast: That looks surprising...htlc5 has a bigger cltv expiry than htlc3 and htlc4, why does it come first?

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
t-bast: Oh yeah since this is in a Try it's ok then.
We can make this return an `Option` in a later PR to keep this focused.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -216,6 +216,11 @@ object Transactions {
   case class CommitmentOutputLink(output: TxOut, redeemScript: Seq[ScriptElt], specItem: CommitmentOutput)
 
   object CommitmentOutputLink {
+    /**
+      * We sort HTLC outputs according to BIP69 + CLVT as tie-breaker for offered HTLC, we do this only for the outgoing
```
t-bast: nit:
```suggestion
      * We sort HTLC outputs according to BIP69 + CLTV as tie-breaker for offered HTLC, we do this only for the outgoing
```

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
t-bast: Oh yeah, gotcha, since in the claim-htlc-timeout for received HTLC the cltv is directly inside the script, BIP 69 is enough. This is subtle, thanks for the comment you added!

eclair-core/src/test/scala/fr/acinq/eclair/transactions/TransactionsSpec.scala
t-bast: Got it, there were two issues: we were using incoming HTLCs instead of outgoing.
I have a small fix in https://github.com/ACINQ/eclair/commit/1341652cc7745046f7e5482865dc32719f822c7f if you want to take a look!

araspitzu: Oh indeed! I applied both your suggested changes in the latest commit 517ed54

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -29,7 +29,12 @@ case object IN extends Direction { def opposite = OUT }
 case object OUT extends Direction { def opposite = IN }
 // @formatter:on
 
-case class DirectedHtlc(direction: Direction, add: UpdateAddHtlc)
+sealed trait CommitmentOutput
+object CommitmentOutput {
+  case object ToLocal extends CommitmentOutput
+  case object ToRemote extends CommitmentOutput
+}
+case class DirectedHtlc(direction: Direction, add: UpdateAddHtlc) extends CommitmentOutput
```
pm47: The model is misleading: we seem to assume that a `DirectedHtlc` will always have a corresponding `CommitmentOutput`, which is wrong.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -206,39 +206,100 @@ object Transactions {
...
+
+  def makeCommitTxOutputs(localIsFunder: Boolean,
+                          localDustLimit: Satoshi,
+                          localRevocationPubkey: PublicKey,
+                          toLocalDelay: CltvExpiryDelta,
+                          localDelayedPaymentPubkey: PublicKey,
+                          remotePaymentPubkey: PublicKey,
+                          localHtlcPubkey: PublicKey,
+                          remoteHtlcPubkey: PublicKey,
+                          spec: CommitmentSpec): Seq[CommitmentOutputLink]  = {
```
pm47: Nit: we may want to define a `type` = `Seq[CommitmentOutputLink]` (or a container class), because the whole point of this is to preserve the order.

t-bast:
I've finished my compatibility testing.
Our behavior is correct (I also tested local / remote force-close).
We're fully compatible with c-lightning master.
However lnd doesn't correctly handle this cltv ordering. I'm opening an issue for them to fix it, it doesn't prevent us from merging this PR IMO.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
t-bast: I see what you're saying (dust HTLCs will not produce a commitment output) but I don't think it's completely wrong. It's just that the documentation should state that `CommitmentOutput` are potential commitment outputs, provided they are above the dust limit. Or rename it `PotentialCommitmentOutput` but that's horrible :)

t-bast: The goal is that the three types listed here are the only three things you can find in the outputs of a commitment tx.

araspitzu: What do you mean by _corresponding_ `CommitmentOutput`? This class is a placeholder used to link the items in `CommitmentSpec` with the outputs in the commitment transactions, instances of `CommitmentOutput` are not actual transaction outputs. Perhaps the name `CommitmentOutput` is misleading? If yes i think we should rename it to `CommitmentSpecOutput`

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -206,39 +206,100 @@ object Transactions {
...
+                        localRevocationPubkey: PublicKey,
+                        toLocalDelay: CltvExpiryDelta,
+                        localDelayedPaymentPubkey: PublicKey,
+                        feeratePerKw: Long): HtlcTimeoutTx = {
     val fee = weight2fee(feeratePerKw, htlcTimeoutWeight)
-    val redeemScript = htlcOffered(localHtlcPubkey, remoteHtlcPubkey, localRevocationPubkey, ripemd160(htlc.paymentHash.bytes))
-    val pubkeyScript = write(pay2wsh(redeemScript))
-    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(htlc.amountMsat.truncateToSatoshi))
+    val redeemScript = output.redeemScript
+    val DirectedHtlc(OUT, htlc) = output.specItem
```
pm47: That's a hidden cast, which could throw at runtime.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
pm47: This plus https://github.com/ACINQ/eclair/pull/790#discussion_r398688368 makes me think that the model has a problem.

araspitzu: I see the contradiction and it's supposed to be mitigated by the fact that we declare `CommitmentOutput` very close to `CommitmentSpec` but this is clearly not enough. How about we rename it `CommitmentSpecOutput`? This name makes it clear is something that lives in our  `CommitmentSpec` and the reader should understand this is not mapping 1:1 with the outputs in the commitment transaction.

pm47: Okay I've tried to invert the relationship between `CommitmentOutput` and `DirectedHtlc` and it seems to fix the issue. As a bonus we are now strongly typed which helps a bit with https://github.com/ACINQ/eclair/pull/790#discussion_r398688368 (not completely, we could go even further by typing the direction).

See 8bdec0e43b5799099c7a719e3a0db3a662e489a6.

t-bast: I like https://github.com/ACINQ/eclair/commit/8bdec0e43b5799099c7a719e3a0db3a662e489a6, the additional typing is really nice, removes a few casts.
I'll do some tests on the variance issue.

araspitzu:
Following yesterday's conversation and hoping to get everyone on board i'm proposing to type `DirectedHtlc` in 748acd4. The changes introduce a small hierarchy of types where `DirectedHtlc` is extended by 2 subclasses, each for IN/OUT direction. There is still the duality in names `Outgoing/Offered HTLC` and `Incoming/Received HTLC`, I started using `Outgoing/Incoming` but we should make the naming more consistent across the codebase (and now it's a good time to make a choice).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -614,8 +614,8 @@ object Helpers {
 
       // remember we are looking at the remote commitment so IN for them is really OUT for us and vice versa
       val txes = remoteCommit.spec.htlcs.collect {
-        // incoming htlc for which we have the preimage: we spend it directly
-        case DirectedHtlc(OUT, add: UpdateAddHtlc) if preimages.exists(r => sha256(r) == add.paymentHash) => generateTx("claim-htlc-success")(Try {
+        // outgoing htlc for which we have the preimage: we spend it directly
```
t-bast: This is really an incoming HTLC (because it's outgoing in the *remote*'s commit). Yeah this is confusing :)
Maybe the comment could be updated to:
```scala
// incoming htlc for which we have the preimage: we spend it directly.
// NB: we are looking at the remote's commitment, from its point of view it's an outgoing htlc.
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -624,8 +624,8 @@ object Helpers {
 
         // (incoming htlc for which we don't have the preimage: nothing to do, it will timeout eventually and they will get their funds back)
 
-        // outgoing htlc: they may or may not have the preimage, the only thing to do is try to get back our funds after timeout
-        case DirectedHtlc(IN, add: UpdateAddHtlc) => generateTx("claim-htlc-timeout")(Try {
+        // incoming htlc: they may or may not have the preimage, the only thing to do is try to get back our funds after timeout
```
t-bast: Same as above.

pm47:
> Following yesterday's conversation and hoping to get everyone on board i'm proposing to type DirectedHtlc in 748acd4. The changes introduce a small hierarchy of types where DirectedHtlc is extended by 2 subclasses, each for IN/OUT direction.

We are going in the right direction (pun intended), but the `Direction` trait should not be needed anymore.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -33,8 +33,21 @@ sealed trait CommitmentOutput
 object CommitmentOutput {
   case object ToLocal extends CommitmentOutput
   case object ToRemote extends CommitmentOutput
+  case class InHtlc(incomingHtlc: IncomingHtlc) extends CommitmentOutput
+  case class OutHtlc(outgoingHtlc: OutgoingHtlc) extends CommitmentOutput
 }
-case class DirectedHtlc(direction: Direction, add: UpdateAddHtlc) extends CommitmentOutput
+
+sealed trait DirectedHtlc {
+  def direction: Direction
```
t-bast: Do we really need to keep `Direction`? It seems like we can go further and remove that entirely, and rely on pattern matching only.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -33,8 +33,21 @@ sealed trait CommitmentOutput
 object CommitmentOutput {
   case object ToLocal extends CommitmentOutput
   case object ToRemote extends CommitmentOutput
+  case class InHtlc(incomingHtlc: IncomingHtlc) extends CommitmentOutput
```
t-bast: If we only keep one `Htlc(htlc: DirectedHtlc) extends CommitmentOutput` the pattern matching that annoyed us yesterday doesn't work properly?

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -123,21 +124,21 @@ object Transactions {
   /** Offered HTLCs below this amount will be trimmed. */
   def offeredHtlcTrimThreshold(dustLimit: Satoshi, spec: CommitmentSpec): Satoshi = dustLimit + weight2fee(spec.feeratePerKw, htlcTimeoutWeight)
 
-  def trimOfferedHtlcs(dustLimit: Satoshi, spec: CommitmentSpec): Seq[DirectedHtlc] = {
+  def trimOfferedHtlcs(dustLimit: Satoshi, spec: CommitmentSpec): Seq[OutgoingHtlc] = {
     val threshold = offeredHtlcTrimThreshold(dustLimit, spec)
     spec.htlcs
-      .filter(_.direction == OUT)
+      .collect { case o:OutgoingHtlc => o }
```
t-bast: nit: you could merge the `filter` with the `collect`

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -123,21 +124,21 @@ object Transactions {
...
 
   /** Received HTLCs below this amount will be trimmed. */
   def receivedHtlcTrimThreshold(dustLimit: Satoshi, spec: CommitmentSpec): Satoshi = dustLimit + weight2fee(spec.feeratePerKw, htlcSuccessWeight)
 
-  def trimReceivedHtlcs(dustLimit: Satoshi, spec: CommitmentSpec): Seq[DirectedHtlc] = {
+  def trimReceivedHtlcs(dustLimit: Satoshi, spec: CommitmentSpec): Seq[IncomingHtlc] = {
     val threshold = receivedHtlcTrimThreshold(dustLimit, spec)
     spec.htlcs
-      .filter(_.direction == IN)
+      .collect { case i:IncomingHtlc => i }
```
t-bast: same

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -211,21 +212,21 @@ object Transactions {
 
    * @param output transaction output
    * @param redeemScript redeem script that matches this output (most of them are p2wsh)
-   * @param specItem commitment spec item this output is built from
+   * @param commitmentOutput commitment spec item this output is built from
    */
-  case class CommitmentOutputLink(output: TxOut, redeemScript: Seq[ScriptElt], specItem: CommitmentOutput)
+  case class CommitmentOutputLink[+T <: CommitmentOutput](output: TxOut, redeemScript: Seq[ScriptElt], commitmentOutput: T)
```
t-bast: Do you really need the `+T`? Can't it only be `T`?

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
t-bast: Or maybe `Htlc[T <: DirectedHtlc](htlc: T)`?

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -50,7 +63,10 @@ object CommitmentSpec {
   }
 
   def addHtlc(spec: CommitmentSpec, direction: Direction, update: UpdateAddHtlc): CommitmentSpec = {
```
t-bast: Maybe we can remove direction from here and directly receive a `DirectedHtlc`?

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
pm47: > Or maybe Htlc[T <: DirectedHtlc](htlc: T)?

I had that in mind but what @araspitzu went with is simpler and will cause less typing headaches imo.

araspitzu:
How do we store a `DirectedHtlc` without the direction? When we encode with scodecs we must encode the direction too (or the type) because we need to remember it when decoding.

t-bast:
> How do we store a DirectedHtlc without the direction? When we encode with scodecs we must encode the direction too (or the type) because we need to remember it when decoding.

If it's only needed in the codecs, that's an enum we can put closer to the codec code? Or we keep that enum here and use it only for the codec.

araspitzu:
Last update only addresses the PR feedback, i will push another commit for the removal of `Direction`

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
araspitzu: We can have a type parameter like `[T <: DirectedHtlc]` that would be used in the pattern matching, the caller can specify the direction as a type: `addHtlc[IncomingHtlc](spec, update)`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -610,23 +613,21 @@ object Helpers {
       val preimages = commitments.localChanges.all.collect { case u: UpdateFulfillHtlc => u.paymentPreimage }
 
       // remember we are looking at the remote commitment so IN for them is really OUT for us and vice versa
-      var outputsAlreadyUsed = Set.empty[Int] // this is needed to handle cases where we have several identical htlcs
       val txes = remoteCommit.spec.htlcs.collect {
-        // incoming htlc for which we have the preimage: we spend it directly
-        case DirectedHtlc(OUT, add: UpdateAddHtlc) if preimages.exists(r => sha256(r) == add.paymentHash) => generateTx("claim-htlc-success")(Try {
+        // incoming htlc for which we have the preimage: we spend it directly.
+        // NB: we are looking at the remote's commitment, from its point of view it's an outgoing htlc.
+        case IncomingHtlc(add: UpdateAddHtlc) if preimages.exists(r => sha256(r) == add.paymentHash) => generateTx("claim-htlc-success")(Try {
```
t-bast: This should stay an `OutgoingHtlc`, since we're in the remote commit

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -610,23 +613,21 @@ object Helpers {
...
           Transactions.addSigs(txinfo, sig, preimage)
         })
 
         // (incoming htlc for which we don't have the preimage: nothing to do, it will timeout eventually and they will get their funds back)
 
         // outgoing htlc: they may or may not have the preimage, the only thing to do is try to get back our funds after timeout
-        case DirectedHtlc(IN, add: UpdateAddHtlc) => generateTx("claim-htlc-timeout")(Try {
-          val txinfo = Transactions.makeClaimHtlcTimeoutTx(remoteCommitTx.tx, outputsAlreadyUsed, localParams.dustLimit, localHtlcPubkey, remoteHtlcPubkey, remoteRevocationPubkey, localParams.defaultFinalScriptPubKey, add, feeratePerKwHtlc)
-          outputsAlreadyUsed = outputsAlreadyUsed + txinfo.input.outPoint.index.toInt
+        case OutgoingHtlc(add: UpdateAddHtlc) => generateTx("claim-htlc-timeout")(Try {
```
t-bast: And this one should stay an `IncomingHtlc`, since we're in the remote commit

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
t-bast: If the caller is ready to specify `IncomingHtlc` vs `OutgoingHtlc`, I think it's simpler that he creates that directly, isn't it?
For example `addHtlc(spec, IncomingHtlc(update))`. Adding a type parameter here seems unnecessary.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -92,9 +93,14 @@ object ChannelCodecs extends Logging {
     (wire: BitVector) => bool.decode(wire).map(_.map(b => if (b) IN else OUT))
   )
 
+  import shapeless.::
   val htlcCodec: Codec[DirectedHtlc] = (
     ("direction" | directionCodec) ::
-      ("add" | updateAddHtlcCodec)).as[DirectedHtlc]
+      ("add" | updateAddHtlcCodec)).xmap(
```
t-bast: Can't you get the result of the `directionCodec` here? I think we have some codecs that do that, I'll look around.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
t-bast: Alright let's keep it like that then, it's true that it's simpler.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: Told you it had the potential to be rabbit hole-y 🤣

araspitzu:
Can we postpone the removal of `Direction` to another PR? It's a pretty big refactoring and in many places is not trivial to replace the explicit use of the direction with pattern matching (and we end up using manifest/classtag). I will address the latest feedback and we could merge this PR as-is and without going too much out of scope.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
t-bast: Let's do this in a follow-up PR.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
t-bast: Actually your proposal works fine, I like it. I just find the indentation a bit confusing, what about:

```scala
val htlcCodec: Codec[DirectedHtlc] = (("direction" | directionCodec) :: ("add" | updateAddHtlcCodec)).xmap(
    {
      case IN :: add :: HNil => IncomingHtlc(add)
      case OUT :: add :: HNil => OutgoingHtlc(add)
    },
    {
      htlc => htlc.direction :: htlc.add :: HNil
    }
  )
```

And we should put the `import shapeless.::` at the top of the file IMHO, not a big fan of scattering imports in the file.

araspitzu: Will do, i also don't like the scattered imports but i'd make an exception for `shapeless.::` if imported at the top of the file it's very hard to understand what `::` does if it's from scala's collection or shapeless.

araspitzu:
> LGTM, thanks for this long-standing work!
If you don't have time in the next few days, I can work on the leftover refactoring we mentioned.

I think i can work on the refactoring, by the way before merging this PR i'd like to do a final E2E test.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -268,25 +338,34 @@ object Transactions {
...
 
-  def makeClaimHtlcSuccessTx(commitTx: Transaction, outputsAlreadyUsed: Set[Int], localDustLimit: Satoshi, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, remoteRevocationPubkey: PublicKey, localFinalScriptPubKey: ByteVector, htlc: UpdateAddHtlc, feeratePerKw: Long): ClaimHtlcSuccessTx = {
+  def makeClaimHtlcSuccessTx(commitTx: Transaction, outputs: CommitmentOutputs, localDustLimit: Satoshi, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, remoteRevocationPubkey: PublicKey, localFinalScriptPubKey: ByteVector, htlc: UpdateAddHtlc, feeratePerKw: Long): ClaimHtlcSuccessTx = {
     val redeemScript = htlcOffered(remoteHtlcPubkey, localHtlcPubkey, remoteRevocationPubkey, ripemd160(htlc.paymentHash.bytes))
-    val pubkeyScript = write(pay2wsh(redeemScript))
-    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(htlc.amountMsat.truncateToSatoshi))
+    val Some(outputIndex) = outputs.zipWithIndex.find {
+      case (CommitmentOutputLink(_, _, OutHtlc(OutgoingHtlc(outgoingHtlc))), _) => outgoingHtlc.id == htlc.id
+      case _ => false
+    }.map(_._2)
```
pm47:
```suggestion
val Some(outputIndex) = outputs.zipWithIndex.collectFirst {
      case (CommitmentOutputLink(_, _, OutHtlc(OutgoingHtlc(outgoingHtlc))), outputIndex) if outgoingHtlc.id == htlc.id => outputIndex
    }
```

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -306,10 +385,13 @@ object Transactions {
...
 
-  def makeClaimHtlcTimeoutTx(commitTx: Transaction, outputsAlreadyUsed: Set[Int], localDustLimit: Satoshi, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, remoteRevocationPubkey: PublicKey, localFinalScriptPubKey: ByteVector, htlc: UpdateAddHtlc, feeratePerKw: Long): ClaimHtlcTimeoutTx = {
+  def makeClaimHtlcTimeoutTx(commitTx: Transaction, outputs: CommitmentOutputs, localDustLimit: Satoshi, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, remoteRevocationPubkey: PublicKey, localFinalScriptPubKey: ByteVector, htlc: UpdateAddHtlc, feeratePerKw: Long): ClaimHtlcTimeoutTx = {
     val redeemScript = htlcReceived(remoteHtlcPubkey, localHtlcPubkey, remoteRevocationPubkey, ripemd160(htlc.paymentHash.bytes), htlc.cltvExpiry)
-    val pubkeyScript = write(pay2wsh(redeemScript))
-    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(htlc.amountMsat.truncateToSatoshi))
+    val Some(outputIndex) = outputs.zipWithIndex.find {
+      case (CommitmentOutputLink(_, _, InHtlc(IncomingHtlc(incomingHtlc))), _) => incomingHtlc.id == htlc.id
+      case _ => false
+    }.map(_._2)
```
pm47:
```suggestion
val Some(outputIndex) = outputs.zipWithIndex.collectFirst {
      case (CommitmentOutputLink(_, _, InHtlc(IncomingHtlc(incomingHtlc))), outputIndex) if incomingHtlc.id == htlc.id => outputIndex
    }
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -92,9 +93,15 @@ object ChannelCodecs extends Logging {
...
-      ("add" | updateAddHtlcCodec)).as[DirectedHtlc]
+  val htlcCodec: Codec[DirectedHtlc] = (("direction" | directionCodec) :: ("add" | updateAddHtlcCodec)).xmap(
+    {
+      case IN :: add :: HNil => IncomingHtlc(add)
+      case OUT :: add :: HNil => OutgoingHtlc(add)
+    },
+    {
+      htlc => htlc.direction :: htlc.add :: HNil
+    }
+  )
```
pm47:
```suggestion
val htlcCodec: Codec[DirectedHtlc] = discriminated[DirectedHtlc].by(directionCodec)
    .typecase(IN, updateAddHtlcCodec.as[IncomingHtlc])
    .typecase(OUT, updateAddHtlcCodec.as[OutgoingHtlc])
```

pm47: Also, any changes to those codecs **mandate** dedicated non-reg tests.

araspitzu:
At 60ba2e4 i applied the latest feedback by @pm47 (including the non-reg test). I also tested the compatibility with CL and it went okay (i compared the commitment transactions after adding HTLCs with same amount and preimage but different CLTV), i didn't test with LND master yet as there is a known issue that is being fixed.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
```diff
@@ -137,6 +137,18 @@ class ChannelCodecsSpec extends FunSuite {
   }
 
   test("encode/decode htlc") {
+    // these encoded HTLC were produced by a previous version of the codec (at commit 8932785e001ddfe32839b3f83468ea19cf00b289)
```
t-bast: Can you extract that (backwards-compatibility check) to a separate test?

araspitzu: done

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -361,10 +361,9 @@ object Transactions {
 
   def makeClaimHtlcSuccessTx(commitTx: Transaction, outputs: CommitmentOutputs, localDustLimit: Satoshi, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, remoteRevocationPubkey: PublicKey, localFinalScriptPubKey: ByteVector, htlc: UpdateAddHtlc, feeratePerKw: Long): ClaimHtlcSuccessTx = {
     val redeemScript = htlcOffered(remoteHtlcPubkey, localHtlcPubkey, remoteRevocationPubkey, ripemd160(htlc.paymentHash.bytes))
-    val Some(outputIndex) = outputs.zipWithIndex.find {
-      case (CommitmentOutputLink(_, _, OutHtlc(OutgoingHtlc(outgoingHtlc))), _) => outgoingHtlc.id == htlc.id
-      case _ => false
-    }.map(_._2)
+    val Some(outputIndex) = outputs.zipWithIndex.collectFirst {
```
pm47: Just realized that this is another "hidden cast". It may be ok here, but we need to be really careful with this pattern, this is equivalent to doing an `Option.get`, and breaks the functional model by throwing exceptions.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
pm47: +1, and also clearly label it as a non-reg test

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
t-bast: I also raised this in a previous comment :).
But as a matter of fact the previous code also threw an exception here, hidden in the `findHtlc...` method. And the upper layer does a `Try` so at least we're not regressing.
But we should definitely update that function to return an `Option` in a later PR.

araspitzu: Also note that the current code used to get the output index (`findPubKeyScriptIndex`) throws an exception as well.

pm47: True, but at least `findPubKeyScriptIndex` throws an explicit `OutputNotFound`. I hesitated with proposing to throw this same exception here, but it's not *exactly* the same thing.

araspitzu: Why is it not the same thing? Both method try to get the outputIndex searching the `outputs` array, if none can be found an exception will be thrown. We can replace the extractor pattern with a `.getOrElse(throw OutputNotFound)` to make it more consistent with the previous behavior. WDYT?

t-bast: I think it would be a better idea to keep it like that and change this function to return an `Option` in the refactoring PR we talked about.

pm47: > Why is it not the same thing?

It's not the same lookup

> I think it would be a better idea to keep it like that

Agreed, that's why I acked

PR #791, opened 2018-12-22
Take channel age and capacity into account when calculating an edge weight

btcontract:
`ChannelUpdate` contains useful info such as age in blocks (taken from `shortChannelId`) and capacity (taken from `htlcMaximumMsat`) which can improve an edge weight calculation in a graph.

It is reasonable to think that older channels have caused less problems before, did not get force-closed and as such are better than newer channels all other things being equal. It is also reasonable to think that channels of larger capacity are more serious at being good routers than smaller ones.


pm47:
This PR is superseded by #821

PR #792, opened 2018-12-25
Fire together wire together

btcontract:
This PR builds on https://github.com/ACINQ/eclair/pull/791 and adds yet another edge weight factor which is past channel's performance. 

Since routing failure can't be reliably punished we at least should endorse a success. The idea therefore is: if this channel has participated in recent successful payments then we can expect it to keep it that way and decrease it's weight in a graph. OTOH if this channel was not participating in recent successful payments (either by being excluded from graph because of failures or just by not being used) then we gradually forget about it's successful history and success factor stops playing a role in weight calculation.

Success factor is updated by exponentially decaying function which has desired properties: converges to either top `DEFAULT_SUCCESS_FACTOR` or bottom `0` depending on direction, substantially changes a factor at start (the very first successful payment already tells us a lot about channel reliability) and slows down with subsequent payments (20th successful payment does not tell us as much as 1st one).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
@@ -26,10 +26,7 @@ case class ShortChannelId(private val id: Long) extends Ordered[ShortChannelId] 
 
   def toLong: Long = id
 
-  override def toString: String = {
-    val TxCoordinates(blockHeight, txIndex, outputIndex) = ShortChannelId.coordinates(this)
-    s"${blockHeight}x${txIndex}x${outputIndex}"
-  }
+  override def toString: String = s"${txCoordinates.blockHeight}x${txCoordinates.outputIndex}x${txCoordinates.outputIndex}"
 
   // we use an unsigned long comparison here
   override def compare(that: ShortChannelId): Int = (this.id + Long.MinValue).compareTo(that.id + Long.MinValue)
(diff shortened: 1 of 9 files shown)


araspitzu:
Thanks for the PRs, I think those kind of heuristics can be useful and lead to better routes, however 2 points: 

- Are there any results that can prove these heuristics? In particular it seems to me that prioritizing newer channels over old ones doesn't bring much to the table (there might be old channels that are very reliable) if not a little incentive toward closing and reopening a channel just to catch some routing fees.  
- There are other changes incoming in the routing so those will have to be rebased on top of them

btcontract:
> Are there any results that can prove these heuristics?

Very informal ones. I basically port what is already being used in my software because user reports are positive.

> In particular it seems to me that prioritizing newer channels over old ones doesn't bring much to the table

It's the other way around: older channels have higher priority as defined here: https://github.com/ACINQ/eclair/blob/93bae977c1a142de84f2e11a057707973ea7652b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala#L312

Newer blocks have higher number, hence more weight and less chance to be selected. The whole `edgeWeightByAmountAgeCapacity` is a bit counter-intuitive since I've decided to stick to summation and multiplication only because this method will be called very often while traversing a graph.

> There are other changes incoming in the routing so those will have to be rebased on top of them

sure, I'll update once needed.

araspitzu:
Please rebase on 7f3b101

araspitzu:
Please rebase again on fd3ff91 I'm looking forward into integrating this into the route selection but there will be some changes, for example we want to keep in account the total CLTV of a route as well. Also storing the 'successFactors' in a map on the DirectedGraph object is not ideal, opens up for concurrency issues and is not persisted which means every time the eclair-mobile starts up the success factors will be empty but the tradeoff of storing a [Long, Long] map in the db network db might be worth the gain.

btcontract:
I'm sorry too busy with other stuff currently, please take it from here or I'll get back to this later.

PR #795, opened 2018-12-28
Limit CLTV value to 2016

araspitzu:
As per the spec 1.1 we should not offer a to_self_delay value greater than 2016, nor accept a channel that offers such a high value.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -163,6 +163,10 @@ object NodeParams {
     val maxAcceptedHtlcs = config.getInt("max-accepted-htlcs")
     require(maxAcceptedHtlcs <= Channel.MAX_ACCEPTED_HTLCS, s"max-accepted-htlcs must be lower than ${Channel.MAX_ACCEPTED_HTLCS}")
 
+    val maxToLocalCLTV = config.getInt("max-to-local-delay-blocks")
+    val offeredCLTV = config.getInt("to-remote-delay-blocks")
+    require(maxToLocalCLTV < 2016 && offeredCLTV < 2016, "CLTV delay values too high, max is 2016")
+
     val overrideFeatures: Map[PublicKey, (BinaryData, BinaryData)] = config.getConfigList("override-features").map { e =>
       val p = PublicKey(e.getString("nodeid"))
       val gf = BinaryData(e.getString("global-features"))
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -163,6 +163,10 @@ object NodeParams {
     val maxAcceptedHtlcs = config.getInt("max-accepted-htlcs")
     require(maxAcceptedHtlcs <= Channel.MAX_ACCEPTED_HTLCS, s"max-accepted-htlcs must be lower than ${Channel.MAX_ACCEPTED_HTLCS}")
 
+    val maxToLocalCLTV = config.getInt("max-to-local-delay-blocks")
+    val offeredCLTV = config.getInt("to-remote-delay-blocks")
+    require(maxToLocalCLTV <= 2016 && offeredCLTV <= 2016, "CLTV delay values too high, max is 2016")
```
pm47: The max value should be defined in the `Channel` object, just like e.g. `Channel.MAX_ACCEPTED_HTLCS`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -62,6 +62,9 @@ object Channel {
   val MIN_CLTV_EXPIRY = 9L
   val MAX_CLTV_EXPIRY = 7 * 144L // one week
 
+  //  since BOLTv1.1
+  val MAX_MAIN_CLTV = 2016L
```
```suggestion
  val MAX_TO_SELF_DELAY = 2016L
```

This is the "official" name of this parameter in the BOLTs, I've never seen a reference to a main cltv.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -134,8 +134,8 @@ object Helpers {
     if (accept.dustLimitSatoshis > accept.channelReserveSatoshis) throw DustLimitTooLarge(accept.temporaryChannelId, accept.dustLimitSatoshis, accept.channelReserveSatoshis)
 
     // if minimum_depth is unreasonably large:
-    // MAY reject the channel.
-    if (accept.toSelfDelay > nodeParams.maxToLocalDelayBlocks) throw ToSelfDelayTooHigh(accept.temporaryChannelId, accept.toSelfDelay, nodeParams.maxToLocalDelayBlocks)
+    // MAY reject the channel. max CLTV is 2016
```
pm47: same bikeshed comment as above
```suggestion
    // MAY reject the channel.
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Also, `toSelfDelay` should be an `Int`, not a `Long`

PR #797, opened 2019-01-02
Compute routes backward

araspitzu:
Compute the routes backward from the target and with the correct channels cost, nodes that are 'neighbor' to the local node (direct channel) will have a routing cost of 0.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -36,14 +36,11 @@ object Graph {
     * @param extraEdges a list of extra edges we want to consider but are not currently in the graph
     * @return
     */
-  def shortestPath(g: DirectedGraph, sourceNode: PublicKey, targetNode: PublicKey, amountMsat: Long, ignoredEdges: Set[ChannelDesc], extraEdges: Set[GraphEdge], reverse: Boolean = false): Seq[Hop] = {
-    dijkstraShortestPath(g, sourceNode, targetNode, amountMsat, ignoredEdges, extraEdges, reverse).map(edge => reverse match {
-      case false => Hop(edge.desc.a, edge.desc.b, edge.update)
-      case true  => Hop(edge.desc.b, edge.desc.a, edge.update)
-    })
+  def shortestPath(g: DirectedGraph, sourceNode: PublicKey, targetNode: PublicKey, amountMsat: Long, ignoredEdges: Set[ChannelDesc], extraEdges: Set[GraphEdge]): Seq[Hop] = {
+    dijkstraShortestPath(g, sourceNode, targetNode, amountMsat, ignoredEdges, extraEdges).map(edge => Hop(edge.desc.b, edge.desc.a, edge.update))
   }
 
-  def dijkstraShortestPath(g: DirectedGraph, sourceNode: PublicKey, targetNode: PublicKey, amountMsat: Long, ignoredEdges: Set[ChannelDesc], extraEdges: Set[GraphEdge], reverse: Boolean = false): Seq[GraphEdge] = {
+  def dijkstraShortestPath(g: DirectedGraph, sourceNode: PublicKey, targetNode: PublicKey, amountMsat: Long, ignoredEdges: Set[ChannelDesc], extraEdges: Set[GraphEdge]): Seq[GraphEdge] = {
 
     // optionally add the extra edges to the graph
     val graphVerticesWithExtra = extraEdges.nonEmpty match {
@@ -95,8 +92,7 @@ object Graph {
             // note: the default value here will never be used, as there is always an entry for the current in the 'cost' map
             // note: when searching on a reversed graph we consider 0 the weight of the edge of the target-neighboring node
             // this is because it will be a direct channel to the source and it pays no fees
-            val isDirectChannel = if(reverse) neighbor == targetNode else neighbor == sourceNode
-            val newMinimumKnownCost = cost.get(current.key) + edgeWeightByAmount(edge, amountMsat, isNeighborTarget = isDirectChannel)
+            val newMinimumKnownCost = cost.get(current.key) + edgeWeightByAmount(edge, amountMsat, neighbor == targetNode)
 
             // we call containsKey first because "getOrDefault" is not available in JDK7
             val neighborCost = cost.containsKey(neighbor) match {
@@ -137,7 +133,7 @@ object Graph {
         }
 
         // if we are searching "backward" the resulting edgePath is already reversed
-        if(!reverse) edgePath.reverse else edgePath
+        edgePath
       }
     }
   }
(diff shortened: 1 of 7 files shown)


araspitzu:
I chose to have an explicit parameter on the graph construction and manipulation functions, alternatively we can have a separate function with reverse in the name i.e "addEdgeReverse", please let me know what do you think of it.

sstone:
Isn't it possible to use the same graph (i.e. not need for the additional `reverse` parameter) and just compute the route backwards internally ? It should be just an implementation choice that is not even exposed in the API ?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -791,7 +791,7 @@ object Router {
   def findRoute(g: DirectedGraph, localNodeId: PublicKey, targetNodeId: PublicKey, amountMsat: Long, extraEdges: Set[GraphEdge] = Set.empty, ignoredEdges: Set[ChannelDesc] = Set.empty): Try[Seq[Hop]] = Try {
     if (localNodeId == targetNodeId) throw CannotRouteToSelf
 
-    Graph.shortestPath(g, localNodeId, targetNodeId, amountMsat, ignoredEdges, extraEdges) match {
+    Graph.shortestPath(g, targetNodeId, localNodeId, amountMsat, ignoredEdges.map(reverseDesc), extraEdges.map(edge => edge.copy(desc = reverseDesc(edge.desc)))) match {
```
sstone: I think we should keep the method and parameters as they were, we are still computing a shortest path from the source to the destination, it's just that internally we compute it backwards

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -90,7 +90,9 @@ object Graph {
             val neighbor = edge.desc.b
 
             // note: the default value here will never be used, as there is always an entry for the current in the 'cost' map
-            val newMinimumKnownCost = cost.get(current.key) + edgeWeightByAmount(edge, amountMsat)
+            // note: when searching on a reversed graph we consider 0 the weight of the edge of the target-neighboring node
+            // this is because it will be a direct channel to the source and it pays no fees
+            val newMinimumKnownCost = cost.get(current.key) + edgeWeightByAmount(edge, amountMsat, neighbor == targetNode)
```
sstone: I don't think this is right: the fees that are applied are the fees for the outgoing edge, and which is then added to the amount that we want to forward. Suppose we have:

A
    \
     C -- D
  /
B
And want to forward X from A to D.
First we look for incoming edges of D that can carry X. We have one: edge_cd
Then we move backwards and look for incoming edges of C that can carry amount + cost(edge_cd, amount) and so on: the amount that is used to compute the relay fee is updated at each hop.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -81,16 +81,17 @@ object Graph {
...
 
-            val neighbor = edge.desc.b
+          // note: 'cost' contains the smallest known cumulative cost (amount + fees) necessary to reach 'current' so far
+          // note: there is always an entry for the current in the 'cost' map
+          val newMinimumKnownCost = edgeWeight(edge, cost.get(current.key), neighbor == targetNode)
 
-            // note: the default value here will never be used, as there is always an entry for the current in the 'cost' map
-            val newMinimumKnownCost = cost.get(current.key) + edgeWeightByAmount(edge, amountMsat)
+          // test for ignored edges
+          if (!(edge.update.htlcMaximumMsat.exists(_ < newMinimumKnownCost) ||
```
pm47: is that the most optimal order for the testing conditions? Given that the first to return `true` will return immediately

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -130,17 +131,25 @@ object Graph {
...
-  private def edgeWeightByAmount(edge: GraphEdge, amountMsat: Long): Long = {
-    nodeFee(edge.update.feeBaseMsat, edge.update.feeProportionalMillionths, amountMsat)
+  /**
+    *
+    * @param edge the edge for which we want to compute the weight
+    * @param amountWithFees the value that this edge will have to carry along
+    * @param isNeighborTarget true if the receiving vertex of this edge is the target node (source in a reversed graph), which has cost 0
+    * @return the new amount updated with the necessary fees for this edge
+    */
+  private def edgeWeight(edge: GraphEdge, amountWithFees: Long, isNeighborTarget: Boolean): Long = isNeighborTarget match {
```
pm47: would an `if` be more optimal here?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -161,35 +170,42 @@ object Graph {
...
+        * Edges are added as "reversed" swapping the starting vertex with the final.
         *
         * @param edge the edge that is going to be added to the graph
         * @return a new graph containing this edge
         */
       def addEdge(edge: GraphEdge): DirectedGraph = {
 
-        val vertexIn = edge.desc.a
-        val vertexOut = edge.desc.b
+        val finalEdge = edge.copy(desc = reverseDesc(edge.desc))
```
pm47: how significant is this extra copy during graph initialization on mobile?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -308,18 +335,22 @@ object Graph {
...
+            case None => mutableMap += someDesc.b -> List.empty[GraphEdge]
             case _ =>
           }
         }
 
         new DirectedGraph(mutableMap.toMap)
       }
 
-      def graphEdgeToHop(graphEdge: GraphEdge): Hop = Hop(graphEdge.desc.a, graphEdge.desc.b, graphEdge.update)
+      def reverseDesc(desc: ChannelDesc): ChannelDesc = desc.copy(a = desc.b, b = desc.a)
```
pm47: Nitpick: I would define this as a member function of `ChannelDesc`

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: Depends on which one you consider more readable, i've used 'match' for testing conditions and this is consistent to it.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: We don't call this method during the graph initialization, we use DirectedGraph.makeGraph...which also makes a copy of the edge desc

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: done

araspitzu:
wiki updated

Issue #773, opened 2018-12-05
checkinvoice API no longer returning useful data due to poor serialisation of PaymentRequest


n1bor:
When calling the checkinvoice API the returned JSON is missing most of the information. E.g. no nodeId or expiry. This is due to the PaymentRequest not being serialised correctly.

The below gets what I need. Also need to update Service.scala to include the new class on line 77

```
class PaymentRequestSerializer extends CustomSerializer[PaymentRequest](format => ({ null },{
  case p: PaymentRequest => JObject(JField("prefix",JString(p.prefix) ) :: 
      JField("amount",JLong(p.amount.getOrElse(MilliSatoshi(-1L)).toLong) ) ::
      JField("timestamp",JLong(p.timestamp) ) ::
      JField("nodeId",JString(p.nodeId.toString()) ) ::
      JField("description",JString(
          p.description match{ case Left(l) => l.toString()
                               case Right(r) => r.toString() } 
          ) ) ::
      
      JField("paymentHash",JString(p.paymentHash.toString()) ) ::
      JField("expiry",JLong(p.expiry.getOrElse(-1L)) ) ::
      JField("minFinalCltvExpiry",JLong(p.minFinalCltvExpiry.getOrElse(-1L)) ) ::
      Nil)
}))
```

araspitzu:
Correct, the API now doesn't return what it is supposed to. Will work out a fix based on your suggestions, thanks for the heads up!

PR #798, opened 2019-01-03
Add JSON serializer for PaymentRequest

n1bor:
Fixes issue #773


eclair-core/src/main/scala/fr/acinq/eclair/api/JsonSerializers.scala
```diff
@@ -132,3 +133,17 @@ class DirectionSerializer extends CustomSerializer[Direction](format => ({ null
...
+  case p: PaymentRequest => JObject(JField("prefix",JString(p.prefix) ) ::
+      JField("amount",JLong(p.amount.getOrElse(MilliSatoshi(-1L)).toLong) ) ::
+      JField("timestamp",JLong(p.timestamp) ) ::
+      JField("nodeId",JString(p.nodeId.toString()) ) ::
+      JField("description",JString(
+          p.description match{ case Left(l) => l.toString()
+                               case Right(r) => r.toString() }
+          ) ) ::
+      JField("paymentHash",JString(p.paymentHash.toString()) ) ::
+      JField("expiry",JLong(p.expiry.getOrElse(-1L)) ) ::
```
araspitzu: Perhaps it's better to use a JNull here? Seems more correct to return a json `null` Instead of a dummy -1 in the case where the value is missing.

pm47:
Can you also take the opportunity to rename `checkinvoice` to `parseinvoice`, which is what the method actually does? Current name makes one think it checks if the invoice has been paid, which is what `checkpayment` does.

eclair-core/src/main/scala/fr/acinq/eclair/api/JsonSerializers.scala
```diff
@@ -132,3 +133,17 @@ class DirectionSerializer extends CustomSerializer[Direction](format => ({ null
   case d: Direction => JString(d.toString)
 }))
 
+class PaymentRequestSerializer extends CustomSerializer[PaymentRequest](format => ({ null },{
+  case p: PaymentRequest => JObject(JField("prefix",JString(p.prefix) ) ::
+      JField("amount",JLong(p.amount.getOrElse(MilliSatoshi(-1L)).toLong) ) ::
```
araspitzu: Please use JNull here as well

n1bor: fixed...

eclair-core/src/main/scala/fr/acinq/eclair/api/JsonSerializers.scala
n1bor: Fixed

eclair-core/src/main/scala/fr/acinq/eclair/api/JsonSerializers.scala
```diff
@@ -132,3 +133,17 @@ class DirectionSerializer extends CustomSerializer[Direction](format => ({ null
   case d: Direction => JString(d.toString)
 }))
 
+class PaymentRequestSerializer extends CustomSerializer[PaymentRequest](format => ({ null },{
+  case p: PaymentRequest => JObject(JField("prefix",JString(p.prefix) ) ::
+      JField("amount",if(p.amount.isDefined) JLong(p.amount.get.toLong) else JNull ) ::
+      JField("timestamp",JLong(p.timestamp) ) ::
+      JField("nodeId",JString(p.nodeId.toString()) ) ::
+      JField("description",JString(
+          p.description match{ case Left(l) => l.toString()
```
sstone: nitpick: formatting issue

Issue #775, opened 2018-12-10
UI enhancement: Total balance


STAWKEYE:
This does not seem to be the same issue mentioned in #636 but I was thinking it would be a great feature if somewhere on the GUI there was a total balance (from within your channels). For example, each open channel already has "your balance" and so could the sum of those be placed on the status bar (where your node ID is) or in a subtotal line underneath the channels? If you want to keep the GUI clean I wouldnt' mind it buried in a menu on-demand, it's just a PITA to tally up 8 digit decimals manually when auditing.

Issue #690, opened 2018-09-06
GUI is confusing. Easy to accidentally close/force close the wrong channel


viaj3ro:
I accidentally closed the wrong channel more than once due to the blue lines making it look as if the close/force close button belongs to the channel above and not below.

It would help if the different channels are separated in a way such that there is no ambiguity about which channel belongs to which button.

STAWKEYE:
I agree completely here - but I do still like the visual of the "fill bars". As a suggestion what about keeping the bars but alternate shadings of the background similar to those alternating dark/light shades of green, blue, or grey on ledger style spreadsheets? (Like what you get with "format as table" in Excel for example)

YaroslavYrm:
IMO, easiest improvement is to move blue bar to the top of channel cell. In this way it will look like separator.

viaj3ro:
Just closed the wrong channel AGAIN :x 

This time it is really annoying since the accidentally closed channel was a very old, well funded, well used and well balanced one. Exactly the kind you REALLY want to keep open :(

Please fix this GUI issue. IMHO dropping the balance bar to the bottom would be the best but moving it to the top like @YaroslavYrm is suggesting would work as well.

anything but the status quo, basically...

dpad85:
Commit https://github.com/ACINQ/eclair/commit/2be2d38b952b4799428167ecf3a4cce60deb7169 adds a confirmation window when clicking on the `close` button. The confirmation dialog now contains some details about the channel as proposed by @viaj3ro.

The channel UI was also reworked, the channel's balance bar should not be confusing anymore: [preview](https://i.imgur.com/GkM6O3L.png).

viaj3ro:
looks good to me. Is the remote node ID and funder (yes/no) missing though? If yes, I'd suggest to re-add it. Otherwise it would be really difficult to not get lost once you open more than 2 or 3 channels.

dpad85:
The peer node id is displayed in bold characters at the top of the channel panel, I've switched its position with the channel id, it makes more sense that way. Good point with the funder boolean, I've put it back.

Issue #542, opened 2018-04-05
Decimal format for bits


rodasmith:
CoinUtils `formatAmountInUnit` uses a static `COIN_FORMAT` based on decimal pattern ` "###,###,###,##0.###########"` but if `unit: CoinUnit` is `BitUnit` then the format would ideally include at least two decimal places `("##,###,###,###,##0.00###")` [updated to range from 1 msat possible in LN to 21 million BTC] to reveal the underlying satoshi value and to help distinguish from satoshi format for round amounts like 0.10 bits, 1.00 bits, 1000.00 bits, etc.

https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/CoinUtils.scala#L197

rodasmith:
Since it's going to be dynamic based on unit option, consider also formatting BTC as `"##,###,##0.00000000###"` and mBTC as `"##,###,###,##0.00000###"`.

PR #801, opened 2019-01-03
GUI Improvements

dpad85:
This PR improves the channels panel layout. The balance progress bar is now smaller, and the balance of the channel and the peer node id are more visible. (#690)

Closing a channel now always prompts a confirmation dialog. This confirmation dialog contains details about the channel such as the peer node id, the channel's state, or the channel's current balance. (#690)

Amounts format is now dynamic and changes with the unit chosen by the user. (#542)

A total balance has been added to the status bar. (#775)

image


rorp:
Does it make sense to display the channel balances in satoshis, like it's done in CLI?

dpad85:
I'm not sure what you mean here. The balance in the GUI is displayed in whatever unit you prefer, be it milli-satoshis, satoshis, bits... You can set that in the conf.

The CLI does not support this feature, and amounts are displayed in a fixed unit.

Issue #799, opened 2019-01-03
Need time that payment was made available in API


n1bor:
checkpayment only returns True/False. Would instead like Payment object returned, which includes the amount (useful of overpayments/donations) and timestamp.

Will all PR soon...

t-bast:
This has been fixed with the `sentinfo` RPC addition.
Thanks for contributing ;)

PR #802, opened 2019-01-03
Add more details to `checkpayment` API call

n1bor:
Fixes issue #799


n1bor:
OK - works me.. is on the todo list!

PR #805, opened 2019-01-05
Allow access to audit of transactions from API

### LINKED ISSUE #615: Send API dangerous and timeout does not mean a payment failed.
Seen examples of payments hanging for 4.5 hours before failing:

> 05:18:39.879 INFO  akka://default/user/$j/payment-initiator/$4 f.a.eclair.payment.PaymentLifecycle - route found: attempt=2/5 route=02ad6fb8d693dc1e4569bcedefadf5f72a931ae027dc0f0c544b34c1c6f3b9a02b-02d539692f55b6d61847dc432bb78166bf52318e80e6b33733bb75f76645656e27 channels=7f96a0003370000->803ed0006250001

> 09:54:44.878 INFO  akka://default/user/$j/payment-initiator/$4 f.a.eclair.payment.PaymentLifecycle - received an error message from local, trying to use a different channel (failure=channel is unavailable (offline or closing))

So do we need another call to poll the API to check if timed out payments have now failed or succeeded?

And actually in the lightning spec once we try to send a payment we can only be sure it has failed once the htlc is rolled back in our commit. So staying open for hours/days is a possibility that could be common?

n1bor:
This would be an alternative to pr #802 as could get exact payment time using this.
Also can use to fix #615 as can check after a timeout if a payment went or not - just poll this till the invoice times out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,8 @@ java -Dlogback.configurationFile=/path/to/logback-custom.xml -jar eclair-node-gu
   forceclose   | channelId                                                                              | force-close a channel by publishing the local commitment tx (careful: this is more expensive than a regular close and will incur a delay before funds are spendable)"
   audit        |                                                                                        | list all send/received/relayed payments
   audit        | from, to                                                                               | list send/received/relayed payments in that interval (from <= timestamp < to)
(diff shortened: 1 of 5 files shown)


pm47:
Will merge #802 instead. Note that the audit database is made for, well, auditing purposes, and is not supposed to be queried during normal operations.

PR #806, opened 2019-01-05
WIP - Database Performance Monitoring

n1bor:
Initial version that just monitors Channels and Network db's.
Any comments on better ways todo this appreciated! As feels like a lot of copy/paste.

If not and you are happy to merge will add in similar classes for other database objects.

### DIFF
--- a/eclair-core/pom.xml
+++ b/eclair-core/pom.xml
@@ -211,6 +211,11 @@
             <artifactId>guava</artifactId>
             <version>${guava.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-core</artifactId>
+            <version>4.0.3</version>
+        </dependency>
         <!-- TESTS -->
         <dependency>
             <groupId>com.whisk</groupId>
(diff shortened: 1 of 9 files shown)


pm47:
Concept ack, but there is indeed a lot of repetition and there may be better options than codahale metrics out there (I used it in the peer-ratelimit branch, but it was just for prototyping).

We plan to work on monitoring starting february, and depending on the metrics library we end up choosing, this branch will then either require heavy refactoring or be deprecated. In the meantime if this fulfills your needs then I would suggest to manually merge that in your local repo.

n1bor:
This library seems fine. My only comment would be to add a LOT of monitoring. Even if it adds 10%+ CPU is worth it - as only has to help you spot one bit of bad code and you can save double that. Even if gets into the 1000's of datapoints think it is worth it.
Do work on trading systems doing 250k++ trades a day and just the database has 1000's of metrics we capture in real-time and historise to enable diagnostics of issues.

t-bast:
This PR is superceded by https://github.com/ACINQ/eclair/pull/1126 which provides database performance monitoring (not enabled yet by default though)

PR #809, opened 2019-01-07
Monitoring of Eventbus and Gauges on Channels/Nodes

n1bor:


n1bor:
@pm47  Been running this for 10 hrs on 280 channels:
https://pastebin.com/raw/vhvvJj3F

Seems to average almost 80 channel updates a second - which seems totally unnecessary.

Wondering why we need updates from all the peers? Surely if we just request updates from 8 that would be enough?

n1bor:
This is a perfect one to add as a plugin...

n1bor:
Add grafana stats from API if anyone interested:
https://lightningconductor.net/grafana/

t-bast:
This PR is superceded by https://github.com/ACINQ/eclair/pull/1126.
We will gradually add more metrics / spans as we test them under load.

PR #813, opened 2019-01-08
Compute the k-shortest routes

araspitzu:


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -776,24 +779,35 @@ object Router {
     */
   val ROUTE_MAX_LENGTH = 20
 
+  // The default amount of routes we'll search for when findRoute is called
+  val DEFAULT_ROUTES_AMOUNT = 3
```
pm47: nitpick: "amount" is misleading
```suggestion
  val DEFAULT_ROUTES_COUNT = 3
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -776,24 +779,35 @@ object Router {
...
 
   /**
-    * Find a route in the graph between localNodeId and targetNodeId, returns the route and its cost
+    * Find a route in the graph between localNodeId and targetNodeId, returns the route.
+    * Will perform a k-shortest path selection given the @param numRoutes and randomly select one of the result.
```
pm47: The fact that we also add a ceiling on cost on top of the base algo deserves a mention here.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -25,6 +24,109 @@ object Graph {
...
+      }
+    }
+
+    shortestPaths
+  }
+
+  // Calculates the cost of a path, direct channels with the source will have a cost of 0 (pay no fees), only the first
+  // edge in the list is a direct channel
+  def pathCost(path: Seq[GraphEdge], amountMsat: Long): Long = {
+    path.drop(1).reverse.foldLeft(amountMsat) { (cost, edge) =>
```
pm47: do we really need the `reverse`?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -25,6 +24,109 @@ object Graph {
...
+      edgeWeight(edge, cost, isNeighborTarget = false)
+    }
+  }
+
+  //helper function implementing the subList function for "Seq[T]" that will return the list with the
+  //first element if indices 0 are used
+  def subList[T](list: Seq[T], from: Int, to: Int): Seq[T] = {
```
pm47: Given that you only need `subList(l, 0, i)` in your code, you might as well use `l.take(i)`

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -593,10 +595,141 @@ class RouteCalculationSpec extends FunSuite {
...
+      makeUpdate(2L, b, c, feeBaseMsat = 2, 0),
+      makeUpdate(3L, c, d, feeBaseMsat = 3, 0),
+      makeUpdate(5L, e, f, feeBaseMsat = 3, 0),
+      makeUpdate(6L, f, d, feeBaseMsat = 3, 0),
+      makeUpdate(7L, e, c, feeBaseMsat = 1, 0)
+    ).toMap)
+
+    val Success(someRoute) = Router.findRoute(g, a, d, DEFAULT_AMOUNT_MSAT, numRoutes = 3)
+
+    val routeFees = pathCost(someRoute, DEFAULT_AMOUNT_MSAT) - DEFAULT_AMOUNT_MSAT
```
pm47: I would have rather computed e.g. 10 times the route and made sure that the result is always of cost 4 or 5.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
araspitzu: Done

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
araspitzu: Comment added

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: I think so, or are you suggesting to use 'foldRight' to avoid doing a reverse? (Just peeped into the scala implementations, foldRight calls reverse before applying foldLeft)

araspitzu: Done with foldRight

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: done

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
araspitzu: Done in 0cf1511

PR #820, opened 2019-01-15
[WIP] JDK10/11 compatiblity: use Scala 2.11.12

sstone:
With this change, you can:
- compile eclair-core and eclair-node on JDK10/11
- run eclair-node on JDK10/11

You cannot compile or run eclair-node-gui, because it uses JavaFX which is now packaged differently from JDK8.

See https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#jdk-9--up-compatibility-notes


sstone:
Superseded by #846

PR #821, opened 2019-01-15
Routing heuristics

araspitzu:
Add heuristics to the cost function of the path finding algorithm, and the ability to tune them by prioritizing for certain types. This enables the initiator of the payment (via `SendPayment`) to search for routes optimized for reliability or timeout where the reliability of a channel is obtained considering its **age** and **capacity**, with the assumption that older and larger channel are more reliable. Note that it is possible to specify the factor(s) by which we want consider each optimization, the `ratios: WeightRatios` parameter allows to specify a floating point value for each optimization heuristic, for example:  
 - `WeightRatios(cltvDeltaFactor = 0, ageFactor = 0.5, capacityFactor = 0.5)`  will consider equally the age and capacity of a channel but won't consider its CLTV delta.
 - `WeightRatios(cltvDeltaFactor = 0, ageFactor = 0, capacityFactor = 0)` will not consider heuristics optimization at all thus falling back to the previous behavior, cost optimized.
 - `WeightRatios(cltvDeltaFactor = 1, ageFactor = 0, capacityFactor = 0)` will find the route with the smallest possible **cltv**, thus minimizing the timeout in case of payment failure.

Optimization ratios can be specified at payment time or via the configuration file, by default they're all set to 0 in order to search for the route with the cheapest fees. Note that this can be combined with the existing bounded-search parameters such as the [fee-threshold-sat](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/resources/reference.conf#L104) to ask the router to find heuristics-optimized routes within the given constraints.


araspitzu:
This supersedes #792

araspitzu:
This is built on top of #844 and rebased to the last commit.

araspitzu:
These changes have no significant performance impact, below a benchmark on a routing table from 29/01/2019 (~23k channels).  
```
 Run complete. Total time: 00:07:24 (routing_heuristic: b1bf584) - Bounded search, no randomization, max-route-length=7, cost-optimized

Benchmark                                   Mode  Cnt  Score   Error  Units
GraphBenchmark.findPath                     avgt    2  0.001           s/op   
GraphBenchmark.findPathWithExtraChannels    avgt    2  0.004           s/op
GraphBenchmark.findPathWithIgnoredChannels  avgt    2  0.032           s/op
GraphBenchmark.routerLoadingTime            avgt    2  5.320           s/op
```

eclair-core/src/main/resources/reference.conf
```diff
@@ -103,6 +103,12 @@ eclair {
       max-cltv =  1008             // max acceptable cltv expiry for the payment (1008 ~ 1 week)
       fee-threshold-sat = 21       // if fee is below this value we skip the max-fee-pct check
       max-fee-pct = 0.03           // route will be discarded if fee is above this value (in percentage relative to the total payment amount); doesn't apply if fee < fee-threshold-sat
+
+      // channel 'weight' is computed with the following formula: channelFee * (cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity)
+      // the following parameters can be used to ask the router to use heuristics to find i.e: 'cltv-optimized' routes
+      ratio-cltv = 0.0             // when computing the weight for a channel, consider its CLTV delta in this proportion
```
sstone: We should document valid bounds for these ratio (between 0 and 1 and sum is 1 for example)

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -246,7 +246,10 @@ object NodeParams {
         searchMaxRouteLength = config.getInt("router.path-finding.max-route-length"),
         searchMaxCltv = config.getInt("router.path-finding.max-cltv"),
         searchMaxFeeBaseSat = config.getLong("router.path-finding.fee-threshold-sat"),
-        searchMaxFeePct = config.getDouble("router.path-finding.max-fee-pct")
+        searchMaxFeePct = config.getDouble("router.path-finding.max-fee-pct"),
+        searchRatioCltv = config.getDouble("router.path-finding.ratio-cltv"),
```
sstone: We should check that configured ratios are valid

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -46,6 +46,7 @@ import fr.acinq.eclair.channel.Register
 import fr.acinq.eclair.crypto.LocalKeyManager
 import fr.acinq.eclair.io.{Authenticator, Server, Switchboard}
 import fr.acinq.eclair.payment._
+import fr.acinq.eclair.router.Graph.WeightRatios
```
sstone: Is it still needed ?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -28,6 +28,7 @@ import fr.acinq.eclair.crypto.Sphinx.{ErrorPacket, Packet}
 import fr.acinq.eclair.crypto.{Sphinx, TransportHandler}
 import fr.acinq.eclair.payment.PaymentLifecycle._
 import fr.acinq.eclair.payment.PaymentRequest.ExtraHop
+import fr.acinq.eclair.router.Graph.WeightRatios
 import fr.acinq.eclair.router._
```
sstone: It seems that now this file not need to be changed

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -254,26 +273,70 @@ object Graph {
...
+
+    // Every edge is weighted by its clvt-delta value, normalized
+    val channelCltvDelta = edge.update.cltvExpiryDelta
+    val cltvFactor = normalize(channelCltvDelta, CLTV_LOW, CLTV_HIGH)
+
+    // NB 'edgeCost' includes the amount to be sent plus the fees that must be paid to traverse this @param edge
+    val edgeCost = if (isNeighborTarget) prev.cost else edgeFeeCost(edge, prev.cost)
+
+    val factor = (cltvFactor * wr.cltvDeltaFactor) + (ageFactor * wr.ageFactor) + (capFactor * wr.capacityFactor) match {
+      case 0 => 0.00001 // if the factor turns out to be 0 we default to a very small number to avoid having a weight of 0 and still take into account the cost
```
sstone: Why not use 1 instead ? this way we revert to exactly the same algorithm as before (i.e use edge costs)

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
araspitzu: Nope, removed in 5c94cef

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
araspitzu: Done in a0caf94

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: Done in 0acbc5d

eclair-core/src/main/resources/reference.conf
araspitzu: Done in 0acbc5d

PR #826, opened 2019-01-16
Upgrade to bitcoin 0.17.1

sstone:
Bitcoin Core 0.18 is about to enter RC cycle and should be release soon (initial target was April). It is not compatible with 0.16 (some of the RPC calls that we use have been removed. They're still available in 0.17 but tagged as deprecated). 

With this PR, eclair will be compatible with 0.17 and the upcoming 0.18, but not with 0.16 any more so it will be a breaking change for some of our users. Supporting the last 2 versions is the right option and we should be ready before 0.18 is actually released (its initial target was April).


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -49,6 +51,13 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
     rpcClient.invoke("signrawtransaction", hex).map(json => {
       val JString(hex) = json \ "hex"
       val JBool(complete) = json \ "complete"
+      if (!complete) {
+        val message = json \ "errors" match {
+          case value: JValue => Serialization.write(value)(DefaultFormats)
```
pm47: We should look in the code of bitcoin core what is normally returned in the `errors` field for `signrawtransaction` try to parse it in a best effort basis here, and return "signrawtransaction failed" if we failed.

eclair-core/src/test/resources/integration/bitcoin.conf
```diff
@@ -1,11 +1,13 @@
 regtest=1
+noprinttoconsole=1 
 server=1
 port=28333
-rpcport=28332
+regtest.rpcport=28332
```
pm47: looks weird to only have this field prefixed, can't we instead make a clean `regtest` section?

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -141,10 +141,10 @@ class BitcoinCoreWalletSpec extends TestKit(ActorSystem("test")) with BitcoindSe
     sender.send(bitcoincli, BitcoinReq("getrawtransaction", fundingTxes(2).txid.toString()))
     assert(sender.expectMsgType[JString](10 seconds).s === fundingTxes(2).toString())
 
-    // NB: bitcoin core doesn't clear the locks when a tx is published
+    // NB: bitcoin core before 0.17.0 doesn't clear the locks when a tx is published
     sender.send(bitcoincli, BitcoinReq("listlockunspent"))
-    assert(sender.expectMsgType[JValue](10 seconds).children.size === 2)
-
+    val expectedLocks = if (bitcoinVersion >= 170000) 0 else 2
```
pm47: We shouldn't have things like `if (bitcoinVersion >= 170000)`: our test runs against a certain version of bitcoin, backward compatibility isn't in scope here.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -177,7 +177,11 @@ class BitcoinCoreWalletSpec extends TestKit(ActorSystem("test")) with BitcoindSe
 
     val pubkeyScript = Script.write(Script.pay2wsh(Scripts.multiSig2of2(randomKey.publicKey, randomKey.publicKey)))
     wallet.makeFundingTx(pubkeyScript, MilliBtc(50), 10000).pipeTo(sender.ref)
-    assert(sender.expectMsgType[Failure].cause.asInstanceOf[JsonRPCError].error.message.contains("Please enter the wallet passphrase with walletpassphrase first."))
+    val error = sender.expectMsgType[Failure].cause.asInstanceOf[JsonRPCError].error
+    // behaviour has changed, bitcoin core will now return a more generic error
+    if (bitcoinVersion < 170000) {
```
pm47: same comment as above

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoindService.scala
```diff
@@ -83,7 +84,13 @@ trait BitcoindService extends Logging {
     logger.info(s"waiting for bitcoind to initialize...")
     awaitCond({
       sender.send(bitcoincli, BitcoinReq("getnetworkinfo"))
-      sender.receiveOne(5 second).isInstanceOf[JValue]
+      sender.receiveOne(5 second) match {
```
pm47: This can be removed.

eclair-core/src/test/resources/integration/bitcoin.conf
```diff
@@ -1,11 +1,13 @@
...
 rpcuser=foo
 rpcpassword=bar
 txindex=1
 zmqpubrawblock=tcp://127.0.0.1:28334
 zmqpubrawtx=tcp://127.0.0.1:28335
 rpcworkqueue=64
 addresstype=bech32
+deprecatedrpc=signrawtransaction
```
pm47: We should instead use the new method for signing transactions, and tell users to update their bitcoin core. Testing version 0.17.3 makes it the currently supported version, so we need to be consistent and be up-to-date with this version.

eclair-core/src/test/resources/integration/bitcoin.conf
```diff
@@ -1,11 +1,13 @@
 regtest=1
+noprinttoconsole=1 
 server=1
 port=28333
```
MarcoFalke: `port` is also a section setting: https://bitcoincore.org/en/releases/0.17.0/#configuration-sections-for-testnet-and-regtest

Though, given that the tests pass, setting the port is probably not needed at all

MarcoFalke:
Might also want to remove the `travis/builddeps.sh` file, which is unused and references Bitcoin Core 0.13?

Thoragh:
https://github.com/ACINQ/eclair/blob/f9ead30b5c7e2f1d3d6ff1eb93c383d4491dbada/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala#L129
You probably want to update that line? And remove the TODO a few lines below?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -46,9 +48,13 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
   def fundTransaction(tx: Transaction, lockUnspents: Boolean, feeRatePerKw: Long): Future[FundTransactionResponse] = fundTransaction(Transaction.write(tx).toString(), lockUnspents, feeRatePerKw)
 
   def signTransaction(hex: String): Future[SignTransactionResponse] =
-    rpcClient.invoke("signrawtransaction", hex).map(json => {
+    rpcClient.invoke("signrawtransactionwithwallet", hex).map(json => {
       val JString(hex) = json \ "hex"
       val JBool(complete) = json \ "complete"
+      if (!complete) {
+        val message = (json \ "errors" \\ classOf[JString]).mkString(",")
```
pm47: This is still a bit hacky and deserves a test with multiple error messages.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -111,20 +111,27 @@ class BitcoinCoreWalletSpec extends TestKit(ActorSystem("test")) with BitcoindSe
...
-    wallet.fundTransaction(unsignedTx, false, 1500).pipeTo(sender.ref)
-    val FundTransactionResponse(fundedTx, _, _) = sender.expectMsgType[FundTransactionResponse]
-
-    // change the index of the UTXO that it spends
-    val fundedTx1 = fundedTx.copy(txIn = fundedTx.txIn.updated(0, fundedTx.txIn(0).copy(outPoint = fundedTx.txIn(0).outPoint.copy(index = fundedTx.txIn(0).outPoint.index + 1))))
-
-    // signing it should fail, and the error message should contain the txid of the UTXO that could not be used
-    wallet.signTransaction(fundedTx1).pipeTo(sender.ref)
+    val unknownTxids = Seq(
+      BinaryData("01"  *32),
```
pm47: minor: format issue on lines 118->120

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -101,7 +101,7 @@ class BitcoinCoreWalletSpec extends TestKit(ActorSystem("test")) with BitcoindSe
     }
   }
 
-  test("handle error when signing transactions") {
+  test("handle errors when signing transactions") {
```
pm47: wouldn't that be nice to have kept the single error message test just in case?

sstone: It's not possible because errors are always returned in an array even if there is just one error.

PR #827, opened 2019-01-16
Add balance and channel lifecycle events to the audit db

pm47:
Note that balance events are logged at most once every 30s, and only when the balance actually changes (e.g. won't log if a payment is failed).

Also, only send `AvailableBalanceChanged` when needed.

We were sending this event everytime we sent a `commit_sig`, which is
incorrect because our balance doesn't change if, say, we are signing an
incoming htlc.

Note that we only send this event in `NORMAL` state, not in `SHUTDOWN`
state, because the balance is not really *available* in the latter.


n1bor:
Would be good to add checking balances in the integration test. As is the most critical part of the software!

And assume adding in all the nasty corner cases once channel is in CLOSING is on the todo list?

Issue #819, opened 2019-01-13
java.lang.RuntimeException: serialization error: alias: 


CatOkHero:
Dear Eclair Team, 

I have the issue when Eclair was consuming 99% of CPU and crashed after 20-30 minutes of work. 
By logs, I have found that my eclair.node-alias was too big

Here is the crash log:
```
2019-01-13 13:00:07,591 INFO  fr.acinq.eclair.Setup  - version=0.2-beta8 commit=52821b8664c329cc2e14e237df1aeff96b5260d5
2019-01-13 13:00:07,689 INFO  o.b.Secp256k1Context  - couldn't find secp256k1 library, defaulting to spongycastle
2019-01-13 13:00:08,357 INFO  fr.acinq.eclair.Setup  - nodeid=022706ea6499a904b0a545fa6212455bb3fbd0b608e901e3e652a85219b2ed0fd8 alias=”BitcoinLightningNodeUkraine”
2019-01-13 13:00:08,358 INFO  fr.acinq.eclair.Setup  - using chain=mainnet chainHash=6fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d6190000000000
2019-01-13 13:00:08,359 INFO  fr.acinq.eclair.Setup  - initializing secure random generator
2019-01-13 13:00:11,627 INFO  fr.acinq.eclair.Setup  - current feeratesPerKB=FeeratesPerKB(7537,2585,2000,2000,2000,2000) feeratesPerKw=FeeratesPerKw(1884,646,500,500,500,500)
2019-01-13 13:00:11,688 INFO  f.a.eclair.router.Router  - loading network announcements from db...
2019-01-13 13:00:11,693 INFO  f.a.eclair.router.Router  - loaded from db: channels=0 nodes=0 updates=0
2019-01-13 13:00:11,693 INFO  f.acinq.eclair.io.Server  - bound on /0:0:0:0:0:0:0:0:9735
2019-01-13 13:00:11,740 ERROR f.a.e.SimpleSupervisor  - supervisor caught error for child=router strategy=Resume akka.actor.ActorInitializationException: akka://eclair-node-gui/user/$h/router: exception during creation
	at akka.actor.ActorInitializationException$.apply(Actor.scala:180)
	at akka.actor.ActorCell.create(ActorCell.scala:607)
	at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:461)
[...]

PR #828, opened 2019-01-17
Impose a max length of 32 bytes on UTF-8 encoded aliases

araspitzu:
The configuration allowed the alias to be at most 32 chars but some chars are encoded in more than one byte (UTF-8) and this results in a spec violation where the alias was actually too long to be sent over the wire.


araspitzu:
Fixes #819

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -81,6 +81,9 @@ case class NodeParams(keyManager: KeyManager,
                       maxPendingPaymentRequests: Int,
                       maxPaymentFee: Double,
                       minFundingSatoshis: Long) {
+
+  require(alias.getBytes("UTF-8").length <= 32, "Invalid alias, too long")
```
pm47: They could be a  `require` here, but the actual message should be in `makeNodeParams` for consistency.

Also "alias must be shorter than 32B"  would be more informative

araspitzu:
> The StartupSpec seems a bit over the top, you could instead load the default configuration, override the `node.alias` key and run `makeNodeParams`.

Addressed in 31854ba

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -170,6 +171,9 @@ object NodeParams {
     val offeredCLTV = config.getInt("to-remote-delay-blocks")
     require(maxToLocalCLTV <= Channel.MAX_TO_SELF_DELAY && offeredCLTV <= Channel.MAX_TO_SELF_DELAY, s"CLTV delay values too high, max is ${Channel.MAX_TO_SELF_DELAY}")
 
+    val nodeAlias = config.getString("node-alias")
+    require(nodeAlias.getBytes("UTF-8").length <= 32, "Invalid alias, too long (max allowed 32 bytes)")
```
pm47:
```suggestion
    require(nodeAlias.getBytes("UTF-8").length <= 32, "invalid alias, too long (max allowed 32 bytes)")
```

PR #838, opened 2019-01-28
Proper handling of gossiped channels being spent

pm47:
While it makes sense to exclude from the routing table channels for
which there is a spending tx in the mempool, we shouldn't blame our
peers for considering the channel still opened if the spending tx hasn't
yet been confirmed.

Also, reworked `ValidateResult` with better types. It appears that the
`NonexistingChannel` error wasn't really useful (a malicious peer would
probably point to an existing funding tx, so there is no real difference
between "txid not found" and "invalid script"), so it was replaced by
`InvalidAnnouncement` error which is a more serious offense (punished
by a disconnection, and probably a ban when we implement that sort of
things).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
@@ -80,8 +80,7 @@ final case class ValidateRequest(ann: ChannelAnnouncement)
 sealed trait UtxoStatus
 object UtxoStatus {
   case object Unspent extends UtxoStatus
-  case object Spent extends UtxoStatus
-  case object SpentAndSpendingTxConfirmed extends UtxoStatus
+  case class Spent(spendingTxConfirmed: Boolean) extends UtxoStatus
 }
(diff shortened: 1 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
```diff
@@ -77,6 +77,12 @@ final case class WatchEventLost(event: BitcoinEvent) extends WatchEvent
   */
 final case class PublishAsap(tx: Transaction)
 final case class ValidateRequest(ann: ChannelAnnouncement)
-final case class ValidateResult(c: ChannelAnnouncement, tx: Option[Transaction], unspent: Boolean, t: Option[Throwable])
+sealed trait UtxoStatus
+object UtxoStatus {
+  case object Unspent extends UtxoStatus
+  case object Spent extends UtxoStatus
+  case object SpentAndSpendingTxConfirmed extends UtxoStatus
+}
```
sstone: Nitpick: why not `case class Spent(confirmed:Boolean)` instead ?

PR #840, opened 2019-01-29
Stop disconnected peer when it has no channels

pm47:
When we are disconnected and we don't have any channel with that peer,
we can stop it and remove its address from db.

We also need to handle the case where we are in `DISCONNECTED` or
`INITIALIZING` state and we are notified that the last channel we had
with that peer just got closed.

~Note that this opens a race condition in the switchboard, if we receive
an incoming connection from that same peer right after we stopped it,
and before the switchboard received the `Terminated` event. If that
happens, then `Peer.Connect` or `Peer.OpenChannel` will timeout. We
could also have the switchboard listens to deadletter events, but that
seems a bit over the top.~ Not anymore since eea7253.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -20,7 +20,7 @@ import java.io.ByteArrayInputStream
 import java.net.InetSocketAddress
 import java.nio.ByteOrder
 
-import akka.actor.{ActorRef, OneForOneStrategy, PoisonPill, Props, Status, SupervisorStrategy, Terminated}
+import akka.actor.{ActorRef, FSM, OneForOneStrategy, PoisonPill, Props, Status, SupervisorStrategy, Terminated}
 import akka.event.Logging.MDC
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{BinaryData, DeterministicWallet, MilliSatoshi, Protocol, Satoshi}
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -153,7 +159,13 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
     case Event(Terminated(actor), d: InitializingData) if d.channels.exists(_._2 == actor) =>
       val h = d.channels.filter(_._2 == actor).keys
       log.info(s"channel closed: channelId=${h.mkString("/")}")
-      stay using d.copy(channels = d.channels -- h)
+      val channels1 = d.channels -- h
+      if (channels1.isEmpty) {
+        // we have no existing channels, we can forget about this peer
+        stop(FSM.Normal)
+      } else {
+        stay using d.copy(channels = d.channels -- h)
```
araspitzu: To avoid redoing the diff
```suggestion
        stay using d.copy(channels = channels1)
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -18,7 +18,7 @@ package fr.acinq.eclair.io
 
 import java.net.InetSocketAddress
 
-import akka.actor.{Actor, ActorLogging, ActorRef, OneForOneStrategy, Props, Status, SupervisorStrategy, Terminated}
+import akka.actor.{Actor, ActorLogging, ActorRef, DeadLetter, OneForOneStrategy, Props, Status, SupervisorStrategy, Terminated}
```
sstone: This import is not used and should be removed.

Ligthmeup:
Thanks for answer. Can you please explain that a bit more? 
Sincerely 

Sendt fra Yahoo Mail på Android 
 
  Den fre. 1. feb.. 2019 kl. 16.28 skrev Fabrice Drouin<notifications@github.com>:   
@sstone approved this pull request.

pm47:
NB: by using `context.children`, we effectively replace a `HashMap` by a `TreeMap`, which means that peer's lookup would be O(log(N)) instead of O(1).

PR #844, opened 2019-01-31
Parametric route search

araspitzu:
This PR aims at enabling a narrower search for routes by exposing a set of parameters to limit the result and the search scope. Routes can now be 'requested' imposing a maximum number of hops, a maximum total CLTV and a maximum fees, this information will be actively used to prune edges when searching the graph for a route.


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -785,14 +793,13 @@ object Router {
...
+  val DEFAULT_ROUTE_MAX_LENGTH = 20
 
   // The default amount of routes we'll search for when findRoute is called
   val DEFAULT_ROUTES_COUNT = 3
 
-  // The default allowed 'spread' between the cheapest route found an the others
-  // routes exceeding this difference won't be considered as a valid result
-  val DEFAULT_ALLOWED_SPREAD = 0.1D
+  // The default params used for route searching, can be overridden by the caller
+  val DEFAULT_ROUTE_PARAMS = RouteParams(maxFeeBaseMsat = 21000, maxFeePct = 0.03, routeMaxCltv = 144, routeMaxLength = 6)
```
sstone: These parameters should be in our configuration file.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -809,20 +816,39 @@ object Router {
...
-    // minimum cost
-    val minimumCost = foundRoutes.head.weight
+    val foundRoutes = Graph.yenKshortestPaths(g, localNodeId, targetNodeId, amountMsat, ignoredEdges, extraEdges, numRoutes, boundaries).toList match {
+      case Nil =>
+        if(routeParams.routeMaxLength < DEFAULT_ROUTE_MAX_LENGTH){
+          return findRoute(g, localNodeId, targetNodeId, amountMsat, numRoutes, extraEdges, ignoredEdges, routeParams.copy(routeMaxLength = DEFAULT_ROUTE_MAX_LENGTH))
+        } else {
+          throw RouteNotFound
+        }
+      case route :: Nil if route.path.isEmpty =>
```
sstone: It would be cleaner to not return an empty route here and have a single "cannot find a route" check

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -809,20 +816,39 @@ object Router {
...
+      weight.length <= routeParams.routeMaxLength && weight.length <= DEFAULT_ROUTE_MAX_LENGTH &&
+      weight.cltv <= routeParams.routeMaxCltv
     }
 
-    // minimum cost
-    val minimumCost = foundRoutes.head.weight
+    val foundRoutes = Graph.yenKshortestPaths(g, localNodeId, targetNodeId, amountMsat, ignoredEdges, extraEdges, numRoutes, boundaries).toList match {
+      case Nil =>
+        if(routeParams.routeMaxLength < DEFAULT_ROUTE_MAX_LENGTH){
+          return findRoute(g, localNodeId, targetNodeId, amountMsat, numRoutes, extraEdges, ignoredEdges, routeParams.copy(routeMaxLength = DEFAULT_ROUTE_MAX_LENGTH))
```
sstone: When we retry we should also relax the max CLTV constraint

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
araspitzu: done 05b5015

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
araspitzu: done bfb9df3

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
araspitzu: @sstone done in 4f7efc3. I took the chance to group the router configurations into one case class for convenience, the configuration values will be used as default unless overridden by call specific parameters. Also I came up with a 1 week (1008 blocks) max CLTV for routes although the spec doesn't define a max CLTV for that.

araspitzu:
Last addition to this PR: config key _eclair.max-payment-fee_ has been removed in favor of _router.search-max-fee-pct_ 0bc2808, the two had near identical purpose and were overlapping with the latest change. CC: @dpad85

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -807,25 +815,36 @@ object Router {
...
       }
     }
 
-    // minimum cost
-    val minimumCost = foundRoutes.head.weight
-
-    // routes paying at most minimumCost + 10%
-    val eligibleRoutes = foundRoutes.filter(_.weight <= (minimumCost + minimumCost * DEFAULT_ALLOWED_SPREAD).round)
-    Random.shuffle(eligibleRoutes).head.path.map(graphEdgeToHop)
+    Random.shuffle(foundRoutes).head.path.map(graphEdgeToHop)
```
pm47: this deserves a comment "at this point we know that foundRoutes is not empty".

eclair-core/src/main/resources/reference.conf
```diff
@@ -97,5 +96,11 @@ eclair {
     channel-exclude-duration = 60 seconds // when a temporary channel failure is returned, we exclude the channel from our payment routes for this duration
     broadcast-interval = 60 seconds // see BOLT #7
     init-timeout = 5 minutes
+
+    // the values below will be used to perform route searching
+    search-max-length = 6         // route max length for the 'first pass'
```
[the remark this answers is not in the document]
araspitzu: Done f95ce08

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
araspitzu: Done 582f2f7

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -9,23 +9,24 @@ import Router._
 
 object Graph {
 
-  case class WeightedNode(key: PublicKey, weight: Long)
-  case class WeightedPath(path: Seq[GraphEdge], weight: Long)
+  case class RichWeight(cost: Long, length: Int, cltv: Int)
```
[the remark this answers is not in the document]
araspitzu: Done daee9fe

pm47:
I propose the following change for the configuration parameters (names, description and units). What do you think @araspitzu?

```
// parameters for the path finding algorithm when looking for a payment route
path-finding {
  max-route-length = 6 // max route length for the 'first pass', if none is found then a second pass is made with no limit
  max-cltv =  1008 // max acceptable cltv expiry for the payment (1008 ~ 1 week)
  fee-threshold-sat = 21 // if fee is below this value we skip the pax-fee-pct check
  max-fee-pct = 0.03 // route will be discarded if fee is above this value (in percentage relative to the total payment amount); doesn't apply if fee < fee-threshold-sat 
}
```

araspitzu:
> I propose the following change for the configuration parameters (names, description and units). What do you think @araspitzu?
> 
> ```
> // parameters for the path finding algorithm when looking for a payment route
> path-finding {
>   max-route-length = 6 // max route length for the 'first pass', if none is found then a second pass is made with no limit
>   max-cltv =  1008 // max acceptable cltv expiry for the payment (1008 ~ 1 week)
>   fee-threshold-sat = 21 // if fee is below this value we skip the max-fee-pct check
>   max-fee-pct = 0.03 // route will be discarded if fee is above this value (in percentage relative to the total payment amount); doesn't apply if fee < fee-threshold-sat 
> }
> ```
I like them, path-finding is a good descriptive name. Mostly we want to make sure there is no confusion with other config keys (i.e max-cltv) and i think path-finding does a good job distinguishing them. Props for the fee threshold in satoshi too, it's a bit better now.

araspitzu:
Addressed comments in c9f93e2

PR #852, opened 2019-02-07
Update jeromq dependency

pm47:


pm47:
Nice catch. Yes, it may improve the numerous-socket-opening issue that we see on windows.

pm47:
@dpad85 updated

PR #867, opened 2019-02-14
Extended API for fine-grained payment tracking

btcontract:
This is (yet another) attempt to sneak in an extended API which tries to improve on current limitations:

- `send` method has a hard timeout so can't be relied upon when payment takes a long time.
- Irregular cases, such as on-chain refunds and lost dusty payments, are not accounted for.

Above limitations are tolerable when Eclair is run by user directly, but are problematic when API is used by a service which sends/receives payments on behalf of it's users since a service has to be able to tell users what is going on with their payments as well as reliably track these payments at all times.

New API strategy is this:

- `send` becomes a fire-and-forget method.
- Two new `Payment` events are added: `PaymentSettlingOnChain` and `PaymentLostOnChain`.
- Outgoing payment is considered PENDING until websoket sends `PaymentSent` | `PaymentLostOnChain` | `PaymentSettlingOnChain(isDone = true)`, alternatively user may call `sentinfo` API endpoint which would return the same objects.
- Incoming payment is considered PENDING until it is expired or websocket sends `PaymentReceived` | `PaymentLostOnChain` | `PaymentSettlingOnChain(isDone = true)`, alternatively user may call `receivedinfo` API endpoint which would return the same objects.

One serious change this requires is https://github.com/btcontract/eclair/commit/a76ec92e3c1d09608c3db37f819de416fc0cbbc1#diff-aeb98076f262543b70a86753bd3719adR53 which adds `paymentHash` to `HtlcTimeoutTx`. This is needed for `PaymentSettlingOnChain` event which has `paymentHash` and `txid` fields. `txid` there belongs to a final CSV-delayed refunding transaction and allows API user to link off-chain failed payment to on-chain refunding transaction.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -408,55 +409,74 @@ object Helpers {
...
-              localDelayedPubkey,
-              localParams.defaultFinalScriptPubKey, feeratePerKwDelayed)
-            val sig = keyManager.sign(claimDelayed, keyManager.delayedPaymentPoint(localParams.channelKeyPath), localPerCommitmentPoint)
-            Transactions.addSigs(claimDelayed, sig)
-          })
+      val preimages = commitments.localChanges.all.collect { case u: UpdateFulfillHtlc => sha256(u.paymentPreimage) -> u.paymentPreimage }.toMap
+      val inFlightHtlcs = commitments.localCommit.spec.htlcs.map(htlc => htlc.add.paymentHash -> htlc.add.amountMsat).toMap
+      var onChainHashes = Set.empty[BinaryData] // this is needed to collect hashes of payments which can be resolved on-chain
+
+      val successAndDelayedTxs = for {
```
btcontract: An advantage of using `for`-sequences here:

- if either `success` or `delayed` fails then no tx is published as opposed to current approach where `success` may be published but `delayed` won't be (because applying on-chain fees would make it dusty for example).
- if either `success` or `delayed` fails then no `PaymentSettlingOnChain` will be sent, but `PaymentLostOnChain` will be sent for this payment instead.

Further, if `LocalCommitPublished` can be modified to store `(success, delayed)` and `(timeout, delayed)` tuples directly then UI and API may inform how much blocks is left for each delayed payment: `CLTV(timeout) + CSV(delayed)` as well as total on-chain fees paid: `timeout.input.amount - delayed.output.amount`.

btcontract:
`IntegrationSpec` seems to be failing, I'm not sure what's going on here, any advice?

araspitzu:
> `IntegrationSpec` seems to be failing, I'm not sure what's going on here, any advice?

Looks like integration-test ecalir nodes couldn't find a config key, maybe because of this addition? https://github.com/ACINQ/eclair/pull/867/files#diff-8f6c9262b4235c9580874fdd69f54b7cR234

pm47:
This is has been superseded by #884 and #885.

PR #873, opened 2019-02-18
Electrum fixes

sstone:
- clear history status when we get disconnected and still have pending history requests
- represent block heights as signed integers
- prepend a version number to persisted electrum wallet data


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -331,9 +336,11 @@ class ElectrumWallet(seed: BinaryData, client: ActorRef, params: ElectrumWallet.
             case (Some(previousHeight), height) if previousHeight != height =>
               // there was a reorg
               context.system.eventStream.publish(TransactionConfidenceChanged(txid, confirmations, data.computeTimestamp(txid, params.walletDb)))
-              downloadHeadersIfMissing(height.toInt)
-              client ! GetMerkle(txid, height.toInt)
-            case (Some(previousHeight), height) if previousHeight == height && data.proofs.get(txid).isEmpty =>
+              if (height > 0) {
```
pm47: Why isn't this check not in the `downloadHeadersIfMissing` function?

sstone: Because you don't want to download headers, but you also don't want to download merkle proofs.

PR #879, opened 2019-02-21
Set default `to-remote-delay-blocks` to 720

pm47:
It sounds like a more reasonable default value, 144 was too short.


pm47:
Hmm, @sstone why haven't you just modified the existing ``commongConfig` instead of duplicating and shadowing it?

PR #883, opened 2019-02-25
Electrum: use current change address as channel closing address

### LINKED ISSUE #882: Electrum: use new "closing addresses" for new channels
In Electrum mode (mobile) when we create a new channel, we give it the wallet's current receive address to be used later when it is closed, which means different channels will use the same onchain closing address if there was no onchain activity. We should use a new address instead.  

See https://github.com/ACINQ/eclair-mobile/issues/153

sstone:
We currently use the current receive address as closing address for new channels.
But if we create several new channels without any onchain activity, they will all use the same
closing address for their closing transaction.

Using the current change address instead is much better since it will be changed as soon as the
transactions is published.

Fixes  #882


sstone:
Additional info: my first idea was to just generate a new receive address and return it but it's not that simple: if or whatever reason theses addresses are not actually used, we may end up with a range of unused received addresses that is larger than the wallet's swipe range. If that happens, then restoring the onchain wallet from its BIP49 seed may not work (the new wallet will stop generating and checking new keys once it has enough unused ones). This is why I propose to use change addresses for our "closing addresses" instead.

sstone:
It turns out it was not a good idea, even probably worse than what we currently have. Simplest solution right now it to ask for the current receive address when the channel is actually closed and not when it is created.

PR #884, opened 2019-03-06
Send events when HTLCs settle on-chain

pm47:
This is a simpler alternative to #867, with the following limitations:
- no `OnChainRefundsDb` and associated API calls
- `PaymentSettlingOnChain`/`PaymentLostOnChain` events will be sent
exactly once per payment and have less information
- we don't touch `HtlcTimeoutTx`
- use json4s type hints instead of manual attributes to case classes


araspitzu:
TODO: update the doc

eclair-core/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -31,6 +49,15 @@ trait Service extends Directives with Logging {
   import JsonSupport.marshaller
   import JsonSupport.formats
   import JsonSupport.serialization
+  // used to send typed messages over the websocket
+  val formatsWithTypeHint = formats.withTypeHintFieldName("type") +
+    ShortTypeHints(List(
```
pm47: How difficult is it to provide custom mapping for type hints?

araspitzu: Not difficult at all, will do

araspitzu: Done in 52b17fe

eclair-core/src/main/scala/fr/acinq/eclair/api/JsonSerializers.scala
```diff
@@ -186,4 +186,15 @@ object JsonSupport extends Json4sSupport {
 
   implicit val shouldWritePretty: ShouldWritePretty = ShouldWritePretty.True
 
+  case class CustomTypeHints(custom: Map[Class[_], String]) extends TypeHints {
```
pm47: Can this be put in a separate file?

eclair-core/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -31,6 +51,15 @@ trait Service extends Directives with Logging {
   import JsonSupport.marshaller
   import JsonSupport.formats
   import JsonSupport.serialization
+  // used to send typed messages over the websocket
+  val formatsWithTypeHint = formats.withTypeHintFieldName("type") +
+    CustomTypeHints(Map(
+      classOf[PaymentSent] -> "payment-sent",
```
pm47: Nice

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1247,6 +1247,11 @@ class Channel(val nodeParams: NodeParams, wallet: EclairWallet, remoteNodeId: Pu
             log.info(s"cannot fail overriden htlc #${add.id} paymentHash=${add.paymentHash} (origin not found)")
         }
       }
+      // for our outgoing payments, let's send events if we know that they will settle on chain
+      Closing
+        .onchainOutgoingHtlcs(d.commitments.localCommit, d.commitments.remoteCommit, d.commitments.remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit), tx)
```
pm47: @sstone can you double check this? this is quite sensitive

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -802,6 +803,38 @@ object Helpers {
...
       }
 
+    /**
+      * Tells if we were the origin of this outgoing htlc
+      *
+      * @param htlcId
+      * @param originChannels
+      * @return
+      */
+    def isSentByLocal(htlcId: Long, originChannels: Map[Long, Origin]) = (originChannels.get(htlcId) collect { case l: Local => l }).isDefined
```
pm47: I wonder if there is a cleaner way to do that

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -802,6 +803,38 @@ object Helpers {
...
+
+    /**
+      * As soon as a local or remote commitment reaches min_depth, we know which htlcs will be settled on-chain (whether
+      * or not they actually have an output in the commitment tx).
+      *
+      * @param localCommit
+      * @param remoteCommit
+      * @param nextRemoteCommit_opt
+      * @param tx
```
pm47:
```suggestion
      * @param tx  a transaction that is sufficiently buried in the blockchain
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -228,10 +228,10 @@ object PaymentRequest {
       f.version match {
         case 17 if prefix == "lnbc" => Base58Check.encode(Base58.Prefix.PubkeyAddress, data)
         case 18 if prefix == "lnbc" => Base58Check.encode(Base58.Prefix.ScriptAddress, data)
-        case 17 if prefix == "lntb" => Base58Check.encode(Base58.Prefix.PubkeyAddressTestnet, data)
-        case 18 if prefix == "lntb" => Base58Check.encode(Base58.Prefix.ScriptAddressTestnet, data)
+        case 17 if prefix == "lntb" || prefix == "lnbcrt" => Base58Check.encode(Base58.Prefix.PubkeyAddressTestnet, data)
```
pm47: @sstone can you double check this?

eclair-core/src/main/scala/fr/acinq/eclair/api/JsonSerializers.scala
araspitzu: I think it belongs here logically, it's where we define our application-wide JsonSupport and all the custom serializers.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
araspitzu: Perhaps this is a bit more readable
```suggestion
    def isSentByLocal(htlcId: Long, originChannels: Map[Long, Origin]) = originChannels.get(htlcId) match {
      case Some(Local(_)) => true
      case _ => false
    }
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
sstone: It's correct because Base58 addresses on testnet and regtest have the same prefix, but not Bech32 addresses :(
But I think handling "lnbcrt" on a separate line looks a bit cleaner

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -228,10 +228,10 @@ object PaymentRequest {
...
+        case 18 if prefix == "lntb" || prefix == "lnbcrt" => Base58Check.encode(Base58.Prefix.ScriptAddressTestnet, data)
         case version if prefix == "lnbc" => Bech32.encodeWitnessAddress("bc", version, data)
-        case version if prefix == "lntb" => Bech32.encodeWitnessAddress("tb", version, data)
+        case version if prefix == "lntb" || prefix == "lnbcrt" => Bech32.encodeWitnessAddress("tb", version, data)
```
sstone: Bech32 prefix should be "bcrt", not "tb"

araspitzu: Done in 5e476fd

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
sstone: Yes LGTM

araspitzu:
LGTM, i'll add the fallback address for /receive in #885

PR #885, opened 2019-03-07
Add a proper payments database

pm47:
This builds on #884.

There is no unique identifier for payments in LN protocol. Critically,
we can't use `payment_hash` as a unique id because there is no way to
ensure unicity at the protocol level.

Also, the general case for a "payment" is to be associated to multiple
`update_add_htlc`s, because of automated retries. We also routinely
retry payments, which means that the same `payment_hash` will be
conceptually linked to a list of lists of `update_add_htlc`s.

In order to address this, we introduce a payment id, which uniquely
identifies a payment, as in a set of sequential `update_add_htlc`
managed by a single `PaymentLifecycle` that ends with a `PaymentSent` or
`PaymentFailed` outcome.

We can then query the api using either `payment_id` or `payment_hash`.
The former will return a single payment status, the latter will return a
set of payment statuses, each identified by their `payment_id`.


pm47:
NB: this is best reviewed using a [diff against #884](https://github.com/ACINQ/eclair/compare/extended-api-pm...payments-db).

araspitzu:
The /send API now immediately returns a UUID, outgoing payments are stored in the paymentsDB and associated with a status, the status is updated every time there is a change by the payment lifecycle. There is also a /paymentinfo API that can be queried via payment-id and returns its status along with some extra info.

araspitzu:
The PR also adds an optional _fallback address_ to the /receive API

araspitzu:
TODO: update the doc

eclair-core/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -68,6 +70,13 @@ trait Service extends Directives with Logging {
   implicit val actorSystem: ActorSystem
   implicit val mat: ActorMaterializer
 
+  // custom directive to fail with HTTP 404 if the element was not found
+  def complete[T](fut: Future[Option[T]])(implicit marshaller: ToResponseMarshaller[T]): Route = onComplete(fut) {
```
pm47: If there is no better practice for that, then I think we should define a directive in a separate file and choose a less-ambiguous name.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1256,8 +1256,8 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       // for our outgoing payments, let's send events if we know that they will settle on chain
       Closing
         .onchainOutgoingHtlcs(d.commitments.localCommit, d.commitments.remoteCommit, d.commitments.remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit), tx)
-        .filter(add => Closing.isSentByLocal(add.id, d.commitments.originChannels)) // we only care about htlcs for which we were the original sender here
-        .foreach(add => context.system.eventStream.publish(PaymentSettlingOnChain(amount = MilliSatoshi(add.amountMsat), add.paymentHash)))
+        .map(add => (add, d.commitments.originChannels.get(add.id).collect { case Local(id, _) => id })) // we resolve the payment id if this was a local payment
```
pm47: Nitpick: replace with a `zip`

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,70 @@
...
+  * @param id           internal payment identifier
+  * @param payment_hash payment_hash
   * @param amount_msat  amount of the payment, in milli-satoshis
-  * @param timestamp    absolute time in seconds since UNIX epoch when the payment was created.
+  * @param lastUpdate   absolute time in seconds since UNIX epoch when the payment was last updated.
   */
-case class Payment(payment_hash: ByteVector32, amount_msat: Long, timestamp: Long)
+case class OutgoingPayment(id: UUID, paymentHash: ByteVector32, amountMsat: Long, lastUpdate: Long, status: OutgoingPaymentStatus.Value)
+
+object OutgoingPaymentStatus extends Enumeration {
```
pm47: Nitpick: I would define this in the `OutgoingPayment` companion object.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,70 @@
...
+
+/**
+  * Outgoing payment is every payment that is sent by this node, they may not be finalized and
+  * when is final it can be failed or successful.
+  *
+  * @param id           internal payment identifier
+  * @param payment_hash payment_hash
   * @param amount_msat  amount of the payment, in milli-satoshis
-  * @param timestamp    absolute time in seconds since UNIX epoch when the payment was created.
+  * @param lastUpdate   absolute time in seconds since UNIX epoch when the payment was last updated.
```
pm47: How about `created`+`updated`?

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,70 @@
...
   */
-case class Payment(payment_hash: ByteVector32, amount_msat: Long, timestamp: Long)
+case class OutgoingPayment(id: UUID, paymentHash: ByteVector32, amountMsat: Long, lastUpdate: Long, status: OutgoingPaymentStatus.Value)
```
pm47:
```suggestion
case class SentPayment(id: UUID, paymentHash: ByteVector32, amountMsat: Long, lastUpdate: Long, status: SentPaymentStatus.Value)
```

For consistency with `ReceivedPayment`? I get that the payment can fail, but it is still sent and the resolution doesn't depend on the sender.

And also this is not consistent with `listSent`

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,70 @@
...
-  * Payments should not be updated nor deleted.
   */
 trait PaymentsDb {
 
-  def addPayment(payment: Payment)
+  def addReceivedPayment(payment: ReceivedPayment)
+
+  def addSentPayments(sent: OutgoingPayment)
+
+  def updateOutgoingStatus(id: UUID, newStatus: OutgoingPaymentStatus.Value)
```
pm47:
```suggestion
  def updateSentStatus(id: UUID, newStatus: OutgoingPaymentStatus.Value)
```

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,70 @@
...
+
+  def receivedByPaymentHash(paymentHash: ByteVector32): Option[ReceivedPayment]
```
pm47:
```suggestion
  def getReceivedByPaymentHash(paymentHash: ByteVector32): Option[ReceivedPayment]
```

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,70 @@
...
 
-  def findByPaymentHash(paymentHash: ByteVector32): Option[Payment]
+  def sentPaymentById(id: UUID): Option[OutgoingPayment]
```
pm47:
```suggestion
  def getSentPaymentById(id: UUID): Option[OutgoingPayment]
```

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
pm47: Or event `getReceived` since arg is typed?

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,70 @@
...
 
-  def listPayments(): Seq[Payment]
+  def sentPaymentByHash(paymentHash: ByteVector32): Option[OutgoingPayment]
```
pm47:
```suggestion
  def getSentByHash(paymentHash: ByteVector32): Option[OutgoingPayment]
```

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -125,6 +127,7 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb {
       var q: Queue[PaymentSent] = Queue()
       while (rs.next()) {
         q = q :+ PaymentSent(
+          id = ChannelCodecs.UNKNOWN_UUID, // TODO: temporary! fix that
```
pm47: We should add a payment id column to the audit.sent table. Or maybe just the payment hash?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -46,11 +62,13 @@ trait Eclair {
...
 
-  def receive(description: String, amountMsat: Option[Long], expire: Option[Long]): Future[String]
+  def receive(description: String, amountMsat: Option[Long], expire: Option[Long], fallbackAddress: Option[String]): Future[String]
 
   def findRoute(targetNodeId: PublicKey, amountMsat: Long, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty): Future[RouteResponse]
 
-  def send(recipientNodeId: PublicKey, amountMsat: Long, paymentHash: ByteVector32, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, minFinalCltvExpiry: Option[Long] = None): Future[PaymentResult]
+  def send(recipientNodeId: PublicKey, amountMsat: Long, paymentHash: ByteVector32, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, minFinalCltvExpiry: Option[Long] = None): Future[UUID]
+
+  def paymentInfo(id: UUID): Future[Option[OutgoingPayment]]
```
pm47:
```suggestion
  def sendInfo(id: UUID): Future[Option[OutgoingPayment]]
```

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -46,11 +62,13 @@ trait Eclair {
...
 
   def checkpayment(paymentHash: ByteVector32): Future[Boolean]
```
pm47:
```suggestion
  def receiveInfo(paymentHash: ByteVector32): Future[Boolean]
```

araspitzu: It seems consistent to call it 'receiveInfo' however 'checkpayment' is really checking if the payment has been received and nothing else, it doesn't return generic information about it but just a boolean. I'd agree to call it 'receiveInfo' if we can do more than the bare check.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -245,7 +245,7 @@ class Setup(datadir: File,
       authenticator = system.actorOf(SimpleSupervisor.props(Authenticator.props(nodeParams), "authenticator", SupervisorStrategy.Resume))
       switchboard = system.actorOf(SimpleSupervisor.props(Switchboard.props(nodeParams, authenticator, watcher, router, relayer, wallet), "switchboard", SupervisorStrategy.Resume))
       server = system.actorOf(SimpleSupervisor.props(Server.props(nodeParams, authenticator, serverBindingAddress, Some(tcpBound)), "server", SupervisorStrategy.Restart))
-      paymentInitiator = system.actorOf(SimpleSupervisor.props(PaymentInitiator.props(nodeParams.nodeId, router, register), "payment-initiator", SupervisorStrategy.Restart))
+      paymentInitiator = system.actorOf(SimpleSupervisor.props(PaymentInitiator.props(nodeParams.nodeId, router, register, database.payments), "payment-initiator", SupervisorStrategy.Restart))
```
pm47: We should really only pass `nodeParams` to the constructor, like for all other actors.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,75 @@
...
+
+/**
+  * Outgoing payment is every payment that is sent by this node, they may not be finalized and
+  * when is final it can be failed or successful.
+  *
+  * @param id           internal payment identifier
+  * @param payment_hash payment_hash
   * @param amount_msat  amount of the payment, in milli-satoshis
-  * @param timestamp    absolute time in seconds since UNIX epoch when the payment was created.
+  * @param updatedAt   absolute time in seconds since UNIX epoch when the payment was last updated.
```
pm47: `createdAt` missing in javadoc, also there are inconsistencies between camel case and underscore names

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,75 @@
...
   */
-case class Payment(payment_hash: ByteVector32, amount_msat: Long, timestamp: Long)
+case class OutgoingPayment(id: UUID, paymentHash: ByteVector32, amountMsat: Long, createdAt: Long, updatedAt: Long, status: OutgoingPaymentStatus.Value)
```
pm47: Have you missed the `OutgoingPayment`->`SentPayment` comment or did you have an objection?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
pm47: This is not clear at all if `checkpayment` refers to an outgoing or incoming payment. Having it return an `Option[ReceivedPayment]` and be symmetrical with `sendInfo` is fine with me.

araspitzu:
The 'id' column in audit is added in 9134a66,  i agree with returning `Option[ReceivedPayment]` and being symmetrical with the `/paymentInfo` API i was waiting to discuss it. I see the point of changing `OutgoingPayment`->`SentPayment` although `SentPayment` really means the payment has been **sent** which is not what the object is mapping, we use it to update the status of an outgoing payment which may never reach the destination (in that case is not sent at all). With that considered i don't like `OutgoingPayment` either but i couldn't find a better name.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
araspitzu: Where are the inconsitencies? We use everywhere the underscore in SQL statements and camel case in scala code.

dpad85:
Nitpick: new files like `ExtraDirectives` are missing the Apache 2 header.

araspitzu:
On top of the new payment-id functionality this PR also adds invoice persistence to the DB. This provides eclair with better handling of incoming payments and allows to offer new APIs as well, it is now possible to list all existing invoices (optionally filtered by from/to date), all _pending_ invoices and query individual invoices by payment hash. The DB migration discards previous received payments.

n1bor:
Did a bit of testing and looks good.

Few issues I noticed...
When sending from GUI get:

> Cannot cast java.util.UUID to fr.acinq.eclair.payment.PaymentLifecycle$PaymentResult

When using CLI and receivedinfo on an invoice which has not been paid get:

> parse error: Invalid numeric literal at line 1, column 4

which is a little confusing.

cli help missing the new commands.

But generally fixes all the issues I have with the API.

araspitzu:
@n1bor thanks for the feedback, all reported issues should be fixed in commits: 31db633,  22d328c, c5a2ef2. Now when the APIs are queried for an element that is not found in the db it replies with HTTP 404 AND a friendler json encoded message (this fixes the cli).

eclair-core/src/main/scala/fr/acinq/eclair/payment/LocalPaymentHandler.scala
```diff
@@ -38,44 +36,30 @@ import scala.util.Try
...
-  def run(hash2preimage: Map[ByteVector32, PendingPaymentRequest]): Receive = {
-
-    case PurgeExpiredRequests =>
-      context.become(run(hash2preimage.filterNot { case (_, pr) => hasExpired(pr) }))
+  override def receive: Receive = {
 
     case ReceivePayment(amount_opt, desc, expirySeconds_opt, extraHops, fallbackAddress_opt) =>
       Try {
-        if (hash2preimage.size > nodeParams.maxPendingPaymentRequests) {
+        if (paymentDb.listPendingPaymentRequests().size > nodeParams.maxPendingPaymentRequests) {
```
pm47: This `Try` pattern isn't great. It was there before though.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -16,40 +16,76 @@
...
+  */
+case class ReceivedPayment(paymentHash: ByteVector32, amountMsat: Long, timestamp: Long)
+
+/**
+  * Sent payment is every payment that is sent by this node, they may not be finalized and
+  * when is final it can be failed or successful.
   *
-  * @param payment_hash identifier of the payment
+  * @param id           internal payment identifier
+  * @param payment_hash payment_hash
```
pm47: The javadoc is out of date.

pm47:
Serialization result is weird. Why the `null`s?
```json
{
    "prefix": "lntb",
    "amount": null,
    "timestamp": 1555089352,
    "nodeId": "03dca9db59c25bfb430667d5c080b55a08b808b39251ec9e5b3d86939262d18e6a",
    "description": "test",
    "paymentHash": "b80ea80ec5d6d1182ea0fa418c7c02ddac3e7d6af686449a46450712155ba2a5",
    "expiry": 3600,
    "minFinalCltvExpiry": null,
    "serialized": "lntb1pwtp37gpp5hq82srk96mg3st4qlfqcclqzmkkrult276ryfxjxg5r3y92m52jsdq8w3jhxaqxqrrssf8gdlty0v67dq3xn6m4e6je84ym65p5uf3a5pmqrlmdyk0ch93ak2pndqhqf5khswq32wmhwq0xyc22qpx4f0zpk77nvq55s37x5sdqqqfwpkw"
}
```

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -34,32 +50,40 @@ trait Eclair {
...
+  def receive(description: String, amountMsat: Option[Long], expire: Option[Long], fallbackAddress: Option[String]): Future[PaymentRequest]
 
-  def receive(description: String, amountMsat: Option[Long], expire: Option[Long]): Future[String]
+  def receivedInfo(paymentHash: ByteVector32): Future[Option[IncomingPayment]]
 
-  def findRoute(targetNodeId: PublicKey, amountMsat: Long, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty): Future[RouteResponse]
+  def send(recipientNodeId: PublicKey, amountMsat: Long, paymentHash: ByteVector32, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, minFinalCltvExpiry: Option[Long] = None): Future[UUID]
 
-  def send(recipientNodeId: PublicKey, amountMsat: Long, paymentHash: ByteVector32, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, minFinalCltvExpiry: Option[Long] = None): Future[PaymentResult]
+  def sentInfo(id: Either[UUID, ByteVector32]): Future[Option[OutgoingPayment]]
```
pm47: There is a problem here. It should be (note the return type):
```scala
def sentInfo(uuid: UUID]): Future[Option[OutgoingPayment]]
def sentInfo(paymentHash: ByteVector32): Future[Seq[OutgoingPayment]]
```

n1bor:
my issues all fixed.

One other point though sentinfo does not return the pre-image? Looks like it is missing from the OutgoingPayment class?
The pre-image is obviously really useful to prove you made the payment. Is in the GUI just not the sent_payments table.

eclair-core/src/main/scala/fr/acinq/eclair/payment/LocalPaymentHandler.scala
araspitzu: It's still useful, the PaymentRequest creation can go wrong and this allows to send a feedback back to the sender.

pm47: This I think would be more scala-esque:

```scala
Try {
        val paymentPreimage = randomBytes32
        val expirySeconds = expirySeconds_opt.getOrElse(nodeParams.paymentRequestExpiry.toSeconds)
        val paymentRequest = PaymentRequest(nodeParams.chainHash, amount_opt, paymentHash, nodeParams.privateKey, desc, fallbackAddress_opt, expirySeconds = Some(expirySeconds), extraHops = extraHops)
        log.debug(s"generated payment request={} from amount={}", PaymentRequest.write(paymentRequest), amount_opt)
        paymentDb.addPaymentRequest(paymentRequest, paymentPreimage)
      } match {
        case Success(paymentRequest) => sender ! paymentRequest
        case Failure(t) => sender ! Status.Failure(t) 
      }
```

Kukks:
Looking forward to this :D

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -17,68 +17,210 @@
...
   val DB_NAME = "payments"
-  val CURRENT_VERSION = 1
+  val CURRENT_VERSION = 2
 
   using(sqlite.createStatement()) { statement =>
-    require(getVersion(statement, DB_NAME, CURRENT_VERSION) == CURRENT_VERSION) // there is only one version currently deployed
-    statement.executeUpdate("CREATE TABLE IF NOT EXISTS payments (payment_hash BLOB NOT NULL PRIMARY KEY, amount_msat INTEGER NOT NULL, timestamp INTEGER NOT NULL)")
+    require(getVersion(statement, DB_NAME, CURRENT_VERSION) <= CURRENT_VERSION) // version 2 is "backward compatible" in the sense that it uses separate tables from version 1. There is no migration though
+    statement.executeUpdate("CREATE TABLE IF NOT EXISTS received_payments (payment_hash BLOB NOT NULL PRIMARY KEY, preimage BLOB NOT NULL, payment_request TEXT NOT NULL, received_msat INTEGER, created_at INTEGER NOT NULL, expire_at INTEGER, received_at INTEGER)")
+    statement.executeUpdate("CREATE TABLE IF NOT EXISTS sent_payments (id TEXT NOT NULL PRIMARY KEY, payment_hash BLOB NOT NULL, preimage BLOB, amount_msat INTEGER NOT NULL, created_at INTEGER NOT NULL, completed_at INTEGER, status VARCHAR NOT NULL)")
```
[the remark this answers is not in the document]
araspitzu: Agreed, done in f7fc22c

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -17,68 +17,210 @@
...
+        Some(preimage, pr)
+      } else {
+        None
+      }
+    }
+  }
+
+  override def listPaymentRequests(from: Long, to: Long): Seq[PaymentRequest] = listPaymentRequests(from, to, pendingOnly = false)
+
+  override def listPendingPaymentRequests(from: Long, to: Long): Seq[PaymentRequest] = listPaymentRequests(from, to, pendingOnly = true)
```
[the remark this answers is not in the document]
araspitzu: Good one, done in 153b62c

PR #886, opened 2019-03-07
NetworkDb: remove stale channels in batch

sstone:
Use batch queries to remove/prune multiple channels  (we currently iterate and use one query per channel). This should allow us to remove stale channels on Android  much more efficiently.


eclair-core/src/main/scala/fr/acinq/eclair/db/NetworkDb.scala
```diff
@@ -49,7 +50,9 @@ trait NetworkDb {
 
   def listChannelUpdates(): Seq[ChannelUpdate]
 
-  def addToPruned(shortChannelId: ShortChannelId)
+  def addToPruned(shortChannelId: ShortChannelId): Unit = addToPruned(Seq(shortChannelId))
```
pm47: Why did you keep `def addToPruned(shortChannelId: ShortChannelId)`?

sstone: No real reason, I've removed it.

eclair-core/src/test/scala/fr/acinq/eclair/db/SqliteNetworkDbSpec.scala
```diff
@@ -104,19 +106,36 @@ class SqliteNetworkDbSpec extends FunSuite {
     db.updateChannelUpdate(channel_update_1)
   }
 
-  test("add/remove/test pruned channels") {
+  test("remove many channels") {
```
pm47: Why is this test so much more complicated?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -344,7 +344,7 @@ class Router(nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Prom
 
     case Event(TickPruneStaleChannels, d) =>
       // first we select channels that we will prune
-      val staleChannels = getStaleChannels(d.channels.values, d.updates)
+      val staleChannels = getStaleChannels(d.channels.values, d.updates).toSeq
```
pm47: can we make take an `removeChannels` take an `Iterable` to remove a copy?

PR #887, opened 2019-03-08
Support option_simplified_commitment

araspitzu:
Add support for the new commitment type with [_option_simplified_commitment_](https://github.com/lightningnetwork/lightning-rfc/pull/513). This PR proposes a non-typed design built on top of the existing Commitments class, there is now a _version_ field that is used to identify the type of commitment being used; when we serialize the STATE_DATA the version is translated into a byte and prepended to the serialized state. Summary of the changes:

- There are two  _pushMe_ outputs of 1000sats each, both paid by the funder but one sends to the fundee 
- The option is sticky, once a channel is past FUNDING_SIGNED it can't be changed
- There are no _update_fee_ messages, the feerate is set to be 253 sat/kw
- HTLCs transactions are signed using SIGHASH_SINGLE | SIGHASH_ANYONECANPAY and using feerate = 0, we'll have to bring our own inputs here as well.
- The negotiation to close the channels is started by the _fundee_
- _to_remote_ output is delayed and the pubkey is not rotated
- [WIP] To close the channel we must CPFP the commitment TX by attaching new inputs (bring your own fee)

At the time of this writing there are still some open questions on the BOLTs themselves regarding _option_simplified_commitment_.


araspitzu:
Closing because `option_simplified_commitment` has been replaced by `option_static_remotekey` https://github.com/lightningnetwork/lightning-rfc/pull/642

PR #653, opened 2018-07-04
Cleanup internal Eclair API

Christewart:
I ended up writing this to use for a suredbits project, but figured it would be better to just open source and allow for a little bit of a cleaner abstraction in eclair that hides all of the actor stuff. This currently implements a subset of the Actor messages `Service`. 

The goal with this is to replace all of the actor asks in `Service` with this eventually. For now I just wanted to get this out here and see what people think. 

I think the return types for some of the Actors (`openChannel`, `connectToNode`) probably should return something different than just a `String`.

PR #894, opened 2019-03-13
API: replace JSON-RPC by named parameters

araspitzu:
Port the existing API functionalities over a new structure of HTTP endpoints, with the biggest difference being the usage of **named parameters** for the requests (responses are unchanged). RPC methods have become endpoints and the parameters for each are now passed via form-params (clients must use the header "Content-Type" : "multipart/form-data"), this allows for a clearer interpretation of the parameters and results in more elegant parsing code on the server side. It is possible to still use the old API version via a configuration key.


eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
```diff
@@ -0,0 +1,391 @@
...
+    }
+  }
+
+  def connect(nodeId_opt: Option[PublicKey], host_opt:Option[String], port_opt: Option[Int], uri_opt: Option[String]): Route = (nodeId_opt, host_opt, port_opt, uri_opt) match {
+    case (None, None, None, Some(uri)) => complete((appKit.switchboard ? Peer.Connect(NodeURI.parse(uri))).mapTo[String])
+    case (Some(nodeId), Some(host), Some(port), None) => complete((appKit.switchboard ? Peer.Connect(NodeURI.parse(s"$nodeId@$host:$port"))).mapTo[String])
+    case _ => reject(UnknownParamsRejection("Wrong arguments for 'connect'"))
+  }
+
+  def open(nodeId: PublicKey, fundingSatoshis: Long, pushMsat: Option[Long], fundingFeerateSatByte: Option[Long], flags: Option[Int]): Future[String] = {
```
Christewart: I know this is WIP, but I think all of this stuff would be a good idea to create a trait and implement that trait in this file. 

Some work i've done before: 

https://github.com/ACINQ/eclair/pull/653/files#diff-9cda5c6eaeb9a6dfc654be6a6db44aadR15

it means it a lot easier as a reader of this code base to reason about these things. Also with this method specifically, I would _highly_ recommend implementing types for `Satoshis` and `MilliSatoshis` as passing around raw `Long` is very dangerous. 

Feel free to copy these implementations: 

https://github.com/bitcoin-s/bitcoin-s-core/blob/master/core/src/main/scala/org/bitcoins/core/currency/CurrencyUnits.scala#L64-L94

https://github.com/bitcoin-s/bitcoin-s-core/blob/master/core/src/main/scala/org/bitcoins/core/protocol/ln/currency/MilliSatoshis.scala#L16

araspitzu: cACK for declaring the operations in a separate trait as it makes the code more readable and composable. I also agree on the typing the currency units but i think this should be part of a bigger PR for eclair as it needs to be refactored in many places, for example even if we had `fr.acinq.bitcoin.MilliSatoshi` in the parameters type i'd still be sending a `Long` down to the payment lifecycle actor.

eclair-core/pom.xml
```diff
@@ -79,10 +79,10 @@
                 <activeByDefault>true</activeByDefault>
             </activation>
             <properties>
-                <bitcoind.url>https://bitcoin.org/bin/bitcoin-core-0.16.3/bitcoin-0.16.3-x86_64-linux-gnu.tar.gz
+                <bitcoind.url>https://bitcoin.org/bin/bitcoin-core-0.17.1/bitcoin-0.17.1-x86_64-linux-gnu.tar.gz
```
pm47: Seems you haven't been starting from master branch?

eclair-core/eclair-cli
```diff
@@ -31,13 +32,21 @@ Full documentation at: <https://github.com/ACINQ/eclair>"
...
     # set password
     if [ -z ${PASSWORD} ]; then auth="eclair-cli";
     else auth="eclair-cli:"${PASSWORD}; fi
-    eval curl "--user ${auth} --silent --show-error -X POST -H \"Content-Type: application/json\" -d '{ \"method\": \"'${1}'\", \"params\": '${2}' }' ${URL}" | jq -r "$jqexp"
+    # collect form data from 2nd parameter
+    form_data=""
+    for param in ${2}; do
+      form_data="$form_data -F \"$param\""
+    done;
+    #echo "FORM_DATA: $form_data"
```
pm47: `#echo "FORM_DATA: $form_data"` should be cleaned up

eclair-core/src/main/resources/reference.conf
```diff
@@ -13,6 +13,7 @@ eclair {
     binding-ip = "127.0.0.1"
     port = 8080
     password = "" // password for basic auth, must be non empty if json-rpc api is enabled
+    use-new-version = false
```
pm47: I would name this `use-old-api` and set to false (make the api default).

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -264,26 +262,50 @@ class Setup(datadir: File,
...
         logger.info(s"json-rpc api enabled on port=${config.getInt("api.port")}")
         implicit val materializer = ActorMaterializer()
-        val api = new Service {
+        val api = if(config.getBoolean("api.use-new-version")){
+          new NewService {
+
+          override val actorSystem = kit.system
 
-          override def scheduler = system.scheduler
+          override val mat = materializer
```
pm47: why do we need that?

eclair-core/src/main/scala/fr/acinq/eclair/api/EclairApi.scala
```diff
@@ -0,0 +1,224 @@
...
+
+  def checkpayment(paymentHash: ByteVector32): Future[Boolean]
+
+  def audit(from_opt: Option[Long], to_opt: Option[Long]): Future[AuditResponse]
+
+  def networkFees(from_opt: Option[Long], to_opt: Option[Long]): Future[Seq[NetworkFee]]
+
+  def channelStats(): Future[Seq[Stats]]
+
+  def help(): List[String]
```
pm47: `help` should'n be part of the programmatic API

eclair-core/src/main/scala/fr/acinq/eclair/api/MetaService.scala
```diff
@@ -0,0 +1,9 @@
+package fr.acinq.eclair.api
+
+import akka.http.scaladsl.server.Route
+
+trait MetaService {
```
pm47: is this really needed?

araspitzu: Yes otherwise we can't access the field "route" (see Setup.scala#L310) on two objects with different type, this is _transitory_ and will be removed as soon as we deprecate the old API.

eclair-core/src/main/scala/fr/acinq/eclair/api/EclairApi.scala
araspitzu: Done via 1fe65f4

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
araspitzu: The materializer is needed to work with akka's Flow in the websocket handler, the scheduler is used to artificially create a delay when the authentication was rejected.

eclair-core/src/main/resources/reference.conf
araspitzu: Done via 8268e70

eclair-core/eclair-cli
araspitzu: Done via 72cf2e8

eclair-core/pom.xml
araspitzu: Ah my bad, i rebased this branch on `use-bitcoin-0.17.1` as there was a failing test (fixed in that branch) and i assumed that was going to master before the API revamp anyway,

araspitzu:
db15e2b reverts the merge of use-bitcoin-0.17.1 and 3e285d1 merges master to get the TransactionSpecs fix

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
```diff
@@ -0,0 +1,251 @@
...
+      .map(TextMessage.apply)
+  }
+
+  val timeoutResponse: HttpRequest => HttpResponse = { r =>
+    HttpResponse(StatusCodes.RequestTimeout).withEntity(ContentTypes.`application/json`, "request timed out")
+  }
+
+  def userPassAuthenticator(credentials: Credentials): Future[Option[String]] = credentials match {
+    case p@Credentials.Provided(id) if p.verify(password) => Future.successful(Some(id))
+    case _ => akka.pattern.after(1 second, using = actorSystem.scheduler)(Future.successful(None)) // force a 1 sec pause to deter brute force
```
pm47: This would allow removing a class-level field:
```suggestion
    case _ => akka.pattern.after(1 second, using = actorSystem.scheduler)(Future.successful(None))(actorSystem.dispatcher) // force a 1 sec pause to deter brute force
```

Full diff:

```diff
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala b/eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
index d07642e1..87d11b2d 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
@@ -33,7 +33,6 @@ trait NewService extends Directives with Logging with MetaService {
   def eclairApi: EclairApi

   implicit val actorSystem: ActorSystem
-  implicit lazy val ec = actorSystem.dispatcher
   implicit val mat: ActorMaterializer

   // a named and typed URL parameter used across several routes, 32-bytes hex-encoded
@@ -76,7 +75,7 @@ trait NewService extends Directives with Logging with MetaService {

   def userPassAuthenticator(credentials: Credentials): Future[Option[String]] = credentials match {
     case p@Credentials.Provided(id) if p.verify(password) => Future.successful(Some(id))
-    case _ => akka.pattern.after(1 second, using = actorSystem.scheduler)(Future.successful(None)) // force a 1 sec pause to deter brute force
+    case _ => akka.pattern.after(1 second, using = actorSystem.scheduler)(Future.successful(None))(actorSystem.dispatcher) // force a 1 sec pause to deter brute force
   }

   val route: Route = {
```

eclair-core/src/main/scala/fr/acinq/eclair/api/MetaService.scala
pm47: How about something like:

```scala
case class ServiceBase(eclairApi: EclairApi, password: String, actorSystem: ActorSystem,  mat: ActorMaterializer)
```

which can then be used like so:

```scala
new ServiceBase(eclair, password, system, materializer) with NewService
```

assuming that `NewService` has been modified like this:
```scala
trait NewService extends Directives with Logging {

  this: ServiceBase =>
```

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
```diff
@@ -0,0 +1,251 @@
...
+  val route: Route = {
+    respondWithDefaultHeaders(customHeaders) {
+      handleExceptions(apiExceptionHandler) {
+        withRequestTimeoutResponse(timeoutResponse) {
+          authenticateBasicAsync(realm = "Access restricted", userPassAuthenticator) { _ =>
+            post {
+              path("getinfo") {
+                complete(eclairApi.getInfoResponse())
+              } ~
+                path("help") {
```
dpad85: The `help` endpoint should be removed. The message cannot be retrieved by `eclair-cli` if the connection to the node does not work. By moving the help message to eclair-cli, we'd have the same maintenance load but with a better UX.

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
```diff
@@ -0,0 +1,251 @@
...
+import akka.http.scaladsl.server.directives.{Credentials, LoggingMagnet}
+import akka.stream.{ActorMaterializer, OverflowStrategy}
+import akka.stream.scaladsl.{BroadcastHub, Flow, Keep, Source}
+import fr.acinq.eclair.payment.{PaymentLifecycle, PaymentReceived, PaymentRequest}
+import grizzled.slf4j.Logging
+import scodec.bits.ByteVector
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+
+trait NewService extends Directives with Logging with MetaService {
```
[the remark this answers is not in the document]
dpad85: I agree, the `NewService` class name is confusing and send a wrong message.

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
```diff
@@ -0,0 +1,251 @@
...
+                  handleWebSocketMessages(makeSocketHandler)
+                }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  val help = List(
```
dpad85: Let's move this to `eclair-cli` as mentioned above.

eclair-core/src/main/scala/fr/acinq/eclair/api/EclairApi.scala
```diff
@@ -0,0 +1,190 @@
...
+import fr.acinq.eclair.io.Peer.{GetPeerInfo, PeerInfo}
+import fr.acinq.eclair.payment.PaymentLifecycle._
+import fr.acinq.eclair.payment.{PaymentLifecycle, PaymentReceived, PaymentRequest}
+import fr.acinq.eclair.router.{ChannelDesc, RouteNotFound, RouteRequest, RouteResponse}
+import fr.acinq.eclair.wire.{ChannelAnnouncement, ChannelUpdate, NodeAddress, NodeAnnouncement}
+import scodec.bits.ByteVector
+import scala.concurrent.duration._
+import scala.concurrent.Future
+
+trait EclairApi {
```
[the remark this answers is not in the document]
dpad85: `Eclair` is too generic.

> @dpad85 could this api be used by Eclair Mobile?

Not with the way it's done here, because it's a field of `NewService` which is removed on android. It could be used if it was a parameter of `fr.acinq.eclair.Kit`, which is returned by `Setup#bootstrap()`.

Since `EclairAPI` is in fact a helper interface over the application `Kit`, why not move all these utility functions to `Kit`. It would be the standard way of interacting with the node, through the Kit.

We could have:

`trait EclairInterface` listing the helper functions

`Kit` implementing `EclairInterface` 

`NewService` would inherit all necessary methods. And I could use this helper functions from Eclair Mobile.

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
```diff
@@ -0,0 +1,251 @@
...
+  def password: String
+
+  def eclairApi: EclairApi
+
+  implicit val actorSystem: ActorSystem
+  implicit lazy val ec = actorSystem.dispatcher
+  implicit val mat: ActorMaterializer
+
+  // a named and typed URL parameter used across several routes, 32-bytes hex-encoded
+  val channelIdNamedParameter = "channelId".as[ByteVector32](sha256HashUnmarshaller)
```
dpad85: Centralizing `channelId` parameter in a variable is good for maintenance, but when reading the file I cannot easily get the endpoints' parameters in a quick glance, and I have to look up what `channelIdNamedParameter` means.

We should either:
- have a consistent and exhaustive way of describing the endpoint parameter:
```scala
formFields("channelId".as[ByteVector32](sha256HashUnmarshaller), "scriptPubKey".as[ByteVector](binaryDataUnmarshaller).?)...
```

- or at least rename `channelIdNamedParameter` to `channelId` so that we know that the `close` endpoint requires the `channelId` param

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
```diff
@@ -0,0 +1,251 @@
...
+                  formFields("description".as[String], "amountMsat".as[Long].?, "expireIn".as[Long].?) { (desc, amountMsat, expire) =>
+                    complete(eclairApi.receive(desc, amountMsat, expire))
+                  }
+                } ~
+                path("parseinvoice") {
+                  formFields("invoice".as[PaymentRequest]) { invoice =>
+                    complete(invoice)
+                  }
+                } ~
+                path("findRouteByInvoice") {
```
dpad85: Consistency issue: some endpoints are all lowercase (`updaterelayfee`, `forceclose`, ...), some are camelCased (`findRouteByNode`, `sendToInvoice`).

I think we should go for camelCase everywhere.

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
araspitzu: Addressed in 7092989

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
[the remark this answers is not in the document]
araspitzu: Done in f07741c

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -264,26 +264,44 @@ class Setup(datadir: File,
       _ <- if (config.getBoolean("api.enabled")) {
         logger.info(s"json-rpc api enabled on port=${config.getInt("api.port")}")
         implicit val materializer = ActorMaterializer()
-        val api = new Service {
+        val getInfo = GetInfoResponse(nodeId = nodeParams.nodeId,
```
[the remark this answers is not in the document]
araspitzu: Done in 8567979  I removed the 'port' field altogether because the information is already available in the pubblic addresses list.

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
araspitzu: Addressed in 26cc39e

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
araspitzu: Fixed in e07e9f6

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
araspitzu: Done in 7f63104

eclair-core/src/main/scala/fr/acinq/eclair/api/NewService.scala
araspitzu: Done in 7f63104

eclair-core/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  private def matchTestJson(apiName: String, overWrite: Boolean, response: String) = {
+    val p = Paths.get(s"src/test/resources/api/$apiName")
+
+    if (overWrite) {
+      Files.writeString(p, response)
+      assert(false, "'overWrite' should be false before commit")
+    } else {
+      val expectedResponse = Source.fromFile(p.toUri).mkString
```
sstone: Accessing resources should not be done through the file API but with `getClass.getResourceAsStream` or `getClass.getResource`, you could try this instead:

``` scala
    val name = s"/api/$apiName"
    val stream = getClass.getResourceAsStream(name)
    val expectedResponse = Try(Source.fromInputStream(stream).mkString).getOrElse(throw new RuntimeException(s"resource /api/$apiName not found, it should contain $response"))
    assert(response == expectedResponse, s"Test mock for $apiName did not match the expected response")
```

eclair-core/src/main/scala/fr/acinq/eclair/api/OldService.scala
```diff
@@ -0,0 +1,422 @@
...
+                      import kit._
+
+                      req.method match {
+                        // utility methods
+                        case "getinfo" => completeRpcFuture(req.id, getInfoResponse)
+                        case "help" => completeRpc(req.id, help)
+
+                        // channel lifecycle methods
+                        case "connect" => req.params match {
+                          case JString(pubkey) :: JString(host) :: JInt(port) :: Nil =>
```
dpad85: Could the port be optional, with default value being `9735`?

We can retrieve the default port value from `fr.acinq.eclair.io.NodeURI#DEFAULT_PORT`.

eclair-core/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -1,422 +1,222 @@
...
-                        case _ => reject(UnknownMethodRejection(req.id))
-                      }
+              post {
+                path("getinfo") {
+                  complete(eclairApi.getInfoResponse())
+                } ~
+                  path("connect") {
+                    formFields("uri".as[String]) { uri =>
+                      complete(eclairApi.connect(uri))
+                    } ~ formFields("nodeId".as[PublicKey], "host".as[String], "port".as[Int]) { (nodeId, host, port) =>
```
dpad85: Could the port be optional, with default value being `9735`?

We can retrieve the default port value from `fr.acinq.eclair.io.NodeURI#DEFAULT_PORT`.

eclair-core/src/main/scala/fr/acinq/eclair/api/OldService.scala
araspitzu: Good catch, let's do it on the new service :+1:

eclair-core/src/main/scala/fr/acinq/eclair/api/Service.scala
araspitzu: Done in 23c3515

eclair-core/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
araspitzu: Done in ba63614

eclair-core/src/main/scala/fr/acinq/eclair/api/Service.scala
araspitzu: Also updated the doc: https://acinq.github.io/eclair/#connect-manually

PR #899, opened 2019-03-14
Extended queries optional

pm47:
This is the implementation of https://github.com/lightningnetwork/lightning-rfc/pull/557.


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -820,6 +883,90 @@ object Router {
...
+
+  /**
+    * Have to split ids because otherwise message could be too big
+    * there could be several reply_channel_range messages for a single query
+    *
+    * @param shortChannelIds
+    * @return
+    */
+  def split(shortChannelIds: SortedSet[ShortChannelId]): List[ShortChannelIdsChunk] = {
+    // TODO: this is wrong because it can split blocks
```
pm47: This should be addressed

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -145,15 +144,25 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
...
           } else {
             // "old" nodes, do as before
+            log.info("peer requested a full routing table dump")
             router ! GetRoutingState
           }
         }
         if (remoteHasChannelRangeQueriesOptional || remoteHasChannelRangeQueriesMandatory) {
           // if they support channel queries, always ask for their filter
-          router ! SendChannelQuery(remoteNodeId, d.transport)
+          // NB: we always add extended info; if peer doesn't understand them it will ignore them
```
pm47: comment is innacurate and should be removed

pm47:
@sstone seems like `includeNode1` and `includeNode2` are not implemented?

BTW there is a related bug in the current implementation:

> The receiver [of a `query_short_channel_ids` message]:
        MUST follow with any node_announcements for each channel_announcement

PR #900, opened 2019-03-14
Disable route randomization by default

araspitzu:
With the current size of the network the randomization of path-finding has shown some memory related drawbacks, this PR aims at disabling it by default to provide a better user experience.


araspitzu:
Closing because of #904

PR #902, opened 2019-03-18
Rework router data structures

pm47:
Instead of using two separate maps (for channels and channel_updates), we now use a single map, which groups channel+channel_updates. This is also true for data storage, resulting in the removal of the channel_updates table.


araspitzu:
Builds on top of #899

t-bast:
Is this ready for review @pm47 ?
If you want to move forward with this PR, I think it's worth integrating it before I start messing around payment and router for MPP and Trampoline. Let me know if you want me to review this PR now.

pm47:
It will be once we have merged the parent PR (hopefully today).

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -106,10 +129,12 @@ case object TickPruneStaleChannels
   * Created by PM on 24/05/2016.
   */
 
-class Router(nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Promise[Done]] = None) extends FSMDiagnosticActorLogging[State, Data] {
+class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Promise[Done]] = None) extends FSMDiagnosticActorLogging[State, Data] {
 
   import Router._
 
+  implicit val np = nodeParams
```
t-bast: This implicit feels a bit hackish...it's only used for `PrivateChannel` to implicitly get our `nodeId`, maybe changing `PrivateChannel` to `PrivateChannel(nodeId: PublicKey, remoteNodeId: PublicKey, etc)` would be cleaner?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -381,15 +402,16 @@ class Router(nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Prom
       stay
 
     case Event('channels, d) =>
-      sender ! d.channels.values
+      sender ! d.channels.values.map(_.ann)
       stay
 
-    case Event('updates, d) =>
-      sender ! (d.updates ++ d.privateUpdates).values
+    case Event('channelsMap, d) =>
```
t-bast: nit: I think it's clearer if this one is called `'channels` and the previous one is renamed `'channelAnnouncements`

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -495,79 +515,121 @@ class Router(nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Prom
...
+        .map(chunk => QueryShortChannelIds(chainHash,
+          shortChannelIds = EncodedShortChannelIds(shortChannelIds.encoding, chunk.map(_.shortChannelId)),
+          if (routingMessage.timestamps_opt.isDefined || routingMessage.checksums_opt.isDefined)
+            TlvStream(QueryShortChannelIdsTlv.EncodedQueryFlags(shortChannelIds.encoding, chunk.map(_.flag)))
+          else
+            TlvStream.empty
+          ))
+        .toList
+      val (sync1, replynow_opt) = updateSync(d.sync, remoteNodeId, replies)
+      // we only send a rely right away if there were no pending requests
```
t-bast:
```suggestion
      // we only send a reply right away if there were no pending requests
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -776,43 +849,160 @@ object Router {
...
+
+  /**
+    * Have to split ids because otherwise message could be too big
+    * there could be several reply_channel_range messages for a single query
+    *
+    * @param shortChannelIds
+    * @return
+    */
+  def split(shortChannelIds: SortedSet[ShortChannelId]): List[ShortChannelIdsChunk] = {
+    // TODO: this is wrong because it can split blocks
```
t-bast: Is this really an issue? Does that mean we need to sometimes send smaller chunks to avoid splitting blocks?

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
```diff
@@ -45,9 +47,18 @@ case class GenericTlv(tag: UInt64, value: ByteVector) extends Tlv
   * @param unknown unknown tlv records.
   * @tparam T the stream namespace is a trait extending the top-level tlv trait.
   */
-case class TlvStream[T <: Tlv](records: Traversable[T], unknown: Traversable[GenericTlv] = Nil)
+case class TlvStream[T <: Tlv](records: Traversable[T], unknown: Traversable[GenericTlv] = Nil) {
+  /**
+    *
+    * @tparam R input type parameter, must be a subtype of the main TLV type
+    * @return the TLV record of of type that matches the input type parameter if any (there can be at most one, since BOLTs specify
```
t-bast:
```suggestion
    * @return the TLV record of type that matches the input type parameter if any (there can be at most one, since BOLTs specify
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
```diff
@@ -45,9 +47,18 @@ case class GenericTlv(tag: UInt64, value: ByteVector) extends Tlv
...
+    *         that TLV records are supposed to be unique
```
t-bast:
```suggestion
    *         that TLV records are supposed to be unique)
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Actually probably not, see https://github.com/ACINQ/eclair/pull/899/commits/361460d3c5b3e9b25134e9b5fafe22b7594aba49.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: I like the `channels`/`updates`/`nodes` triplet though.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: How about removing this line and replace the other line by:

```scala
PrivateChannel(remoteNodeId, None, None)(nodeParams)
```

pm47: Actually went with your proposal.

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
pm47: Fixed in 11c12be

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
pm47: Fixed in 11c12be

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Fixed in 11c12be

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Ok, I guess it maps cleanly to the spec structure (channel_announcement, channel_update, node_announcement)

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -56,6 +63,24 @@ case class RouterConf(randomizeRouteSelection: Boolean,
...
 
 case class ChannelDesc(shortChannelId: ShortChannelId, a: PublicKey, b: PublicKey)
+
+case class PublicChannel(ann: ChannelAnnouncement, fundingTxid: ByteVector32, capacity: Satoshi, update_1_opt: Option[ChannelUpdate], update_2_opt: Option[ChannelUpdate]) {
+  update_1_opt.foreach(u => assert(Announcements.isNode1(u.channelFlags)))
+  update_2_opt.foreach(u => assert(!Announcements.isNode1(u.channelFlags)))
+  def getNodeIdSameSideAs(u: ChannelUpdate): PublicKey = if (Announcements.isNode1(u.channelFlags)) ann.nodeId1 else ann.nodeId2
+  def getSameSideAs(u: ChannelUpdate): Option[ChannelUpdate] = if (Announcements.isNode1(u.channelFlags)) update_1_opt else update_2_opt
+  def updateSameSideAs(u: ChannelUpdate): PublicChannel = if (Announcements.isNode1(u.channelFlags)) copy(update_1_opt = Some(u)) else copy(update_2_opt = Some(u))
+  def updateFor(n: PublicKey): Option[ChannelUpdate] = if (n == ann.nodeId1) update_1_opt else if (n == ann.nodeId2) update_2_opt else throw new IllegalArgumentException("this node is unrelated to this channel")
```
sstone: Naming here is confusing,  in `updateSameSideAs` update means "modify", here it means "get" , why not simply call it `get` which would be consistent with  `getSameSideAs` ?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -294,12 +319,12 @@ class Router(nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Prom
           case (d, (n, origins)) => origins.foldLeft(d) { case (d, origin) => handle(n, origin, d) } // we reprocess the same node_announcement for every origins (to preserve origin information)
         }
         stay using d3
-      } else {
+        case None =>
```
sstone: There seems to be an indentation issue for this `match {}` statement

PR #904, opened 2019-03-19
Ensure the cost function in path-finding is monotonic

araspitzu:
With fe31f2d9d2754d6dd979bc1f454ae4bee7fc8b6d the heuristics introduced a cost function that is not monotonic and that led scenarios where the path-finding would produce an invalid result and even get stuck in an endless computation. This PR fixes that by making sure the cost function is monotonic.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -221,8 +221,8 @@ object Graph {
         // note: there is always an entry for the current in the 'weight' map
         val currentWeight = weight.get(current.key)
 
-        // for each neighbor
-        currentNeighbors.foreach { edge =>
+        // for each neighbor with fees > 0
+        currentNeighbors.filterNot(edgeHasZeroFee).foreach { edge =>
 
           val neighbor = edge.desc.a
 
@@ -276,6 +276,10 @@ object Graph {
 
         edgePath
     }
+  }
+
+  private def edgeHasZeroFee(edge: GraphEdge): Boolean = {
+    edge.update.feeBaseMsat == 0 && edge.update.feeProportionalMillionths == 0
   }
 
   // Computes the compound weight for the given @param edge, the weight is cumulative and must account for the previous edge's weight.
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -306,7 +306,7 @@ object Graph {
       val factor = (cltvFactor * wr.cltvDeltaFactor) + (ageFactor * wr.ageFactor) + (capFactor * wr.capacityFactor)
       val edgeWeight = if (isNeighborTarget) prev.weight else edgeCost * factor
 
-      RichWeight(cost = edgeCost, length = prev.length + 1, cltv = prev.cltv + channelCltvDelta, weight = edgeWeight)
+      RichWeight(cost = edgeCost, length = prev.length + 1, cltv = prev.cltv + channelCltvDelta, weight = prev.weight + edgeWeight)
```
sstone: It would be more consistent to use instead:
```scala
      val edgeWeight = if (isNeighborTarget) prev.weight else prev.weight + edgeCost * factor

      RichWeight(cost = edgeCost, length = prev.length + 1, cltv = prev.cltv + channelCltvDelta, weight = edgeWeight)
```

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -837,6 +837,36 @@ class RouteCalculationSpec extends FunSuite {
...
+    val amount = 351000
+
+    Globals.blockCount.set(567634) // simulate mainnet block for heuristic
+    val Success(route) = Router.findRoute(g, thisNode, targetNode, amount, 1, Set.empty, Set.empty, params)
+
+    assert(route.size == 2)
+    assert(route.last.nextNodeId == targetNode)
+  }
+
 }
```
sstone: We also need more tests.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: Addressed in 5039da3

PR #905, opened 2019-03-19
Deal with channels with fees=0 when computing a route

araspitzu:
Channels with feeBase=0 and feeProportional=0 cause issues to the path-finding algorithm, this PR makes eclair consider those channel as feeBase=1msat when computing a route but then when the payment is sent out we use its actual fee (zero).

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
@@ -590,6 +590,21 @@ class RouteCalculationSpec extends FunSuite {
     assert(route1.map(hops2Ids) === Success(1 :: 2 :: 4 :: 5 :: Nil))
   }
 
+  test("treat channels with fees=0 as if they had feeBase=1msat") {
+
+    val updates = List(
+      makeUpdate(1L, a, b, 10, 10),
+      makeUpdate(2L, b, c, 10, 10),
+      makeUpdate(3L, b, e, 0, 0), // goes straight to target with no fees!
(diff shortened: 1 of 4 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -590,6 +590,21 @@ class RouteCalculationSpec extends FunSuite {
...
+      makeUpdate(2L, b, c, 10, 10),
+      makeUpdate(3L, b, e, 0, 0), // goes straight to target with no fees!
+      makeUpdate(4L, c, d, 10, 10),
+      makeUpdate(5L, d, e, 10, 10)
+    ).toMap
+
+    val g = makeGraph(updates)
+
+    val route1 = Router.findRoute(g, a, e, DEFAULT_AMOUNT_MSAT, numRoutes = 1, routeParams = DEFAULT_ROUTE_PARAMS)
+    assert(route1.map(hops2Ids) === Success(1 :: 3 :: Nil))
```
sstone: I don't see where we test that we "treat channels with fees=0 as if they had feeBase=1msat" ?

araspitzu: The name has become misleading after i changed the behavior of the implementation, initially we wanted to ignore the 0-fee channels and that test was asserting exactly that.

araspitzu: How about 077a4a4?

PR #909, opened 2019-03-21
Graph fuzzy testing

araspitzu:
Using a subgraph of the current mainnet network db ( with ~1k nodes and ~30k channels) we randomly select two nodes and run the path-finding algorithm against that. Currently 500 iterations are executed against both heuristic path-finding and non heuristic.


araspitzu:
To be done: adjust the mock graph size to something smaller and more git-friendly (currently is ~30MB).

eclair-core/src/test/scala/fr/acinq/eclair/router/FuzzyGraph.scala
```diff
@@ -0,0 +1,227 @@
...
+import fr.acinq.eclair.router.Graph.WeightRatios
+import fr.acinq.eclair.wire.{ChannelAnnouncement, ChannelUpdate, Color, NodeAddress}
+import org.json4s.{FileInput, jackson}
+import org.scalatest.FunSuite
+import scodec.bits.ByteVector
+import scala.util.{Failure, Random, Success}
+import scala.collection.immutable.TreeMap
+import scala.collection.mutable
+
+class FuzzyGraph extends FunSuite {
```
sstone: This should be call `FuzzyGraphSpec` instead

araspitzu: Done in 1e24241

araspitzu:
The mock network db is now 3.5MB, it's CSV formatted, zip compressed and stripped from the signatures. The whole test takes a bit less than 40s to run on a high-end laptop.

t-bast:
Closing this stale PR, we've moved to an A/B testing model instead.

PR #912, opened 2019-03-22
Don't send updates if no filter has been set

pm47:
We mistakenly implemented `Peer.timestampInRange` as opt-out whereas it
should be opt-in.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -325,7 +325,7 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
         case (count, (_, origins)) if origins.contains(self) =>
           // the announcement came from this peer, we don't send it back
           count
-        case (count, (msg: HasTimestamp, _)) if !timestampInRange(msg, d.gossipTimestampFilter) =>
+        case (count, (msg, _)) if !timestampInRange(msg, d.gossipTimestampFilter) =>
           // the peer has set up a filter on timestamp and this message is out of range
           count
         case (count, (msg, _)) =>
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -596,14 +596,15 @@ object Peer {
     *
     * @param gossipTimestampFilter_opt optional gossip timestamp range
     * @return
-    *           - true if the msg's timestamp is in the requested range, or if there is no filtering
-    *           - false otherwise
+    *           - true if the msg's timestamp is in the requested range or if msg has no timestamps
```
sstone:
```suggestion
    *           - true if there is a filter and msg has no timestamp, or has one that matches the filter 
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -596,14 +596,15 @@ object Peer {
...
+    *           - false otherwise, or if there is no filtering
```
sstone:
```suggestion
    *           - false otherwise
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -596,14 +596,15 @@ object Peer {
...
-  def timestampInRange(msg: HasTimestamp, gossipTimestampFilter_opt: Option[GossipTimestampFilter]): Boolean = {
+  def timestampInRange(msg: RoutingMessage, gossipTimestampFilter_opt: Option[GossipTimestampFilter]): Boolean = {
     // check if this message has a timestamp that matches our timestamp filter
-    gossipTimestampFilter_opt match {
-      case None => true // no filtering
-      case Some(GossipTimestampFilter(_, firstTimestamp, timestampRange)) => msg.timestamp >= firstTimestamp && msg.timestamp <= firstTimestamp + timestampRange
+    (msg, gossipTimestampFilter_opt) match {
+      case (_, None) => false // BOLT 7: A node which wants any gossip messages would have to send this, otherwise [...] no gossip messages would be received.
+      case (hasTs: HasTimestamp, Some(GossipTimestampFilter(_, firstTimestamp, timestampRange))) => hasTs.timestamp >= firstTimestamp && hasTs.timestamp <= firstTimestamp + timestampRange
+      case _ => true // if message doesn't have a timestamp (e.g. channel_announcement), then we send it
```
sstone:
```suggestion
      case _ => true // if there is a filter and message doesn't have a timestamp (e.g. channel_announcement), then we send it
```

PR #926, opened 2019-03-29
Fix eclair-cli to work with equal sign in arguments

araspitzu:
Eclair-cli was not working as advertised in the documentation, this PR fixes the behavior.

### DIFF
--- a/eclair-core/eclair-cli
+++ b/eclair-core/eclair-cli
@@ -35,7 +35,6 @@ and COMMAND is one of:
 
 Examples
 --------
-  eclair-cli help                           display available commands
   eclair-cli -a localhost:1234 peers        list the peers of a node hosted on localhost:1234
   eclair-cli close --channelId=006fb...     closes the channel with id 006fb...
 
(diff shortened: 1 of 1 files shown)


pm47:
It works, but the error message is a bit cryptic if user forget to pur the `=`:

```shell
./eclair-cli -s channels --nodeId 029b2ff755e1fc37fc735ba8070b3c3d1b7cc9b20a4cd3b95dcfc66d289e3e09bc
incorrect argument 029b2ff755e1fc37fc735ba8070b3c3d1b7cc9b20a4cd3b95dcfc66d289e3e09bc
```

araspitzu:
> It works, but the error message is a bit cryptic if user forget to pur the `=`:
> 
> ```shell
> ./eclair-cli -s channels --nodeId 029b2ff755e1fc37fc735ba8070b3c3d1b7cc9b20a4cd3b95dcfc66d289e3e09bc
> incorrect argument 029b2ff755e1fc37fc735ba8070b3c3d1b7cc9b20a4cd3b95dcfc66d289e3e09bc
> ```

At 5b6e35c we print a more explanatory error message ` incorrect argument, please use --arg=value` and then the usage message (which again contains instructions on how to pass arguments).

PR #927, opened 2019-04-01
Added simple plugin support

pm47:
Using org.clapper:classutil library and a very simple `Plugin` interface.


eclair-node/src/main/scala/fr/acinq/eclair/Plugin.scala
```diff
@@ -0,0 +1,29 @@
...
+import java.io.File
+import java.net.{URL, URLClassLoader}
+
+import org.clapper.classutil.ClassFinder
+
+trait Plugin {
+
+  def onSetup(setup: Setup): Unit
+
+  def onKit(kit: Kit): Unit
```
araspitzu: Not sure i would expose the entire `Kit`, in this way plugins enlarge the attack surface of eclair. An insecure plugin can be exploited to access channel data.

pm47: It allows to do powerful stuff on top of eclair, and you are always taking risks by running someone else's code.

n1bor:
+1 for this. Is there any way we could interface from the REST API to the plugin? Even if could just forward a string from API -> Plugin and get a string back? Say:
**_plugin "pluginname" "String of stuff to send"_**

And there would be another method in the trait:
 **_def onReqeust(string: param): Future[String]_**

pm47:
@n1bor that seems over the top, and a plugin can open its own web interface if needed.

PR #928, opened 2019-04-01
Added a timeout for channel open request

pm47:
Currently, if the peer is unresponsive (typically doesn't respond to
`open_channel` or `funding_created`), we wait indefinitely, or until the
connection closes.

This translates to an API timeout for users, and uncertainty about the
state of the channel.


pm47:
TODO: API doc @araspitzu

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -70,13 +70,16 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.switchboard ? Peer.Connect(NodeURI.parse(uri))).mapTo[String]
   }
 
-  override def open(nodeId: PublicKey, fundingSatoshis: Long, pushMsat: Option[Long], fundingFeerateSatByte: Option[Long], flags: Option[Int]): Future[String] = {
+  override def open(nodeId: PublicKey, fundingSatoshis: Long, pushMsat: Option[Long], fundingFeerateSatByte: Option[Long], flags: Option[Int], timeout_opt: Option[Timeout]): Future[String] = {
+    // we want this to expire *before* the ask times out, otherwise the user won't get a usable response
+    val timeoutCappedAt50s_opt = timeout_opt.map(value => Timeout(value.duration.max(50 seconds)))
```
pm47: @sstone I think that's actually not needed. Users may want a short api response, but still let time for channel opening.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -328,6 +334,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     case Event(INPUT_DISCONNECTED, _) =>
       replyToUser(Left(LocalError(new RuntimeException("disconnected"))))
       goto(CLOSED)
+
+    case Event(TickChannelOpenTimeout, _) =>
```
araspitzu: It would be nice if we send a `shutdown` before disconnecting.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -70,13 +70,16 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.switchboard ? Peer.Connect(NodeURI.parse(uri))).mapTo[String]
   }
 
-  override def open(nodeId: PublicKey, fundingSatoshis: Long, pushMsat: Option[Long], fundingFeerateSatByte: Option[Long], flags: Option[Int]): Future[String] = {
+  override def open(nodeId: PublicKey, fundingSatoshis: Long, pushMsat: Option[Long], fundingFeerateSatByte: Option[Long], flags: Option[Int], timeout_opt: Option[Timeout]): Future[String] = {
+    // we want the default timeout to expire *before* the default ask times out, otherwise user won't get a usable response
+    val openTimeout = timeout_opt.getOrElse(Timeout(30 seconds))
     (appKit.switchboard ? Peer.OpenChannel(
```
araspitzu: Why not setting an explicit timeout here?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -70,13 +70,16 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.switchboard ? Peer.Connect(NodeURI.parse(uri))).mapTo[String]
   }
 
-  override def open(nodeId: PublicKey, fundingSatoshis: Long, pushMsat: Option[Long], fundingFeerateSatByte: Option[Long], flags: Option[Int]): Future[String] = {
+  override def open(nodeId: PublicKey, fundingSatoshis: Long, pushMsat: Option[Long], fundingFeerateSatByte: Option[Long], flags: Option[Int], timeout_opt: Option[Timeout]): Future[String] = {
+    // we want the default timeout to expire *before* the default ask times out, otherwise user won't get a usable response
+    val openTimeout = timeout_opt.getOrElse(Timeout(30 seconds))
```
araspitzu: 'orElse' should use nodeParams.apiTimeout

juscamarena:
Looks good, I think I'll test it out live in production as I need this. ;)

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
pm47: Good catch, addressed with 59ede841.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -198,6 +199,7 @@ object NodeParams {
       reserveToFundingRatio = config.getDouble("reserve-to-funding-ratio"),
       maxReserveToFundingRatio = config.getDouble("max-reserve-to-funding-ratio"),
       db = database,
+      apiTimeout = FiniteDuration(config.getDuration("api.timeout").getSeconds, TimeUnit.SECONDS),
```
sstone: If feels wrong to have the API t/o here

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -67,19 +67,22 @@ trait Eclair {
 class EclairImpl(appKit: Kit) extends Eclair {
 
   implicit val ec = appKit.system.dispatcher
-  implicit val timeout = Timeout(60 seconds) // used by akka ask
+  implicit val timeout = Timeout(appKit.nodeParams.apiTimeout) // used by akka ask
```
sstone: It feels wrong here (see comment below). Adding an implicit timeout parameter to each method call in the `Eclair` interface would be cleaner

pm47: This is what it would be in Service.scala:

```scala
          formFields(timeout.?) { tm_opt =>
            implicit val timeout = tm_opt.getOrElse(Timeout(30 seconds))
            withRequestTimeout(timeout.duration + 2.seconds) {
              withRequestTimeoutResponse(timeoutResponse) {
```

PR #929, opened 2019-04-02
Minor API fixes

sstone:
Move GetInfoResponse and AuditResponse out of the api package
Fix parameter handling for audit() and networkFees()


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -148,10 +151,8 @@ class EclairImpl(appKit: Kit) extends Eclair {
   }
 
   override def audit(from_opt: Option[Long], to_opt: Option[Long]): Future[AuditResponse] = {
-    val (from, to) = (from_opt, to_opt) match {
-      case (Some(f), Some(t)) => (f, t)
-      case _ => (0L, Long.MaxValue)
-    }
+    val from = from_opt.getOrElse(0L)
```
pm47: That's not equivalent to the previous version.

sstone: Yes previous version was buggy and did not handle all cases (spotted by @araspitzu)

Issue #906, opened 2019-03-20
Feature request: set max payment attempts from configuration


araspitzu:
When eclair sends a payment it retries up to (currently) 5 times by default and there is no option to override it via the APIs or via configuration, would be nice to have it.

PR #931, opened 2019-04-03
Set max payment attempts from configuration

rorp:
Fixes https://github.com/ACINQ/eclair/issues/906


eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -261,7 +261,7 @@ class IntegrationSpec extends TestKit(ActorSystem("test")) with BitcoindService
     val pr = sender.expectMsgType[PaymentRequest]
     // then we make the actual payment
     sender.send(nodes("A").paymentInitiator,
-      SendPayment(amountMsat.amount, pr.paymentHash, nodes("D").nodeParams.nodeId, routeParams = integrationTestRouteParams))
+      SendPayment(amountMsat.amount, pr.paymentHash, nodes("D").nodeParams.nodeId, routeParams = integrationTestRouteParams, maxAttempts = 5))
```
araspitzu: Although that was the previous behavior, there is no need to retry payments up to five times in integration test. Just tried locally and it works fine setting them to 1.

araspitzu:
PR needs to be rebased on master to fix conflicts.

araspitzu:
Concept is ACK, I wanted to do this but to be fair we should also expose the parameter via the APIs.

rorp:
I added `maxAttempts` parameter to `send` call. It would be great to update https://acinq.github.io/eclair/ but I  don't know how.

araspitzu:
There are [instructions](https://github.com/ACINQ/eclair/blob/master/BUILD.md#building-the-api-documentation) but you need commit access to update it, for the moment I'm doing it manually as the APIs are under heavy changes.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -135,10 +135,10 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.router ? RouteRequest(appKit.nodeParams.nodeId, targetNodeId, amountMsat, assistedRoutes)).mapTo[RouteResponse]
   }
 
-  override def send(recipientNodeId: PublicKey, amountMsat: Long, paymentHash: ByteVector32, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, minFinalCltvExpiry: Option[Long] = None): Future[PaymentResult] = {
+  override def send(recipientNodeId: PublicKey, amountMsat: Long, paymentHash: ByteVector32, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, minFinalCltvExpiry: Option[Long] = None, maxAttempts: Option[Int] = None): Future[PaymentResult] = {
     val sendPayment = minFinalCltvExpiry match {
-      case Some(minCltv) => SendPayment(amountMsat, paymentHash, recipientNodeId, assistedRoutes, finalCltvExpiry = minCltv)
-      case None  => SendPayment(amountMsat, paymentHash, recipientNodeId, assistedRoutes)
+      case Some(minCltv) => SendPayment(amountMsat, paymentHash, recipientNodeId, assistedRoutes, finalCltvExpiry = minCltv, maxAttempts = maxAttempts.getOrElse(appKit.nodeParams.maxPaymentAttempts))
```
pm47: I'd prefer:

```scala
  override def send(recipientNodeId: PublicKey, amountMsat: Long, paymentHash: ByteVector32, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, minFinalCltvExpiry_opt: Option[Long] = None, maxAttempts_opt: Option[Int] = None): Future[PaymentResult] = {
    val maxAttempts = maxAttempts_opt.getOrElse(appKit.nodeParams.maxPaymentAttempts)
    val sendPayment = minFinalCltvExpiry_opt match {
      case Some(minCltv) => SendPayment(amountMsat, paymentHash, recipientNodeId, assistedRoutes, finalCltvExpiry = minCltv, maxAttempts = maxAttempts))
      case None  => SendPayment(amountMsat, paymentHash, recipientNodeId, assistedRoutes, maxAttempts = maxAttempts)
    }
```

@araspitzu  note that I suffixed options with `_opt`, just noticed this isn't the case for a whole bunch of parameters in the Eclair interface.

PR #933, opened 2019-04-05
Turbo channels and more websocket events

btcontract:
This adds support for so called "turbo channels" in funder mode only for now. These are basically zeroconf channels with some amount pushed by funder which can be spent by fundee right away.

This is implemented by defining `zeroConfSpendablePush` bits in `ChannelFlags` to signal for public and private channels: https://github.com/btcontract/eclair/blob/f9ef2f41b2d1e5b66e8c3eed168260d3797d3823/eclair-core/src/test/scala/fr/acinq/eclair/FeaturesSpec.scala#L54-L58.

When bit is set funder sends `FundingLocked` once funding tx is published and fundee sends `FundingLocked` once it sees it in a mempool. Non-supporting fundees will simply wait until funding reaches a required number of confirmations so for them this degrades into a usual channel, same will happen if supporting fundee does not see a funding in a mempool and only gets it confirmed in a block.

Additionally, this adds more events for websocket to track a channel opening process on a more fine grained level. Turbo channels require some initial trust: fundee has to trust funder to publish a funding, push a certain amount (typically paid by fundee beforehand off-band) and not double-spend it. OTOH funder needs to know what exactly happened to channel if something went wrong (funder disconnected or sent an error in an opening phase).


juscamarena:
Would be great to have this merged in. Seeing no discussion from devs here makes it seem like it's not something wanted. :/

pm47:
@juscamarena I priority is releasing eclair 0.3, we haven't looked enough into this to have a opinion.

Kixunil:
What happened here? Why was this closed? The branch disappeared, is there a maintained version available somewhere?

btcontract:
Looks like there were no interest in this plus this branch got very outdated over a year, I had no desire to fix it.

Kixunil:
Oh, so I have no way to upgrade backwards-compatibly now?

btcontract:
Upgrading should be without issues since this did not change anything in storage, but opening of new turbo channels won't work obviously.

Kixunil:
Ah, great, that's very helpful as I wasn't opening Turbo channels lately. Thanks!

PR #934, opened 2019-04-05
Expose the websocket over HTTP GET

araspitzu:
This PR fixes the broken behavior of the websocket when used over HTTP POST together with basic auth. A test has been added to verify we can use the socket WITH authentication.


juscamarena:
Thanks for this, will test it out locally, thought I was going mad wondering why it wouldn't work last night.

juscamarena:
Tested and works great, I think I'll update the documentation so others aren't as confused as me when first using ws with eclair.

PR #938, opened 2019-04-09
Enrich test for internal eclair API implementation (fr.acinq.eclair.Eclair.scala)

araspitzu:
This PR adds more tests to the logic implemented in EclairApiImpl.


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -225,13 +227,18 @@ class EclairImpl(appKit: Kit) extends Eclair {
...
-  def sendToChannel(channelIdentifier: String, request: Any)(implicit timeout: Timeout): Future[Any] =
-    for {
-      fwdReq <- Future(Register.ForwardShortId(ShortChannelId(channelIdentifier), request))
-        .recoverWith { case _ => Future(Register.Forward(ByteVector32.fromValidHex(channelIdentifier), request)) }
-        .recoverWith { case _ => Future.failed(new RuntimeException(s"invalid channel identifier '$channelIdentifier'")) }
-      res <- appKit.register ? fwdReq
-    } yield res
+  def sendToChannel(channelIdentifier: String, request: Any)(implicit timeout: Timeout): Future[Any] = {
+    (Try(ForwardShortId(ShortChannelId(channelIdentifier), request)) match {
+      case Success(value) => Left(value.shortChannelId)
```
sstone: Why not call `appKit.register ? ForwardShortId(shortChannelId, request)` here ?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -225,13 +227,18 @@ class EclairImpl(appKit: Kit) extends Eclair {
...
+      case Failure(_) => Try(Forward(ByteVector32.fromValidHex(channelIdentifier), request)) match {
+        case Success(v) => Right(v.channelId)
```
sstone: Why not call `appKit.register ? Forward(channelId, request)` here ?

araspitzu: How about 125cbbf?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -225,13 +227,15 @@ class EclairImpl(appKit: Kit) extends Eclair {
...
     * @return
     */
-  def sendToChannel(channelIdentifier: String, request: Any)(implicit timeout: Timeout): Future[Any] =
-    for {
-      fwdReq <- Future(Register.ForwardShortId(ShortChannelId(channelIdentifier), request))
-        .recoverWith { case _ => Future(Register.Forward(ByteVector32.fromValidHex(channelIdentifier), request)) }
-        .recoverWith { case _ => Future.failed(new RuntimeException(s"invalid channel identifier '$channelIdentifier'")) }
-      res <- appKit.register ? fwdReq
-    } yield res
+  def sendToChannel(channelIdentifier: String, request: Any)(implicit timeout: Timeout): Future[Any] = {
```
pm47: What is the purpose of this change?

araspitzu: Purpose was so get rid of unnecessary "Future" and simplify the code but #969 got a better solution so i'm removing this.

PR #940, opened 2019-04-10
BOLT11 numeric tagged field padding

araspitzu:
This PR forces idempotent behavior on the serialization/deserialization of numeric tagged fields in the BOLT11 payment requests. When the fields are created we now use the minimum necessary bits to represent the value but we also (left) pad this to a multiple of five, this avoids losing information when the data is serialized to bech32.


araspitzu:
Merged in master by #885

PR #946, opened 2019-04-14
Multipart payments

btcontract:
This adds basic server-side support for multipart payments by adding a special LNURL tag and enabling generation of multiple payment requests at once.

Overview and demo: https://medium.com/@akumaigorodski/multipart-payments-647cdf5cc15a

Technical description: https://github.com/btcontract/lnurl-rfc/blob/master/spec.md#5-multipart-payments


btcontract:
Seems like a dead end approach in light of this conversation: https://github.com/ElementsProject/lightning/issues/2527 and possibly better approach described here: https://github.com/lightningnetwork/lightning-rfc/pull/593#discussion_r278835899

PR #949, opened 2019-04-16
Print stack trace when crashing during boot sequence

araspitzu:
Adds a more verbose output when the application crashes during the Boot procedure.


eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -48,8 +48,8 @@ object Boot extends App with Logging {
 
   def onError(t: Throwable): Unit = {
     val errorMsg = if (t.getMessage != null) t.getMessage else t.getClass.getSimpleName
-    System.err.println(s"fatal error: $errorMsg")
-    logger.error(s"fatal error: $errorMsg")
+    logger.error(s"fatal error: $errorMsg", t)
+    t.printStackTrace()
```
dpad85: `errorMsg` should be removed because the message is already printed with the stack trace

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteChannelsDb.scala
```diff
@@ -34,7 +34,7 @@ class SqliteChannelsDb(sqlite: Connection) extends ChannelsDb {
   val CURRENT_VERSION = 1
 
   using(sqlite.createStatement()) { statement =>
-    require(getVersion(statement, DB_NAME, CURRENT_VERSION) == CURRENT_VERSION) // there is only one version currently deployed
+    require(getVersion(statement, DB_NAME, CURRENT_VERSION) == CURRENT_VERSION, s"incompatible versions of $DB_NAME DB found") // there is only one version currently deployed
```
dpad85: Nitpick: `versions` should be `version`

PR #951, opened 2019-04-17
Live channel database backup

sstone:
Backups are handled by a custom actor that listens to persistence events sent by channels. They are done in 2 steps:

- create a sqlite backup in a temporary file
- rename the temporary file, which should be atomic (though it actually depends on the OS and file system).

Users would typically rename/move the backup file then make a safe copy of it.

Backups are sequential. If backup requests pile up while a backup is still in progress, only the last one will be kept.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ The files that you need to backup are located in your data directory. You must b
 Your seed never changes once it is created, but your channels do change whenever you receive our send payments.
 `eclair.bak` is safe to backup even when your system is running. We recommend that you implement your backup 
 process in 2 steps:
-- first, rename `eclair.bak` to a new local file
+- first, move `eclair.bak` to a new local file
 - then, backup the renamed file using whatever tool you like.
 
 ## Docker
--- a/README.md
+++ b/README.md
@@ -142,10 +142,19 @@ The files that you need to backup are located in your data directory. You must b
 - your seed (`seed.dat`)
 - your channel database (`eclair.bak` under directory `mainnet`, `testnet` or `regtest` depending on which chain you're running on)
 
-Your seed never changes once it has been created, but your channels will change whenever you receive our send payments. Eclair will
+Your seed never changes once it has been created, but your channels will change whenever you receive or send payments. Eclair will
 create and maintain a snapshot of its database, named `eclair.bak`, in your data directory, and update it when needed. This file is 
-always consistent and safe to use even when Eclair is running, and this is what you should backup (for example, you could 
-move/rename it and upload the renamed file to a dedicated backup system, ...)
+always consistent and safe to use even when Eclair is running, and this is what you should backup regularly, e.g. with a `cron` task.
+
+Note that depending on your filesystem, we recommend first moving `eclair.bak` to some temporary file before copying that file to your final backup location.
+
+You can configure an optional script/exe to be called by eclair once a new database snapshot has been created, using the following option:
+
+```
+eclair.backup-notify-script = "absolute-path-to-your-script"
+```
+
+Make sure that your script is executable and uses an absolute path name for `eclair.bak`
 
 ## Docker
 
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ eclair.bitcoind.rpcpassword=<your-mainnet-rpc-password-here>
 
 The files that you need to backup are located in your data directory. You must backup:
 - your seed (`seed.dat`)
-- your channel database (`eclair.bak` in`mainnet`, `testnet` or `regtest` depending on which chain you're running on)
+- your channel database (`eclair.bak` under directory `mainnet`, `testnet` or `regtest` depending on which chain you're running on)
 
 Your seed never changes once it is created, but your channels do change whenever you receive our send payments.
 `eclair.bak` is safe to backup even when your system is running. We recommend that you implement your backup 
--- a/README.md
+++ b/README.md
@@ -144,15 +144,13 @@ The files that you need to backup are located in your data directory. You must b
 
 Your seed never changes once it has been created, but your channels will change whenever you receive or send payments. Eclair will
 create and maintain a snapshot of its database, named `eclair.bak`, in your data directory, and update it when needed. This file is 
-always consistent and safe to use even when Eclair is running, and this is what you should backup regularly, e.g. with a `cron` task.
+always consistent and safe to use even when Eclair is running, and this is what you should backup regularly.
 
-You can configure an optional notification script/exe to be called by eclair once a new database snapshot has been created, using the following option:
-
+For example you could configure a `cron` task for your backup job. Or you could configure an optional notification script to be called by eclair once a new database snapshot has been created, using the following option:
 ```
 eclair.backup-notify-script = "absolute-path-to-your-script"
 ```
-
-Make sure that your script is executable and uses an absolute path name for `eclair.bak`
+Make sure that your script is executable and uses an absolute path name for `eclair.bak`.
 
 Note that depending on your filesystem, in your backup process we recommend first moving `eclair.bak` to some temporary file 
 before copying that file to your final backup location.
--- a/README.md
+++ b/README.md
@@ -202,6 +202,21 @@ eclair.bitcoind.rpcuser=<your-mainnet-rpc-user-here>
 eclair.bitcoind.rpcpassword=<your-mainnet-rpc-password-here>
 ```
 
+### Backups
+
+The files that you need to backup are located in your data directory. You must backup:
+- your seed (`seed.dat`)
+- your channel database (`eclair.bak` in`mainnet`, `testnet` or `regtest` depending on which chain you're running on)
+
+Your seed never changes once it is created, but your channels do change whenever you receive our send payments.
+`eclair.bak` is safe to backup even when your system is running. We recommend that you implement your backup 
+process in 2 steps:
+- first, rename `eclair.bak` to a new local file
+- then, backup the renamed file using whatever tool you like.
+
+
+
+
 ## Resources
 - [1] [The Bitcoin Lightning Network: Scalable Off-Chain Instant Payments](https://lightning.network/lightning-network-paper.pdf) by Joseph Poon and Thaddeus Dryja
 - [2] [Reaching The Ground With Lightning](https://github.com/ElementsProject/lightning/raw/master/doc/deployable-lightning.pdf) by Rusty Russell
--- a/README.md
+++ b/README.md
@@ -146,15 +146,17 @@ Your seed never changes once it has been created, but your channels will change 
 create and maintain a snapshot of its database, named `eclair.bak`, in your data directory, and update it when needed. This file is 
 always consistent and safe to use even when Eclair is running, and this is what you should backup regularly, e.g. with a `cron` task.
 
-Note that depending on your filesystem, we recommend first moving `eclair.bak` to some temporary file before copying that file to your final backup location.
-
-You can configure an optional script/exe to be called by eclair once a new database snapshot has been created, using the following option:
+You can configure an optional notification script/exe to be called by eclair once a new database snapshot has been created, using the following option:
 
 ```
 eclair.backup-notify-script = "absolute-path-to-your-script"
 ```
 
 Make sure that your script is executable and uses an absolute path name for `eclair.bak`
+
+Note that depending on your filesystem, in your backup process we recommend first moving `eclair.bak` to some temporary file 
+before copying that file to your final backup location.
+
 
 ## Docker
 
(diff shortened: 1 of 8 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -18,6 +18,8 @@ eclair {
 
   watcher-type = "bitcoind" // other *experimental* values include "electrum"
 
+  backup-file = "eclair.sqlite.backup"
```
```suggestion
  backup-file = "eclair.sqlite.bak"
```

Not sure why we need to make this configurable while other files (e.g. `eclair.sqlite`) have a fixed name.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -220,11 +220,11 @@ class Setup(datadir: File,
       routerTimeout = after(FiniteDuration(config.getDuration("router.init-timeout").getSeconds, TimeUnit.SECONDS), using = system.scheduler)(Future.failed(new RuntimeException("Router initialization timed out")))
       _ <- Future.firstCompletedOf(routerInitialized.future :: routerTimeout :: Nil)
 
+      chaindir = new File(datadir, chain)
```
pm47: There must be a better way to do this than to redefine the directory here. There was even a `TODO` comment that you missed when you moved it.

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,87 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class BackupHandler(databases: Databases, wip: File, destination: File)(implicit ec: ExecutionContext) extends Actor with ActorLogging {
+  import BackupHandler._
+
+  // we listen to ChannelPersisted events, which will trigger a backup
+  override def preStart(): Unit = {
+    super.preStart()
+    context.system.eventStream.subscribe(self, classOf[ChannelPersisted])
```
pm47: This may be better to suscribe to events here, but we do it differently in all other actors. Maybe stay consistent for now and do another PR to square all actors up at the same time?

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  override def unhandled(message: Any): Unit = {
+    super.unhandled(message)
+    log.warning(s"unhandled message $message")
```
pm47: Logging of unhandled messages should be done using akka conf.

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  // busy mode: there is already a backup in progress
+  // `again` tells us if we have to do another one when we're done with the current one
+  def busy(again: Option[ChannelPersisted]): Receive = {
+    case persisted: ChannelPersisted =>
+      log.info(s"replacing pending backup with a newer one")
+      context become busy(Some(persisted))
+
+    case Done if again.isDefined =>
```
pm47: I thought we agree to stop or at least limit doing several handlers for the same message? How about:

```scala
case Done =>
  again match {
    case Some(persisted) =>  
      doBackup(databases, wip, destination, again.get) pipeTo self
      context become busy(None)
    case None =>
      context become idle
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,87 @@
...
+      doBackup(databases, wip, destination, again.get) pipeTo self
+      context become busy(None)
+
+    case Done =>
+      context become idle
+
+      // we use the pipe pattern so errors are wrapped in akka.Status.Failure
+    case Status.Failure(cause) if again.isDefined =>
```
pm47: Have you considered using a dedicated thread for this actor instead?

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,87 @@
...
+      doBackup(databases, wip, destination, again.get) pipeTo self
+      context become busy(None)
+
+    case Status.Failure(cause) =>
+      log.error("database backup failed: {}", cause)
+      context become idle
+  }
+}
+
+object BackupHandler extends Logging {
```
pm47: Using contextless logging should be discouraged, in favor of passing `ActorLogging` parameter as implicit to companion object methods (like we did in `Commitment`).

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  case object Done
```
pm47: Why not use `akka.Done`?

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -45,6 +46,12 @@ object Databases {
     override val peers = new SqlitePeersDb(eclairJdbc)
     override val payments = new SqlitePaymentsDb(eclairJdbc)
     override val pendingRelay = new SqlitePendingRelayDb(eclairJdbc)
+    override def backup(file: File): Unit = {
```
pm47: This API makes the assumption that we will only use a single file to store the databases, which is a bit strange.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -345,8 +352,11 @@ class Setup(datadir: File,
 
 // @formatter:off
 sealed trait Bitcoin
+
```
pm47: Formatting issue

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+
+
+/**
+  * README !
+  * This actor will synchronously make a backup of the database it was initialized with whenever it receives
+  * a ChannelPersisted event.
+  * To avoid piling up messages and entering an endless backup loop, it is supposed to be used with a bounded mailbox
+  * with a single item:
+  *
+  * backup-mailbox {
```
pm47: Isn't there a way to enforce this configuration? That seems a bit error prone.

Overall the pattern is pretty nice.

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+  context.system.eventStream.subscribe(self, classOf[ChannelPersisted])
+
+  def receive = {
+    case persisted: ChannelPersisted =>
+      val start = System.currentTimeMillis()
+      databases.backup(wip)
+      val result = wip.renameTo(destination)
+      require(result, s"cannot rename $wip to $destination")
+      val end = System.currentTimeMillis()
+      log.info(s" database backup triggered by ${persisted.channelId} to $destination done in ${end - start} ms")
```
pm47:
```suggestion
      log.info(s"database backup triggered by channelId=${persisted.channelId} took ${end - start}ms")
```

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -234,7 +234,14 @@ class Setup(datadir: File,
       _ = wallet.getFinalAddress.map {
         case address => logger.info(s"initial wallet address=$address")
       }
-
+      backupHandler = system.actorOf(
+        SimpleSupervisor.props(
+          BackupHandler.props(
+            nodeParams.db,
+            new File(chaindir, "eclair.bak.wip"),
```
pm47:
```suggestion
            new File(chaindir, "eclair.bak.tmp"),
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+object BackupHandler {
+  def props(databases: Databases, wip: File, destination: File) = Props(new BackupHandler(databases: Databases, wip: File, destination: File)).withMailbox("eclair.backup-mailbox")
```
pm47:
```suggestion
  def props(databases: Databases, tmpFile: File, backupFile: File) = Props(new BackupHandler(databases: Databases, wip: File, destination: File)).withMailbox("eclair.backup-mailbox")
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+  *   mailbox-push-timeout-time = 0
+  * }
+  *
+  * Messages that cannot be processed will be sent to dead letters
+  *
+  * @param databases database to backup
+  * @param wip work-in-progress file
+  * @param destination destination file
+  */
+class BackupHandler(databases: Databases, wip: File, destination: File) extends Actor with ActorLogging {
```
pm47:
```suggestion
class BackupHandler(databases: Databases, tmpFile: File, backupFile: File) extends Actor with ActorLogging {
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+  * backup-mailbox {
+  *   mailbox-type = "akka.dispatch.BoundedMailbox"
+  *   mailbox-capacity = 1
+  *   mailbox-push-timeout-time = 0
+  * }
+  *
+  * Messages that cannot be processed will be sent to dead letters
+  *
+  * @param databases database to backup
+  * @param wip work-in-progress file
```
pm47:
```suggestion
  * @param tmpFile temporary file
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+  * @param destination destination file
```
pm47:
```suggestion
  * @param backupFile final backup file
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+  */
+class BackupHandler(databases: Databases, wip: File, destination: File) extends Actor with ActorLogging {
+
+  // we listen to ChannelPersisted events, which will trigger a backup
+  context.system.eventStream.subscribe(self, classOf[ChannelPersisted])
+
+  def receive = {
+    case persisted: ChannelPersisted =>
+      val start = System.currentTimeMillis()
+      databases.backup(wip)
```
pm47:
```suggestion
      databases.backup(tmpFile)
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+      val result = wip.renameTo(destination)
```
pm47:
```suggestion
      val result = tmpFile.renameTo(backupFile)
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+      require(result, s"cannot rename $wip to $destination")
```
pm47:
```suggestion
      require(result, s"cannot rename $tmpFile to $backupFile")
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
pm47: From [javadoc](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#renameTo(java.io.File)):

> Many aspects of the behavior of this method are inherently platform-dependent: The rename operation might not be able to move a file from one filesystem to another, **it might not be atomic**, and it might not succeed if a file with the destination abstract pathname already exists. The return value should always be checked to make sure that the rename operation was successful.

README.md
```diff
@@ -136,6 +136,18 @@ Eclair uses [`logback`](https://logback.qos.ch) for logging. To use a different
...
+#### Backup
+
+The files that you need to backup are located in your data directory. You must backup:
+- your seed (`seed.dat`)
+- your channel database (`eclair.bak` under directory `mainnet`, `testnet` or `regtest` depending on which chain you're running on)
+
+Your seed never changes once it is created, but your channels do change whenever you receive our send payments.
+`eclair.bak` is safe to backup even when your system is running. We recommend that you implement your backup 
+process in 2 steps:
+- first, rename `eclair.bak` to a new local file
```
pm47:
```suggestion
- first, move `eclair.bak` to a new local file
```

PR #952, opened 2019-04-17
Send to route

araspitzu:
Implement sending a payment to a pre-computed route, the route input format is compatible with the output of the `/findroute` API. As the route input format is a simple list of nodeId(s) it is mandatory that all the channels involved in the pre-computed route are public, thus sending to a private channel is not supported in this API. The behavior is very similar to other payments API, in a fire-and-forget fashion the endpoint immediately returns a UUID of the payment attempt and more information can then be retrieved via the websocket or payments APIs, the payment will not be retried with different routes (maxAttempts=1).


eclair-core/src/main/scala/fr/acinq/eclair/api/FormParamExtractors.scala
```diff
@@ -57,4 +57,10 @@ object FormParamExtractors {
     Timeout(str.toInt.seconds)
   }
 
+  implicit val pubkeyListUnmarshaller: Unmarshaller[String, List[PublicKey]] = Unmarshaller.strict { str =>
+    serialization.read[List[String]](str).map { el =>
```
pm47: Does that parse URL-encoded json?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
```diff
@@ -19,18 +19,18 @@ package fr.acinq.eclair.payment
...
+import fr.acinq.eclair.payment.PaymentLifecycle.{GenericSendPayment, SendPayment, SendPaymentToRoute}
 
 /**
   * Created by PM on 29/08/2016.
   */
 class PaymentInitiator(nodeParams: NodeParams, router: ActorRef, register: ActorRef) extends Actor with ActorLogging {
 
   override def receive: Receive = {
-    case c: SendPayment =>
+    case c:GenericSendPayment =>
```
pm47:
```suggestion
    case c: GenericSendPayment =>
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
```diff
@@ -19,18 +19,18 @@ package fr.acinq.eclair.payment
...
       val paymentId = UUID.randomUUID()
       val payFsm = context.actorOf(PaymentLifecycle.props(nodeParams, paymentId, router, register))
-      payFsm forward c
+      payFsm.forward(c)
```
pm47: Previous syntaxt was a convention we use elsewhere (to be close to `actor ! msg`)

```suggestion
      payFsm forward c
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -67,6 +67,7 @@ case class RouteRequest(source: PublicKey,
                         ignoreChannels: Set[ChannelDesc] = Set.empty,
                         routeParams: Option[RouteParams] = None)
 
+case class PartialRouteRequest(hops:Seq[PublicKey])
```
pm47: The name is misleading, it makes it seem like we have a partial route and we need the rest. How about (credits @sstone):

```suggestion
case class FinalizeRoute(hops:Seq[PublicKey])
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -45,6 +45,12 @@ class PaymentLifecycle(nodeParams: NodeParams, id: UUID, router: ActorRef, regis
   startWith(WAITING_FOR_REQUEST, WaitingForRequest)
 
   when(WAITING_FOR_REQUEST) {
+    case Event(SendPaymentToRoute(amountMsat, paymentHash, hops, finalCltvExpiry), WaitingForRequest) =>
```
pm47: nit: that's not consistent with the handler style below (I found that argument extraction is less maintenable, everytime you add an arg they don't compile).

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
```diff
@@ -19,18 +19,18 @@ package fr.acinq.eclair.payment
 import java.util.UUID
 import akka.actor.{Actor, ActorLogging, ActorRef, Props}
 import fr.acinq.eclair.NodeParams
-import fr.acinq.eclair.payment.PaymentLifecycle.SendPayment
+import fr.acinq.eclair.payment.PaymentLifecycle.{GenericSendPayment, SendPayment, SendPaymentToRoute}
```
pm47: Unused imports

```suggestion
import fr.acinq.eclair.payment.PaymentLifecycle.GenericSendPayment
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
araspitzu: I came up with the name because the route here is not complete, I like the suggestion though!

araspitzu: Addressed in 44acb29

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
araspitzu: I'm okay with this proposed change even though in general i like the compiler to do the heavy lifting and error-catching, but in the case of the extractor pattern the compile time errors are typically not very useful, addressed in 52e4938.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
araspitzu: Addressed in 3584079

eclair-core/src/main/scala/fr/acinq/eclair/api/FormParamExtractors.scala
araspitzu: There is no URL-encoding here, this only parses the form data parameter (as json array of strings) into a list of public keys.

araspitzu:
> Wouldn't that be possible to add a `goingThroughNodes` attributes to the existing `SendPayment`? Or have a `Hint` parameter that can be `NoHint`, `GoesThrough(nodeId: PublicKey)` or `StaticRoute(nodeIds: Seq[PublicKey])`

Yes that was an option but adding "hints" to `SendPayment` feels wrong because it would have a double meaning, on one hand it would refer to the BOLT11 hints found on the payment request, on the other we would store an entire static route in there. This would lead to some branching code where we take different action depending on the value of the hints, and i believe it's not the best approach in term of readability. The two use cases are captured by the duo SendPayment/SendPaymentToRoute.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -410,6 +411,13 @@ class Router(nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Prom
       sender ! d
       stay
 
+    case Event(FinalizeRoute(partialHops), d) =>
+      // split into sublists [(a,b),(b,c), ...] then get the edges between each of those pair, then select the largest edge between them
```
```suggestion
      // split into sublists [(a,b),(b,c), ...] then get the edges between each of those pairs, then select the largest edge between them
```

eclair-core/src/test/scala/fr/acinq/eclair/router/RouterSpec.scala
```diff
@@ -130,10 +130,10 @@ class RouterSpec extends BaseRouterSpec {
   test("handle bad signature for NodeAnnouncement") { fixture =>
     import fixture._
     val sender = TestProbe()
-    val buggy_ann_a = ann_a.copy(signature = ann_b.signature, timestamp = ann_a.timestamp + 1)
-    sender.send(router, PeerRoutingMessage(null, remoteNodeId, buggy_ann_a))
-    sender.expectMsg(TransportHandler.ReadAck(buggy_ann_a))
-    sender.expectMsg(InvalidSignature(buggy_ann_a))
+    val buggy_ann_b = ann_b.copy(signature = ann_c.signature, timestamp = ann_b.timestamp + 1)
```
pm47: Why did you need to change that?

araspitzu: Good catch that was some leftover from development, in 3aa79f7 i reverted the changes to RouterSpec and BaseRouterSpec (along with updating the copyright header).

araspitzu:
TODO: update doc

pm47:
Just realized there is no API-level test, ideally it would test both input formats (json and comma-separated):
- `["nodeid1", "nodeid2", ... ]`
- `nodeid1,nodeid2`

PR #954, opened 2019-04-18
Sync whitelist

pm47:
We will only sync with whitelisted peers. If the whitelist is empty then we sync with everyone.


eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -69,20 +60,35 @@ class PeerSpec extends TestkitBaseClass with StateTestsHelperMethods {
     val transport = TestProbe()
     val wallet: EclairWallet = new TestWallet()
     val remoteNodeId = Bob.nodeParams.nodeId
+
+    import com.softwaremill.quicklens._
+    val aliceParams = TestConstants.Alice.nodeParams
+      .modify(_.syncWhitelist).setToIf(test.tags.contains("sync-whitelist-bob"))(Set(remoteNodeId))
```
t-bast: nit: I would find this slightly better:
```suggestion
      .modify(_.syncWhitelist).setToIf(test.tags.contains("sync-whitelist-bob"))(Set(Bob.nodeParams.nodeId))
```

pm47: Problem is we reuse `remoteNodeId` a few lines below and it has to be the same value.

PR #957, opened 2019-04-19
resolve hostname ip in docker level

Kukks:
Good to note that this includes #956

This PR adds a new feature to the docker deployment.

By specifying a hostname through an environment variable name `PUBLIC_HOST`, it will resolve the ip and set the first public ip to it.


Kukks:
Seems like this is actually supported internally. Closing :)

PR #960, opened 2019-04-23
Backup: explicitely specify move options

sstone:
We now specify that we want to atomically overwrite the existing backup file with the new one (fixes
a potential issue on Windows).
We also publish a specific notification when the backup process has been completed.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -17,7 +17,7 @@ eclair {
   }
 
   // override this with a script/exe that will be called everytime a new database backup has been created
-  # backup-notify-script = ""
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -17,7 +17,7 @@ eclair {
   }
 
   // override this with a script/exe that will be called everytime a new database backup has been created
-  backup-notify-script = ""
+  # backup-notify-script = ""
```
dpad85: Could we provide an example instead of an empty string (which would not work anyway)?

```suggestion
  # backup-notify-script = "/path/to/script.sh"
```

PR #962, opened 2019-04-24
Add `balances` API endpoint

btcontract:
Currently balances can be obtained from `channels` call but this requires a lot of work on caller side and also some specific knowledge (reserves, commit tx fee, in-flight payments), so this new `balances` endpoint only returns a correct balance info for each channel. 

Data is taken from `Relayer` actor which already has exactly what we need: a map of `OutgoingChannel` objects which now contain correctly calculated `can send / can receive` balances for each `NORMAL` channel. Additionally, obtaining balances this way has a minimal performance hit when compared to querying each channel used in `channels`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -73,13 +73,13 @@ case class Commitments(localParams: LocalParams, remoteParams: RemoteParams,
 
   val announceChannel: Boolean = (channelFlags & 0x01) != 0
 
-  def availableBalanceForSendMsat: Long = {
+  lazy val availableBalanceForSendMsat: Long = {
     val reduced = CommitmentSpec.reduce(remoteCommit.spec, remoteChanges.acked, localChanges.proposed)
     val feesMsat = if (localParams.isFunder) Transactions.commitTxFee(Satoshi(remoteParams.dustLimitSatoshis), reduced).amount * 1000 else 0
     reduced.toRemoteMsat - remoteParams.channelReserveSatoshis * 1000 - feesMsat
   }
 
-  def availableBalanceForReceiveMsat: Long = {
+  lazy val availableBalanceForReceiveMsat: Long = {
     val reduced = CommitmentSpec.reduce(localCommit.spec, localChanges.acked, remoteChanges.proposed)
     val feesMsat = if (localParams.isFunder) 0 else Transactions.commitTxFee(Satoshi(localParams.dustLimitSatoshis), reduced).amount * 1000
     reduced.toRemoteMsat - localParams.channelReserveSatoshis * 1000 - feesMsat
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -40,7 +40,6 @@ case class PublishableTxs(commitTx: CommitTx, htlcTxsAndSigs: List[HtlcTxAndSigs
 case class LocalCommit(index: Long, spec: CommitmentSpec, publishableTxs: PublishableTxs)
 case class RemoteCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, remotePerCommitmentPoint: Point)
 case class WaitingForRevocation(nextRemoteCommit: RemoteCommit, sent: CommitSig, sentAfterLocalCommitIndex: Long, reSignAsap: Boolean = false)
-case class UsableBalances(localMsat: Long, remoteMsat: Long, isPublic: Boolean)
 // @formatter:on
 
 /**
@@ -74,12 +73,16 @@ case class Commitments(localParams: LocalParams, remoteParams: RemoteParams,
 
   val announceChannel: Boolean = (channelFlags & 0x01) != 0
 
-  def usableBalances: UsableBalances = {
+  def availableBalanceForSendMsat: Long = {
     val reduced = CommitmentSpec.reduce(remoteCommit.spec, remoteChanges.acked, localChanges.proposed)
-    val commitTxFee = Transactions.commitTxFee(Satoshi(remoteParams.dustLimitSatoshis), reduced).amount * 1000
-    val localBalance = reduced.toRemoteMsat - remoteParams.channelReserveSatoshis * 1000 - { if (localParams.isFunder) commitTxFee else 0 }
-    val remoteBalance = reduced.toLocalMsat - localParams.channelReserveSatoshis * 1000 - { if (localParams.isFunder) 0 else commitTxFee }
-    UsableBalances(localMsat = localBalance, remoteMsat = remoteBalance, isPublic = announceChannel)
+    val feesMsat = if (localParams.isFunder) Transactions.commitTxFee(Satoshi(remoteParams.dustLimitSatoshis), reduced).amount * 1000 else 0
+    reduced.toRemoteMsat - remoteParams.channelReserveSatoshis * 1000 - feesMsat
+  }
+
+  def availableBalanceForReceiveMsat: Long = {
+    val reduced = CommitmentSpec.reduce(localCommit.spec, localChanges.acked, remoteChanges.proposed)
+    val feesMsat = if (localParams.isFunder) 0 else Transactions.commitTxFee(Satoshi(remoteParams.dustLimitSatoshis), reduced).amount * 1000
+    reduced.toRemoteMsat - localParams.channelReserveSatoshis * 1000 - feesMsat
   }
 }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -40,6 +40,7 @@ case class PublishableTxs(commitTx: CommitTx, htlcTxsAndSigs: List[HtlcTxAndSigs
 case class LocalCommit(index: Long, spec: CommitmentSpec, publishableTxs: PublishableTxs)
 case class RemoteCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, remotePerCommitmentPoint: Point)
 case class WaitingForRevocation(nextRemoteCommit: RemoteCommit, sent: CommitSig, sentAfterLocalCommitIndex: Long, reSignAsap: Boolean = false)
+case class UsableBalances(localMsat: Long, remoteMsat: Long, isPublic: Boolean)
 // @formatter:on
 
 /**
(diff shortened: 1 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -73,10 +73,12 @@ case class Commitments(localParams: LocalParams, remoteParams: RemoteParams,
 
   def announceChannel: Boolean = (channelFlags & 0x01) != 0
 
-  def availableBalanceForSendMsat: Long = {
+  def availableBalances: (Long, Long) = {
```
dpad85: It's quite hard to understand what this tuple of `Long` is. I think that `availableBalances` should be typed, with something like:

```scala
case class AvailableBalance(local: MilliSatoshi, remote: MilliSatoshi)
```

pm47:
I am not against adding this feature, but this should be done independently of the `Relayer`, which should only have one job.

btcontract:
Even if that would mean precisely copying what `Relayer` does already? Reading off `Relayer` state seems innocent in a sense that it does not make `Relayer` change anything.

btcontract:
Now there's a `liveBalanceTracker` actor which is only created if API is enabled, will this work?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -40,6 +40,7 @@ case class PublishableTxs(commitTx: CommitTx, htlcTxsAndSigs: List[HtlcTxAndSigs
 case class LocalCommit(index: Long, spec: CommitmentSpec, publishableTxs: PublishableTxs)
 case class RemoteCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, remotePerCommitmentPoint: Point)
 case class WaitingForRevocation(nextRemoteCommit: RemoteCommit, sent: CommitSig, sentAfterLocalCommitIndex: Long, reSignAsap: Boolean = false)
+case class AvailableBalances(localMsat: Long, remoteMsat: Long, isPublic: Boolean)
```
pm47: The `isPublic` attribute has nothing to do with balances.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -71,12 +72,14 @@ case class Commitments(localParams: LocalParams, remoteParams: RemoteParams,
 
   def addRemoteProposal(proposal: UpdateMessage): Commitments = Commitments.addRemoteProposal(this, proposal)
 
-  def announceChannel: Boolean = (channelFlags & 0x01) != 0
+  val announceChannel: Boolean = (channelFlags & 0x01) != 0
```
pm47: :+1:

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
```diff
@@ -333,4 +329,10 @@ object Relayer {
       case _ => requestedShortChannelId // we don't have a channel_update for this short_channel_id
     }
   }
+
+  def updateChannelUpdatesIfPresent(channelUpdates: Map[ShortChannelId, OutgoingChannel], shortChannelId: ShortChannelId, commitments: Commitments): Map[ShortChannelId, OutgoingChannel] =
```
pm47: If this is used from several different packages then it probably doesn't belong here. There must be a better way to factor this.

btcontract:
> But more importantly this does not return all balances: it will ignore channel being created, or in the process of closing. The logic from Relayer was custom tailored for relaying payments and doesn't take care of all those cases.

It also ignores offline channels and this is by design, the purpose of this method as I see is to return exactly what can be sent and received right now (including a possibility to ignore private channels are they are useless for routing).

pm47:
> Use case for this is to show a service user how much can be received/withdrawn from a service at this very moment.

Oh, I see! So it is basically about `availableForSend`/`availableForReceive` at one point in time. With that in mind I have far less concern with putting this in the `Relayer`.

> Maybe "balances" is a wrong word for it.

Yup, the name is clearly misleading. How about `availablebalance` or `usablebalance`? Not very satisfactory.

btcontract:
Is some kind of test needed for this?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -71,12 +72,14 @@ case class Commitments(localParams: LocalParams, remoteParams: RemoteParams,
...
 
-  def availableBalanceForSendMsat: Long = {
+  def usableBalances: UsableBalances = {
     val reduced = CommitmentSpec.reduce(remoteCommit.spec, remoteChanges.acked, localChanges.proposed)
-    val feesMsat = if (localParams.isFunder) Transactions.commitTxFee(Satoshi(remoteParams.dustLimitSatoshis), reduced).amount * 1000 else 0
-    reduced.toRemoteMsat - remoteParams.channelReserveSatoshis * 1000 - feesMsat
+    val commitTxFee = Transactions.commitTxFee(Satoshi(remoteParams.dustLimitSatoshis), reduced).amount * 1000
+    val localBalance = reduced.toRemoteMsat - remoteParams.channelReserveSatoshis * 1000 - { if (localParams.isFunder) commitTxFee else 0 }
```
pm47: `localBalance` needs to be computed from the local commitment (see the difference ine the reduced commitment in `sendAdd` and `receiveAdd`).

How about:
1) keep `availableBalanceForSendMsat` as it was
2) add an `availableBalanceForReceiveMsat` that is symmetrical (it doesn't duplicate code because the commitment is computed differently)
3) define the `UsableBalance` class in `Relayer`, and make it contain the `isFunder` boolean and `remoteNodeId` and `channel_update`, so that we don't mix concerns.

pm47:
Hey @btcontract, are you still on this?

btcontract:
Yes!

btcontract:
Actually just started, I don't yet understand why `UsableBalance` with `isFunder` boolean, `remoteNodeId` and `channel_update` is needed in `Relayer` if there is already defined a 
```scala
case class OutgoingChannel(nextNodeId: PublicKey, channelUpdate: ChannelUpdate ...
```

Perhaps it should be something like this in `Relayer`:
```scala
case class UsableBalances(canSend: Long, canReceive: Long, isPublic: Boolean)
case class OutgoingChannel(nextNodeId: PublicKey, channelUpdate: ChannelUpdate, usableBalances: UsableBalances)```

Basically almost like it is now but `UsableBalances` moved to `Relayer`?

**Comment by pm47** [comment]
I wrote `isFunder` but I meant `isPublic`, does that make more sense?

The goal is to separate concerns and keep internal classes clean. `UsableBalance` would be an "API" class and you can put things as distinct as `availableBalanceForSendMsat` and `isPublic` in it.

**Comment by btcontract** [comment]
OK, but where should `UsableBalance` be located, still inside `OutgoingChannel` or separate map in `Relayer` (or something else)?

File: eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala [hunk 6]
```diff
@@ -74,12 +73,16 @@ case class Commitments(localParams: LocalParams, remoteParams: RemoteParams,
...
-    val localBalance = reduced.toRemoteMsat - remoteParams.channelReserveSatoshis * 1000 - { if (localParams.isFunder) commitTxFee else 0 }
-    val remoteBalance = reduced.toLocalMsat - localParams.channelReserveSatoshis * 1000 - { if (localParams.isFunder) 0 else commitTxFee }
-    UsableBalances(localMsat = localBalance, remoteMsat = remoteBalance, isPublic = announceChannel)
+    val feesMsat = if (localParams.isFunder) Transactions.commitTxFee(Satoshi(remoteParams.dustLimitSatoshis), reduced).amount * 1000 else 0
+    reduced.toRemoteMsat - remoteParams.channelReserveSatoshis * 1000 - feesMsat
+  }
+
+  def availableBalanceForReceiveMsat: Long = {
+    val reduced = CommitmentSpec.reduce(localCommit.spec, localChanges.acked, remoteChanges.proposed)
+    val feesMsat = if (localParams.isFunder) 0 else Transactions.commitTxFee(Satoshi(remoteParams.dustLimitSatoshis), reduced).amount * 1000
```
pm47: That's wrong, you should use the fees as computed in `receiveAdd`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
```diff
@@ -66,27 +69,35 @@ class Relayer(nodeParams: NodeParams, register: ActorRef, paymentHandler: ActorR
 
   private val commandBuffer = context.actorOf(Props(new CommandBuffer(nodeParams, register)))
 
-  override def receive: Receive = main(Map.empty, new mutable.HashMap[PublicKey, mutable.Set[ShortChannelId]] with mutable.MultiMap[PublicKey, ShortChannelId])
+  override def receive: Receive = main(Map.empty, new mutable.HashMap[PublicKey, mutable.Set[ShortChannelId]] with mutable.MultiMap[PublicKey, ShortChannelId], Map.empty)
 
-  def main(channelUpdates: Map[ShortChannelId, OutgoingChannel], node2channels: mutable.HashMap[PublicKey, mutable.Set[ShortChannelId]] with mutable.MultiMap[PublicKey, ShortChannelId]): Receive = {
+  def main(channelUpdates: Map[ShortChannelId, OutgoingChannel],
```
pm47: I'd rather redefine `OutgoingChannel` as:

```scala
case class OutgoingChannel(nextNodeId: PublicKey, channelUpdate: ChannelUpdate, commitments: Commitments)
```
Change is minimal, and then you can implement `GetUsableBalances` in a self-contained way by iterating over the `channelUpdates` map.

pm47:
Nice, now it is just missing a simple test in `RelayerSpec`.

eclair-core/src/test/scala/fr/acinq/eclair/TestUtils.scala
```diff
@@ -28,4 +31,9 @@ object TestUtils {
     .get("buildDirectory") // this is defined if we run from maven
     .getOrElse(new File(sys.props("user.dir"), "target").getAbsolutePath) // otherwise we probably are in intellij, so we build it manually assuming that user.dir == path to the module
 
+  def makeCommitments(channelId: ByteVector32, availableBalanceMsat: Long = 50000000L) =
```
pm47: This doesn't belong here (this is for methods pertaining to the test environment itself) but probably in `HtlcGenerationSpec` with the reset of the generated content.

eclair-core/src/test/scala/fr/acinq/eclair/payment/RelayerSpec.scala
```diff
@@ -416,4 +411,11 @@ class RelayerSpec extends TestkitBaseClass {
     assert(fwd.channelId === origin.originChannelId)
     assert(fwd.message.id === origin.originHtlcId)
   }
+
+  test("get usable balances") { f =>
+    import f._
+    val sender = TestProbe()
+    sender.send(relayer, GetUsableBalances)
```
pm47: Come on, this test is meaningless, the list is empty.

At the very least (and this doesn't even test e.g. inversion of send/receive amounts):

```scala
  test("get usable balances") { f =>
    import f._
    val sender = TestProbe()
    relayer ! LocalChannelUpdate(null, channelId_ab, channelUpdate_ab.shortChannelId, a , None, channelUpdate_ab, makeCommitments(channelId_ab))
    relayer ! LocalChannelUpdate(null, channelId_bc, channelUpdate_bc.shortChannelId, c, None, channelUpdate_bc, makeCommitments(channelId_bc))
    sender.send(relayer, GetUsableBalances)
    assert(sender.expectMsgType[Iterable[UsableBalances]].size === 2)
  }
```

btcontract: Perhaps I should revise my notion of really simple test :)

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -71,13 +71,19 @@ case class Commitments(localParams: LocalParams, remoteParams: RemoteParams,
...
-  def announceChannel: Boolean = (channelFlags & 0x01) != 0
+  val announceChannel: Boolean = (channelFlags & 0x01) != 0
 
   def availableBalanceForSendMsat: Long = {
     val reduced = CommitmentSpec.reduce(remoteCommit.spec, remoteChanges.acked, localChanges.proposed)
     val feesMsat = if (localParams.isFunder) Transactions.commitTxFee(Satoshi(remoteParams.dustLimitSatoshis), reduced).amount * 1000 else 0
     reduced.toRemoteMsat - remoteParams.channelReserveSatoshis * 1000 - feesMsat
   }
+
+  def availableBalanceForReceiveMsat: Long = {
```
pm47: Can you make this a `lazy val`? That's my last comment.

btcontract:
Yes, I have forgot to override `val`, not `def` in tests. What do you think about this though: https://github.com/btcontract/eclair/commit/1f18c597c9515fe676ad683a162dffc173164776

An idea is to move subtle computations in a same place (commitments object) instead of repeating them, tests pass locally with this change.

araspitzu:
TODO: once merged, update the docs

pm47:
@btcontract sounds like premature optimization, let's keep it for later

btcontract:
@pm47 An idea was not so much to save on computation but rather to remove duplicate logic as I've made a couple of mistakes while trying to copy it. Anyway, tests are fixed now.

PR #971, opened 2019-04-26
Various fix and improvements in time/timestamp handling

### LINKED ISSUE #970: Audit Database timestamp has inconsistent units.
The sent/received/relayed tables the timestamp is seconds.
The network_fees/channel_events/balance_updated table is milliseconds.

araspitzu:
This PR standardizes the way we compute the current time as unix timestamp 

- Scala's Platform is used and the conversion is done via scala's concurrent.duration facilities
- Java's Instant has been replaced due to broken compatibility with android
- AuditDB events use milliseconds (fixes #970)
- PaymentDB events use milliseconds
- Query filters for AuditDB and PaymentDB use seconds

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
@@ -29,6 +29,11 @@ import concurrent.duration._
 import scala.collection.immutable.Queue
 import scala.compat.Platform
 
+/**
+  * AUDIT DB stores timestamp in milliseconds instead of seconds, objects instances will still have the timestamp in seconds but
+  * when written/read from the DB a conversion must occur, all conversions are handled in this class.
+  * @param sqlite
+  */
 class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
 
   import SqliteUtils._
@@ -86,7 +91,7 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
       statement.setLong(3, e.localBalanceMsat)
       statement.setLong(4, e.commitments.commitInput.txOut.amount.toLong)
       statement.setLong(5, e.commitments.remoteParams.channelReserveSatoshis) // remote decides what our reserve should be
-      statement.setLong(6, Platform.currentTime.milliseconds.toSeconds)
+      statement.setLong(6, Platform.currentTime.milliseconds.toMillis) // in audit DB we store timestamps as milliseconds
       statement.executeUpdate()
     }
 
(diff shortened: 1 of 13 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -87,7 +86,7 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
       statement.setLong(3, e.localBalanceMsat)
       statement.setLong(4, e.commitments.commitInput.txOut.amount.toLong)
       statement.setLong(5, e.commitments.remoteParams.channelReserveSatoshis) // remote decides what our reserve should be
-      statement.setLong(6, Platform.currentTime)
+      statement.setLong(6, Platform.currentTime.milliseconds.toSeconds)
```
pm47: This should be `.toMilliseconds`

araspitzu: Okay so audit DB unfortunately has milliseconds timestamp in it, commit 294cb0a  makes sure we still store then in that time unit but when the objects are instantiated we convert the millis to actual seconds.

araspitzu:
@n1bor

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -180,7 +181,7 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
     using(sqlite.prepareStatement(queryStmt)) { statement =>
       statement.setLong(1, from)
       statement.setLong(2, to)
-      if (pendingOnly) statement.setLong(3, Instant.now().getEpochSecond)
+      if (pendingOnly) statement.setLong(3, Platform.currentTime.milliseconds.toSeconds)
```
pm47: Shouldn't that be `Platform.currentTime.milliseconds.toMillis`?

araspitzu: No because Payment Request is a protocol object and use seconds, we also store "expiry" in seconds and the "expire_at" field of the payment db uses seconds as time unit [here](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala#L139).

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -760,7 +760,7 @@ object Router {
   def isStale(u: ChannelUpdate): Boolean = {
     // BOLT 7: "nodes MAY prune channels should the timestamp of the latest channel_update be older than 2 weeks (1209600 seconds)"
     // but we don't want to prune brand new channels for which we didn't yet receive a channel update
-    val staleThresholdSeconds = Platform.currentTime / 1000 - 1209600
+    val staleThresholdSeconds = Platform.currentTime.milliseconds.toSeconds - 1209600
```
```suggestion
    val staleThresholdSeconds = (Platform.currentTime.milliseconds - 14.days).toSeconds
```

PR #972, opened 2019-04-26
Extended Queries: use TLV format for optional data

sstone:
The RFC PR has already been updated, and this adds more flexibility as node that don't understand some fields can just skip over them. Timestamps and checksums have also been separated.


pm47:
This probably need to be rebased from #1045.

t-bast:
Yes, I discussed it with @sstone and he said he would rebase this on top of #1045 once #1045 is stabilized (hopefully today).

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
```diff
@@ -222,77 +220,163 @@ case class ChannelUpdate(signature: ByteVector,
...
+case class QueryChannelRangeExtension(flag: Byte) extends Tlv {
+  override val `type` = UInt64(1)
+
+  val wantTimestamps = QueryChannelRangeExtension.wantTimestamps(flag)
+
+  val wantChecksums = QueryChannelRangeExtension.wantChecksums(flag)
+}
+
+case object QueryChannelRangeExtension {
+  val WANT_TIMSTAMPS: Byte = 1
```
pm47:
```suggestion
  val WANT_TIMESTAMPS: Byte = 1
```

PR #974, opened 2019-04-28
[WIP] Add logging of database timings

n1bor:
Been adding metric of my node using logstash to scrape the log files and push the number to graphite/grafana and works well.

This is a proposal for including database timings in the logs so we can monitor that too. Is very light and adds very little code.

If you agree this is a good approach will update the other database classes too.

With this I could add time spent on database to https://lightningconductor.net/grafana/d/zHCFILzWz/channel-actor-details?orgId=1


n1bor:
Build failure looks like a travis issue?
Another build has same error:
https://travis-ci.org/ACINQ/eclair/builds/525413389

n1bor:
@pm any thoughts on this? If agree way to go will add to all classes. As I said logstash is great for realtime parsing the logs and pushing to whatever analysis tool you want.

pm47:
I'm working on an alternative based on [Kamon](https://kamon.io/), which is more powerful than simple metrics. Functionnally it will include what you propose here.

t-bast:
This PR is superceded by https://github.com/ACINQ/eclair/pull/1126.
We will gradually add more metrics / spans as we test them under load.

PR #975, opened 2019-04-29
Fixed overflow issue with max duration

pm47:
This is a regression caused by #971, because `Duration` has a max value of `Long.MaxValue` *nanoseconds*, not *seconds*.


araspitzu:
LGTM.

PR #976, opened 2019-04-29
Variable-length onion payloads

t-bast:
This pull request implements the variable-length onion proposal submitted in https://github.com/lightningnetwork/lightning-rfc/pull/619

I also took this opportunity to heavily refactor our onion code:

- Moved everything encoding-related to `scodec`
- Split the contents of the Sphinx object into sub-objects (in particular this will let us easily add support for onions of different sizes such as trampoline onions)
- Renamed most methods to (hopefully) make them clearer
- We only returned `BAD_ONION` errors and not the more fine-grained errors detailed in the spec -> that's now fixed
- Added more tests
- Removed the misleading `LAST_PACKET`: this was misleading because this was the (invalid) next packet created when unwrapping the last packet, not actually the last packet
- I considered using the same abstraction for the payment onion and the failure onion, but unfortunately they have very different mechanisms: I think it would be a bad idea to try to fit them in the same abstraction

There are two minor spec details that may change depending on the discussion on the spec PR:

- The bits used for the feature flag will likely align with the feature bit unification effort
- The termination condition may (or not) be simplified (to only the hmac being `0x00`)

These are fairly minor changes so I think you can start reviewing now to fix all eclair/scala specific stuff @pm47 @sstone


t-bast:
We discussed yesterday a new proposal to use variable-length frames.
This will require many changes and make most old commits obsolete, so hold on before you review - the next commit will change everything ;)

t-bast:
For some reason Travis doesn't pick up this branch, I'll delete it and re-push.

t-bast:
@pm47 @sstone this is now ready for review :tada: 
The code is up-to-date with the latest spec proposal (see https://github.com/lightningnetwork/lightning-rfc/pull/619) and the official test vector works.

pm47:
Is there any dependency between this PR and #1045?

t-bast:
Yes we need the `varint` codec (for the length of each payload).
There is some new discussion on the onion spec PR to potentially re-add the `realm` byte which would require a small change to the current implementation, I hope we resolve this during tonight's meeting.
I'll rebase this PR on top of #1045 once #1045 is stabilized.

pm47:
Ok, is this still ready for review?

t-bast:
I think most of the code will not change, the only changes still being discussed would be keeping a first byte to indicate `realm` and potentially defining a new stop condition. Those would be small-ish changes added to another commit on this PR.
But it might be worth waiting until tomorrow to start reviewing, I'll update this PR with a summary of the IRC meeting and an evaluation of whether this is worth reviewing as-is or will need many changes :)

t-bast:
We addressed all open design comments during yesterday's meeting, only a few typos and clarifications should be needed to be able to merge the spec PR (unless something new comes out).

I think we all agreed yesterday that this is how variable-length onion will be merged, so it should be safe to review now and rebase on top of this if you need variable-length onions.

t-bast:
I just rebased on master, this should be easier to review now.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -36,12 +35,13 @@ object Features {
   val CHANNEL_RANGE_QUERIES_BIT_MANDATORY = 6
   val CHANNEL_RANGE_QUERIES_BIT_OPTIONAL = 7
 
+  val OPTION_VARIABLE_LENGTH_ONION_MANDATORY = 0
```
pm47: I guess that's still TBD.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Mac.scala
```diff
@@ -0,0 +1,59 @@
...
+import scodec.bits.ByteVector
+
+/**
+  * Created by t-bast on 04/07/19.
+  */
+
+/**
+  * Create and verify message authentication codes.
+  */
+trait Mac {
```
pm47: Given that you constrained the `mac` to be 32 bytes:

```suggestion
trait Mac32 {
```

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: Yes exactly, there are two small moving parts that are pending comments from Christian on the spec, I detailed that in the description of this PR.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Mac.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  def verify(mac: ByteVector32, message: ByteVector): Boolean
+
+}
+
+case class Hmac256(key: ByteVector) extends Mac {
+
+  override def mac(message: ByteVector): ByteVector32 = Mac.hmac256(key, message)
+
+  override def verify(mac: ByteVector32, message: ByteVector): Boolean = this.mac(message) === mac
```
pm47: Looking at scodec javadoc, triple equals seems to be the correct choice, but we are using double equals everywhere else and never had issues. Not sure what's the best practice here.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -74,298 +48,324 @@ object Sphinx extends Logging {
...
+    payload.head match {
+      case 0 =>
+        // The 1.0 BOLT spec used 65-bytes frames inside the onion payload.
+        // The first byte of the frame (called `realm`) is set to 0x00, followed by 32 bytes of per-hop data, followed by a 32-bytes mac.
+        65
+      case _ =>
+        // The 1.1 BOLT spec changed the frame format to use variable-length per-hop payloads.
+        // The first bytes contain a varint encoding the length of the payload data (not including the trailing mac).
+        // Since messages are always smaller than 65535 bytes, this varint will either be 1 or 3 bytes long.
+        val dataLength = CommonCodecs.varintoverflow.decode(BitVector(payload.take(3))).require.value.toInt
```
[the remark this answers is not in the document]
pm47: How about a rewrite to avoid leaking varint-specific details into this:

```scala
val lengthPrefix = payload.take(3)
val decodeResult = CommonCodecs.varintoverflow.decode(lengthPrefix.bits).require
val dataLength = decodeResult.value
val varintLength = lengthPrefix.length - decodeResult.remainder.length
(varintLength + dataLength + MacLength).toInt
```

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Mac.scala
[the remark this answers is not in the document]
t-bast: It should be slightly faster, but probably by a rather insignificant amount.
Should I revert to `==` to harmonize with the rest of the codebase?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -36,35 +32,13 @@ import scala.util.{Failure, Success, Try}
...
 
-  // length of a payload: 33 bytes (1 bytes for realm, 32 bytes for a realm-specific packet)
-  val PayloadLength = 33
-
-  // max number of hops
-  val MaxHops = 20
-
-  // onion packet length
-  val PacketLength = 1 + 33 + MacLength + MaxHops * (PayloadLength + MacLength)
+  def mac(key: ByteVector, message: ByteVector): ByteVector32 = Mac.hmac256(key, message)
```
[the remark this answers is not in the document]
t-bast: Premature optimization :D
To avoid creating objects all the time (creating more garbage) when I could just call the method directly.
But if you prefer I change that.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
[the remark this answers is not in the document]
t-bast: Good suggestion, I'll update.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -74,298 +48,324 @@ object Sphinx extends Logging {
...
+      * Create an encrypted onion packet that contains payloads for all nodes in the list.
+      *
+      * @param sessionKey     session key.
+      * @param publicKeys     node public keys (one per node).
+      * @param payloads       payloads (one per node).
+      * @param associatedData associated data.
+      * @return An onion packet with all shared secrets. The onion packet can be sent to the first node in the list, and
+      *         the shared secrets (one per node) can be used to parse returned failure messages if needed.
+      */
+    def create(sessionKey: PrivateKey, publicKeys: Seq[PublicKey], payloads: Seq[ByteVector], associatedData: ByteVector32): PacketAndSecrets = {
```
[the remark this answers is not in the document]
t-bast: How would you check that? Each payload may have a different size...

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -74,298 +48,324 @@ object Sphinx extends Logging {
...
-    require(packet.length == ErrorPacketLength, s"invalid error packet length ${packet.length}, must be $ErrorPacketLength")
-    val key = generateKey("ammag", sharedSecret)
-    val stream = generateStream(key, ErrorPacketLength)
-    logger.debug(s"ammag key: $key")
-    logger.debug(s"error stream: $stream")
-    packet xor stream
-  }
+  object PaymentPacket extends OnionPacket {
+
+    override val PayloadLength = 1300
```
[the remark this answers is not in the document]
t-bast: It's not really halfway-through, because `PaymentPacket` isn't constrained to `PerHopPayload`s.
`PaymentPacket` is simply the packet that's used with `update_add_htlc` but it can either contain `PerHopPayload`s, variable-length payloads (tlv streams), or a mixture of both.

The way I view this Sphinx component is:

- it only deals with `bytes` (not typed payloads)
- `OnionPacket` implements the algorithm from the sphinx paper (where a packet starts with a version, then public key, then an onion-encrypted payload, then a mac)
- The sphinx paper doesn't constrain the size of the onion-encrypted payload at all, but it needs it to correctly generate the filler
- Payment packets will always have a fixed size of 1300 bytes
- Trampoline packets will always have a fixed size of 692 bytes
- If we implement Hornet at some point, we will want a lot of different packet sizes

I used that extension/inheritance scheme to be able to inject the `PayloadLength` value (necessary in all the computations inside `OnionPayload` methods) because the only other alternative I saw was to add the `payloadLength` as an argument of each function, which felt weird.

Once the computation is done it doesn't matter if the result is a trampoline onion or a payment onion as the higher layers should only treat it as a blob, so it makes sense that all returns an `OnionPacket` (imo).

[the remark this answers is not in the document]
t-bast: I do agree that the architecture may look complex, maybe it's because I'm not experience enough in Scala to find a better pattern...but we can refactor this later when we find one.

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
```diff
@@ -29,7 +31,8 @@ import scodec.Attempt
 
 // @formatter:off
 sealed trait FailureMessage { def message: String }
-sealed trait BadOnion extends FailureMessage { def onionHash: ByteVector32 }
+sealed trait FailureCode { def failureCode: Int }
```
pm47: Since all `failure_message`s have a `failure_code`, it should be:

```scala
sealed trait FailureMessage {
  def code: Int
  def message: String
}
```
But then we have to define `failureCode` for each message, right? And then it gets nasty...

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
```diff
@@ -38,9 +41,22 @@ case object InvalidRealm extends Perm { def message = "realm was not understood
...
+}
+case class InvalidOnionHmac(onionHash: ByteVector32) extends BadOnion with Perm {
+  def failureCode = FailureMessageCodecs.BADONION | FailureMessageCodecs.PERM | 5
+  def message = "onion HMAC was incorrect when it reached the processing node"
+}
+case class InvalidOnionKey(onionHash: ByteVector32) extends BadOnion with Perm {
+  def failureCode = FailureMessageCodecs.BADONION | FailureMessageCodecs.PERM | 6
+  def message = "ephemeral key was unparsable by the processing node"
+}
+case class InvalidOnionUnknown(onionHash: ByteVector32) extends BadOnion with Perm {
```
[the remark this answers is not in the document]
t-bast: I found it strange that Bolt 4 doesn't have an explicit error for that...so indeed to be compliant we fallback to the generic BadOnion error. Since the trait is already named `BadOnion` it causes issues when I rename this to `BadOnion`...but maybe simply `InvalidOnion`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
```diff
@@ -215,34 +214,33 @@ object Relayer {
...
+      case Right(p@Sphinx.DecryptedPacket(payload, nextPacket, _)) =>
+        LightningMessageCodecs.perHopPayloadCodec.decode(BitVector(payload)) match {
+          case Attempt.Successful(DecodeResult(perHopPayload, _)) if p.isLastPacket =>
+            Right(FinalPayload(add, perHopPayload))
+          case Attempt.Successful(DecodeResult(perHopPayload, _)) =>
+            Right(RelayPayload(add, perHopPayload, nextPacket))
+          case Attempt.Failure(_) =>
+            Left(InvalidOnionUnknown(Sphinx.PaymentPacket.hash(add.onionRoutingPacket)))
+        }
+      case Left(badOnion) => Left(badOnion)
```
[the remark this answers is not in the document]
t-bast: I can't get this to compile (it complains about the right type not being the same)...

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
```diff
@@ -222,6 +222,11 @@ case class ChannelUpdate(signature: ByteVector64,
   require(((messageFlags & 1) != 0) == htlcMaximumMsat.isDefined, "htlcMaximumMsat is not consistent with messageFlags")
 }
 
+case class OnionPacket(version: Int,
```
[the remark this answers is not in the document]
t-bast: That touches on the layering we discussed indeed...
I want to keep the business logic associated to that guy in Sphinx.scala while defining the codecs in `wire` which splits a bit the code.
I can definitely move those two in a separate file.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
```diff
@@ -222,6 +222,11 @@ case class ChannelUpdate(signature: ByteVector64,
...
+                       publicKey: ByteVector,
```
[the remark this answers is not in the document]
t-bast: Because I need to return a specific error if that cannot be parsed as a public key (spec compliance) and I think this logic shouldn't bleed in the encoding layer here (that error is returned in Sphinx.scala when we convert to a public key).

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
```diff
@@ -125,4 +126,23 @@ object CommonCodecs {
 
   def zeropaddedstring(size: Int): Codec[String] = fixedSizeBytes(32, utf8).xmap(s => s.takeWhile(_ != '\u0000'), s => s)
 
+  /**
+    * When encoding, prepend a valid mac to the output of the given codec.
+    * When decoding, verify that a valid mac is prepended.
+    */
+  def prependmac[A](codec: Codec[A], mac: Mac) = Codec[A](
```
pm47: Alternate proposal:

```scala
def prependmac[A](codec: Codec[A], mac: Mac) = Codec[A](
    (a: A) => codec.encode(a).map(bits => mac.mac(bits.bytes).bits ++ bits),
    (bits: BitVector) => ("mac" | bytes32).decode(bits) match {
      case Attempt.Successful(DecodeResult(msgMac, remainder)) if mac.verify(msgMac, remainder.bytes) => codec.decode(remainder)
      case Attempt.Successful(_) => Attempt.Failure(scodec.Err("invalid mac"))
      case Attempt.Failure(err) => Attempt.Failure(err)
    }
  )
```

eclair-core/src/test/scala/fr/acinq/eclair/wire/LightningMessageCodecsSpec.scala
```diff
@@ -102,29 +92,40 @@ class LightningMessageCodecsSpec extends FunSuite {
...
+      val payload2 = perHopPayloadCodec.decode(bin1.toBitVector).require.value
       assert(payload2 === payload1)
     }
   }
 
   test("decode channel_update with htlc_maximum_msat") {
     // this was generated by c-lightning
     val bin = hex"010258fff7d0e987e2cdd560e3bb5a046b4efe7b26c969c2f51da1dceec7bcb8ae1b634790503d5290c1a6c51d681cf8f4211d27ed33a257dcc1102862571bf1792306226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf188910f0005a100000200005bc75919010100060000000000000001000000010000000a000000003a699d00"
-    val update = LightningMessageCodecs.lightningMessageCodec.decode(BitVector(bin.toArray)).require.value.asInstanceOf[ChannelUpdate]
+    val update = lightningMessageCodec.decode(bin.toBitVector).require.value.asInstanceOf[ChannelUpdate]
```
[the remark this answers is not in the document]
t-bast: I agree, it's shorter :)

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
t-bast: Yeah exactly, the reason we have to do this for onion errors is that the spec is a bit messy on that.
The only place where we really need that mapping in the other direction is for BadOnion because it's used in FailureMessage and in the malformed onion message.
The other FailureMessages don't need that "reverse lookup" so I didn't bother doing it for all.
But it wouldn't be a lot of work, if you want I can harmonize this.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Mac.scala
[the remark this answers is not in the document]
pm47: This is very small so let's keep it, and fix the rest of the codebase later.

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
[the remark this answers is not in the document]
pm47: Ok, but that's still not compliant right? There is no spec-defined message with that code *and* an enclosed `sha256_of_onion`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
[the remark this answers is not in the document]
pm47: Right, should be:
```scala
case l@Left(_) => l
```

Nevermind, it's ugly.

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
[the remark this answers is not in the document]
t-bast: You're right, but in fact we're never sending any `BadOnion` failures in failure messages because when a `BadOnion` error occurs this should be an `update_fail_malformed_htlc` that's returned instead.
Unless I'm missing something it seems like the spec is confusing here because Bolt 2 separates `update_fail_malformed_htlc` from `update_fail_htlc` and Bolt 4 makes it look like there's no separation between those two.
Should we update the spec in Bolt 4 to remove all the `BadOnion` errors from the failure message and mention that all BadOnion errors should instead use only the `update_fail_malformed_htlc` that needs `sha_of_onion` and `failureCode`?

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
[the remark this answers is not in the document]
pm47: Isn't that an argument for a tighter coupling between the onion and the codec?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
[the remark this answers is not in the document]
pm47: Right, they are padded to `PayloadLength`. Couldn't we at least check that they are smaller than that?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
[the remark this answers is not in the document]
pm47: Well I saw that you were using it somewhere else. Otherwise I was wondering why introduce this abstraction in the first place. I think you can keep your optimization and still use the interface.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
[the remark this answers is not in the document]
t-bast: I don't know, for crypto stuff I think it makes sense that the wire layer splits the different fields but doesn't interpret them, and lets another layer interpret the fields...

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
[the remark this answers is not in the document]
t-bast: Yes definitely, we can (and should) check that a single payload can't exceed the maximum length. I'll add that check.

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
pm47: The problem is that the result would probably not look good, we would have to define companion objects to all classes (or use the `FailureMessage` object to store the bindings).

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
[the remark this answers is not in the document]
t-bast: The interface is necessary for the `prependmac` codec, which is why I took some code out of Sphinx to the Mac file so that codecs can use it.

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
t-bast: Ah that's true.
Let me investigate a bit though, because I think that `BadOnion` errors should in fact be completely separated from `FailureMessage` which would fix this problem.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -74,298 +48,323 @@ object Sphinx extends Logging {
...
-      case Success(ParsedPacket(_, nextPacket, sharedSecret)) if nextPacket.isLastPacket => Success(acc :+ sharedSecret)
-      case Success(ParsedPacket(_, nextPacket, sharedSecret)) => extractSharedSecrets(nextPacket.serialize, privateKey, associatedData, acc :+ sharedSecret)
-      case Failure(t) => Failure(t)
+  def peekPayloadLength(payload: ByteVector): Int = {
+    payload.head match {
+      case 0 =>
+        // The 1.0 BOLT spec used 65-bytes frames inside the onion payload.
+        // The first byte of the frame (called `realm`) is set to 0x00, followed by 32 bytes of per-hop data, followed by a 32-bytes mac.
+        65
+      case _ =>
```
sstone: Shouldn't there be a special case for `1` ?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -74,298 +48,323 @@ object Sphinx extends Logging {
...
-      routingInfo2.dropRight(routingInfoFiller.length) ++ routingInfoFiller
+  case class DecryptedPacket(payload: ByteVector, nextPacket: wire.OnionPacket, sharedSecret: ByteVector32) {
+
+    val isLastPacket: Boolean = payload.head match {
+      // In Bolt 1.0 the last hop is signaled via an empty hmac.
+      case 0 => nextPacket.hmac == ByteVector32.Zeroes
+      // In Bolt 1.1 the last hop can also be signaled via a dedicated TLV type with type=0x00.
+      case 0xfd => payload(3) == 0 || nextPacket.hmac == ByteVector32.Zeroes
+      case _ => payload(1) == 0 || nextPacket.hmac == ByteVector32.Zeroes
     }
```
sstone: Is it possible to have `nextPacket.hmac == ByteVector32.Zeroes` and not be the last packet ?
Is it an error is `nextPacket.hmac == ByteVector32.Zeroes` but `payload(1) != 0` ?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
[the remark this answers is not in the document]
t-bast: Actually that's already done inside `wrap`

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -0,0 +1,56 @@
...
+import fr.acinq.eclair.crypto.Sphinx
+import scodec.Codec
+import scodec.bits.ByteVector
+import scodec.codecs._
+
+/**
+  * Created by t-bast on 05/07/2019.
+  */
+
+case class OnionPacket(version: Int,
```
sstone: To be consistent with how we map BOLT names to our own classes, `OnionPacket` should be `OnionRoutingPacket`, which may also help reduce the ambiguity described above between wire and Sphinx OnionPacket definitions

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
t-bast: Why? 1 is an invalid value, but not at this layer, it will fail when trying to parse to a TLV stream.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
t-bast: There is one chance in `2^256` that `ByteVector32.Zeroes` ends up being the correct mac value for a non-last packet, but that's already the case with the current scheme :)
Apart from that it shouldn't change what we have now, we just added another termination condition. It's unclear yet how it will be used, but we wanted to have that functionality ready to reclaim some space inside the onion (instead of wasting 32 bytes).

eclair-core/src/test/scala/fr/acinq/eclair/crypto/SphinxSpec.scala
```diff
@@ -71,108 +71,277 @@ class SphinxSpec extends FunSuite {
...
+  /*
+  filler = 0xb77d99c935d3f32469844f7e09340a91ded147557bdd0456c369f7e449587c0f5666faab58040146db49024db88553729bce12b860391c29c1779f022ae48a9cb314ca35d73fc91addc92632bcf7ba6fd9f38e6fd30fabcedbd5407b6648073c38331ee7ab0332f41f550c180e1601f8c25809ed75b3a1e78635a2ef1b828e92c9658e76e49f995d72cf9781eec0c838901d0bdde3ac21c13b4979ac9e738a1c4d0b9741d58e777ad1aed01263ad1390d36a18a6b92f4f799dcf75edbb43b7515e8d72cb4f827a9af0e7b9338d07b1a24e0305b5535f5b851b1144bad6238b9d9482b5ba6413f1aafac3cdde5067966ed8b78f7c1c5f916a05f874d5f17a2b7d0ae75d66a5f1bb6ff932570dc5a0cf3ce04eb5d26bc55c2057af1f8326e20a7d6f0ae644f09d00fac80de60f20aceee85be41a074d3e1dda017db79d0070b99f54736396f206ee3777abd4c00a4bb95c871750409261e3b01e59a3793a9c20159aae4988c68397a1443be6370fd9614e46108291e615691729faea58537209fa668a172d066d0efff9bc77c2bd34bd77870ad79effd80140990e36731a0b72092f8d5bc8cd346762e93b2bf203d00264e4bc136fc142de8f7b69154deb05854ea88e2d7506222c95ba1aab065c8a
```
pm47: Shouldn't that be removed?

eclair-core/src/test/scala/fr/acinq/eclair/crypto/SphinxSpec.scala
```diff
@@ -71,108 +71,277 @@ class SphinxSpec extends FunSuite {
...
-    val Success(Sphinx.ParsedPacket(payload2, nextPacket2, sharedSecret2)) = Sphinx.parsePacket(privKeys(2), associatedData, nextPacket1.serialize)
-    val Success(Sphinx.ParsedPacket(payload3, nextPacket3, sharedSecret3)) = Sphinx.parsePacket(privKeys(3), associatedData, nextPacket2.serialize)
-    val Success(Sphinx.ParsedPacket(payload4, nextPacket4, sharedSecret4)) = Sphinx.parsePacket(privKeys(4), associatedData, nextPacket3.serialize)
-    assert(Seq(payload0, payload1, payload2, payload3, payload4) == payloads)
+  test("peek at per-hop payload length") {
+    val testCases = Map(
+      34 -> hex"01",
+      41 -> hex"08",
+      65 -> hex"00",
+      285 -> hex"fc",
```
pm47: I can't get the reference results for the last 3 cases, can you explain how you compute the length? E.g. for `fx` I find a length of `303`.

eclair-core/src/test/scala/fr/acinq/eclair/crypto/SphinxSpec.scala
t-bast: It's true that it's a bit redundant to have it both in the comment and in the test itself.
I'll remove it.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Mac.scala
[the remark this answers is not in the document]
pm47: I'm nitpicking, but to be precise the interface should probably be called `HMac32` and the implementation `HMacSha256` ([source](https://en.wikipedia.org/wiki/HMAC)). I know, that's not particularly elegant.

eclair-core/src/test/scala/fr/acinq/eclair/crypto/SphinxSpec.scala
t-bast: `fc` will be `32 (mac length) + 252 (fc decimal value) + 1 (length of the `fc` byte) = 285`
Can you detail how you got to `303`?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Mac.scala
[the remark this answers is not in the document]
t-bast: On the contrary, this interface works for any MAC algorithm that produces a 32-bytes tag, not only for hmac. It loses generality if we call it `HMac32`.
It's true though that `Hmac256` could explicitly mention that we're using SHA (because that's what the rfc does), but it's an internal implementation detail for the users of the class so a smaller name is better isn't it?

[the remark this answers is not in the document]
pm47: You are right!

eclair-core/src/test/scala/fr/acinq/eclair/crypto/SphinxSpec.scala
pm47: Ahem, I was doing 0xfc + 0x32 + 0x01...

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
```diff
@@ -93,4 +97,26 @@ object FailureMessageCodecs {
...
+    * +----------------+----------------------------------+-----------------+----------------------+-----+
+    * with failure message length + pad length = 256
+    */
+  def failureOnionCodec(mac: Mac32): Codec[FailureMessage] = CommonCodecs.prependmac(
+    paddedFixedSizeBytesDependent(
+      260,
+      "failureMessage" | variableSizeBytes(uint16, FailureMessageCodecs.failureMessageCodec),
+      nBits => {
+        val nBytes = (nBits / 8).toInt
+        variableSizeBytes(uint16, bytes(nBytes - 2)).unit(ByteVector.empty)
```
pm47: Your implementation works, but I don't know how. You have defined the *zero* as an empty byte vector, but it should be 0x00?

```suggestion
        variableSizeBytes(uint16, bytes(nBytes - 2)).unit(ByteVector.fromByte(0))
```

pm47: And also it should be:

```suggestion
        variableSizeBytes(uint16, bytes(nBytes - 2).unit(ByteVector.empty))
```

Right?

eclair-core/src/test/scala/fr/acinq/eclair/wire/FailureMessageCodecsSpec.scala
```diff
@@ -42,30 +44,84 @@ class FailureMessageCodecsSpec extends FunSuite {
...
+      hex"ba6e122b2941619e2106e8437bf525356ffc8439ac3b2245f68546e298a08cc6 000a 400f 000000000000002a 00f6 000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
+    ) zip Seq(
```
pm47: This construct with `zip` is a bit error prone, why not define a `Map` instead?

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
t-bast: You're right, it worked the same with an empty ByteVector simply because it treats `0x00` as the default value if not specified, but it's better to be explicit.
And it's better to put the unit inside the variableSizeBytes as well, I'll update that.

pm47: How about: 😎 

```scala
def failureOnionCodec(mac: Mac32): Codec[FailureMessage] = CommonCodecs.prependmac(
    paddedFixedSizeBytesDependent(
      260,
      "failureMessage" | variableSizeBytes(uint16, FailureMessageCodecs.failureMessageCodec),
      nBits => variableSizeBytes(uint16, ignore(nBits - 2 * 8)) // two bytes are used to encode the length of the padding
    ).as[FailureMessage], mac)
```

pm47: Nitpick:

```scala
def failureOnionCodec(mac: Mac32): Codec[FailureMessage] = CommonCodecs.prependmac(
    paddedFixedSizeBytesDependent(
      260,
      "failureMessage" | variableSizeBytes(uint16, FailureMessageCodecs.failureMessageCodec),
      nBits => "padding" | variableSizeBytes(uint16, ignore(nBits - 2 * 8)) // two bytes are used to encode the length of the padding
    ).as[FailureMessage], mac)
```

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -74,298 +48,319 @@ object Sphinx extends Logging {
...
+      * @param packet         packet received by this node.
+      * @return a DecryptedPacket(payload, packet, shared secret) object where:
+      *         - payload is the per-hop payload for this node.
+      *         - packet is the next packet, to be forwarded using the info that is given in the payload.
+      *         - shared secret is the secret we share with the node that sent the packet. We need it to propagate
+      *         failure messages upstream.
+      *         or a BadOnion error containing the hash of the invalid onion.
+      */
+    def peel(privateKey: PrivateKey, associatedData: ByteVector, packet: wire.OnionRoutingPacket): Either[wire.BadOnion, DecryptedPacket] = packet.version match {
+      case 0 => Try(PublicKey(packet.publicKey, checkValid = true)) match {
```
pm47: The field `publicKey` isn't typed in `OnionRoutingPacket` because we want to handle the "invalid pubkey" case at this layer.

But for optimization reasons (it is pretty expensive to fully validate a public key), our codec currently doesn't do any validation: we just read 33 bytes and ensure that the key begins with `02` or `03`.

How about we properly type the field to `PublicKey` in `OnionRoutingPacket`, and we do a proper validation here (that is, we verify that the point is indeed on the curve here).

The tradeoff here would be that un unparseable key would probably not even start with a `02` or `03`, and the failure would happen in the codec. Okay, not sure about this.

@sstone WDYT?

t-bast: We definitely need to verify here that the point is on the curve (which is done with the `checkValid = true` flag. So if we do validation here, it feels a bit weird to do partial validation in the codec...
Especially since that means that we need to make the codec more complex to correctly return the right `BadOnion` error in case the public key is invalid (not sure it's easy to do with the current decoding flow).

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -0,0 +1,65 @@
...
+                              payload: ByteVector,
+                              hmac: ByteVector32)
+
+case class PerHopPayload(shortChannelId: ShortChannelId,
+                         amtToForward: Long,
+                         outgoingCltvValue: Long)
+
+object OnionCodecs {
+
+  def onionPacketCodec(payloadLength: Int): Codec[OnionRoutingPacket] = (
```
pm47: For consistency:

```suggestion
  def onionRoutingPacketCodec(payloadLength: Int): Codec[OnionRoutingPacket] = (
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
```diff
@@ -215,34 +213,34 @@ object Relayer {
...
-              Success(RelayPayload(add, perHopPayload, nextPacket))
-            case Attempt.Failure(cause) =>
-              Failure(new RuntimeException(cause.messageWithContext))
-          }
-      }
+  def decryptPacket(add: UpdateAddHtlc, privateKey: PrivateKey): Either[BadOnion, NextPayload] =
+    Sphinx.PaymentPacket.peel(privateKey, add.paymentHash, add.onionRoutingPacket) match {
+      case Right(p@Sphinx.DecryptedPacket(payload, nextPacket, _)) =>
+        OnionCodecs.perHopPayloadCodec.decode(payload.bits) match {
+          case Attempt.Successful(DecodeResult(perHopPayload, _)) if p.isLastPacket =>
```
pm47: Should we check that `remainder` is empty?

t-bast: I don't think it's this layer's responsibility.
For 65-bytes frames, the sphinx layer ensures that what we receive in the `Relayer` has the proper format so we will always have an empty remainder.
For variable-length frames, the sphinx layer also guarantees that we'll receive exactly the bytes of the tlv stream, so it should always be empty too.
Do you want to add a `require` to catch bugs (since this is a case that should never happen if the other layers behave properly)?

t-bast: I think it's always re-assuring to add such a `require` check to mention that the other layers should make sure we end up with an empty remainder here.

pm47: A `require` here would result in an infinite loop. Doing this:

```scala
 OnionCodecs.perHopPayloadCodec.decode(payload.bits) match {
          case Attempt.Successful(DecodeResult(perHopPayload, remainder)) if p.isLastPacket && remainder.isEmpty =>
            Right(FinalPayload(add, perHopPayload))
          case Attempt.Successful(DecodeResult(perHopPayload, remainder)) if remainder.isEmpty =>
            Right(RelayPayload(add, perHopPayload, nextPacket))
          case _ =>
            // Onion is correctly encrypted but the content of the per-hop payload couldn't be parsed.
            Left(InvalidOnionPayload(Sphinx.PaymentPacket.hash(add.onionRoutingPacket)))
        }
```

would correctly return an `InvalidOnionPayload` but it doesn't look very good.

Alternatively:
```scala
 OnionCodecs.perHopPayloadCodec.decode(payload.bits) match {
          case Attempt.Successful(DecodeResult(_, remainder)) if remainder.nonEmpty =>
            Left(InvalidOnionPayload(Sphinx.PaymentPacket.hash(add.onionRoutingPacket)))
          case Attempt.Successful(DecodeResult(perHopPayload, _)) if p.isLastPacket =>
            Right(FinalPayload(add, perHopPayload))
          case Attempt.Successful(DecodeResult(perHopPayload, _)) =>
            Right(RelayPayload(add, perHopPayload, nextPacket))
          case Attempt.Failure(_) =>
            // Onion is correctly encrypted but the content of the per-hop payload couldn't be parsed.
            Left(InvalidOnionPayload(Sphinx.PaymentPacket.hash(add.onionRoutingPacket)))
        }
```
doesn't look much better.

t-bast: What about this?

```scala
OnionCodecs.perHopPayloadCodec.decode(payload.bits) match {
          case Attempt.Successful(DecodeResult(perHopPayload, remainder)) =>
            require(remainder.isEmpty, "there shouldn't be any remaining bytes after decoding sphinx payload")
            if (p.isLastPacket)
              Right(FinalPayload(add, perHopPayload))
            else
              Right(RelayPayload(add, perHopPayload, nextPacket))
          case Attempt.Failure(_) =>
            // Onion is correctly encrypted but the content of the per-hop payload couldn't be parsed.
            Left(InvalidOnionPayload(Sphinx.PaymentPacket.hash(add.onionRoutingPacket)))
        }
```

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
sstone: It bothers me a bit too to not use types here but it makes parsing and error handling easy to read and compare to the BOLTs so I can live with it. 
Also our process is already split in 2 and we do not check the htlc's onion packet until we've received a revocation so using a stricter type should probably be part of a larger refactoring where the htlc is fully parsed when we receive it

t-bast:
Christian updated the spec PR to address the latest comments (with the updated feature bit) and Laolu got the LN implementation working.
I'm going to freeze this PR because it's already pretty big, and I'm going to add support for the onion tlvs in another PR (with a cross-implementation test in Christian's repo).

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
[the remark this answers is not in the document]
pm47: Discussed with @sstone, in this particular case it is easier to not type the field.

PR #981, opened 2019-05-06
Wireshark dissector support

t-bast:
The wireshark lightning-dissector used a hack to work with eclair: it read our log file in debug mode and parsed the encryptions keys via a very brittle regex. Unfortunately our log format had changed, thus breaking the regex.

The recommended way to integrate with the dissector is to write keys to a log file whenever key rotation happens (see https://github.com/nayutaco/lightning-dissector/blob/master/CONTRIBUTING.md).

This PR adds a new logger dedicated to that task, disabled by default. I also made the necessary changes to the `lightning-dissector` lua plugin and will send a PR upstream once we agree on the eclair changes.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
@@ -96,6 +96,9 @@ object ChaCha20 {
   */
 object ChaCha20Poly1305 extends Logging {
 
+  // This logger is used to dump encryption keys to enable traffic analysis by the lightning-dissector.
+  // See https://github.com/nayutaco/lightning-dissector for more details.
+  // It is disabled by default (in the logback.xml configuration file).
   val keyLogger = Logger("keylog")
 
   /**
--- a/eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
@@ -19,6 +19,7 @@ package fr.acinq.eclair.crypto
 import java.nio.ByteOrder
 
 import fr.acinq.bitcoin.{ByteVector32, Protocol}
+import grizzled.slf4j.Logger
 import grizzled.slf4j.Logging
 import org.spongycastle.crypto.engines.ChaCha7539Engine
 import org.spongycastle.crypto.params.{KeyParameter, ParametersWithIV}
@@ -31,8 +32,8 @@ import scodec.bits.ByteVector
 object Poly1305 {
   /**
     *
-    * @param key  input key
-    * @param data input data
+    * @param key   input key
+    * @param datas input data
     * @return a 16 byte authentication tag
     */
   def mac(key: ByteVector, datas: ByteVector*): ByteVector = {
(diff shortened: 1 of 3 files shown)


t-bast:
Here is what it looks like when used in wireshark:

wireshark-lightning-dissector

This is very helpful, there's a lot of information easily accessible for troubleshooting / understanding the protocol.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
```diff
@@ -94,6 +96,8 @@ object ChaCha20 {
   */
 object ChaCha20Poly1305 extends Logging {
 
+  val keyLogger = Logger("keylog")
```
araspitzu: How about adding a small comment here? In some time, it will be hard to remember why we've put this here.

t-bast: Sure, good idea.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
```diff
@@ -106,7 +113,12 @@ object ChaCha20Poly1305 extends Logging {
     val polykey = ChaCha20.encrypt(ByteVector32.Zeroes, key, nonce)
     val ciphertext = ChaCha20.encrypt(plaintext, key, nonce, 1)
     val tag = Poly1305.mac(polykey, aad, pad16(aad), ciphertext, pad16(ciphertext), Protocol.writeUInt64(aad.length, ByteOrder.LITTLE_ENDIAN), Protocol.writeUInt64(ciphertext.length, ByteOrder.LITTLE_ENDIAN))
+
     logger.debug(s"encrypt($key, $nonce, $aad, $plaintext) = ($ciphertext, $tag)")
+    if (nonce === ByteVector.fill(ChaCha20.NonceLength)(0.byteValue)) {
```
sstone: I'd rather compare nonce to a `Zeroes` val, with a comment that explains that nonce is incremental and initialised with 0.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
```diff
@@ -124,7 +136,12 @@ object ChaCha20Poly1305 extends Logging {
     val tag = Poly1305.mac(polykey, aad, pad16(aad), ciphertext, pad16(ciphertext), Protocol.writeUInt64(aad.length, ByteOrder.LITTLE_ENDIAN), Protocol.writeUInt64(ciphertext.length, ByteOrder.LITTLE_ENDIAN))
     require(tag == mac, "invalid mac")
     val plaintext = ChaCha20.decrypt(ciphertext, key, nonce, 1)
+
     logger.debug(s"decrypt($key, $nonce, $aad, $ciphertext, $mac) = $plaintext")
+    if (nonce === ByteVector.fill(ChaCha20.NonceLength)(0.byteValue)) {
```
sstone: Same as above

eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
t-bast: Sounds good, done.

PR #982, opened 2019-05-06
Add bot support for code coverage (codecov)

araspitzu:
This PR adds support for the scala test coverage tool ([scoverage](http://scoverage.org/)), it will help providing info on how much the coverage changes for each pull request, also a badge indicating the current coverage is added to the readme.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ scala:
  - 2.11.12
 env:
  - export LD_LIBRARY_PATH=/usr/local/lib
- global:
-    - CODECOV_TOKEN="eadc3ab4-14c9-441a-841b-9a5ae511059f"
 before_install:
  - wget http://apache.crihan.fr/dist/maven/maven-3/3.6.0/binaries/apache-maven-3.6.0-bin.zip
  - unzip -qq apache-maven-3.6.0-bin.zip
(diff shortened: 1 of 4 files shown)


t-bast:
LGTM

One thing that I usually do with codecov is provide a `.codecov.yml` configuration file (at the root of the repo). Codecov's default configuration expects unreasonable code coverage percentage which makes the status check fail: this doesn't prevent you from merging, but when looking at the commit history it will look like the commit build failed (you'll see a big red failure cross). To prevent that I usually set reasonable expectations in that custom configuration.

But that's definitely something that can be added afterwards, once we get tired of seeing failed status checks :)

araspitzu:
Currently scoverage is configured to pass the build with any coverage value (travis run `mvn scoverage:report`), so it will just let you know if it's decreasing but it shouldn't fail the build.

PS. example report [here](https://codecov.io/gh/acinq/eclair/branch/scoverage-build-plugin)

.travis.yml
```diff
@@ -13,7 +13,7 @@ before_install:
  - export M2_HOME=$PWD/apache-maven-3.6.0
  - export PATH=$M2_HOME/bin:$PATH
 script:
- - mvn install
+ - mvn scoverage:report
```
pm47: Does this do everything that `mvn install` does? (compile+run tests)

pm47: It looks like it does (https://scoverage.github.io/scoverage-maven-plugin/1.4.0-M5/plugin-info.html).

PR #988, opened 2019-05-09
Accept `commit_sig` without changes

pm47:
LND sometimes sends a new signature without any changes, which is a
(harmless) spec violation.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -395,8 +395,9 @@ object Commitments {
     // we will reply to this sig with our old revocation hash preimage (at index) and our next revocation hash (at index + 1)
     // and will increment our index
 
-    if (!remoteHasChanges(commitments))
-      throw CannotSignWithoutChanges(commitments.channelId)
+    // LND sometimes sends a new signature without any changes, which is a (harmless) spec violation
+    //if (!remoteHasChanges(commitments))
```
araspitzu: How about adding a log here? Whether it's harmless or not it's still a spec violation and it's worth knowing when it happens.

PR #990, opened 2019-05-09
Set tcp client timeout to 20s

pm47:
So that it fails before the ask/api time out.


araspitzu:
Looks good, although a user can still specify an http timeout that is smaller than the tcp timeout, we now prevent this from default.

PR #991, opened 2019-05-09
Restrict scope of the code coverage report

araspitzu:
Ignore subprojects *eclair-node* and *eclair-node-gui* when generating a coverage report (codecov).

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,3 +5,7 @@ coverage:
 
 comment:
   layout: "diff, files"
+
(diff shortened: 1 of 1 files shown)


araspitzu:
> I'm not sure that's what you want.
> Per codecov's documentation, `folder/*` will not match recursively in the folder.
> I think you want `eclair-node/**/*` and `eclair-gui/**/*` don't you?

Yes that was the idea.

PR #995, opened 2019-05-10
Recovery tool

araspitzu:
Adds a channel recovery functionality to eclair. The recovery is only capable of recovering our main output of a channel if the following conditions are met:

- The user has the seed 
- The user has the channel backup, it's *static* string and can be obtained calling the `/backup` API
- The user has information about the channel funding transaction, such as channelShortId or fundingTxId
- The remote node is still online
- The remote node supports `option_data_loss_protect`

Assuming the above conditions are met then the recovery tool is able to trigger the publishing of the latest commitment transaction owned by the remote node, at which point eclair will spend its main output back to the wallet. It is possible to trigger the recovery procedure both from an API call `/recovery` and from a console based interactive procedure where the user is asked for the necessary information. The recovery has been tested against the following implementations: eclair (latest master), c-lightning (commit fefe7dfb) and lnd v0.6.0.


pm47:
I only took a cursory look but I would have done this in a separate module and I don't think this should have an impact on existing code (e.g. `case class Connect(uri: NodeURI, withCommitments: Set[HasCommitments] = Set.empty)`).

araspitzu:
> I only took a cursory look but I would have done this in a separate module and I don't think this should have an impact on existing code (e.g. `case class Connect(uri: NodeURI, withCommitments: Set[HasCommitments] = Set.empty)`).

I agree we shouldn't change existing code, Peer.Connect was just a low hanging fruit to trigger the peer level connection + INPUT_RECONNECTED.

araspitzu:
Will open a new shiny PR

PR #1004, opened 2019-05-14
Increase Service and EclairImpl test coverage

araspitzu:
Improves the API test to match the ~80% test coverage that we have on eclair, it also add the test-scoped dependency of [mockito](https://github.com/mockito/mockito-scala) to simplify mocking services in the tests.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -38,8 +38,11 @@ import fr.acinq.eclair.wire.{ChannelAnnouncement, ChannelUpdate, NodeAddress, No
 import TimestampQueryFilters._
 
 case class GetInfoResponse(nodeId: PublicKey, alias: String, chainHash: ByteVector32, blockHeight: Int, publicAddresses: Seq[NodeAddress])
+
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -182,43 +182,46 @@ class EclairImpl(appKit: Kit) extends Eclair {
...
 
-    appKit.nodeParams.db.payments.listPendingPaymentRequests(from, to)
+    appKit.nodeParams.db.payments.listPendingPaymentRequests(filter.from, filter.to)
   }
 
   override def getInvoice(paymentHash: ByteVector32)(implicit timeout: Timeout): Future[Option[PaymentRequest]] = Future {
     appKit.nodeParams.db.payments.getPaymentRequest(paymentHash)
   }
 
+  def getDefaultTimestampFilters(from_opt: Option[Long], to_opt: Option[Long]) = {
```
pm47: Maybe make that a method of `TimestampQueryFilter` companion object?

araspitzu: In 409a7fa I moved the function to the companion object and grouped together the case classes.

PR #1008, opened 2019-05-17
Use 64B representation instead of DER for sigs

pm47:
It is more compact, and as an added bonus it frees us from the
completely unrelated Bitcoin-specific `0x01` trailing sig hash.

Note that we already used the 64B representation for storage everywhere,
except in `ChannelCodecs.htlcTxAndSigsCodec`, which required a backward
compatibility codec. Added a nonreg test for this.

@sstone it seems like `Transactions.sign` is misplaced and belongs to
`LocalKeyManager`, but I'm not completely sure because it feels nice to
have all the signature-related stuff at the same place in
`Transactions`.


pm47:
Superseded by #1014

PR #1009, opened 2019-05-20
Peer reconnection address from node announcements

### LINKED ISSUE #1003: Due to failed reconnection, paying an invoice incurs unnecessary fees.
I have two nodes. I created an invoice on one and paid it from the other. I paid it from Eclair Desktop. There is a channel (575638x611x1) from my Eclair node to my other node (MemeracingLN). Eclair says MemeracingLN is offline. However, MemeracingLN was the target of the payment and the invoice has been settled on the MemeracingLN node.

Eclair (Desktop, v 0.3) paid an invoice to a node without using its direct channel to that node because the node is recorded in EClair as _offline_.

If we are paying a node to which we have a channel then use the channel, and if the channel is offline, try a simple connection to it, and if that doesn't work, ask the user.

I have not attempted to reproduce.  I would expect the code is written to have the behavior I got, simply because the reliability of the OFFLINE indication was assumed to be high.  It isn't, so I propose the following solution:

Before finding a route to the receiving node, check if we have a channel to it and if so, **do not route**.  Instead:
1. Is the channel online?
   1. Yes - negotiate the payment with the channel.
   1. No - Attempt a "simple connection" to the receiving node.  Did it succeed?
      1. Yes - Go back to step 1.
      1. No - Return an error: "Receiving node is OFFLINE." - Optionally, allow the user to have Eclair route the payment anyway.

I am on Windows 10 using JDK 11.

araspitzu:
Currently the node address information in `node_announcement` messages is not used when establishing a connection. Peer's addresses are known only when the user initiated the connection providing a destination address, in case of incoming connection the peer's address is blanked out and we can't reconnect. According to the [spec](https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md#the-node_announcement-message) the `addresses` field of `node_announcement` should contain the network addresses where a peer expects an incoming connection, this PR leverages this information to provide the following enhancements:

- Connect-to a pubkey (no address specified, will use `node_announcements`)
- Reconnect to peers using their `node_announcements` address as fallback in case we don't know the address

Fixes #1003

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -23,6 +23,7 @@ import java.nio.ByteOrder
 import akka.actor.{ActorRef, FSM, OneForOneStrategy, PoisonPill, Props, Status, SupervisorStrategy, Terminated}
 import akka.event.Logging.MDC
 import akka.util.Timeout
+import com.google.common.net.HostAndPort
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{ByteVector32, DeterministicWallet, MilliSatoshi, Protocol, Satoshi}
 import fr.acinq.eclair.blockchain.EclairWallet
@@ -59,8 +60,17 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
   }
 
   when(DISCONNECTED) {
-    case Event(Peer.Connect(NodeURI(_, hostAndPort)), d: DisconnectedData) =>
-      val address = new InetSocketAddress(hostAndPort.getHost, hostAndPort.getPort)
+    case Event(Peer.Connect(_, address_opt), d: DisconnectedData) =>
+      val address = address_opt match {                                                                        // The connect command has an optional address that is specified by the user,
+        case Some(hostAndPort) => new InetSocketAddress(hostAndPort.getHost, hostAndPort.getPort)              // if present we use it otherwise we use the address from the node_announcement
+        case None => nodeParams.db.network.getNode(remoteNodeId).flatMap(_.addresses.headOption) match {       //
+          case Some(announcementAddress) => announcementAddress.socketAddress                                  //
+          case None =>
+            log.warning(s"Unable to connect to $remoteNodeId no address found")
+            stopPeer()                                                                                         // if there is no address from the node_announcement we stop
+            throw new IllegalArgumentException(s"Unable to connect to $remoteNodeId no address found")
+        }
+      }
       if (d.address_opt.contains(address)) {
         // we already know this address, we'll reconnect automatically
         sender ! "reconnection in progress"
@@ -73,8 +83,8 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
 
     case Event(Reconnect, d: DisconnectedData) =>
       d.address_opt match {
-        case None => stay // no-op (this peer didn't initiate the connection and doesn't have the ip of the counterparty)
         case _ if d.channels.isEmpty => stay // no-op (no more channels with this peer)
+        case None => stay // no-op (this peer didn't initiate the connection and doesn't have the ip of the counterparty)
         case Some(address) =>
           context.actorOf(Client.props(nodeParams, authenticator, address, remoteNodeId, origin_opt = None))
           // exponential backoff retry with a finite max
@@ -549,7 +559,14 @@ object Peer {
   case object CONNECTED extends State
 
   case class Init(previousKnownAddress: Option[InetSocketAddress], storedChannels: Set[HasCommitments])
-  case class Connect(uri: NodeURI)
+  case class Connect(nodeId: PublicKey, address_opt: Option[HostAndPort]) {
+    def uri: Option[NodeURI] = address_opt.map(NodeURI(nodeId, _))
+  }
+
+  object Connect {
+    def apply(uri: NodeURI): Connect = new Connect(uri.nodeId, Some(uri.address))
+  }
+
   case object Reconnect
   case object Disconnect
   case object ResumeAnnouncements
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -81,7 +81,7 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
       }
 
     case Event(Reconnect, d: DisconnectedData) =>
-      if(d.channels.isEmpty) stay
+      if (d.channels.isEmpty) stay
 
       d.address_opt.orElse(getPeerAddressFromNodeAnnouncement()) match {
         case None =>
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -85,7 +85,6 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
           log.warning(s"Unable to reconnect to peer, no address known")
           stay
         case Some(address) =>
-          // InetAddress.getHostAddress returns the IP address as string
           context.actorOf(Client.props(nodeParams, authenticator, address, remoteNodeId, origin_opt = None))
           log.info(s"reconnecting to $address")
           // exponential backoff retry with a finite max
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -78,12 +78,9 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
       }
 
     case Event(Reconnect, d: DisconnectedData) =>
-      if (d.channels.isEmpty) stay
-
       d.address_opt.orElse(getPeerAddressFromNodeAnnouncement()) match {
-        case None =>
-          log.warning(s"Unable to reconnect to peer, no address known")
-          stay
+        case _ if d.channels.isEmpty => stay // no-op, no more channels with this peer
+        case None                    => stay // no-op, we don't know any address to this peer and we won't try reconnecting again
         case Some(address) =>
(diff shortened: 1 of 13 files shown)


pm47:
What would be nice is also implement a `disconnect` API call.

pm47:
We should also probably make the exponential backoff delay go all the way to 1 hour every reconnect attempt

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -549,7 +567,14 @@ object Peer {
   case object CONNECTED extends State
 
   case class Init(previousKnownAddress: Option[InetSocketAddress], storedChannels: Set[HasCommitments])
-  case class Connect(uri: NodeURI)
+  case class Connect(nodeId: PublicKey, address_opt: Option[HostAndPort]) {
```
pm47: Why didn't you use the `Reconnect` object here? Instead of a `Connect` without address?

araspitzu: Because logically a connection to a peer does not necessarily require knowledge of his address, eclair can support connect-to-pubkey (with an optional address).

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -59,26 +58,39 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
...
-      } else {
-        // we immediately process explicit connection requests to new addresses
-        context.actorOf(Client.props(nodeParams, authenticator, address, remoteNodeId, origin_opt = Some(sender())))
-        stay
+    case Event(Peer.Connect(_, address_opt), d: DisconnectedData) =>
+      address_opt.map(hAndp => new InetSocketAddress(hAndp.getHost, hAndp.getPort)).orElse {
+        getPeerAddressFromNodeAnnouncement()
+      } match {
+        case None =>
+          log.warning(s"Unable to connect to $remoteNodeId no address found")
```
pm47: This log line is not needed (also, the node id is in the MDC)

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -59,26 +58,39 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
...
+          sender() ! s"Unable to connect to $remoteNodeId no address found"
```
```suggestion
          sender() ! "no address found"
```

Consistency with existing messages

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -59,26 +58,39 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
...
+          stopPeer()
```
pm47: ⚠️ bug: we may very well have channels here!

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -59,26 +58,39 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
...
+            context.actorOf(Client.props(nodeParams, authenticator, address, remoteNodeId, origin_opt = Some(sender())))
+            stay using d.copy(address_opt = Some(address))
+          }
       }
 
     case Event(Reconnect, d: DisconnectedData) =>
-      d.address_opt match {
-        case None => stay // no-op (this peer didn't initiate the connection and doesn't have the ip of the counterparty)
-        case _ if d.channels.isEmpty => stay // no-op (no more channels with this peer)
+      if (d.channels.isEmpty) stay
```
pm47: It is not obvious at all that the `stay` statement breaks the execution, which is what you seem to be assuming. I find this construction much more error prone than the previous one

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -59,26 +58,39 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
...
+
+      d.address_opt.orElse(getPeerAddressFromNodeAnnouncement()) match {
+        case None =>
+          log.warning(s"Unable to reconnect to peer, no address known")
```
pm47: I don't think that deserves a log, or a most a debug (consistency: log lines don't start uppercase)

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -59,26 +58,39 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
...
+          stay
         case Some(address) =>
+          // InetAddress.getHostAddress returns the IP address as string
```
pm47: This comment seems useless

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -91,7 +103,7 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
         case Some((gf, lf)) => wire.Init(globalFeatures = gf, localFeatures = lf)
         case None => wire.Init(globalFeatures = nodeParams.globalFeatures, localFeatures = nodeParams.localFeatures)
       }
-      log.info(s"using globalFeatures=${localInit.globalFeatures.toBin} and localFeatures=${localInit.localFeatures.toBin}")
+      log.debug(s"using globalFeatures={} and localFeatures={}", localInit.globalFeatures.toBin, localInit.localFeatures.toBin)
```
pm47: This log is useful as information because there can be peer specific params. Ok for the formatting changes but then you also need to update line 138. As a rule of thumb I mostly keep the `{}` notation for debug statements on hot paths.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -412,6 +424,7 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
       stay using d.copy(behavior = d.behavior.copy(fundingTxAlreadySpentCount = 0, ignoreNetworkAnnouncement = false))
 
     case Event(Disconnect, d: ConnectedData) =>
+      log.info(s"disconnecting")
```
pm47: This event should also be handled in state `INITIALIZING`

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -478,7 +491,7 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
 
   onTransition {
     case INSTANTIATING -> DISCONNECTED if nodeParams.autoReconnect && nextStateData.address_opt.isDefined => self ! Reconnect // we reconnect right away if we just started the peer
-    case _ -> DISCONNECTED if nodeParams.autoReconnect && nextStateData.address_opt.isDefined => setTimer(RECONNECT_TIMER, Reconnect, 1 second, repeat = false)
+    case _ -> DISCONNECTED if nodeParams.autoReconnect => setTimer(RECONNECT_TIMER, Reconnect, 1 second, repeat = false)
```
pm47: the same change should be applied to the line immediately below.

araspitzu:
At commit 3a944e2  all the comments have been addressed, @pm47

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -177,6 +184,13 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
       } else {
         stay using d.copy(channels = channels1)
       }
+
+    case Event(Disconnect(nodeId), d: ConnectedData) if nodeId == remoteNodeId =>
```
pm47: Wrong data type, should be `InitializingData`. More tests needed?

pm47:
How about this refactoring:

```diff
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
index d7c1f4422..8560a4622 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -59,9 +59,9 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor

   when(DISCONNECTED) {
     case Event(Peer.Connect(_, address_opt), d: DisconnectedData) =>
-      address_opt.map(hAndp => new InetSocketAddress(hAndp.getHost, hAndp.getPort)).orElse {
-        getPeerAddressFromNodeAnnouncement()
-      } match {
+      address_opt
+        .map(hostAndPort2InetSocketAddress)
+        .orElse(getPeerAddressFromNodeAnnouncement) match {
         case None =>
           sender ! "no address found"
           stay
@@ -78,7 +78,7 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
       }

     case Event(Reconnect, d: DisconnectedData) =>
-      d.address_opt.orElse(getPeerAddressFromNodeAnnouncement()) match {
+      d.address_opt.orElse(getPeerAddressFromNodeAnnouncement) match {
         case _ if d.channels.isEmpty => stay // no-op, no more channels with this peer
         case None                    => stay // no-op, we don't know any address to this peer and we won't try reconnecting again
         case Some(address) =>
@@ -518,7 +518,7 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
   }

   // TODO gets the first of the list, improve selection?
-  def getPeerAddressFromNodeAnnouncement(): Option[InetSocketAddress] = {
+  def getPeerAddressFromNodeAnnouncement: Option[InetSocketAddress] = {
     nodeParams.db.network.getNode(remoteNodeId).flatMap(_.addresses.headOption.map(_.socketAddress))
   }

@@ -643,4 +643,6 @@ object Peer {
       case _ => true // if there is a filter and message doesn't have a timestamp (e.g. channel_announcement), then we send it
     }
   }
+
+  def hostAndPort2InetSocketAddress(hostAndPort: HostAndPort): InetSocketAddress = new InetSocketAddress(hostAndPort.getHost, hostAndPort.getPort)
 }
```

NB: for methods without side effect parentheses should be omitted

PR #1012, opened 2019-05-20
Set `handleCommandError` to `WARN`

pm47:
Those are not fatal errors for channels.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1704,7 +1704,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   }
 
   def handleCommandError(cause: Throwable, cmd: Command) = {
-    log.error(s"${cause.getMessage} while processing cmd=${cmd.getClass.getSimpleName} in state=$stateName")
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1704,7 +1704,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   }
 
   def handleCommandError(cause: Throwable, cmd: Command) = {
-    log.error(s"${cause.getMessage} while processing cmd=${cmd.getClass.getSimpleName} in state=$stateName")
+    log.warning(s"${cause.getMessage} while processing cmd=${cmd.getClass.getSimpleName} in state=$stateName")
```
araspitzu: How about passing the error to the logger? It will print the stacktrace in the logs and makes it easier to debug errors.
```suggestion
    log.warning(s"${cause.getMessage} while processing cmd=${cmd.getClass.getSimpleName} in state=$stateName", cause)
```

pm47: We print the stack trace for unknown errors (two lines below). For known errors printing the stack trace doesn't give information and will only pollute the logs.

PR #1013, opened 2019-05-21
Make `htlc-reaper` a top-level actor

pm47:
The way we handle requests to `Switchboard` assume that all children are `Peer`s. This led the `peer` API request to timeout.


araspitzu:
LGTM

PR #1014, opened 2019-05-23
Use compact encoding for signatures (64 bytes) instead of DER

sstone:
This is based on #1008 with modification to use our latest secp256k1 JNI bindings

We now always use public keys in compressed format (33 bytes, DER encoded). All operations on public keys (LN or bip32) are done using this format so we never need to convert keys.

Compared to our current master, which used bitcoin-lib 0.11:

- signatures are created and verified in compact format (unless secp256k1 bindings are not available, which is not the case in our production targets including Android). Note: we still need to convert signature to DER format when we add them to transaction's witness scripts, but it's very fast.
- pubkey addition also uses secp256k1 and will be much faster (it is used to compute bip32 keys, and also to compute commitment keys and points)
- memory usage will be reduced because compact keys are smaller than DER keys and no conversion means less object creation.

The only place where we need to use "compressed" private keys (i,.e with a `0x01` suffix) is in our key manager, when we generate channel secrets, so as to be compatible with previous versions of eclair (i.e. users can upgrade safely).

We still use Bouncy Castle for pubkey recovery, which is used when parsing payment requests. It could be faster but the current version takes about 0.1s on Android and has no significant impact on UX.


eclair-core/src/main/scala/fr/acinq/eclair/transactions/Scripts.scala
```diff
@@ -18,27 +18,15 @@ package fr.acinq.eclair.transactions
...
-
-  def expiry2cltv(in: Long): Long = ???
-
-  /*in match {
-      case locktime(Blocks(blocks)) => blocks
-      case locktime(Seconds(seconds)) => seconds
-    }*/
+  def der(sig: ByteVector64): ByteVector = Crypto.compact2der(sig) :+ 1
 
   def multiSig2of2(pubkey1: PublicKey, pubkey2: PublicKey): Seq[ScriptElt] = if (LexicographicalOrdering.isLessThan(pubkey1.toBin, pubkey2.toBin))
```
araspitzu: Does it respect the spec ordering, defined as `the numerically lesser of the two DER-encoded`?

sstone: Yes because we use lexicographical ordering for pubkeys (and their encoding does not change here) but not for signatures.

sstone: More precisely: we still use DER encoding for public keys (the 33 bytes compressed representation of a public is DER)

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -435,14 +435,12 @@ object PaymentRequest {
     val data = string2Bits(lowercaseInput.slice(separatorIndex + 1, lowercaseInput.size - 6)) // 6 == checksum size
     val bolt11Data = Codecs.bolt11DataCodec.decode(data).require.value
     val signature = bolt11Data.signature
-    val r = new BigInteger(1, signature.take(32).toArray)
-    val s = new BigInteger(1, signature.drop(32).take(32).toArray)
     val message: ByteVector = ByteVector.view(hrp.getBytes) ++ data.dropRight(520).toByteVector // we drop the sig bytes
-    val (pub1, pub2) = Crypto.recoverPublicKey((r, s), Crypto.sha256(message))
+    val (pub1, pub2) = Crypto.recoverPublicKey(ByteVector64(signature.take(64)), Crypto.sha256(message))
```
pm47: Maybe define:
```scala
val sig = ByteVector64(signature.take(64))
```
since it is used twice

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Scripts.scala
```diff
@@ -105,6 +93,9 @@ object Scripts {
     // @formatter:on
   }
 
+  def scriptPubKeyHtlcSend(ourkey: PublicKey, theirkey: PublicKey, abstimeout: Long, reltimeout: Long, rhash: ByteVector32, commit_revoke: ByteVector): Seq[ScriptElt]
```
pm47: This doesn't seem consistent with how sigs are handled. Also, it looks like `OP_PUSHDATA` can take a PublicKey as argument?

PR #1016, opened 2019-05-24
Close channel if funding tx times out

pm47:
Instead of going to state `ERR_FUNDING_TIMEOUT`.

This only happens when we are fundee. We *could* have some funds at
stake if there was a non-zero `push_msat`, but we already allows 5 days
for the funding tx to confirm so the best option is probably to forget
about the channel.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1561,8 +1561,6 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
   when(ERR_INFORMATION_LEAK)(errorStateHandler)
 
-  when(ERR_FUNDING_TIMEOUT)(errorStateHandler)
-
   when(ERR_FUNDING_LOST)(errorStateHandler)
 
   whenUnhandled {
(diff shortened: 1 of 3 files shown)


btcontract:
I'm afraid 5 days is not enough, there were cases of txs taking weeks to get confirmed eventually.

pm47:
> there were cases of txs taking weeks to get confirmed eventually

Sure but then the funder will do a local close and get back his funds. It cost resources to keep waiting for channels and that could be an easy attack vector.

btcontract:
> that could be an easy attack vector

were there incidents already? This attack vector seems unlikely to me because fundee has to commit sizable amount of funds + miner fees per each pending channel. OTOH cases like https://github.com/lightningnetwork/lnd/issues/2897 do happen already. Can timeout at least be configurable?

pm47:
> were there incidents already?

not that I know of

> fundee has to commit sizable amount of funds

You mean funder? That's the thing, it doesn't have to commit to anything, only pretend to publish a funding tx.

Yes we could make the timeout configurable. And we could keep waiting if we see that a tx has been published but not yet confirmed.

btcontract:
Yes, I specifically mean a case where remote funding is seen in a mempool, otherwise 5 days is too long even. And sadly there are complications with this logic when neutrino is used: https://github.com/lightningnetwork/lnd/issues/2897#issuecomment-480794094

PR #1018, opened 2019-05-28
Removed `PaymentRequest.MAX_AMOUNT`

pm47:
It was obsole since
https://github.com/lightningnetwork/lightning-rfc/commit/068b0bccf94e8cdaf5f298dade0fcc8cc8421ef6.

Note that we use a signed long, but it doesn't matter since 2^64
milli-satoshis is greater than the total number of bitcoins.


pm47:
Thanks @juscamarena for reporting this.

PR #1023, opened 2019-06-03
Add feeThresholdSat and maxFeePct parameters to payments API

### LINKED ISSUE #1021: API Payments Feature Request: Expose Custom Parameter for fees
sendtonode payinvoice sendtoroute do not have fee parameters which is essential as every transaction may not have the same fee policy preference. 

Having both options fee-threshold-sat and max-fee-pct would be nice ass well.

araspitzu:
This PR enriches the `/payinvoice` and `/sendtonode` API to accept extra parameters for path-finding, now it's possible to specify the `feeThresholdSat` which puts a hard limit on how much the payment route can pay in terms of fees, and `maxFeePct` which is a proportional limit for the fees (note this is ignored if fees are below the threshold). Both parameters are optional and if not specified eclair will use the values from the configuration.

Fixes #1021

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -17,6 +17,7 @@
 package fr.acinq.eclair
 
 import java.util.UUID
+
 import akka.actor.ActorRef
 import akka.pattern._
 import akka.util.Timeout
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -181,11 +183,18 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.paymentInitiator ? SendPaymentToRoute(amountMsat, paymentHash, route, finalCltvExpiry)).mapTo[UUID]
   }
 
-  override def send(recipientNodeId: PublicKey, amountMsat: Long, paymentHash: ByteVector32, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, minFinalCltvExpiry_opt: Option[Long] = None, maxAttempts_opt: Option[Int] = None)(implicit timeout: Timeout): Future[UUID] = {
+  override def send(recipientNodeId: PublicKey, amountMsat: Long, paymentHash: ByteVector32, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, minFinalCltvExpiry_opt: Option[Long] = None, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Long], maxFeePct_opt: Option[Double])(implicit timeout: Timeout): Future[UUID] = {
     val maxAttempts = maxAttempts_opt.getOrElse(appKit.nodeParams.maxPaymentAttempts)
+
+    val defaultRouteParams = Router.getDefaultRouteParams(appKit.nodeParams.routerConf)
```
araspitzu: This allows to easily override only one parameter of the `Routeparams`, CC: @dpad85

PR #1025, opened 2019-06-04
Actor params: small refactoring

t-bast:
While it's convenient to be able to create most actors from a `NodeParams` instance it also breaks modularity (all actors depend on `NodeParams` which contains a lot of unrelated configuration fields).
It makes it very hard to use our code as a library.

I propose overloading the companion object's `props` and removing unused parameters from actor's constructors. The benefit is that changes are only scoped to each specific actor (by providing an overloaded `props`) and makes clearer what fields each actor really needs.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -19,7 +19,6 @@ package fr.acinq.eclair
 import java.io.File
 import java.net.InetSocketAddress
 import java.nio.file.Files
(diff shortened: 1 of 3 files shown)


t-bast:
Ping @pm47 :)

pm47:
I prefer the `trait` alternative but I'll defer the decision to @sstone.

sstone:
For now I prefer the current implementation which is simple and non-intrusive. Traits don't seem to be flexible enough so I'd rather not change things until we find a satisfying solution.

t-bast:
Closing for now in favor of a better long-term solution (tbd)

PR #1026, opened 2019-06-04
Deterministic channel keypath

araspitzu:
This PR introduces deterministic keypath construction for the derivation of the keys used in the channel operations, the construction is made in such a way that we guarantee we never reuse the same keypath in order to prevent reusing keys across channels.The schema can be summarized as follow:

Funder scenario: 
> keyPath = m/47'/2'/SHA256(funding_tx.input[0].outpoint)/0

Fundee scenario: 
> fundingPubkey.keyPath = m/47'/2'/SHA256(blockchain_height || counter)/2
> keyPath = m/47'/2'/SHA256(funding_tx.output[channel_output_index].scriptPubkey)/1

The fundee scenario involves using 2 different keypaths, one to derive the funding pubkey and another to derive all the other points required to construct `accept_channel` message. The usage of 2 different keypaths is required to compute the scriptPubkey of the funding transaction which is later used to construct the second keypath which is then used to derive all the other points for the channel.
As side effect of this PR all the static channel data, as in the data derived from the channel points exchanged in the open/accept messages, can now be derived only from the seed.

Credits: @armelinw


eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteChannelsDb.scala
```diff
@@ -49,11 +49,40 @@ class SqliteChannelsDb(sqlite: Connection) extends ChannelsDb with Logging {
...
+        using(sqlite.prepareStatement("INSERT INTO block_key_counter VALUES(?, ?)")) { insert =>
+          insert.setLong(1, blockHeight)
+          insert.setLong(2, counter)
+          insert.executeUpdate()
+        }
+      }
+    }
+  }
+
+
```
sstone: It is possible to infer the counter value from the data we read from the database (check all channels fro a given peer, then get their block heights from the short channel id) ?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/KeyManager.scala
```diff
@@ -42,6 +43,25 @@ trait KeyManager {
...
+  def deterministicPaymentPoint(localParams: LocalParams): ExtendedPublicKey
+
+  def deterministicDelayedPaymentPoint(localParams: LocalParams): ExtendedPublicKey
+
+  def deterministicHtlcPoint(localParams: LocalParams): ExtendedPublicKey
+
+  def deterministicCommitmentSecret(localParams: LocalParams, index: Long): Crypto.Scalar
+
+  def deterministicCommitmentPoint(localParams: LocalParams, index: Long): Crypto.Point
+
```
sstone: Can we keep the old names instead ?

araspitzu: Yes these changes have been reverted

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -588,7 +588,7 @@ object Peer {
   def makeChannelParams(nodeParams: NodeParams, defaultFinalScriptPubKey: ByteVector, isFunder: Boolean, fundingSatoshis: Long, channelKeyPath: DeterministicWallet.KeyPath): LocalParams = {
```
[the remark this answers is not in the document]
araspitzu: Done in 5ded959

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -51,9 +56,15 @@ object ChannelCodecs extends Logging {
...
       ("parent" | int64)).as[ExtendedPrivateKey]
 
-  val localParamsCodec: Codec[LocalParams] = (
+  def localParamsCodec(version: CommitmentVersion): Codec[LocalParams] = (
     ("nodeId" | publicKey) ::
-      ("channelPath" | keyPathCodec) ::
+      (version match {
+        case CommitmentV0 => ("channelPath" | keyPathCodec.xmap[Either[KeyPath, KeyPathFundee]](Left(_), {
+          case Left(kp) => kp
+          case Right(_) => throw new IllegalArgumentException(s"Found unexpected value for version=$version")
```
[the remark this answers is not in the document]
araspitzu: Yes, this has been fixed in 847eb06

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteChannelsDb.scala
araspitzu: The alternative is to have an atomic counter stored in volatile memory, however while this is a more elegant solution it opens up to new attacker scenarios.

araspitzu:
## Summary of the changes

### Key path generation
Eclair uses BIP32 key paths to derive all the various keys needed for channel operation, instead of generating a new random keypath when negotiating a channel we now use an external source of entropy to feed randomness in the path construction. The entropy is taken from the funding input's outpoint and in the case we're fundee we use a combination of block height and an incrementing counter to ensure we never reuse keys, the full construction is highlighted below:
 
Funder scenario:
>keyPath = m/47'/2'/SHA256(funding_tx.input[0].outpoint)/0

Fundee scenario:
>fundingPubkey.keyPath = m/47'/2'/SHA256(blockchain_height || counter)/2
>keyPath = m/47'/2'/SHA256(funding_tx.output[channel_output_index].scriptPubkey)/1

### Channel creation changes
**Funder scenario** : in order to send an `open_channel` we need to compute the basepoints for the channel operation and indeed we need the key path to do so. This means that before starting the negotiation of the channel opening we must ask the wallet to provide an input for the channel funding, we will use the input outpoint to construct the channel key path. Such particular _flow_ modified the `Channel` FSM to have some new intermediate states, namely where we ask the wallet to create a funding transaction [WAIT_FOR_FUNDING_INTERNAL_CREATED](https://github.com/ACINQ/eclair/blob/518947ae9890a1918c42ac22bdf289276c333718/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala#L55)  and where we ask the wallet to sign it [WAIT_FOR_FUNDING_SIGNED_INTERNAL](https://github.com/ACINQ/eclair/blob/518947ae9890a1918c42ac22bdf289276c333718/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala#L54). Note that after we ask the wallet to make a funding transaction we strip the signatures from it and we later query the wallet again to add new signatures, the wallet will sign the updated transaction with the script containing the funding pubkey from the remote peer (we have the remote funding pubkey only after we receive `accept_channel`). 
**Fundee scenario**: this scenario uses 2 different keypaths to operate the channel, one for the funding pubkey and one for all the other points. This is necessary because we want to use the scriptPubkey from the funding transaction to construct the keypath but to do so we need to derive out funding pubkey first, in order to do that we use a combination of block height and a _counter_. This scenario doesn't alter the channel opening flow.

Reference: [funder construction](https://github.com/ACINQ/eclair/blob/518947ae9890a1918c42ac22bdf289276c333718/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L94) and [fundee construction](https://github.com/ACINQ/eclair/blob/518947ae9890a1918c42ac22bdf289276c333718/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L99)

### Codecs and Commitments upgrades
The usage of 2 different *kinds* of keypaths (one for _funder_ scenario, two for _fundee_) forced to upgrade the `LocalParams` class to store an `Either[KeyPath, KeyPathFundee]`, we also need to upgrade the codecs to store these new objects and we must do it in a backward compatible manner. 
The codec has been modified to support [commitment versioning](https://github.com/ACINQ/eclair/blob/bdcd0aba7e6c2f6eb6acc5e665aa4d3a57cfda20/eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala#L400) and depending on the version we'll fetch the correct key paths. 

### Database layer upgrades
There is no db migration needed here but we add a new table to channel db, `block_key_counter` table. This is used to emulate an atomic auto incrementing counter used for the fundee key path construction, the alternative implementation using the `AtomicLong` has been ditched because of security concerns.

### Testing
The PR has been manual tested against eclair master, there are also [new codec tests](https://github.com/ACINQ/eclair/blob/53b3bee80cbd02970f3b121835d328a379c7c35f/eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala#L58) and backward compatibility test for fundee scenario in DATA_NORMAL: [here](https://github.com/ACINQ/eclair/blob/53b3bee80cbd02970f3b121835d328a379c7c35f/eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala#L212). To ensure previous version with yet-to-open channels can safely upgrade there are some test where we assert we can do FSM transition like "WAIT_FOR_FUNDING_CONFIRMED -> WAIT_FOR_FUNDING_LOCKED", test [here](https://github.com/ACINQ/eclair/blob/bdcd0aba7e6c2f6eb6acc5e665aa4d3a57cfda20/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/c/WaitForFundingConfirmedStateSpec.scala#L87) and [here](https://github.com/ACINQ/eclair/blob/bdcd0aba7e6c2f6eb6acc5e665aa4d3a57cfda20/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/c/WaitForFundingLockedStateSpec.scala#L82).

araspitzu:
Replaced by #1086

PR #1034, opened 2019-06-13
Close the channel if the funding tx never confirms

### LINKED ISSUE #1029: Feature Request: Periodic Sanity Check for "Offline" channels
image
I have attached a screenshot showing a channel with a funding transaction ID that **_is not in the blockchain_**.  The TxID is 30eccbb2752cc1e57a01298e38c1f0c1e9f05ada5b15effef6836433ec5b53c2
My request is that at startup, Eclair collect the funding TxIDs for all of its OFFLINE channels and ask its connected full node to getrawtransaction that TxID.  If it errors out, let the user choose to keep the channel or remove it.  

It's easy enough to force-close the channel, but I wouldn't want to do that if the channel were real.  The feature request automates the work of establishing whether or not the funding transaction actually exists.

While it makes sense that this situation (a channel with a non-existent funding Tx) should never exist, I can see a few ways that it might happen - orphaned blocks, sybil attacks, and bugs.

I am (very slowly) learning scala, so I could work on this, and it will get done faster is someone points me in the right direction.  For example, PATH/FILE.scala has startup code such as FUNCTION, and you can find examples of making RPC calls in ...

Thanks!

pm47:
If we are fundee and after 5 days the funding isn't even in the mempool,
then we give up waiting and consider the channel closed. Note that if
the funding tx stays unconfirmed forever we won't give up waiting.

If we are funder, we never give up until the funding tx is double spent,
and we periodically republish it.

This applies to states `WAIT_FOR_FUNDING_CONFIRMED` and `CLOSING` (and
also `OFFLINE`/`SYNCING` when underlying state is `WAIT_FOR_FUNDING_CONFIRMED`).

Also, added a generic way of passing context to `ElectrumClient`
requests/responses.

Fixes #1029.


pm47:
Two caveats:
- we handle the case where we send a `GetTxWithMeta` to `ElectrumWatcher` when it is disconnected (we stash it and retry when we reconnect), but we don't handle the case when `ElectrumClient` gets disconnected after `ElectrumWatcher` forwards the request. I think it is acceptable, but it's not perfect;
- we don't handle the case where bitcoind returns 'tx not found' because it is reindexing. I think it is ok too.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -121,13 +120,15 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
   for {
     exists <- getTransaction(tx.txid).map(_ => true).recover { case _ => false }
     doublespent <- if (exists) {
-      // if the tx is in the blockchain, it can't have been doublespent
+      // if the tx is in the blockchain, it can't have been double-spent
```
sstone: If `getTransaction` fails we should explicitly check the returned error which should be `No such mempool or blockchain transaction. Use gettransaction for wallet transactions. (code: -5)`  and should not contain `Blockchain transactions are still in the process of being indexed` (we cannot use error codes to differentiate here, it would also be -5)

pm47: Fixed with cc8d7fb

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -455,23 +455,10 @@ class ElectrumWallet(seed: ByteVector, client: ActorRef, params: ElectrumWallet.
...
-
-          // list the tx utxos
-          val utxos = tx.txIn.map(_.outPoint).toSet
-
-          // check if one of our transactions spends the same inputs as our tx and has a different txid; if this is the
-          // case, then the tx has been double spent
-          utxos.exists(utxo => ourSpentUtxos.contains(utxo) && !data.transactions.contains(tx.txid))
-      }
+      val isDoubleSpent = data.heights
+        .filter { case (_, height) => data.blockchain.height - height > 0 } // we only consider tx that have been confirmed (NB: > 1 means 2 confirmations)
```
sstone: We need an additional check that `height > 0` (it would be 0 or -1 for an unconfirmed)

PR #1038, opened 2019-06-14
Electrum: add data to TxIdFromPosResponse.

t-bast:
To make the response usable, we need to copy the request input (block height and transaction position).
Otherwise an actor receiving the response can't map the received transaction id to a set of short channel ids.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumClientSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumClientSpec.scala
@@ -69,15 +69,19 @@ class ElectrumClientSpec extends TestKit(ActorSystem("test")) with FunSuiteLike 
 
   test("get transaction id from position") {
     probe.send(client, GetTransactionIdFromPosition(height, position))
-    val GetTransactionIdFromPositionResponse(txid, merkle) = probe.expectMsgType[GetTransactionIdFromPositionResponse]
+    val GetTransactionIdFromPositionResponse(txid, height1, position1, merkle) = probe.expectMsgType[GetTransactionIdFromPositionResponse]
     assert(txid === referenceTx.txid)
+    assert(height1 === height)
+    assert(position1 === position)
     assert(merkle === Nil)
   }
 
   test("get transaction id from position with merkle proof") {
     probe.send(client, GetTransactionIdFromPosition(height, position, merkle = true))
-    val GetTransactionIdFromPositionResponse(txid, merkle) = probe.expectMsgType[GetTransactionIdFromPositionResponse]
+    val GetTransactionIdFromPositionResponse(txid, height1, position1, merkle) = probe.expectMsgType[GetTransactionIdFromPositionResponse]
     assert(txid === referenceTx.txid)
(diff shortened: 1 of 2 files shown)


pm47:
@sstone WDYT? this makes us diverge from electrumx protocol

t-bast:
Other options I see:

- Embedding the whole request in the response
- Allowing an arbitrary payload to be given in requests, and will be copied in responses
- Adding an optional RequestId in requests, returned in responses

WDYT?

sstone:
I like this change and it's similar to what is done with `GetMerkle` for example. We already "enrich" responses with some context info since the protocol is supposed to be used sequentially from a single client which does not match our use case.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumClientSpec.scala
```diff
@@ -69,15 +69,19 @@ class ElectrumClientSpec extends TestKit(ActorSystem("test")) with FunSuiteLike
 
   test("get transaction id from position") {
     probe.send(client, GetTransactionIdFromPosition(height, position))
-    val GetTransactionIdFromPositionResponse(txid, merkle) = probe.expectMsgType[GetTransactionIdFromPositionResponse]
+    val GetTransactionIdFromPositionResponse(txid, height1, position1, merkle) = probe.expectMsgType[GetTransactionIdFromPositionResponse]
     assert(txid === referenceTx.txid)
+    assert(height1 === height)
+    assert(position1 === position)
     assert(merkle === Nil)
   }
```
sstone: nitpick: since we know the expected response, we could instead use `probe.expectMsg()` here

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumClientSpec.scala
```diff
@@ -69,15 +69,19 @@ class ElectrumClientSpec extends TestKit(ActorSystem("test")) with FunSuiteLike
...
 
   test("get transaction id from position with merkle proof") {
     probe.send(client, GetTransactionIdFromPosition(height, position, merkle = true))
-    val GetTransactionIdFromPositionResponse(txid, merkle) = probe.expectMsgType[GetTransactionIdFromPositionResponse]
+    val GetTransactionIdFromPositionResponse(txid, height1, position1, merkle) = probe.expectMsgType[GetTransactionIdFromPositionResponse]
```
sstone: same as above. otherwise lgtm !

SomberNight:
> the protocol is supposed to be used sequentially from a single client 

Not sure what you mean by that. If you mean that there is only ever one pending request / on the wire, for sure Electrum itself does not use it like that; as that would be horribly slow. If you mean from a code flow perspective then it's true that modern Electrum has async networking and so all information available to the request is always also available where the response arrives as it is usually just a few lines below. :)

The Electrum Protocol is JSON-RPC based with each request containing an id that lets you match responses to requests, which is what you are presumably already doing.
"Enriching" responses by locally injecting some data from the corresponding requests into them is just an implementation detail. (this is also what pre-asyncio Electrum used to do)

PR #1040, opened 2019-06-17
Connect immediately on restart, then wait

pm47:
This is to allow herd effect when we restart the app and have numerous
peers.

Also removed the unnecessary transition and cleaned up delay
computation.


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -645,4 +645,9 @@ object Peer {
   }
 
   def hostAndPort2InetSocketAddress(hostAndPort: HostAndPort): InetSocketAddress = new InetSocketAddress(hostAndPort.getHost, hostAndPort.getPort)
+
+  /**
+    * Exponential backoff retry with a finite max
+    */
+  def nextReconnectionDelay(currentDelay: FiniteDuration): FiniteDuration = (2 * currentDelay).max(MAX_RECONNECT_INTERVAL)
```
araspitzu: That's not really exponential but a linear increment of the delay, it'd suggest something like `Math.pow(2, currentDelay).max(MAX_RECONNECT_INTERVAL)`

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -493,7 +492,6 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
   }
 
   onTransition {
-    case INSTANTIATING -> DISCONNECTED if nodeParams.autoReconnect && nextStateData.address_opt.isDefined => self ! Reconnect // we reconnect right away if we just started the peer
     case _ -> DISCONNECTED if nodeParams.autoReconnect => setTimer(RECONNECT_TIMER, Reconnect, 1 second, repeat = false)
```
araspitzu: How about starting with `delay = 0 second` and let it be handled by  `nextReconnectionDelay` ? In this way even on mobile we're going to connect asap, but it might trigger herd effect on nodes with a lot of connections.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
araspitzu: Okay on a second thought this wouldn't work because we need to store the number of attempts if we are going to do the exponential backoff.

pm47:
LGTM

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -54,7 +55,7 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
         channel ! INPUT_RESTORED(state)
         FinalChannelId(state.channelId) -> channel
       }.toMap
-      goto(DISCONNECTED) using DisconnectedData(previousKnownAddress, channels)
+      goto(DISCONNECTED) using DisconnectedData(previousKnownAddress, channels, MAX_RECONNECT_INTERVAL) // when we restart, we will attempt to reconnect right away, but then we'll wait
```
sstone: Randomizing `nextReconnectionDelay` here would help reduce the effect

pm47: @araspitzu can you add some randomization here? Maybe, say, between 50% and 100% of `MAX_RECONNECT_INTERVAL`?

araspitzu: cACK, in commit 22cb049 I added a randomized delay for the first reconnection attempt between 50% and 100% of the maximum delay ( so the first reconnection attempt will be between 30min and 1h)

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -55,7 +55,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
         channel ! INPUT_RESTORED(state)
         FinalChannelId(state.channelId) -> channel
       }.toMap
-      goto(DISCONNECTED) using DisconnectedData(previousKnownAddress, channels, MAX_RECONNECT_INTERVAL) // when we restart, we will attempt to reconnect right away, but then we'll wait
+      val firstReconnectionMillis = (MAX_RECONNECT_INTERVAL.toMillis / 2) + Random.nextInt((MAX_RECONNECT_INTERVAL.toMillis / 2).toInt) // between 50% and 100% of MAX_RECONNECT_INTERVAL
```
pm47: Why not type `firstReconnectionMillis` right here? Resolving to milliseconds seems a bit overkill.

Alternatively:

```scala
val firstNextReconnectionDelay: FiniteDuration = MAX_RECONNECT_INTERVAL.minus(Random.nextInt(MAX_RECONNECT_INTERVAL.toSeconds.toInt / 2).seconds)
```

araspitzu: Yes it's a bit overkill considering we're picking a value between 30min and 1h

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -492,9 +493,17 @@ class Peer(nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: Actor
...
+    *   - Eclair restart: The switchboard creates the peers and sends Init and then Peer.Reconnect to trigger reconnection attempts
+    *
+    * So when we see this transition we NO-OP because we don't want to start a Reconnect timer but the peer will receive the trigger
+    * (Connect/Reconnect) messages from the switchboard.
+    */
   onTransition {
-    case INSTANTIATING -> DISCONNECTED if nodeParams.autoReconnect && nextStateData.address_opt.isDefined => self ! Reconnect // we reconnect right away if we just started the peer
-    case _ -> DISCONNECTED if nodeParams.autoReconnect => setTimer(RECONNECT_TIMER, Reconnect, 1 second, repeat = false)
+    case INSTANTIATING -> DISCONNECTED => ()
+    case _ -> DISCONNECTED if nodeParams.autoReconnect => setTimer(RECONNECT_TIMER, Reconnect, Random.nextInt(5000).millis, repeat = false) // we add some randomization to not have peers reconnect to each other exactly at the same time
```
sstone: We could use a configuration setting instead of hardcoding 5000 milliseconds here

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -544,6 +553,8 @@ object Peer {
 
   val IGNORE_NETWORK_ANNOUNCEMENTS_PERIOD = 5 minutes
 
+  val MAX_RECONNECT_INTERVAL = 1 hour
```
sstone: MAX_RECONNECT_INTERVAL should also be a configuration setting

araspitzu: ACK for having this in the configuration settings.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
araspitzu: Isn't it a bit too much have this configurable? The randomization here prevents from starting the reconnection attempts exactly at the same time and we just want to make sure it happens as soon as possible but not exactly synchronized with the remote peer (if the remote is using eclair as well).

pm47: I think it doesn't hurt, will add.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: done with b8a056c

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: done with b8a056c

PR #1044, opened 2019-06-19
getsentinfo returns timestamps in millis

pvyhnal-generalbytes:


araspitzu:
LGTM and thanks for the contribution @pvyhnal-generalbytes

PR #1045, opened 2019-06-21
Add TLV and TLV stream codec support

t-bast:
Implement the latest TLV/TLV stream proposal (on which we all agreed during the last IRC meeting, spec PR is only waiting for final language edits and test vectors).

I took this opportunity to refactor the wire codecs (move stuff around and harmonize) and update/remove some obsolete code (features that were missing in scodec and that are now available there).

I recommend reviewing commit by commit, this will make it easier to ignore the noise generated by moving many codec from `LightningMessageCodecs` to `CommonCodecs`.


eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -45,8 +45,51 @@ object LightningMessageCodecs {
   // (for something smarter see https://github.com/yzernik/bitcoin-scodec/blob/master/src/main/scala/io/github/yzernik/bitcoinscodec/structures/UInt64.scala)
   val uint64: Codec[Long] = int64.narrow(l => if (l >= 0) Attempt.Successful(l) else Attempt.failure(Err(s"overflow for value $l")), l => l)
 
+  val uint64L: Codec[Long] = int64L.narrow(l => if (l >= 0) Attempt.Successful(l) else Attempt.failure(Err(s"overflow for value $l")), l => l)
```
pm47: NB: `uint64` codec is only suitable when we now for sure that decoded values will be < 2^63 because java longs are signed. Is it the case here? If not, then we have to use the purpose-build `UINT64` class.

eclair-core/src/test/scala/fr/acinq/eclair/wire/LightningMessageCodecsSpec.scala
```diff
@@ -58,6 +59,53 @@ class LightningMessageCodecsSpec extends FunSuite {
...
+
+    for ((long, ref) <- expected) {
+      val encoded = uint64L.encode(long).require
+      assert(ref === encoded)
+      val decoded = uint64L.decode(encoded).require.value
+      assert(long === decoded)
+    }
+  }
+
+  test("encode/decode with varint codec") {
```
pm47: Does Bitcoin Core have test vectors for the varint encoding?

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -45,8 +45,51 @@ object LightningMessageCodecs {
...
+
   val uint64ex: Codec[UInt64] = bytes(8).xmap(b => UInt64(b), a => a.toByteVector.padLeft(8))
 
+  // Bitcoin-style varint codec (CompactSize)
+  val varInt = Codec[Long](
```
pm47: Same comment as above: should we use `UInt64`?

Have you considered naming it `compactsize`? apparently there seem to be confusion about multiple varint encodings in Bitcoin.

nit²: for some (probably bad) reason we use all lowercase codec names for codecs

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -45,8 +45,51 @@ object LightningMessageCodecs {
   // (for something smarter see https://github.com/yzernik/bitcoin-scodec/blob/master/src/main/scala/io/github/yzernik/bitcoinscodec/structures/UInt64.scala)
   val uint64: Codec[Long] = int64.narrow(l => if (l >= 0) Attempt.Successful(l) else Attempt.failure(Err(s"overflow for value $l")), l => l)
 
+  val uint64L: Codec[Long] = int64L.narrow(l => if (l >= 0) Attempt.Successful(l) else Attempt.failure(Err(s"overflow for value $l")), l => l)
+
   val uint64ex: Codec[UInt64] = bytes(8).xmap(b => UInt64(b), a => a.toByteVector.padLeft(8))
 
+  // Bitcoin-style varint codec (CompactSize)
```
pm47: For reference: https://bitcoin.org/en/developer-reference#compactsize-unsigned-integers

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
```diff
@@ -0,0 +1,36 @@
...
+package fr.acinq.eclair.wire
+
+import scodec.bits.ByteVector
+
+/**
+  * Created by t-bast on 20/06/2019.
+  */
+
+// @formatter:off
+trait Tlv
```
pm47: I see you didn't make this a `sealed trait` on purpose, that makes sense.

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
```diff
@@ -0,0 +1,36 @@
...
+sealed trait OnionTlv extends Tlv
+// @formatter:on
+
+/**
+  * Generic tlv type we fallback to if we don't understand the incoming type.
+  *
+  * @param `type` tlv type.
+  * @param value  tlv value (length is implicit, and encoded as a varint).
+  */
+case class GenericTlv(`type`: Long, value: ByteVector) extends Tlv
```
pm47: Ok, so this is where the varint is used. I'm fully expecting test tlv to use very high types, so probably better to use `UInt64`, wdyt?

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+  * Created by t-bast on 20/06/2019.
+  */
+
+object TlvCodecs {
+
+  val genericTlv: Codec[GenericTlv] = (("type" | varInt) :: variableSizeBytesLong(varInt, bytes)).as[GenericTlv]
+
+  def tlvFallback(codec: Codec[Tlv]): Codec[Tlv] = discriminatorFallback(genericTlv, codec).xmap(_ match {
```
pm47: Nice find, `discriminatorFallback` seems like the right tool for this job.

Nit: should you type the codec to `DiscriminatorCodec`? Probably not, otherwise the right codec would have had this type in the definition of `discriminatorFallback`.

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
```diff
@@ -109,7 +104,7 @@ object CommonCodecs {
 
   val ipv4address: Codec[Inet4Address] = bytes(4).xmap(b => InetAddress.getByAddress(b.toArray).asInstanceOf[Inet4Address], a => ByteVector(a.getAddress))
 
-  val ipv6address: Codec[Inet6Address] = bytes(16).exmap(b => attemptFromTry(Inet6Address.getByAddress(null, b.toArray, null)), a => attemptFromTry(ByteVector(a.getAddress)))
+  val ipv6address: Codec[Inet6Address] = bytes(16).exmap(b => Attempt.fromTry(Try(Inet6Address.getByAddress(null, b.toArray, null))), a => Attempt.fromTry(Try(ByteVector(a.getAddress))))
```
pm47: Nice find

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
t-bast: Right, I must admit I copy-pasted that part at the beginning and wanted to come back to it, this is a bit unclear. I'll clean that up.

eclair-core/src/test/scala/fr/acinq/eclair/wire/LightningMessageCodecsSpec.scala
t-bast: Good idea, I'll find them.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
t-bast: I'll think about it more but indeed it probably makes sense to use UInt64 here.

> Have you considered naming it compactsize?

It's true that it's missing a clear spec reference, I don't know yet what name will be used most...I tend to prefer varInt because CompactSize could be any type (doesn't mention Int) but I'll choose what comes out of the spec and will rename if we use compact size everywhere.

> nit²: for some (probably bad) reason we use all lowercase codec names for codecs

I was wondering why the all-lower case names :)
I'll do the same for consistency.

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
t-bast: Yes, it is currently necessary to define test tlv namespaces elsewhere...
Once we have enough real tlv types defined maybe we can switch the tests to use those real tlvs and remove the test ones and make this a sealed trait.

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
t-bast: Yes I agree, I'll update that.

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
t-bast: > Nice find, discriminatorFallback seems like the right tool for this job.

Full credit goes to @sstone who found it :)

I didn't notice the `DiscriminatorCodec` provided by `scodec`, I'll try it out to see if it makes sense.

eclair-core/src/test/scala/fr/acinq/eclair/wire/LightningMessageCodecsSpec.scala
t-bast: It looks like I actually came up with almost the same test suite as they did ;)
Adding a few more for completeness.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
t-bast: As discussed offline, we should probably retire the `uint64` codec and replace it with proper `UInt64` (even though it's a bit more painful to use than `Long`).
I changed this one to really use the 64 bits.

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
t-bast: From what I understand the `DiscriminatorCodec` type is mainly used internally to `scodec` when we use functions like `discriminatedBy`. I think what we have works well, I don't see a meaningful way of changing that to be more idiomatic.

t-bast:
Yesterday's meeting confirmed that the spec PR (https://github.com/lightningnetwork/lightning-rfc/pull/607) is accepted, we're only waiting for test vectors to be added.
You can thus safely review this PR and rebase on it if you need TLV.

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
```diff
@@ -38,40 +38,41 @@ object CommonCodecs {
   // (for something smarter see https://github.com/yzernik/bitcoin-scodec/blob/master/src/main/scala/io/github/yzernik/bitcoinscodec/structures/UInt64.scala)
   val uint64: Codec[Long] = int64.narrow(l => if (l >= 0) Attempt.Successful(l) else Attempt.failure(Err(s"overflow for value $l")), l => l)
 
-  val uint64L: Codec[Long] = int64L.narrow(l => if (l >= 0) Attempt.Successful(l) else Attempt.failure(Err(s"overflow for value $l")), l => l)
-
   val uint64ex: Codec[UInt64] = bytes(8).xmap(b => UInt64(b), a => a.toByteVector.padLeft(8))
 
+  val uint64L: Codec[UInt64] = bytes(8).xmap(b => UInt64(b.reverse), a => a.toByteVector.padLeft(8).reverse)
```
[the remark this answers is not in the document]
t-bast: I hesitated, I'll do that

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
```diff
@@ -84,14 +85,18 @@ object CommonCodecs {
         case Attempt.Successful(b) =>
           b.value match {
             case 0xff => verifyMinimalEncoding(0x100000000L, uint64L.decode(b.remainder))
-            case 0xfe => verifyMinimalEncoding(0x10000L, uint32L.decode(b.remainder))
-            case 0xfd => verifyMinimalEncoding(0xfdL, uint16L.decode(b.remainder).map(b => b.map(_.toLong)))
-            case _ => Attempt.Successful(DecodeResult(b.value.toLong, b.remainder))
+            case 0xfe => verifyMinimalEncoding(0x10000L, uint32L.decode(b.remainder).map(b => b.map(UInt64(_))))
```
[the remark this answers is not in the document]
t-bast: Yeah I don't find it perfect either, but I didn't find a better way...

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
```diff
@@ -84,14 +85,18 @@ object CommonCodecs {
...
+            case 0xfd => verifyMinimalEncoding(0xfdL, uint16L.decode(b.remainder).map(b => b.map(UInt64(_))))
+            case _ => Attempt.Successful(DecodeResult(UInt64(b.value), b.remainder))
           }
         case Attempt.Failure(err) => Attempt.Failure(err)
       }
     })
 
+  // This codec can be safely used for values < 2^63 and will fail otherwise.
+  // It is useful in combination with variableSizeBytesLong to encode/decode TLV lengths because those will always be < 2^63.
+  val varlong: Codec[Long] = varint.narrow(l => if (l <= UInt64(Long.MaxValue)) Attempt.successful(l.toBigInt.toLong) else Attempt.failure(Err(s"overflow for value $l")), l => UInt64(l))
```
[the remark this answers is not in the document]
t-bast: This one need to be a `Codec[Long]` because there is no safe conversion from `UInt64` to `Long` (since `UInt64` can overflow a `Long`) and I need to use that in a `variableSizeBytesLong` codec for TLV (and `variableSizeBytesLong` requires a `Codec[Long]`).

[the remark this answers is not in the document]
t-bast: But if there's some type magic that allows me to directly use the `varint` codec in `variableSizeBytesLong` even though it's a `Codec[UInt64]` and not a `Codec[Long]`, I'm interested!

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
```diff
@@ -0,0 +1,74 @@
...
+  */
+case class TlvStream(records: Seq[Tlv]) {
+
+  // @formatter:off
+  sealed trait Error { val message: String }
+  case object RecordsNotOrdered extends Error { override val message = "tlv records must be ordered by monotonically-increasing types" }
+  case object DuplicateRecords extends Error { override val message = "tlv streams must not contain duplicate records" }
+  case object UnknownEvenTlv extends Error { override val message = "tlv streams must not contain unknown even tlv types" }
+  // @formatter:on
+
```
sstone: These Error objects should be defined in TlvStream's companion object, not in the class itself

t-bast: Done.

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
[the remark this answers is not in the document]
t-bast: Done.

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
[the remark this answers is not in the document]
pm47:
```suggestion
  val varintoverflow: Codec[Long] = varint.narrow(l => if (l <= UInt64(Long.MaxValue)) Attempt.successful(l.toBigInt.toLong) else Attempt.failure(Err(s"overflow for value $l")), l => UInt64(l))
```

Ok if you rename it to `varintoverflow` it makes more sense and is consistent with the `unit64`/`unit64overflow`.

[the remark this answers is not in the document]
t-bast: Good point, done.

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvTypes.scala
pm47: Why did you define your own `Error` class instead of reusing `scodec.Error`?

pm47: How about:

```scala
case class TlvStream(records: Seq[Tlv]) {

  records.foldLeft(Option.empty[Tlv]) {
    case (None, record) => Some(record)
    case (Some(previousRecord), record) =>
      require(record.`type` == previousRecord.`type`, "tlv streams must not contain duplicate records")
      require(record.`type` > previousRecord.`type`, "tlv records must be ordered by monotonically-increasing types")
      require(record.`type`.toBigInt % 2 != 0, "tlv streams must not contain unknown even tlv types")
      Some(record)
  }
}
```

Codec would be:

```scala
def tlvStream(codec: Codec[Tlv]): Codec[TlvStream] = {
    Codec[TlvStream](
      (s: TlvStream) => {
        val recordTypes = s.records.map(_.`type`)
        if (recordTypes.length != recordTypes.distinct.length) {
          Attempt.Failure(scodec.Err("duplicate tlv records aren't allowed"))
        } else {
          Encoder.encodeSeq(codec)(s.records.sortBy(_.`type`).toList)
        }
      },
      (buf: BitVector) =>
        Decoder.decodeCollect[List, Tlv](codec, None)(buf).map(_.map(TlvStream(_)))
    )
  }
```

t-bast:
Allright this should be ready for final review now (thanks for @pm47's love for `scodec` ;))

PR #1047, opened 2019-06-25
Handle unknown fields in network announcements

pm47:
All the data contained in `node_announcement`, `channel_announcement`
and `channel_update` is to be included in the signature, including
unknown trailing fields. We were ignoring them, causing signature
verification to fail when there was unknown fields.

In the case of `channel_update` there is a backward compatibility issue
to handle, because when persisting channel data in state `NORMAL`, we
used to store the `channel_update` followed by other data, and without
prefixing it with size information.

To work around that we use the same trick as before, based on an
additional discriminator codec. It works well but it forces us to keep
legacy codecs.


pm47:
@t-bast with your latest experience on scodec, curious on your thoughts on bf95a92?

t-bast:
I don't know of another way to achieve that with `scodec`, it looks like a good solution. I have another idea to approach this, curious to know if I'm missing something. Instead of fixing that by reading old DB entries with a decode-only codec, why not do a DB migration to add the length prefix to every stored `channel_update`? Is it breaking the signatures in a non-reconcilable way (the signature verification could strip out the length prefix for old records, couldn't it?)?

On a side-note, why don't we have the same issue with `channel_announcements` and `node_announcements`? We're not storing those in the DB?

pm47:
> why not do a DB migration

To do a DB migration, we would need the same codecs, plus we need to do the actual migration, so that's strictly more work? The problem with signatures is just for data that we receive: we currently strip data that the signer has included in the sig, which makes the verification fail.

> On a side-note, why don't we have the same issue with channel_announcements and node_announcements? We're not storing those in the DB?

We are, but those are in a single "record" (database column). The issue is when we have several independent messages stored contiguously, which the case for `DATA_NORMAL`: without length info the codec doesn't know where to stop.

pm47:
>> On a side-note, why don't we have the same issue with channel_announcements and node_announcements? We're not storing those in the DB?

>We are, but those are in a single "record" (database column). The issue is when we have several independent messages stored contiguously, which the case for DATA_NORMAL: without length info the codec doesn't know where to stop.

Argh, we *do* store a `channel_announcement` is `DATA_NORMAL`, and I was wondering why my tests weren't passing!

t-bast:
> To do a DB migration, we would need the same codecs, plus we need to do the actual migration, so that's strictly more work?

That's true, you can forget that proposal :)
It scopes the usage of that obsolete codec to only inside a DB migration though, so it could be isolated from non-legacy code, but that might not be worth the extra work.

With my limited newly acquired `scodec` experience, I don't see a better way of achieving that so lgtm for the proposed solution.

eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
```diff
@@ -32,20 +32,20 @@ import scala.concurrent.duration._
   */
 object Announcements {
 
-  def channelAnnouncementWitnessEncode(chainHash: ByteVector32, shortChannelId: ShortChannelId, nodeId1: PublicKey, nodeId2: PublicKey, bitcoinKey1: PublicKey, bitcoinKey2: PublicKey, features: ByteVector): ByteVector =
-    sha256(sha256(serializationResult(LightningMessageCodecs.channelAnnouncementWitnessCodec.encode(features :: chainHash :: shortChannelId :: nodeId1 :: nodeId2 :: bitcoinKey1 :: bitcoinKey2 :: HNil))))
+  def channelAnnouncementWitnessEncode(chainHash: ByteVector32, shortChannelId: ShortChannelId, nodeId1: PublicKey, nodeId2: PublicKey, bitcoinKey1: PublicKey, bitcoinKey2: PublicKey, features: ByteVector, unknownFields: ByteVector): ByteVector =
```
sstone: a bit off-topic, but it feels like inputs to all these `*WitnessEncode` method should be actual gossip messages

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -266,12 +266,35 @@ object ChannelCodecs extends Logging {
       ("shortChannelId" | shortchannelid) ::
       ("lastSent" | fundingLockedCodec)).as[DATA_WAIT_FOR_FUNDING_LOCKED]
 
+  // All channel_announcement's written prior to supporting unknown trailing fields had the same size
+  val noUnknownFieldsChannelAnnouncementSizeCodec: Codec[Int] = provide(430)
```
sstone: Comment should also explain that this works because we're only reading channel announcements that we create ourselves, with an empty `features` field

eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
pm47: True, not sure why it's done that way, will fix.

pm47: I know why: we don't have an object for this, only a codec.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: Very good point, I missed that completely.

pm47: Fixed with 02b23f4

pm47:
@sstone please revalidate again, had to add a cosmetic commit

PR #1051, opened 2019-06-26
Strict linting

araspitzu:
Apply strict linting to the code as dictated by the `-Xlint` compiler option, this will prevent from doing a series of mistakes typically caused by auto-tupling or type inference by the scala compiler.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/DBCompatChecker.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/DBCompatChecker.scala
@@ -39,7 +39,7 @@ object DBCompatChecker extends Logging {
     * @param nodeParams
     */
   def checkNetworkDBCompatibility(nodeParams: NodeParams): Unit =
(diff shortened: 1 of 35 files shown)


pm47:
To be honest Linting looked kind of nice no? If this doesn't fix anything (and I believe it doesn't prevent the `object`/`type` confusion), then do we really need it?

araspitzu:
> To be honest Linting looked kind of nice no? If this doesn't fix anything (and I believe it doesn't prevent the `object`/`type` confusion), then do we really need it?

Right it doesn't prevent the `object`/`type` confusion but it does prevent other kind of errors, for example [typos in the interpolation](https://github.com/ACINQ/eclair/pull/1051/files#diff-4f266586dd1c36dd04d7e87db7a2cac5L320), or when the compiler automatically lift the type to `Any` which is very often a mistake (the linting found one entry for this https://github.com/ACINQ/eclair/pull/1051/files#diff-a3f9e1bff39d3bbe4a8eee05b8f821b8R259 but in this case is okay).

araspitzu:
Rationale not good enough to be merged

PR #1053, opened 2019-06-26
Add more data to `usablebalances` API method

btcontract:
Now it additionally returns remote `nodeId` and `shortChannelId` for each channel. Also negative balances are represented as `0`s.

Should there be a related test in `ApiServiceSpec`? I'm not sure I understand how to bring `Relayer` there.


araspitzu:
In `ApiServiceSpec` you can test the format of the requests and responses, for example [here](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala#L172) we assert how the parameters for /close are handled, in L177 we mock the internal eclair API (and the necessary response for /close) then in L179 we instantiate the concrete `Service`. Lines 181-191 are the actual test, in L189 we're asserting the content of the response, in L190 that the internal eclair API was called with the expected parameters and in L191 we match the response against a mock response we have in the resources. If you don't want to go all the way down to save a mock response you can just assert the presence of some properties.

btcontract:
Thanks, I did it similar to how `peers` API test is implemented.

araspitzu:
One final remark: the `Relayer` will receive `LocalChannelUpdate` and compute the available balances only when we re-emit a channel_update. This is fine because the functionality is meant to aggregate balances for available channels, however [here](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1693) we do emit a local channel update also for transitions NORMAL -> OFFLINE and this causes the functionality to display available balances also for channels that are offline.

araspitzu:
On a second thought my last comment is actually blocking, since the API is meant to display available balances we shouldn't include OFFLINE channels in the response.

btcontract:
Does this in turn means `Relayer` treats an OFFLINE channel as operational and may try to route payments through it?

araspitzu:
It will not because the `channel_update` will contain the flag disabled for the upstream channel and the `Relayer` will fail [here](https://github.com/ACINQ/eclair/blob/a35d50d4c8cf9909ed087fe4a0a54c2deb022d3c/eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala#L347)

btcontract:
I'll filter by `Announcements.isEnabled(channelUpdate.channelFlags)` in `GetUsableBalances`, that should solve it.

btcontract:
Actually is there a point to store disabled channels in `Relayer` at all? Perhaps we can change https://github.com/ACINQ/eclair/blob/a35d50d4c8cf9909ed087fe4a0a54c2deb022d3c/eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala#L79 to be:  
```scala
case LocalChannelUpdate(_, channelId, shortChannelId, remoteNodeId, _, channelUpdate, commitments) if Announcements.isEnabled(channelUpdate.channelFlags) =>
```

btcontract:
Running tests with this change yields `UnknownNextPeer` instead of `ChannelDisabled` which I gess is meaningful difference for other peers? So `Relayer` should stay as is?

pm47:
Yes there is a point: "no channel" is different from "channel found but offline"

Let's not change sensitive relaying logic for what is essentially a helper API method.

araspitzu:
I disagree with the change it wouldn't be consistent with the current logic. The `LocalChannelUpdate`/`LocalChannelDown` messages are meant to be distinguish when a channel has it's link level in a usable state or not. For "link level" i mean that we hold an usable commitment for that channel, in fact we send the first `LocalChannelUpdate` when we reach state NORMAL for the first time and then we send  `LocalChannelDown` only when the commitment we hold is not usable anymore (i.e we move to state SHUTDOWN/NEGOTIATING/CLOSING/ERR)

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
```diff
@@ -48,7 +48,7 @@ case class ForwardFail(fail: UpdateFailHtlc, to: Origin, htlc: UpdateAddHtlc) ex
 case class ForwardFailMalformed(fail: UpdateFailMalformedHtlc, to: Origin, htlc: UpdateAddHtlc) extends ForwardMessage
 
 case object GetUsableBalances
-case class UsableBalances(canSendMsat: Long, canReceiveMsat: Long, isPublic: Boolean)
+case class UsableBalances(canSendMsat: Long, canReceiveMsat: Long, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, isPublic: Boolean)
```
pm47: Reordering parameters would make more sense and be consistent with other API objects.

```suggestion
case class UsableBalances(remoteNodeId: PublicKey, shortChannelId: ShortChannelId, canSendMsat: Long, canReceiveMsat: Long, isPublic: Boolean)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
```diff
@@ -74,7 +74,14 @@ class Relayer(nodeParams: NodeParams, register: ActorRef, paymentHandler: ActorR
   def main(channelUpdates: Map[ShortChannelId, OutgoingChannel], node2channels: mutable.HashMap[PublicKey, mutable.Set[ShortChannelId]] with mutable.MultiMap[PublicKey, ShortChannelId]): Receive = {
 
     case GetUsableBalances =>
-      sender ! channelUpdates.values.map(o => UsableBalances(o.commitments.availableBalanceForSendMsat, o.commitments.availableBalanceForReceiveMsat, o.commitments.announceChannel))
+      sender ! channelUpdates.values
+          .filter(o => Announcements.isEnabled(o.channelUpdate.channelFlags))
+          .map(o => UsableBalances(
+            canSendMsat = if (o.commitments.availableBalanceForSendMsat < 0) 0 else o.commitments.availableBalanceForSendMsat,
```
pm47: `availableBalanceForSendMsat ` and `availableBalanceForReceiveMsat ` should not return negative values imo. Also this is better implemented with a `max`.

eclair-core/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -144,6 +144,27 @@ class ApiServiceSpec extends FunSuite with ScalatestRouteTest with IdiomaticMock
       }
   }
 
+  test("'usabelnbalances' asks router for current usable balances") {
```
pm47:
```suggestion
  test("'usablebalances' asks router for current usable balances") {
```

btcontract:
Making commitment methods return `max(..., 0)` breaks certain tests (https://github.com/ACINQ/eclair/blob/master/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala#L106) so I've just changed `if ... else` to `max` in `Relayer` for now.

pm47:
@btcontract Let's update the test with:
```scala
assert(bob.stateData.asInstanceOf[DATA_NORMAL].commitments.availableBalanceForSendMsat == Math.max(pushMsat - TestConstants.Alice.channelParams.channelReserveSatoshis * 1000, 0))
```

PR #1057, opened 2019-07-02
Reject payments for expired invoices

### LINKED ISSUE #1055: Fail incoming payments for expired invoices
Eclair should fail incoming payments for expired invoices, instead the behavior is lax and accepts payments for expired requests.

araspitzu:
Ensure that we don't accept incoming HTLCs for expired invoices, fixes #1055.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
@@ -39,6 +39,7 @@ trait PaymentsDb {
 
   def getPaymentRequest(paymentHash: ByteVector32): Option[PaymentRequest]
 
+  // returns non paid payment request
   def getPendingPaymentRequestAndPreimage(paymentHash: ByteVector32): Option[(ByteVector32, PaymentRequest)]
 
   def listPaymentRequests(from: Long, to: Long): Seq[PaymentRequest]
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
@@ -155,9 +155,8 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
   }
 
   override def getPendingPaymentRequestAndPreimage(paymentHash: ByteVector32): Option[(ByteVector32, PaymentRequest)] = {
-    using(sqlite.prepareStatement("SELECT payment_request, preimage FROM received_payments WHERE payment_hash = ? AND received_at IS NULL AND (expire_at > ? OR expire_at IS NULL)")) { statement =>
+    using(sqlite.prepareStatement("SELECT payment_request, preimage FROM received_payments WHERE payment_hash = ? AND received_at IS NULL")) { statement =>
       statement.setBytes(1, paymentHash.toArray)
-      statement.setLong(2, Platform.currentTime)
       val rs = statement.executeQuery()
       if (rs.next()) {
         val preimage = rs.getByteVector32("preimage")
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
@@ -155,8 +155,9 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
   }
 
   override def getPendingPaymentRequestAndPreimage(paymentHash: ByteVector32): Option[(ByteVector32, PaymentRequest)] = {
-    using(sqlite.prepareStatement("SELECT payment_request, preimage FROM received_payments WHERE payment_hash = ? AND received_at IS NULL")) { statement =>
+    using(sqlite.prepareStatement("SELECT payment_request, preimage FROM received_payments WHERE payment_hash = ? AND received_at IS NULL AND (expire_at > ? OR expire_at IS NULL)")) { statement =>
       statement.setBytes(1, paymentHash.toArray)
+      statement.setLong(2, Platform.currentTime)
       val rs = statement.executeQuery()
       if (rs.next()) {
         val preimage = rs.getByteVector32("preimage")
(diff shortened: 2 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -155,8 +155,9 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
   }
 
   override def getPendingPaymentRequestAndPreimage(paymentHash: ByteVector32): Option[(ByteVector32, PaymentRequest)] = {
-    using(sqlite.prepareStatement("SELECT payment_request, preimage FROM received_payments WHERE payment_hash = ? AND received_at IS NULL")) { statement =>
+    using(sqlite.prepareStatement("SELECT payment_request, preimage FROM received_payments WHERE payment_hash = ? AND received_at IS NULL AND (expire_at > ? OR expire_at IS NULL)")) { statement =>
       statement.setBytes(1, paymentHash.toArray)
+      statement.setLong(2, Platform.currentTime)
```
pm47: ok because we store `expire_at` as milliseconds:

https://github.com/ACINQ/eclair/blob/667444cd0eb8917e2cf6e8f10773d46fb7f0f742/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala#L140

eclair-core/src/test/scala/fr/acinq/eclair/db/SqlitePaymentsDbSpec.scala
```diff
@@ -158,26 +158,30 @@ class SqlitePaymentsDbSpec extends FunSuite {
...
-    assert(db.listPaymentRequests(0, (Platform.currentTime.milliseconds + 1.minute).toSeconds) == Seq(i2, i1))
+    assert(db.listPaymentRequests(0, (Platform.currentTime.milliseconds + 1.minute).toSeconds) == Seq(i2, i1, i3))
     assert(db.getPaymentRequest(i1.paymentHash) == Some(i1))
     assert(db.getPaymentRequest(i2.paymentHash) == Some(i2))
+    assert(db.getPaymentRequest(i3.paymentHash) == Some(i3))
 
     assert(db.listPendingPaymentRequests(0, (Platform.currentTime.milliseconds + 1.minute).toSeconds) == Seq(i2, i1))
     assert(db.getPendingPaymentRequestAndPreimage(paymentHash1) == Some((ByteVector32.Zeroes, i1)))
     assert(db.getPendingPaymentRequestAndPreimage(paymentHash2) == Some((ByteVector32.One, i2)))
+    assert(db.getPendingPaymentRequestAndPreimage(paymentHash3) == None) // expired invoice won't be returned
```
pm47: This test isn't really meaningful because `getPendingPaymentRequestAndPreimage` uses `Platform.currentTime`, and here we use a constant `someTimestamp` way in the past.

araspitzu: In cae244e the test has been improved using a more recent timestamp for the expired invoice.

pm47:
> That's better, but still not compliant because BOLT 11 sets a default one-hour expiry

This is in fact mitigated by (h/t @araspitzu):

https://github.com/ACINQ/eclair/blob/32145e8d6a619a93d54124c1be3f69023485e19b/eclair-core/src/main/scala/fr/acinq/eclair/payment/LocalPaymentHandler.scala#L50

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
sstone: I don't like having non-deterministic db calls, why not check for expired PR at the call site instead ?

araspitzu: The `expire_at` db column was meant to provide a database level filter for non expired invoice, in this way we don't have to retrieve all of them and apply the filter later.

sstone: Sorry but I really think it's the wrong place and also makes testing more complex, I would like to see an explicit check in the upper layer

araspitzu:
@sstone i see the point of not introducing complexity at the db layer but the rationale behind it was that we can have a fast lookup for expired invoices. The case of `getPendingPaymentRequestAndPreimage` is a bit particular though because it is invoked when we receive an `update_add_htlc` and we lookup the invoice by payment hash (which is indexed) so the `expire_at` filter here is a bit redundant, also there is some added value (in code readibility) in doing the check in the  payment handler. Would it be okay for you to move the check up to the payment handler but leave the `expire_at` in the db for the other calls?

sstone:
It's not really complexity that worries me, it's rather having a "get data"  method that returns different results even if the database you're reading from has not changed. Having `expire_at` in the db is fine, but checking it there does not feel right, it should be next to the other checks in our payment handler (another option which I like less would be to explicitly pass a `current time` parameter to the method instead of using `Platform.currentTime`)

araspitzu:
Right, but then if we want to remove non-determinism from these db calls I see only two options:
- Get rid of `expire_at` and do all the checks/filters in an upper layer
- Keep `expire_at` and have the current time passed as a parameter to the call

I propose doing option #2 but as a special case for `getPendingPaymentRequestAndPreimage` we don't pass the current time to the db call instead we explicitly check the expiry in the payment handler. Other calls that use `expire_at`, namely `listPaymentRequests` and `listPendingPaymentRequests` can have the current time as function parameter.

sstone:
Ok I propose that for this PR we just fix `getPendingPaymentRequestAndPreimage` and check the expiry in payment handler.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -39,6 +39,7 @@ trait PaymentsDb {
 
   def getPaymentRequest(paymentHash: ByteVector32): Option[PaymentRequest]
 
+  // returns non paid payment request
```
sstone: I don't think that comment is true, it will return all payment requests included the ones that have been paid

araspitzu: In [SqlitePaymentDb#L158](https://github.com/ACINQ/eclair/blob/e128992c50022243b0609662ef7c39ae80cc2c8b/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala#L158) we filter for `received_at IS NULL` so this retrieves only non paid payment requests, note that we update the record setting `received_at` when receiving a payment.

sstone: Ah ok my bad.... I think we're good to go!

PR #1058, opened 2019-07-02
Enhance bitcoind checks during startup

araspitzu:
Check the `initialblockdownload` flag from `getblockchaininfo` bitcoin RPC during startup, this is an additional check to ensure we're using a fully synced bitcoin instance.


pm47:
Nice, find, if it works shouldn't we get rid of the `verificationprogress > 0.999` check altogether?

araspitzu:
Their meaning is almost overlapping but slightly different, the node is considered to be doing `IBD` whenever it has to download a large number of blocks (i.e for new nodes, or if it was offline for a long time), while the verification progress indicates at which point the verification is so far. There is a situation in which you haven't verified all the blocks but you're not technically in IBD

PR #1059, opened 2019-07-02
Added a channel version to `Commitments` object

pm47:
In a backward-compatible way, by using the fact that the first object of
a legacy `Commitments` was a public key, starting from 0x02 or 0x03.


eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -50,6 +50,14 @@ object ChannelCodecs extends Logging {
       ("path" | keyPathCodec) ::
       ("parent" | int64)).as[ExtendedPrivateKey]
 
+  val channelVersionCodec: Codec[ChannelVersion] = discriminatorWithDefault[ChannelVersion](
+    discriminator = discriminated[ChannelVersion].by(byte)
+      .typecase(0x01, provide(ChannelVersion.STANDARD))
```
araspitzu: How about encoding the type byte inside the `ChannelVersion` object?

pm47: Why would we do that? The object and its encoding are two different things.

araspitzu: Each subtype of `ChannelVersion` will be associated with a particular byte that is used in the codec, since all versions will have a version byte it seemed reasonable to have it embedded in `ChannelVersion` itself. The code would become something like:
```
.typecase(ChannelVersion.STANDARD.typeByte, provide(ChannelVersion.STANDARD))
```

pm47: You are thinking in terms of code factorization but I'm more concerned about proper separation of layers. With your proposal the codec would be partly defined in the `ChannelCodecs` file and partly in the `ChannelTypes` file. Meaning that modifying `ChannelTypes` could create backward incompatibility.

araspitzu: Yes i agree that layer separation is quite important here and we definitely don't want to define the codec in two different places, however it felt reasonable to declare the codec byte close to the definition of the scala type for a certain commitment version (after all they are always and uniquely associated), i did something similar [here](https://github.com/ACINQ/eclair/pull/1026/files#diff-ca253561a8b809afd49e646777f7c9bbR381) although i did not include it in the object.

PR #1063, opened 2019-07-04
Add eclair-cli to eclair's docker image

araspitzu:
Add the command line tool `eclair-cli` to the eclair docker image, this will allow users to easily interact with their eclair instance via `docker exec`.

cc: @Kukks


pm47:
Is there a way to have docker hub pick up this branch?

araspitzu:
> Is there a way to have docker hub pick up this branch?

Yes, the tag i pushed was successfully built https://hub.docker.com/r/acinq/eclair/builds

PR #1070, opened 2019-07-12
Add fee info to GetSentInfo

btcontract:
A `feeMsat` field has been added to `OutgoingPayment`, it is set to `0` for all states except `SUCCEDDED`.


eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -55,20 +74,21 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
     }
   }
 
-  override def updateOutgoingPayment(id: UUID, newStatus: OutgoingPaymentStatus.Value, preimage: Option[ByteVector32] = None) = {
+  override def updateOutgoingPayment(id: UUID, newStatus: OutgoingPaymentStatus.Value, preimage: Option[ByteVector32], feeMsat: Long): Unit = {
     require((newStatus == SUCCEEDED && preimage.isDefined) || (newStatus == FAILED && preimage.isEmpty), "Wrong combination of state/preimage")
```
araspitzu: How about enhancing this check to assert we can't put `feeMsat` in a payment if `newStatus == FAILED`?

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -33,14 +33,33 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
...
-  val CURRENT_VERSION = 2
+  val CURRENT_VERSION = 3
 
-  using(sqlite.createStatement()) { statement =>
-    require(getVersion(statement, DB_NAME, CURRENT_VERSION) <= CURRENT_VERSION, s"incompatible version of $DB_NAME DB found") // version 2 is "backward compatible" in the sense that it uses separate tables from version 1. There is no migration though
+  def migration123(statement: Statement): Unit = {
+    // 1 -> 2 was backwards compatible so this method is used to migrate both 1 and 2 to 3
     statement.executeUpdate("CREATE TABLE IF NOT EXISTS received_payments (payment_hash BLOB NOT NULL PRIMARY KEY, preimage BLOB NOT NULL, payment_request TEXT NOT NULL, received_msat INTEGER, created_at INTEGER NOT NULL, expire_at INTEGER, received_at INTEGER)")
     statement.executeUpdate("CREATE TABLE IF NOT EXISTS sent_payments (id TEXT NOT NULL PRIMARY KEY, payment_hash BLOB NOT NULL, preimage BLOB, amount_msat INTEGER NOT NULL, created_at INTEGER NOT NULL, completed_at INTEGER, status VARCHAR NOT NULL)")
+    statement.executeUpdate("ALTER TABLE sent_payments ADD COLUMN fee_msat INTEGER DEFAULT 0 NOT NULL")
```
araspitzu: Why not using a nullable field here?

btcontract: And use `feeMsat: Option[Long]` instead of `0` by default in `OutgoingPayment`?

araspitzu: Right, having `feeMsat: Option[Long]` may be a stretch, the alternative would be to do the conversion in the db layer (keeping `feeMsat: Long` and nullable in the table) but that's even worse. I think we can keep it NOT NULL and use `0` as default value, after all if the payment wasn't successful we spent exactly `0` msat in fees.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -77,7 +77,7 @@ case class IncomingPayment(paymentHash: ByteVector32, amountMsat: Long, received
   * @param completedAt absolute time in seconds since UNIX epoch when the payment succeeded.
```
araspitzu: Please update the javadoc

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
btcontract: I think it can be made even stronger:
```scala
override def updateOutgoingPayment(id: UUID, newStatus: OutgoingPaymentStatus.Value, preimage: Option[ByteVector32], feeMsat: Long): Unit = {
    require((newStatus == SUCCEEDED && preimage.isDefined) || (newStatus != SUCCEEDED && preimage.isEmpty && feeMsat == 0L), "Wrong combination of state/preimage")
```

araspitzu: Yes looks good

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
btcontract: So no changes here then?

araspitzu: Let's wait for more reviewers but for me is okay as it is, cc @sstone @pm47

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -33,14 +33,33 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
   import SqliteUtils.ExtendedResultSet._
 
   val DB_NAME = "payments"
-  val CURRENT_VERSION = 2
+  val CURRENT_VERSION = 3
 
-  using(sqlite.createStatement()) { statement =>
-    require(getVersion(statement, DB_NAME, CURRENT_VERSION) <= CURRENT_VERSION, s"incompatible version of $DB_NAME DB found") // version 2 is "backward compatible" in the sense that it uses separate tables from version 1. There is no migration though
+  def migration123(statement: Statement): Unit = {
```
pm47: `migaration123` doesn't mean anything, I'd prefer taking inspiration from `SqliteAuditDb` (with `migration12` and `migration23`, maybe one of them being noop).

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
araspitzu: I don't have strong feelings about this but having it nullable forces the use of `Option[Long]` which is very correct but will lead to (IMHO) less elegant code where we need to check for the presence of the `feeMsat` on a finalized outgoing payment. On the other hand one could argue that using a default value of `0` is logically correct because a non finalized payment pays exactly `0` msat in fees.

btcontract: FWIW on a client side I'm working on using `0` is bit more comfortable as I can just `value + fee` instead of `value + feeOpt.getOrElse(0)`.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -33,14 +33,36 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
...
   val DB_NAME = "payments"
-  val CURRENT_VERSION = 2
+  val CURRENT_VERSION = 3
 
-  using(sqlite.createStatement()) { statement =>
-    require(getVersion(statement, DB_NAME, CURRENT_VERSION) <= CURRENT_VERSION, s"incompatible version of $DB_NAME DB found") // version 2 is "backward compatible" in the sense that it uses separate tables from version 1. There is no migration though
+  // 1 -> 2 was backwards compatible so this method is noop
+  def migration12(statement: Statement): Unit = ()
+
+  def migration23(statement: Statement): Unit = {
```
pm47: That doesn't seem right: if we are in version 2 then we already have those tables.

That should be:

```scala
def migration23(statement: Statement): Unit = {
    statement.executeUpdate("CREATE INDEX IF NOT EXISTS payment_hash_idx ON sent_payments(payment_hash)")
  }

Right?
```

pm47: Also the original comment for version 2 was much more informative

btcontract: All `CREATE TABLE/INDEX` has `IF NOT EXISTS` which neutralizes them.

pm47: Sure, that's how the `case CURRENT_VERSION =>` (because starting from nothing or being up to date is the same from a migration point of view) but there is no need to use that in the `2->3` scenario.

araspitzu:
Needs rebase on master and the new field `feeMsat: Long` should become `fee: MilliSatoshi`

t-bast:
This change has been included in #1130 :tada:

PR #1071, opened 2019-07-12
add in utility script to startup a regtest Network

n1bor:
it has 2 GUI nodes (1 and 5) and 3 cli nodes 2,3,4
Network is:
1<->2<->3<->4<->5
includes aliases for cli interaction with all nodes

### DIFF
--- a/contrib/regtest.sh
+++ b/contrib/regtest.sh
@@ -1,0 +1,197 @@
+#set -x
+
+# VARs
+
+# location of data dirs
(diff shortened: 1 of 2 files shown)


sstone:
This is interesting but should not go into our source repo. Instead, it should be a standalone project with an explanation and a few examples of how to use it do understand how LN works (I use something very similar for coding sessions/hands-on demos: https://github.com/sstone/ln-in-action).

PR #1072, opened 2019-07-12
Extended Queries: use TLV format for optional data

sstone:
Optional query extensions now use TLV instead of a custom format.
Flags are encoded as varint instead of bytes as originally proposed. With the current proposal they will all fit on a single byte, but will be much easier to extends this way.


eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
```diff
@@ -229,69 +229,150 @@ case class PerHopPayload(shortChannelId: ShortChannelId,
...
                          outgoingCltvValue: Long)
 
-// @formatter:off
+
+/**
+  * Optional TLV-based query message that can be appended to QueryShortChannelIds
+  * @param encoding 0 means uncompressed, 1 means compressed with zlib
+  * @param array array of query flags, each flags specifies the info we want for a given channel
+  */
+case class EncodedQueryFlags(encoding: EncodingType, array: List[Long]) extends Tlv
```
pm47: I feel like the TLV should be defined in a separate file.

pm47: Also this is not right: you should define a `sealed trait` for each TLV namespace

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
```diff
@@ -229,69 +229,150 @@ case class PerHopPayload(shortChannelId: ShortChannelId,
...
 
 case class EncodedShortChannelIds(encoding: EncodingType,
                                   array: List[ShortChannelId])
 
-case class EncodedQueryFlags(encoding: EncodingType,
-                             array: List[Byte])
 
 case class QueryShortChannelIds(chainHash: ByteVector32,
+                                shortChannelIds: EncodedShortChannelIds,
+                                extensions: TlvStream[Tlv]) extends RoutingMessage with HasChainHash {
```
pm47: Once TLV are properly separated in traits, this should be typed.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
```diff
@@ -229,69 +229,150 @@ case class PerHopPayload(shortChannelId: ShortChannelId,
...
+  val queryFlags: Option[EncodedQueryFlags] = extensions.records collectFirst { case flags: EncodedQueryFlags => flags }
```
pm47: Should we make a generic implicit `.getRecord[T]` available for all TLV, as we will always use this pattern? cc @t-bast

t-bast: That's a good idea!

t-bast: Hum looks like we can't do that without reflection because of type erasure...
I'll let the Scala guru try to find a way to do this but my naive attempt didn't work, Scala generics aren't C++ templates ;)

pm47: We can simply do:
```scala
case class TlvStream[T <: Tlv](records: Traversable[T], unknown: Traversable[GenericTlv] = Nil) {
  def get[R <: T](implicit m: Manifest[R]): Option[R] = records.collectFirst { case r: R => r }
}
```

And then:
```scala
test("extract record by type") {
  val stream = TlvStream[TestTlv](Seq(TestType254(42), TestType1(42)), Seq(GenericTlv(13, hex"2a"), GenericTlv(11, hex"2b")))
  assert(stream.get[TestType1].contains(TestType1(42)))
  assert(stream.get[TestType2].isEmpty)
  assert(stream.get[OtherType1].isEmpty) // <- doesn't compile
}
```

😎

t-bast: Yes but if we use `Manifest` this won't work an Android right?

pm47: I'm not sure, let's ask @sstone

sstone: Not sure how yet but it does seem to work on Android (with a real device) I'll update this discussion asap

sstone: Implemented with:

``` scala
case class TlvStream[T <: Tlv](records: Traversable[T], unknown: Traversable[GenericTlv] = Nil) {
  def get[R <: T : ClassTag]: Option[R] = records.collectFirst { case r: R => r }
}
```

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -76,7 +77,7 @@ case class NodeParams(keyManager: KeyManager,
                       routerConf: RouterConf,
                       socksProxy_opt: Option[Socks5ProxyParams],
                       maxPaymentAttempts: Int) {
-
+  require(feeBase <= MilliSatoshi(0xFFFFFFFFL))
```
pm47: This should be in `NodeParams.makeNodeParams` for consistency, with a human understandable message (and also a comment explaining why we have to cap the value)

PR #1073, opened 2019-07-12
Replace traits by bitfield for ChannelVersion

pm47:
It is better suited for composition of features.

This  paves the way for #1026.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
@@ -19,14 +19,14 @@ package fr.acinq.eclair.channel
 import java.util.UUID
 
 import akka.actor.ActorRef
-import fr.acinq.bitcoin.Crypto.{PublicKey}
+import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{ByteVector32, DeterministicWallet, OutPoint, Satoshi, Transaction}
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -52,7 +52,7 @@ object ChannelCodecs extends Logging {
 
   val channelVersionCodec: Codec[ChannelVersion] = discriminatorWithDefault[ChannelVersion](
     discriminator = discriminated[ChannelVersion].by(byte)
-      .typecase(0x01, provide(ChannelVersion.STANDARD))
+      .typecase(0x01, bits(4 * 8).as[ChannelVersion])
```
araspitzu: Why hardcoding the bit values instead of using `ChannelVersion.STANDARD.bits` here?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -223,8 +223,10 @@ object ChannelFlags {
   val Empty = 0x00.toByte
 }
 
-sealed trait ChannelVersion
+case class ChannelVersion(bits: BitVector) {
+  require(bits.size == 4 * 8, "channel version takes 4 bytes")
```
sstone: Size requirement is hardcoded in 3 different places, which not define it in `ChannelVersion`'s companion object ?

PR #1079, opened 2019-07-17
Handle fulfill not acked upstream

t-bast:


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -849,8 +849,11 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         }
       }
 
-    case Event(c@CurrentBlockCount(count), d: DATA_NORMAL) if d.commitments.timedoutOutgoingHtlcs(count).nonEmpty =>
-      handleLocalError(HtlcTimedout(d.channelId, d.commitments.timedoutOutgoingHtlcs(count)), d, Some(c))
+    case Event(c@CurrentBlockCount(count), d: DATA_NORMAL) if d.commitments.timedOutOutgoingHtlcs(count).nonEmpty =>
```
t-bast: It's a bit redundant to have these two cases duplicated in 4 states...
Shall I refactor that? Maybe by creating a partial receive that handles these two cases and re-using it in the 4 states that need it?

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -188,6 +189,7 @@ object NodeParams {
       maxHtlcValueInFlightMsat = UInt64(config.getLong("max-htlc-value-in-flight-msat")),
       maxAcceptedHtlcs = maxAcceptedHtlcs,
       expiryDeltaBlocks = config.getInt("expiry-delta-blocks"),
+      upstreamTimeoutSafetyBlocks = config.getInt("upstream-timeout-safety-blocks"),
```
araspitzu: Should we check that `upstreamTimeoutSafetyBlocks` is strictly greater than `expiryDeltaBlocks`? If the cltv delta is smaller than the safety timeout the htlc will be considered "almost" timed out as soon as we receive a fulfill from downstream.

t-bast: That's the other way around :)
`upstreamTimeoutSafetyBlocks` needs to be strictly smaller than `expiryDeltaBlocks`, should I assert that in the code itself and prevent the node from starting if that's not satisfied?

araspitzu: Oh indeed it has to be strictly smaller :), i think an assertion in `NodeParams` should be enough. The user configuring this parameter must know that if they are too close it risks to have unnecessary channel closures.

eclair-core/src/main/resources/reference.conf
```diff
@@ -68,6 +68,11 @@ eclair {
   max-to-local-delay-blocks = 2016 // maximum number of blocks that we are ready to accept for our own delayed outputs (2016 ~ 2 weeks)
   mindepth-blocks = 3
   expiry-delta-blocks = 144
+  // When we receive the pre-image for an HTLC and want to fulfill it but the upstream peer stops responding, we want to
+  // avoid letting its HTLC-timeout transaction become enforceable on-chain (otherwise there is a race condition between
+  // our HTLC-success and their HTLC-timeout).
+  // We will close the channel when the HTLC-timeout will happen in less than this number.
+  upstream-timeout-safety-blocks = 6
```
```suggestion
  fulfill-safety-before-timeout-blocks = 6
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
```diff
@@ -388,7 +392,65 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
...
+
+    sender.send(alice, INPUT_DISCONNECTED)
+    sender.send(bob, INPUT_DISCONNECTED)
+    awaitCond(alice.stateName == OFFLINE)
+    awaitCond(bob.stateName == OFFLINE)
+
+    // We simulate a pending fulfill on that HTLC but not relayed.
+    // When it is close to expiring upstream, we should close the channel.
+    sender.send(commandBuffer, CommandSend(htlc.channelId, htlc.id, CMD_FULFILL_HTLC(htlc.id, r, commit = true)))
+    sender.send(bob, CurrentBlockCount(htlc.cltvExpiry - bob.underlyingActor.nodeParams.fulfillSafetyBeforeTimeoutBlocks))
```
sstone: We should also check that a `HtlcWillTimeoutUpstream` error has been published.

PR #1080, opened 2019-07-17
Handle fees increases when channel is OFFLINE

araspitzu:
### Rationale
We need to check on the feerates used in the commitment transaction even when the channel is in state offline, the risk here is to end up holding a commitment that never confirms once broadcast.

### Implementation
The `Channel` class has been tweaked to handle `CurrentFeerates` events when OFFLINE, the fees mismatch is checked against `max-feerate-mismatch` and we only perform the check if the network fees are increasing (higher than current feerate of the commitment). The check is performed both in funder and fundee scenarios.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1428,6 +1428,16 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       // -> in CLOSING we either have mutual closed (so no more htlcs), or already have unilaterally closed (so no action required), and we can't be in OFFLINE state anyway
       handleLocalError(HtlcTimedout(d.channelId, d.commitments.timedoutOutgoingHtlcs(count)), d, Some(c))
 
+    case Event(c@CurrentFeerates(feeratesPerKw), d: HasCommitments) =>
```
pm47: This needs to also be done in state `SYNCING` (we can stay in that state a long time in some corner cases).

Can we also factorize in a `handleCurrentFeerate` handler?

araspitzu: ACK for applying the logic in `SYNCING` too, factorizing the code in a handler is nice too but the fee-mismatch checks are slightly different between the normal case and `OFFLINE`/`SYNCING`. For example when `OFFLINE` we only want to fail the channel if the fees are raising over a certain threshold, as opposed to a more general mismatch used during `NORMAL` or `SHUTDOWN`

araspitzu: Commit e98301 handles the feerates in `SYNCING` and refactored the feerate logic into 2 different handlers, one used during normal channel operation and the other used when the channel is not operational (`OFFLINE`/`SYNCING`)

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1732,6 +1722,35 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+    * This is used to check for the commitment fees when the channel is not operational but we have something at stake
+    * @param c the new feerates
+    * @param d the channel commtiments
+    * @return
+    */
+  def handleOfflineFeerate(c: CurrentFeerates, d: HasCommitments) = {
+    val networkFeeratePerKw = c.feeratesPerKw.blocks_2
+    val currentFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw
+    // if the fees are too high we risk to not be able to confirm our current commitment
+    if(networkFeeratePerKw > currentFeeratePerKw && Helpers.isFeeDiffTooHigh(currentFeeratePerKw, networkFeeratePerKw, nodeParams.maxFeerateMismatch)){
```
pm47: `networkFeeratePerKw > currentFeeratePerKw` we should do this check in `handleCurrentFeerate` too. We don't want to close the channel if the commitment fees are too high right?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1732,6 +1722,35 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+
+  /**
+    * This is used to check for the commitment fees when the channel is not operational but we have something at stake
+    * @param c the new feerates
+    * @param d the channel commtiments
+    * @return
+    */
+  def handleOfflineFeerate(c: CurrentFeerates, d: HasCommitments) = {
+    val networkFeeratePerKw = c.feeratesPerKw.blocks_2
+    val currentFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw
```
pm47: Defining `currentFeeratePerKw ` is nice and should also be done in `handleCurrentFeerate`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
araspitzu: Why should we check for the mismatch only if the network fees are higher than the current fees? In the non-offline case this can happen and is perfectly valid, the offline case checks only for increases because if fees go down we're safe with the current commitment fees.

araspitzu:
Rebased on master, and a log entry has been added for when an offline channel gets failed due to a fee mismatch.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1728,6 +1718,37 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+    * @param c the new feerates
+    * @param d the channel commtiments
+    * @return
+    */
+  def handleOfflineFeerate(c: CurrentFeerates, d: HasCommitments) = {
+    val networkFeeratePerKw = c.feeratesPerKw.feePerBlock(target = nodeParams.onChainFeeConf.feeTargets.commitmentBlockTarget)
+    val currentFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw
+    // if the fees are too high we risk to not be able to confirm our current commitment
+    if(networkFeeratePerKw > currentFeeratePerKw && Helpers.isFeeDiffTooHigh(currentFeeratePerKw, networkFeeratePerKw, nodeParams.onChainFeeConf.maxFeerateMismatch)){
+      log.warning(s"closing OFFLINE ${d.channelId} due fee mismatch: currentFeeratePerKw=$currentFeeratePerKw networkFeeratePerKw=$networkFeeratePerKw")
```
```suggestion
      log.warning(s"closing OFFLINE ${d.channelId} due to fee mismatch: currentFeeratePerKw=$currentFeeratePerKw networkFeeratePerKw=$networkFeeratePerKw")
```

pm47:
How about we add a parameter to disable closing in case of fee mismatch? I can see that being useful in some cases.

pm47:
e.g. `close-on-feerate-mismatch` right below `max-feerate-mismatch` with a default to true.

araspitzu:
> e.g. `close-on-feerate-mismatch` right below `max-feerate-mismatch` with a default to true.

Yeah that's a good idea but i would call it `close-on-offline-feerate-mismatch`. Users can try (conservatively) enabling the option and if it doesn't close unnecessary channels it should just be set to true.

eclair-core/src/main/resources/reference.conf
```diff
@@ -90,6 +90,7 @@ eclair {
     // maximum local vs remote feerate mismatch; 1.0 means 100%
     // actual check is abs((local feerate - remote fee rate) / (local fee rate + remote fee rate)/2) > fee rate mismatch
     max-feerate-mismatch = 1.56 // will allow remote fee rates up to 8x bigger or smaller than our local fee rate
+    close-on-offline-feerate-mismatch = true // if false eclair will not close offline channel on feerate mismatch
```
```suggestion
    close-on-offline-feerate-mismatch = true // if false eclair will not close offline channels on feerate mismatch
```

I don't think this comment adds much information. How about the good old "do not change this unless you know what you are doing" instead?

araspitzu: I tried my best :) but in fact this config key should be self explanatory, will add.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1419,6 +1403,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     // -> in CLOSING we either have mutual closed (so no more htlcs), or already have unilaterally closed (so no action required), and we can't be in OFFLINE state anyway
     case Event(c: CurrentBlockCount, d: HasCommitments) => handleNewBlock(c, d)
 
+    case Event(c: CurrentFeerates, d: HasCommitments) if nodeParams.onChainFeeConf.closeOnOfflineMismatch =>
```
araspitzu: Alternatively we can do the check for `nodeParams.onChainFeeConf.closeOnOfflineMismatch` inside `handleOfflineFeerate` with the advantage that we can print a log line if the mismatch is over the threshold but the flag is disabled, WDTY @pm47 ?

pm47: I think that's better, we want to see the warning even if we do nothing.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
```diff
@@ -459,4 +464,70 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
...
+    val localFeeratePerKw = aliceStateData.commitments.localCommit.spec.feeratePerKw
+    val tooHighFeeratePerKw = ((alice.underlyingActor.nodeParams.onChainFeeConf.maxFeerateMismatch + 6) * localFeeratePerKw).toLong
+    val highFeerate = FeeratesPerKw.single(tooHighFeeratePerKw)
+
+    // alice is funder
+    sender.send(alice, CurrentFeerates(highFeerate))
+    alice2blockchain.expectMsg(PublishAsap(aliceCommitTx))
+  }
+
+  test("handle feerate changes while offline (disabled flag)", Tag("disable-offline-mismatch")) { f =>
```
pm47:
```suggestion
  test("handle feerate changes while offline (don't close on mismatch)", Tag("disable-offline-mismatch")) { f =>
```

PR #1082, opened 2019-07-18
Made using/storing/sending consistent

pm47:


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2201,7 +2201,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     context.system.eventStream.publish(NetworkFeePaid(self, remoteNodeId, channelId, tx, fee, desc))
   }
 
-  def store[T](d: T)(implicit tp: T <:< HasCommitments): T = {
```
sstone: I had forgotten about this. New method signature is so much better!

pm47:
@sstone had to merge from master, your review has been discarded.

PR #1083, opened 2019-07-20
Configurable transaction confirmation target

### LINKED ISSUE #1028: Feature request: better (manual) fee control
In the last 4 months I paid a total of around ~0.22 BTC for roughly 600 channel opening/closing transactions even though my opening tx are always only paying a maximum of 1-5 sat/b and only maybe 75% of those channels were opened by myself. Pretty much all of those tx happened during times of insignificant network congestion. As soon as congestion goes up, things will get much worse and 0.22 BTC is already pretty bad!

If possible, this would be the best imho:

- manual fee setting whenever humanly possible (the algos just suck big time)
- stop overpaying (sometimes by more than 400% according to blockstream.info)
- use a standard target of 144 blocks instead of 2
- an option to manually adjust this target via config
- make it simple to get stuck TXs unstuck via CFP
- use less TXs, but that's already in the works: https://github.com/ACINQ/eclair/pull/1026
- get rid of unnecessary force closures (mostly an LND issue, afaik)
- option to combine closing and opening TXs

Hopefully at least some of it can be implemented before the bull market starts picking up steam. Otherwise I fear that operating a routing node will become VERY expensive. I'm already paying hundreds of dollars each month...

araspitzu:
### Rationale
Fee computation is a sensitive topic and directly affects the profitability of node operators, thus saving as much as possible is a priority and a nice-to-have feature. Currently eclair uses a very conservative estimation for most of the transactions, resulting in unnecessary high fees under certain scenarios; this PR aims at making this behavior configurable and give the node operators the opportunity to fine tune on chain transaction costs.

### High level overview
There is a new configuration block that can be used to define default values for the block target used in the fee calculation for various transactions, currently supported are types: "funding", "commitment", "mutual-close" and "claim-main". Eclair will target the block number defined there when computing the fees, this means that according to the current feerate (still taken from the usual fee providers) we calculate the absolute transaction fee trying to target the confirmation on the N-th block where N is user defined default. 
- `funding` is the block target for the funding transaction (you can override this when opening a channel)
- `commitment` is the block target using when computing the commitment transaction, this is used when force closing the channel.
- `mutual-close` block target used when negotiating the closing transaction fee with the remote peer.
- `claim-main` block target is used to compute fee for the claim-main transaction, this used to spend our main channel output to our on-chain wallet 

### Implementation details
The `claim-main` block target is applied to both the claim-main-delayed and claim-main transaction, those are logically equal and none of them is competing with the counterparty for confirmation.
 
We still don't provide configurable block target for htlc-2nd stage transactions and main penalty transaction, this is due to the fact that in these scenarios we're competing with the counterparty and we 
must act swiftly.
  
### Transactions and block target
- Funding transaction: blockTarget=`funding`
- Commitment transaction: blockTarget=`commitment`
- HTLC-success transaction: blockTarget=2 (hardcoded)
- HTLC-timeout transaction: blockTarget=2 (hardcoded)
- Main penalty transaction: blockTarget=2 (hardcoded)
- HTLC penalty transaction: blockTarget=2 (hardcoded)
- Mutual close transaction: blockTarget=`mutual-close` ( must be >= `commitment`)
- Claim main output transaction (remote unilateral-close): blockTarget=`claim-main`
- Claim main delayed transaction (local unilateral-close): blockTarget=`claim-main`
- Claim htlc-delayed transaction: blockTarget=`claim-main`

Fixes #1028


eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -188,7 +188,8 @@ class Setup(datadir: File,
           blocks_6 = config.getLong("default-feerates.delay-blocks.6"),
           blocks_12 = config.getLong("default-feerates.delay-blocks.12"),
           blocks_36 = config.getLong("default-feerates.delay-blocks.36"),
-          blocks_72 = config.getLong("default-feerates.delay-blocks.72")
+          blocks_72 = config.getLong("default-feerates.delay-blocks.72"),
+          blocks_144 = config.getLong("default-feerates.delay-blocks.144")
```
sstone: I'm not sure that a target of 144 blocks is necessary or even realistic, afaict all the estimators that we use will return the same estimates when target > 60 blocks

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -310,7 +311,7 @@ object Commitments {
     (commitments1, fee)
   }
 
-  def receiveFee(commitments: Commitments, fee: UpdateFee, maxFeerateMismatch: Double): Commitments = {
+  def receiveFee(feeTargets: FeeTargets, commitments: Commitments, fee: UpdateFee, maxFeerateMismatch: Double, feeEstimator: FeeEstimator): Commitments = {
```
sstone: commitments should be the first parameter

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -423,21 +424,26 @@ object Helpers {
       }
     }
 
-    def firstClosingFee(commitments: Commitments, localScriptPubkey: ByteVector, remoteScriptPubkey: ByteVector)(implicit log: LoggingAdapter): Satoshi = {
+    def firstClosingFee(feeratePerKw: Long, commitments: Commitments, localScriptPubkey: ByteVector, remoteScriptPubkey: ByteVector)(implicit log: LoggingAdapter): Satoshi = {
```
sstone: `feeratePerKw` should be the last parameter here imho

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -490,16 +496,15 @@ object Helpers {
       * @param commitments our commitment data, which include payment preimages
       * @return a list of transactions (one per HTLC that we can claim)
       */
-    def claimCurrentLocalCommitTxOutputs(keyManager: KeyManager, commitments: Commitments, tx: Transaction)(implicit log: LoggingAdapter): LocalCommitPublished = {
+    def claimCurrentLocalCommitTxOutputs(feeTargets: FeeTargets, keyManager: KeyManager, commitments: Commitments, tx: Transaction, feeEstimator: FeeEstimator)(implicit log: LoggingAdapter): LocalCommitPublished = {
```
sstone: `feeTargets` should be after `feeEstimator`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -650,7 +654,7 @@ object Helpers {
       *
       * @return a [[RevokedCommitPublished]] object containing penalty transactions if the tx is a revoked commitment
       */
-    def claimRevokedRemoteCommitTxOutputs(keyManager: KeyManager, commitments: Commitments, tx: Transaction, db: ChannelsDb)(implicit log: LoggingAdapter): Option[RevokedCommitPublished] = {
+    def claimRevokedRemoteCommitTxOutputs(feeTargets: FeeTargets, keyManager: KeyManager, commitments: Commitments, tx: Transaction, db: ChannelsDb, feeEstimator: FeeEstimator)(implicit log: LoggingAdapter): Option[RevokedCommitPublished] = {
```
sstone: `feeTargets` should be after `feeEstimator`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -619,11 +624,10 @@ object Helpers {
       * @param tx                       the remote commitment transaction that has just been published
       * @return a list of transactions (one per HTLC that we can claim)
       */
-    def claimRemoteCommitMainOutput(keyManager: KeyManager, commitments: Commitments, remotePerCommitmentPoint: PublicKey, tx: Transaction)(implicit log: LoggingAdapter): RemoteCommitPublished = {
+    def claimRemoteCommitMainOutput(feeTargets: FeeTargets, keyManager: KeyManager, commitments: Commitments, remotePerCommitmentPoint: PublicKey, tx: Transaction, feeEstimator: FeeEstimator)(implicit log: LoggingAdapter): RemoteCommitPublished = {
```
sstone: `feeTargets` should be after `feeEstimator`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -562,7 +567,7 @@ object Helpers {
       * @param tx           the remote commitment transaction that has just been published
       * @return a list of transactions (one per HTLC that we can claim)
       */
-    def claimRemoteCommitTxOutputs(keyManager: KeyManager, commitments: Commitments, remoteCommit: RemoteCommit, tx: Transaction)(implicit log: LoggingAdapter): RemoteCommitPublished = {
+    def claimRemoteCommitTxOutputs(feeTargets: FeeTargets, keyManager: KeyManager, commitments: Commitments, remoteCommit: RemoteCommit, tx: Transaction, feeEstimator: FeeEstimator)(implicit log: LoggingAdapter): RemoteCommitPublished = {
```
sstone: `feeTargets` should be after `feeEstimator`

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -65,6 +65,7 @@ object Transactions {
   case object OutputNotFound extends RuntimeException(s"output not found (probably trimmed)") with TxGenerationSkipped
   case object AmountBelowDustLimit extends RuntimeException(s"amount is below dust limit") with TxGenerationSkipped
 
+  case class FeeTargets(fundingBlockTarget: Int, commitmentBlockTarget: Int, mutualCloseBlockTarget: Int, claimMainBlockTarget: Int)
```
sstone: Why is `FeeTargets` not in `fr.acinq.eclair.blockchain.fee' alongside `FeeEstimator` ?

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -192,7 +200,10 @@ object NodeParams {
       toRemoteDelayBlocks = config.getInt("to-remote-delay-blocks"),
       maxToLocalDelayBlocks = config.getInt("max-to-local-delay-blocks"),
       minDepthBlocks = config.getInt("mindepth-blocks"),
-      smartfeeNBlocks = 3,
+      feeEstimator = new FeeEstimator {
```
sstone: fee estimator should be provided to `makeNodeParams`, not defined here

araspitzu: Addressed in f0649dd

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
araspitzu: ACK, I moved `FeeTargets` in `FeeEstimator.scala` in commit f0649dd

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
araspitzu: Agree, done in f0649dd

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
araspitzu: Done in f0649dd

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
araspitzu: Also done in f0649dd

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
araspitzu: Done in f0649dd

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
araspitzu: Agreed, reorg'd in f0649dd

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
araspitzu: Done in f0649dd

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
araspitzu: We could have 144 to be the fallback-lowest fee applicable even if the estimator doesn't give an exact estimate for more than 60 blocks.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package fr.acinq.eclair.blockchain.fee
+
+trait FeeEstimator {
+
+  def getFeeratePerKb(target: Int) : Long
+  def getFeeratePerKw(target: Int) : Long
+
```
sstone: nitpick: formatting (needs a newline after  `getFeeratePerKb`)

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -38,6 +40,17 @@ object TestConstants {
   val pushMsat = 200000000L
   val feeratePerKw = 10000L
 
+  class TestFeeEstimator extends FeeEstimator {
+    private var currentFeerates = FeeratesPerKw.single(feeratePerKw)
+
+    override def getFeeratePerKb(target: Int): Long = feerateKw2KB(currentFeerates.feePerBlock(target))
+    override def getFeeratePerKw(target: Int): Long = currentFeerates.feePerBlock(target)
```
sstone: same formatting issue as above

pm47:
> HTLC-success transaction: blockTarget=2 (hardcoded)
> HTLC-timeout transaction: blockTarget=2 (hardcoded)
> Main penalty transaction: blockTarget=2 (hardcoded)
> HTLC penalty transaction: blockTarget=2 (hardcoded)

@t-bast I think this is the value we were looking for for #1079.

t-bast:
> @t-bast I think this is the value we were looking for for #1079.

Do you think we should hard-code this for 1079? I think it's better to keep it configurable, because this is linked to the amount of time you can expect to be safely offline, so if you are on a not-so-reliable node (raspberry pi for example) you might want to use a value higher than 2 here.

pm47:
> Do you think we should hard-code this for 1079?

Not necessarily, it's just that now we have a rationale for the default value. OTOH changing the default commitment target is definitely risky.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -423,21 +424,26 @@ object Helpers {
       }
     }
 
-    def firstClosingFee(commitments: Commitments, localScriptPubkey: ByteVector, remoteScriptPubkey: ByteVector)(implicit log: LoggingAdapter): Satoshi = {
+    def firstClosingFee(commitments: Commitments, localScriptPubkey: ByteVector, remoteScriptPubkey: ByteVector, feeratePerKw: Long)(implicit log: LoggingAdapter): Satoshi = {
```
pm47: I don't see the point of splitting this method in two.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -423,21 +424,26 @@ object Helpers {
...
       val closingWeight = Transaction.weight(Transactions.addSigs(dummyClosingTx, dummyPublicKey, remoteParams.fundingPubKey, Transactions.PlaceHolderSig, Transactions.PlaceHolderSig).tx)
-      // no need to use a very high fee here, so we target 6 blocks; also, we "MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction"
-      val feeratePerKw = Math.min(Globals.feeratesPerKw.get.blocks_6, commitments.localCommit.spec.feeratePerKw)
       log.info(s"using feeratePerKw=$feeratePerKw for initial closing tx")
       Transactions.weight2fee(feeratePerKw, closingWeight)
     }
 
+    def firstClosingFee(commitments: Commitments, localScriptPubkey: ByteVector, remoteScriptPubkey: ByteVector, feeEstimator: FeeEstimator, feeTargets: FeeTargets)(implicit log: LoggingAdapter): Satoshi = {
+      // we "MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction"
+      val blockTarget = Math.max(feeTargets.mutualCloseBlockTarget, feeTargets.commitmentBlockTarget)
```
pm47: We need to compare feerates, not block targets.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -237,6 +242,22 @@ class PeerSpec extends TestkitBaseClass {
...
+    val probe = TestProbe()
+    system.eventStream.subscribe(probe.ref, classOf[ChannelCreated])
+    connect(remoteNodeId, authenticator, watcher, router, relayer, connection, transport, peer)
+
+    assert(peer.stateData.channels.isEmpty)
+    probe.send(peer, Peer.OpenChannel(remoteNodeId, Satoshi(12300), MilliSatoshi(0), None, None, None))
+    awaitCond(peer.stateData.channels.nonEmpty)
+
+    val channelCreated = probe.expectMsgType[ChannelCreated]
+    assert(channelCreated.initialFeeratePerKw == feeEstimator.getFeeratePerKw(Alice.nodeParams.feeTargets.commitmentBlockTarget))
```
pm47: No need to change the fixture, and also there shouldn't be calls to `Alice.nodeParams` because the peer is only basing its parameters on those.

```scala
val nodeParams = peer.underlyingActor.nodeParams
assert(channelCreated.initialFeeratePerKw == nodeParams.feeEstimator.getFeeratePerKw(nodeParams.feeTargets.commitmentBlockTarget))
assert(channelCreated.fundingTxFeeratePerKw.get == nodeParams.feeEstimator.getFeeratePerKw(nodeParams.feeTargets.fundingBlockTarget))
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -102,7 +105,7 @@ class ClosingStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
 
       awaitCond(alice.stateName == NORMAL)
       awaitCond(bob.stateName == NORMAL)
-      withFixture(test.toNoArgTest(FixtureParam(alice, bob, alice2bob, bob2alice, alice2blockchain, bob2blockchain, relayerA, relayerB, channelUpdateListener, bobCommitTxes)))
+      withFixture(test.toNoArgTest(FixtureParam(alice, bob, alice2bob, bob2alice, alice2blockchain, bob2blockchain, relayerA, relayerB, channelUpdateListener, bobCommitTxes, feeEstimator)))
```
pm47: Same comment as in `PeerSpec`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
araspitzu: We are comparing feerates, this code is the result of adapting:
```
// no need to use a very high fee here, so we target 6 blocks; also, we "MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction"
val feeratePerKw = Math.min(Globals.feeratesPerKw.get.blocks_6, commitments.localCommit.spec.feeratePerKw)
```
into
```
// we "MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction"
val blockTarget = Math.max(feeTargets.mutualCloseBlockTarget, feeTargets.commitmentBlockTarget)
val feeratePerKw = feeEstimator.getFeeratePerKw(blockTarget)
```
It's just a different way to extract the feerate, note that the `Math.min` of the 2 feerates has become `Math.max` of the confirmation target because the higher the target the lower the feerate.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
araspitzu: This was done for testability and used in [ClosingStateSpec](https://github.com/ACINQ/eclair/pull/1083/files#diff-0d859eeb997ad6e4086f3c413770423fR158), i can change the test to use a hardcoded value, it will be harder to read in the future though.

araspitzu:
Latest commit 8856051 makes use of the `feeEstimator` from the underlying actor in all the test

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
araspitzu: Done at 8856051

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
araspitzu: Also done at 8856051

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -310,7 +311,7 @@ object Commitments {
     (commitments1, fee)
   }
 
-  def receiveFee(commitments: Commitments, fee: UpdateFee, maxFeerateMismatch: Double): Commitments = {
+  def receiveFee(commitments: Commitments, fee: UpdateFee, maxFeerateMismatch: Double, feeEstimator: FeeEstimator, feeTargets: FeeTargets): Commitments = {
```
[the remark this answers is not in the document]
araspitzu: Done in c66b3b8

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
```diff
@@ -30,13 +30,33 @@ trait FeeProvider {
 }
 
 // stores fee rate in satoshi/kb (1 kb = 1000 bytes)
-case class FeeratesPerKB(block_1: Long, blocks_2: Long, blocks_6: Long, blocks_12: Long, blocks_36: Long, blocks_72: Long) {
-  require(block_1 > 0 && blocks_2 > 0 && blocks_6 > 0 && blocks_12 > 0 && blocks_36 > 0 && blocks_72 > 0, "all feerates must be strictly greater than 0")
+case class FeeratesPerKB(block_1: Long, blocks_2: Long, blocks_6: Long, blocks_12: Long, blocks_36: Long, blocks_72: Long, blocks_144: Long) {
+  require(block_1 > 0 && blocks_2 > 0 && blocks_6 > 0 && blocks_12 > 0 && blocks_36 > 0 && blocks_72 > 0 && blocks_144 > 0, "all feerates must be strictly greater than 0")
+
+  def feePerBlock(target: Int) = target match {
```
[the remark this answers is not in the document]
araspitzu: Well, at the moment these are just an internal representation of how feerates are stored and we can't access them from anywhere in the code (only the feeEstimator is exposed).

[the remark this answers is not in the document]
sstone: I agree that it would be nice to hide this completely and use `FeeEstimator` instead of `FeeProvider` but that too could be a separate PR ?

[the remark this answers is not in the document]
araspitzu: Yes that sounds like a good candidate for a refactoring-focused PR

araspitzu:
Summary of latest changes:

- All on-chain fee related configurations have been grouped together, both in the configuration file and in `NodeParams` under the `OnChainFeeConf` object.
- There are now helper functions to extract the on-chain fee conf during the test (this avoids lengthy `alice.underlyingActor.nodeParams.feeEstimator`)
- Default confirmation target for claim transactions is now 12 blocks

eclair-core/src/main/resources/reference.conf
```diff
@@ -77,13 +79,23 @@ eclair {
   fee-base-msat = 1000
   fee-proportional-millionths = 100 // fee charged per transferred satoshi in millionths of a satoshi (100 = 0.01%)
 
-  // maximum local vs remote feerate mismatch; 1.0 means 100%
-  // actual check is abs((local feerate - remote fee rate) / (local fee rate + remote fee rate)/2) > fee rate mismatch
-  max-feerate-mismatch = 1.56 // will allow remote fee rates up to 8x bigger or smaller than our local fee rate
+  on-chain-fees {
```
pm47: Not a fan of that name, but can't find a better one. Any idea @sstone ?

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -143,8 +158,12 @@ object TestConstants {
       pingInterval = 30 seconds,
       pingTimeout = 10 seconds,
       pingDisconnect = true,
-      maxFeerateMismatch = 1.0,
-      updateFeeMinDiffRatio = 0.1,
+      onChainFeeConf = OnChainFeeConf(
```
sstone: Not a big issue but I don't like the name, I would call it either `OnChainFeeConfiguration` or  `OnChainFeeSettings`

eclair-core/src/main/resources/reference.conf
```diff
@@ -77,13 +79,23 @@ eclair {
...
+    // number of blocks to target when computing fees for each transaction type
+    fee-targets {
```
```suggestion
    target-blocks {
```

And this should be made consistent with `default-feerates.delay-blocks`. Btw why aren't default feerates under the `on-chain-fees` section?

eclair-core/src/main/resources/reference.conf
```diff
@@ -77,13 +79,23 @@ eclair {
...
+    // maximum local vs remote feerate mismatch; 1.0 means 100%
+    // actual check is abs((local feerate - remote fee rate) / (local fee rate + remote fee rate)/2) > fee rate mismatch
+    max-feerate-mismatch = 1.56 // will allow remote fee rates up to 8x bigger or smaller than our local fee rate
 
-  // funder will send an UpdateFee message if the difference between current commitment fee and actual current network fee is greater
-  // than this ratio.
-  update-fee_min-diff-ratio = 0.1
+    // funder will send an UpdateFee message if the difference between current commitment fee and actual current network fee is greater
+    // than this ratio.
+    update-fee_min-diff-ratio = 0.1
```
pm47: Ew, didn't notice this ugly underscore before!

```suggestion
    update-fee-min-diff-ratio = 0.1
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -165,4 +162,7 @@ trait StateTestsHelperMethods extends TestKitBase {
 
   def channelId(a: TestFSMRef[State, Data, Channel]) = Helpers.getChannelId(a.stateData)
 
+  def feeConfOfChannel(a: TestFSMRef[State, Data, Channel]) = a.underlyingActor.nodeParams.onChainFeeConf
```
pm47: Why not overload the name?

```suggestion
  def feeConf(a: TestFSMRef[State, Data, Channel]) = a.underlyingActor.nodeParams.onChainFeeConf
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -165,4 +162,7 @@ trait StateTestsHelperMethods extends TestKitBase {
...
+  def feeConfOfPeer(a: TestFSMRef[Peer.State, Peer.Data, Peer]) = a.underlyingActor.nodeParams.onChainFeeConf
```
pm47:
```suggestion
  def feeConf(a: TestFSMRef[Peer.State, Peer.Data, Peer]) = a.underlyingActor.nodeParams.onChainFeeConf
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
pm47: And why didn't you make it return a `TestFeeEstimator`?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/g/NegotiatingStateSpec.scala
```diff
@@ -52,7 +52,14 @@ class NegotiatingStateSpec extends TestkitBaseClass with StateTestsHelperMethods
       reachNormal(setup)
       val sender = TestProbe()
       // alice initiates a closing
-      if (test.tags.contains("fee2")) Globals.feeratesPerKw.set(FeeratesPerKw.single(4319)) else Globals.feeratesPerKw.set(FeeratesPerKw.single(10000))
+      if (test.tags.contains("fee2")) {
+        feeConfOfChannel(alice).feeEstimator.asInstanceOf[TestFeeEstimator].setFeerate(FeeratesPerKw.single(4319))
```
sstone: it would be better if we could write `setFeeRate(alice, FeeratesPerKw.single(4319))` and not have to cast `feeEstimator` here

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
sstone: Type erasure will not let you overload method names here, but returning a `TestFeeEstimator` or implemented a `setFee` method here would be cleaner

pm47: Right, WDYT about using the pimp pattern here? `alice.feeEstimator`?

eclair-core/src/main/resources/reference.conf
araspitzu: Right, `default-feerates.delay-blocks` should go under under the `on-chain-fees` section too

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
araspitzu: I returned an `OnChainFeeConf` object because in some test we need both the estimator and the targets, the estimator is much more common though and i can return the `TestFeeEstimator` from the specialized helper method.

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
araspitzu: I tried to be consistent with `RouterConf` but i reckon is not a great name (especially because it recalls "configuration" while it contains an estimator too).

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
pm47: Probably better to follow @sstone's recommendation and add a `setFeerate` method. You can still use `getFeeConf` when you need the targets.

Also the file isn't properly formatted.

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
pm47: Well this is taken from `reference.CONF` so it's not that bad.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/g/NegotiatingStateSpec.scala
araspitzu: 9da4996 uses pimp pattern to make `TestFSMRef` (both for Channel and Peer) return an instance of `TestFeeEstimator` and `FeeTargets`

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
araspitzu: Discarded by 9da4996

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
pm47: @araspitzu has updated with pimp pattern, looks nice IMO

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: I understand what you did, but it doesn't work: there is no guarantee that the resulting fee will be lesser than the one used in the last commitment, because fees may vary widely. We *must* compare to `commitments.localCommit.spec.feeratePerKw`.

araspitzu: That's correct the resulting feerate must be compared with the one used in the commitment transaction and it's not the case currently.

araspitzu: In 419bf60 i used `Math.min(feeEstimator.getFeeratePerKw(blockTarget), commitments.localCommit.spec.feeratePerKw)` to do the proper check.

pm47: Right, but the line above with the max on block targets is misleading. This convey the intent better:

```scala
// we "MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction"
val feeratePerKw = Math.min(requestedFeerate, commitments.localCommit.spec.feeratePerKw)
```

araspitzu: Sounds good to me, but note that with this change if `mutualCloseBlockTarget < commitmentBlockTarget` we are (likely) to make a spec violation, I propose to have a check done in the `NodeParam` to avoid that.

araspitzu: @pm47 Nevermind my comment above, the suggested change is correct.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -65,8 +65,7 @@ case class NodeParams(keyManager: KeyManager,
                       pingInterval: FiniteDuration,
                       pingTimeout: FiniteDuration,
                       pingDisconnect: Boolean,
-                      maxFeerateMismatch: Double,
-                      updateFeeMinDiffRatio: Double,
+                      onChainFeeConf: OnChainFeeConf,
```
pm47: Last nit: `onChainFeeConf` is lost in between peer-specific params, can you move it somewhere where it makes more sense?

araspitzu: Done in 36a8d5e

araspitzu:
Sorry for discarding the review but commit 36a8d5e set the wrong `maxFeerateMismatch` on `TestConstant.Bob`, it has been fixed in 8f1bcc0

PR #1087, opened 2019-07-23
Activate support for variable-length onion

t-bast:
While we have support for variable-length onions at the Sphinx layer, no code path uses it.
This PR moves the feature bit to the configuration and adds support for receiving and relaying variable-length onion payloads.

Note that the current code on master is advertising the optional variable-length onion feature bit regardless of the configuration file's `globalFeatures`. I think this was a mistake on my side.

The PR can easily be reviewed commit by commit: they should be well isolated.

If you are working on things that require variable-length payloads, don't hesitate to rebase on this branch (this is what trampoline and amp are currently doing).


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -219,6 +219,10 @@ object PaymentLifecycle {
   case class UnreadableRemoteFailure(route: Seq[Hop]) extends PaymentFailure
   case class PaymentFailed(id: UUID, paymentHash: ByteVector32, failures: Seq[PaymentFailure]) extends PaymentResult
 
+  sealed trait PaymentOptions
```
t-bast: The reason I introduced that is because it's useful for AMP and Trampoline.
This allows changing the `case object TlvPayload` to a `sealed trait` instead with methods that allow setting additional tlvs inside the hops.

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -17,49 +17,133 @@
...
+        case _ => None
+      }
+  }
+
+}
+
+object OnionPerHopPayload {
+
+  // @formatter:off
+  implicit def legacyPayloadToPerHopPayload(legacy: OnionForwardInfo): OnionPerHopPayload = OnionPerHopPayload(Right(legacy))
```
t-bast: I'm not a big fan of this, do you have a suggestion on how to improve this?

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -17,49 +17,133 @@
...
+    case Left(tlv) =>
+      tlv.records.foldLeft((None, None): (Option[Long], Option[Long])) {
+        case ((_, cltv), AmountToForward(amount)) => (Some(amount), cltv)
+        case ((amount, _), OutgoingCltv(cltv)) => (amount, Some(cltv))
+        case (res, _) => res
+      } match {
+        case (Some(amount), Some(cltv)) => Some(OnionPaymentInfo(amount, cltv))
+        case _ => None
+      }
+  }
```
sstone: Something like:
```scala
      for {
        amount <- tlv.records.collectFirst { case AmountToForward(amountMsat) => amountMsat }
        cltv <- tlv.records.collectFirst { case OutgoingCltv(cltv) => cltv }
      } yield OnionPaymentInfo(amount, cltv)
```
would be easier to read, we go through the records twice but it should not be a performance concern

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -17,49 +17,133 @@
...
+    case Left(tlv) =>
+      tlv.records.foldLeft((None, None, None): (Option[ShortChannelId], Option[Long], Option[Long])) {
+        case ((_, amount, cltv), OutgoingChannelId(channelId)) => (Some(channelId), amount, cltv)
+        case ((channelId, _, cltv), AmountToForward(amount)) => (channelId, Some(amount), cltv)
+        case ((channelId, amount, _), OutgoingCltv(cltv)) => (channelId, amount, Some(cltv))
+        case (res, _) => res
+      } match {
+        case (Some(channelId), Some(amount), Some(cltv)) => Some(OnionForwardInfo(channelId, amount, cltv))
+        case _ => None
+      }
```
sstone: Same here, for comprehensions are probably easier to read

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
sstone: You could override `OnionPerHopPayload.apply()` and use `OnionPerHopPayload(...)` explicitely but it's not really "lean"

t-bast: Yeah it would force us to always use the `OnionPerHopPayload` constructor explicitly which is a bit cumbersome...could I somehow define an extension method `toOnionPerHopPayload` for `OnionForwardInfo` and `TlvStream[OnionTlv]`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -228,41 +232,48 @@ object PaymentLifecycle {
...
     *         - firstExpiry is the cltv expiry for the first htlc in the route
     *         - a sequence of payloads that will be used to build the onion
     */
-  def buildPayloads(finalAmountMsat: Long, finalExpiry: Long, hops: Seq[Hop]): (Long, Long, Seq[PerHopPayload]) =
-    hops.reverse.foldLeft((finalAmountMsat, finalExpiry, PerHopPayload(ShortChannelId(0L), finalAmountMsat, finalExpiry) :: Nil)) {
-      case ((msat, expiry, payloads), hop) =>
-        val nextFee = nodeFee(hop.lastUpdate.feeBaseMsat, hop.lastUpdate.feeProportionalMillionths, msat)
-        (msat + nextFee, expiry + hop.lastUpdate.cltvExpiryDelta, PerHopPayload(hop.lastUpdate.shortChannelId, msat, expiry) +: payloads)
+  def buildPayloads(finalAmountMsat: Long, finalExpiry: Long, hops: Seq[Hop], opts: PaymentOptions = LegacyPayload): (Long, Long, Seq[OnionPerHopPayload]) = {
+    val finalPayload: Seq[OnionPerHopPayload] = opts match {
```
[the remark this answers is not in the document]
t-bast: Because you fold it below to return a `Seq[]` so you need a `Seq[]` from the start right?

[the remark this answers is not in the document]
pm47: Why aren't `finalAmountMsat` and `finalExpiry` provided in some kind of `LegacyPayload` ?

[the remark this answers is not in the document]
t-bast: Can you detail? I don't understand what you mean by that.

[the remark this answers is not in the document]
pm47: Something like (naively):

```scala
sealed trait Payload
case class LegacyPayload(finalAmountMsat: Long, finalExpiry: Long) extends Payload
case class TlvPayload(finalAmountMsat: Long, finalExpiry: Long, tlv: TlvStream) extends Payload
```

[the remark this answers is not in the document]
t-bast: We could do something like that but it loses generality.
This trait doesn't apply only to the last payload, but potentially to every payload.
For now we use the legacy onion format for all intermediate payloads but that's not going to be the case forever. We might need `PaymentOptions` to provide us additional tlvs to inject to intermediate payloads that support variable-length onion (for example if we need to save space on intermediate hops).

Here is how I see it evolving long-term:
```scala
sealed trait PaymentOptions
case object LegacyPayload extends PaymentOptions
sealed trait TlvPayload extends PaymentOptions {
  def finalNodeRecords(): Seq[OnionTlv] = Nil
  def intermediateNodeRecords(): Seq[OnionTlv] = Nil
}
case class MultiPartPayload(totalAmountMsat: Long) extends TlvPayload {
  override def finalNodeRecords() = MultiPartPayment(totalAmountMsat) :: Nil
}
```

or probably something a bit more involved to mix-in stuff for the last payload and stuff for intermediate payloads...

And the `buildOnion` function would use `TlvPayload`'s functions to enrich the "default" payload (amount, expiry, shortChannelId).

But I might be trying to forecast too far in the future, maybe it makes sense to assume that in the short-term we will only want to inject tlv records for the final node and in that case we can indeed add `finalAmountMsat` and `finalExpiry` to those case classes. Let me know!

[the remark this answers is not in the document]
pm47: How about a hierarchy like (again, based on my cursory understanding on the topic):

```scala
sealed trait Payload
sealed trait TlvPayload extends Payload

sealed trait IntermediatePayload
case class IntermediateLegacyPayload(...) extends IntermediatePayload
case class IntermediateTlvPayload(...) extends IntermediatePayload with TlvPayload

sealed trait FinalPayload
case class FinalLegacyPayload(...) extends FinalPayload
case class FinalTlvPayload(...) extends FinalPayload with TlvPayload
```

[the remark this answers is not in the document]
t-bast: I think it's better to forget about the intermediate node cases, it changes the signature of the function too much and we're not sure yet we will need it. Better to do that when we know we need it.
Let's do something like what you propose where this `PaymentOption` only applies to the final node and we can put the amount and expiry in there to simplify the signature of `buildPayloads`.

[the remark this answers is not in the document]
t-bast: I refactored the Payment classes, I think this puts us in a good position for adding new payment features.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -111,7 +112,7 @@ trait Eclair {
 
 class EclairImpl(appKit: Kit) extends Eclair {
 
-  implicit val ec = appKit.system.dispatcher
+  implicit val ec: ExecutionContextExecutor = appKit.system.dispatcher
```
pm47: nit: we really want that to be an `ExecutorContext`

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -219,9 +214,23 @@ object PaymentLifecycle {
...
+  /**
+   * Options to help build the final payload of the payment route.
+   */
+  sealed trait PaymentOptions {
+    // The final htlc amount in millisatoshis.
+    val finalAmountMsat: Long
+    // The final htlc expiry in number of blocks.
+    val finalExpiry: Long
+  }
+  case class LegacyPayload(finalAmountMsat: Long, finalCltvExpiry: Long) extends PaymentOptions {
```
pm47: `finalCltvExpiry` should be `finalCltvExpiryDelta`

t-bast: Yeah I hesitated a lot with that one :)

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: Interesting, I let IntelliJ suggest it, I'll update that.

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
```diff
@@ -97,6 +96,7 @@ object FailureMessageCodecs {
     .typecase(18, ("expiry" | uint32).as[FinalIncorrectCltvExpiry])
     .typecase(19, ("amountMsat" | millisatoshi).as[FinalIncorrectHtlcAmount])
     .typecase(21, provide(ExpiryTooFar))
+    .typecase(PERM, sha256.as[InvalidOnionPayload])
```
t-bast: Since the only thing that was blocking this PR (from the spec point of view) was a decision for that specific error code, I'm currently simply using PERM (which is valid and very generic) which should unblock merging this PR.
We can add the final failure code once the spec stabilizes in another PR.

t-bast: I'm not a big fan of mapping PERM to InvalidOnionPayload bidirectionnally (even if this is temporary).
LND mentioned in the IRC meeting that they currently use InvalidOnionKey for that scenario: maybe we should start by doing that until the error code is correctly spec-ed?

t-bast:
All the follow-up work done on MPP and Trampoline is based on the refactorings and design choices made in this PR, so I'd like this PR to be reviewed and merged to make sure I won't spend too much time in rebase hell ;)

t-bast:
This PR now contains the latest error code from the spec (see https://github.com/lightningnetwork/lightning-rfc/pull/627).
We don't fully provide debug hints because they are hard to extract from scodec.
We do provide the `tag` hint when a tlv field is missing (but it duplicates the tag value in `Onion.scala`: I don't see a cleaner way of doing this without adding too much code).

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -38,18 +35,24 @@ object Features {
...
-  def hasFeature(features: BitSet, bit: Int): Boolean = features.get(bit)
+  // Note that BitVector indexes from left to right whereas the specification indexes from right to left.
+  // This is why we have to reverse the bits to check if a feature is set.
+
+  def hasFeature(features: BitVector, bit: Int): Boolean = if (features.sizeLessThanOrEqual(bit)) false else features.reverse.get(bit)
 
-  def hasFeature(features: ByteVector, bit: Int): Boolean = hasFeature(BitSet.valueOf(features.reverse.toArray), bit)
+  def hasFeature(features: ByteVector, bit: Int): Boolean = hasFeature(features.bits, bit)
+
+  def hasVariableLengthOnion(features: ByteVector): Boolean = hasFeature(features, VARIABLE_LENGTH_ONION_MANDATORY) || hasFeature(features, VARIABLE_LENGTH_ONION_OPTIONAL)
```
pm47: This isn't consistent with what we currently have in the `Peer`:

https://github.com/ACINQ/eclair/blob/71da6776e133e1c4caf18f2c1ac091b48c4f5c8f/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala#L137-L138

https://github.com/ACINQ/eclair/blob/71da6776e133e1c4caf18f2c1ac091b48c4f5c8f/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala#L154

Not sure what's the best approach.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -218,6 +214,18 @@ object PaymentLifecycle {
   case class UnreadableRemoteFailure(route: Seq[Hop]) extends PaymentFailure
   case class PaymentFailed(id: UUID, paymentHash: ByteVector32, failures: Seq[PaymentFailure]) extends PaymentResult
 
+  /**
+   * Options to help build the final payload of the payment route.
+   */
+  sealed trait PaymentOptions {
```
pm47: Can we find a name that better conveys the fact that this targets the final node?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -46,25 +47,22 @@ class PaymentLifecycle(nodeParams: NodeParams, id: UUID, router: ActorRef, regis
...
 
   when(WAITING_FOR_ROUTE) {
     case Event(RouteResponse(hops, ignoreNodes, ignoreChannels), WaitingForRoute(s, c, failures)) =>
       log.info(s"route found: attempt=${failures.size + 1}/${c.maxAttempts} route=${hops.map(_.nextNodeId).mkString("->")} channels=${hops.map(_.lastUpdate.shortChannelId).mkString("->")}")
       val firstHop = hops.head
-      // we add one block in order to not have our htlc fail when a new block has just been found
-      val finalExpiry = (c.finalCltvExpiryDelta + 1).toCltvExpiry
-
-      val (cmd, sharedSecrets) = buildCommand(id, c.amount, finalExpiry, c.paymentHash, hops)
+      val (cmd, sharedSecrets) = buildCommand(id, c.paymentHash, hops, c.paymentOptions)
```
pm47: Where did the `c.finalCltvExpiryDelta + 1` go?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
```diff
@@ -213,8 +218,8 @@ object Relayer extends Logging {
 
   // @formatter:off
   sealed trait NextPayload
-  case class FinalPayload(add: UpdateAddHtlc, payload: PerHopPayload) extends NextPayload
-  case class RelayPayload(add: UpdateAddHtlc, payload: PerHopPayload, nextPacket: OnionRoutingPacket) extends NextPayload {
+  case class FinalPayload(add: UpdateAddHtlc, payload: OnionPerHopPayload) extends NextPayload
```
pm47: The payload isn't consistent between `FinalPayload`/`RelayPayload`. This is due to the choices made in `Onion` and I'll comment further there.

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -18,48 +18,136 @@ package fr.acinq.eclair.wire
...
-case class PerHopPayload(shortChannelId: ShortChannelId,
-                         amtToForward: MilliSatoshi,
-                         outgoingCltvValue: CltvExpiry)
+case class OnionForwardInfo(shortChannelId: ShortChannelId,
+                            amtToForward: MilliSatoshi,
+                            outgoingCltvValue: CltvExpiry)
+
+case class OnionPaymentInfo(amount: MilliSatoshi, cltvExpiry: CltvExpiry)
+
+case class OnionPerHopPayload(payload: Either[TlvStream[OnionTlv], OnionForwardInfo]) {
```
pm47: Have you considered using a proper trait hierarchy?

```scala
sealed trait Payload

sealed trait LegacyPayload extends Payload
sealed trait TlvPayload extends Payload

sealed trait RelayPayload extends Payload
sealed trait FinalPayload extends Payload

case class RelayLegacy(...) extends LegacyPayload with RelayPayload
case class FinalLegacy(...) extends LegacyPayload with FinalPayload
case class RelayTlv(...) extends TlvPayload with RelayPayload
case class FinalTlv(...) extends TlvPayload with FinalPayload
```

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: Strictly speaking it would be more correct to handle `variable_length_onion_optional` and `variable_length_onion_mandatory` separately, when mandatory is set we should reject legacy payloads.
I don't think anyone wants to set mandatory at the moment though, I don't know if we should correctly implement the distinction between the two.
But you're right that separating the two checks makes it more explicit that we're not handling the distinction yet, I'll change that.

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
t-bast: This is more verbose, but you're right that it will likely make the code more explicit and allow us to separate legacy and tlv more cleanly. I'll investigate that.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
t-bast: Good catch, it got lost in the subsequent refactor / rebase steps!

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
t-bast: This is much better, thanks for the suggestion.
We are really going to have different code-paths for tlv vs legacy and relay vs final, so it really makes sense to separate them explicitly at the type level.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
```diff
@@ -17,25 +17,48 @@
...
 
   override def receive: Receive = {
-    case c: GenericSendPayment =>
+    case p: PaymentInitiator.SendPaymentRequest =>
       val paymentId = UUID.randomUUID()
+      // We add one block in order to not have our htlc fail when a new block has just been found.
+      val finalExpiry = (p.finalExpiryDelta + 1).toCltvExpiry
       val payFsm = context.actorOf(PaymentLifecycle.props(nodeParams, paymentId, router, register))
-      payFsm forward c
+      p.predefinedRoute match {
```
pm47: I would add a comment here to remove all ambiguity:

```scala
// NB: we only generate legacy payload for now
```

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -38,18 +35,29 @@ object Features {
...
     * Check that the features that we understand are correctly specified, and that there are no mandatory features that
-    * we don't understand (even bits)
+    * we don't understand (even bits).
     */
-  def areSupported(bitset: BitSet): Boolean = {
-    val supportedMandatoryFeatures = Set(OPTION_DATA_LOSS_PROTECT_MANDATORY)
-    for (i <- 0 until bitset.length() by 2) {
-      if (bitset.get(i) && !supportedMandatoryFeatures.contains(i)) return false
+  def areSupported(features: BitVector): Boolean = {
+    val supportedMandatoryFeatures = Set[Long](OPTION_DATA_LOSS_PROTECT_MANDATORY)
```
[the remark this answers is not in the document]
t-bast: This is currently independent because variable-length onion is a global feature, and this method is only used to check against local features.
However https://github.com/lightningnetwork/lightning-rfc/pull/666 will flatten those so it probably makes sense to update this `areSupported` method.

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -18,48 +18,160 @@ package fr.acinq.eclair.wire
...
+    case Left(tlvs) => Attempt.successful(RelayTlvPayload(tlvs))
+    case Right(legacy) => Attempt.successful(legacy)
+  }, {
+    case legacy: RelayLegacyPayload => Right(legacy)
+    case RelayTlvPayload(tlvs) => Left(tlvs)
+  })
+
+  val finalPerHopPayloadCodec: Codec[FinalPayload] = fallback(tlvPerHopPayloadCodec, legacyFinalPerHopPayloadCodec).narrow({
+    case Left(tlvs) if tlvs.get[AmountToForward].isEmpty => Attempt.failure(MissingRequiredTlv(UInt64(2)))
+    case Left(tlvs) if tlvs.get[OutgoingCltv].isEmpty => Attempt.failure(MissingRequiredTlv(UInt64(4)))
```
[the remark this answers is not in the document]
t-bast: Yes but it's a sender restriction. I think it's harmless as a receiver to accept and ignore it instead of failing the payment, that means we would be rejecting payments that are otherwise valid and are giving us money :D

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
```diff
@@ -213,58 +215,58 @@ object Relayer extends Logging {
...
+            perHopPayload match {
+              case finalPayload: Onion.FinalPayload => Right(FinalPayload(add, finalPayload))
+              case relayPayload: Onion.RelayPayload => Right(RelayPayload(add, relayPayload, nextPacket))
             }
-          case Attempt.Failure(_) =>
-            // Onion is correctly encrypted but the content of the per-hop payload couldn't be parsed.
-            Left(InvalidOnionPayload(Sphinx.PaymentPacket.hash(add.onionRoutingPacket)))
+          case Attempt.Failure(e: OnionCodecs.MissingRequiredTlv) => Left(e.failureMessage)
+          // Onion is correctly encrypted but the content of the per-hop payload couldn't be parsed.
+          // It's hard to provide tag and offset information from scodec failures, so we currently don't do it.
```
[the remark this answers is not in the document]
t-bast: Yes but that's covered by the previous case.
If we get to that case we don't have such information, and we can't provide an offset either.

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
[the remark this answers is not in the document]
pm47: Fair enough

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -18,48 +18,160 @@ package fr.acinq.eclair.wire
...
+  /** Legacy fixed-size 65-bytes onion payload. */
+  sealed trait LegacyPayload extends PerHopPayload
+
+  /** Variable-length onion payload with optional additional tlv records. */
+  sealed trait TlvPayload extends PerHopPayload {
+    val records: TlvStream[OnionTlv]
+  }
+
+  /** Per-hop payload for an intermediate node. */
+  sealed trait RelayPayload extends PerHopPayload {
```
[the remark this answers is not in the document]
t-bast: I think the ambiguity was that Legacy/Tlv is a description of the encoding, whereas Relay/Final is a description of the content.
I separated those in two distinct traits, I hope it's cleaner. It still allows easy extension for trampoline.

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -49,19 +49,21 @@ object Onion {
 
   import OnionTlv._
 
-  /** Per-hop payload from an HTLC's payment onion (after decryption and decoding). */
-  sealed trait PerHopPayload
+  sealed trait PerHopPayloadEncoding
 
   /** Legacy fixed-size 65-bytes onion payload. */
-  sealed trait LegacyPayload extends PerHopPayload
+  sealed trait LegacyPayload extends PerHopPayloadEncoding
```
pm47: `PerHopPayloadEncoding` -> `PerHopPayloadFormat`

`LegacyPayload` -> `LegacyFormat`

eclair-core/src/test/scala/fr/acinq/eclair/FeaturesSpec.scala
```diff
@@ -35,8 +35,8 @@ class FeaturesSpec extends FunSuite {
     assert(hasFeature(hex"02", Features.OPTION_DATA_LOSS_PROTECT_OPTIONAL))
   }
 
-  test("'initial_routing_sync' and 'data_loss_protect' feature") {
-    val features = hex"0a"
+  test("'initial_routing_sync', 'data_loss_protect' and 'variable_length_onion' features") {
+    val features = hex"010a"
```
pm47: This test seems unfinished

t-bast:
:tada: :tada:

PR #1088, opened 2019-07-25
Typed amounts

araspitzu:
This PR aims at typing all the currency amounts used in eclair, this will provide type-checker safety and help spot potential bugs regarding operation between different currency units (i.e comparing Satoshis to MilliSatoshis), the feature makes use of the new `eclair.MilliSatoshi` class that implements comparison and arithmetic operation only with itself. 

The class fields (or method parameters) that used `Long` to represent a currency amount have been typed and renamed (i.e amountMsat -> amount or fundingSatoshis -> fundingAmount), most of the comparison have been replaced by typed comparisons.

Protocol messages (in `wire` package) have been typed but not renamed, same for the http APIs, the form params parsing have been typed but not renamed to avoid breaking API changes.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -150,12 +150,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       forwarder ! remote
       val open = OpenChannel(nodeParams.chainHash,
         temporaryChannelId = temporaryChannelId,
-        fundingSatoshis = fundingSatoshis,
-        pushMsat = pushMsat,
-        dustLimitSatoshis = localParams.dustLimitSatoshis,
+        fundingSatoshis = Satoshi(fundingSatoshis),
```
t-bast: It's going to be a bit painful, but as we correctly type these amounts maybe we should rename the variables, they don't need to explicitly name the unit anymore?
`fundingSatoshis -> fundingAmount`
`pushMsat -> pushAmount`
`dustLimitSatoshis -> dustLimit`
etc?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -289,10 +289,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
             firstPerCommitmentPoint = keyManager.commitmentPoint(localParams.channelKeyPath, 0))
           val remoteParams = RemoteParams(
             nodeId = remoteNodeId,
-            dustLimitSatoshis = open.dustLimitSatoshis,
+            dustLimitSatoshis = open.dustLimitSatoshis.toLong,
```
t-bast: Why are the amount not typed in `RemoteParams`? I think we should push those types everywhere we have an amount, don't we?

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
```diff
@@ -52,9 +52,14 @@ object CommonCodecs {
   // this codec can be safely used for values < 2^63 and will fail otherwise
   // (for something smarter see https://github.com/yzernik/bitcoin-scodec/blob/master/src/main/scala/io/github/yzernik/bitcoinscodec/structures/UInt64.scala)
   val uint64overflow: Codec[Long] = int64.narrow(l => if (l >= 0) Attempt.Successful(l) else Attempt.failure(Err(s"overflow for value $l")), l => l)
-
   val uint64: Codec[UInt64] = bytes(8).xmap(b => UInt64(b), a => a.toByteVector.padLeft(8))
 
+  val uint64overflowSat: Codec[Satoshi] = uint64overflow.xmapc(l => Satoshi(l))(_.toLong)
+  val uint64overflowMSat: Codec[MilliSatoshi] = uint64overflow.xmapc(l => MilliSatoshi(l))(_.toLong)
```
t-bast: nit: `Msat`

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
```diff
@@ -52,9 +52,14 @@ object CommonCodecs {
...
+
+  val uint64Sat: Codec[Satoshi] = uint64.xmapc(l => Satoshi(l.toBigInt.longValue()))(sat => UInt64(sat.toLong))
+  val uint64MSat: Codec[MilliSatoshi] = uint64.xmapc(l => MilliSatoshi(l.toBigInt.longValue()))(msat => UInt64(msat.toLong))
```
t-bast: nit: `Msat`

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
```diff
@@ -52,9 +52,14 @@ object CommonCodecs {
   // this codec can be safely used for values < 2^63 and will fail otherwise
   // (for something smarter see https://github.com/yzernik/bitcoin-scodec/blob/master/src/main/scala/io/github/yzernik/bitcoinscodec/structures/UInt64.scala)
   val uint64overflow: Codec[Long] = int64.narrow(l => if (l >= 0) Attempt.Successful(l) else Attempt.failure(Err(s"overflow for value $l")), l => l)
-
   val uint64: Codec[UInt64] = bytes(8).xmap(b => UInt64(b), a => a.toByteVector.padLeft(8))
 
+  val uint64overflowSat: Codec[Satoshi] = uint64overflow.xmapc(l => Satoshi(l))(_.toLong)
+  val uint64overflowMSat: Codec[MilliSatoshi] = uint64overflow.xmapc(l => MilliSatoshi(l))(_.toLong)
+
+  val uint64Sat: Codec[Satoshi] = uint64.xmapc(l => Satoshi(l.toBigInt.longValue()))(sat => UInt64(sat.toLong))
```
t-bast: Do we even need this one? It seems like `Long` is definitely enough for all amounts, even for amounts in msat a `Long` should be able to contain all the Bitcoins ever created...
So we should only use the `overflow` version because a value that overflows is likely not a valid amount, right?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
```diff
@@ -114,29 +114,29 @@ class WaitForAcceptChannelStateSpec extends TestkitBaseClass with StateTestsHelp
   test("recv AcceptChannel (reserve below dust limit)") { f =>
     import f._
     val accept = bob2alice.expectMsgType[AcceptChannel]
-    val reserveTooSmall = accept.dustLimitSatoshis - 1
+    val reserveTooSmall = accept.dustLimitSatoshis.toLong - 1
```
t-bast: Isn't `-` defined for `Satoshi` and `Millisatoshi`?
It seems to me that we could (should) get rid of most calls to `toLong` and leverage supported operators and implicit conversions...

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
araspitzu: I thought about that but we have to consider those objects map messages from the spec and there is value in keeping exactly the same name. Will definitely rename variables that are not part of the spec messages.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
araspitzu: Sure, i'm starting from the `wire` messages (LN messages) and the PR is in draft, as the changes reach more part of the code it will be upgraded to typed amounts.

araspitzu:
> This is going to be helpful to have the compiler let us check when we're using satoshis instead of millisatoshis :)
> I think we can go further and push those typed amounts almost everywhere, unless it's really painful to use when developing...what do you think?

Agreed that it will be very helpful to have all the amounts typed :), the plan is to have types _everywhere_ and i'm starting from the wire messages, not sure if i'll break it down into smaller PRs (after the wire msg).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -146,7 +146,7 @@ object Commitments {
     // the HTLC we are about to create is outgoing, but from their point of view it is incoming
     val outgoingHtlcs = reduced.htlcs.filter(_.direction == IN)
 
-    val htlcValueInFlight = UInt64(outgoingHtlcs.map(_.add.amountMsat).sum)
+    val htlcValueInFlight = UInt64(outgoingHtlcs.map(_.add.amountMsat.toLong).sum)
```
sstone: We should be able to write `outgoingHtlcs.map(_.add.amountMsat).sum` here (all that is needed is an implementation of `Numeric[MilliSatoshi]`, it's missing in bitcoin-lib but can be added here for now)

araspitzu: How about adding `Numeric[MilliSatoshi]` to https://github.com/ACINQ/bitcoin-lib/pull/36 ? Then i can build this PR on top of the upgraded bitcoin-lib.

araspitzu:
> I think that the goal of this PR should be to eliminate all explicit call to `.toLong` as well as implicit conversions as much as possible. If there's too much friction and changes in `bitcoin-lib` are needed then it should also be documented. I'm not sure it would be that useful to have one PR for using stricter types in wire messages types and another one for our internal types.

I started with the main goal of typing the amounts, while doing so i incurred in comparison between typed/non-typed quantities so i used `.toLong` to preserve the comparison result (note that currrently when comparing `MilliSatoshi` with `Satoshi` we lose precision). I had the idea of breaking down the changes in different PRs mostly for ease of review but if that's not a problem i can keep adding changes here.

araspitzu:
The lastest commit b3d4a03 replaces the `MilliSatoshi` from bitcoin-lib with an equivalent implementation under the package `eclair`, the rationale for this is that bitcoin-lib's `BtcAmount` should deal only with bitcoin amounts and that does not include the millisatoshis. The new `MilliSatoshi` class implements comparison operators only with itself and offers a conversion to `Satoshi`, also all `BtcAmount` can be converted to `MilliSatoshi` via [pimp pattern](https://github.com/ACINQ/eclair/pull/1088/files#diff-e77b1bb47acf28a7f3afd7c9aca35f79R184). With this change all the comparison between `MilliSatoshi` and `BtcAmount` have to be explicitly handled (and depend on the context), i.e when computing transaction fees [here](https://github.com/ACINQ/eclair/pull/1088/files#diff-aeb98076f262543b70a86753bd3719adR223) we use `Satoshi`'s numeric ops and comparison but when necessary we can use msat-to-msat comparison, [here](https://github.com/ACINQ/eclair/pull/1088/files#diff-5fa783866be9932536bd1825c7998eb7R95).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -289,10 +289,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
             maxHtlcValueInFlightMsat = open.maxHtlcValueInFlightMsat,
-            channelReserveSatoshis = open.channelReserveSatoshis, // remote requires local to keep this much satoshis as direct payment
-            htlcMinimumMsat = open.htlcMinimumMsat,
+            channelReserveSatoshis = open.channelReserveSatoshis.toLong, // remote requires local to keep this much satoshis as direct payment
```
sstone: this is supposed to be temporary, right ?

araspitzu: Absolutely! It's temporary until `LocalParams` and `RemoteParams` are migrated to the typed amounts.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -339,7 +339,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           val localFundingPubkey = keyManager.fundingPublicKey(localParams.channelKeyPath).publicKey
           val fundingPubkeyScript = Script.write(Script.pay2wsh(Scripts.multiSig2of2(localFundingPubkey, remoteParams.fundingPubKey)))
           wallet.makeFundingTx(fundingPubkeyScript, Satoshi(fundingSatoshis), fundingTxFeeratePerKw).pipeTo(self)
-          goto(WAIT_FOR_FUNDING_INTERNAL) using DATA_WAIT_FOR_FUNDING_INTERNAL(temporaryChannelId, localParams, remoteParams, fundingSatoshis, pushMsat, initialFeeratePerKw, accept.firstPerCommitmentPoint, open)
+          goto(WAIT_FOR_FUNDING_INTERNAL) using DATA_WAIT_FOR_FUNDING_INTERNAL(temporaryChannelId, localParams, remoteParams, Satoshi(fundingSatoshis), pushMsat, initialFeeratePerKw, accept.firstPerCommitmentPoint, open)
```
sstone: Same here, fundingSatoshis is supposed to be typed ?

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommonCodecs.scala
sstone: Yes Long should be enough: we have 21*1e6 btc, which is 21*1e6*1e8 = 21*1e14 sat, which is 21*1e17 msat which we can still represent with a Long.
And I'd rather have the codecs named satoshi and millisatoshi ?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
araspitzu: It will be typed but that change shouldn't be here **now**. To clarify i applied the following pattern when typing the amounts in `wire` messages:
- When doing a comparison between typed a non typed amounts use raw data types ( Long vs Long)
- Changes should applied only to `wire` messages, other classes using the same fields (i.e `LocalParams` have remained unchanged.
- Classes that use the new typed amounts have been migrated to typed only if all the fields are now typed , [example](https://github.com/ACINQ/eclair/pull/1088/files#diff-b90a51e9dd7869ff814b488a4271daad)
- Functions that mimic constructor also pulled the change from the underlying type [example](https://github.com/ACINQ/eclair/pull/1088/files#diff-d57afc529bb6dd213ae17d552113061dR40)

So the change you commented shouldn't be there, `DATA_WAIT_FOR_FUNDING_INTERNAL.fundingSatoshis` should stay `Long` for the moment. Alternatively i can propagate the typing (and diverge from the PR-wide heuristics) to this class and type "pushMsat" as well.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -137,9 +136,9 @@ object NodeParams {
       case _ => BITCOIND
     }
 
-    val dustLimitSatoshis = config.getLong("dust-limit-satoshis")
+    val dustLimitSatoshis = Satoshi(config.getLong("dust-limit-satoshis"))
     if (chainHash == Block.LivenetGenesisBlock.hash) {
-      require(dustLimitSatoshis >= Channel.MIN_DUSTLIMIT, s"dust limit must be greater than ${Channel.MIN_DUSTLIMIT}")
+      require(dustLimitSatoshis >= Satoshi(Channel.MIN_DUSTLIMIT), s"dust limit must be greater than ${Channel.MIN_DUSTLIMIT}")
```
sstone: Channel.MIN_DUSTLIMIT should be typed

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
sstone: INPUT_INIT_FUNDER sat/msat fields should be typed, same will all command messages

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -159,4 +159,55 @@ package object eclair {
...
+    def -(other: MilliSatoshi) = MilliSatoshi(amount - other.amount)
+    def *(m: Long) = MilliSatoshi(amount * m)
+    def /(d: Long) = MilliSatoshi(amount / d)
+    def compare(other: MilliSatoshi): Int = if (amount == other.amount) 0 else if (amount < other.amount) -1 else 1
+    def <= (that: MilliSatoshi): Boolean = compare(that) <= 0
+    def >= (that: MilliSatoshi): Boolean = compare(that) >= 0
+    def <  (that: MilliSatoshi): Boolean = compare(that) < 0
+    def >  (that: MilliSatoshi): Boolean = compare(that) > 0
+    def unary_-() = MilliSatoshi(-amount)
+    def toSatoshi: Satoshi = Satoshi(amount / 1000)
```
sstone: conversion here is not lossless, this should be reflected in the method's name, I would suggest `truncateToSatoshi`

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -218,8 +219,8 @@ object Transactions {
     val fee = weight2fee(feeratePerKw, htlcTimeoutWeight)
     val redeemScript = htlcOffered(localHtlcPubkey, remoteHtlcPubkey, localRevocationPubkey, ripemd160(htlc.paymentHash.bytes))
     val pubkeyScript = write(pay2wsh(redeemScript))
-    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(Satoshi(htlc.amountMsat / 1000)))
-    val amount = MilliSatoshi(htlc.amountMsat) - fee
+    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(Satoshi(htlc.amountMsat.toLong / 1000)))
```
sstone: this should use msat to sat conversion

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -266,7 +267,7 @@ object Transactions {
   def makeClaimHtlcSuccessTx(commitTx: Transaction, outputsAlreadyUsed: Set[Int], localDustLimit: Satoshi, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, remoteRevocationPubkey: PublicKey, localFinalScriptPubKey: ByteVector, htlc: UpdateAddHtlc, feeratePerKw: Long): ClaimHtlcSuccessTx = {
     val redeemScript = htlcOffered(remoteHtlcPubkey, localHtlcPubkey, remoteRevocationPubkey, ripemd160(htlc.paymentHash.bytes))
     val pubkeyScript = write(pay2wsh(redeemScript))
-    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(Satoshi(htlc.amountMsat / 1000)))
+    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(Satoshi(htlc.amountMsat.toLong / 1000)))
```
sstone: this should use msat to sat conversion

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -289,7 +290,7 @@ object Transactions {
   def makeClaimHtlcTimeoutTx(commitTx: Transaction, outputsAlreadyUsed: Set[Int], localDustLimit: Satoshi, localHtlcPubkey: PublicKey, remoteHtlcPubkey: PublicKey, remoteRevocationPubkey: PublicKey, localFinalScriptPubKey: ByteVector, htlc: UpdateAddHtlc, feeratePerKw: Long): ClaimHtlcTimeoutTx = {
     val redeemScript = htlcReceived(remoteHtlcPubkey, localHtlcPubkey, remoteRevocationPubkey, ripemd160(htlc.paymentHash.bytes), htlc.cltvExpiry)
     val pubkeyScript = write(pay2wsh(redeemScript))
-    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(Satoshi(htlc.amountMsat / 1000)))
+    val outputIndex = findPubKeyScriptIndex(commitTx, pubkeyScript, outputsAlreadyUsed, amount_opt = Some(Satoshi(htlc.amountMsat.toLong / 1000)))
```
sstone: this should use msat to sat conversion

eclair-core/src/test/scala/fr/acinq/eclair/payment/RelayerSpec.scala
```diff
@@ -63,7 +63,7 @@ class RelayerSpec extends TestkitBaseClass {
     // we use this to build a valid onion
     val (cmd, _) = buildCommand(UUID.randomUUID(), finalAmountMsat, finalExpiry, paymentHash, hops)
     // and then manually build an htlc
-    val add_ab = UpdateAddHtlc(channelId = channelId_ab, id = 123456, cmd.amountMsat, cmd.paymentHash, cmd.cltvExpiry, cmd.onion)
+    val add_ab = UpdateAddHtlc(channelId = channelId_ab, id = 123456, MilliSatoshi(cmd.amountMsat), cmd.paymentHash, cmd.cltvExpiry, cmd.onion)
```
sstone: Same comment as above, this should go once `amountMsat` is typed

eclair-core/src/main/scala/fr/acinq/eclair/payment/Autoprobe.scala
```diff
@@ -54,7 +54,7 @@ class Autoprobe(nodeParams: NodeParams, router: ActorRef, paymentInitiator: Acto
         case Some(targetNodeId) =>
           val paymentHash = randomBytes32 // we don't even know the preimage (this needs to be a secure random!)
           log.info(s"sending payment probe to node=$targetNodeId payment_hash=$paymentHash")
-          paymentInitiator ! SendPayment(PAYMENT_AMOUNT_MSAT, paymentHash, targetNodeId, maxAttempts = 1)
+          paymentInitiator ! SendPayment(MilliSatoshi(PAYMENT_AMOUNT_MSAT), paymentHash, targetNodeId, maxAttempts = 1)
```
sstone: PAYMENT_AMOUNT_MSAT should be typed

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -869,14 +869,14 @@ object Router {
     val currentBlockHeight = Globals.blockCount.get()
 
     val boundaries: RichWeight => Boolean = { weight =>
-      ((weight.cost - amountMsat) < routeParams.maxFeeBaseMsat || (weight.cost - amountMsat) < (routeParams.maxFeePct * amountMsat)) &&
+      ((weight.cost - amount) < routeParams.maxFeeBase || (weight.cost - amount) < MilliSatoshi(amount.toLong * routeParams.maxFeePct.toLong)) &&
```
sstone: Do we still need to `.toLong` calls here ?

araspitzu: Not really but `maxFeePct` is a `Double` and can't be multiplied with satoshis as-is, however there is a cleaner way to do this:
```
((weight.cost - amount) < routeParams.maxFeeBase || (weight.cost - amount) < amount * routeParams.maxFeePct.toLong) &&
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -16,7 +16,7 @@
 
 package fr.acinq.eclair.blockchain.bitcoind.rpc
 
-import fr.acinq.bitcoin._
+import fr.acinq.bitcoin.{Block, ByteVector32, Transaction}
```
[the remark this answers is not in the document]
araspitzu: It was a leftover from the transition to `eclair.MilliSatoshi`, they're gone in 7b72dca

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
araspitzu: Fixed in c3eae80

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
araspitzu: Agreed and i like the name, implemented in bda3655

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -159,4 +159,54 @@ package object eclair {
...
+    override def toInt(x: MilliSatoshi): Int = x.toLong.toInt
+    override def toLong(x: MilliSatoshi): Long = x.toLong
+    override def toFloat(x: MilliSatoshi): Float = x.toLong
+    override def toDouble(x: MilliSatoshi): Double = x.toLong
+    override def compare(x: MilliSatoshi, y: MilliSatoshi): Int = x.compare(y)
+  }
+
+  private def satoshi2millisatoshi(input: Satoshi): MilliSatoshi = MilliSatoshi(input.amount * 1000L)
+
+  def maxOf(x: MilliSatoshi, y: MilliSatoshi) = MilliSatoshi(Math.max(x.amount, y.amount))
```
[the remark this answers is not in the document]
araspitzu: Somehow i find more explicit `Math.max` since using the standard(java) library functions enables the reader to assume the implementation is correct, it's not a big deal to change it though.

PR #1089, opened 2019-07-25
Publish transactions during transitions

pm47:
Follow up to #1082.

The goal is to be able to publish transactions only after we have
persisted the state. Otherwise we may run into corner cases like [1]
where a refund tx has been published, but we haven't kept track of it
and generate a different one (with different fees) the next time.

As a side effect, we can now remove the special case that we were
doing when publishing the funding tx, and remove the `store` function.

NB: the new `calling` transition method isn't restricted to publishing
transactions but that is the only use case for now.

[1] https://github.com/ACINQ/eclair-mobile/issues/206


pm47:
NB: I haven't been consistent between the syntax `calling doPublish(closingTx)` (recommended by intellij) or `calling(doPublish(closingTx))`. WDYT @sstone?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1210,25 +1212,15 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-          }
-          val nextRemoteCommitPublished1 = d.nextRemoteCommitPublished.map {
-            remoteCommitPublished =>
-              val remoteCommitPublished1 = Helpers.Closing.claimRemoteCommitTxOutputs(keyManager, commitments1, commitments1.remoteCommit, remoteCommitPublished.commitTx)
-              doPublish(remoteCommitPublished1)
-              remoteCommitPublished1
+          val localCommitPublished1 = d.localCommitPublished.map(localCommitPublished => Helpers.Closing.claimCurrentLocalCommitTxOutputs(keyManager, commitments1, localCommitPublished.commitTx))
+          val remoteCommitPublished1 = d.remoteCommitPublished.map(remoteCommitPublished => Helpers.Closing.claimRemoteCommitTxOutputs(keyManager, commitments1, commitments1.remoteCommit, remoteCommitPublished.commitTx))
+          val nextRemoteCommitPublished1 = d.nextRemoteCommitPublished.map(remoteCommitPublished => Helpers.Closing.claimRemoteCommitTxOutputs(keyManager, commitments1, commitments1.remoteCommit, remoteCommitPublished.commitTx))
+          def republish(): Unit = {
```
sstone: nit: there should be a new line before `def republish()`

pm47:
> I think that it behaves differently in case of error: currently if `doPublish` fails for example we won't transition to the next state or trigger transition handlers, but here I think we would

That's a good point, I haven't tested but I think the transition would fail. Anyway even in previous code there was an implicit (bold?) assumption that these `doPublish` calls never fail.

PR #1090, opened 2019-07-26
Use correct cost comparison when evaluating candidate channels

araspitzu:
This PR fixes bug https://github.com/ACINQ/eclair-mobile/issues/207 which was caused by an incorrect channel cost evaluation against its `htlcMaximumMsat`. The fix is to consider the correct channel cost.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -88,7 +88,8 @@ object Graph {
     val candidates = new mutable.PriorityQueue[WeightedPath]
 
     // find the shortest path, k = 0
-    val shortestPath = dijkstraShortestPath(graph, sourceNode, targetNode, amountMsat, ignoredEdges, extraEdges, RichWeight(amountMsat, 0, 0, 0), boundaries, currentBlockHeight, wr)
+    val initialWeight = RichWeight(cost = amountMsat, 0, 0, 0)
+    val shortestPath = dijkstraShortestPath(graph, sourceNode, targetNode, ignoredEdges, extraEdges, initialWeight, boundaries, currentBlockHeight, wr)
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -230,8 +230,8 @@ object Graph {
           val newMinimumKnownWeight = edgeWeight(edge, currentWeight, initialWeight.length == 0 && neighbor == sourceNode, currentBlockHeight, wr)
 
           // test for ignored edges
-          if (edge.update.htlcMaximumMsat.forall(newMinimumKnownWeight.cost + amountMsat <= _) &&
-            newMinimumKnownWeight.cost + amountMsat >= edge.update.htlcMinimumMsat &&
+          if (edge.update.htlcMaximumMsat.forall(newMinimumKnownWeight.cost <= _) &&
```
sstone: If `amountMsat` is not used here then it's not used at all in `dijkstraShortestPath` and should be removed from the parameter list. We then need to make sure that https://github.com/ACINQ/eclair/blob/352e4d00cd20c3a1e589d6392c71a1b43298ef48/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala#L127 is correct

araspitzu: With this change `dijkstraShortestPath` will use only the cost found in the `initialWeight` parameter which is `RichWeight(amountMsat, 0, 0, 0)` for the initial step and when calculating the 'spurPath' for an alternative path this is the result of `pathWeight` so it's the cost to go from the starting node to this intermediate node (from here the algorithm will search for a different path). I will remove `amountMsat` from the parameter list.

araspitzu: f62f472 removes the unused parameter and also extracts the `initialWeight` parameter in the first invocation of `dijkstraShortestPath`, for more readability.

PR #1092, opened 2019-07-31
Documentation update

t-bast:
* Clean-up the readme formatting warnings
* Update the OpenJDK 11 link everywhere
* Added a section linking to our wiki and API doc
* Most of the wiki pages have been updated prior to this PR

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,39 +1,57 @@
 # Building Eclair
 
 ## Requirements
-- [OpenJDK 11](https://jdk.java.net/11/).
+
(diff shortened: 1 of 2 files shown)


t-bast:
I think we should create new wiki pages to use a guides for common scenario.
For example we could have a wiki page with examples on how to use the `sendtoroute` command, or once we have MPP a page on how to use MPP will be useful.
Checking payment status via the CLI could be documented a bit more too.

araspitzu:
Nice to see a 'Documentation' section, how about moving it closer to the top for quicker inspection by newcomers?

PR #1093, opened 2019-08-01
Derive channel keys from the channel funding pubkey

sstone:
This PR changes how we compute channel keys: 
- Each new channel is assigned a random BIP32 "funding key" path, which can be used to compute its funding privkey/pubkey pair.
- All channel secrets/private keys are derived from a BIP32 path that can be deterministically computed from the channel's funding pubkey.

User funds can now be recovered from their seed, without using backups (we could say that "we use the bitcoin blockchain to backup channels"). 
If a user loses their channel backup but still has their BIP32 seed, they can use DLP to request peers they have active channels with to publish their current commitment transaction, extract their funding pubkey from its witness, re-compute their `to-remote` key and spend their output.

Compatibility with older versions of eclair is achieved by adding a `version` field to `LocalParams` using the same pattern as for channel versions (since legacy  `LocalParams` also starts with a public key). It has been tested with the following scenarios:
- new nodes can create and use channels with legacy nodes
- legacy nodes can be upgraded. Channels created before the upgrade will have their `LocalParam.version` field set to 0, and new channels will have their `LocalParam.version` field set to 1.

For users, it means that this feature will only be available for new channels, they still need their backup data to recover legacy channels (note: for recovering channels with peers that support DLP, any backup will do and not just the last one)


SomberNight:
> User funds can now be recovered from their seed, without using backups (we could say that "we use the bitcoin blockchain to backup channels").
If a user loses their channel backup but still has their BIP32 seed, they can use DLP to request peers they have active channels with to publish their current commitment transaction, extract their funding pubkey from its witness, re-compute their to-remote key and spend their output.

To be clear, you assume that the user performing disaster-recovery has
- their BIP32 seed
- knows which peer (e.g. 33 byte node id) they had what channel (e.g. 8 byte short channel id) with
- the node is online and reachable
- the channel had DLP enabled

right?

sstone:
Not necessarily, just knowing your seed helps a lot: If commitment transactions have already been published, all you need to to is scan the blockchain for txs that spend a 2-of-2 p2wsh multisig, extract pubkeys, and check if you can spend its to-remote output. This may be enough in many cases, as your peers will eventually close channels that are never used anymore and where they have funds locked in, even if they don't support DLP.

Also, if you're funded channels it's very easy to re-compute their channel ids, just look at transactions coming out of your bitcoin wallet. If channels are public you can then find which node you were connected to

And in most cases users open channels to very few nodes and should be able to guess which ones

PR #1096, opened 2019-08-08
Use MilliSatoshi to type 'maxHtlcValueInFlightMsat'

araspitzu:
Use a specialized scodec to encode/decode 64 bit unsigned integers into scala's `Long`, then we use MilliSatoshi (backed by `Long`) to type 'maxHtlcValueInFlightMsat' and when is necessary to compare it against other MilliSatoshis we use an unsigned comparison.


eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -75,7 +75,7 @@ object ChannelCodecs extends Logging {
   val remoteParamsCodec: Codec[RemoteParams] = (
     ("nodeId" | publicKey) ::
       ("dustLimit" | satoshi) ::
-      ("maxHtlcValueInFlightMsat" | uint64) ::
+      ("maxHtlcValueInFlightMsat" | millisatoshi64) ::
```
sstone: we should be able to use the `millisatoshi` codec here, since 21 million btc = 2100000000000000000 msat = 1d24b2dfac520000 msat, we don't need BigInt (but need to check potential impact on serialisation and storage)

araspitzu: Okay then in the deserialization we would need truncate excess values (encoded in the uint64) to 2100000000000000000 msat

sstone: Ah yes I see the pb now :( what happens if it is set to 0xFFFFFFFFFFFFFFFF ... I don't like truncating because with end up with `decode(encode(x)) != x`

araspitzu:
9dfea99 uses unsigned Long comparison and parsing to store an uint64 and convert it to MilliSatoshi, however we still target java 7 so we need to write the functions to compare/convert unsigned longs.

PR #1097, opened 2019-08-09
Derive channel keys from the channel funding pubkey

sstone:
This PR changes how we compute channel keys: 
- Each new channel is assigned a random BIP32 "funding key" path, which can be used to compute its funding privkey/pubkey pair.
- All channel secrets/private keys are derived from a BIP32 path that can be deterministically computed from the channel's funding pubkey.

User funds can now be recovered from their seed, without using backups (we could say that "we use the bitcoin blockchain to backup channels"). 
If a user loses their channel backup but still has their BIP32 seed, they can use DLP to request peers they have active channels with to publish their current commitment transaction, extract their funding pubkey from its witness, re-compute their `to-remote` key and spend their output.

Compatibility with older versions of eclair is achieved by adding a `version` field to `LocalParams` using the same pattern as for channel versions (since legacy  `LocalParams` also starts with a public key). It has been tested with the following scenarios:
- new nodes can create and use channels with legacy nodes
- legacy nodes can be upgraded. Channels created before the upgrade will have their `LocalParam.version` field set to 0, and new channels will have their `LocalParam.version` field set to 1.

For users, it means that this feature will only be available for new channels, they still need their backup data to recover legacy channels (note: for recovering channels with peers that support DLP, any backup will do and not just the last one)

Node: this is #1093 resurrected, it got corrupted when I forced-push changes :(


sstone:
As pointed out in the old PR, this is not just useful when your peers support DLP (Data Loss Protect) and you can still connect to them and use DLP to get them to publish their commit tx.

If you've lost everything (except for you seed of course) you will still be able to scan the blockchain for commit tx that you can spend, which is good because it's very likely that your peers will eventually close their channels with you if they're never used.

It's also very easy to retrieve the ids of channels you've funded just by looking at your onchain wallet transactions. 

And in most cases users should be able to remember/guess which nodes they opened a channel to.

SomberNight:
> If you've lost everything (except for you seed of course) you will still be able to scan the blockchain for commit tx that you can spend, which is good because it's very likely that your peers will eventually close their channels with you if they're never used.

If the remote closed with their commitment transaction, and you notice while scanning the blockchain, you probably won't have the per commitment secret to derive the `to_remote` key. Or am I missing something? Are you assuming `static_remotekey` or some future feature?

sstone:
You're right it was wishful thinking (I want `toRemoteKey` based on the "remote" point so much that I keep forgetting it's not the case). 

This change will also help with `static_remotekey`, especially if we keep the CSV delay (this is still an open point) but as is, it's better than what we currently have.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -630,17 +630,23 @@ object Peer {
...
     secureRandom.nextBytes(entropy)
     val bis = new ByteArrayInputStream(entropy)
-    val channelKeyPath = DeterministicWallet.KeyPath(Seq(Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), Protocol.uint32(bis, ByteOrder.BIG_ENDIAN)))
-    makeChannelParams(nodeParams, defaultFinalScriptPubKey, isFunder, fundingAmount, channelKeyPath)
+    // we make sure that funder and fundee key path end differently
+    val last = isFunder match {
+      case false => DeterministicWallet.hardened(0)
+      case true => DeterministicWallet.hardened(1)
+    }
+    val fundingKeyPath = DeterministicWallet.KeyPath(Seq(Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), last))
```
t-bast: Why do we reduce the depth of the funding key path compared to before?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/KeyManager.scala
```diff
@@ -73,5 +76,36 @@ trait KeyManager {
...
+    */
+  def signChannelAnnouncement(fundingKeyPath: DeterministicWallet.KeyPath, chainHash: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, remoteFundingKey: PublicKey, features: ByteVector): (ByteVector64, ByteVector64)
+
+  def signChannelAnnouncement(fundingKey: DeterministicWallet.ExtendedPublicKey, chainHash: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, remoteFundingKey: PublicKey, features: ByteVector): (ByteVector64, ByteVector64)
+  = signChannelAnnouncement(fundingKey.path, chainHash, shortChannelId, remoteNodeId, remoteFundingKey, features)
 }
+
+object KeyManager {
+  /**
+    * Create a BIP32 path from a public key. This path will be used to derive channel keys.
```
t-bast: It might be worth explaining in details why we derive channel keys from the funding key (what it brings for recovery when being funder, and that it doesn't help fundee that much).

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
sstone: What we want to avoid here is reusing keys (either funding keys, or channel keys that are derived from them) and I think 8 bytes (i.e. 2^64 keys) is enough.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/KeyManager.scala
sstone: Yers, I'll update the comment, and we'll add (with another PR) a recovery tool that will allow users to get their funds back if all they have is their seed, the id of a node they had a channel to, and the channel id (which could be inferred from wallet transactions for "funder" channels).

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -641,17 +641,23 @@ object Peer {
...
 
   def makeChannelParams(nodeParams: NodeParams, defaultFinalScriptPubKey: ByteVector, isFunder: Boolean, fundingAmount: Satoshi): LocalParams = {
-    val entropy = new Array[Byte](16)
+    val entropy = new Array[Byte](8)
     secureRandom.nextBytes(entropy)
     val bis = new ByteArrayInputStream(entropy)
-    val channelKeyPath = DeterministicWallet.KeyPath(Seq(Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), Protocol.uint32(bis, ByteOrder.BIG_ENDIAN)))
-    makeChannelParams(nodeParams, defaultFinalScriptPubKey, isFunder, fundingAmount, channelKeyPath)
+    // we make sure that funder and fundee key path end differently
+    val last = isFunder match {
```
pm47: How about:

```scala
  val last = DeterministicWallet.hardened(if (isFunder) 1 else 0)
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -641,17 +641,23 @@ object Peer {
...
+      case false => DeterministicWallet.hardened(0)
+      case true => DeterministicWallet.hardened(1)
+    }
+    val fundingKeyPath = DeterministicWallet.KeyPath(Seq(Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), Protocol.uint32(bis, ByteOrder.BIG_ENDIAN), last))
```
pm47: How about:

```scala
    val path = Seq(secureRandom.nextLong(), secureRandom.nextLong(), last)
    val fundingKeyPath = DeterministicWallet.KeyPath(path)
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -159,12 +161,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         feeratePerKw = initialFeeratePerKw,
         toSelfDelay = localParams.toSelfDelay,
         maxAcceptedHtlcs = localParams.maxAcceptedHtlcs,
-        fundingPubkey = keyManager.fundingPublicKey(localParams.channelKeyPath).publicKey,
-        revocationBasepoint = keyManager.revocationPoint(localParams.channelKeyPath).publicKey,
-        paymentBasepoint = keyManager.paymentPoint(localParams.channelKeyPath).publicKey,
-        delayedPaymentBasepoint = keyManager.delayedPaymentPoint(localParams.channelKeyPath).publicKey,
-        htlcBasepoint = keyManager.htlcPoint(localParams.channelKeyPath).publicKey,
-        firstPerCommitmentPoint = keyManager.commitmentPoint(localParams.channelKeyPath, 0),
+        fundingPubkey = fundingPubKey.publicKey,
```
pm47: I don't see the point of not creating `fundingPubkey` inline?

```suggestion
        fundingPubkey = keyManager.fundingPublicKey(fundingKeyPath),
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -281,12 +285,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
             htlcMinimumMsat = localParams.htlcMinimum,
             toSelfDelay = localParams.toSelfDelay,
             maxAcceptedHtlcs = localParams.maxAcceptedHtlcs,
-            fundingPubkey = keyManager.fundingPublicKey(localParams.channelKeyPath).publicKey,
-            revocationBasepoint = keyManager.revocationPoint(localParams.channelKeyPath).publicKey,
-            paymentBasepoint = keyManager.paymentPoint(localParams.channelKeyPath).publicKey,
-            delayedPaymentBasepoint = keyManager.delayedPaymentPoint(localParams.channelKeyPath).publicKey,
-            htlcBasepoint = keyManager.htlcPoint(localParams.channelKeyPath).publicKey,
-            firstPerCommitmentPoint = keyManager.commitmentPoint(localParams.channelKeyPath, 0))
+            fundingPubkey = fundingPubKey.publicKey,
```
```suggestion
            fundingPubkey = keyManager.fundingPublicKey(fundingKeyPath),
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -200,7 +203,20 @@ final case class LocalParams(nodeId: PublicKey,
                              isFunder: Boolean,
                              defaultFinalScriptPubKey: ByteVector,
                              globalFeatures: ByteVector,
-                             localFeatures: ByteVector)
+                             localFeatures: ByteVector) {
+
+  def fundingPubKey(keyManager: KeyManager) = keyManager.fundingPublicKey(fundingKeyPath)
```
pm47: This doesn't seem necessary and breaks consistency with other pubkey generators.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -189,8 +191,9 @@ final case class DATA_CLOSING(commitments: Commitments,
 
 final case class DATA_WAIT_FOR_REMOTE_PUBLISH_FUTURE_COMMITMENT(commitments: Commitments, remoteChannelReestablish: ChannelReestablish) extends Data with HasCommitments
 
-final case class LocalParams(nodeId: PublicKey,
-                             channelKeyPath: DeterministicWallet.KeyPath,
+final case class LocalParams(version: Int,
```
pm47: I'm not sure about adding a version here instead of using a `ChannelVersion` for this.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -200,7 +203,20 @@ final case class LocalParams(nodeId: PublicKey,
...
+
+  def fundingPubKey(keyManager: KeyManager) = keyManager.fundingPublicKey(fundingKeyPath)
+
+  def channelKeyPath(keyManager: KeyManager) = version match {
+    case 0 =>
+      // legacy mode:  we reuse the funding key path as our channel key path
+      fundingKeyPath
+    case 1 =>
+      // deterministic mode: use the funding pubkey to compute the channel key path
+      KeyManager.channelKeyPath(fundingPubKey(keyManager))
```
araspitzu: Would it be possible to store the channel key path on a lazy val to avoid doing the `hash160` each time is used?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -148,6 +148,8 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     case Event(initFunder@INPUT_INIT_FUNDER(temporaryChannelId, fundingSatoshis, pushMsat, initialFeeratePerKw, fundingTxFeeratePerKw, localParams, remote, _, channelFlags, _), Nothing) =>
       context.system.eventStream.publish(ChannelCreated(self, context.parent, remoteNodeId, isFunder = true, temporaryChannelId, initialFeeratePerKw, Some(fundingTxFeeratePerKw)))
       forwarder ! remote
+      val fundingPubKey = keyManager.fundingPublicKey(localParams.fundingKeyPath).publicKey
```
pm47: This line needs to be removed.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -524,7 +529,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         case Success(_) =>
           log.info(s"channelId=${commitments.channelId} was confirmed at blockHeight=$blockHeight txIndex=$txIndex")
           blockchain ! WatchLost(self, commitments.commitInput.outPoint.txid, nodeParams.minDepthBlocks, BITCOIN_FUNDING_LOST)
-          val nextPerCommitmentPoint = keyManager.commitmentPoint(commitments.localParams.channelKeyPath, 1)
+          val fundingPubKey = keyManager.fundingPublicKey(d.commitments.localParams.fundingKeyPath).publicKey
```
pm47: Line needs to be removed.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1399,7 +1407,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       forwarder ! r
 
       val yourLastPerCommitmentSecret = d.commitments.remotePerCommitmentSecrets.lastIndex.flatMap(d.commitments.remotePerCommitmentSecrets.getHash).getOrElse(ByteVector32.Zeroes)
-      val myCurrentPerCommitmentPoint = keyManager.commitmentPoint(d.commitments.localParams.channelKeyPath, d.commitments.localCommit.index)
+      val fundingPubKey = keyManager.fundingPublicKey(d.commitments.localParams.fundingKeyPath)
```
pm47: Line needs to be removed.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/KeyManager.scala
```diff
@@ -73,5 +76,40 @@ trait KeyManager {
...
+    * @param shortChannelId short channel id
+    * @param remoteNodeId remote node id
+    * @param remoteFundingKey remote funding pubkey
+    * @param features channel features
+    * @return a (nodeSig, bitcoinSig) pair. nodeSig is the signature of the channel announcement with our node's
+    *         private key, bitcoinSig is the signature of the channel announcement with our funding private key
+    */
+  def signChannelAnnouncement(fundingKeyPath: DeterministicWallet.KeyPath, chainHash: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, remoteFundingKey: PublicKey, features: ByteVector): (ByteVector64, ByteVector64)
+
+  def signChannelAnnouncement(fundingKey: DeterministicWallet.ExtendedPublicKey, chainHash: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, remoteFundingKey: PublicKey, features: ByteVector): (ByteVector64, ByteVector64)
```
pm47: I don't see the point of this additional method since it's easier to just do `signChannelAnnouncement(commitments.localParams.fundingKeyPath, ...)`

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -224,9 +224,19 @@ object ChannelFlags {
...
 }
+
 object ChannelVersion {
+  import scodec.bits._
   val LENGTH_BITS = 4 * 8
-  val STANDARD = ChannelVersion(BitVector.fill(LENGTH_BITS)(false))
+  val ZEROES = ChannelVersion(bin"00000000000000000000000000000000")
+  val STANDARD = ZEROES
+  val USE_PUBKEY_KEYPATH_BIT = 0 // bit numbers start at 0
+  val USE_PUBKEY_KEYPATH = STANDARD | ChannelVersion(bin"00000000000000000000000000000001")
```
pm47: Can't we use `USE_PUBKEY_KEYPATH_BIT` here?

PR #1100, opened 2019-08-19
Add API call to get general network statistics

n1bor:
allchannels is now so slow it timesout.

This API call returns general stats on the network that is useful to know.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -38,7 +38,19 @@ import fr.acinq.eclair.payment.{GetUsableBalances, PaymentReceived, PaymentRelay
 import fr.acinq.eclair.wire.{ChannelAnnouncement, ChannelUpdate, NodeAddress, NodeAnnouncement}
 import TimestampQueryFilters._
 
+
 case class GetInfoResponse(nodeId: PublicKey, alias: String, chainHash: ByteVector32, blockHeight: Int, publicAddresses: Seq[NodeAddress])
+
+case class FlagCounter(data : Map[Int,Int] = Map(0->0, 1->0, 2->0, 3->0, 4->0, 5->0, 6->0, 7->0)) {
+  def +(that: Byte) = {
+    FlagCounter((0 to 7).map{ b =>
+      (b,data.getOrElse(b,0) + ((that & (1<<b)) >> b).toInt)
+    }.toMap)
+  }
+}
+
+case class GetNetworkInfoResponse(totalChannelCount: Long, totalNodes: Long, totalUpdates:Long, avgCltvExpiryDelta: Long, avgHtlcMinimumMsat: Long,
+                           avgFeeBaseMsat: Long, avgFeeProportionalMillionths: Long, avgHtlcMaximumMsat: Long, messageFlags: FlagCounter, channelFlags: FlagCounter )
 
 case class AuditResponse(sent: Seq[PaymentSent], received: Seq[PaymentReceived], relayed: Seq[PaymentRelayed])
 
@@ -105,6 +117,8 @@ trait Eclair {
   def allUpdates(nodeId_opt: Option[PublicKey])(implicit timeout: Timeout): Future[Iterable[ChannelUpdate]]
 
   def getInfoResponse()(implicit timeout: Timeout): Future[GetInfoResponse]
+
+  def getNetworkInfoResponse()(implicit timeout: Timeout): Future[GetNetworkInfoResponse]
 
   def usableBalances()(implicit timeout: Timeout): Future[Iterable[UsableBalances]]
 }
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -177,7 +191,30 @@ class EclairImpl(appKit: Kit) extends Eclair {
...
 
+  override def getNetworkInfoResponse()(implicit timeout: Timeout): Future[GetNetworkInfoResponse] = {
+    for {
+      nodeCount <- allNodes().map{_.size}
+      channelCount <- allChannels().map{_.size}
+
+      (updateCount: Long,avgCltvExpiryDelta: Long,avgHtlcMinimumMsat: Long,avgFeeBaseMsat: Long,avgFeeProportionalMillionths: Long,avgHtlcMaximumMsat: Long, mf: FlagCounter, cf:FlagCounter ) <- allUpdates(None).map{
+        updates => updates.foldLeft(Tuple8(0L, 0L, 0L, 0L, 0L, 0L,FlagCounter(), FlagCounter())) {
+          (t, c) =>
+            Tuple8(t._1 + 1,
```
t-bast: Instead of this hard-to-read `Tuple8` why don't you directly use a `GetNetworkInfoResponse` here?
I think it would be cleaner.

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -155,6 +158,46 @@ class EclairImplSpec extends TestKit(ActorSystem("mySystem")) with fixture.FunSu
       }
     })
   }
+import scala.collection.mutable.ListBuffer
```
t-bast: Can you please order imports and run the formatter?
If you're using IntelliJ, `ctrl+alt+o` and `ctrl+alt+l`.

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -155,6 +158,46 @@ class EclairImplSpec extends TestKit(ActorSystem("mySystem")) with fixture.FunSu
...
+  test("Get network Info works") {f =>
```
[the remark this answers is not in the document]
t-bast: This test is quite hard to follow, could you clean it up?
Something like arrange-act-assert pattern with meaningful names for values we assert on would be easier to read.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
n1bor: Not sure this is possible. As GetNetworkInfoResponse include results of the earlier futures too.
Could fill with dummy zeros but think that would make more confusing.
Especially if additional calls/futures were added to get more data.

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -155,6 +158,46 @@ class EclairImplSpec extends TestKit(ActorSystem("mySystem")) with fixture.FunSu
...
+
+    val ann_a = makeNodeAnnouncement(randomKey, "node-A", Color(15, 10, -70), Nil)
+    val eclair = new EclairImpl(kit)
+    val fResp = eclair.getNetworkInfoResponse() // ask updates filtered by 'b'
+    f.router.expectMsg('nodes)
+    f.router.reply(Seq(ann_a))
+    f.router.expectMsg('channels)
+    val channelId_ab = ShortChannelId(420000, 1, 0)
+    val channelId_bc = ShortChannelId(420000, 2, 0)
+    val chan_ab = fr.acinq.eclair.router.BaseRouterSpec.channelAnnouncement(channelId_ab, randomKey, randomKey, randomKey, randomKey)
```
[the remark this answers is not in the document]
n1bor: Am in EclairImpSpec so do need it. Wanted to make clear was coming from another Spec.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: I would do something like (I slightly renamed the fields, you'll have to adapt):

```
override def getNetworkInfoResponse()(implicit timeout: Timeout): Future[GetNetworkInfoResponse] = {
    for {
      nodeCount <- allNodes().map(_.size)
      channelCount <- allChannels().map(_.size)
      channelUpdates <- allUpdates(None)
      val networkInfo = channelUpdates.foldLeft(GetNetworkInfoResponse(channelCount, nodeCount, channelUpdates.size, 0, MilliSatoshi(0), MilliSatoshi(0), 0, MilliSatoshi(0))) {
        (current, next) => current.copy(
          avgCltvExpiryDelta = current.avgCltvExpiryDelta + next.cltvExpiryDelta,
          avgFeeBaseMsat = current.avgFeeBaseMsat + next.feeBaseMsat
          // etc
        )
      }
    } yield networkInfo.copy(
      avgFeeBaseMsat = networkInfo.avgFeeBaseMsat / networkInfo.updates
      // etc
    )
  }
```

And I would refactor to make dividing all `avg` fields by the `updateCount` be done in a method in the case class or something like that.

t-bast:
@n1bor I merged #1116 which might be useful to you: you could now simply expose that via the API. Would that be enough information for your use-case?

PR #1101, opened 2019-08-20
Route computation: fix fee check

sstone:
Fee check during route computation is:
- fee is below maximum value
- OR fee is below amout * maximum percentage

The second check was buggy and route computation would fail when fees we above maximum value but below maximum percentage of amount being paid.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -867,7 +867,7 @@ object Router {
 
     val currentBlockHeight = Globals.blockCount.get()
 
-    def feeBaseOk(fee: MilliSatoshi): Boolean = fee < routeParams.maxFeeBase
+    def feeBaseOk(fee: MilliSatoshi): Boolean = fee <= routeParams.maxFeeBase
 
     def feePctOk(fee: MilliSatoshi, amount: MilliSatoshi): Boolean = {
       val maxFee = amount * routeParams.maxFeePct
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -871,7 +871,7 @@ object Router {
 
     def feePctOk(fee: MilliSatoshi, amount: MilliSatoshi): Boolean = {
       val maxFee = amount * routeParams.maxFeePct
-      fee < maxFee
+      fee <= maxFee
     }
 
     def feeOk(fee: MilliSatoshi, amount: MilliSatoshi): Boolean = feeBaseOk(fee) || feePctOk(fee, amount)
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -868,10 +867,21 @@ object Router {
 
     val currentBlockHeight = Globals.blockCount.get()
 
+    def feeBaseOk(fee: MilliSatoshi): Boolean = fee < routeParams.maxFeeBase
+
+    def feePctOk(fee: MilliSatoshi, amount: MilliSatoshi): Boolean = {
+      val maxFee = amount * routeParams.maxFeePct
+      fee < maxFee
```
```suggestion
      fee <= maxFee
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -868,10 +867,21 @@ object Router {
 
     val currentBlockHeight = Globals.blockCount.get()
 
+    def feeBaseOk(fee: MilliSatoshi): Boolean = fee < routeParams.maxFeeBase
```
```suggestion
    def feeBaseOk(fee: MilliSatoshi): Boolean = fee <= routeParams.maxFeeBase
```

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -56,6 +56,30 @@ class RouteCalculationSpec extends FunSuite {
...
 
+  test("check fee against max pct properly") {
+
+    // fee is acceptable is it is either
+    // - below our maximum fee base
+    // - below our maximum fraction of the paid amount
+
+    // here we here a maximum fee base of 1 msat, and all our updates have a base fee of 10 msat
+    // so our fee will always be above the base fee, and we will always check that is is below our maximum percentage
+    // of the amount being paid
```
t-bast: nit: a few typos in the comment: should probably be:
```
here we have a maximum fee base of 1 msat, and all our updates have a base fee of 10 msat
so our fee will always be above the base fee, and we will always check that it is below our maximum percentage
of the amount being paid
```

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -56,6 +56,30 @@ class RouteCalculationSpec extends FunSuite {
...
+    val updates = List(
+      makeUpdate(1L, a, b, MilliSatoshi(10), 10, cltvDelta = 1),
+      makeUpdate(2L, b, c, MilliSatoshi(10), 10, cltvDelta = 1),
+      makeUpdate(3L, c, d, MilliSatoshi(10), 10, cltvDelta = 1),
+      makeUpdate(4L, d, e, MilliSatoshi(10), 10, cltvDelta = 1)
+    ).toMap
+
+    val g = makeGraph(updates)
+
+    val route = Router.findRoute(g, a, e, DEFAULT_AMOUNT_MSAT, numRoutes = 1, routeParams = DEFAULT_ROUTE_PARAMS.copy(maxFeeBase = MilliSatoshi(1)))
```
t-bast: I don't see any test verifying that:
* Fee percent are ok but fee base is not -> no route found
* Fee base are ok but fee percent is not -> no route found

Do we have that somewhere? Or would it be useful to add?

sstone: It's because the actual test is fee base below threshold OR fee percent below threshold so in both cases we should find a route. The bug was that if fee base was above threshold but fee percent was acceptable we would still fail to find a route.

PR #1102, opened 2019-08-21
Move http APIs to subproject eclair-node

araspitzu:
This PR moves the `Service` implementing eclair's HTTP API to the subproject eclair-node, it will produce a smaller eclair-core jar and simplify the merging of the android branch. Note that json serialization is still part of eclair-core as it is used for testing and to provide channel debug information.


eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -39,7 +43,29 @@ object Boot extends App with Logging {
     val setup = new Setup(datadir)
     plugins.foreach(_.onSetup(setup))
     setup.bootstrap onComplete {
-      case Success(kit) => plugins.foreach(_.onKit(kit))
+      case Success(kit) =>
+        plugins.foreach(_.onKit(kit))
```
t-bast: I would initialize the plugins after the API is up (imo initializing external things like plugins should always come last)

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -39,7 +43,29 @@ object Boot extends App with Logging {
...
+        val config = NodeParams.loadConfiguration(datadir, ConfigFactory.empty())
```
t-bast: It's a bit weird to have to reload the configuration, could we had the relevant parts to `Kit` instead?
Maybe that is tied to a bigger config refactoring and should be done later though.

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -145,11 +141,12 @@ class ApiServiceSpec extends FunSuite with ScalatestRouteTest with IdiomaticMock
 
   test("'usablebalances' asks router for current usable balances") {
 
+    val remoteNodeId = PublicKey(hex"03af0ed6052cf28d670665549bc86f4b721c9fdb309d40c58f5811f63966e005d0")
```
t-bast: nit: this is `aliceNodeId` right?

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
araspitzu: Agreed, in commit 3469eb5 i don't reload the config anymore but directly access the field on the Setup class.

araspitzu: Agreed, in commit 3469eb5 i don't reload the config anymore but directly access the field on the Setup class.

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
araspitzu: Good catch thanks :) , fixed in 3469eb5 (for some reason github is not showing the last commit here on the PR)

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
araspitzu: I agree that plugins initialization should happen last but to be fair our plugin initialization is already interleaved with the bootstrap,  we first call `onSetup` then `onKit`.

t-bast:
LGTM

sstone:
Have you tested that it is still possible to enable the API on eclair-node-gui ?

araspitzu:
@sstone I didn't...and it wasn't working, commit e917f5b8 allows to enable the API when running the GUi as well (also GUI in headless mode)

sstone:
Is it also possible to move `JsonSerializers` (including `akka-http-json4s`) to eclair-node ?

araspitzu:
@sstone we can't get rid of json serizliation in eclair-core because of [this](https://github.com/ACINQ/eclair/blob/api_to_eclair-node/eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala#L299) test that uses the custom defined serializers. An alternative would be to keep `json4s` in eclair-core but move `akka-http-json4s` on eclair-node, i've tried that but there were issues, i will give it another shot.

sstone:
Yes we still need `json4s` (we also use it to talk to bitcoind) but not `akka-http-json4s`, and the compatibility test is very limited and could be improved (in another PR).  If would suggest that for now we move JsonSerializers to eclair-node, and add just what we need in the compatibility test to make it work.

PR #1103, opened 2019-08-21
Add more numeric utilities to MilliSatoshi

t-bast:
It's quite cumbersome to manipulate MilliSatoshis right now, everything needs to be wrapped in `MilliSatoshi()`.
While I understand that we don't want an implicit conversion from `Long` to `MilliSatoshi`, arithmetic operations and comparisons with `BtcAmount`s would make sense.


araspitzu:
I think those are valuable changes even though we still can't compare `MilliSatoshi`s to `Long`, an example appication of this would be [here](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala#L211). I think it's very important to not create confusion on the result type when composing operations and i like that in your changes all the operations result in `MilliSatoshi`. To provide postfix ops conversions like "123 msat" you don't need implicit conversions but you can define an implicit wrapper around `Long` like this one:
```
  implicit class LongToMsat(l: Long) {
    def msat: MilliSatoshi = MilliSatoshi(l)
  }
  val x = 124 msat // x is of type MiiliSatoshi
```

t-bast:
> To provide postfix ops conversions like "123 msat" you don't need implicit conversions but you can define an implicit wrapper around Long like this one

That's very cool, thanks for the tip! With that I think I should remove the numeric operations that use `Long` and instead have our code explicitly use the `msat` or `sat` postfix when using numeric values, that will be better. I'll update that right now.

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -156,32 +154,45 @@ package object eclair {
...
+   */
   val MaxEpochSeconds = Duration.fromNanos(Long.MaxValue).toSeconds
 
   /**
-    * One MilliSatoshi is a thousand of a Satoshi, the smallest unit usable in bitcoin
-    * @param amount
-    */
+   * One MilliSatoshi is a thousand of a Satoshi, the smallest unit usable in bitcoin
+   */
   case class MilliSatoshi(amount: Long) {
```
sstone: `amount` should be a `private val`

t-bast:
It looks like the default formatter in IntelliJ slightly changed whitespace rules.
I suggest toggling the "ignore whitespace changes" when reviewing ;)

eclair-core/src/test/java/fr/acinq/eclair/MilliSatoshiTest.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package fr.acinq.eclair;
+
+import fr.acinq.bitcoin.Satoshi;
+
+/**
+ * This class is a compile-time check that we are able to compile Java code that uses MilliSatoshi utilities.
+ */
+public final class MilliSatoshiTest {
```
t-bast: I think that if we want real Java tests we'll need to add JUnit support (akka is using JUnit for their Java tests)

eclair-core/src/main/scala/fr/acinq/eclair/CoinUtils.scala
```diff
@@ -17,14 +17,16 @@
 package fr.acinq.eclair
 
 import java.text.{DecimalFormat, NumberFormat}
+
```
araspitzu: nit: unnecessary blank lines?

eclair-core/src/main/scala/fr/acinq/eclair/MilliSatoshi.scala
```diff
@@ -0,0 +1,76 @@
...
+  def -(other: BtcAmount) = MilliSatoshi(underlying - other.toMilliSatoshi.underlying)
+  def *(m: Long) = MilliSatoshi(underlying * m)
+  def *(m: Double) = MilliSatoshi((underlying * m).toLong)
+  def /(d: Long) = MilliSatoshi(underlying / d)
+  def unary_-() = MilliSatoshi(-underlying)
+  override def compare(other: MilliSatoshi): Int = if (underlying == other.underlying) 0 else if (underlying < other.underlying) -1 else 1
+  // Since BtcAmount is a sealed trait that MilliSatoshi cannot extend, we need to redefine comparison operators.
+  def compare(other: BtcAmount): Int = {
+    val otherMsat = other.toMilliSatoshi.underlying
+    if (underlying == otherMsat) 0 else if (underlying < otherMsat) -1 else 1
```
araspitzu: Not a strong opinion but i would try to reuse the code as much as possible. For example here you could reuse the `compare(other: MilliSatoshi)` implementation instead of re-writing again the comparison logic, even the method above could use the `compareTo` offered by the java api.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -398,7 +398,7 @@ class Router(nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Prom
 
     case Event(FinalizeRoute(partialHops), d) =>
       // split into sublists [(a,b),(b,c), ...] then get the edges between each of those pairs, then select the largest edge between them
-      val edges = partialHops.sliding(2).map { case List(v1, v2) => d.graph.getEdgesBetween(v1, v2).maxBy(_.update.htlcMaximumMsat) }
+      val edges = partialHops.sliding(2).map { case List(v1, v2) => d.graph.getEdgesBetween(v1, v2).maxBy(_.update.htlcMaximumMsat.getOrElse(0 msat)) }
```
araspitzu: I think this is unnecessary since `maxBy` can work with options too, for example this works in the REPL:
```
scala> case class Foo(elem: Option[Int])
defined class Foo

scala> List(Foo(Some(1)), Foo(Some(2)), Foo(None))
res2: List[Foo] = List(Foo(Some(1)), Foo(Some(2)), Foo(None))

scala> res2.maxBy(_.elem)
res3: Foo = Foo(Some(2))

scala> res2.minBy(_.elem)
res4: Foo = Foo(None)
```
However the `maxBy` operation can't be performed on empty lists and if this happens this like will throw an exception (this happened in the wild already).

t-bast: It doesn't compile if I remove it because of evil implicits saying it's ambiguous to choose the right `Ordering`...

eclair-core/src/main/scala/fr/acinq/eclair/CoinUtils.scala
t-bast: It's the default import formatting (`ctrl+alt+o`).
One paragraph for java stdlib imports, then one paragraph for libraries imports, then one for scala stdlib imports.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitcoinCoreFeeProvider.scala
```diff
@@ -64,16 +64,16 @@ object BitcoinCoreFeeProvider {
...
             // negative value means failure
             feerate.toLong
           case JInt(feerate) =>
             // should (hopefully) never happen
-            btc2satoshi(Btc(feerate.toLong)).amount
+            btc2satoshi(Btc(feerate.toLong)).toLong
         }
       case JArray(errors) =>
-        val error = errors collect { case JString(error) => error } mkString (", ")
+        val error = errors collect { case JString(error) => error } mkString ", "
```
araspitzu: nit: IIRC as a convention we use `.` as much as possible so this would become:
`val error = errors.collect { case JString(error) => error }.mkString(", ")`

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -607,9 +606,9 @@ object Peer {
   case object ResumeAnnouncements
   case class OpenChannel(remoteNodeId: PublicKey, fundingSatoshis: Satoshi, pushMsat: MilliSatoshi, fundingTxFeeratePerKw_opt: Option[Long], channelFlags: Option[Byte], timeout_opt: Option[Timeout]) {
     require(fundingSatoshis < Channel.MAX_FUNDING, s"fundingSatoshis must be less than ${Channel.MAX_FUNDING}")
-    require(pushMsat.amount <= 1000 * fundingSatoshis.amount, s"pushMsat must be less or equal to fundingSatoshis")
-    require(fundingSatoshis.amount >= 0, s"fundingSatoshis must be positive")
-    require(pushMsat.amount >= 0, s"pushMsat must be positive")
+    require(pushMsat <= fundingSatoshis, s"pushMsat must be less or equal to fundingSatoshis")
+    require(fundingSatoshis.toLong >= 0, s"fundingSatoshis must be positive")
```
araspitzu: How about using a typed comparison like: `fundingSatoshis >= 0.sat`?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -643,7 +642,7 @@ object Peer {
       channelKeyPath,
       dustLimit = nodeParams.dustLimit,
       maxHtlcValueInFlightMsat = nodeParams.maxHtlcValueInFlightMsat,
-      channelReserve = maxOf(Satoshi((nodeParams.reserveToFundingRatio * fundingAmount.toLong).toLong), nodeParams.dustLimit), // BOLT #2: make sure that our reserve is above our dust limit
+      channelReserve = Satoshi((nodeParams.reserveToFundingRatio * fundingAmount.toLong).toLong).max(nodeParams.dustLimit), // BOLT #2: make sure that our reserve is above our dust limit
```
araspitzu: This could be made more readable using:
```suggestion
      (fundingAmount.toMilliSatoshi * nodeParams.reserveToFundingRatio).max(nodeParams.dustLimit), // BOLT #2: make sure that our reserve is above our dust limit
```

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/GUIUpdater.scala
```diff
@@ -86,7 +86,7 @@ class GUIUpdater(mainController: MainController) extends Actor with ActorLogging
       val (channelPaneController, root) = createChannelPanel(channel, peer, remoteNodeId, isFunder, channelId)
       channelPaneController.updateBalance(currentData.commitments)
       val m1 = m + (channel -> channelPaneController)
-      val totalBalance = MilliSatoshi(m1.values.map(_.getBalance.amount).sum)
+      val totalBalance = MilliSatoshi(m1.values.map(_.getBalance.toLong).sum)
```
araspitzu: This can use `sum` on the `MilliSatoshi` class since we also have defined `NumericMilliSatoshi extends Numeric[MilliSatoshi]`

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/controllers/ReceivePaymentController.scala
```diff
@@ -71,7 +71,7 @@ class ReceivePaymentController(val handlers: Handlers, val stage: Stage) extends
       case "" => createPaymentRequest(None)
       case GUIValidators.amountDecRegex(_*) =>
         Try(CoinUtils.convertStringAmountToMsat(amount.getText, unit.getValue)) match {
-          case Success(amountMsat) if amountMsat.amount < 0 =>
+          case Success(amountMsat) if amountMsat.toLong < 0 =>
```
araspitzu:
```suggestion
          case Success(amountMsat) if amountMsat < 0.msat =>
```

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/GUIUpdater.scala
t-bast: Good catch!

eclair-core/src/main/scala/fr/acinq/eclair/MilliSatoshi.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  override def compare(other: MilliSatoshi): Int = underlying.compareTo(other.underlying)
+  // Since BtcAmount is a sealed trait that MilliSatoshi cannot extend, we need to redefine comparison operators.
+  def compare(other: BtcAmount): Int = compare(other.toMilliSatoshi)
+  def <=(other: BtcAmount): Boolean = compare(other) <= 0
+  def >=(other: BtcAmount): Boolean = compare(other) >= 0
+  def <(other: BtcAmount): Boolean = compare(other) < 0
+  def >(other: BtcAmount): Boolean = compare(other) > 0
+
+  def max(other: MilliSatoshi): MilliSatoshi = if (this > other) this else other
```
pm47: Maybe a comment explaining why comparisons are assymetric?

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -156,58 +154,40 @@ package object eclair {
...
-    }
-  }
-
   implicit object NumericMilliSatoshi extends Numeric[MilliSatoshi] {
-    override def plus(x: MilliSatoshi, y: MilliSatoshi): MilliSatoshi = MilliSatoshi(x.amount + y.amount)
-    override def minus(x: MilliSatoshi, y: MilliSatoshi): MilliSatoshi = MilliSatoshi(x.amount - y.amount)
-    override def times(x: MilliSatoshi, y: MilliSatoshi): MilliSatoshi = MilliSatoshi(x.amount * y.amount)
-    override def negate(x: MilliSatoshi): MilliSatoshi = MilliSatoshi(-x.amount)
+    // @formatter:off
+    override def plus(x: MilliSatoshi, y: MilliSatoshi): MilliSatoshi = x + y
```
pm47: A comment explaining why we also need to implement `Numeric[MilliSatoshi]` would be helpful.

eclair-core/src/test/java/fr/acinq/eclair/MilliSatoshiTest.java
pm47: Just making sure that it compiles seems good enough for now.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -16,54 +16,54 @@
...
+ * @param amount    amount to pay (empty string means no amount is specified)
+ * @param timestamp request timestamp (UNIX format)
+ * @param nodeId    id of the node emitting the payment request
+ * @param tags      payment tags; must include a single PaymentHash tag
+ * @param signature request signature that will be checked against node id
+ */
 case class PaymentRequest(prefix: String, amount: Option[MilliSatoshi], timestamp: Long, nodeId: PublicKey, tags: List[PaymentRequest.TaggedField], signature: ByteVector) {
 
-  amount.map(a => require(a.amount > 0, s"amount is not valid"))
+  amount.map(a => require(a.toLong > 0, s"amount is not valid"))
```
pm47:
```suggestion
  amount.map(a => require(a > 0.msat, "amount is not valid"))
```

PR #1104, opened 2019-08-22
Typed cltv expiry

t-bast:
When implementing payment changes (MPP and Trampoline) I found it very confusing that cltv expiry wasn't typed (and that expiry variables were half of the time incorrectly named, often not mentioning that it was a delta).

This change makes it simpler to reason about expiries and let the compiler help us when we're messing things up between expiry delta and absolute expiry.

Apologies for the long and tedious review since it touches most files :D


eclair-core/src/main/scala/fr/acinq/eclair/CltvExpiry.scala
```diff
@@ -0,0 +1,67 @@
...
+ * Created by t-bast on 21/08/2019.
+ */
+
+/**
+ * Bitcoin scripts (in particular HTLCs) need an absolute block expiry (greater than the current block count) to work
+ * with OP_CLTV.
+ *
+ * @param get the absolute cltv expiry value (current block count + some delta).
+ */
+case class CltvExpiry(get: Long) {
```
pm47:
```suggestion
case class CltvExpiry(private val underlying: Long) {
```

And add a `def toLong: Long` getter

eclair-core/src/main/scala/fr/acinq/eclair/CltvExpiry.scala
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Channels advertise a cltv expiry delta that should be used when routing through them.
+ * This value needs to be converted to a [[fr.acinq.eclair.CltvExpiry]] to be used in OP_CLTV.
+ *
+ * CltvExpiryDelta can also be used when working with OP_CSV which is by design a delta.
+ *
+ * @param delta the cltv expiry delta value.
+ */
+case class CltvExpiryDelta(delta: Int) {
```
pm47:
```suggestion
case class CltvExpiryDelta(private val underlying: Int) {
```

And add a `def toInt: Int` getter

pm47:
How should we handle comparisons?

t-bast:
> How should we handle comparisons?

I made both classes extend `Ordered`. I don't think it makes sense to compare `CltvExpiry` to `CltvExpiryDelta`, so we shouldn't need anything else (comparing with `Long` or `Int` could be misleading too so I didn't implement anything for that).

PR #1105, opened 2019-08-22
Use unsigned comparison for 'maxHtlcValueInFlightMsat'

araspitzu:
While in most places we can store the currency values in a signed `Long` because the maximum number of millisatoshi is < 2^63 we can't do the same with `maxHtlcValueInFlightMsat` because this is a limit declared by one party of the channel and such limit can be set to any arbitrary number up to 2^64-1.This PR makes use of unsigned comparison between the current in-flight HTLC value and the `maxHtlcValueInFlightMsat`.


t-bast:
Maybe this is the right occasion to work on backing `UInt64` by a Long instead of a `BigInt`?
I know we've talked about it several times, now that we are in "fix our types madness mode" we could tackle that :)

araspitzu:
> backing UInt64 by a Long instead of a BigInt

I totally agree with that although i'm not sure if this is the right PR.

t-bast:
> I totally agree with that although i'm not sure if this is the right PR.

This should definitely be done in a separate PR, but I think backing `UInt64` by a `Long` would change how you would implement compare for `maxHtlcValueInflightMsat`. I think doing the `UInt64` re-work could be a pre-requisite before doing the work you're doing in this PR. What do you think?

araspitzu:
In this PR i'm already storing 'maxHtlcValueInFlightMsat' in `MilliSatoshi` (so backed by Long) and using unsigned comparisons when needed, do you think we should move it back to `UInt64` once it's backed by `Long`?

t-bast:
> In this PR i'm already storing 'maxHtlcValueInFlightMsat' in MilliSatoshi 

On the contrary, you're keeping it as a `UInt64`...

What I meant is that depending on how we re-wor `UInt64` to be backed by a `Long`, the unsigned compare you're introducing might not make sense anymore, so it might be worth starting with the `UInt64` re-work before doing comparisons between `UInt64` and `MilliSatoshi`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -148,8 +149,9 @@ object Commitments {
     // the HTLC we are about to create is outgoing, but from their point of view it is incoming
     val outgoingHtlcs = reduced.htlcs.filter(_.direction == IN)
 
-    val htlcValueInFlight = UInt64(outgoingHtlcs.map(_.add.amountMsat).sum.toLong)
-    if (htlcValueInFlight > commitments1.remoteParams.maxHtlcValueInFlightMsat) {
+    val htlcValueInFlight = outgoingHtlcs.map(_.add.amountMsat).sum
```
t-bast: I'm afraid this can wrap around since `amountMsat` is a `Long`. If that sum exceeds `Long.MaxValue` it will wrap to a negative value instead of being the correct `UInt64` value we want to use for the comparison, right?

t-bast:
But I'm not sure, I haven't looked enough into the `UInt64` backing by a `Long`, maybe it doesn't change anything for this PR?

araspitzu:
I don't think it changes much for this PR, here we're treating `maxHtlcValueInFlightMsat` as a special case because it can have values like `0xffffffffffffffff` but we still want to type it with MilliSatoshi, so using a `Long` while keeping it unsigned. The danger here is to compare signed values with unsigned values, in this case we need to make sure the first term of the comparison is greater than zero (i.e [here](https://github.com/ACINQ/eclair/pull/1105/files#diff-0383ea0b8b273b3430e02133c31301aaR152) `htlcValueInFlight` is always positive). If we change `UInt64` to be backed by `Long` this field is not affected (because it's a MilliSatoshi) but surely the change will affect other parts of the codebase :)

PR #1106, opened 2019-08-22
Bolt4: remove final_expiry_too_soon error message

t-bast:
It allowed probing attacks and the spec deprecated it in favor of IncorrectOrUnknownPaymentDetails.


t-bast:
Done. I had to get rid of using traits for `PERM` and `NODE` because that made it hard to apply to unknown failure messages.
I think the only useful information we want to get out of unknown failure messages is whether it's a temporary or permanent failure (which impacts retries), We can't really leverage `UPDATE` and `BADONION` because we don't know how to decode the payload, so we can't know where the onion hash and channel update is. I think it's ok.

eclair-core/src/main/scala/fr/acinq/eclair/wire/FailureMessage.scala
```diff
@@ -52,13 +58,25 @@ case class AmountBelowMinimum(amount: MilliSatoshi, update: ChannelUpdate) exten
...
 case object ExpiryTooFar extends FailureMessage { def message = "payment expiry is too far in the future" }
+
+/**
+ * We allow remote nodes to send us unknown failure codes (e.g. deprecated failure codes).
+ * By reading the PERM and NODE bits we can still extract useful information for payment retry even without knowing how
+ * to decode the failure payload (but we can't extract a channel update or onion hash).
+ */
+sealed trait UnknownFailureMessage extends FailureMessage {
+  def message = "unknown failure message"
+  override def toString = s"$message ($code)"
```
pm47:
```suggestion
  override def toString = s"$message (0x${code.toHexString})"
```

PR #1107, opened 2019-08-22
Fix regression in `Commitments.availableForSend`

pm47:
We must consider `nextRemoteCommit` when applicable.

This is a regression caused in #784. The core bug only exists when we
have a pending unacked `commit_sig`, but since we only send the
`AvailableBalanceChanged` event when sending a signature (not when
receiving a revocation), actors relying on this event to know the
current available balance (e.g. the `Relayer`) will have a wrong
value in-between two outgoing sigs.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -85,7 +85,9 @@ case class Commitments(channelVersion: ChannelVersion,
   val announceChannel: Boolean = (channelFlags & 0x01) != 0
 
   lazy val availableBalanceForSend: MilliSatoshi = {
-    val reduced = CommitmentSpec.reduce(remoteCommit.spec, remoteChanges.acked, localChanges.proposed)
+    // we need to base the next current commitment on the last sig we sent, even if we didn't yet receive their revocation
+    val remoteCommit1 = remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit)
+    val reduced = CommitmentSpec.reduce(remoteCommit1.spec, remoteChanges.acked, localChanges.proposed)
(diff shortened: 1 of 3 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/channel/CommitmentsSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+    import setup._
+    within(30 seconds) {
+      reachNormal(setup, test.tags)
+      awaitCond(alice.stateName == NORMAL)
+      awaitCond(bob.stateName == NORMAL)
+      withFixture(test.toNoArgTest(setup))
+    }
+  }
+
+  test("correct values for availableForSend/availableForReceive") { f =>
```
t-bast: Nice test :)
I think it would be good to make almost the same test but with an HTLC_FAIL at the end, what do you think?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -85,7 +85,9 @@ case class Commitments(channelVersion: ChannelVersion,
   val announceChannel: Boolean = (channelFlags & 0x01) != 0
 
   lazy val availableBalanceForSend: MilliSatoshi = {
-    val reduced = CommitmentSpec.reduce(remoteCommit.spec, remoteChanges.acked, localChanges.proposed)
+    // we need to base the next current commitment on the last sig we sent, even if we didn't yet receive their revocation
+    val remoteCommit1 = remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit)
```
pm47: NB: this works too:

```suggestion
    val remoteCommit1 = remoteNextCommitInfo.left.map(_.nextRemoteCommit).left.getOrElse(remoteCommit)
```

pm47:
NB: those things are subtle, please review carefully...

PR #1108, opened 2019-08-23
Channel range queries: send back node announcements

sstone:
This PR adds support for sending back node announcements when replying to channel range queries:
- when explicitly requested (bit is set in the optional query flag)
- when query flags are not used and a channel announcement is sent (as per the BOLTs)

A new configuration option `request-node-announcements` has been added in the `router` section. If set to true, we will request node announcements when we receive a channel id (through channel range queries) that we don't know of.
This is a setting that we will probably turn off on mobile devices.


eclair-core/src/main/resources/reference.conf
```diff
@@ -116,6 +116,7 @@ eclair {
     channel-exclude-duration = 60 seconds // when a temporary channel failure is returned, we exclude the channel from our payment routes for this duration
     broadcast-interval = 60 seconds // see BOLT #7
     init-timeout = 5 minutes
+    request-node-announcements = true // if true we will ask for node annnouncements when we receive channel ids that we don't know
```
```suggestion
    request-node-announcements = true // if true we will ask for node announcements when we receive channel ids that we don't know
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -573,26 +574,58 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
...
+          }
+          if (flag_opt.map(QueryShortChannelIdsTlv.QueryFlagType.includeUpdate2).getOrElse(true)) {
+            d.updates.get(ChannelDesc(ca.shortChannelId, ca.nodeId2, ca.nodeId1)).foreach { u =>
+              transport ! u
+              numcu1 = numcu1 + 1
+            }
+          }
+          if (flag_opt.map(QueryShortChannelIdsTlv.QueryFlagType.includeNodeAnnouncement1).getOrElse(true) && !sent1.contains(ca.nodeId1)) {
+            d.nodes.get(ca.nodeId1).foreach { a =>
+              transport ! a
```
pm47: Our convention is `c` for channels, `u` for updates and `n` for nodes.

```suggestion
              transport ! n
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -573,26 +574,58 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
...
+            d.nodes.get(ca.nodeId2).foreach { a =>
+              transport ! a
+              sent1 = sent1 + ca.nodeId2
+            }
+          }
+          loop(tail, flags.drop(1), numca1, numcu1, sent1)
+      }
+
+      val (channelCount, updateCount, nodeCount) = loop(shortChannelIds.array, flags)
+      log.info("received query_short_channel_ids with {} items, sent back {} channels and {} updates and {} node announcements", shortChannelIds.array.size, channelCount, updateCount, nodeCount)
```
```suggestion
      log.info("received query_short_channel_ids with {} items, sent back {} channels and {} updates and {} node", shortChannelIds.array.size, channelCount, updateCount, nodeCount)
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/QueryShortChannelIdsTlv.scala
```diff
@@ -20,13 +20,19 @@ object QueryShortChannelIdsTlv {
     val INCLUDE_CHANNEL_ANNOUNCEMENT: Long = 1
     val INCLUDE_CHANNEL_UPDATE_1: Long = 2
     val INCLUDE_CHANNEL_UPDATE_2: Long = 4
-    val INCLUDE_ALL: Long = (INCLUDE_CHANNEL_ANNOUNCEMENT | INCLUDE_CHANNEL_UPDATE_1 | INCLUDE_CHANNEL_UPDATE_2)
+    val INCLUDE_NODE_ANNOUNCEMENT_1: Long = 8
+    val INCLUDE_NODE_ANNOUNCEMENT_2: Long = 16
+    val INCLUDE_ALL: Long = (INCLUDE_CHANNEL_ANNOUNCEMENT | INCLUDE_CHANNEL_UPDATE_1 | INCLUDE_CHANNEL_UPDATE_2 | INCLUDE_NODE_ANNOUNCEMENT_1 | INCLUDE_NODE_ANNOUNCEMENT_2)
```
pm47: `INCLUDE_ALL` doesn't seem very useful

PR #1109, opened 2019-08-26
Use Long to back the UInt64 type

araspitzu:
The java `Long` uses 64 bits to store a signed integer, since we want to store 64 bits unsigned integers we have a `UInt64` type backed by `BigInt`. This PR improves this by storing an unsigned 64 bits integer in a `Long` (providing unsigned comparison) and using that to back the `UInt64` type.


eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,36 @@
...
 
-case class UInt64(private val underlying: BigInt) extends Ordered[UInt64] {
+case class UInt64(private val underlying: Long) extends Ordered[UInt64] {
 
-  require(underlying >= 0, s"uint64 must be positive (actual=$underlying)")
-  require(underlying <= UInt64.MaxValueBigInt, s"uint64 must be < 2^64 -1 (actual=$underlying)")
+  override def compare(o: UInt64): Int = UnsignedLongs.compare(underlying, o.underlying)
 
-  override def compare(o: UInt64): Int = underlying.compare(o.underlying)
+  def toByteVector: ByteVector = ByteVector.fromValidHex(UnsignedLongs.toString(underlying, 16))
```
sstone: Using a codec would be much more efficient that parsing an hex string

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,36 @@
...
-  val MaxValue = UInt64(MaxValueBigInt)
+  val MaxValue = UInt64(hex"0xffffffffffffffff")
 
-  def apply(bin: ByteVector) = new UInt64(new BigInteger(1, bin.toArray))
-
-  def apply(value: Long) = new UInt64(BigInt(value))
+  def apply(bin: ByteVector): UInt64 = bin.toHex match {
+    case hex if hex.isEmpty => UInt64(0)
+    case hex => new UInt64(UnsignedLongs.parseUnsignedLong(hex, 16))
+  }
```
sstone: Same here, a codec would be more efficient

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
araspitzu: Is it a problem if `toByteVector` here produces padded byte arrays? This is how the `uint64` codec in `CommonsCodec` behaves.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
araspitzu: Do you mean using the `uint64` scodec? We can't use it here because it's defined using this constructor so it would cause an infinite loop.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,42 @@
...
+  val MaxValue = UInt64(hex"0xffffffffffffffff")
 
-  val MaxValue = UInt64(MaxValueBigInt)
-
-  def apply(bin: ByteVector) = new UInt64(new BigInteger(1, bin.toArray))
-
-  def apply(value: Long) = new UInt64(BigInt(value))
+  def apply(bin: ByteVector): UInt64 = UInt64(bin.toLong(signed = false))
 
   object Conversions {
```
t-bast: I think these conversions aren't needed anymore

araspitzu: Actually they're still in use by [TlvCodecs](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala#L36), we could use a more explicit conversion there but i'm afraid it becomes quite uglier because we'd need to do unsigned comparisons (which is currently achieved by implicitly converting the right hand side to a `UInt64` and using its `compare` implementation)

t-bast: Ah no you're right, I completely messed that up :)

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,42 @@
...
-case class UInt64(private val underlying: BigInt) extends Ordered[UInt64] {
+case class UInt64(private val underlying: Long) extends Ordered[UInt64] {
 
-  require(underlying >= 0, s"uint64 must be positive (actual=$underlying)")
-  require(underlying <= UInt64.MaxValueBigInt, s"uint64 must be < 2^64 -1 (actual=$underlying)")
+  override def compare(o: UInt64): Int = UnsignedLongs.compare(underlying, o.underlying)
 
-  override def compare(o: UInt64): Int = underlying.compare(o.underlying)
+  def toByteVector: ByteVector = this match {
+    case x if x <= UInt64(255) => ByteVector.fromLong(underlying, size = 1)
```
t-bast: A more compact way of doing this:
```scala
def toByteVector: ByteVector = underlying match {
  case 0 => hex"00"
  case _ => ByteVector.fromLong(underlying).dropWhile(_ == 0)
}
```

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,42 @@
...
+    case x if x <= UInt64(281474976710655L) => ByteVector.fromLong(underlying, size = 6)
+    case x if x <= UInt64(72057594037927935L) => ByteVector.fromLong(underlying, size = 7)
+    case _ => ByteVector.fromLong(underlying, size = 8)
+  }
 
-  def toByteVector: ByteVector = ByteVector.view(underlying.toByteArray.takeRight(8))
+  def toBigInt: BigInt = BigInt(toString)
 
-  def toBigInt: BigInt = underlying
+  def toLong: Long = underlying
```
t-bast: I would remove this: it's unused and can be confusing when UInt64 overflows a positive Long, don't you think?

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,42 @@
 
 package fr.acinq.eclair
 
-import java.math.BigInteger
-
+import com.google.common.primitives.UnsignedLongs
 import scodec.bits.ByteVector
+import scodec.bits._
```
t-bast: nit: you only need `scodec.bits.HexStringSyntax`, no need to import everything

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,42 @@
...
 
-case class UInt64(private val underlying: BigInt) extends Ordered[UInt64] {
+case class UInt64(private val underlying: Long) extends Ordered[UInt64] {
 
-  require(underlying >= 0, s"uint64 must be positive (actual=$underlying)")
-  require(underlying <= UInt64.MaxValueBigInt, s"uint64 must be < 2^64 -1 (actual=$underlying)")
+  override def compare(o: UInt64): Int = UnsignedLongs.compare(underlying, o.underlying)
```
t-bast: I would move the `compareUnsigned` that's currently in `MilliSatoshi` and put it here: this way the use of `UnsignedLongs` is constrained to this file, which is really the only place where we want to hide it.
What do you think?

eclair-core/src/test/scala/fr/acinq/eclair/UInt64Spec.scala
```diff
@@ -31,6 +31,7 @@ class UInt64Spec extends FunSuite {
     assert(b < a)
     assert(z < a && z < b && z < c)
     assert(a == a)
+    assert(a == UInt64.MaxValue)
```
t-bast: I think it's also worth enriching these tests with more edge cases around `Long.MaxValue`.
I'd also test the toString method.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,42 @@
...
+  }
 
-  def toByteVector: ByteVector = ByteVector.view(underlying.toByteArray.takeRight(8))
+  def toBigInt: BigInt = BigInt(toString)
 
-  def toBigInt: BigInt = underlying
+  def toLong: Long = underlying
 
-  override def toString: String = underlying.toString
+  override def toString: String = UnsignedLongs.toString(underlying, 10)
```
t-bast: Maybe adding `toHexString` can be useful to debug big values?

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,42 @@
...
+    case x if x <= UInt64(16777215) => ByteVector.fromLong(underlying, size = 3)
+    case x if x <= UInt64(4294967295L) => ByteVector.fromLong(underlying, size = 4)
+    case x if x <= UInt64(1099511627775L) => ByteVector.fromLong(underlying, size = 5)
+    case x if x <= UInt64(281474976710655L) => ByteVector.fromLong(underlying, size = 6)
+    case x if x <= UInt64(72057594037927935L) => ByteVector.fromLong(underlying, size = 7)
+    case _ => ByteVector.fromLong(underlying, size = 8)
+  }
 
-  def toByteVector: ByteVector = ByteVector.view(underlying.toByteArray.takeRight(8))
+  def toBigInt: BigInt = BigInt(toString)
```
t-bast: It feels wasteful to go a string to convert, I'd use `BigInt(toByteVector.toArray)`

eclair-core/src/test/scala/fr/acinq/eclair/UInt64Spec.scala
t-bast: Overall this `UInt64` is full of subtleties and everything could go wrong if we didn't handle an edge case well, so I'd add a lot more things here to feel safe...
That's also a good candidate for fuzzing when we add that to eclair.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
araspitzu: Agreed, it's confusing and mostly wrong because a signed Long can't accommodate an uint64.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
araspitzu: Nice one! I was a bit wary of doing so many comparisons and this solves it elegantly.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
araspitzu: Not sure about that, using `toByteVector` the caller can easily get the hex string calling `toHex`.

eclair-core/src/test/scala/fr/acinq/eclair/UInt64Spec.scala
araspitzu: I did some similar testing in [here](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/test/scala/fr/acinq/eclair/CoinUtilsSpec.scala#L32), it's related but i think is worth repeating the tests in `UInt64Spec`. As you said this is plenty of subtleties and thorough testing is the first step to get them right.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
t-bast: True

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
araspitzu: I like this proposal, will do a `UInt64.compareUnsigned(other: MilliSatoshi)`

araspitzu: bfd7f6b moves the unsigned comparison into the `UInt64` class, also the comparison now takes into account the possible negative values for MilliSatoshi and guarantees that `UInt64(_) > MilliSatoshi(y)` for any `y` smaller than `0`

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,38 @@
...
 
-case class UInt64(private val underlying: BigInt) extends Ordered[UInt64] {
-
-  require(underlying >= 0, s"uint64 must be positive (actual=$underlying)")
-  require(underlying <= UInt64.MaxValueBigInt, s"uint64 must be < 2^64 -1 (actual=$underlying)")
+case class UInt64(private val underlying: Long) extends Ordered[UInt64] {
 
-  override def compare(o: UInt64): Int = underlying.compare(o.underlying)
+  override def compare(o: UInt64): Int = UnsignedLongs.compare(underlying, o.underlying)
+  def compareUnsigned(other: MilliSatoshi) = other.toLong match {
```
t-bast: I would even rename that to `compare` (since it's implicit that we compare unsigned because UInt64 is unsigned).
And I would also define:
```scala
def <(other: MilliSatoshi): Boolean
def >(other: MilliSatoshi): Boolean
def <=(other: MilliSatoshi): Boolean
def >=(other: MilliSatoshi): Boolean
```
like what we do in MilliSatoshi to compare with BtcAmount.
What do you think?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -148,7 +148,7 @@ object Commitments {
 
     val htlcValueInFlight = outgoingHtlcs.map(_.add.amountMsat).sum
     // we must use unsigned comparison here because 'maxHtlcValueInFlightMsat' is effectively an uint64 and can exceed the capacity of MilliSatoshi class
-    if (htlcValueInFlight.compareUnsigned(commitments1.remoteParams.maxHtlcValueInFlightMsat) > 0) {
+    if (commitments1.remoteParams.maxHtlcValueInFlightMsat.compareUnsigned(htlcValueInFlight) < 0) {
```
t-bast: If you implement `<` you can write `commitments1.remoteParams.maxHtlcValueInFlightMsat < htlcValueInFlight` which is much better :)

eclair-core/src/test/scala/fr/acinq/eclair/UInt64Spec.scala
```diff
@@ -22,23 +22,52 @@ import scodec.bits._
...
     assert(b.toString === "18446744073709551614")
     assert(c.toByteVector === hex"0x2a")
     assert(c.toString === "42")
     assert(z.toByteVector === hex"0x00")
     assert(z.toString === "0")
+    assert(UInt64(hex"0xff").toByteVector == hex"0xff")
+    assert(UInt64(hex"0x800").toByteVector == hex"0x800")
+  }
+
+  test("use unsigned comparison when comparing millisatoshis to uint64") {
```
t-bast: And those tests would also be much easier to read if it used `<` and `>`!

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
araspitzu: I see the point of your suggestion but i also like having this (unsigned comparison) very explicit, after all is a bit of a special case; defining those comparison operators feels like hiding the complexity because at first glance the reader won't see that a special comparison is used. With that said i'm not strongly opposed and if we want to do it this way i'm okay as well (probably a comment in the code is enough?).

t-bast: I think it's weird that `UInt64` (which is explicitly unsigned) defines a method `compareUnsigned` :)
Especially since it's not really a `compareUnsigned` anymore: we take into account the `MilliSatoshi` sign: if negative it's smaller than our `UInt64`, otherwise we are just comparing two positive values (and the fact that we internally use a compareUnsigned on long is just a side effect of how we encode our UInt64).

araspitzu: That's a fair point, `compareUnsigned` is a bit weird since is defined on an unsigned type. Okay i will do that, let's see how it goes through the review :)

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,43 @@
...
-  override def compare(o: UInt64): Int = underlying.compare(o.underlying)
+  def <(other: MilliSatoshi): Boolean = compare(other) < 0
+  def >(other: MilliSatoshi): Boolean = compare(other) > 0
+  def <=(other: MilliSatoshi): Boolean = compare(other) <= 0
+  def >=(other: MilliSatoshi): Boolean = compare(other) >= 0
 
-  def toByteVector: ByteVector = ByteVector.view(underlying.toByteArray.takeRight(8))
+  def toByteVector: ByteVector = underlying match {
+    case 0 => hex"00"
+    case _ => ByteVector.fromLong(underlying).dropWhile(_ == 0)
```
pm47: The `dropWhile` shouldn't be necessary anymore

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -183,8 +182,7 @@ object Commitments {
     val incomingHtlcs = reduced.htlcs.filter(_.direction == IN)
 
     val htlcValueInFlight = incomingHtlcs.map(_.add.amountMsat).sum
-    // we must use unsigned comparison here because 'maxHtlcValueInFlightMsat' is effectively an uint64 and can exceed the capacity of MilliSatoshi class
-    if (htlcValueInFlight.compareUnsigned(commitments1.localParams.maxHtlcValueInFlightMsat) > 0) {
+    if (commitments1.localParams.maxHtlcValueInFlightMsat < htlcValueInFlight) {
```
pm47: The removed comment would probably have helped somewhere around line 30 in UInt64.scala.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
t-bast: I think we need it, otherwise we will always have something that's 8-bytes long?

pm47:
Did we keep the multiple `UInt64.toBigInt.toLong` on purpose?

t-bast:
> Did we keep the multiple UInt64.toBigInt.toLong on purpose?

What's the behavior for a UInt64 that overflows Long.MaxValue?
Is `.toBigInt.toLong` different from `underlying`?

araspitzu:
The behavior or `toBigInt.toLong` is equivalent to reading the `underlying` Long of `UInt64`, reference api https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html#longValue() . There is also the `longValueExact` function that avoids losing precision but unfortunately is only available for JDK >= 8.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
```diff
@@ -16,40 +16,43 @@
...
+  }
 
-  def toBigInt: BigInt = underlying
+  def toBigInt: BigInt = BigInt(signum = 1, toByteVector.toArray)
```
pm47: Feels a bit weird to go through `toByteVector` for converting to `BigInt`. Stackoverflow suggests:

```suggestion
  def toBigInt: BigInt = (BigInt(underlying >>> 1) << 1) + (underlying & 1)
```

source: https://stackoverflow.com/questions/21212993/unsigned-variables-in-scala

Tests pass with this change.

eclair-core/src/test/scala/fr/acinq/eclair/UInt64Spec.scala
```diff
@@ -22,23 +22,52 @@ import scodec.bits._
...
+    assert(a == UInt64.MaxValue)
+    assert(l.toByteVector == hex"7fffffffffffffff")
+    assert(l.toString == Long.MaxValue.toString)
+    assert(l.toBigInt == BigInt(Long.MaxValue))
+    assert(l1.toByteVector == hex"8000000000000000")
+    assert(l1.toString == "9223372036854775808")
+    assert(l1.toBigInt == BigInt("9223372036854775808"))
     assert(a.toByteVector === hex"0xffffffffffffffff")
-    assert(a.toString === "18446744073709551615")
+    assert(a.toString === "18446744073709551615") // 2^64 - 1
```
pm47: I feel we are missing a few tests about `toBigInt`. Patch:

```diff
diff --git a/eclair-core/src/test/scala/fr/acinq/eclair/UInt64Spec.scala b/eclair-core/src/test/scala/fr/acinq/eclair/UInt64Spec.scala
index 16f58f83..cbd75988 100644
--- a/eclair-core/src/test/scala/fr/acinq/eclair/UInt64Spec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/UInt64Spec.scala
@@ -38,20 +38,31 @@ class UInt64Spec extends FunSuite {
     assert(z < a && z < b && z < c && z < l && c < l && l < l1 && l < b && l < a)
     assert(a == a)
     assert(a == UInt64.MaxValue)
+
     assert(l.toByteVector == hex"7fffffffffffffff")
     assert(l.toString == Long.MaxValue.toString)
     assert(l.toBigInt == BigInt(Long.MaxValue))
+
     assert(l1.toByteVector == hex"8000000000000000")
     assert(l1.toString == "9223372036854775808")
     assert(l1.toBigInt == BigInt("9223372036854775808"))
+
     assert(a.toByteVector === hex"0xffffffffffffffff")
     assert(a.toString === "18446744073709551615") // 2^64 - 1
+    assert(a.toBigInt === BigInt("18446744073709551615"))
+
     assert(b.toByteVector === hex"0xfffffffffffffffe")
     assert(b.toString === "18446744073709551614")
+    assert(b.toBigInt === BigInt("18446744073709551614"))
+
     assert(c.toByteVector === hex"0x2a")
     assert(c.toString === "42")
+    assert(c.toBigInt === BigInt("42"))
+
     assert(z.toByteVector === hex"0x00")
     assert(z.toString === "0")
+    assert(z.toBigInt === BigInt("0"))
+
     assert(UInt64(hex"0xff").toByteVector == hex"0xff")
     assert(UInt64(hex"0x800").toByteVector == hex"0x800")
   }

```

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
pm47: Replacing the whole method by:

```scala
  def toByteVector: ByteVector = ByteVector.fromLong(underlying)
```

should work (except that it makes cosmetic tests fail). I think it is simpler that way, do you see a potential problem?

araspitzu:
> Replacing the whole method by:
>
>  def toByteVector: ByteVector = ByteVector.fromLong(underlying)
> should work (except that it makes cosmetic tests fail). I think it is simpler that way, do you see a potential problem?

I don't see problems with that, currently `toByteVector` is used by `CommonsCodecs`, `TlvCodecs` and `toBigInt`. I just tried to remove the truncation and the codec tests are passing, unless in some part of the code we require the `UInt64` to be minimally encoded in a byte vector this change should be fine.

eclair-core/src/main/scala/fr/acinq/eclair/UInt64.scala
araspitzu: The change have been implemented in 8c60d6a, while a bit less readable i think this solution is elegant and more efficient than going through the byte vector conversion.

PR #659, opened 2018-07-16
[WIP] Rework fulfill/fail commands acknowledgements

pm47:
When we receive an `UpdateFulfillHtlc` from a downstream channel, it is critical that we don't lose it because it allows us to pull funds from some upstream channel. But this upstream channel may very well be currently `OFFLINE` and unable to update the commitment right away, so we need to remember it for later. Same applies to an `UpdateFailHtlc` (even if it is less critical), because we don't want the upstream htlc to timeout and cause a channel to get closed.

We were previously relying on a `CommandBuffer` actor, which made the flow a little cumbersome and not easy to understand. 

Instead, when a channel is in `OFFLINE` state and can't process a `CMD_FULFILL_HTLC`/`CMD_FAIL_HTLC`/`CMD_FAIL_MALFORMED_HTLC`, the channel itself is responsible for storing the command, and will reprocess it when it goes online again or gets closed.

PR #1111, opened 2019-08-28
Avoid pendingRelays db call each time a channel becomes operational

btcontract:
Currently every time each channel is reestablished there is a `pendingRelays` db call in `CommandBuffer`, this perhaps can become too slow if there are a lot of channels, also it may be abused by flapping channels.

This PR changes this by keeping pending relays in `CommandBuffer` map which gets filled from db once when `CommandBuffer` is instantiated.


eclair-core/src/main/scala/fr/acinq/eclair/db/PendingRelayDb.scala
```diff
@@ -33,14 +33,12 @@ import fr.acinq.eclair.channel.Command
...
+  def addPendingRelay(channelId: ByteVector32, cmd: HasHtlcIdCommand)
 
   def removePendingRelay(channelId: ByteVector32, htlcId: Long)
 
-  def listPendingRelay(channelId: ByteVector32): Seq[Command]
+  def listPendingRelay(channelId: ByteVector32): Seq[HasHtlcIdCommand]
 
-  def listPendingRelay(): Set[(ByteVector32, Long)]
-
-  def close(): Unit
```
t-bast: Why did you remove `close()`?

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -207,14 +207,14 @@ object Switchboard extends Logging {
...
     // the command from the pending relay db.
-    val channel2Htlc: Set[(ByteVector32, Long)] =
-    channels
-      .flatMap(_.commitments.remoteCommit.spec.htlcs)
-      .filter(_.direction == OUT)
-      .map(htlc => (htlc.add.channelId, htlc.add.id))
-      .toSet
-
-    val pendingRelay: Set[(ByteVector32, Long)] = relayDb.listPendingRelay()
+    val channel2Htlc: Seq[(ByteVector32, Long)] = for {
```
t-bast: I don't get why you're changing that method...

eclair-core/src/main/scala/fr/acinq/eclair/db/PendingRelayDb.scala
btcontract: it was highlighted as not used anywhere

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
btcontract: it was just easier for me to reason what's going on by replacing with method calls with `for`

btcontract: Or, if you mean `Set -> Seq`, it's because `.toSet` in old method was superfluous because it's only used in `val toClean = pendingRelay -- channel2Htlc` where `pendingRelay` is a `Set` already and `channel2Htlc` can be just a `Seq`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/CommandBuffer.scala
```diff
@@ -29,44 +31,49 @@ import fr.acinq.eclair.channel._
...
 
-    case CommandSend(channelId, htlcId, cmd) =>
+    case CommandSend(channelId, cmd) =>
       // save command in db
-      register forward Register.Forward(channelId, cmd)
+      register ! Register.Forward(channelId, cmd)
       // we also store the preimage in a db (note that this happens *after* forwarding the fulfill to the channel, so we don't add latency)
-      pendingRelay.addPendingRelay(channelId, htlcId, cmd)
+      nodeParams.db.pendingRelay.addPendingRelay(channelId, cmd)
+      context become main(pendingRelays.addBinding(channelId, cmd.modify(_.commit).setTo(false)))
```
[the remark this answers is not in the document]
btcontract: It's because of this line: https://github.com/ACINQ/eclair/pull/1111/files#diff-5a1f33d183f796f6859fb09367579dafR64

Once stashed they all should have `commit = false` because `CommandBuffer` sends all of them and then a final `CMD_SIGN`, it's faster this way.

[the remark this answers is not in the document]
araspitzu: Right, then i think this should be in a comment maybe above the declaration of the `CommandBuffer` class. A more fancy solution would be to have a strategy pattern approach to be able to select the desired behavior of the buffer but probably we can stick with this default strategy (but it's nice to leave a comment).

pm47:
You're effectively introducing a cache (which is the second most complicated things in computer science after naming things), and adding it to a sensitive part of the app. For those reasons I'm not too keen on making the change without hard data to back it up. How much do we really gain in practice?

I like the the `HasHtlcIdCommand ` trait, but I would define it like:

```scala
sealed trait HasHtlcId extends Command {
  def id: Long
  def commit: Boolean
}
```

pm47:
Also the whole pending relay db thing is awkward. There was already a failed attempt at redoing it with #659, but we are going to get back to it soonish.

btcontract:
Well I must admit this is not something which causes problems right now, at least to me. 

This is a part of my work of implementing hosted channels in Eclair, as these channels are supposed to be created automatically with each app install I can imagine a single Eclair instance having __a lot__ of them if this thing takes off, so I try to identify and eliminate places which could be a performance bottleneck.

If you do have a different plan for handling offline fails/fulfills, perhaps you could outline it? I may attempt to implement it.

pm47:
This seems premature, and is an easy fix if needed (see https://github.com/ACINQ/eclair/pull/1476#issuecomment-651858852).

PR #1112, opened 2019-08-29
Make tests run in parallel

pm47:
There are two level of parallelization possible:
- Between test suites (a suite = a test file)
Ok, for all tests, since #1127
- Within a suite
Depends on tests suites, some rely on sequential execution of tests, some don't.


pm47:
@sstone what's your opinion on edf53ec?

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/controllers/OpenChannelController.scala
```diff
@@ -56,7 +56,7 @@ class OpenChannelController(val handlers: Handlers, val stage: Stage) extends Lo
   @FXML def initialize() = {
     fundingUnit.setItems(Constants.FX_UNITS_ARRAY_NO_MSAT)
     fundingUnit.setValue(FxApp.getUnit.label)
-    feerateField.setText((Globals.feeratesPerKB.get().blocks_6 / 1000).toString)
+    //feerateField.setText((Globals.feeratesPerKB.get().blocks_6 / 1000).toString) TODO: fix this
```
pm47: @dpad85 can you help with this?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
```diff
@@ -231,7 +232,7 @@ object ElectrumWatcher extends App {
     def receive = {
       case ElectrumClient.ElectrumReady(_, _, _) =>
         log.info(s"starting watcher")
-        context become running(context.actorOf(Props(new ElectrumWatcher(client)), "watcher"))
+        context become running(context.actorOf(Props(new ElectrumWatcher(new AtomicLong(0), client)), "watcher"))
```
t-bast: The first property of `ElectrumReady` is the current height, shouldn't you use that instead of `0`?
Overall I must admit I don't understand how this is going to work. If you're using your own atomic `blockCount` instead of the shared one, the `ElectrumWatcher` will never have the correct value in `this.blockCount.get()` :confused:

eclair-core/src/test/scala/fr/acinq/eclair/TestkitBaseClass.scala
```diff
@@ -30,10 +30,6 @@ import scala.concurrent.Await
   */
 abstract class TestkitBaseClass extends TestKit(ActorSystem("test")) with fixture.FunSuiteLike with BeforeAndAfterEach with BeforeAndAfterAll {
```
t-bast: You can remove the `BeforeAndAfterAll` trait

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -69,7 +69,7 @@ class NormalStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     val initialState = alice.stateData.asInstanceOf[DATA_NORMAL]
     val sender = TestProbe()
     val h = randomBytes32
-    val add = CMD_ADD_HTLC(MilliSatoshi(50000000), h, CltvExpiryDelta(144).toCltvExpiry, TestConstants.emptyOnionPacket, upstream = Left(UUID.randomUUID()))
+    val add = CMD_ADD_HTLC(MilliSatoshi(50000000), h, CltvExpiryDelta(144).toCltvExpiry(alice.underlyingActor.nodeParams.currentBlockHeight), TestConstants.emptyOnionPacket, upstream = Left(UUID.randomUUID()))
```
t-bast: It might be a bit clearer if the StateTestsHelperMethods fixture exposed `currentBlockHeight` directly (or maybe `aliceCurrentBlockHeight` and `bobCurrentBlockHeight` to be able to test when their block count is desynchronized).

eclair-core/src/main/scala/fr/acinq/eclair/CltvExpiry.scala
```diff
@@ -49,7 +49,7 @@ case class CltvExpiryDelta(private val underlying: Int) extends Ordered[CltvExpi
   /**
    * Adds the current block height to the given delta to obtain an absolute expiry.
    */
-  def toCltvExpiry = CltvExpiry(Globals.blockCount.get() + underlying)
+  def toCltvExpiry(blockHeight: Long) = CltvExpiry(blockHeight + underlying)
```
t-bast: Crazy idea: maybe we could also define a second helper one like that:
```scala
def toCltvExpiry(implicit nodeParams: NodeParams): CltvExpiry
```

pm47:
@araspitzu I don't think it is going to make a huge difference in terms of overall build time because the longest test will define that, but it can't hurt.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
pm47: I did that at first, then removed it with 2154d98d4d6f3518030ae3f890acab65636a327a because it looked much cleaner. We could still add a `currentBlockHeight` arg with a default value when we need it.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
pm47: This was actually unused code, I removed it in 427f4e7.

eclair-core/src/test/scala/fr/acinq/eclair/TestkitBaseClass.scala
pm47: I can't, we override `afterEach` and `afterAll`.

t-bast: Oh yeah I confused that with one that I added in my amp branch (and that I can remove once this is on master)

eclair-core/src/test/scala/fr/acinq/eclair/channel/FuzzySpec.scala
```diff
@@ -86,100 +86,95 @@ class FuzzySpec extends TestkitBaseClass with StateTestsHelperMethods with Loggi
...
-    system.actorOf(Props(new SenderActor(alice, paymentHandlerB, latch)))
-    system.actorOf(Props(new SenderActor(alice, paymentHandlerB, latch)))
+    val senders = 2
+    val totalMessages = 100
+    val latch = new CountDownLatch(senders)
+    for (_ <- 0 until senders) system.actorOf(Props(new SenderActor(alice, paymentHandlerB, latch, totalMessages / senders)))
     awaitCond(latch.getCount == 0, max = 2 minutes)
     assert(alice.stateName == NORMAL || alice.stateName == OFFLINE)
-    assert(bob.stateName == NORMAL || alice.stateName == OFFLINE)
+    assert(bob.stateName == NORMAL || bob.stateName == OFFLINE)
```
t-bast: Nice catch

eclair-core/src/test/scala/fr/acinq/eclair/channel/FuzzySpec.scala
```diff
@@ -86,100 +86,95 @@ class FuzzySpec extends TestkitBaseClass with StateTestsHelperMethods with Loggi
...
     awaitCond(latch.getCount == 0, max = 2 minutes)
     val sender = TestProbe()
     awaitCond({
       sender.send(alice, CMD_CLOSE(None))
       sender.expectMsgAnyClassOf(classOf[String], classOf[Status.Failure]) == "ok"
     }, max = 30 seconds)
-    senders.foreach(_ ! 'stop)
-    awaitCond(alice.stateName == CLOSING)
-    awaitCond(alice.stateName == CLOSING)
+    awaitCond( {
```
t-bast: Maybe use a one-liner like below?

t-bast:
I've been re-starting the builds several times to "stress-test" them ;)
Looks good so far!

Christewart:
A suggestion with test architecture: 

We have had pretty good success over at [bitcoin-s](https://github.com/bitcoin-s/bitcoin-s) by segregating test cases into separate projects:

If you look at our project setup, you will see there is usually a project, and then a corresponding `project-test`. 

For instance, we have [`eclair-rpc`](https://github.com/bitcoin-s/bitcoin-s/blob/7eb883428e03ac62850ff36525e239587de2dbfa/build.sbt#L34) and [`eclair-rpc-test`](https://github.com/bitcoin-s/bitcoin-s/blob/master/build.sbt#L294). 

What this allows us to do is build a [`testkit`](https://github.com/bitcoin-s/bitcoin-s/blob/master/build.sbt#L342) with all of the src files (such as `eclair-rpc`, [`bitcoind-rpc`](https://github.com/bitcoin-s/bitcoin-s/blob/master/build.sbt#L33)) packaged together in a library to be used as a testing tool for other projects. We actually [publish it as a dependency for others to use](https://oss.sonatype.org/content/repositories/snapshots/org/bitcoin-s/bitcoin-s-testkit_2.12/). 

We use the testkit stuff internally and have found it very useful, and I though I would drop a line for you guys to consider a `eclair-testkit` depedency since this PR seems to be touching a lot of test related configuration. You guys use maven, so i'm not sure how well these things are supported, but it is worth considering :-)

pm47:
Hmm.... now we reproduce the build failure with semaphore (a25fc89).

PR #1116, opened 2019-09-02
Router computes network stats

t-bast:
This PR contains three logically independent commits.

* I first added comments and fixed warnings to graph processing while I was reading the code.
* Then I added a small feature to set the htlcMaximumMsat for routing hints (otherwise the graph processing algorithm used a minimum value which slightly reduced the benefits of those routing hints).
* And finally I'm adding the computation of network statistics to the router: this will be useful for multi-part payments to decide what thresholds should be used to split a payment


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -29,10 +29,21 @@ import scala.collection.mutable
...
   case class RichWeight(cost: MilliSatoshi, length: Int, cltv: CltvExpiryDelta, weight: Double) extends Ordered[RichWeight] {
-    override def compare(that:  RichWeight): Int = this.weight.compareTo(that.weight)
+    override def compare(that: RichWeight): Int = this.weight.compareTo(that.weight)
   }
+  /**
+   * We use heuristics to calculate the weight of an edge based on channel age, cltv delta and capacity.
+   * We favor older channels, with bigger capacity and small cltv delta.
+   */
   case class WeightRatios(cltvDeltaFactor: Double, ageFactor: Double, capacityFactor: Double) {
     require(0 < cltvDeltaFactor + ageFactor + capacityFactor && cltvDeltaFactor + ageFactor + capacityFactor <= 1, "The sum of heuristics ratios must be between 0 and 1 (included)")
```
araspitzu: It's currently possible to set values such as `capacityFactor=1234.5` and `ageFactor=-1234.4` which summed would lie within the interval [0, 1] but effectively screw up the cost function. How about adding a check to enforce individual correctness of the parameters?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -815,18 +837,23 @@ object Router {
...
-
+  def toFakeUpdate(extraHop: ExtraHop, htlcMaximum: MilliSatoshi): ChannelUpdate = {
+    // the `direction` bit in flags will not be accurate but it doesn't matter because it is not used
+    // what matters is that the `disable` bit is 0 so that this update doesn't get filtered out
+    ChannelUpdate(signature = ByteVector64.Zeroes, chainHash = ByteVector32.Zeroes, extraHop.shortChannelId, Platform.currentTime.milliseconds.toSeconds, messageFlags = 1, channelFlags = 0, extraHop.cltvExpiryDelta, htlcMinimumMsat = 0 msat, extraHop.feeBase, extraHop.feeProportionalMillionths, Some(htlcMaximum))
+  }
 
-  def toAssistedChannels(extraRoute: Seq[ExtraHop], targetNodeId: PublicKey): Map[ShortChannelId, AssistedChannel] = {
+  def toAssistedChannels(extraRoute: Seq[ExtraHop], targetNodeId: PublicKey, amount: MilliSatoshi): Map[ShortChannelId, AssistedChannel] = {
     // BOLT 11: "For each entry, the pubkey is the node ID of the start of the channel", and the last node is the destination
```
araspitzu: What's the advantage in computing the `htlcMaximumMsat` ourselves? The path finding would work fine even if there is no `htlcMaximumMsat`.

t-bast: Yes but if there's no `htlcMaximumMsat` the path finding uses the default lower bound, which penalizes the channel whereas we'd like to select those as much as possible since they were explicitly added to the invoice

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -812,17 +812,23 @@ object Router {
...
+    // should be able to route the payment, so we'll compute an htlcMaximumMsat accordingly.
+    // We could also get the channel capacity from the blockchain (since we have the shortChannelId) but that's more expensive.
     val nextNodeIds = extraRoute.map(_.nodeId).drop(1) :+ targetNodeId
-    extraRoute.zip(nextNodeIds).map {
-      case (extraHop: ExtraHop, nextNodeId) => extraHop.shortChannelId -> AssistedChannel(extraHop, nextNodeId)
-    }.toMap
+    extraRoute.zip(nextNodeIds).reverse.foldLeft((amount, Map.empty[ShortChannelId, AssistedChannel])) {
+      case ((amount, acs), (extraHop: ExtraHop, nextNodeId)) =>
+        val nextAmount = amount + nodeFee(extraHop.feeBase, extraHop.feeProportionalMillionths, amount)
+        (nextAmount, acs + (extraHop.shortChannelId -> AssistedChannel(extraHop, nextNodeId, nextAmount.max(RoutingHeuristics.CAPACITY_CHANNEL_LOW))))
```
pm47: I would have done it a bit differently:

```scala
    // BOLT 11: "For each entry, the pubkey is the node ID of the start of the channel", and the last node is the destination
    // The invoice doesn't explicitly specify the channel's htlcMaximumMsat, but we can safely assume that the channel
    // should be able to route the payment, so we'll compute an htlcMaximumMsat accordingly.
    // We could also get the channel capacity from the blockchain (since we have the shortChannelId) but that's more expensive.
    // We also need to make sure the channel isn't excluded by our heuristics
    val lastChannelCapacity = amount.max(RoutingHeuristics.CAPACITY_CHANNEL_LOW)
    val nextNodeIds = extraRoute.map(_.nodeId).drop(1) :+ targetNodeId
    extraRoute.zip(nextNodeIds).reverse.foldLeft((lastChannelCapacity, Map.empty[ShortChannelId, AssistedChannel])) {
      case ((amount, acs), (extraHop: ExtraHop, nextNodeId)) =>
        val nextAmount = amount + nodeFee(extraHop.feeBase, extraHop.feeProportionalMillionths, amount)
        (nextAmount, acs + (extraHop.shortChannelId -> AssistedChannel(extraHop, nextNodeId, nextAmount)))
    }._2
```

pm47: Note: Have you made sure that the comparison with `RoutingHeuristics.CAPACITY_CHANNEL_LOW` isn't strict in our algorithm?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -162,8 +165,7 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
 
   setTimer(TickBroadcast.toString, TickBroadcast, nodeParams.routerConf.routerBroadcastInterval, repeat = true)
   setTimer(TickPruneStaleChannels.toString, TickPruneStaleChannels, 1 hour, repeat = true)
-
-  val SHORTID_WINDOW = 100
+  setTimer(TickNetworkStats.toString, TickNetworkStats, 6 hour, repeat = true)
```
```suggestion
  setTimer(TickNetworkStats.toString, TickNetworkStats, 6 hours, repeat = true)
```

pm47: Should that be configurable?

eclair-core/src/main/scala/fr/acinq/eclair/router/NetworkStats.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import fr.acinq.bitcoin.Satoshi
+import fr.acinq.eclair.wire.ChannelUpdate
+import fr.acinq.eclair.{CltvExpiryDelta, MilliSatoshi}
+
+/**
+ * Created by t-bast on 30/08/2019.
+ */
+
+case class Stats[T](min: T, max: T, median: T, percentile5: T, percentile10: T, percentile25: T, percentile75: T, percentile90: T, percentile95: T)
```
pm47: There may be off-the-shelf libs for that, e.g. `SummaryStatistics` from apache or guava.

Those could have the added advantage to be incremental, we wouldn't have to use a `TickNetworkStats`.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Yes, we're just using it in `htlcMaximumMsat.getOrElse(CAPACITY_CHANNEL_LOW)`

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: It would be a good idea, yeah. I was lazy for that first iteration :)

eclair-core/src/main/scala/fr/acinq/eclair/router/NetworkStats.scala
t-bast: Guava works fine and is more efficient, switching to that!

eclair-core/src/main/scala/fr/acinq/eclair/router/NetworkStats.scala
```diff
@@ -0,0 +1,64 @@
...
+      val cltvStats = Stats(publicChannels.flatMap(pc => getChannelUpdateField(pc, u => u.cltvExpiryDelta.toInt.toLong)), d => CltvExpiryDelta(d.toInt))
+      val feeBaseStats = Stats(publicChannels.flatMap(pc => getChannelUpdateField(pc, u => u.feeBaseMsat.toLong)), d => MilliSatoshi(d.toLong))
+      val feeProportionalStats = Stats(publicChannels.flatMap(pc => getChannelUpdateField(pc, u => u.feeProportionalMillionths)), d => d.toLong)
+      val nodes = publicChannels.flatMap(pc => pc.ann.nodeId1 :: pc.ann.nodeId2 :: Nil).toSet.size
+      Some(NetworkStats(publicChannels.size, nodes, capacityStats, cltvStats, feeBaseStats, feeProportionalStats))
+    }
+  }
+
+  private def getChannelUpdateField[T](pc: PublicChannel, f: ChannelUpdate => T): Seq[T] =
+    pc.update_1_opt.map(u => f(u) :: Nil).getOrElse(Nil) ++ pc.update_2_opt.map(u => f(u) :: Nil).getOrElse(Nil)
```
pm47:
```suggestion
    (pc.update_1_opt.toSeq ++ pc.update_2_opt.toSeq).map(f)
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -136,20 +145,15 @@ case class Data(nodes: Map[PublicKey, NodeAnnouncement],
...
+// @formatter:off
 sealed trait State
-
 case object NORMAL extends State
 
 case object TickBroadcast
-
 case object TickPruneStaleChannels
-
+case object TickNetworkStats
```
pm47: This should be a verb:

```suggestion
case object TickComputeNetworkStats
```

araspitzu:
Is this superseding #1100 ?

t-bast:
> Is this superseding #1100 ?

Not completely because I don't expose it yet to the API.
But #1100 should probably use this instead of computing stats in the API layer indeed ;)

PR #1117, opened 2019-09-02
Api: reject expired invoices

t-bast:
https://github.com/ACINQ/eclair-mobile/issues/209 mentions that we don't check invoice expiry on the sender side.
I'm adding that check at the http API level, but I think it also needs to be added inside eclair-mobile (which directly uses the Java interface if I'm not mistaken).

### DIFF
--- a/eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
+++ b/eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
@@ -31,7 +31,6 @@ import akka.stream.scaladsl.{BroadcastHub, Flow, Keep, Source}
 import akka.stream.{ActorMaterializer, OverflowStrategy}
 import akka.util.Timeout
 import com.google.common.net.HostAndPort
-import de.heikoseeberger.akkahttpjson4s.Json4sSupport
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{ByteVector32, Satoshi}
 import fr.acinq.eclair.api.FormParamExtractors._
@@ -74,6 +73,9 @@ trait Service extends ExtraDirectives with Logging {
   val paramParsingTimeout = 5 seconds
 
   val apiExceptionHandler = ExceptionHandler {
+    case t: IllegalArgumentException =>
+      logger.error(s"API call failed with cause=${t.getMessage}", t)
+      complete(StatusCodes.BadRequest, ErrorResponse(t.getMessage))
     case t: Throwable =>
       logger.error(s"API call failed with cause=${t.getMessage}", t)
       complete(StatusCodes.InternalServerError, ErrorResponse(t.getMessage))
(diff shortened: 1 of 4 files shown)


eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -223,6 +222,7 @@ trait Service extends ExtraDirectives with Logging {
                         } ~
                         path("payinvoice") {
                           formFields(invoiceFormParam, amountMsatFormParam.?, "maxAttempts".as[Int].?, "feeThresholdSat".as[Satoshi].?, "maxFeePct".as[Double].?) {
+                            case (invoice, _, _, _, _) if invoice.isExpired => reject(MalformedFormFieldRejection("invoice", "The invoice provided has expired"))
```
t-bast: This isn't exactly a "malformed" field, but I'm not very familiar with akka http's DSL, do you have a suggestion on what to use there?

araspitzu:
If we add the check in the internal API layer `Eclair.scala` (see `EclairImpl`) then we should be able to address the issue in both the http API and the mobile app. The check there would return a failed `Future` and from the http API this should get translated into an `ErrorResponse` with http status code 400

t-bast:
Yeah but currently `Eclair.scala` doesn't have this information because it doesn't receive the full invoice.
The useful invoice data is extracted before forwarding to `Eclair.scala` (which in my opinion makes sense too).

araspitzu:
Yes it makes sense to extract the data from the invoice here (http layer) but we also lose some bits of information, in this case what's necessary to be fully BOLT compliant. We could add the `expiry` and `timestamp` of the invoice to the `send` internal API or we just leave here but duplicate the check in the mobile app.

dpad85:
@araspitzu Eclair mobile does not use the API to send the payment, instead it directly talks to the `PaymentInitiator`.

We already check the expiration field in the app when initiating the payment. However we do not check it again after that, so if the app is paused for a long time and then resumes at the same point, you'll be able to send the payment.

eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
dpad85: I think `ValidationRejection` or even an `IllegalArgumentException` would be more appropriate, since the field is correctly formed but semantically invalid.

t-bast:
> Eclair mobile does not use the API to send the payment, instead it directly talks to the PaymentInitiator.

Interesting. So in that case we have two choices:

- check expiry in `Service.scala` and in eclair-mobile (and in eclair-mobile that check should be updated to fix https://github.com/ACINQ/eclair-mobile/issues/209)
- add expiry to the parameters required by `PaymentInitiator` and check it there (I feel this shouldn't be `PaymentInitiator`'s responsibility to check that though, this looks like a client pre-condition)

What do you think?

dpad85:
I think we should check the expiry in `Service` and in the client side. For a good UX and payment flow, the UI already has to check payment parameters beforehand without relying on the backend so we will have duplicate code anyway.

araspitzu:
To be frank i'm not fully convinced the API is the proper place to do this check. If eclair-core wants to be fully BOLT compliant than we're putting the check in the wrong place (`Service` is in eclair-node), suppose someone wants to use eclair-core as a base library to interact with the network then it would expect to not be able to pay expired invoices since it's in the [spec](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#payer--payee-requirements)

t-bast:
True, it would make sense that what's inside `eclair-core` is fully spec-compliant.
I'll try modifying slightly the service API to move that check inside `eclair-core`.

PR #1118, opened 2019-09-02
Native CLI

dpad85:
# Description

The current CLI is a bash file, which brings several limitations and makes it hard to maintain. The goal of this pull request is to improve maintainability, bring better error feedback to the user, display exhaustive commands usage, and support correct option syntax.

This PR creates a new side project written in Java with the [picocli](https://github.com/remkop/picocli) library (which brings many features out of the box) and leverages [GraalVM's SubstrateVM](https://www.graalvm.org/docs/reference-manual/aot-compilation/) to generate a single executable binary.

# Example

The code relies heavily on annotations. Each command gets its own class file, which makes the project a bit imposing at first.

```java
@CommandLine.Command(name = "channel", description = "Returns detailed information about a local channel.", sortOptions = false)
public class Channel extends BaseSubCommand {

  @CommandLine.Option(names = { "--channelId", "-c" }, required = true, description = "Id of a channel (32 bytes HexString)")
  String channelId;

  @Override
  public Integer call() throws Exception {
    // http call with okhttp
    return 0;
  }
}
```

### Usage output

```shell
$ ./eclair-cli channel -h

Usage: eclair-cli channel [-hV] -c=<channelId>

Returns detailed information about a local channel.

Options:
  -c, --channelId=<channelId>
                  Id of a channel (32 bytes HexString)
  -h, --help      Show this help message and exit.
  -V, --version   Print version information and exit.
```

Picocli allows for quite a lot of customization regarding the help/usage messages.

### Raw output

The CLI outputs raw JSON by default, which can be piped with `jq`.

```shell
$ ./eclair-cli channel -c 5d86d5536972145e9cf92bcccfeec0aa1653415d165074d4317e88cbb4b194be | jq
{
  "nodeId": "022eb0195aa68832e27ab8fc7b7f509423cf32a9a898e4d5e2cbd84f5ae1c3e520",
  "channelId": "5d86d5536972145e9cf92bcccfeec0aa1653415d165074d4317e88cbb4b194be",
  "state": "NORMAL",
  "data": {
    "commitments": {
      "localParams": {
        "nodeId": "02f716273936cde931f0639965eab2934cdd791630549bc600c780b18f278ad355",
       ...
```

### Pretty output

As a proof of concept, I've added Moshi for json parsing on some calls that output large data sets (this brings some overhead regarding reflection with Graal). We can then print these large data sets into a more human friendly output directly from the CLI, which might be useful.

```shell
$ ./eclair-cli --human-readable channels
You have 1 channels with a total balance of 1,999,980 sat

▸ Peer:    022eb0195aa68832e27ab8fc7b7f509423cf32a9a898e4d5e2cbd84f5ae1c3e520 (1 channels)
  Balance: 1,999,980 sat

            STATE |        BALANCE |       CAPACITY |         SHORT ID |                                                       CHANNEL ID | CHANNEL POINT
           NORMAL |      1,999,980 |      2,100,000 |         5313x1x0 | 5d86d5536972145e9cf92bcccfeec0aa1653415d165074d4317e88cbb4b194be | be94b1b4cb887e31d47450165d415316aac0eecfcc2bf99c5e14726953d5865d:0
```

# Native image generation

The native image is not built directly from Maven (this is possible but would force developers to install GraalVM + SubstrateVM on their current machine which may not be ideal). 

Instead, Maven will build a regular JAR file (with assembly, meaning that dependencies' classes are embedded). On a separate machine/VM, where GraalVM+SubstrateVM is installed, you can then build the actual native file with a simple command.

```shell
$ native-image -jar target/eclair-cli-x.y.z.jar
```

# Notes

- We use Java instead of Scala to limit the size of the native binary file (reaching 30MB already). The code is also very simple since it heavily relies on picocli for the actual CLI logic, so Scala would not bring that much.

- This project is not a module of the eclair project, so it has no impact on the current build. This also prevent packaging the scala-lib dependency (defined in the root pom) into the assembly JAR.

- GraalVM support for Windows is limited and we will most probably not be able to provide a windows .exe CLI for now. 

# Default options

User must provide the API password/url to the CLI. For convenience reasons, he can save these options into a `~/.eclair/eclair-cli.conf` file.


t-bast:
Do we still have bash auto-completion? 
Or do we still need to provide `eclair-cli.bash-completion` (and maybe update it)?

dpad85:
> Do we still have bash auto-completion?
> Or do we still need to provide `eclair-cli.bash-completion` (and maybe update it)?

Picocli can generate an autocompletion script (see https://picocli.info/autocomplete.html) which can also handle options, so we should even improve there. I have not tested it yet.

eclair-cli/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>fr.acinq.eclair</groupId>
+    <artifactId>eclair-cli</artifactId>
+    <version>1.0.0</version>
```
t-bast: Shouldn't we align version numbers with `eclair`?
Can't we inherit that from the parent?

eclair-cli/src/main/java/fr/acinq/eclair/cli/payment/PayInvoice.java
```diff
@@ -0,0 +1,72 @@
...
+import fr.acinq.eclair.cli.utils.BaseSubCommand;
+import okhttp3.ResponseBody;
+import picocli.CommandLine;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static fr.acinq.eclair.cli.utils.Utils.print;
+
+@CommandLine.Command(name = "payinvoice", description = "Deserialize a BOLT-11 invoice into human readable data.", sortOptions = false)
```
t-bast: nit: wrong copy-paste, this should be "Pay a BOLT-11 invoice"

eclair-cli/src/main/java/fr/acinq/eclair/cli/payment/SendToRoute.java
```diff
@@ -0,0 +1,69 @@
...
+import okhttp3.ResponseBody;
+import picocli.CommandLine;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static fr.acinq.eclair.cli.utils.Utils.print;
+
+@CommandLine.Command(name = "sendtoroute", description = "Sends money to a node without an invoice. The payment will be retried several times if needed.", sortOptions = false)
```
t-bast: nit: wrong copy-paste: should be "sends money to a user-defined route without an invoice"

eclair-cli/src/main/resources/messages.properties
```diff
@@ -0,0 +1,27 @@
...
+#
+
+opts.node_id = Id of the node (33 bytes HexString)
+opts.channel_id = Id of a channel (32 bytes HexString)
+opts.short_channel_id = Short id of a channel (<block_height>x<tx_index>x<output_index>)
+opts.invoice = BOLT-11 invoice
+opts.payment_hash = Payment hash for this payment (32 bytes HexString)
+opts.payment_uuid = Payment attempt unique identifier
+opts.payment.amount_msat_send = Amount to send, in millisatoshi
+opts.payment.max_retries = Payment hash for this payment (32 bytes HexString)
```
t-bast: nit: should be "number of times to retry the payment"

eclair-cli/pom.xml
dpad85: The project is not a child module of the `eclair` project for technical reasons:

> has no impact on the current build [...] also prevent packaging the scala-lib dependency (defined in the root pom) into the assembly JAR.

But this can probably be solved by tweaking the assembly plugin configuration.

t-bast: > The project is not a child module of the eclair project

Yeah that makes sense, but I don't know how we can keep their versions aligned...
Maybe we don't really care though :)

eclair-cli/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+
+    <name>eclair-cli</name>
+
+    <properties>
+        <!-- base -->
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
+        <!-- deps -->
+        <graal.version>19.2.0</graal.version>
```
araspitzu: nit: unused

dpad85:
If we want to make a better CLI, we should rather use something based on Kotlin MP.

PR #1120, opened 2019-09-03
Fix maven mirror

t-bast:
We used a mirror in France, which doesn't make sense since Travis builds run on VM in GCP US.
Also upgraded to maven 3.6.2.


t-bast:
I think that as long as we stay in the 3.6.* there's no need to update our BUILD.md, these versions should be compatible.
If we upgrade to 3.7.* or 4.* (if those are released at some point) then we'll update instructions.

PR #1122, opened 2019-09-03
Store channel data on restore

pm47:
This may seem odd, but it is necessary in some cases when we have a data
format change that involves creating fields that contain a timestamp.

If we don't store that data, we will generate a new version of the data
at every restart, with a more recent timestamp.

It is the case for the `DATA_WAIT_FOR_FUNDING_CONFIRMED_COMPAT_01_Codec`
codec, or `DATA_CLOSING_COMPAT_06_Codec` codec, which results in
channels never been cleaned up.


pm47:
So this has effectively fixed a case of a channel never being cleaned up on Eclair Mobile, but I'm still unsure about storing data in the `WAIT_FOR_INIT_INTERNAL` handler. Seems hackish and we could reasonably expect this state to not change the data. WDYT @sstone?

sstone:
First I though we could add a `wasUpgraded` or `needsSaving` field to DATA instances and have this field set to `true` in the "legacy" codec but it quickly becomes a mess.
Knowing which codec was actually used to deserialize channel data would also help but it's handled in the channel db so it would be also clumsy to retrieve this info (and it breaks encapsulation). So even if the proposed solution here is hacky I can't think of a better one yet...

pm47:
The more I think of this, the less comfortable I am with rewriting all channels every we restart startup. We need to find something better.

PR #1124, opened 2019-09-04
Made sync params configurable

pm47:
This allows us to choose smaller parameters for tests and reduce cpu
requirement during testing.


eclair-core/src/main/resources/reference.conf
```diff
@@ -121,6 +121,8 @@ eclair {
     sync {
       request-node-announcements = true // if true we will ask for node announcements when we receive channel ids that we don't know
       encoding-type = zlib // encoding for short_channel_ids and timestamps in query channel sync messages; other possible value is "uncompressed"
+      channel-range-chunk-size = 3500 // how many short channel ids we are going to answer at once; we can theoretically fit 4091 uncompressed channel ids in a single lightning message (max size 65 Kb)
```
pm47: I'm not satisfied with the naming of both parameters. If anyone has a better idea...

eclair-core/src/main/resources/reference.conf
```diff
@@ -121,8 +121,8 @@ eclair {
     sync {
       request-node-announcements = true // if true we will ask for node announcements when we receive channel ids that we don't know
       encoding-type = zlib // encoding for short_channel_ids and timestamps in query channel sync messages; other possible value is "uncompressed"
-      channel-range-chunk-size = 3500 // how many short channel ids we are going to answer at once; we can theoretically fit 4091 uncompressed channel ids in a single lightning message (max size 65 Kb)
-      short-id-window = 100 // how many channels are we going to request at once
+      channel-range-chunk-size = 2500 // do not change this unless you know what you are doing
```
t-bast: Why not also keep the previous message? I think it was useful

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1070,15 +1069,14 @@ object Router {
    * @param shortChannelIds
    * @return
    */
-  def split(shortChannelIds: SortedSet[ShortChannelId]): List[ShortChannelIdsChunk] = {
+  def split(shortChannelIds: SortedSet[ShortChannelId], channelRangeChunkSize: Int): List[ShortChannelIdsChunk] = {
     // this algorithm can split blocks (meaning that we can in theory generate several replies with the same first_block/num_blocks
     // and a different set of short_channel_ids) but it doesn't matter
-    val SPLIT_SIZE = 3500 // we can theoretically fit 4091 uncompressed channel ids in a single lightning message (max size 65 Kb)
```
pm47: @sstone pretty sure this calculation was wrong, can you reproduce it?

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -246,6 +246,8 @@ object NodeParams {
         randomizeRouteSelection = config.getBoolean("router.randomize-route-selection"),
         requestNodeAnnouncements = config.getBoolean("router.sync.request-node-announcements"),
         encodingType = routerSyncEncodingType,
+        channelRangeChunkSize = config.getInt("router.sync.channel-range-chunk-size"),
```
t-bast: I think it's also good to add a require on the max value (like you did in the first commit), isn't it?

pm47: Yres, but for that we have to agree on what are the actual max values, and it has to be maintained. Given that this is very low level, and that we strongly discourage users from changing it, I'm not sure it is worth it to change.

t-bast: Fair enough

eclair-core/src/main/resources/reference.conf
```diff
@@ -121,6 +121,8 @@ eclair {
     sync {
       request-node-announcements = true // if true we will ask for node announcements when we receive channel ids that we don't know
       encoding-type = zlib // encoding for short_channel_ids and timestamps in query channel sync messages; other possible value is "uncompressed"
+      channel-range-chunk-size = 2500 // do not change this unless you know what you are doing
```
```suggestion
      channel-range-chunk-size = 2500 // max number of channel ids (+ timestamps + checksums) in reply_channel_range *do not change this unless you know what you are doing*
```

eclair-core/src/main/resources/reference.conf
```diff
@@ -121,6 +121,8 @@ eclair {
...
+      channel-query-chunk-size = 100 // do not change this unless you know what you are doing
```
```suggestion
      channel-query-chunk-size = 100 // max number of short_channel_ids in query_short_channel_ids *do not change this unless you know what you are doing*
```

eclair-core/src/main/resources/reference.conf
araspitzu: What's the rationale for changing the default to from 3500 to 2500? Is it improving performances?

pm47: This is related to:

> @sstone pretty sure this calculation was wrong, can you reproduce it?

The max value was wrong because we didn't take timestamps+checksums into account. It is somewhere around 2700.

PR #1125, opened 2019-09-05
Add Semaphore CI

pm47:


pm47:
@araspitzu I have no idea how to configure scoverage on semaphore. Docs look cryptic...

PR #1126, opened 2019-09-05
Add monitoring with Kamon (disabled by default)

pm47:
For now:
- we only track some tasks (especially in the router, but not even
`node_announcement` and `channel_update`)
- all db calls are monitored


eclair-core/src/main/scala/fr/acinq/eclair/PimpKamon.scala
```diff
@@ -0,0 +1,25 @@
...
+      f
+    } finally {
+      timer.stop()
+    }
+  }
+
+  def timeFuture[T](name: String)(f: => Future[T])(implicit ec: ExecutionContext): Future[T] = {
+    val timer = Kamon.timer(name).withoutTags().start()
+    val res = f
+    res onComplete { case _ => timer.stop }
```
araspitzu: Shouldn't this be `res onComplete { case Success(_) => timer.stop }`? I'd assume we don't want to mix measurements for failed calls with the successful ones.

pm47: We need to stop the timer somehow? I don't think it matters anyway, because Kamon has first-class handling of distribution of latencies, and we would detect outliers easily.

pm47:
I'm marking this as reviewable to minimize conflicts with other branches. It is just a first step.

Thanks @ivantopo for all the support!

eclair-core/src/main/scala/fr/acinq/eclair/PimpKamon.scala
```diff
@@ -0,0 +1,25 @@
+package fr.acinq.eclair
```
t-bast: nit: no copyright header?

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -40,6 +42,9 @@ object Boot extends App with Logging {
     implicit val system: ActorSystem = ActorSystem("eclair-node")
     implicit val ec: ExecutionContext = system.dispatcher
     val setup = new Setup(datadir)
+
+    Kamon.init(setup.appConfig)
```
pm47:
```suggestion
    // TODO: disable Kamon for now
    //Kamon.init(setup.appConfig)
```

eclair-core/src/main/scala/fr/acinq/eclair/PimpKamon.scala
t-bast: I think that's where we'd use tags or some equivalent. 
Usually we should split our measurements on values of a specific tag with Success, Failure1, Failure2, etc depending on our set of possible failures.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -149,26 +150,36 @@ class ExtendedBitcoinClient(val rpcClient: BitcoinJsonRPCClient) {
...
-      unspent <- isTransactionOutputSpendable(txid, outputIndex, includeMempool = true)
-      fundingTxStatus <- if (unspent) {
-        Future.successful(UtxoStatus.Unspent)
-      } else {
-        // if this returns true, it means that the spending tx is *not* in the blockchain
-        isTransactionOutputSpendable(txid, outputIndex, includeMempool = false).map {
-          case res => UtxoStatus.Spent(spendingTxConfirmed = !res)
+      val span = Kamon.spanBuilder("validate-bitcoin-client").start()
+      for {
+        _ <- Future.successful(0)
```
t-bast: Why is that needed?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -149,26 +150,36 @@ class ExtendedBitcoinClient(val rpcClient: BitcoinJsonRPCClient) {
...
-      }
-      tx <- getRawTransaction(txid)
-      unspent <- isTransactionOutputSpendable(txid, outputIndex, includeMempool = true)
-      fundingTxStatus <- if (unspent) {
-        Future.successful(UtxoStatus.Unspent)
-      } else {
-        // if this returns true, it means that the spending tx is *not* in the blockchain
-        isTransactionOutputSpendable(txid, outputIndex, includeMempool = false).map {
-          case res => UtxoStatus.Spent(spendingTxConfirmed = !res)
+      val span = Kamon.spanBuilder("validate-bitcoin-client").start()
```
t-bast: What's kamon recommendation for naming spans? Do they automatically included the containing namespace/class/object? It's usually recommended to use some kind of namespacing in spans, so either Kamon does it automatically for us or maybe we should do something like `blockchain.bitcoind.rpc.extendedbitcoinclient.validate`?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -149,26 +150,36 @@ class ExtendedBitcoinClient(val rpcClient: BitcoinJsonRPCClient) {
...
+      for {
+        _ <- Future.successful(0)
+        span0 = Kamon.spanBuilder("getblockhash").start()
```
t-bast: Is that automagically included as a child span of the `span` defined line 153?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -149,26 +150,36 @@ class ExtendedBitcoinClient(val rpcClient: BitcoinJsonRPCClient) {
...
+        txid: String <- rpcClient.invoke("getblock", blockHash).map {
+          case json => Try {
+            val JArray(txs) = json \ "tx"
+            txs(txIndex).extract[String]
+          } getOrElse ByteVector32.Zeroes.toHex
         }
-      }
-    } yield ValidateResult(c, Right((Transaction.read(tx), fundingTxStatus)))
+        _ = span1.finish()
+        span2 = Kamon.spanBuilder("getrawtx").start()
```
t-bast: It looks like we'd benefit a lot from a utility function in KamonExt to wrap a future in a span.
It would be nice to just write `tx <- KamonExt.wrapInSpan("getrawtx")(getRawTransaction(txid))` which would automatically do the `span.finish()`.
The name `wrapInSpan` isn't very good, needs some bikeshedding.

eclair-core/src/main/scala/fr/acinq/eclair/io/Authenticator.scala
```diff
@@ -43,6 +44,7 @@ class Authenticator(nodeParams: NodeParams) extends Actor with DiagnosticActorLo
   def ready(switchboard: ActorRef, authenticating: Map[ActorRef, PendingAuth]): Receive = {
     case pending@PendingAuth(connection, remoteNodeId_opt, address, _) =>
       log.debug(s"authenticating connection to ${address.getHostString}:${address.getPort} (pending=${authenticating.size} handlers=${context.children.size})")
+      Kamon.counter("peers.connecting.count").withTag("state", "authenticating").increment()
```
t-bast: It might be useful to centralize counters for each component (like the `Metrics` object you created for `Peer`).
For example in Authenticator's companion object, define `private val connectingPeersCounter = Kamon.counter("peers.connecting.count")`.
This is usually easier for future maintenance.

eclair-core/src/main/scala/fr/acinq/eclair/io/Server.scala
```diff
@@ -52,6 +53,7 @@ class Server(nodeParams: NodeParams, authenticator: ActorRef, address: InetSocke
   def listening(listener: ActorRef): Receive = {
     case Connected(remote, _) =>
       log.info(s"connected to $remote")
+      Kamon.counter("peers.connecting.count").withTag("state", "connected").increment()
```
t-bast: This is the same counter used in `Authenticator`. So definitely worth centralizing somewhere to make sure these two actors keep using the same counter correctly

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -520,7 +536,11 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
         stay
       } else {
         log.info("validating shortChannelId={}", c.shortChannelId)
-        watcher ! ValidateRequest(c)
+        Kamon.runWithContextEntry(shortChannelIdKey, c.shortChannelId) {
+          Kamon.runWithSpan(Kamon.spanBuilder("validate-channel").tag("shortChannelId", c.shortChannelId.toString).start(), finishSpan = false) {
```
t-bast: How will that span be stopped? I'm not sure it's useful to start it here since we're just sending a message to another actor, it's more useful to do all the spanning inside the watcher isn't it?

eclair-node/pom.xml
```diff
@@ -102,6 +120,32 @@
...
+            <groupId>io.kamon</groupId>
+            <artifactId>kamon-jdbc_${scala.version.short}</artifactId>
+            <version>${kamon.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.kamon</groupId>
+            <artifactId>kamon-annotation_${scala.version.short}</artifactId>
+            <version>2.0.0-RC1</version>
+        </dependency>
+        <!--dependency>
```
t-bast: Maybe remove entirely? I don't expect it to be useful to monitor akka-http since we only use it for the RPC API...

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -40,6 +42,10 @@ object Boot extends App with Logging {
     implicit val system: ActorSystem = ActorSystem("eclair-node")
     implicit val ec: ExecutionContext = system.dispatcher
     val setup = new Setup(datadir)
+
+    // TODO: disable Kamon for now
+    //Kamon.init(setup.appConfig)
```
t-bast: Might be worth making this configurable?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
pm47: It's just a limitation of the `for` construct.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
pm47: Yes

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
t-bast: Right, got it.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
pm47: That's a good point, let's address this in a 2nd iteration

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: It is stopped in the `ValidateResult` handler. The point is to be able to measure time spent in the queue, which I think is important to measure backlogs.

eclair-node/pom.xml
pm47: I think monitoring the API is still useful, and I'm waiting for a compatible version.

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
pm47: Yes, I was just being lazy

pm47:
@t-bast I addressed the most pressing points. I think we can take care of the rest in a later iteration?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Interesting, got it.

eclair-node/pom.xml
```diff
@@ -102,6 +120,32 @@
...
+            <groupId>io.kamon</groupId>
+            <artifactId>kamon-system-metrics_${scala.version.short}</artifactId>
+            <version>${kamon.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.kamon</groupId>
+            <artifactId>kamon-jdbc_${scala.version.short}</artifactId>
+            <version>${kamon.version}</version>
+        </dependency>
+        <dependency>
```
pm47:
```suggestion
        <!--dependency>
```

eclair-node/pom.xml
```diff
@@ -102,6 +120,32 @@
...
+            <groupId>io.kamon</groupId>
+            <artifactId>kamon-annotation_${scala.version.short}</artifactId>
+            <version>2.0.0-RC1</version>
+        </dependency>
```
pm47:
```suggestion
        </dependency-->
```

PR #1127, opened 2019-09-05
Removed Globals class

pm47:
This is a prerequisite to parallelization of tests.


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
```diff
@@ -37,7 +37,7 @@ class PaymentInitiator(nodeParams: NodeParams, router: ActorRef, register: Actor
     case p: PaymentInitiator.SendPaymentRequest =>
       val paymentId = UUID.randomUUID()
       // We add one block in order to not have our htlc fail when a new block has just been found.
-      val finalExpiry = (p.finalExpiryDelta + 1).toCltvExpiry
+      val finalExpiry = (p.finalExpiryDelta + 1).toCltvExpiry(nodeParams.currentBlockHeight)
```
t-bast: nit: it's a little bit clearer like that imo:
```suggestion
      val finalExpiry = p.finalExpiryDelta.toCltvExpiry(nodeParams.currentBlockHeight + 1)
```

eclair-core/src/test/scala/fr/acinq/eclair/TestUtils.scala
```diff
@@ -27,4 +28,16 @@ object TestUtils {
     .props
     .get("buildDirectory") // this is defined if we run from maven
     .getOrElse(new File(sys.props("user.dir"), "target").getAbsolutePath) // otherwise we probably are in intellij, so we build it manually assuming that user.dir == path to the module
+
+  def availablePort: Int = synchronized {
```
t-bast: This change is to prepare parallelization, right?

eclair-core/src/test/scala/fr/acinq/eclair/channel/FuzzySpec.scala
```diff
@@ -81,107 +81,101 @@ class FuzzySpec extends TestkitBaseClass with StateTestsHelperMethods with Loggi
       bob ! WatchEventConfirmed(BITCOIN_FUNDING_DEPTHOK, 400000, 42, fundingTx)
       alice2blockchain.expectMsgType[WatchLost]
       bob2blockchain.expectMsgType[WatchLost]
-      awaitCond(alice.stateName == NORMAL)
-      awaitCond(bob.stateName == NORMAL)
+      awaitCond(alice.stateName == NORMAL, 1 minute)
```
t-bast: 1 minute looks huge!

eclair-core/src/test/scala/fr/acinq/eclair/channel/FuzzySpec.scala
```diff
@@ -81,107 +81,101 @@ class FuzzySpec extends TestkitBaseClass with StateTestsHelperMethods with Loggi
...
   test("fuzzy test with both parties sending HTLCs", Tag("fuzzy")) { f =>
     import f._
-    val latch = new CountDownLatch(100)
-    system.actorOf(Props(new SenderActor(alice, paymentHandlerB, latch)))
-    system.actorOf(Props(new SenderActor(alice, paymentHandlerB, latch)))
-    system.actorOf(Props(new SenderActor(bob, paymentHandlerA, latch)))
-    system.actorOf(Props(new SenderActor(bob, paymentHandlerA, latch)))
+    val senders = 2
+    val totalMessages = 100
+    val latch = new CountDownLatch(senders)
```
t-bast: Shouldn't the latch be `2*senders`?
Since you create `senders` actors with `alice` and `senders` actors with `bob`.

t-bast: Maybe `totalMessages` should also be halved (not sure)

t-bast: Or like you did in the other test, separate in two latches.

eclair-core/src/test/scala/fr/acinq/eclair/TestUtils.scala
pm47: Yes, we now use a different set of ports each time we need to spawn bitcoind, in order to be able to run them concurrently.

eclair-core/src/test/scala/fr/acinq/eclair/channel/FuzzySpec.scala
pm47: Yes, was kind of a desperate measure. I should be able to reduce/remove once I figure out the root cause.

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -794,7 +795,27 @@ class IntegrationSpec extends TestKit(ActorSystem("test")) with BitcoindService
     forwardHandlerC.forward(buffer.ref)
     sigListener.expectMsgType[ChannelSignatureReceived]
     send(130000000 msat, paymentHandlerC, nodes("F5").paymentInitiator)
-    forwardHandlerC.expectMsgType[UpdateAddHtlc]
+    Try(forwardHandlerC.expectMsgType[UpdateAddHtlc]).recover {
```
t-bast: This is temporary, right? It seems that after recovering you're not re-throwing the failure so the test might wrongly pass. But I guess you'll remove this once the investigation is over, so not important.

pm47: Yes it's a horrible hack

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -794,7 +796,10 @@ class IntegrationSpec extends TestKit(ActorSystem("test")) with BitcoindService
     forwardHandlerC.forward(buffer.ref)
     sigListener.expectMsgType[ChannelSignatureReceived]
     send(130000000 msat, paymentHandlerC, nodes("F5").paymentInitiator)
+    val t1 = Platform.currentTime.milliseconds
     forwardHandlerC.expectMsgType[UpdateAddHtlc]
```
pm47:
```suggestion
    forwardHandlerC.expectMsgType[UpdateAddHtlc](1 hour)
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/zmq/ZMQActor.scala
```diff
@@ -80,10 +80,12 @@ class ZMQActor(address: String, connected: Option[Promise[Done]] = None) extends
     case event: Event => event.getEvent match {
       case ZMQ.EVENT_CONNECTED =>
         log.info(s"connected to ${event.getAddress}")
+        println(s"connected to ${event.getAddress}")
```
araspitzu: Is this temporary?

eclair-core/src/test/resources/integration/bitcoin.conf
```diff
@@ -1,5 +1,5 @@
 regtest=1
-noprinttoconsole=1 
+noprinttoconsole=0
```
araspitzu: Is this option still iavailable? I can't find it in the [bitcoin.conf](https://github.com/bitcoin/bitcoin/blob/v0.17.1/share/examples/bitcoin.conf)

eclair-core/src/test/scala/fr/acinq/eclair/TestUtils.scala
```diff
@@ -27,4 +28,18 @@ object TestUtils {
...
+    var serverSocket: ServerSocket = null
+    try {
+      serverSocket = new ServerSocket(0)
+      val port = serverSocket.getLocalPort
+      println(s"attributing port=$port")
```
araspitzu: nit: replace with log entry

eclair-core/src/test/resources/integration/bitcoin.conf
pm47: Apparently it works...

PR #1128, opened 2019-09-06
Check funds in millisatoshi when sending/receiving an HTLC

pm47:
Instead of satoshi, which could introduce rounding errors.

Also, we check first the balance before the max-inflight amount, because
it makes more sense in terms of error management.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -152,7 +152,7 @@ object Commitments {
     // we look from remote's point of view, so if local is funder remote doesn't pay the fees
     val fees = if (commitments1.localParams.isFunder) commitTxFee(commitments1.remoteParams.dustLimit, reduced) else 0.sat
     val missing = reduced.toRemote - commitments1.remoteParams.channelReserve - fees
-    if (missing < 0.sat) {
+    if (missing < 0.msat) {
       return Left(InsufficientFunds(commitments.channelId, amount = cmd.amount, missing = -missing.truncateToSatoshi, reserve = commitments1.remoteParams.channelReserve, fees = fees))
     }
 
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -148,6 +148,14 @@ object Commitments {
     // the HTLC we are about to create is outgoing, but from their point of view it is incoming
     val outgoingHtlcs = reduced.htlcs.filter(_.direction == IN)
 
+    // a node cannot spend pending incoming htlcs, and need to keep funds above the reserve required by the counterparty, after paying the fee
+    // we look from remote's point of view, so if local is funder remote doesn't pay the fees
+    val fees = if (commitments1.localParams.isFunder) commitTxFee(commitments1.remoteParams.dustLimit, reduced) else 0.sat
+    val missing = reduced.toRemote - commitments1.remoteParams.channelReserve - fees
+    if (missing < 0.sat) {
```
t-bast: nit:
```suggestion
    if (missing < 0.msat) {
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -183,6 +183,13 @@ object Commitments {
     val reduced = CommitmentSpec.reduce(commitments1.localCommit.spec, commitments1.localChanges.acked, commitments1.remoteChanges.proposed)
     val incomingHtlcs = reduced.htlcs.filter(_.direction == IN)
 
+    // a node cannot spend pending incoming htlcs, and need to keep funds above the reserve required by the counterparty, after paying the fee
+    val fees = if (commitments1.localParams.isFunder) 0.sat else Transactions.commitTxFee(commitments1.localParams.dustLimit, reduced)
+    val missing = reduced.toRemote - commitments1.localParams.channelReserve - fees
+    if (missing < 0.sat) {
```
t-bast: nit: same:
```suggestion
    if (missing < 0.msat) {
```

PR #1074, opened 2019-07-14
Store failure messages for outgoing payments

rorp:
Currently there's no way to get payment failures using `getsentinfo` RPC. This PR addresses that.

PR #1130, opened 2019-09-06
Payment lifecycle refactoring

t-bast:
* Unify payment events (we really didn't need two separate type hierarchies, it became messy once `PaymentResult` started being sent to the `eventStream`)
* Move interactions with the DB and eventStream out of `PaymentLifecycle`: this paves the way for payments-that-are-not-really-payments for which we don't want to store in the DB and don't want to emit events (sub-payments in AMP or relayed trampoline payments for example)
* Remove information from AuditDB:
  * channel id won't make sense anymore in an AMP world and wasn't used
* Add more information to the PaymentDB:
  * external id (to help lightning apps reconcile with their own DB)
  * parent id (AMP)
  * target node ID
  * bolt 11 invoice
  * fees paid
  * payment route in case of success
  * payment failures
* Both Payment DB and Audit DB have a new version and migration

In particular, this supercedes #1048, #1070, #1074 and #1099

Here is what a payment success looks like:
payment_success

And here is a payment failure:
payment_failure

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -78,13 +78,13 @@ trait Eclair {
 
   def receivedInfo(paymentHash: ByteVector32)(implicit timeout: Timeout): Future[Option[IncomingPayment]]
 
-  def send(externalId: Option[UUID], recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest] = None, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
+  def send(externalId_opt: Option[String], recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest] = None, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
 
   def sentInfo(id: Either[UUID, ByteVector32])(implicit timeout: Timeout): Future[Seq[OutgoingPayment]]
 
   def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty)(implicit timeout: Timeout): Future[RouteResponse]
 
-  def sendToRoute(externalId: Option[UUID], route: Seq[PublicKey], amount: MilliSatoshi, paymentHash: ByteVector32, finalCltvExpiryDelta: CltvExpiryDelta)(implicit timeout: Timeout): Future[UUID]
+  def sendToRoute(externalId_opt: Option[String], route: Seq[PublicKey], amount: MilliSatoshi, paymentHash: ByteVector32, finalCltvExpiryDelta: CltvExpiryDelta)(implicit timeout: Timeout): Future[UUID]
 
   def audit(from_opt: Option[Long], to_opt: Option[Long])(implicit timeout: Timeout): Future[AuditResponse]
 
@@ -112,6 +112,9 @@ trait Eclair {
 class EclairImpl(appKit: Kit) extends Eclair {
 
   implicit val ec: ExecutionContext = appKit.system.dispatcher
+
+  // We constrain external identifiers. This allows uuid, long and pubkey to be used.
+  private val externalIdMaxLength = 66
 
   override def connect(target: Either[NodeURI, PublicKey])(implicit timeout: Timeout): Future[String] = target match {
     case Left(uri) => (appKit.switchboard ? Peer.Connect(uri)).mapTo[String]
@@ -186,30 +189,35 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.router ? RouteRequest(appKit.nodeParams.nodeId, targetNodeId, amount, assistedRoutes)).mapTo[RouteResponse]
   }
 
-  override def sendToRoute(externalId: Option[UUID], route: Seq[PublicKey], amount: MilliSatoshi, paymentHash: ByteVector32, finalCltvExpiryDelta: CltvExpiryDelta)(implicit timeout: Timeout): Future[UUID] = {
-    (appKit.paymentInitiator ? SendPaymentRequest(amount, paymentHash, route.last, 1, finalCltvExpiryDelta, None, externalId, route)).mapTo[UUID]
+  override def sendToRoute(externalId_opt: Option[String], route: Seq[PublicKey], amount: MilliSatoshi, paymentHash: ByteVector32, finalCltvExpiryDelta: CltvExpiryDelta)(implicit timeout: Timeout): Future[UUID] = {
+    externalId_opt match {
+      case Some(externalId) if externalId.length > externalIdMaxLength => Future.failed(new IllegalArgumentException("externalId is too long: cannot exceed 66 characters"))
+      case _ => (appKit.paymentInitiator ? SendPaymentRequest(amount, paymentHash, route.last, 1, finalCltvExpiryDelta, None, externalId_opt, route)).mapTo[UUID]
+    }
   }
 
-  override def send(externalId: Option[UUID], recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest], maxAttempts_opt: Option[Int], feeThreshold_opt: Option[Satoshi], maxFeePct_opt: Option[Double])(implicit timeout: Timeout): Future[UUID] = {
+  override def send(externalId_opt: Option[String], recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest], maxAttempts_opt: Option[Int], feeThreshold_opt: Option[Satoshi], maxFeePct_opt: Option[Double])(implicit timeout: Timeout): Future[UUID] = {
     val maxAttempts = maxAttempts_opt.getOrElse(appKit.nodeParams.maxPaymentAttempts)
-
     val defaultRouteParams = Router.getDefaultRouteParams(appKit.nodeParams.routerConf)
     val routeParams = defaultRouteParams.copy(
       maxFeePct = maxFeePct_opt.getOrElse(defaultRouteParams.maxFeePct),
       maxFeeBase = feeThreshold_opt.map(_.toMilliSatoshi).getOrElse(defaultRouteParams.maxFeeBase)
     )
 
-    invoice_opt match {
-      case Some(invoice) if invoice.isExpired => Future.failed(new IllegalArgumentException("invoice has expired"))
-      case Some(invoice) =>
-        val sendPayment = invoice.minFinalCltvExpiryDelta match {
-          case Some(minFinalCltvExpiryDelta) => SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts, minFinalCltvExpiryDelta, invoice_opt, externalId, assistedRoutes = invoice.routingInfo, routeParams = Some(routeParams))
-          case None => SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts, paymentRequest = invoice_opt, externalId = externalId, assistedRoutes = invoice.routingInfo, routeParams = Some(routeParams))
-        }
-        (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
-      case None =>
-        val sendPayment = SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts = maxAttempts, externalId = externalId, routeParams = Some(routeParams))
-        (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
+    externalId_opt match {
+      case Some(externalId) if externalId.length > externalIdMaxLength => Future.failed(new IllegalArgumentException("externalId is too long: cannot exceed 66 characters"))
+      case _ => invoice_opt match {
+        case Some(invoice) if invoice.isExpired => Future.failed(new IllegalArgumentException("invoice has expired"))
+        case Some(invoice) =>
+          val sendPayment = invoice.minFinalCltvExpiryDelta match {
+            case Some(minFinalCltvExpiryDelta) => SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts, minFinalCltvExpiryDelta, invoice_opt, externalId_opt, assistedRoutes = invoice.routingInfo, routeParams = Some(routeParams))
+            case None => SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts, paymentRequest = invoice_opt, externalId = externalId_opt, assistedRoutes = invoice.routingInfo, routeParams = Some(routeParams))
+          }
+          (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
+        case None =>
+          val sendPayment = SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts = maxAttempts, externalId = externalId_opt, routeParams = Some(routeParams))
+          (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
+      }
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -78,13 +78,13 @@ trait Eclair {
 
   def receivedInfo(paymentHash: ByteVector32)(implicit timeout: Timeout): Future[Option[IncomingPayment]]
 
-  def send(recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest] = None, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
+  def send(externalId: Option[UUID], recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest] = None, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
 
   def sentInfo(id: Either[UUID, ByteVector32])(implicit timeout: Timeout): Future[Seq[OutgoingPayment]]
 
   def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty)(implicit timeout: Timeout): Future[RouteResponse]
 
-  def sendToRoute(route: Seq[PublicKey], amount: MilliSatoshi, paymentHash: ByteVector32, finalCltvExpiryDelta: CltvExpiryDelta)(implicit timeout: Timeout): Future[UUID]
+  def sendToRoute(externalId: Option[UUID], route: Seq[PublicKey], amount: MilliSatoshi, paymentHash: ByteVector32, finalCltvExpiryDelta: CltvExpiryDelta)(implicit timeout: Timeout): Future[UUID]
 
   def audit(from_opt: Option[Long], to_opt: Option[Long])(implicit timeout: Timeout): Future[AuditResponse]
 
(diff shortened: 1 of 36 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -182,17 +178,44 @@ class PaymentLifecycle(nodeParams: NodeParams, id: UUID, router: ActorRef, regis
...
-  def props(nodeParams: NodeParams, id: UUID, router: ActorRef, register: ActorRef) = Props(classOf[PaymentLifecycle], nodeParams, id, router, register)
+  def props(nodeParams: NodeParams, progressHandler: PaymentProgressHandler, router: ActorRef, register: ActorRef) = Props(classOf[PaymentLifecycle], nodeParams, progressHandler, router, register)
+
+  /** This handler notifies other components of payment progress. */
+  trait PaymentProgressHandler {
+    val id: UUID
+
+    // @formatter:off
+    def onSend(paymentHash: ByteVector32, finalAmount: MilliSatoshi): Unit
+    def onSucceed(sender: ActorRef, eventStream: EventStream, result: PaymentSent): Unit
```
pm47:
```suggestion
    def onSucceed(sender: ActorRef, result: PaymentSent)(ctx: ActorContext): Unit
```

t-bast:
After offline discussion, I think I'll make the following changes:

- remove payment failed events from audit DB (or should I keep them?)
- add payment failures to the payment DB
- convert `Hop` to a simpler `HopSummary(nodeId, nextNodeId, channelId)` (we don't need all the channel updates in the DB when we're only interested in logging the route when a payment succeeded) -> do it in a cleaner way than currently
- do the same trick for failures (convert to a lighter `FailureSummary` to be stored in DB)

Since DB changes require migrations and careful (painful?) testing, are we ok with that plan?

araspitzu:
Nice changes, especially storing the route for a successful payment (can be used later for path finding heuristics). On the payment events: is it possible to unify the events extending `fr.acinq.eclair.payment.PaymentEvent` with `fr.acinq.eclair.db.IncomingPayment` and `fr.acinq.eclair.db.OutgoingPayment`? They are used in different DBs but in fact contain almost exactly the same data and it feels a bit redundant.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -78,13 +78,13 @@ trait Eclair {
 
   def receivedInfo(paymentHash: ByteVector32)(implicit timeout: Timeout): Future[Option[IncomingPayment]]
 
-  def send(recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest] = None, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
+  def send(externalId: Option[UUID], recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest] = None, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
```
sstone: Why not use a more generic type (string of 32 chars or less) for external ids ?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -182,17 +178,44 @@ class PaymentLifecycle(nodeParams: NodeParams, id: UUID, router: ActorRef, regis
...
-  def props(nodeParams: NodeParams, id: UUID, router: ActorRef, register: ActorRef) = Props(classOf[PaymentLifecycle], nodeParams, id, router, register)
+  def props(nodeParams: NodeParams, progressHandler: PaymentProgressHandler, router: ActorRef, register: ActorRef) = Props(classOf[PaymentLifecycle], nodeParams, progressHandler, router, register)
+
+  /** This handler notifies other components of payment progress. */
+  trait PaymentProgressHandler {
+    val id: UUID
+
+    // @formatter:off
+    def onSend(): Unit
+    def onSucceed(sender: ActorRef, result: PaymentSent)(ctx: ActorContext): Unit
```
sstone: nit: `onSuccess` / `onFailure` seems more natural

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -19,70 +19,131 @@ package fr.acinq.eclair.db
...
 
-  def listOutgoingPayments(): Seq[OutgoingPayment]
+  /** Get all the outgoing payment attempts in the given time range. */
+  def listOutgoingPayments(from: Long, to: Long): Seq[OutgoingPayment]
 
+  /** Add a new payment request (to receive a payment). */
   def addPaymentRequest(pr: PaymentRequest, preimage: ByteVector32)
 
+  /** Get the payment request for the given payment hash, if any. */
   def getPaymentRequest(paymentHash: ByteVector32): Option[PaymentRequest]
```
dpad85: All the `xxxPaymentRequest` method are very confusing and do not point out that they deal with `IncomingPayment`. You actually need to read the `SqlitePaymentDb` code to understand that they return the `PaymentRequest` for `IncomingPayment` only. 

Instead of returning a `PaymentRequest` object, these methods should return the actual `IncomingPayment` object containing the payment request, which can be unwrapped afterwards if needed.

Also, the `getIncomingPayment` and `listIncomingPayments` methods only return the settled incoming payments which is not obvious.

- change `getPaymentRequest` to:

```scala
/** Get the incoming payment (paid or not) for the given payment hash, if any. */
def getIncomingPayment(paymentHash: ByteVector32): Option[IncomingPayment]
```

- change `getIncomingPayment` to:

```scala
/** Get the paid incoming payment for the given payment hash, if any. If the payment exists but has expired or has not been paid yet, returns nothing. */
def getPaidIncomingPayment(paymentHash: ByteVector32): Option[IncomingPayment]
```

- change `getPendingPaymentRequestAndPreimage` to:

```scala
/** Get a pending incoming payment for the given payment hash, if any. If the payment exists but has been paid or has expired, returns nothing. */
def getPendingIncomingPayment(paymentHash: ByteVector32): Option[IncomingPayment]
```

- change `listPaymentRequests` , `listPendingPaymentRequests` and `listIncomingPayments` to:

```scala
/** List all incoming payments (pending, expired and fulfilled) in the given time range */
def listIncomingPayments(from: Long, to: Long): Seq[IncomingPayment]

/** List all pending incoming payments in the given time range */
def listPendingIncomingPayments(from: Long, to: Long): Seq[IncomingPayment]

/** List all paid incoming payments in the (optional) given time range */
def listPaidIncomingPayments(from: Option[Long], to: Option[Long]): Seq[IncomingPayment]
```

t-bast: Completely agree, this is a good opportunity to re-work on this interface and make it more intuitive.

dpad85:
I'd like to use this PR to add a method in `PaymentDB` that retrieve a list of incoming+outgoing payments in a single call. This method would query the `sent_payments` and `received_payments` DB and build a list of objects designed to be as light as possible for good performance (thinking of mobile app here).

```scala
Payment(id: Either[UUID, ByteVector32], 
               amount: Option[MilliSatoshi],
               paymentRequest: Option[String],
               status: Either[OutgoingPaymentStatus, IncomingPaymentStatus],
               completedAt: Option[Long])
```

Note that `paymentRequest` is a `String` because deserializing a `PaymentRequest` is costly especially if the list is large.

t-bast:
> I'd like to use this PR to add a method in PaymentDB that retrieve a list of incoming+outgoing payments in a single call.

That sounds reasonable, I'll add that too.
Many changes to the PaymentsDb interface coming in the next commits ;)

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -19,70 +19,131 @@ package fr.acinq.eclair.db
...
 trait PaymentsDb {
 
-  // creates a record for a non yet finalized outgoing payment
+  /** Create a record for a non yet finalized outgoing payment. */
   def addOutgoingPayment(outgoingPayment: OutgoingPayment)
 
-  // updates the status of the payment, if the newStatus is SUCCEEDED you must supply a preimage
-  def updateOutgoingPayment(id: UUID, newStatus: OutgoingPaymentStatus.Value, preimage: Option[ByteVector32] = None)
+  /** Update the status of the payment in case of success. */
+  def updateOutgoingPayment(paymentResult: PaymentSent)
```
pm47: Nit: could have been an `Either[PaymentFailed, PaymentSent]`. Not sure what's best

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -19,70 +19,131 @@ package fr.acinq.eclair.db
...
-  // all the outgoing payment (attempts) to pay the given paymentHash
+  /** Get all the outgoing payment attempts to pay the given paymentHash. */
   def getOutgoingPayments(paymentHash: ByteVector32): Seq[OutgoingPayment]
 
-  def listOutgoingPayments(): Seq[OutgoingPayment]
+  /** Get all the outgoing payment attempts in the given time range. */
+  def listOutgoingPayments(from: Long, to: Long): Seq[OutgoingPayment]
 
+  /** Add a new payment request (to receive a payment). */
   def addPaymentRequest(pr: PaymentRequest, preimage: ByteVector32)
```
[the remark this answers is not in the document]
pm47: One could say that calling an invoice an "incoming payment" is a bit premature. Who knows if there will indeed be an incoming payment for this invoice... But `addPaymentRequest` is indeed too ambiguous.

Why not `addIncomingPayment`/`setPaidIncomingPayment` to be consistent with your suggestions below?

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
pm47: Why are the `from`/`to` fiels optional only for `listPaidIncomingPayments`?

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -19,70 +19,131 @@ package fr.acinq.eclair.db
...
-  * @param amount      amount of the payment, in milli-satoshis
-  * @param receivedAt  absolute time in seconds since UNIX epoch when the payment was received.
-  */
+ * Incoming payment object stored in DB.
+ *
+ * @param paymentHash identifier of the payment.
+ * @param amount      amount of the payment, in milli-satoshis.
+ * @param receivedAt  absolute time in seconds since UNIX epoch when the payment was received.
+ */
 case class IncomingPayment(paymentHash: ByteVector32, amount: MilliSatoshi, receivedAt: Long)
```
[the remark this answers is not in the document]
pm47: Why don't we add the payment request?

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -19,70 +19,131 @@ package fr.acinq.eclair.db
...
+                           externalId: Option[UUID],
+                           paymentHash: ByteVector32,
+                           amount: MilliSatoshi,
+                           targetNodeId: PublicKey,
+                           createdAt: Long,
+                           status: OutgoingPaymentStatus.Value,
+                           paymentRequest_opt: Option[PaymentRequest],
+                           completedAt: Option[Long] = None,
+                           successSummary: Option[PaymentSuccessSummary] = None,
+                           failureSummary: Option[PaymentFailureSummary] = None)
```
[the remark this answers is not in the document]
pm47: Having:

```scala
sealed trait OutgoingPaymentStatus
case object Pending extends OutgoingPaymentStatus
case class PaymentSuccessSummary(paymentPreimage: ByteVector32, feesPaid: MilliSatoshi, route: Seq[HopSummary]) extends OutgoingPaymentStatus
case class PaymentFailureSummary(failures: Seq[FailureSummary]) extends OutgoingPaymentStatus
```

Would reduce the size of `OutgoingPayment` and remove consistencies issues between the enum and the options.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -49,20 +51,42 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
       statement.executeUpdate("CREATE INDEX IF NOT EXISTS channel_errors_timestamp_idx ON channel_errors(timestamp)")
     }
 
+    def migration34(statement: Statement) = {
+      statement.executeUpdate("DROP index sent_timestamp_idx")
```
pm47: Why drop the index and recreate it later?

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -49,20 +51,42 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
+      statement.executeUpdate("ALTER TABLE sent RENAME TO _sent_old")
+      statement.executeUpdate("CREATE TABLE sent (id BLOB NOT NULL, amount_msat INTEGER NOT NULL, fees_msat INTEGER NOT NULL, payment_hash BLOB NOT NULL, payment_preimage BLOB NOT NULL, timestamp INTEGER NOT NULL)")
+      statement.executeUpdate("INSERT INTO sent (id, amount_msat, fees_msat, payment_hash, payment_preimage, timestamp) SELECT id, amount_msat, fees_msat, payment_hash, payment_preimage, timestamp FROM _sent_old")
```
pm47: We should put all those migrations inside transactions

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -49,20 +51,42 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
       statement.executeUpdate("CREATE INDEX IF NOT EXISTS channel_errors_timestamp_idx ON channel_errors(timestamp)")
     }
 
+    def migration34(statement: Statement) = {
+      statement.executeUpdate("DROP index sent_timestamp_idx")
+      statement.executeUpdate("ALTER TABLE sent RENAME TO _sent_old")
+      statement.executeUpdate("CREATE TABLE sent (id BLOB NOT NULL, amount_msat INTEGER NOT NULL, fees_msat INTEGER NOT NULL, payment_hash BLOB NOT NULL, payment_preimage BLOB NOT NULL, timestamp INTEGER NOT NULL)")
```
pm47: I would keep the field `to_channel_id` and only input sub-payments in case of AMP. We can easily group subpayments using the `id` field.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: Agreed, changing that to a string limited to 66 characters (to allow UUID, Long and Public Keys)

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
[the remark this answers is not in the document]
t-bast: I'm adding the payment request, an incoming payment status and optionally the `received_amount` and `received_at` (in case of success).

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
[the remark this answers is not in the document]
t-bast: Good idea, I'll explore that.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
t-bast: Not a strong opinion, but I slightly prefer overloading in that case (it's easier for the caller).

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: Because since we're renaming the table and creating a new one, the index won't follow automatically to the new table.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: I'll re-work that part, that makes sense.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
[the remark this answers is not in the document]
dpad85: > [...] calling an invoice an "incoming payment" is a bit premature. Who knows if there will indeed be an incoming payment for this invoice[...]

A node generates an invoice only because it expects this invoice to be paid with an incoming payment. Of course this payment may never actually come, just as an outgoing payment may never be actually sent, though we will store it all the same.

`setPaidIncomingPayment ` sounds great!

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
dpad85: > Why are the `from`/`to` fields optional only for `listPaidIncomingPayments`?

My bad, the `from`/`to` fields should be optional in all 3 methods. But now I'm thinking we should not add to much stuff into this PR, and perhaps should just change the methods' name. I think at some point we will have to add more powerful filtering parameters.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: Actually I'm not sure it makes sense. I think it's better that AuditDb doesn't know about the internal details that a payment was AMP (this is an implementation detail that you can ignore when you only want to look at things from a high-level view, especially since either all sub-payments complete or none of them complete).

The sub-payment details (with the whole route, hence the channelId) are already in the PaymentsDb if you want to have a more detailed look at how the payment was split and what routes it took. Let's discuss that offline.

t-bast:
The latest commits should resolve all the pending comments (please mark them resolved if everything looks good). I chose a slightly different naming which I think makes sense.

New things that were added and not necessarily discussed:

- Harmonized `Long` to always be milliseconds (even for the `from` and `to` fields used in `listXXX` functions)
- Made payment expiry explicitly not null -> the spec says in the absence of an explicit expiry, we use 1 hour. That means the `expire_at` column in the `PaymentsDb` should always be set to a non-null values to avoid discrepancies between that column and calling `paymentRequest.isExpired`
- Run migrations/init inside SQL transactions: there was an issue with our previous transaction code, it didn't commit the changes nor rollback...that made it look like it worked but only because we're using a single connection to the DB and it was memory-inefficient (intermediate results were kept locally in the connection instead of committed to the DB)

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: As discussed offline, I'll investigate this further ;)

eclair-core/src/main/scala/fr/acinq/eclair/payment/Relayer.scala
```diff
@@ -138,8 +136,9 @@ class Relayer(nodeParams: NodeParams, register: ActorRef, paymentHandler: ActorR
         case Local(id, None) =>
           // we sent the payment, but we probably restarted and the reference to the original sender was lost,
           // we publish the failure on the event stream and update the status in paymentDb
-          nodeParams.db.payments.updateOutgoingPayment(id, OutgoingPaymentStatus.FAILED)
-          context.system.eventStream.publish(PaymentFailed(id, paymentHash, Nil))
+          val result = PaymentFailed(id, paymentHash, Nil)
+          nodeParams.db.payments.updateOutgoingPayment(result)
```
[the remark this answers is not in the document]
t-bast: I'll investigate that.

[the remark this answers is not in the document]
t-bast: I'm not a big fan, it feels hackish to use a deadletters sender.
I'll re-visit the relayer part in the MPP PR because we'll have to add more logic here, so let's keep it like that for now?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -20,20 +20,97 @@ import java.util.UUID
...
+  require(parts.nonEmpty, "sent payment is empty")
+  val amount: MilliSatoshi = parts.map(_.amount).sum
+  val feesPaid: MilliSatoshi = parts.map(_.feesPaid).sum
+  val timestamp: Long = parts.map(_.timestamp).min
 }
 
-case class PaymentSent(id: UUID, amount: MilliSatoshi, feesPaid: MilliSatoshi, paymentHash: ByteVector32, paymentPreimage: ByteVector32, toChannelId: ByteVector32, timestamp: Long = Platform.currentTime) extends PaymentEvent
+object PaymentSent {
+
+  case class PartialPayment(id: UUID, amount: MilliSatoshi, feesPaid: MilliSatoshi, toChannelId: ByteVector32, route: Seq[Hop], timestamp: Long = Platform.currentTime)
```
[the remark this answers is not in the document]
t-bast: Actually you really need to provide a route all the time so I like having directly a Seq.
You're referring to the hackish part in the AuditDB where we set this to `Nil` because it's not really relevant for Audit right? I wanted to avoid creating yet another type, but I think putting an option there only because of this small AuditDB hack would be too harmful for the normal code...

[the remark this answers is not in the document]
pm47: Not I was referring to the relayer

PR #1131, opened 2019-09-10
Halt on fatal validation errors

araspitzu:
This PR aims at improving the error handling when we can't validate channel announcements, the behavior is now to halt completely to avoid continuing without being able to validate the network updates.


araspitzu:
Closing because it's superseded by a more general approach in #1132

PR #1132, opened 2019-09-10
Better handling of bitcoind failures

araspitzu:
Currently eclair checks for bitcoin connectivity only at startup but if for any reason bitcoin _becomes_ unreachable eclair won't stop the channel operations, this is dangerous and can lead to fund loss (undetected channel breaches). This PR introduces a `FatalErrorMonitor` actor that will monitor for unrecoverable/fatal errors and halt eclair when they are detected, currently only the ZMQ disconnection is handled.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -27,6 +27,7 @@ eclair {
     rpcpassword = "bar"
     zmqblock = "tcp://127.0.0.1:29000"
     zmqtx = "tcp://127.0.0.1:29000"
+    disconnect-grace-seconds = 60
   }
 
   min-feerate = 3 // minimum feerate in satoshis per byte
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -263,7 +263,7 @@ class Setup(datadir: File,
       server = system.actorOf(SimpleSupervisor.props(Server.props(nodeParams, authenticator, serverBindingAddress, Some(tcpBound)), "server", SupervisorStrategy.Restart))
       paymentInitiator = system.actorOf(SimpleSupervisor.props(PaymentInitiator.props(nodeParams, router, register), "payment-initiator", SupervisorStrategy.Restart))
       _ = for (i <- 0 until config.getInt("autoprobe-count")) yield system.actorOf(SimpleSupervisor.props(Autoprobe.props(nodeParams, router, paymentInitiator), s"payment-autoprobe-$i", SupervisorStrategy.Restart))
-      _ = system.actorOf(SimpleSupervisor.props(Props(new FatalErrorMonitor), "fatal-error-actor", SupervisorStrategy.Restart))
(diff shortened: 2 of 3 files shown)


araspitzu:
There is no test here because testing `System.exit( )` is quite hard, the test suite terminates and breaks the CI. I couldn't find an option to fork the jvm only for one test, suggestions are very welcome.

araspitzu:
I found [this check](https://github.com/ACINQ/eclair/blob/f724efaa76b256048de18f706e9cb58ecbebd6aa/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala#L125) which is checking for the same error that i observer "in the wild" to some eclair users (running eclair with bitcoind **reindexing**), however something did not go as planned. Looking at  `BitcoinCoreWallet.scala` line 127 is never reached and eclair doesn't stop, instead the call returns `true` which leads to the channel being abruptly closed and removed from the channel db.

Relevant part of the logs where we wrongly detect the funding tx being double spent: 
```
f.a.e.channel.Channel  - funding tx has been double spent! fundingTxid=YYYYY fundingTx=XXXXX
f.a.e.channel.Channel  n:AAAAAA c:BBBBBBB - failed to publish funding tx
f.a.e.channel.Channel  n:AAAAAA c:BBBBBBB - deleting database record for channelId=BBBBBBB
```
Note that the user's bitcoind was reindexing and it was responding:
`No such mempool transaction. Blockchain transactions are still in the process of being indexed. Use gettransaction for wallet transactions.`

This comes from [this line](https://github.com/bitcoin/bitcoin/commit/6d772a3d441875fbdcd7c15aaa8d9b97f61aa3a9#diff-01aa7d1d32f1b9e5a836c9c411978918R190) and it's no surprise that it doesn't match with our wallet [check](https://github.com/ACINQ/eclair/blob/f724efaa76b256048de18f706e9cb58ecbebd6aa/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala#L125) since the error from bitcoind does **not** contain "indexing". I wonder if fixing that would have avoided troubles and perhaps even this PR is not necessary if the check works.

pm47:
Good catch. The problem with this kind of hacks is that it's very difficult to test. We should probably just look for string "index" here.

araspitzu:
Yeah string based matching is very error prone, the only way we can detect errors/changes in advance is through extensive regression testing. After the findings from my last comment what do you think of this PR? The `indexed`/`indexing` issue can be solved separately from the PR but i think it still makes sense to work on a general monitor mechanism for bitcoind (am i the only one feels this need?). I just added a tiny improvement on the `FatalMonitorActor` to make it less trigger happy.

pm47:
I would first fix the `indexed`/`indexing` mismatch.

I'm unconvinced that monitoring ZMQ connections is the right way to go. I believe stopping eclair when there are RPC errors at e.g. eclair startup or channel opening makes more sense, because there is an interaction with the user and he could notice that something is wrong. Just having eclair stop when the user is asleep doesn't seem to be a good thing.

araspitzu:
Okay then let's close this PR and fix the `indexed`/`indexing` issue on another one, after all it seems to be a fairly common error the users are hitting.

PR #1134, opened 2019-09-11
Add codecov integration to semaphore CI

araspitzu:


.semaphore/semaphore.yml
```diff
@@ -1,23 +1,23 @@
...
+        # the cache util. It must be set in all blocks.
+        - name: MAVEN_OPTS
+          value: "-Dmaven.repo.local=.m2"
+      jobs:
+        - name: Build & Test
+          commands:
+            - sem-version java 11
+            - checkout
+            - cache restore maven
+            - mvn scoverage:report && bash <(curl -s https://codecov.io/bash)
```
pm47: Is it possible to put this on the next line?
```
- mvn scoverage:report
- bash <(curl -s https://codecov.io/bash)
```

araspitzu: I already tried it in 7756c5e but it doesn't work (codecov upload was executed but the report wasn't pushed on the PR)

pm47:
Looking at the last build report I see:

```
-> Pinging Codecov
https://codecov.io/upload/v4?package=bash-8a28df4&token=secret&branch=semaphore_codecov&commit=0e390c5ab6d533bd542041596d5f699f1b9ba195&build=&build_url=&name=&tag=&slug=ACINQ%2Feclair&service=semaphore&flags=&pr=&job=
HTTP 400
Please provide the repository token to upload reports via `-t :repository-token`
```

araspitzu:
Turns out the token is not required for public projects **if** they're using a CI service from a certain list and semaphore isn't in the list.

pm47:
Shouldn't that token have been put in the "semaphore secrets"? Not sure how sensitive it is.

araspitzu:
It's certainly better to put it in the secrets, although the codecov token doesn't allow to commit or change code but only to upload coverage reports (still bad).

PR #1135, opened 2019-09-12
Drop support for Java 8

pm47:
We already have Java 7 (for Android) and Java 11. Supporting Java 8
would require crossbuilding, which we are not doing (two recent PRs
broke the build on Java 8).

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -6,8 +6,6 @@
 - [Maven](https://maven.apache.org/download.cgi) 3.6.0 or newer
 - [Docker](https://www.docker.com/) 18.03 or newer (optional) if you want to run all tests
 
-:warning: You can also use [Oracle JDK 1.8](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html) to build and run eclair, but we recommend you use OpenJDK 11.
-
 ## Build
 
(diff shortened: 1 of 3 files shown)


sstone:
LGTM, just one question: how does it interact with GraalVM ? (I thought that you had to use Oracle JDK8 to build for GraalVM)

pm47:
I'll let @araspitzu answer.

araspitzu:
The graal build takes a few workarounds to circumvent JDK8 compatibility issues, for example replacing  `java.util.zip.CRC32C` with `com.google.common.hash.Hashing`, there is also a test related incompatibility due to the usage of `Files.writeString` in BitcoindService that can be replaced with an older Files interface. All in all is not a huge deal to do those compatibility fixes as long as they are not too many.

pm47:
Do you mean that dropping Java 8 make the Holy Build easier or more difficult?

araspitzu:
At this stage wouldn't change much because the fixes are already in place, but if dropping support means we will diverge even more from java 8 compatibility then yes it will make the holy build harder.

pm47:
We won't diverge because we have to stick as much as possible to Java 7 for Android, so I guess that's not a blocker.

PR #1137, opened 2019-09-16
Sphinx: accept invalid downstream errors

### LINKED ISSUE #1110: Two expired HTLCs followed by remote force-close with testnet Eclair node
Remote node is endurance, channel id is `f3f72886a022c36329257fc459478201512d2744facffc3a3b343a4e25e9ce34`, there were two stuck HTLCs with expiration at block `1575802` but endurance was not failing them after that deadline has passed.

Channel was eventually force-closed by endurance with this commit tx: https://testnet.smartbit.com.au/tx/5176657c307a30ae61d3217cfcca8b7d5e0aba297dfa2b2911e826d559db2abb

t-bast:
When a downstream node sends us an onion error with an invalid length, we must forward the failure.
The recipient won't be able to extract the error but at least it knows the payment failed.

Fixes #1110 (I'm having a quick look at lnd and c-lightning to see which one has the bug of creating these invalid errors in the first place)

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
@@ -325,6 +325,8 @@ object Sphinx extends Logging {
       val stream = generateStream(key, PacketLength)
       logger.debug(s"ammag key: $key")
       logger.debug(s"error stream: $stream")
+      // If we received a packet with an invalid length, we trim and pad to forward a packet with a normal length upstream.
+      // This is a poor man's attempt at increasing the likelihood of the sender receiving the error.
       packet.take(PacketLength).padLeft(PacketLength) xor stream
     }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
@@ -318,12 +318,14 @@ object Sphinx extends Logging {
       * @return an encrypted failure packet that can be sent to the destination node.
       */
     def wrap(packet: ByteVector, sharedSecret: ByteVector32): ByteVector = {
-      require(packet.length == PacketLength, s"invalid error packet length ${packet.length}, must be $PacketLength")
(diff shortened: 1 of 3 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -1329,6 +1328,25 @@ class NormalStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
...
 
+  test("recv UpdateFailHtlc (invalid onion error length)") { f =>
+    import f._
+    val sender = TestProbe()
+    val (_, htlc) = addHtlc(50000000 msat, alice, bob, alice2bob, bob2alice)
+    crossSign(alice, bob, alice2bob, bob2alice)
+    // Bob receives a failure with a completely invalid onion error (missing mac)
+    sender.send(bob, CMD_FAIL_HTLC(htlc.id, Left(ByteVector.fill(260)(42))))
+    sender.expectMsg("ok")
+    val fail = bob2alice.expectMsgType[UpdateFailHtlc]
```
pm47: I think we should check that the `fail` message was correctly built. Not sure the 2nd part of the check are really useful as it is already checked in other tests.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -318,12 +318,14 @@ object Sphinx extends Logging {
...
-      require(packet.length == PacketLength, s"invalid error packet length ${packet.length}, must be $PacketLength")
+      if (packet.length != PacketLength) {
+        logger.warn(s"invalid error packet length ${packet.length}, must be $PacketLength (malicious or buggy downstream node)")
+      }
       val key = generateKey("ammag", sharedSecret)
       val stream = generateStream(key, PacketLength)
       logger.debug(s"ammag key: $key")
       logger.debug(s"error stream: $stream")
-      packet xor stream
+      packet.take(PacketLength).padLeft(PacketLength) xor stream
```
pm47: Maybe a comment to explain what we are doing here?

PR #1138, opened 2019-09-17
Update string to match on bitcoind while it's indexing

araspitzu:
Since we can't check if bitcoind is indexing during setup we had a string based match to detect it during the execution. The check occurs only in the `doubleSpent()` call of `BitcoinCoreWallet`, and it's currently not working because the response from bitcoind does **not** contain the word "indexing" anymore [source](https://github.com/bitcoin/bitcoin/blob/v0.18.1/src/rpc/rawtransaction.cpp#L197).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -156,6 +156,11 @@ class Setup(datadir: File,
             .filter(value => (value \ "spendable").extract[Boolean])
             .map(value => (value \ "address").extract[String])
         }
+        _ <- chain match {
+          case "mainnet" => bitcoinClient.invoke("getrawtransaction", "2157b554dcfda405233906e461ee593875ae4b1b97615872db6a25130ecc1dd6") // coinbase of #500000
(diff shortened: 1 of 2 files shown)


araspitzu:
How about moving the error check up in the `getTransaction` call? This way we'll be able to notice if bitcoind is reindexing a lot before and even prevent the user from opening a channel.

sstone:
Checking error messages feels a bit brittle and they could be changed, what about adding a new check at startup that gets and checks well known blocks and transactions (with a height close the the current tip on mainnet and testnet, 500000 and 1500000 for example) ?

pm47:
Yes I think that would be a good idea. What would be nice too is find a way to have a check at startup. I wonder what happens if we do a `getrawtransaction` on e.g. txid `000000000..00000000`. Will it raise the indexing error? If so, we could consider that a "not found" error is OK, and an "indexing" error makes eclair stop.

araspitzu:
Yes the way we can check for this during startup is to call `getrawtransction` on a well known transaction in a block, this will have to be adapted for both mainnet and testnet; I'm not sure about checking for `0000...000` but it seems brittle (we would be relying that the indexing status trumps the unknown id error).

sstone:
If it cannot retrieve a well known block/tx at startup, it does not really matter what the error is the app should just stop.

araspitzu:
> If it cannot retrieve a well known block/tx at startup, it does not really matter what the error is the app should just stop.

Commit 87f812d checks for the coinbase of block #500000 (in mainnet) during startup, should we add the `getblock` call as well?

Issue #1085, opened 2019-07-21
Building eclair with graalvm native image


Christewart:
Hi all,

Graalvm has a ahead of time compiler (AOTC) that allows you to create native binaries for jars. I would like to attempt to use this for bitcoin-s and the testing suites we have to reduce test time. I've attempted to build the eclair-node with the graalvm native image and am running into a few issues. I thought I would file an issue here asking for official support for this

Here is my branch

https://github.com/Christewart/eclair/tree/native-image-build

Here is some issues i have open o the graal repo https://github.com/oracle/graal/issues/1513

This is some what unrelated, but there is _allegedly_ support for compiling native iOS code with graal as well. https://gluonhq.com/java-on-ios-for-real/

sstone:
Hi Chris,
Thanks, It looks interesting. I suspect that your main pb is startup time, and it may be caused by capsule which we use to package eclair (I guess it may also cause issue with GraalVM ?). I guess that packaging the app as a traditional fat jar (which has its share of issues) or as a directory with a `lib` subdir that includes all dependencies could make it start much quicker too

Christewart:
Graal compiles a native binary, which circumvents the JVM startup time together. 

You can read this blog post for an example of how the native image reduced netty startup time by 90%. This does come at a cost for production applications, but  for testing applications it seems like a no brainer to me.

https://medium.com/graalvm/instant-netty-startup-using-graalvm-native-image-generation-ed6f14ff7692

Another project adopting it is Spring, which is a notoriously heavy weight project: 

https://github.com/spring-projects/spring-framework/wiki/GraalVM-native-image-support#graalvm

The payoff: 


> GraalVM native image allows to compile Spring applications to native executable with very fast startup (less than 100ms) with low memory consumption (usually 5x less than its regular JVM equivalent) at the price of lower throughput and various limitations. Reflection and dynamic proxies are supported but need to be configured manually or via dedicated support. It also allows to produce small container images.

Again, it isn't a silver bullet. But it would be a great option to have IMO.

PR #1140, opened 2019-09-18
Native build

araspitzu:
## Context
Using [GraalVM](https://github.com/oracle/graal) we can provide build options to compile eclair to a native executable that can be run without using any (J)VM, this has the advantage of faster startup times and improved memory consumption making it suitable for embedded devices. This PR is worked on top of `upgrade_scala_and_libs` branch where the scala version has been bumped to 2.12 and several dependencies upgraded to the latest version (i.e akka -> 2.5.23) 

### Known limitations
With GraalVM being still an experimental technology, although already used in production, eclair-native is also experimental this means that not everything might work as expected. Below there is a non exhaustive list of known limitations:
- Logging: does not log to file but only to _stdout_
- NodeAnnouncement alias and color are not serialized correctly.
- Electrum watcher (untested).
- JavaFX GUI (untested).
- Kamon

### Target platforms
The eclair-native build is currently available for linux and macOS while it's Windows support is experimental because GraalVM is not fully available on that platform, we support only one CPU architecture on the native build: x86_64. 

As specified in the [build instructions](https://github.com/ACINQ/eclair/blob/the_holy_build/BUILD.md#native-build-experimental) eclair-native needs some particular dependencies that are in the process of being reviewed, until then the CI pipeline won't work.

Related issue #1085 

For testing purposes i've setup an eclair-native node in testnet at:
`03d5592d8f27d89a8c1be27ea5628b78e5154f70a4d5c044709c5c58dad55d4a4d@172.81.180.147:9737`


sstone:
Closed, GraalVM is not yet a realistic option for us and if we resume working on this we'll probably have to start from scratch.

PR #1141, opened 2019-09-19
Implement option_static_remotekey

araspitzu:
Add support for the upcoming feature `option_static_remotekey` https://github.com/lightningnetwork/lightning-rfc/pull/642


araspitzu:
Commit a1934f5 introduces a new `ChannelVersion` to be able to "tag" the channel instance with `STATIC_REMOTEKEY` version. This was necessary because on reconnection we [always overwrite](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1413) local features but `option_static_remotekey` is sticky and if negotiated during channel opening it applies to all commitment transactions.

pm47:
@sstone There is a new `ChannelVersion` in this PR

araspitzu:
For those interested in testing i'm running 12bf3ec on a testnet node at `021eace881c1080ff4238e7e8b89aaac171c229f6e3e6650e21135cb4f3b398658@172.81.180.147:9737`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -550,12 +552,13 @@ object Commitments {
...
   }
 
-  def makeRemoteTxs(keyManager: KeyManager, commitTxNumber: Long, localParams: LocalParams, remoteParams: RemoteParams, commitmentInput: InputInfo, remotePerCommitmentPoint: PublicKey, spec: CommitmentSpec, channelVersion: ChannelVersion): (CommitTx, Seq[HtlcTimeoutTx], Seq[HtlcSuccessTx]) = {
+  def makeRemoteTxs(keyManager: KeyManager, channelVersion: ChannelVersion, commitTxNumber: Long, localParams: LocalParams, remoteParams: RemoteParams, commitmentInput: InputInfo, remotePerCommitmentPoint: PublicKey, spec: CommitmentSpec): (CommitTx, Seq[HtlcTimeoutTx], Seq[HtlcSuccessTx]) = {
+    val channelKeyPath = keyManager.channelKeyPath(localParams, channelVersion)
     val localPaymentPubkey = channelVersion match {
-      case STANDARD => Generators.derivePubKey(keyManager.paymentPoint(localParams.channelKeyPath).publicKey, remotePerCommitmentPoint)
-      case STATIC_REMOTEKEY => keyManager.paymentPoint(localParams.channelKeyPath).publicKey
+      case STANDARD => Generators.derivePubKey(keyManager.paymentPoint(channelKeyPath).publicKey, remotePerCommitmentPoint)
+      case STATIC_REMOTEKEY => keyManager.paymentPoint(channelKeyPath).publicKey
```
sstone: The intent behind static remote key is to use wallet keys so that users who have lost everything will still eventually get their funds back because the remote commit tx sends directly to their wallet.
Here it's not the case and we will still need one extra tx to return funds to the user's wallet. We don't need extra data though as we will be able to to compute the static key from the funding pubkey, but I think it's worth looking at hoiw we would implement using wallet keys here.

araspitzu: 8c8fb83 implements `toRemote` using the wallet keys, the PoC uses "defaultFinalScriptPubKey" as the script pubkey for our main output in the remote commit transaction. Among the difficulties this forces us to ask the wallet for the private key associated with 'defaultFinalScriptPubKey' to be able to send our `paymentBasepoint` in Open/Accept messages, also because the `toRemote` is a P2WPKH we'd need to drop support for bitcoin core `addresstype=p2sh-segwit` because that would be a P2SH(P2WPKH) and the remote peer can't possibly know that (unless we change the way we compute defaultFinalScriptPubKey). Note that 8c8fb83 is using a few "shortcuts" to get the job done while the final implementation of would store our `paymentBasepoint` in the `LocalParams` for later use (without querying the wallet too often)

sstone: Asking for the wallet's private key is a no-go, but I think there's something else we could try:
- call `getnewaddress` and ask for a `bech32`address. this should work even if the wallet's address type is `p2sh-segwit`
- call `getaddressinfo` on the generated address. It's a new API call that was added in 0.17, and it should return the raw pubkey for this address

Sending to a `p2wpkh` address for this pubkey (which should match the address returned by `getnewaddress`) should be detected by Bitcoin Core even in `p2sh-segwit` mode. 

We would need to add a `getpubkey` method to our ElectrumWallet but that's easy

araspitzu: As per suggestion 1ac4c48 uses `getaddressinfo` to get the wallet public key associated with the final address, also it's now storing the `localPaymentBasepoint` on the `LocalParams` instead of calling the wallet.

araspitzu:
At the current status the PR implements `option_static_remotekey` using the wallet keys in order to avoid doing a claim transaction for our main output, the key associated to our output is stored in the `LocalParam` for convenience during commitment computation. The option is sticky and once negotiated during channel opening it stays enabled (thus the new `ChannelVersion` bit), it also honors the `option_data_loss_protect` fields when reestablishing a connection to the remote peer. Feature support is enabled by default.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -64,16 +66,23 @@ trait StateTestsHelperMethods extends TestKitBase with fixture.TestSuite with Pa
...
   def reachNormal(setup: SetupFixture,
                   tags: Set[String] = Set.empty): Unit = {
     import setup._
-    val channelVersion = ChannelVersion.STANDARD
+    val channelVersion = if(tags.contains("static_remotekey")) ChannelVersion.STATIC_REMOTEKEY else ChannelVersion.STANDARD
     val channelFlags = if (tags.contains("channels_public")) ChannelFlags.AnnounceChannel else ChannelFlags.Empty
     val pushMsat = if (tags.contains("no_push_msat")) 0.msat else TestConstants.pushMsat
-    val (aliceParams, bobParams) = (Alice.channelParams, Bob.channelParams)
+    val (aliceParams, bobParams) = if(tags.contains("static_remotekey")) {
+      // TODO improve key generation
```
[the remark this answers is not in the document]
araspitzu: The `TestWallet` currently has the keys hardcoded and they are the same for both Alice and Bob, this could be misleading when reading test failures. Improvement would be to generate them randomly at every test run.

eclair-core/src/test/scala/fr/acinq/eclair/FeaturesSpec.scala
```diff
@@ -41,12 +41,28 @@ class FeaturesSpec extends FunSuite {
...
 
+  test("'option_static_remotekey feature") {
+    val optionalSupport = hex"2000"
+    val mandatorySupport = hex"1000"
+    val noSupport = hex"0000"
+
+    assert(hasFeature(mandatorySupport, STATIC_REMOTEKEY_MANDATORY))
+    assert(hasFeature(optionalSupport, STATIC_REMOTEKEY_OPTIONAL))
+
+    assert(Features.canUseStaticRemoteKey(localFeatures = optionalSupport, remoteFeatures = mandatorySupport) == true)
```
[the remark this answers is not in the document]
araspitzu: Although generally ugly I found more readable to explicit the `== true` when the assertion condition is very long, mostly because the negation is hard to spot with naked eyes. Not a big deal though, i can remove it.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -49,6 +52,20 @@ object Features {
    */
   def hasVariableLengthOnion(features: ByteVector): Boolean = hasFeature(features, VARIABLE_LENGTH_ONION_MANDATORY) || hasFeature(features, VARIABLE_LENGTH_ONION_OPTIONAL)
 
+  /**
+    * We can use STATIC_REMOTEKEY if both peer have at least optional support.
+    */
+  def canUseStaticRemoteKey(localFeatures: ByteVector, remoteFeatures: ByteVector): Boolean = {
```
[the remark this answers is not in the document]
araspitzu: Nice, I will add that. Do you think is premature to create a generic method to check if we can use a certain feature with the remote?

araspitzu:
> Nice, this is shaping up!
> Have you been able to test compatibility with LND and CL? I think they both have it on their master branch.

I've done some testing with @rustyrussel's node in testnet, i will repeat them with LND now that is in master branch.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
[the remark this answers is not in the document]
t-bast: Yeah, I think it will likely be a case-by-case basis, probably too early to generalize

t-bast:
> I've done some testing with @rustyrussel's node in testnet, i will repeat them with LND now that is in master branch.

I think especially the channel reestablish case needs to be tested. From what I understood LND may be weird there, they requested us to add to the spec the requirement to keep sending a dummy but valid point in my_last_per_commitment_point so we should check they accept what we sent them.

araspitzu:
> I think especially the channel reestablish case needs to be tested. From what I understood LND may be weird there, they requested us to add to the spec the requirement to keep sending a dummy but valid point in my_last_per_commitment_point so we should check they accept what we sent them.

Are you referring to this [comment](https://github.com/lightningnetwork/lightning-rfc/pull/642#discussion_r325444865)? I purposely left the DLP fields in place in the `ChannelReestablish`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -199,9 +202,39 @@ final case class LocalParams(nodeId: PublicKey,
                              maxAcceptedHtlcs: Int,
                              isFunder: Boolean,
                              defaultFinalScriptPubKey: ByteVector,
+                             localPaymentBasepoint: Option[PublicKey],
```
sstone: There should be a check that `localPaymentBasepoint` can be defined iff option `static_remote_key` is set

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/EclairWallet.scala
```diff
@@ -30,6 +31,8 @@ trait EclairWallet {
 
   def getFinalAddress: Future[String]
 
+  def getPubkeyForAddress(address: String): Future[PublicKey]
```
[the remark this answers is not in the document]
araspitzu: I prefer `getReceivePubkey` as it tells that it's the receiving key/address of the wallet, i'll update `getReceiveAddress` too.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
araspitzu: I've added the check in `Commitments` since we don't have the `channelVersion` down in `LocalParams`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -163,7 +164,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         maxAcceptedHtlcs = localParams.maxAcceptedHtlcs,
         fundingPubkey = fundingPubKey,
         revocationBasepoint = keyManager.revocationPoint(channelKeyPath).publicKey,
-        paymentBasepoint = keyManager.paymentPoint(channelKeyPath).publicKey,
+        paymentBasepoint = channelVersion match {
+          case v if v.isSet(USE_STATIC_REMOTEKEY_BIT) => localParams.localPaymentBasepoint.get
+          case STANDARD => keyManager.paymentPoint(channelKeyPath).publicKey
+        },
```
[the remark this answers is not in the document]
araspitzu: Good catch, i pushed the fix in cd4149d but using the wildcard instead of `if .. else ..`, it felt more consistent with other part of the code where is used the pattern matching

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -551,8 +555,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
   }
 
   def createNewChannel(nodeParams: NodeParams, funder: Boolean, fundingAmount: Satoshi, origin_opt: Option[ActorRef]): (ActorRef, LocalParams) = {
-    val defaultFinalScriptPubKey = Helpers.getFinalScriptPubKey(wallet, nodeParams.chainHash)
-    val localParams = makeChannelParams(nodeParams, defaultFinalScriptPubKey, funder, fundingAmount)
+    val localPaymentPubkey = Helpers.getWalletPaymentBasepoint(wallet)
+    val localParams = makeChannelParams(nodeParams, localPaymentPubkey, funder, fundingAmount)
```
sstone: With this change, our final pubkeyscript will always be a 'p2wpkh' script even if we don't set `static_remote_key`. This will break eclair-mobile which only understands `p2sh-of-p2wpkh` for now.

araspitzu: This brings up another point as well, the PR is removing the `claimP2WPKHOutputTx` transaction because our `toRemote` output of the remote commit is a key from the BIP32 wallet and we don't need to 'spend back'. However eclair-mobile doesn't currently understands `p2wpkh` outputs so it won't be able to spend out `toRemote` output. Possible solutions:
- improve eclair-mobile to understand `p2wpkh` outputs
- do the `claimP2WPKHOutputTx` to spend back to a `p2sh-p2wpkh`

sstone: We will add support for `p2wpkh` to eclair-mobile, but for now the simplest solution is to not change how it works

araspitzu: Latest commit d956374 is reverting the logic for the creation of the finalScriptPubkey for channels that don't support `option_static_remotekey`

araspitzu:
Commit 096641c moves uses `option_static_remotekey` from global features, see https://github.com/lightningnetwork/lightning-rfc/pull/680

araspitzu:
After some offline discussion i've revived and simplified this PR :tada: so here is a gist of the changes, their rationale and why we should merge them.

Option `option_static_remotekey` simplifies the commitment format by **not** rotating the key for the main output, to fully take advantage of this we want to have our main output spendable by the BIP32 wallet (bitcoin-core in our case) so we ask bitcoin-core to generate an address and get us the raw pubkey for it. Note that we don't need to do the claim-main-output-tx anymore.

Supporting this new option requires us to remember what features that were negotiated during channel opening (see [comment](https://github.com/ACINQ/eclair/pull/1141#issuecomment-533111413)) to achieve so we add a new `ChannelVersion` bit that tells us if this channel uses `option_static_remotekey`. The alternative was to store the byte vector of the features but it would be more impactful on the codecs and we already have the `ChannelVersion` so it comes natural to use that.

There is a new field in the [`LocalParams` codecs](https://github.com/ACINQ/eclair/pull/1141/files#diff-ca253561a8b809afd49e646777f7c9bbR74) this is necessary to remember the payment basepoint (our static key) that was used in the channel. There are some alternatives here: 
- query bitcoin-core for the raw key given our toLocal address
- store the open_channel message that we sent during channel creation

Both options are more impactful on the code or end up doing more changes to the codecs.

We now carry the `ChannelVersion` down to `Commitments` and `Helpers` in order to make decisions on how to derive the local payment basepoint.

Note that those changes are also preparation work to support more commitment formats in the future, i expect that we add new `ChannelVersion`s and use that to support the new formats (i.e dual-anchor). Because there are now multiple commitment formats in the network we might end up in a situation where some nodes reject eclair's connections because we don't support a certain commitment format.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -1102,6 +1115,43 @@ class NormalStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     assert(forward.htlc === htlc)
   }
 
+  test("recv RevokeAndAck (one htlc sent, static_remotekey)", Tag("static_remotekey")) { f =>
```
t-bast, line 1118: I think it would be helpful to run more tests of this suite with `static_remotekey`.
I think at least `recv CMD_FULFILL_HTLC`, `recv UpdateFulfillHtlc`, `recv CMD_FAIL_HTLC` and `recv UpdateFailHtlc` would be useful.

A simple way to do this would be to extract the test content to a function that takes the fixtures as input (for example, `def testReceiveCmdFulfillHtlc(f: FixtureParam): Unit`. Then the existing test simply calls this function, and we can create a duplicate of that test that calls the same function but adds the static_remotekey tag.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -722,6 +722,22 @@ class ClosingStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     relayerA.expectNoMsg(100 millis)
   }
 
+  test("recv BITCOIN_TX_CONFIRMED (remote commit, option_static_remotekey)", Tag("static_remotekey")) { f =>
```
t-bast, line 725: nit: I think this should move to line 550, closer to the same test without static_remotekey

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -775,6 +791,49 @@ class ClosingStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     awaitCond(alice.stateName == CLOSED)
   }
 
+
+  test("recv BITCOIN_TX_CONFIRMED (future remote commit, option_static_remotekey)", Tag("static_remotekey")) { f =>
```
t-bast, line 795: I think this one should be updated with the additional timeout HTLCs, like the "normal" `future remote commit` test.
Ideally if you could extract most of it to a shared function that both tests would call, and only do the last part (sending the `WatchEventSpent`s) that's custom in each test it would be nice.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -797,6 +856,27 @@ class ClosingStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     awaitCond(alice.stateData.asInstanceOf[DATA_CLOSING].copy(revokedCommitPublished = Nil) == initialState)
   }
 
+
```
t-bast, line 859: nit: empty line

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -797,6 +856,27 @@ class ClosingStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
...
+  test("recv BITCOIN_FUNDING_SPENT (one revoked tx, option_static_remotekey)", Tag("static_remotekey")) { f =>
```
t-bast, line 860: Same as others, I think it would be great to factorize these tests so that we run it in both version (no static_remotekey / static_remotekey) without duplicating too much code.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
t-bast, line 725: I think we also need the `next remote commit` case to be tested with static_remotekey.
Those tests are too important to risk missing something :)

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -825,6 +827,62 @@ class IntegrationSpec extends TestKit(ActorSystem("test")) with BitcoindService
     assert(outgoingPayments.forall(p => p.status.isInstanceOf[OutgoingPaymentStatus.Failed]), outgoingPayments)
   }
 
+  test("send payments and close the channel C -> F6 with option_static_remotekey") {
```
t-bast, line 830: Really nice test!

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -825,6 +827,62 @@ class IntegrationSpec extends TestKit(ActorSystem("test")) with BitcoindService
...
+    // toRemote output of C as seen by F6
+    val Some(toRemoteOutC) = initialStateDataF6.commitments.localCommit.publishableTxs.commitTx.tx.txOut.find(_.publicKeyScript == Script.write(toRemoteAddress))
+
+    // let's make a payment to advance the commit index
+    val amountMsat = 4200000.msat
+    sender.send(nodes("F6").paymentHandler, ReceivePayment(Some(amountMsat), "1 coffee"))
+    val pr = sender.expectMsgType[PaymentRequest]
+
+    // then we make the actual payment
+    sender.send(nodes("C").paymentInitiator, SendPaymentRequest(amountMsat, pr.paymentHash, nodes("F6").nodeParams.nodeId, routeParams = integrationTestRouteParams, maxAttempts = 1))
```
t-bast, line 853: I don't think you need the `integrationTestRouteParams` here

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -825,6 +827,62 @@ class IntegrationSpec extends TestKit(ActorSystem("test")) with BitcoindService
...
+    val stateDataF6 = sender.expectMsgType[DATA_NORMAL]
+    val commitmentIndex = stateDataF6.commitments.localCommit.index
+    val commitTx = stateDataF6.commitments.localCommit.publishableTxs.commitTx.tx
+    val Some(toRemoteOutCNew) = commitTx.txOut.find(_.publicKeyScript == Script.write(toRemoteAddress))
+
+    // there is a new commitment index in the channel state
+    assert(commitmentIndex == initialCommitmentIndex + 1)
+
+    // script pubkeys of toRemote output remained the same across commitments
+    assert(toRemoteOutC.publicKeyScript == toRemoteOutCNew.publicKeyScript)
```
t-bast, line 868: Maybe also add:
```
assert(toRemoteOutCNew.amount < toRemoteOutC.amount)
```

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -825,6 +827,62 @@ class IntegrationSpec extends TestKit(ActorSystem("test")) with BitcoindService
...
+    awaitCond({
+      sender.send(nodes("C").register, Forward(channelId, CMD_GETSTATE))
+      sender.expectMsgType[State] == CLOSING
+    }, max = 20 seconds, interval = 1 second)
+
+    sender.send(bitcoincli, BitcoinReq("getrawtransaction", commitTx.txid.toHex))
+    val JString(rawTx) = sender.expectMsgType[JValue](10 seconds)
+
+    // the unilateral close contains the static toRemote output
+    assert(Transaction.read(rawTx).txOut.exists(_.publicKeyScript == toRemoteOutC.publicKeyScript))
```
t-bast, line 883: Can we also test that the channel correctly goes to the `CLOSED` state?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumEclairWallet.scala
```diff
@@ -32,7 +32,9 @@ class ElectrumEclairWallet(val wallet: ActorRef, chainHash: ByteVector32)(implic
 
   override def getBalance = (wallet ? GetBalance).mapTo[GetBalanceResponse].map(balance => balance.confirmed + balance.unconfirmed)
 
-  override def getFinalAddress = (wallet ? GetCurrentReceiveAddress).mapTo[GetCurrentReceiveAddressResponse].map(_.address)
+  override def getReceiveAddress = (wallet ? GetCurrentReceiveAddress).mapTo[GetCurrentReceiveAddressResponse].map(_.address)
+
+  override def getReceivePubkey(receiveAddress: Option[String] = None): Future[Crypto.PublicKey] = Future.failed(new RuntimeException("Not implemented"))
```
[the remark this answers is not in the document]
araspitzu: This feature is not (yet) compatible with eclair-mobile/phoenix because our electrum wallet can't spend `p2wpkh` outputs, this is also why the feature should stay disabled in the eclair-core builds for the mobile apps.

eclair-core/src/test/scala/fr/acinq/eclair/FeaturesSpec.scala
```diff
@@ -51,6 +51,14 @@ class FeaturesSpec extends FunSuite {
     assert(hasFeature(hex"0200", VariableLengthOnion, Some(FeatureSupport.Optional)))
   }
 
+  test("'option_static_remotekey' feature") {
```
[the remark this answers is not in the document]
araspitzu: I think it's okay to make this feature opt-in only and we're going to disable it anyway for the mobile apps.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -138,20 +140,28 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, switchboard: Act
         sender ! Status.Failure(new RuntimeException(s"fundingSatoshis=${c.fundingSatoshis} is too big for the current settings, increase 'eclair.max-funding-satoshis' (see eclair.conf)"))
         stay
       } else {
-        val (channel, localParams) = createNewChannel(nodeParams, funder = true, c.fundingSatoshis, origin_opt = Some(sender))
+        val channelVersion = canUseFeature(d.localInit.features, d.remoteInit.features, StaticRemoteKey) match {
+          case false => ChannelVersion.STANDARD
+          case true => ChannelVersion.STATIC_REMOTEKEY
+        }
+        val (channel, localParams) = createNewChannel(nodeParams, funder = true, c.fundingSatoshis, origin_opt = Some(sender), channelVersion)
```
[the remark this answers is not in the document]
araspitzu: Done in 4abf88d

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2077,12 +2088,17 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-    val remotePerCommitmentPoint = d.remoteChannelReestablish.myCurrentPerCommitmentPoint
-    val remoteCommitPublished = Helpers.Closing.claimRemoteCommitMainOutput(keyManager, d.commitments, remotePerCommitmentPoint, commitTx, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)
-    val nextData = DATA_CLOSING(d.commitments, fundingTx = None, waitingSince = now, Nil, futureRemoteCommitPublished = Some(remoteCommitPublished))
-
-    goto(CLOSING) using nextData storing() calling (doPublish(remoteCommitPublished))
+    d.commitments.channelVersion match {
+      case v if v.isSet(USE_STATIC_REMOTEKEY_BIT) =>
+        val remoteCommitPublished = RemoteCommitPublished(commitTx, None, List.empty, List.empty, Map.empty)
+        val nextData = DATA_CLOSING(d.commitments, fundingTx = None, waitingSince = now, Nil, futureRemoteCommitPublished = Some(remoteCommitPublished))
+        goto(CLOSING) using nextData storing()
```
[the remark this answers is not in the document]
araspitzu: Done in 4abf88d

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -56,7 +57,19 @@ class NormalStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
   implicit val log: akka.event.LoggingAdapter = akka.event.NoLogging
 
   override def withFixture(test: OneArgTest): Outcome = {
-    val setup = init()
+    val testWallet = if(test.tags.contains("static_remotekey")){
```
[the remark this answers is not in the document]
araspitzu: Yeah i messed up my autoformatting some time ago (the format config diverged from the default intellij's) and now i have it disabled :cry: .

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
araspitzu: Good point! I factored out the test body into a function (declared just above the actual tests), now we repeat `CMD_FULFILL_HTLC`, `UpdateFulfillHtlc`, `CMD_FAIL_HTLC` and `UpdateFailHtlc` using `option_static_remotekey` too.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
araspitzu: Done in 4abf88d

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
araspitzu: Done in 4abf88d

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
araspitzu: Good catch ;) done the test in 4abf88d

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
araspitzu: Yes it makes sense especially because with static remote key we don't make claim transaction so it's worth to test that we move to `CLOSED`  , done in 4abf88d

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumEclairWallet.scala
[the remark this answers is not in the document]
t-bast: Got it, that makes sense then let's keep it off by default, and let users turn it on if/when they need it.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -775,14 +821,31 @@ class ClosingStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     awaitCond(alice.stateName == CLOSED)
   }
 
-  test("recv BITCOIN_FUNDING_SPENT (one revoked tx)") { f =>
+  test("recv BITCOIN_TX_CONFIRMED (future remote commit, option_static_remotekey)", Tag("static_remotekey")) { f =>
+    import f._
+    val bobCommitTx = testFutureRemoteCommitTxConfirmed(f)
+    // using option_static_remotekey alice doesn't need to swipe her output
```
t-bast, line 827: nit:
```suggestion
    // using option_static_remotekey alice doesn't need to sweep her output
```

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -194,4 +195,20 @@ class PeerSpec extends TestkitBaseClass with StateTestsHelperMethods {
     assert(channelCreated.initialFeeratePerKw == peer.feeEstimator.getFeeratePerKw(peer.feeTargets.commitmentBlockTarget))
     assert(channelCreated.fundingTxFeeratePerKw.get == peer.feeEstimator.getFeeratePerKw(peer.feeTargets.fundingBlockTarget))
   }
+
+  test("use correct final script if option_static_remotekey is negotiated", Tag("static_remotekey")) { f =>
+    import f._
+
+    val probe = TestProbe()
+    connect(remoteNodeId, switchboard, peer, peerConnection, remoteInit = wire.Init(hex"2200")) // Bob supports option_static_remotekey
+    peer.stateData.channels.foreach { case (_, channelRef) =>
```
t-bast, line 204: This doesn't test anything as `peer.stateData.channels` is empty :)
I think you should test that when we send a message to trigger a channel open, the peer uses the right channel version when sending the channel command.

araspitzu: I wanted to test that we both use the right `ChannelVersion` and we compute the final scriptPubkey with the localPaymentBasepoint (as done in [`createNewChannel()`](https://github.com/ACINQ/eclair/pull/1141/files#diff-51851274f6488b212bc8bfa2fa89797aR284)). I fixed the test in dd3326f

araspitzu:
Rebased on master

araspitzu:
I've done some E2E tests with other implementations and I've simulated the upgrade from an eclair node running the latest master branch, the test were successful and no errors were encountered.

Test 1: running `option_static_remotekey` open a channel to LND (make sure you activated the feature in your conf), then process a few payments to make the commit index increase and check that the main outputs don't change scriptPubKey across commitments. Then force close the channel from the remote side and check that eclair doesn't spend/redeem its main output but instead it should be immediately available to bitcoind. Repeat the test initiating the channel from the remote side (LND).

In test 2 we want to make sure it's safe to upgrade from a node currently running master with several channels already opened or in closing states.
Running `master` open a few channels (both as funder and fundee) then force-close one of them and open a new one leaving it waiting for the funding to be locked. Now stop and restart eclair upgrading to `option_static_remotekey`, on restart all channels should function normally. 

Note for reviewers: to reproduce my tests easily here are some go-to instructions to run LND in your local machine (assumes docker + ubuntu):
```
1 git clone github.com:lightningnetwork/lnd && cd lnd
2 docker build -t lnd-image-master .
// adjust to your local bitcoind instance
3 docker create --net=host --name lnd lnd-image-master  --bitcoin.active --bitcoin.node=bitcoind --bitcoin.regtest --bitcoind.rpchost=127.0.0.1:19234 --bitcoind.rpcuser=foobaz --bitcoind.rpcpass=barbaz --bitcoind.zmqpubrawblock=127.0.0.1:29001 --bitcoind.zmqpubrawtx=127.0.0.1:29002 --no-macaroons
4 docker start lnd
5 docker exec -it lnd lncli --no-macaroons create 
// create invoice from remote
6 docker exec -it lnd lncli --no-macaroons addinvoice
// force-close the channel from remote
7 docker exec -it lnd lncli --no-macaroons closechannel --force --funding_txid <funding_txid>
// check on bitcoind that funds arrived from the commit tx 
8 bitcoin-cli listunspent | jq '.[] | select(.txid == "<commit_txid>")'
```

t-bast:
I think it would be great to finalize this PR, maybe we can then start doing some work on anchor outputs since this is where spec efforts to fix the eclipse attacks are focusing (rust-lightning said they will always require `static_remotekey`).

@sstone will you have some time to review this PR?
I'll then reproduce the E2E tests before we merge.

araspitzu:
I did some extra E2E test with c-lightning v0.8.2.1 and eclair 0.3.4. 

The test with c-lightning reproduced what i did previously with LND, I opened a channel with the `static_remotekey` commitment type and checked that across different commit tx(s) we use the same to_remote address. Then I force closed the channel from the remote side and checked that our funds are sent straight into bitcoin-core bip32 wallet via a p2wkh output.
Here are some quick steps to run c-lightning in your local system and reproduce the test (assumes docker+ubuntu):
```
1. git clone -b v0.8.2.1 https://github.com/ElementsProject/lightning && cd lightning
2. docker build -t cl-image .
3. docker run --net=host cl-image --network=regtest --bitcoin-rpcconnect=127.0.0.1:19234 --bitcoin-rpcuser=foobaz --bitcoin-rpcpassword=barbaz // adjust to your local bitcoind instance
4. docker exec -it jolly_mcclintock lightning-cli --network=regtest getinfo // replace 'jolly_mcclintock' with your container name
// open channel from eclair
// ...
// create invoice from remote side
5. docker exec -it jolly_mcclintock lightning-cli --network=regtest invoice msatoshi=5000000 label="" description="
// force close channel from remote
6. docker exec -it jolly_mcclintock lightning-cli --network=regtest close <channel_id> 1
// check on bitcoind that funds arrived from the commit tx 
7. bitcoin-cli listunspent | jq '.[] | select(.txid == "<commit_txid>")'
```

The second test with an older version of eclair aimed and checking we're still backward compatible, the test comprised of opening channels, making payments and closing the channels (both as funder and fundee). This test was successful since no errors was encountered (note there isn't anything specific to `option_static_remotekey` to be checked)

sstone:
I tested with lnd and c-lightning, and checked that it can be merged and work on Android as well (static_remote_key will be disabled for now on our android apps) and everything is fine. I think that once we have both test vectors and have a second approval we're good to go!

t-bast:
> I think that once we have both test vectors and have a second approval we're good to go!

Great, I'll test and review today!

araspitzu:
Pushed last commit with both versions of test vectors being used in the tests :rocket:

eclair-core/src/test/scala/fr/acinq/eclair/transactions/TestVectorsSpec.scala
```diff
@@ -115,8 +100,11 @@ class TestVectorsSpec extends AnyFunSuite with Logging {
     val revocation_basepoint = revocation_basepoint_secret.publicKey
     val funding_privkey = PrivateKey(hex"1552dfba4f6cf29a62a0af13c8d6981d36d0ef8d61ba10fb0fe90da7634d7e1301")
     val funding_pubkey = funding_privkey.publicKey
-    val payment_privkey = Generators.derivePrivKey(payment_basepoint_secret, Local.per_commitment_point)
-    val per_commitment_point = PublicKey(hex"022c76692fd70814a8d1ed9dedc833318afaaed8188db4d14727e2e99bc619d325")
+    val htlc_privkey = Generators.derivePrivKey(payment_basepoint_secret, Local.per_commitment_point)
+    val payment_privkey = channelVersion.isSet(USE_STATIC_REMOTEKEY_BIT) match {
+      case true => payment_basepoint_secret
+      case false => Generators.derivePrivKey(payment_basepoint_secret, Local.per_commitment_point)
```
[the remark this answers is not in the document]
araspitzu: Done in bb05160

eclair-core/src/test/scala/fr/acinq/eclair/transactions/TestVectorsSpec.scala
```diff
@@ -75,36 +79,17 @@ class TestVectorsSpec extends AnyFunSuite with Logging {
     val delayed_payment_basepoint = delayed_payment_basepoint_secret.publicKey
     val funding_privkey = PrivateKey(hex"30ff4956bbdd3222d44cc5e8a1261dab1e07957bdac5ae88fe3261ef321f374901")
     val funding_pubkey = funding_privkey.publicKey
-
     val per_commitment_point = PublicKey(hex"025f7117a78150fe2ef97db7cfc83bd57b2e2c0d0dd25eaf467a4a1c2a45ce1486")
-    val payment_privkey = Generators.derivePrivKey(payment_basepoint_secret, per_commitment_point)
+    val htlc_privkey = Generators.derivePrivKey(payment_basepoint_secret, per_commitment_point)
+    val payment_privkey = channelVersion.isSet(USE_STATIC_REMOTEKEY_BIT) match {
+      case true => payment_basepoint_secret
+      case false => Generators.derivePrivKey(payment_basepoint_secret, per_commitment_point)
```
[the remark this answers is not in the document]
araspitzu: Done in bb05160

Roasbeef:
🚀

Super happy about this, we're actually looking at making the feature bit start to be _required_ in a release or so.

t-bast:
Sounds good, in that case we'll probably align and at least turn it on optionally by default!
We still need some work done on mobile to fully support it (Bech32 support and stuff) which is why it took a bit long and still isn't enabled in our default config.

PR #1145, opened 2019-09-23
Port the API service to spray.io for ease of testing

araspitzu:


pm47:
Can you explain what the goal of this PR is? Hasn't spray been replaced long ago by `akka-http`?

araspitzu:
@pm47 the goal is to help @sstone testing the android branch by enabling the API again. Since on the android branch we use akka 2.3 it's not possible to use akka-http but spray is the closest thing i could find. The PR is merging into `android-update` which is a WIP for merging `master` into `android`, in fact I opened the PR mostly for internal review.

PR #1148, opened 2019-09-26
Use bitcoin 0.18.1 in the test

araspitzu:
Upgrade the version of bitcoin used in the test, this is preparation work to drop support for bitcoin 0.17. The most notable change is the deprecation of `generate` RPC call which is replaced by `generatetoaddress`, all the test have been updated accordingly.

### DIFF
--- a/eclair-core/pom.xml
+++ b/eclair-core/pom.xml
@@ -79,10 +79,10 @@
                 <activeByDefault>true</activeByDefault>
             </activation>
             <properties>
-                <bitcoind.url>https://bitcoin.org/bin/bitcoin-core-0.17.1/bitcoin-0.17.1-x86_64-linux-gnu.tar.gz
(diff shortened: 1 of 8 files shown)


araspitzu:
Hey nice suggestion, i started mechanically replacing the previous calls with the new ones because it was the safest approach, sometimes we also want to reuse an address that was generated in the same unit test. I'll integrate your suggestion (maybe with a small change).

sstone:
Our policy is to support the last 2 versions of Bitcoin Core, and I don't think there's an ETA yet for 0.19. But there are often API changes even between minor versions so it's good to keep track of them and update our documentation.

araspitzu:
@sstone the PR is only updating the test as prep. work for when bitcoin 0.19 is out and we drop support for bitcoin 0.17, sooner or later these changes had to be done. Should i update the documentation?

sstone:
I was wrong, as pointed out by @pm47 our users will now run eclair with bitcoin 0.18, not 0.17, so it makes sense to test against it (and this PR only changes test code). ACK :)

PR #1149, opened 2019-09-26
Electrum: improve coin selection (fixes #1146)

### LINKED ISSUE #1146: Electrum wallet: cannot fund tx is amount is close to the entire wallet balance
Wallets with only one utxo may fail to fund a transaction that sends all the available balance and the target fee rate is low, and will incorrectly return an `insufficient funds` error. 

See https://github.com/ACINQ/eclair-mobile/issues/211

sstone:
This PR changes how we select UTXOs when funding transactions, and fixes an issue where trying to use the amount returned by "spend all" with only 1 utxo and a low fee rate would result in an incorrect "insufficient funds" error (see #1146)

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
@@ -945,14 +945,15 @@ object ElectrumWallet {
       val amount = tx.txOut.map(_.amount).sum
       require(amount > dustLimit, "amount to send is below dust limit")
 
-      // select utxos that are not locked by pending txs
-      val lockedOutputs = locks.map(_.txIn.map(_.outPoint)).flatten
-      val unlocked = utxos.filterNot(utxo => lockedOutputs.contains(utxo.outPoint))
-      val unlocked1 = if (allowSpendUnconfirmed) unlocked else unlocked.filter(_.item.height > 0)
-
-      // sort utxos by amount, in increasing order
-      // this way we minimize the number of utxos in the wallet, and so we minimize the fees we'll pay for them
-      val unlocked2 = unlocked1.sortBy(_.item.value)
+      val unlocked = {
+        // select utxos that are not locked by pending txs
+        val lockedOutputs = locks.flatMap(_.txIn.map(_.outPoint))
+        val unlocked1 = utxos.filterNot(utxo => lockedOutputs.contains(utxo.outPoint))
+        val unlocked2 = if (allowSpendUnconfirmed) unlocked1 else unlocked1.filter(_.item.height > 0)
+        // sort utxos by amount, in increasing order
+        // this way we minimize the number of utxos in the wallet, and so we minimize the fees we'll pay for them
+        unlocked2.sortBy(_.item.value)
+      }
 
       // computes the fee what we would have to pay for our tx with our candidate utxos and an optional change output
       def computeFee(candidates: Seq[Utxo], change: Option[TxOut]): Satoshi = {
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -971,35 +945,61 @@ object ElectrumWallet {
...
+      val tx1 = addUtxosWithDummySig(tx, selected)
+      val tx2 = change_opt.map(out => tx1.addOutput(out)).getOrElse(tx1)
       val tx3 = signTransaction(tx2)
-      //Transaction.correctlySpends(tx3, utxos.map(utxo => utxo.outPoint -> TxOut(Satoshi(utxo.item.value), computePublicKeyScript(utxo.key.publicKey))).toMap, ScriptFlags.STANDARD_SCRIPT_VERIFY_FLAGS)
 
-      // and add the completed tx to the lokcs
+      // and add the completed tx to the locks
       val data1 = this.copy(locks = this.locks + tx3)
-      val fee3 = spent - tx3.txOut.map(_.amount).sum
+      val fee = selected.map(s => Satoshi(s.item.value)).sum - tx3.txOut.map(_.amount).sum
```
t-bast: Why don't we put a `lazy val fee` inside `Transaction` that does that?
It seems like it's unnecessary to separately return the fee, it's a property of the `Transaction`.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -971,35 +945,61 @@ object ElectrumWallet {
...
 
-      // add change output only if non-dust, otherwise change is added to the fee
-      val (tx2, fee2, pos) = (spent - amount - fee1) match {
-        case dustChange if dustChange < dustLimit => (tx1, fee1 + dustChange, -1) // if change is below dust we add it to fees
-        case change => (tx1.addOutput(TxOut(change, computePublicKeyScript(currentChangeKey.publicKey))), fee1, 1) // change output index is always 1
+      val dummyChange = TxOut(Satoshi(0), computePublicKeyScript(currentChangeKey.publicKey))
+
+      @tailrec
+      def loop(current: Seq[Utxo], remaining: Seq[Utxo]) : (Seq[Utxo], Option[TxOut]) = {
+        totalAmount(current) match {
```
t-bast: nit: small proposal that makes it more obvious that each time we're calling `head` or `tail` we're sure we still have elements (can be ignored if you prefer your version):

```scala
totalAmount(current) match {
  case total if total - amount - computeFee(current, None) < Satoshi(0) && remaining.isEmpty => remaining match {
    // not enough funds to send amount and pay fees even without a change output
    case Nil => throw new IllegalArgumentException("insufficient funds")
    // not enough funds, try with an additional input
    case remaining => loop(remaining.head +: current, remaining.tail)
  }
  case total if total - amount - computeFee(current, None) <= dustLimit =>
    // change output would be below dust, we don't add one and just overpay fees
    (current, None)
  case total if total - amount - computeFee(current, Some(dummyChange)) <= dustLimit => remaining match {
    // change output is above dust limit but cannot pay for it's own fee, and we have no more utxos => we overpay a bit
    case Nil => (current, None)
    // try with an additional input
    case remaining => loop(remaining.head +: current, remaining.tail)
  }
  case total =>
    val fee = computeFee(current, Some(dummyChange))
    val change = dummyChange.copy(amount = total - amount - fee)
    (current, Some(change))
}
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -971,35 +945,61 @@ object ElectrumWallet {
       val amount = tx.txOut.map(_.amount).sum
       require(amount > dustLimit, "amount to send is below dust limit")
 
-      // start with a hefty fee estimate
-      val utxos = chooseUtxos(amount + Transactions.weight2fee(feeRatePerKw, 1000), allowSpendUnconfirmed)
-      val spent = totalAmount(utxos)
+      // select utxos that are not locked by pending txs
+      val lockedOutputs = locks.map(_.txIn.map(_.outPoint)).flatten
```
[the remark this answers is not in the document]
sstone: :+1:

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
sstone: Because a transaction cannot compute its onchain fees since it is missing the values of the UTXOs it is spending, it just know their outpoints.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -971,35 +945,61 @@ object ElectrumWallet {
...
-      // add change output only if non-dust, otherwise change is added to the fee
-      val (tx2, fee2, pos) = (spent - amount - fee1) match {
-        case dustChange if dustChange < dustLimit => (tx1, fee1 + dustChange, -1) // if change is below dust we add it to fees
-        case change => (tx1.addOutput(TxOut(change, computePublicKeyScript(currentChangeKey.publicKey))), fee1, 1) // change output index is always 1
+      val dummyChange = TxOut(Satoshi(0), computePublicKeyScript(currentChangeKey.publicKey))
+
+      @tailrec
+      def loop(current: Seq[Utxo], remaining: Seq[Utxo]) : (Seq[Utxo], Option[TxOut]) = {
+        totalAmount(current) match {
+          case total if total - amount - computeFee(current, None) < Satoshi(0) && remaining.isEmpty =>
```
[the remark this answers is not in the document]
sstone: :+!:

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
sstone: It's highly subjective for I find the current version with an explicit check on `remaining.isEmpty` easier to read

t-bast: Agreed that it's subjective, no need to take this into account if you prefer your version.

sstone: :+1:

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
t-bast: Indeed, the outpoint doesn't have enough information to compute this, my bad.

PR #1150, opened 2019-09-27
HTTP API: add type hints for payment status

t-bast:
Cleans up the JSON payment status (easier to interpret for callers).
@darwin you can probably leverage this to simplify your script (check for the `type` field inside the `status` object, see the JSON examples in this PR).


eclair-node/src/main/scala/fr/acinq/eclair/api/JsonSerializers.scala
```diff
@@ -216,17 +302,9 @@ object JsonSupport extends Json4sSupport {
...
-    val reverse: Map[String, Class[_]] = custom.map(_.swap)
-
-    override val hints: List[Class[_]] = custom.keys.toList
-    override def hintFor(clazz: Class[_]): String = custom.getOrElse(clazz, {
-      throw new IllegalArgumentException(s"No type hint mapping found for $clazz")
-    })
-    override def classFor(hint: String): Option[Class[_]] = reverse.get(hint)
-  }
-
+    new JavaUUIDSerializer).withTypeHintFieldName("type") +
```
araspitzu: Is it necessary to apply the `withTypeHintFieldName` to all the Formats? It seems that it's being applied to all serializers and not only the payment events, although i see that other classes aren't serialized with the `type` (perhaps because they're not in the list of custom type names) and it's a bit hard to find documentation for this method on json4s.

araspitzu: Alternatively (but it's not strictly better) we could define two sets of formats in this way:
```
val formats = ... // default generic formats
val formatsWithTypeHint = formats.withTypeHintFieldName("type") + CustomTypeHint
```

Then use the latter only where necessary.

eclair-node/src/test/scala/fr/acinq/eclair/api/JsonSerializersSpec.scala
```diff
@@ -76,14 +73,12 @@ class JsonSerializersSpec extends FunSuite with Matchers {
   }
 
   test("type hints") {
-    implicit val formats = JsonSupport.formats.withTypeHintFieldName("type") + CustomTypeHints(Map(classOf[PaymentSettlingOnChain] -> "payment-settling-onchain")) + new MilliSatoshiSerializer
     val e1 = PaymentSettlingOnChain(UUID.randomUUID, 42 msat, randomBytes32)
-    assert(Serialization.writePretty(e1).contains("\"type\" : \"payment-settling-onchain\""))
+    assert(Serialization.writePretty(e1)(JsonSupport.formats).contains("\"type\" : \"payment-settling-onchain\""))
```
araspitzu:
```suggestion
    assert(JsonSupport.serialization.writePretty(e1)(JsonSupport.formats).contains("\"type\" : \"payment-settling-onchain\""))
```

araspitzu: Actually we should replace all the instances of `org.json4s.jackson.Serialization` with `JsonSupport.serialization` otherwise our test would be overriding the serialization backend for json4s.

eclair-node/src/main/scala/fr/acinq/eclair/api/JsonSerializers.scala
t-bast: From what I understand it will only have an effect for types for which you actually define a type hint.
I honestly don't see a reason to separate the two, having those type hints is beneficial everywhere (and the code is less messy if we don't have two separate formats).

araspitzu: Yes that seems to be the effect, i left the comment because i wasn't sure of its behavior and i can't find any documentation about it.

PR #1152, opened 2019-09-30
Commitments: take HTLC fee into account

t-bast:
Our balance computation was slightly incorrect. If you want to know how much you can send (or receive), you need to take into account the fact that you'll add a new HTLC which adds weight to the commit tx (and thus adds fees).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -88,27 +88,41 @@ case class Commitments(channelVersion: ChannelVersion,
     // we need to base the next current commitment on the last sig we sent, even if we didn't yet receive their revocation
     val remoteCommit1 = remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit)
     val reduced = CommitmentSpec.reduce(remoteCommit1.spec, remoteChanges.acked, localChanges.proposed)
-    // The funder always pays the on-chain fees, so we must subtract that from the amount we can send.
-    val feesMsat = if (localParams.isFunder) commitTxFee(remoteParams.dustLimit, reduced).toMilliSatoshi else 0.msat
-    val balance = (reduced.toRemote - remoteParams.channelReserve.toMilliSatoshi - feesMsat).max(0 msat)
-    // We also need to take into account the fact that adding an htlc increases the commit tx fee if it's not trimmed.
-    if (localParams.isFunder && balance >= offeredHtlcTrimThreshold(remoteParams.dustLimit, reduced)) {
-      (balance - htlcOutputFee(reduced.feeratePerKw)).max(0 msat)
+    val balanceNoFees = (reduced.toRemote - remoteParams.channelReserve).max(0 msat)
+    if (localParams.isFunder) {
+      // The funder always pays the on-chain fees, so we must subtract that from the amount we can send.
+      val commitFees = commitTxFee(remoteParams.dustLimit, reduced).toMilliSatoshi
+      val htlcFees = htlcOutputFee(reduced.feeratePerKw)
+      if (balanceNoFees - commitFees < offeredHtlcTrimThreshold(remoteParams.dustLimit, reduced)) {
+        // htlc will be trimmed
+        (balanceNoFees - commitFees).max(0 msat)
+      } else {
+        // htlc will have an output in the commitment tx, so there will be additional fees.
+        (balanceNoFees - commitFees - htlcFees).max(0 msat)
+      }
     } else {
-      balance
+      // The fundee doesn't pay on-chain fees.
+      balanceNoFees
     }
   }
 
   lazy val availableBalanceForReceive: MilliSatoshi = {
     val reduced = CommitmentSpec.reduce(localCommit.spec, localChanges.acked, remoteChanges.proposed)
-    // The funder always pays the on-chain fees, so we must subtract that from the amount we can send.
-    val feesMsat = if (localParams.isFunder) 0.msat else commitTxFee(localParams.dustLimit, reduced).toMilliSatoshi
-    val balance = (reduced.toRemote - localParams.channelReserve.toMilliSatoshi - feesMsat).max(0 msat)
-    // We also need to take into account the fact that adding an htlc increases the commit tx fee if it's not trimmed.
-    if (!localParams.isFunder && balance >= receivedHtlcTrimThreshold(localParams.dustLimit, reduced)) {
-      (balance - htlcOutputFee(reduced.feeratePerKw)).max(0 msat)
+    val balanceNoFees = (reduced.toRemote - localParams.channelReserve).max(0 msat)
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -95,6 +95,7 @@ object Transactions {
    * these values are defined in the RFC
    */
   val commitWeight = 724
+  val htlcWeight = 172
```
pm47:
```suggestion
  val htlcOutputWeight = 172
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -104,7 +105,8 @@ trait StateTestsHelperMethods extends TestKitBase with fixture.TestSuite with Pa
     bob2blockchain.expectMsgType[WatchConfirmed] // deeply buried
     awaitCond(alice.stateName == NORMAL)
     awaitCond(bob.stateName == NORMAL)
-    assert(bob.stateData.asInstanceOf[DATA_NORMAL].commitments.availableBalanceForSend == (pushMsat - aliceParams.channelReserve).max(0 msat))
+    val htlcFee = Transactions.htlcFee(TestConstants.feeratePerKw)
```
pm47: Current feerate should be taken from `nodeParams`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -88,14 +88,17 @@ case class Commitments(channelVersion: ChannelVersion,
     // we need to base the next current commitment on the last sig we sent, even if we didn't yet receive their revocation
     val remoteCommit1 = remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit)
     val reduced = CommitmentSpec.reduce(remoteCommit1.spec, remoteChanges.acked, localChanges.proposed)
+    // The funder always pays the on-chain fees, so we must subtract that from the amount we can send.
     val feesMsat = if (localParams.isFunder) commitTxFee(remoteParams.dustLimit, reduced).toMilliSatoshi else 0.msat
-    (reduced.toRemote - remoteParams.channelReserve.toMilliSatoshi - feesMsat).max(0 msat)
+    // We also need to take into account the fact that adding an htlc increases the commit tx fee (regardless of which side is funding).
+    (reduced.toRemote - remoteParams.channelReserve.toMilliSatoshi - feesMsat - htlcFee(reduced.feeratePerKw)).max(0 msat)
```
pm47: `htlcFee` should only be paid by the funder.

Also depending on the amount available, the htlc may be trimmed and there wouldn't be any htlc fee

t-bast: Good point, I fixed that and I think this resolves the questions I add (I was confusing myself there, with the latest changes it should be what we want).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -88,14 +88,28 @@ case class Commitments(channelVersion: ChannelVersion,
     // we need to base the next current commitment on the last sig we sent, even if we didn't yet receive their revocation
     val remoteCommit1 = remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit)
     val reduced = CommitmentSpec.reduce(remoteCommit1.spec, remoteChanges.acked, localChanges.proposed)
+    // The funder always pays the on-chain fees, so we must subtract that from the amount we can send.
     val feesMsat = if (localParams.isFunder) commitTxFee(remoteParams.dustLimit, reduced).toMilliSatoshi else 0.msat
-    (reduced.toRemote - remoteParams.channelReserve.toMilliSatoshi - feesMsat).max(0 msat)
+    val balance = (reduced.toRemote - remoteParams.channelReserve.toMilliSatoshi - feesMsat).max(0 msat)
```
pm47: WDYT?

```scala
lazy val availableBalanceForSend: MilliSatoshi = {
    // we need to base the next current commitment on the last sig we sent, even if we didn't yet receive their revocation
    val remoteCommit1 = remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit)
    val reduced = CommitmentSpec.reduce(remoteCommit1.spec, remoteChanges.acked, localChanges.proposed)
    val commitFees = commitTxFee(remoteParams.dustLimit, reduced).toMilliSatoshi
    val htlcFees = htlcOutputFee(reduced.feeratePerKw)
    val balanceNoFees = reduced.toRemote - remoteParams.channelReserve.toMilliSatoshi
    if (localParams.isFunder) {
      // The funder always pays the on-chain fees, so we must subtract that from the amount we can send.
      if (balanceNoFees - commitFees < offeredHtlcTrimThreshold(remoteParams.dustLimit, reduced)) {
        // htlc will be trimmed
        balanceNoFees - commitFees
      } else {
        // htlc will have an output in the commitment tx, so there will be additional fees 
        balanceNoFees - commitFees - htlcFees
      }
    } else {
      // fundee doesn't pay on-chain fees
      balanceNoFees
    }
  }
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -88,14 +88,42 @@ case class Commitments(channelVersion: ChannelVersion,
...
     val remoteCommit1 = remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit)
     val reduced = CommitmentSpec.reduce(remoteCommit1.spec, remoteChanges.acked, localChanges.proposed)
-    val feesMsat = if (localParams.isFunder) commitTxFee(remoteParams.dustLimit, reduced).toMilliSatoshi else 0.msat
-    (reduced.toRemote - remoteParams.channelReserve.toMilliSatoshi - feesMsat).max(0 msat)
+    val balanceNoFees = (reduced.toRemote - remoteParams.channelReserve).max(0 msat)
+    if (localParams.isFunder) {
+      // The funder always pays the on-chain fees, so we must subtract that from the amount we can send.
+      val commitFees = commitTxFee(remoteParams.dustLimit, reduced).toMilliSatoshi
+      val htlcFees = htlcOutputFee(reduced.feeratePerKw)
+      if (balanceNoFees - commitFees < offeredHtlcTrimThreshold(remoteParams.dustLimit, reduced)) {
```
SomberNight: Why "offered" htlc?
`availableBalanceForSend` considers sending an HTLC. The commitment tx considered is the remote's. Hence it is a "received" HTLC.
`availableBalanceForReceive` considers receiving an HTLC. The commitment tx considered is the local's. Hence it is a "received" HTLC.
i.e. I believe both methods should be considering "received" HTLCs (in terms of on-chain outputs).

PR #1153, opened 2019-10-01
Multi part payments

t-bast:
This PR implements https://github.com/lightningnetwork/lightning-rfc/pull/643 (receive and send).

The PR spec is quite stable so I think we can start reviewing this and even merge to master soon-ish (since it doesn't require local/global feature bits, is turned off by default and we just did a release so we have time before the next one). This will make development of other features that depend on MPP easier. 

Pending:

- [x] Finalize spec PR (some TLV fields / timeout defaults / feature bits might change)
- [x] Test memory usage
- [x] Add Kamon metrics
- [ ] Test compatibility with CL / LND


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -88,7 +88,7 @@ trait Eclair {
 
   def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty)(implicit timeout: Timeout): Future[RouteResponse]
 
-  def sendToRoute(externalId_opt: Option[String], route: Seq[PublicKey], amount: MilliSatoshi, paymentHash: ByteVector32, finalCltvExpiryDelta: CltvExpiryDelta)(implicit timeout: Timeout): Future[UUID]
+  def sendToRoute(externalId_opt: Option[String], route: Seq[PublicKey], amount: MilliSatoshi, paymentHash: ByteVector32, finalCltvExpiryDelta: CltvExpiryDelta, invoice_opt: Option[PaymentRequest] = None)(implicit timeout: Timeout): Future[UUID]
```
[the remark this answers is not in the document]
t-bast: Yes, this allows using `sendToRoute` to send partial payments (especially useful for E2E tests when you want to stop the receiver or sender node in the middle of a multi-part payment)

eclair-core/src/main/scala/fr/acinq/eclair/Logs.scala
```diff
@@ -16,17 +16,20 @@
...
 
 object Logs {
 
-  def mdc(remoteNodeId_opt: Option[PublicKey] = None, channelId_opt: Option[ByteVector32] = None): MDC =
+  def mdc(remoteNodeId_opt: Option[PublicKey] = None, channelId_opt: Option[ByteVector32] = None, paymentId_opt: Option[UUID] = None): MDC =
     Seq(
       remoteNodeId_opt.map(n => "nodeId" -> s" n:$n"), // nb: we preformat MDC values so that there is no white spaces in logs
-      channelId_opt.map(c => "channelId" -> s" c:$c")
+      channelId_opt.map(c => "channelId" -> s" c:$c"),
+      paymentId_opt.map(p => "paymentId" -> s" p:$p")
```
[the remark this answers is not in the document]
t-bast: Right now I only use this in `PaymentLifecycle` and `MultiPartPaymentLifecycle` which are really created on a per-payment basis, so it doesn't have that issue. If I wanted that hint to be logged from other components it's true that it wouldn't always work, but it's quite useful at the `*PaymentLifecycle` level when grepping the logs.

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
```diff
@@ -0,0 +1,330 @@
...
+    // For other channels, we need to split the amount based on network statistics and pessimistic fees estimates.
+    val maxFeePct = request.routeParams.map(_.maxFeePct).getOrElse(nodeParams.routerConf.searchMaxFeePct)
+    val maxFeeBase = request.routeParams.map(_.maxFeeBase).getOrElse(nodeParams.routerConf.searchMaxFeeBase.toMilliSatoshi)
+    val channels = if (randomize) {
+      Random.shuffle(balances.filter(p => p.remoteNodeId != request.targetNodeId))
+    } else {
+      balances.filter(p => p.remoteNodeId != request.targetNodeId).sortBy(_.canSend)
+    }
+    channels.foldLeft((remaining, localPayments)) {
+      case ((MilliSatoshi(0), payments), _) => (0 msat, payments)
```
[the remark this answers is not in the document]
t-bast: We can't, this is pattern matching ;)

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
```diff
@@ -0,0 +1,330 @@
...
+      require(d.request.nonEmpty && d.sender.nonEmpty, "multi-part payment request must be set")
+      val r = d.request.get
+      s.stats match {
+        case Some(networkStats) =>
+          log.debug(s"network stats available: ${networkStats.capacity}")
+          relayer ! GetUsableBalances
+          goto(PAYMENT_IN_PROGRESS) using PaymentProgress(d.sender.get, r, networkStats, r.amount, r.maxAttempts, Map.empty, Nil)
+        case None =>
+          log.debug("network stats not available: retrying...")
+          router ! TickComputeNetworkStats
```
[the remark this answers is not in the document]
t-bast: It's okay-ish to attempt the payment without network stats, but it's likely better to do it with accurate network stats.

I updated the code to do without network stats if we don't have them, but trigger a computation on the router for potential next payments. But I think we can improve when network stats are computed, and I'm interested in your feedback on that. Right now I only compute network stats when we finish a sync (and I don't compute them from the "old" public channels data we had in the DB). Maybe it would make more sense to compute a first version of the network stats when the router boots from what's inside the DB, and then re-compute it once a sync completes?
If we do that maybe we don't need the `MultiPartPaymentLifecycle` to tell the router to re-compute?

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
```diff
@@ -0,0 +1,330 @@
...
+  import MultiPartPaymentLifecycle._
+
+  val paymentsDb = nodeParams.db.payments
+
+  startWith(PAYMENT_INIT, PaymentInit(None, None, None))
+
+  when(PAYMENT_INIT) {
+    case Event(r: SendPaymentRequest, d: PaymentInit) =>
+      require(d.request.isEmpty, "multi-part payment lifecycle must receive only one payment request")
+      router ! GetNetworkStats
```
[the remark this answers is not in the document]
t-bast: The issue with that is that it bleeds into every other downstream actor...
`MultiPartPaymentLifecycle` also needs to subscribe to stats and balance events while receiving an initial value for those. I quite like this separation of tasks where only one actor is responsible for keeping local balances up-to-date.

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
```diff
@@ -0,0 +1,330 @@
...
+            (remaining, payments)
+          }
+        } else {
+          val childCount = math.ceil(channel.canSend.min(remaining).toLong.toDouble / maxAmount.toLong).toInt
+          // We can't use all the available balance because we need to take fees into account and we don't know the
+          // exact fee before-hand because we don't know the rest of the route yet.
+          // Splitting into multiple HTLCs in the same channel will also increase the size of the CommitTx (and thus its
+          // fee), which decreases the available balance.
+          // We need to take that into account when trying to send multiple payments through the same channel.
+          val htlcsFees = Transactions.htlcOutputFee(nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(nodeParams.onChainFeeConf.feeTargets.commitmentBlockTarget)) * childCount
```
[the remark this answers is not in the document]
t-bast: That's true. That means we need to change a bit what's exposed in UsableBalances, I'll see how impactful that is.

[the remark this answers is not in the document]
t-bast: I'm deep down into changing this and it changes the `splitPayment` method a bit. I'm having some troubles because creating test `Commitments` objects is a bit painful, I'm trying to simplify it.
Please hold off reviewing the `splitPayment` method until I update it to use `Commitments`.

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -119,16 +132,22 @@ object OnionCodecs {
...
 
+  private val totalAmount: Codec[TotalAmount] = ("total_msat" | tu64overflow).xmap(amountMsat => TotalAmount(MilliSatoshi(amountMsat)), (a: TotalAmount) => a.amount.toLong)
+
   private val amountToForward: Codec[AmountToForward] = ("amount_msat" | tu64overflow).xmap(amountMsat => AmountToForward(MilliSatoshi(amountMsat)), (a: AmountToForward) => a.amount.toLong)
 
   private val outgoingCltv: Codec[OutgoingCltv] = ("cltv" | tu32).xmap(cltv => OutgoingCltv(CltvExpiry(cltv)), (c: OutgoingCltv) => c.cltv.toLong)
 
   private val outgoingChannelId: Codec[OutgoingChannelId] = (("length" | constant(hex"08")) :: ("short_channel_id" | shortchannelid)).as[OutgoingChannelId]
 
+  private val paymentSecret: Codec[PaymentSecret] = (("length" | constant(hex"20")) :: ("payment_secret" | bytes32)).as[PaymentSecret]
```
pm47: Why not use `varsize`-type codecs instead of hardcoding the length?

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentRequestSpec.scala
```diff
@@ -291,6 +297,38 @@ class PaymentRequestSpec extends FunSuite {
     assert(pr.features.bitmask === BitVector.empty)
   }
 
+  test("payment request features") {
```
pm47: test name isn't great

pm47: Should we test handling of unsupported feature bits

t-bast: Yes to both, good points ;)

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentHandler.scala
```diff
@@ -0,0 +1,143 @@
...
+  when(WAITING_FOR_HTLC) {
+    case Event(PaymentTimeout, d: WaitingForHtlc) =>
+      goto(PAYMENT_FAILED) using PaymentFailed(d.paidAmount, wire.PaymentTimeout, d.parts)
+
+    case Event(MultiPartHtlc(totalAmount2, htlc), d: WaitingForHtlc) =>
+      require(htlc.paymentHash == paymentHash, s"invalid payment hash (expected $paymentHash, received ${htlc.paymentHash}")
+      val pp = PartialPayment(htlc.amountMsat, htlc.channelId)
+      if (totalAmount != totalAmount2) {
+        log.warning(s"multi-part payment total amount mismatch: previously $totalAmount, now $totalAmount2")
+        goto(PAYMENT_FAILED) using PaymentFailed(d.paidAmount, IncorrectOrUnknownPaymentDetails(totalAmount2, nodeParams.currentBlockHeight), PendingPayment(htlc.id, pp, sender) :: d.parts)
```
pm47: I would have used `+:` here, probably doesn't matter much.

```suggestion
        goto(PAYMENT_FAILED) using PaymentFailed(d.paidAmount, IncorrectOrUnknownPaymentDetails(totalAmount2, nodeParams.currentBlockHeight), PendingPayment(htlc.id, pp, sender) +: d.parts)
```

pm47: A rewrite with less repetition:

```scala
case Event(mph@MultiPartHtlc(_, htlc), d: WaitingForHtlc) =>
  require(htlc.paymentHash == paymentHash, s"invalid payment hash (expected $paymentHash, received ${htlc.paymentHash}")
  val paidAmount1 = d.paidAmount + htlc.amountMsat
  val part = PendingPayment(htlc.id, PartialPayment(htlc.amountMsat, htlc.channelId), sender)
  val parts1 = part +: d.parts
  if (totalAmount != mph.totalAmount) {
    log.warning(s"multi-part payment total amount mismatch: previously $totalAmount, now ${mph.totalAmount}")
    goto(PAYMENT_FAILED) using PaymentFailed(d.paidAmount, IncorrectOrUnknownPaymentDetails(mph.totalAmount, nodeParams.currentBlockHeight), parts1)
  } else if (paidAmount1 >= totalAmount) {
    goto(PAYMENT_SUCCEEDED) using PaymentSucceeded(paidAmount1, parts1)
  } else {
    stay using d.copy(paidAmount = paidAmount1, parts = parts1)
  }
```

pm47: NB: I think in the failure case we should arguably return `paidAmount1`:

```scala
goto(PAYMENT_FAILED) using PaymentFailed(paidAmount1, IncorrectOrUnknownPaymentDetails(mph.totalAmount, nodeParams.currentBlockHeight), parts1)

```

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentHandler.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  // @formatter:off
+  sealed trait State
+  case object WAITING_FOR_HTLC extends State
+  case object PAYMENT_SUCCEEDED extends State
+  case object PAYMENT_FAILED extends State
+  // @formatter:on
+
+  // @formatter:off
+  sealed trait Data
```
pm47: You can factor the `parts` in the trait, and use a `lazy val` to keep the paid amount:

```scala
// @formatter:off
sealed trait Data {
  def parts: List[PendingPayment]
  lazy val paidAmount = parts.map(_.payment.amount).sum
}
case class WaitingForHtlc(parts: List[PendingPayment]) extends Data
case class PaymentSucceeded(parts: List[PendingPayment]) extends Data
case class PaymentFailed(failure: FailureMessage, parts: List[PendingPayment]) extends Data
// @formatter:on
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentHandler.scala
```diff
@@ -0,0 +1,143 @@
...
+    // The LocalPaymentHandler will create a new instance of MultiPartPaymentHandler to handle a new attempt.
+    case Event(MultiPartHtlc(_, htlc), PaymentFailed(_, failure, _)) =>
+      require(htlc.paymentHash == paymentHash, s"invalid payment hash (expected $paymentHash, received ${htlc.paymentHash}")
+      parent ! ExtraHtlcReceived(paymentHash, PendingPayment(htlc.id, PartialPayment(htlc.amountMsat, htlc.channelId), sender), Some(failure))
+      stay
+
+    case Event("ok", _) => stay
+  }
+
+  onTransition {
```
pm47: You can have multiple `onTransition`, there are additive and it is helpful here to cancel the timer:

```scala
  onTransition {
    case WAITING_FOR_HTLC -> _ => cancelTimer(PaymentTimeout.toString)
  }

  onTransition {
    ...
  }

```

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentHandler.scala
```diff
@@ -0,0 +1,143 @@
...
+    case _ -> PAYMENT_SUCCEEDED =>
+      cancelTimer(PaymentTimeout.toString)
+      nextStateData match {
+        case PaymentSucceeded(_, parts) =>
+          // We expect the parent actor to send us a PoisonPill after receiving this message.
+          parent ! MultiPartHtlcSucceeded(paymentHash, parts.reverse)
```
pm47: I don't quite understand why you didn't send `MultiPartHtlcSucceeded`/`MultiPartHtlcFailed` right before leaving the `WAITING_FOR_HTLC` state:

```scala
if (totalAmount != totalAmount2) {
        log.warning(s"multi-part payment total amount mismatch: previously $totalAmount, now $totalAmount2")
        parent ! MultiPartHtlcFailed(paymentHash, failure, parts.reverse)
        goto(PAYMENT_FAILED) using PaymentFailed(d.paidAmount, IncorrectOrUnknownPaymentDetails(totalAmount2, nodeParams.currentBlockHeight), PendingPayment(htlc.id, pp, sender) :: d.parts)
      } else if (htlc.amountMsat + d.paidAmount >= totalAmount) {
        parent ! MultiPartHtlcSucceeded(paymentHash, parts.reverse)
        goto(PAYMENT_SUCCEEDED) using PaymentSucceeded(htlc.amountMsat + d.paidAmount, PendingPayment(htlc.id, pp, sender) :: d.parts)
      } else {
        stay using d.copy(paidAmount = d.paidAmount + htlc.amountMsat, parts = PendingPayment(htlc.id, pp, sender) :: d.parts)
      }
```

pm47: Using a `Queue` for `parts` would allow an O(1) append and remove the need for `parts.reverse`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentHandler.scala
```diff
@@ -0,0 +1,143 @@
...
+    // A sender should not send us additional htlcs for that payment once we've already reached the total amount.
+    // However if that happens the only rational choice is to fulfill it, because the pre-image has been released so
+    // intermediate nodes will be able to fulfill that htlc anyway. This is a harmless spec violation.
+    case Event(MultiPartHtlc(_, htlc), _) =>
+      require(htlc.paymentHash == paymentHash, s"invalid payment hash (expected $paymentHash, received ${htlc.paymentHash}")
+      log.info(s"received extraneous htlc for payment hash $paymentHash")
+      parent ! ExtraHtlcReceived(paymentHash, PendingPayment(htlc.id, PartialPayment(htlc.amountMsat, htlc.channelId), sender), None)
+      stay
+
+    case Event("ok", _) => stay
```
pm47: This should be factorized in a `whenUnhandled`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentHandler.scala
pm47: Not sure if states `PAYMENT_SUCCEEDED`/`PAYMENT_FAILED` are really needed. I would just `stop()` the FSM with a `Normal`/`Failure` reason. Extra htlcs would be just failed, which isn't a bad thing I think?

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
```diff
@@ -0,0 +1,407 @@
...
+        stay using d.copy(toSend = d.toSend + failedPayment.finalPayload.amount, pending = d.pending - pf.id, failures = d.failures ++ pf.failures)
+      }
+
+    case Event(ps: PaymentSent, d: PaymentProgress) =>
+      require(ps.parts.length == 1, "child payment must contain only one part")
+      // As soon as we get the preimage we can consider that the whole payment succeeded (we have a proof of payment).
+      goto(PAYMENT_SUCCEEDED) using PaymentSucceeded(d.sender, d.request, ps.paymentPreimage, ps.parts, d.pending.keySet - ps.id)
+  }
+
+  when(PAYMENT_ABORTED) {
```
pm47: I'm not sure the `PAYMENT_ABORTED`/`PAYMENT_SUCCEEDED` distinction really make sense, because, as you have noticed, we can receive a payment proof even in the failure state.

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
```diff
@@ -0,0 +1,407 @@
...
+                                 maxChildAmount: MilliSatoshi,
+                                 maxFeeBase: MilliSatoshi,
+                                 maxFeePct: Double,
+                                 channel: OutgoingChannel,
+                                 channelCommitments: Commitments,
+                                 channelPayments: Seq[SendPayment]): Seq[SendPayment] = {
+    // We can't use all the available balance because we need to take the fees for each child payment into account and
+    // we don't know the exact fee before-hand because we don't know the rest of the route yet (so we assume the worst
+    // case where the max fee is used).
+    val previousFees = channelPayments.map(p => maxFeeBase.max(p.finalPayload.amount * maxFeePct))
```
pm47: Seems like the fee calculation is wrong, base+proportional should be added together:

```suggestion
    val previousFees = channelPayments.map(p => maxFeeBase + p.finalPayload.amount * maxFeePct)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
```diff
@@ -62,6 +78,18 @@ object PaymentInitiator {
...
+    def finalExpiry(currentBlockHeight: Long) = finalExpiryDelta.toCltvExpiry(currentBlockHeight + 1)
+  }
+
+  case class SendPaymentConfig(id: UUID,
+                               parentId: UUID,
+                               externalId: Option[String],
+                               paymentHash: ByteVector32,
+                               targetNodeId: PublicKey,
+                               paymentRequest: Option[PaymentRequest],
+                               storeInDb: Boolean,
```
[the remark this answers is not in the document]
pm47: Ok, maybe because we don't want to store child payments, but if so why not check the `parent_id`?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -78,7 +78,7 @@ trait Eclair {
 
   def peersInfo()(implicit timeout: Timeout): Future[Iterable[PeerInfo]]
 
-  def receive(description: String, amount_opt: Option[MilliSatoshi], expire_opt: Option[Long], fallbackAddress_opt: Option[String], paymentPreimage_opt: Option[ByteVector32])(implicit timeout: Timeout): Future[PaymentRequest]
+  def receive(description: String, amount_opt: Option[MilliSatoshi], expire_opt: Option[Long], fallbackAddress_opt: Option[String], paymentPreimage_opt: Option[ByteVector32], allowMultiPart: Boolean)(implicit timeout: Timeout): Future[PaymentRequest]
```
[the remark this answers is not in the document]
t-bast: I think we want it to be opt-in at first to allow us to have "reckless" users test it and report any bugs before we turn it on by default.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -234,12 +234,20 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
...
-      statement.setLong(1, amount.toLong)
-      statement.setLong(2, receivedAt)
-      statement.setBytes(3, paymentHash.toArray)
-      val res = statement.executeUpdate()
-      if (res == 0) throw new IllegalArgumentException("Inserted a received payment without having an invoice")
+    using(sqlite.prepareStatement("SELECT received_msat FROM received_payments WHERE payment_hash = ?"), inTransaction = true) { query =>
+      query.setBytes(1, paymentHash.toArray)
+      val rs = query.executeQuery()
+      if (rs.next()) {
+        val receivedAmount = getNullableLong(rs, "received_msat").map(_ + amount.toLong).getOrElse(amount.toLong)
```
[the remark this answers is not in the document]
t-bast: Nice, didn't know about COALESCE, I'll test it out.

eclair-core/src/main/scala/fr/acinq/eclair/payment/LocalPaymentHandler.scala
```diff
@@ -55,37 +64,116 @@ class LocalPaymentHandler(nodeParams: NodeParams) extends Actor with ActorLoggin
...
-              context.system.eventStream.publish(PaymentReceived(htlc.paymentHash, PaymentReceived.PartialPayment(htlc.amountMsat, htlc.channelId) :: Nil))
-          }
+    case p: FinalPayload => paymentDb.getIncomingPayment(p.add.paymentHash) match {
+      case Some(record) => validatePayment(p, record) match {
+        case Some(cmdFail) =>
+          sender ! cmdFail
         case None =>
-          sender ! CMD_FAIL_HTLC(htlc.id, Right(IncorrectOrUnknownPaymentDetails(htlc.amountMsat, nodeParams.currentBlockHeight)), commit = true)
+          log.info(s"received payment for paymentHash=${p.add.paymentHash} amount=${p.add.amountMsat} totalAmount=${p.payload.totalAmount}")
+          pendingPayments.get(p.add.paymentHash) match {
```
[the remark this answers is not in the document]
t-bast: This has already been fixed in latest commit.

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentHandler.scala
t-bast: I think the state machine is easier to reason about (and avoid hidden race conditions) with the `PAYMENT_SUCCEEDED / PAYMENT_FAILED` states. It allows us to easily log these small spec violations to detect bugs in sender implementations.

That's also why I'm sending the response to the parent in the `onTransition`: it factors that into a single place which is simpler than having it in potentially multiple `if` branches if later commits update this FSM.

In the success case I think we really want to fulfill additional HTLCs, not fail them. They indicate a sender bug, but the only way we can refund the user is if we fulfilled the extra payments (otherwise an intermediate node will claim them but we can't refund because we didn't accept them).

t-bast: Given the number of HTLC in a set I think this is really overkill :)
We need proper perf testing for that but I'd even argue that it's likely that for small arrays like this a `Queue` will end-up being _more_ costly.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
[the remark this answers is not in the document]
t-bast: For trampoline you never want to store in the DB when you're a relaying node. Or maybe in the future after a refactoring we won't want to store multi-part child payments (we currently do store them when we're the sender).
We could use `parentId != id` as an implicit signal, but I prefer an explicit signal, otherwise users of this class will have to know that this is implicitly inferred and take it into account for their use-case.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
```diff
@@ -62,6 +78,18 @@ object PaymentInitiator {
                                 externalId: Option[String] = None,
                                 predefinedRoute: Seq[PublicKey] = Nil,
                                 assistedRoutes: Seq[Seq[ExtraHop]] = Nil,
-                                routeParams: Option[RouteParams] = None)
+                                routeParams: Option[RouteParams] = None) {
+    // We add one block in order to not have our htlcs fail when a new block has just been found.
+    def finalExpiry(currentBlockHeight: Long) = finalExpiryDelta.toCltvExpiry(currentBlockHeight + 1)
+  }
+
+  case class SendPaymentConfig(id: UUID,
```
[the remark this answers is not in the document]
t-bast: IMO this would just defer that code to somewhere else. Since in the DB we want to have two separate columns for those, if we use an Option here or a trait, the DB will have to know that and do the proper mapping when reading from the table...WDYT?

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
```diff
@@ -0,0 +1,407 @@
...
+
+  import MultiPartPaymentLifecycle._
+
+  require(cfg.id == cfg.parentId, "multi-part payment cannot have a parent payment")
+
+  val id = cfg.id
+
+  startWith(PAYMENT_INIT, PaymentInit(None, None, None))
+
+  when(PAYMENT_INIT) {
```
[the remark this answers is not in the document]
t-bast: That's true, let's change that. It's probably a leftover because I used to do the `GetNetworkStats` call in parallel of waiting for the first HTLC.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentInitiator.scala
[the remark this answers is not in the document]
pm47: Ok, let's add a comment then:

```suggestion
                               storeInDb: Boolean, // e.g. for trampoline we don't want to store in the DB when we're relaying payments
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
```diff
@@ -0,0 +1,407 @@
...
+      // We tell the router to compute those stats though: in case our payment attempt fails, they will be available for
+      // another payment attempt.
+      if (s.stats.isEmpty) {
+        router ! TickComputeNetworkStats
+      }
+      relayer ! GetOutgoingChannels()
+      goto(PAYMENT_IN_PROGRESS) using PaymentProgress(d.sender.get, r, s.stats, r.totalAmount, r.maxAttempts, Map.empty, Nil)
+  }
+
+  when(PAYMENT_IN_PROGRESS) {
```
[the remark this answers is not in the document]
t-bast: If we split that into two states, we need to handle the `PaymentFailed` and `PaymentSent` almost exactly the same way (but with a very small difference on whether to ask for updated balances or not) in both, do you think it's ok to have that duplication? I agree that the state machine feels cleaner with two states, but I didn't like the duplication...

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
t-bast: But if we have a single state for both we'll have ugly if/switch everywhere to be able to handle the "spec violation" case (and it feels a bit cleaner to express the difference via different states, because you can move from `PAYMENT_ABORTED` to `PAYMENT_SUCCEEDED` but not the other way around.

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
t-bast: This isn't the actual fee (because we don't know it yet since we haven't sent the payment), it's the worst-case estimation that we may pay the maximum fee. So it's either the maximum by flat max or by percentage max.

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
```diff
@@ -0,0 +1,407 @@
...
+    case Event(OutgoingChannels(channels), d: PaymentProgress) if d.toSend > 0.msat =>
```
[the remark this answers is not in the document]
pm47: fair enough

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
pm47: Got it

eclair-core/src/main/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
pm47: See https://github.com/ACINQ/eclair/pull/1202/commits/0d019864a778b934eb0b5c8dbfc77b92f1c7fe26

eclair-node/src/main/resources/logback.xml
```diff
@@ -35,7 +35,7 @@
             <totalSizeCap>5GB</totalSizeCap>
         </rollingPolicy>
         <encoder>
-            <pattern>%d %-5level %logger{24} %X{nodeId}%X{channelId} - %msg%ex{24}%n</pattern>
+            <pattern>%d %-5level %logger{24} %X{nodeId}%X{channelId}%X{parentPaymentId}%X{paymentId} - %msg%ex{24}%n</pattern>
```
pm47:
```suggestion
            <pattern>%d %-5level %logger{24}%logger{24}%X{nodeId}%X{channelId}%.-11X{parentPaymentId}%.-11X{paymentId} - %msg%ex{24}%n</pattern>
```

PR #1157, opened 2019-10-03
Upgrade new unit tests to bitcoin 0.18.1 API

sstone:
We had 2 open PRs, one that added new tests using the 0.API, one that switched to 0.18.1, when they
were merged the new tests failed since they had not been upgraded....


pm47:
The README hasn't been updated: we should require bitcoind 0.18.1, which means that we don't officially support 0.17, even if it happens to work.

PR #1159, opened 2019-10-03
Sqlite: use TEXT type for strings

sstone:
STRING type does not exist and is mapped to NUMERIC, which will strip leading 0s. It's a problem for the `txid` column in our network db which we read back as a `ByteVector32`, but a minor one since the network db can be deleted  (it's not backed up for example).
There's no migration here because it's not possible to change the type of a sqlite3 column, but again since it only affects our network db it's a minor issue.


sstone:
Additional info:
This problem (truncation of leading 0s in txid) seems to only impact txids which have an hex representation that only includes digits, not letters (i.e `0000000000000000000000000000000000000000000000000000000000000001` would be truncated but not `000000000000000000000000000000000000000000000000000000000000000a`). 

The probability of having a such a txid is (10/16)^64 which is about 10^-13, which is why we never saw this (and probably never will) in production, only when testing with special values (all 0s, "01010101..." etc...)

eclair-core/src/test/scala/fr/acinq/eclair/db/SqliteNetworkDbSpec.scala
```diff
@@ -165,6 +173,34 @@ class SqliteNetworkDbSpec extends FunSuite {
     simpleTest(sqlite)
   }
 
+  test("non-regression test") {
```
pm47: This is a lot of code for just making sure that creating a table with type `STRING` and then reading as `TEXT` works.

Also, test names matter.

The following pseudo code would have worked the same and would not hide the actual test behind a layer of complexity:

```scala
test("non-regression test for changing column type STRING->TEXT") {
  val sqlite = TestConstants.sqliteInMemory()
  statement.execute("CREATE TABLE IF NOT EXISTS test (txid STRING NOT NULL))
  val txids = for (i <- 0 until 1000) random
  txids.foreach(INSERT INTO...)
  statement.execute("CREATE TABLE IF NOT EXISTS test (txid TEXT NOT NULL))
  val res = statement.execute("SELECT txids")
  assert(txids == res)
}
```

sstone: I simplified the test, and add a specific check to verify that creating a table that already exists but with different column types is simply ignored.

sstone:
Column type is stored in the sqlite database, so for an existing database it will remain unchanged (`CREATE TABLE IF NOT EXIST` will be ignored even if we specify different column types). 

How we read and convert data is not linked to how the table was created but defined in the code we use to actually read data, and we always use `rs.getString`  which is mapped to `sqlite3_column_text` which will basically return raw data.  

=> Existing databases will keep their column type as `NUMERIC`. In practice this won't be an issue as explained above, and we don't need to migrate their data.

PR #1161, opened 2019-10-03
Add execution time limit

t-bast:
We have a few flaky tests that sometimes block the build.
The default expiry in that case is 1 hour, so we pay for a semaphore VM for no reason.
We should fail earlier than that (15 minutes sounds reasonable).
We should also fix the flaky tests, but that's for another PR :innocent:

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -25,5 +25,5 @@ blocks:
             - checkout
             - cache restore maven
             - mvn scoverage:report
-            - bash <(curl -s https://codecov.io/bash) -t $CODECOV_TOKEN -v
+            - bash <(curl -s https://codecov.io/bash) -t $CODECOV_TOKEN
             - cache store maven .m2
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -25,5 +25,5 @@ blocks:
             - checkout
             - cache restore maven
             - mvn scoverage:report
-            - bash <(curl -s https://codecov.io/bash) -t $CODECOV_TOKEN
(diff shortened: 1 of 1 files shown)


araspitzu:
Somehow related: codecov is not pushing anymore the coverage reports, #1134 makes it upload them but they never get pushed to the PR.

t-bast:
> Somehow related: codecov is not pushing anymore the coverage reports

That's true! Maybe it's something that an admin of the repo can configure...

t-bast:
> #1134 makes it upload them but they never get pushed to the PR.

In fact I think this doesn't work. When connecting to codecov it doesn't show recent commits...
There's probably something broken with the codecov upload :/

araspitzu:
> > #1134 makes it upload them but they never get pushed to the PR.
> 
> In fact I think this doesn't work. When connecting to codecov it doesn't show recent commits...
> There's probably something broken with the codecov upload :/

Well there is this in the end of the semaphore build https://pm47.semaphoreci.com/jobs/46a90cad-1b7e-4b00-9f42-de24e2234a39#L1462

Christewart:
Something you might find useful is `AsyncTimeLimited` and `TimeLimited` as well. We just started using these recently and they can apply for local development as well. 

https://github.com/bitcoin-s/bitcoin-s/pull/767/files#diff-564dba04c892d9e86ea674a4c9b60ec8R24

If you have an over generic eclair test trait that all of your unit tests extend, it probably is worth mixing this in :-)

t-bast:
> Something you might find useful is AsyncTimeLimited and TimeLimited as well.

Nice, we'll have a look at it (probably when we do fix the flaky test...), thanks for the suggestion!

PR #1162, opened 2019-10-03
Update android branch

sstone:


sstone:
There are lots of changes that impact many files, but areas that require particular attention are:
- network db: new structure is different from the old one, and instead of upgrading we just drop tables which means that users will need to re-sync their network db after they've upgraded. That's what we did before with channel updates anyway
- the impact of the extended optional queries: we need the `chain hash` field to compute channel update checksums, which we used to strip. It is still stripped when we store data but we now provide it when we load data from back the db, and it does not feel consistent and stripping/restoring stripped fields should probably be done in the router  
- payment lifecycle which has changed a lot

PR #1163, opened 2019-10-03
Value in flight fix

btcontract:
`reduced.htlcs` is a `Set`, we could have different HTLCs with same value in it, when mapped to `htlcValueInFlight` it could look like `Set(100 msat, 100 msat).sum` which would be reduced to just `Set(100 msat).sum`


pm47:
They can't be identical though, because they have a unique identifier?

btcontract:
@pm47 yes, but if their amounts are the same they become identical here: https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala#L191

t-bast:
Good catch! :scream:

araspitzu:
Great catch!

pm47:
can you merge from master just to trigger a new build?

btcontract:
Done, what do you think about replacing current solution with keeping `outgoingHtlcs` as `Set` and then
```scala
outgoingHtlcs.foldLeft(0.msat)(_ + _.add.amountMsat)
```
Seems more explicit to me.

pm47:
I like it better, but it needs a comment explaining why we can't simply sum.

Actually what's missing here is a comment that reproduces the issue and would serve as a non-reg in the future.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -188,7 +188,8 @@ object Commitments {
       }
     }
 
-    val htlcValueInFlight = outgoingHtlcs.map(_.add.amountMsat).sum
+    // We can't obtain value in flight by `outgoingHtlcs.map(_.add.amountMsat).sum` because `outgoingHtlcs` is a Set where HTLCs may contain identical amounts and Set does not allow duplicates
+    val htlcValueInFlight = outgoingHtlcs.foldLeft(0 msat)(_ + _.add.amountMsat)
```
pm47: How about this? It is compact enough to be duplicated on line 233 and being in-lined has less risk of being accidentally removed.

```suggestion
    val htlcValueInFlight = outgoingHtlcs.foldLeft(0 msat)(_ + _.add.amountMsat) // can't use .map(...).sum on a set because duplicates would be removed
```

t-bast:
Thanks @btcontract looks good.
For some reason semaphore doesn't pick up the `push` build, blocking the merge.
Could you merge master or squash-rebase on top of master to trigger semaphore?

t-bast:
@btcontract it seems semaphore CI is completely buggy and didn't handle correctly the fact that we whitelisted you for builds *after* this PR was created - and that blocks merging this PR.
Could you close this PR and push this fix in a new one, **with a signed commit** (this isn't currently the case either)?
I can do that for you if you want, but you should have the credit for finding this bug ;)

btcontract:
Please go on, it's fine with me. I'll set up signing later, busy with other stuff currently.

t-bast:
Thanks, superceded by #1180

PR #1164, opened 2019-10-04
Electrum improvements

sstone:
- electrum client: downgrade log level for some connection events from INFO to DEBUG
- electrum watcher: don't ask for Merkle Proofs for unconfirmed txs (which by definition cannot exist)

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumClient.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumClient.scala
@@ -97,7 +97,7 @@ class ElectrumClient(serverAddress: InetSocketAddress, ssl: SSL)(implicit val ec
   })
 
   // Start the client.
-  log.info("connecting to server={}", serverAddress)
+  log.debug("connecting to server={}", serverAddress)
 
   val channelOpenFuture = b.connect(serverAddress.getHostName, serverAddress.getPort)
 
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumClient.scala
```diff
@@ -92,7 +92,7 @@ class ElectrumClient(serverAddress: InetSocketAddress, ssl: SSL)(implicit val ec
   })
 
   // Start the client.
-  log.info("connecting to server={}", serverAddress)
+  log.debug("connecting to server={}", serverAddress)
```
pm47: Not convinced that hiding connection issues is the right thing to do.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -867,7 +867,7 @@ object ElectrumWallet {
           val confirmedBalance = confirmedReceived.map(_.amount).sum - confirmedSpents.map(_.amount).sum
           val unconfirmedBalance = unconfirmedReceived.map(_.amount).sum - unconfirmedSpents.map(_.amount).sum
 
-          logger.debug(s"scriptHash=$scriptHash confirmedBalance=$confirmedBalance unconfirmedBalance=$unconfirmedBalance)")
+          logger.trace(s"scriptHash=$scriptHash confirmedBalance=$confirmedBalance unconfirmedBalance=$unconfirmedBalance)")
```
pm47: This is the only `trace` in the entire code base. Is it really such a special case that we need it here and not anywhere else?

PR #1170, opened 2019-10-10
Add bech32 support for electrum wallet

araspitzu:
Add support for `p2wpkh` scriptPubKeys in our electrum wallet, this will enable eclair-mobile to receive to bech32 addresses. The electrum wallet now accepts a `WalletType` parameter that is used to decide whether the wallet creates outputs of type `p2sh-of-p2wpkh` (current type) or the new `p2wpkh`, the wallet only supports one of the two and they can't be mixed, if started with `p2wpkh` it won't find unspent outputs of the other type.


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -252,7 +253,7 @@ class ElectrumWallet(seed: ByteVector, client: ActorRef, params: ElectrumWallet.
     case Event(ElectrumClient.ScriptHashSubscriptionResponse(scriptHash, status), data) =>
       val key = data.accountKeyMap.getOrElse(scriptHash, data.changeKeyMap(scriptHash))
       val isChange = data.changeKeyMap.contains(scriptHash)
-      log.info(s"received status=$status for scriptHash=$scriptHash key=${segwitAddress(key, chainHash)} isChange=$isChange")
+      log.info(s"received status=$status for scriptHash=$scriptHash keyIndex=${key.path.lastChildNumber}  isChange=$isChange")
```
araspitzu: nit: this is a less descriptive than logging the actual address, although it's possible to reconstruct the address given the seed and this log line

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumAddressStrategy.scala
```diff
@@ -0,0 +1,172 @@
...
+  override def computeAddress(key: PublicKey, chainHash: ByteVector32): String = {
+    val script = Script.pay2wpkh(key)
+    val hash = Crypto.hash160(Script.write(script))
+    chainHash match {
+      case Block.RegtestGenesisBlock.hash | Block.TestnetGenesisBlock.hash => Base58Check.encode(Base58.Prefix.ScriptAddressTestnet, hash)
+      case Block.LivenetGenesisBlock.hash => Base58Check.encode(Base58.Prefix.ScriptAddress, hash)
+    }
+  }
+
+  def addUtxos(tx: Transaction, utxos: Seq[Utxo]): Transaction = {
```
sstone: previous name `addUtxosWithDummySig` was more explicit

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -55,8 +55,8 @@ class ElectrumWallet(seed: ByteVector, client: ActorRef, params: ElectrumWallet.
 
   val master = DeterministicWallet.generate(seed)
 
-  val accountMaster = accountKey(master, chainHash)
-  val changeMaster = changeKey(master, chainHash)
+  val accountMaster = accountKey(master, rootPath(walletType, chainHash))
+  val changeMaster = changeKey(master, rootPath(walletType, chainHash))
```
sstone: This should be handled by the key store/strategy

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -511,8 +512,8 @@ object ElectrumWallet {
   case object GetBalance extends Request
   case class GetBalanceResponse(confirmed: Satoshi, unconfirmed: Satoshi) extends Response
 
-  case object GetXpub extends Request
-  case class GetXpubResponse(xpub: String, path: String) extends Response
+  case object GetRootPub extends Request
```
sstone: We should keep the old name. Xpub is widely used even when it's actually a Upub/Zpub etc... and conveys more meaning.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumAddressStrategy.scala
```diff
@@ -0,0 +1,172 @@
...
+      val utxo = d.utxos.find(_.outPoint == txIn.outPoint).getOrElse(throw new RuntimeException(s"cannot sign input that spends from ${txIn.outPoint}"))
+      val key = utxo.key
+      val sig = Transaction.signInput(tx, i, Script.pay2pkh(key.publicKey), SIGHASH_ALL, Satoshi(utxo.item.value), SigVersion.SIGVERSION_WITNESS_V0, key.privateKey)
+      val sigScript = ByteVector.empty
+      val witness = ScriptWitness(sig :: key.publicKey.value :: Nil)
+      txIn.copy(signatureScript = sigScript, witness = witness)
+    })
+  }
+
+  override def addUtxos(tx: Transaction, utxos: Seq[Utxo]): Transaction = {
```
sstone: We should keep the previous name `addUtxosWithDummySig` which is more explicit (here is not obvious that the output tx contains dummy wittness data)

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWalletBasicSpec.scala
```diff
@@ -76,24 +87,52 @@ class ElectrumWalletBasicSpec extends FunSuite with Logging {
...
-
-  test("compute addresses") {
+  test("compute addresses") { f =>
     val priv = PrivateKey.fromBase58("cRumXueoZHjhGXrZWeFoEBkeDHu2m8dW5qtFBCqSAt4LDR2Hnd8Q", Base58.Prefix.SecretKeyTestnet)._1
     assert(Base58Check.encode(Base58.Prefix.PubkeyAddressTestnet, priv.publicKey.hash160) == "ms93boMGZZjvjciujPJgDAqeR86EKBf9MC")
-    assert(segwitAddress(priv, Block.RegtestGenesisBlock.hash) == "2MscvqgGXMTYJNAY3owdUtgWJaxPUjH38Cx")
+    assert(p2shStrategy.computeAddress(priv, Block.RegtestGenesisBlock.hash) == "2MscvqgGXMTYJNAY3owdUtgWJaxPUjH38Cx")
+  }
+
+  test("derive bip84 keys", Tag("bech32")) { f =>
```
sstone: These tests should be move to specific unit tests for BIP49/BIP84 key stores

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWalletBasicSpec.scala
```diff
@@ -76,24 +87,52 @@ class ElectrumWalletBasicSpec extends FunSuite with Logging {
...
+    val seed = MnemonicCode.toSeed("abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about", passphrase = "")
+    val master = DeterministicWallet.generate(seed)
+    val (accountZpub, _) = nativeSegwitStrategy.computeRootPub(master, Block.LivenetGenesisBlock.hash)
+    assert(accountZpub == "zpub6rFR7y4Q2AijBEqTUquhVz398htDFrtymD9xYYfG1m4wAcvPhXNfE3EfH1r1ADqtfSdVCToUG868RvUUkgDKf31mGDtKsAYz2oz2AGutZYs") // m/84'/0'/0'
+
+    val firstReceivingKey = DeterministicWallet.derivePrivateKey(accountKey(master, rootPath(f.state.walletType, Block.LivenetGenesisBlock.hash)), 0)
+    assert(firstReceivingKey.publicKey.value == hex"0330d54fd0dd420a6e5f8d3624f5f3482cae350f79d5f0753bf5beef9c2d91af3c")
+  }
+
+  test("compute bech32 addresses", Tag("bech32")) { f =>
```
sstone: These tests should be move to specific unit tests for BIP49/BIP84 key stores

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWalletBasicSpec.scala
```diff
@@ -76,24 +87,52 @@ class ElectrumWalletBasicSpec extends FunSuite with Logging {
...
+  }
+
+  // from https://github.com/spesmilo/electrum/blob/master/electrum/tests/test_wallet.py#L218
+  test("electrum bech32 compatibility") { f =>
+    val (privKey, _) = PrivateKey.fromBase58("L4jkdiXszG26SUYvwwJhzGwg37H2nLhrbip7u6crmgNeJysv5FHL", Base58.Prefix.SecretKey)
+    assert(nativeSegwitStrategy.computeAddress(privKey.publicKey, Block.LivenetGenesisBlock.hash) == "bc1q2ccr34wzep58d4239tl3x3734ttle92a8srmuw")
   }
 
-  test("implement BIP49") {
+  test("implement BIP49") { f =>
```
sstone: These tests should be move to specific unit tests for BIP49/BIP84 key stores

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -557,77 +558,51 @@ object ElectrumWallet {
...
-   * @param key public key
-   * @return the hash of the public key script for this key, as used by Electrum's hash-based methods
-   */
-  def computeScriptHashFromPublicKey(key: PublicKey): ByteVector32 = Crypto.sha256(Script.write(computePublicKeyScript(key))).reverse
+    * We define a root path the first 3 elements in the BIP44 derivation scheme, which include
+    * the purpose, coin type (testnet/mainnet) and account number, example: m/purpose'/coin_type'/account'.
+    * We always use the first account.
+    * @return the root path
+    */
+  def rootPath(walletType: WalletType, chainHash: ByteVector32) = walletType match {
```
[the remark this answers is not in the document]
araspitzu: It's quite hard to move it in the strategy since that _lives_ in `ElectrumWallet.Data`, this causes  a deadlock-like initialization error because [here](https://github.com/ACINQ/eclair/pull/1170/files#diff-2d0a62c2e3bc4b5904735f7004030ce7R115) we read the accountMaster/changeMaster before initializing the data. I'm moving the implementation into companion objects of the key store.

sstone:
@araspitzu for reference I've implemented some of the proposed changes in branch https://github.com/ACINQ/eclair/tree/native_segwit_support_fd, including a different way of having specific BIP49/BIP84 tests.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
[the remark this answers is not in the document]
sstone: Yes it's because some of the initial setup (deriving the first keys etc...) copies wallet logic instead of simply asking the wallet to do the actual work, I've a proposal to improve that (see comment below)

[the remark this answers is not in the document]
araspitzu: Nice changes, i thought it was overkill to rework the initialization but it's a more elegant solution (and logically the accountPath goes in the key store). I will integrate that and the parametrized test as well

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/db/sqlite/SqliteWalletDb.scala
```diff
@@ -197,13 +197,18 @@ object SqliteWalletDb {
...
+  val legacyVersion = 0x0000
+  val bech32Version = 0x0001
+
+  val walletTypeCodec = discriminated[WalletType].by(uint32)
+    .typecase(legacyVersion, provide(P2SH_SEGWIT))
+    .typecase(bech32Version, provide(NATIVE_SEGWIT))
 
   val persistentDataCodec: Codec[PersistentData] = (
-    ("version" | constant(BitVector.fromInt(version))) ::
+    ("walletType" | walletTypeCodec) ::
```
sstone: `version` here is not supposed to encoded a specific property, but the version of the persistence format. Changing it means that legacy cannot be read, and in this case the onchain wallet behaves as if it had been restored from seed (it is equivalent to resetting the wallet cache from the wallet GUI). 
Here we need a breaking change (add a new `walletType` field) so I would just set `version` to 1 and add the new field to the codec

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
sstone: It should be possible to get the address from the key store ?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/db/sqlite/SqliteWalletDb.scala
```diff
@@ -197,13 +197,26 @@ object SqliteWalletDb {
...
   /**
     * change this value
-    * -if the new codec is incompatible with the old one
+    * - if the new codec is incompatible with the old one
     * - OR if you want to force a full sync from Electrum servers
     */
-  val version = 0x0000
+  val version = 0x0001
+
+  def walletTypeCodec(vers: Long): Codec[WalletType] = vers match {
```
sstone: why not simply set version to `0x0001` and add new walletType field (a byte for example, 0 for P2SH_SEGWIT, 1 for BECH32, rest not defined yet) ?

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWalletSpec.scala
```diff
@@ -90,13 +90,13 @@ class ElectrumWalletSpec extends TestKit(ActorSystem("test")) with FunSuiteLike
 
   test("wait until wallet is ready") {
     electrumClient = system.actorOf(Props(new ElectrumClientPool(new AtomicLong(), Set(ElectrumServerAddress(new InetSocketAddress("localhost", electrumPort), SSL.OFF)))))
-    wallet = system.actorOf(Props(new ElectrumWallet(seed, electrumClient, WalletParameters(Block.RegtestGenesisBlock.hash, new SqliteWalletDb(DriverManager.getConnection("jdbc:sqlite::memory:")), minimumFee = 5000 sat))), "wallet")
+    wallet = system.actorOf(Props(new ElectrumWallet(seed, electrumClient, WalletParameters(BECH32, Block.RegtestGenesisBlock.hash, new SqliteWalletDb(DriverManager.getConnection("jdbc:sqlite::memory:")), minimumFee = 5000 sat))), "wallet")
```
sstone: Why not reuse the way tests are defined for `ElectrumWalletSimulatedClientSpecBase` here and have BIP49 and BIP84 tests ?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/db/sqlite/SqliteWalletDb.scala
araspitzu: I thought a boolean was enough (after all is a binary choice here) but with the byte we are more _future-proof_, changes have been implemented in e5e6373

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWalletSpec.scala
araspitzu: Done in 7544fb0, i also added the wallet type to the test names of parametrized test because it was getting confusing to see all those test with the same name.

araspitzu:
bf6d7ba adds the clear cache functionality to the electrum wallet, this can be used by eclair-mobile to trigger a full resync with the electrum server by sending a `ClearWalletCache` to the wallet.

araspitzu:
Eclair-mobile integration with this PR needs to handle wallet restore in a special case. On restore the user will be asked whether he's restoring a bech32 or p2sh wallet, to avoid errors eclair-mobile can use special naming for the backup, this will avoid consistency issue because it forces the channel backup to match the electrum wallet type. However this still leaves room for user errors: if the user selects the wrong type then when eclair-mobile reconnects (i.e to ACINQ node) it will trigger the DLP and the remote will force close the channels. A potential solution to avoid that is to pair the wallet type with the `nodeId`, however this also have several disadvantages.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala
```diff
@@ -610,6 +606,11 @@ object ElectrumWallet {
 
     val chainHash = blockchain.chainHash
 
+    def keyStore = walletType match {
```
sstone: Why is it a `def` and not a `val` ?

araspitzu: Slipped through, it doesn't need to be a `def` so it can be a `val`

araspitzu:
#### High level interaction between eclair-mobile and eclair-core in the restore scenario

During the restore scenario the user is prompted to choose a `wallet-type` and depending on the choice eclair-mobile will load a different backup so the following scenarios can happen:

- User has no existing open channels and no existing backup: in this case whatever wallet-type is selected there are no issues.
- User has existing open channels from a previous wallet but selects the **wrong wallet type**: in this case when reconnecting to remote the existing channel will get force-closed and eclair-mobile won't notice it had funds at stake, to resolve this scenario the user can redo the restore procedure choosing the correct wallet-type then eclair should be able to track the existing channel closure and claim the funds.

dpad85:
The restore scenario for eclair mobile can be tricky and it would be nice to avoid any consistency issues.

When you say `User` in fact I would say `node id`, since there is no such thing as a `User`. So if the `node id` changes with wallet type, then the remote will not attempt to reestablish the channel. The p2sh and bech32 wallet would be completely isolated and as such we cannot have consistency issues.

I'm not sure I understand why changing the node id would cause any technical issues, since it's virtually the same thing as using a different seed? 

However I understand that we would have to write a special case in the code generating the node id and maintain it for a long time, just to prevent a onetime issue in bech32 migration for eclair mobile.

araspitzu:
The argument against changing the _nodeid_ is that it's often used as identifier in authentication prototocols and/or signature schemes, for example both [c-lightning](https://github.com/ElementsProject/lightning/blob/master/doc/lightning-signmessage.7.md) and [lnd](https://api.lightning.community/#signmessage) support signing a message with your node key (note that eclair don't support that at the moment); there might be other integrations of which we're aren't aware now. On the other hand if changing the _nodeid_ won't break any existing service/integration then i'm in favor of it.

araspitzu:
2ee9b13 adds the exception to use a different derivation path for the node key if the wallet is electrum in bech32 mode.

t-bast:
We've removed support for Electrum entirely, so this PR is now obsolete.

PR #1172, opened 2019-10-14
Update assisted channels

pm47:
When sending a payment, if a node on the route answers with an
`UPDATE`-type error, we update our routing table with the new
`channel_update` that is attached to the error message.

But additional routing info (provided by receiver in payment requests)
take precedence over whatever data is in the routing table. If one of
the nodes contained in the routing info replies with a newer
`channel_update`, we will indeed update our routing table, before
retrying, but then we will override it by the (untouched) assisted routes.

We know also update the assisted routes.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
@@ -139,8 +139,17 @@ class PaymentLifecycle(nodeParams: NodeParams, progressHandler: PaymentProgressH
             }
             // in any case, we forward the update to the router
             router ! failureMessage.update
+            // we also update assisted routes, because they take precedence over the router's routing table
+            val assistedRoutes1 = c.assistedRoutes.map(_.map {
+              case extraHop: ExtraHop if extraHop.shortChannelId == failureMessage.update.shortChannelId => extraHop.copy(
+                cltvExpiryDelta = failureMessage.update.cltvExpiryDelta,
+                feeBase = failureMessage.update.feeBaseMsat,
+                feeProportionalMillionths = failureMessage.update.feeProportionalMillionths
+              )
+              case extraHop => extraHop
+            })
             // let's try again, router will have updated its state
-            router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, c.assistedRoutes, ignoreNodes, ignoreChannels, c.routeParams)
+            router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, assistedRoutes1, ignoreNodes, ignoreChannels, c.routeParams)
           } else {
             // this node is fishy, it gave us a bad sig!! let's filter it out
             log.warning(s"got bad signature from node=$nodeId update=${failureMessage.update}")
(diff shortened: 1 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
```diff
@@ -358,6 +359,57 @@ class PaymentLifecycleSpec extends BaseRouterSpec {
...
+    // we change the cltv expiry
+    val channelUpdate_bc_modified = makeChannelUpdate(Block.RegtestGenesisBlock.hash, priv_b, c, channelId_bc, CltvExpiryDelta(42), htlcMinimumMsat = channelUpdate_bc.htlcMinimumMsat, feeBaseMsat = channelUpdate_bc.feeBaseMsat, feeProportionalMillionths = channelUpdate_bc.feeProportionalMillionths, htlcMaximumMsat = channelUpdate_bc.htlcMaximumMsat.get)
+    val failure = IncorrectCltvExpiry(CltvExpiry(5), channelUpdate_bc_modified)
+    // and node replies with a failure containing a new channel update
+    sender.send(paymentFSM, UpdateFailHtlc(ByteVector32.Zeroes, 0, Sphinx.FailurePacket.create(sharedSecrets1.head._1, failure)))
+
+    // payment lifecycle forwards the embedded channelUpdate to the router
+    routerForwarder.expectMsg(channelUpdate_bc_modified)
+    awaitCond(paymentFSM.stateName == WAITING_FOR_ROUTE && paymentDb.getOutgoingPayment(id).exists(_.status === OutgoingPaymentStatus.Pending)) // 1 failure but not final, the payment is still PENDING
+    routerForwarder.expectMsgType[RouteRequest]
```
t-bast: Can't you instead assert that the route request here contains the updated channel update (instead of the lines below)?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
```diff
@@ -139,8 +139,17 @@ class PaymentLifecycle(nodeParams: NodeParams, progressHandler: PaymentProgressH
             }
             // in any case, we forward the update to the router
             router ! failureMessage.update
+            // we also update assisted routes, because they take precedence over the router's routing table
+            val assistedRoutes1 = c.assistedRoutes.map(_.map {
+              case extraHop: ExtraHop if extraHop.shortChannelId == failureMessage.update.shortChannelId => extraHop.copy(
+                cltvExpiryDelta = failureMessage.update.cltvExpiryDelta,
```
t-bast: I think you should copy almost everything (htlcMin and htlcMax are theoretically useful to copy because they play a small role in path-finding scoring iirc)

pm47: I can't? This is an `ExtraHop`, which only has a subset of what `ChannelUpdate`'s fields.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
pm47: Same reason as below

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentLifecycle.scala
t-bast: Indeed, got it.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
t-bast: It's not exactly the same, it's more correct to test here that the RouteRequest contains the expiry of 42 in the updated ExtraHop than what the test currently does

PR #1175, opened 2019-10-15
Check configuration for obsolete keys in startup

sstone:
We now check the loaded configuration for obsolete keys (that have been moved to a new section)
and throw an error if any are found, which will prevent eclair from starting.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -93,6 +93,14 @@ object NodeParams {
 
   object ELECTRUM extends WatcherType
 
+  def checkConfiguration(config: Config): Config = {
+    // this check is needed because somme keys were moved to a new section in v0.3.2
+    if (config.hasPath("eclair.default-feerates") || config.hasPath("eclair.max-feerate-mismatch") || config.hasPath("eclair.update-fee_min-diff-ratio")) {
+      throw new IllegalArgumentException("Your configuration uses keys that have moved to a new section, please check our release notes for more details")
+    }
+    config
+  }
+
   /**
    * Order of precedence for the configuration parameters:
    * 1) Java environment variables (-D...)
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -93,18 +93,28 @@ object NodeParams {
 
   object ELECTRUM extends WatcherType
 
+  def checkConfiguration(config: Config): Config = {
+    // this check is needed because somme keys were moved to a new section in v0.3.2
+    if (config.hasPath("eclair.default-feerates") || config.hasPath("eclair.max-feerate-mismatch") || config.hasPath("update-fee_min-diff-ratio")) {
```
araspitzu: Shouldn't `config.hasPath("update-fee_min-diff-ratio")` be `config.hasPath("eclair.update-fee_min-diff-ratio")`? I see it inside the `eclair` block in [0.3.1](https://github.com/ACINQ/eclair/blob/v0.3.1/eclair-core/src/main/resources/reference.conf#L81)

sstone: Good catch thanks!

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -93,18 +93,28 @@ object NodeParams {
 
   object ELECTRUM extends WatcherType
 
+  def checkConfiguration(config: Config): Config = {
+    // this check is needed because somme keys were moved to a new section in v0.3.2
```
pm47:
```suggestion
    // this check is needed because some keys were moved to a new section in v0.3.2
```

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -128,6 +128,8 @@ object NodeParams {
   }
 
   def makeNodeParams(config: Config, keyManager: KeyManager, torAddress_opt: Option[NodeAddress], database: Databases, blockCount: AtomicLong, feeEstimator: FeeEstimator): NodeParams = {
+    val keyPaths = Seq("default-feerates", "max-feerate-mismatch", "update-fee_min-diff-ratio")
+    keyPaths.foreach(keyPath => require(!config.hasPath(keyPath), s"configuration key $keyPath has been renamed/moved to a new section"))
```
pm47: How about:

```scala
    val deprecatedKeyPaths = Map(
      "default-feerates" -> "on-chain-fees.default-feerates",
      "max-feerate-mismatch" -> "on-chain-fees.max-feerate-mismatch",
      "update-fee_min-diff-ratio" -> "on-chain-fees.update-fee-min-diff-ratio"
    )
    deprecatedKeyPaths.foreach {
      case (old, new_) => require(!config.hasPath(old), s"configuration key '$old' has been replaced by '$new_'")
    }
```

PR #1176, opened 2019-10-15
Refactor Upstream/Origin

t-bast:
This is a small refactoring of how we handle upstream/origin HTLC.
This is in preparation of Trampoline where we'll add a `NodeRelayed` origin that will contain a list of incoming HTLCs (of potentially only one HTLC if MPP isn't used).
This isn't a very impacting change but doing it on master saves me from painful rebases in my long-lived trampoline branch and will help the review of the trampoline branch (better to separate refactoring from new feature code).

At some point I wanted to remove the distinction between `Origin` and `Upstream`, but the retry logic in HTLC relay requires `Upstream` to contain the whole `UpdateAddHtlc` whereas we want to store something much lighter in the DB (`Origin`).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
@@ -106,8 +106,10 @@ case class BITCOIN_PARENT_TX_CONFIRMED(childTx: Transaction) extends BitcoinEven
 
 sealed trait Upstream
 object Upstream {
+  /** Our node is the payer. */
   final case class Local(id: UUID) extends Upstream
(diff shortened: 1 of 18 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -105,8 +104,16 @@ case class BITCOIN_PARENT_TX_CONFIRMED(childTx: Transaction) extends BitcoinEven
        "Y8888P"   "Y88888P"  888       888 888       888 d88P     888 888    Y888 8888888P"   "Y8888P"
  */
 
+sealed trait Upstream
+object Upstream {
+  /** Our node is the payer. */
```
pm47:
```suggestion
  /** Our node is the origin of the payment. */
```

PR #1178, opened 2019-10-16
Support pruned bitcoind

pm47:
Impacted parts are:
  - `WatchConfirmed`: can't use `getrawtransaction` anymore; instead we
  use `importaddress`+`rescanblockchain`+`gettransaction`.
  - `WatchSpent`: untouched (there is no indexation for that in bitcoind
  so we were looking in the mempool and in the blockchain, and this
  still works)
  - `ValidateRequest`: doesn't work anymore

Note that we still use `getrawtransaction` when looking for spending
txes in the mempool. The `gettransaction`/`getrawtransaction`
differenciation is quite messy currently.

Left to do:
- handle restart with a brand new bitcoin core (which doesn't watch the
relevant scripts)
- handle (disable? delegate to electrum servers?) authentication of public
channels

Regarding pruning, we can probably just prune everything below
block height 500 000.


araspitzu:
Update:
- Channel shortId validation fully works
- Pruning is limited to block 500000 (if higher than that eclair won't start)
- We always re-add `WatchConfirmed` on channel restore (can be optimized?)

Left to be done: lots of test.

Is worth to be noted that bitcoin-core won't be able to import and watch any arbitrary number of scripts, although the number is supposedly very high we do have a hard ceiling on the number of channels we can handle.

araspitzu:
d53f5e8 makes sure we don't rely on bitcoind indexing when checking a `WatchSpent`. This caused troubles when the user switches to a new bitcoind instance during the breach scenario, in that case the funding transaction is spent but bitcoind is not watching it's output scripts; eclair missed the spent because the zmq watcher was still using [`gettransaction`](https://github.com/ACINQ/eclair/pull/1178/files#diff-67bdb81814859654032b95a44799d74eL120). There are now integration tests that use both spent and unspent non wallet transactions.

araspitzu:
Note for reviewers: the PR relies on bitcoind to keep an extra state for eclair, this is necessary to use `gettransaction` RPC and is achieved via `importaddress` + `rescanblockchain`. This requirement adds extra complexity because we need to deal with the assumption that at any given time the user can restart eclair with a fresh bitcoind instance without any script imported, currently there is no integration test for these scenarios.

araspitzu:
At commit f9fd989 there is no reimport procedure anymore, the watcher is autonomous and can handle chain rescans. The rescans are triggered every time a new block arrives (if and only if there are pending `WatchConfirmed`) and they are limited to the last block. Also importing addresses has been optimized by first looking up if the script/address of a `WatchConfirmed` was already imported, if not we don't reimport it again.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -34,15 +34,46 @@ class ExtendedBitcoinClient(val rpcClient: BitcoinJsonRPCClient) {
 
   implicit val formats = org.json4s.DefaultFormats
 
+  def listReceivedByAddress()(implicit ec: ExecutionContext): Future[List[String]] = {
```
sstone: `minconf` `include_empty` and `include_watchonly` should be explicitly passed to `listReceivedByAddress()`, here we use default values that do not match bitcoin core's

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -80,12 +79,20 @@ class ZmqWatcher(blockCount: AtomicLong, client: ExtendedBitcoinClient)(implicit
...
+          count
       }
+
+      // if there are pending WatchConfirmed(s) we trigger a rescan to be able to check if any of them was confirmed
+      // this operation is blocking but we only rescan from the last -1 block which should be quick
+      for {
+        blockHeigth <- blockHeigthF
+        _ <- if(watches.exists(w => w.isInstanceOf[WatchConfirmed])) client.rescanBlockChain(blockHeigth - 1) else Future.successful(Unit)
+      } yield ()
+
```
sstone: I don't understand how limiting rescans to the last block works. Is there a test for this ? What happends when the tx was confirmed while eclair was offline. At least there needs to be a more explicit comment in the code and the commit description, and a specific test

araspitzu: +1 for the test, i will add it. If eclair was offline it will re-add the watchers when restarted, i see the point that if it was offline for too long and the transaction went buried in several blocks this won't work. But how should we handle this? If the watcher is added and eclair goes offline how deep should it rescan when restarted? Or how deep should it rescan when the watcher is re-added? I struggle to see how we can get this right without scanning from the beginning every time.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
araspitzu: Will do, i will also add the `address_filter` as parameter to be able to query for a single address instead of getting all the list.

araspitzu: This has been addressed in d25fe7b

araspitzu:
### Update

The ZMQ watcher is now capable of providing the same functionality as it does in `master` branch, with the difference that when a new block arrives it *might* trigger a chain rescan. The `WatchConfirmed` contains a scriptPubKey which is imported as watch-only address in bitcoind if it wasn't imported already (d25fe7b). Every `WatchConfirmed` now contains `rescanHeight`: the earliest block height at which we need to scan to index the watched transaction, those rescans are queued until a new block arrives, if there are pending `WatchConfirmed` the rescan is performed using the earliest of the heights of all watchers. This ensures that we can lookup a transaction if we need to and we never do unnecessary work before that (i.e scanning twice). The `rescanHeight` of `WatchConfirmed` is chosen by the Channel with the following logic:
- If we want to watch a brand new transaction then use the current block height
- If there is a `short_channel_id` then use its block height
- If there is a `waitingSince` (in DATA_WAIT_FOR_FUNDING_CONFIRMED) compute the block height corresponding to that timestamp

`waitingSince` is used when eclair has been stopped before the funding is confirmed, because we don't know if bitcoind has already indexed the transaction re-adding the watcher triggers a rescan but we limit the height based on when the transaction was created. 

### Test
`ZmqWatcherSpec` has been enhanced to assert we don't import addresses if not necessary, we queue rescans and we rescan from the earliest height of all watchers. There is also a test where the bitcoin wallet is resetted, we then add new `WatchConfirmed` and assert they work properly. There are also [test](https://github.com/ACINQ/eclair/pull/1178/files#diff-7a611410d2aebfdc5090f528b3689ab4R64) asserting we correctly compute the block height from `waitingSince`.

sstone:
I think that there may be a way to use labels and properly check which watch-only addresses have been properly imported:
- when we start, get all addresses labelled "imported" and compare to our own list
- for the ones that are missing, use `importmulti` with a label set to "pending"
- once `importmulti` returns successfully, change the label to "imported"

araspitzu:
> I think that there may be a way to use labels and properly check which watch-only addresses have been properly imported:
> 
> * when we start, get all addresses labelled "imported" and compare to our own list
> * for the ones that are missing, use `importmulti` with a label set to "pending"
> * once `importmulti` returns successfully, change the label to "imported"

Sounds nice but what's the difference with using `listreceivedbyaddress` as implemented in the [bitcoin client](https://github.com/ACINQ/eclair/pull/1178/files#diff-fc6a493e6d83603a7e2848bde1fb1cf3R60-R71)?  The assumptions for `importmulti` are such that if it terminates it must have correctly imported the given addresses, we can later query which addresses were imported and optimize to not re-import twice (note that we use `include_empty=true` so `listreceivedbyaddress` will return addresses that were imported but haven't received anything).

sstone:
because with`importmulti`, importing and scanning is not atomic: all addresses/scripts will be imported even if scanning fails or bitcoind is stopped. Then there is no way to know if scanning was successful or not, and I'm not such that bicoind will resume scanning when it restarts.

araspitzu:
Ah i see your point, by using labels we can distinguish between those addresses that were _actually_ imported and those that were not, perhaps because scanning was aborted and the address stay in "pending". This will be useful for sure because it's much more reliable than just assuming `importmulti` worked, i'm going to include it in my changes.

araspitzu:
Superceded by https://github.com/ACINQ/eclair/pull/1230

PR #1180, opened 2019-10-21
Fix ignored duplicate amounts in htlc-value-in-flight

t-bast:
Using `.map()` on a `Set` can have surprising results: in our case we end up ignoring htlcs that have the same amount, so the htlc-value-in-flight may be less than what it should be.

Credits to @btcontract for finding the bug (this PR supercedes #1163 because of build issues)

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
@@ -20,8 +20,8 @@ import fr.acinq.eclair.MilliSatoshi
 import fr.acinq.eclair.wire._
 
 /**
-  * Created by PM on 07/12/2016.
-  */
+ * Created by PM on 07/12/2016.
+ */
 
 // @formatter:off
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -36,10 +36,10 @@ final case class CommitmentSpec(htlcs: Set[DirectedHtlc], feeratePerKw: Long, to
 }
 
 object CommitmentSpec {
-  def removeHtlc(changes: List[UpdateMessage], id: Long): List[UpdateMessage] = changes.filterNot(_ match {
+  def removeHtlc(changes: List[UpdateMessage], id: Long): List[UpdateMessage] = changes.filterNot {
     case u: UpdateAddHtlc if u.id == id => true
```
pm47:
```suggestion
    case u: UpdateAddHtlc => u.id == id
```

jonatack:
@t-bast this looks like the bug you mentioned in Berlin?

t-bast:
Exactly ;)

PR #1181, opened 2019-10-22
Fix codecov unknown tag

t-bast:
The `ci` tag was unknown to codecov and I suspect that's why they weren't honoring the `ignore` on eclair-node and GUI.
I also changed our target range to 70-90% (impacts the colors of the badge, it's nit).

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,5 @@
 coverage:
-  ci:
-    - pm47.semaphoreci.com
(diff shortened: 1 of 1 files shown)


t-bast:
I used the following command to check our `.codecov.yaml` and it was reporting it as invalid:
```sh
curl --data-binary @.codecov.yml https://codecov.io/validate
```

PR #1183, opened 2019-10-23
Rework payment handling

pm47:
Instead of replacing payment handler actor, we compose partial methods.


eclair-core/src/main/scala/fr/acinq/eclair/payment/handlers/DefaultHandler.scala
```diff
@@ -14,35 +14,36 @@
...
-  implicit val ec: ExecutionContext = context.system.dispatcher
-  val paymentDb = nodeParams.db.payments
+  val paymentsDb = nodeParams.db.payments
+  var pendingPayments: Map[ByteVector32, (ByteVector32, ActorRef)] = Map.empty
 
-  override def receive: Receive = main(Map.empty)
+  override def handle(implicit ctx: ActorContext, log: LoggingAdapter): PartialFunction[Any, Unit] = {
 
-  def main(pendingPayments: Map[ByteVector32, (ByteVector32, ActorRef)]): Receive = {
     case ReceivePayment(amount_opt, desc, expirySeconds_opt, extraHops, fallbackAddress_opt, paymentPreimage_opt, allowMultiPart) =>
```
t-bast: `ReceivePayment` is currently defined in `PaymentLifecycle` which doesn't really make sense...
Maybe a good opportunity to move it here in the `DefaultHandler` companion object?

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -72,6 +52,30 @@ trait PaymentsDb {
 
   /** List all received (paid) incoming payments in the given time range (milli-seconds). */
   def listReceivedIncomingPayments(from: Long, to: Long): Seq[IncomingPayment]
+}
+
+trait OugoingPaymentsDb {
```
t-bast:
```suggestion
trait OutgoingPaymentsDb {
```

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -26,29 +26,9 @@ import fr.acinq.eclair.{MilliSatoshi, ShortChannelId}
...
-
-  /** List all the outgoing payment attempts that are children of the given id. */
-  def listOutgoingPayments(parentId: UUID): Seq[OutgoingPayment]
-
-  /** List all the outgoing payment attempts that tried to pay the given payment hash. */
-  def listOutgoingPayments(paymentHash: ByteVector32): Seq[OutgoingPayment]
-
-  /** List all the outgoing payment attempts in the given time range (milli-seconds). */
-  def listOutgoingPayments(from: Long, to: Long): Seq[OutgoingPayment]
+trait PaymentsDb extends IncomingPaymentsDb with OugoingPaymentsDb
```
t-bast:
```suggestion
trait PaymentsDb extends IncomingPaymentsDb with OutgoingPaymentsDb
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/handlers/MultiPartHandler.scala
```diff
@@ -14,35 +14,48 @@
...
+                       db: IncomingPaymentsDb) extends ReceiveHandler {
+
+  import MultiPartHandler._
+
+  // NB: this is safe because this handler will be called from within an actor
+  private var pendingPayments: Map[ByteVector32, (ByteVector32, ActorRef)] = Map.empty
 
-  import LocalPaymentHandler._
+  /**
+   * Can be overridden for a more fine-grained controle of whether or not to handle this payload.
```
t-bast:
```suggestion
   * Can be overridden for a more fine-grained control of whether or not to handle this payload.
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/handlers/MultiPartHandler.scala
```diff
@@ -14,35 +14,48 @@
...
+   * If the call returns false, then the pattern matching will fail and the payload will be passed to other handlers.
+   */
+  def doHandle(p: FinalPayload): Boolean = true
 
-  implicit val ec: ExecutionContext = context.system.dispatcher
-  val paymentDb = nodeParams.db.payments
+  /**
+   * Can be overridden to do custom processing on successfully received payments.
+   */
+  def onSuccess(paymentReceived: PaymentReceived)(implicit log: LoggingAdapter): Unit = ()
```
t-bast: This is currently unused. Not sure yet this is the most elegant way of customizing that, I need to think about it more...

eclair-core/src/main/scala/fr/acinq/eclair/payment/handlers/MultiPartHandler.scala
```diff
@@ -14,35 +14,48 @@
...
+
+  // NB: this is safe because this handler will be called from within an actor
+  private var pendingPayments: Map[ByteVector32, (ByteVector32, ActorRef)] = Map.empty
 
-  import LocalPaymentHandler._
+  /**
+   * Can be overridden for a more fine-grained controle of whether or not to handle this payload.
+   * If the call returns false, then the pattern matching will fail and the payload will be passed to other handlers.
+   */
+  def doHandle(p: FinalPayload): Boolean = true
```
t-bast: If I understand correctly, you would have `MultiPartHandler` as the last handler, right?
I'm not sure this is the right level of customization. If another handler wants to handle some `FinalPayload`s, I would not make him extends `MultiPartHandler`. Instead I would just create another handler that pattern matches on `p: FinalPayload if myHandlersCondition(p)`

eclair-core/src/main/scala/fr/acinq/eclair/payment/handlers/ReceiveHandler.scala
```diff
@@ -0,0 +1,27 @@
...
+
+package fr.acinq.eclair.payment.handlers
+
+import akka.actor.ActorContext
+import akka.event.LoggingAdapter
+
+
+trait ReceiveHandler {
+
+  def handle(implicit ctx: ActorContext, log: LoggingAdapter): PartialFunction[Any, Unit]
```
sstone: Why not use `Actor.Receive` instead of `PartialFunction[Any, Unit]` ? It would be more explicit

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentHandler.scala
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Generic payment handler that delegates handling of incoming messages to a list of handlers.
+ *
+ * @param nodeParams
+ */
+class PaymentHandler(nodeParams: NodeParams) extends Actor with ActorLogging {
+
+  override def receive: Receive = {
+    val defaultHandler = new MultiPartHandler(nodeParams, nodeParams.db.payments)
+    normal(Seq(defaultHandler), defaultHandler.handle(context, log))
```
sstone: It looks like we're going to create a new  MultiPartHandler every time receive is called (I have not tested this)

eclair-core/src/main/scala/fr/acinq/eclair/payment/handlers/ReceiveHandler.scala
pm47: Addressed with 0f4d028.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentHandler.scala
pm47: Addressed with 0f4d028.

Issue #1179, opened 2019-10-18
Broken /allupdates API


araspitzu:
The `/allupdates` API is broken since #902, thanks @rorp for the report!

araspitzu:
Closed by #1187

araspitzu:
Closed by #1187

PR #1187, opened 2019-10-23
Fix /allupdates API 

araspitzu:
The /allupdates API was broken when used with a public key filter (see #1179), this PR restores the previous behavior and adds some more extensive test between the `Router` and `EclairImpl`. The API now returns all the updates created by the given predicate key.


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -470,6 +470,10 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
       sender ! updates
       stay
 
+    case Event('updatesMap, d) =>
```
t-bast: This is the same as 'channelsMap, probably don't need to add it, just call 'channelsMap instead?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -181,7 +181,13 @@ class EclairImpl(appKit: Kit) extends Eclair {
 
   override def allUpdates(nodeId_opt: Option[PublicKey])(implicit timeout: Timeout): Future[Iterable[ChannelUpdate]] = nodeId_opt match {
     case None => (appKit.router ? 'updates).mapTo[Iterable[ChannelUpdate]]
-    case Some(pk) => (appKit.router ? 'updatesMap).mapTo[Map[ChannelDesc, ChannelUpdate]].map(_.filter(e => e._1.a == pk || e._1.b == pk).values)
+    case Some(pk) => (appKit.router ? 'updatesMap).mapTo[Map[ShortChannelId, PublicChannel]].map { channels =>
```
t-bast: nit: use 'channelsMap

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
araspitzu: Yes they're the same and 'updatesMap can go, i removed it in fe70ae2.

PR #1189, opened 2019-10-23
add networkstats API call to allow display of network stats

n1bor:
Small change to add network stats to API.

Useful as allnodes and allchannels no longer of much use as return too much data.


eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -173,6 +174,28 @@ class EclairImplSpec extends TestKit(ActorSystem("test")) with fixture.FunSuiteL
...
+    val feePropStat=Stats(32l, 11l, 13l, 22l, 42l, 51l, 53l)
+    val eclair = new EclairImpl(kit)
+    val fResp = eclair.networkStats()
+    f.router.expectMsg(GetNetworkStats)
+
+    f.router.reply( Some(new NetworkStats(1,2,capStat,cltvStat,feeBaseStat,feePropStat) ) )
+
+    awaitCond({
+      fResp.value match {
+        case Some(Success(Some(res))) => res.channels==1
```
araspitzu: nit: formatting
```suggestion
        case Some(Success(Some(res))) => res.channels == 1
```

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -173,6 +174,28 @@ class EclairImplSpec extends TestKit(ActorSystem("test")) with fixture.FunSuiteL
...
+
+    val capStat=Stats(30 sat, 12 sat, 14 sat, 20 sat, 40 sat, 46 sat, 48 sat)
+    val cltvStat=Stats(CltvExpiryDelta(32), CltvExpiryDelta(11), CltvExpiryDelta(13), CltvExpiryDelta(22), CltvExpiryDelta(42), CltvExpiryDelta(51), CltvExpiryDelta(53))
+    val feeBaseStat=Stats(32 msat, 11 msat, 13 msat, 22 msat, 42 msat, 51 msat, 53 msat)
+    val feePropStat=Stats(32l, 11l, 13l, 22l, 42l, 51l, 53l)
+    val eclair = new EclairImpl(kit)
+    val fResp = eclair.networkStats()
+    f.router.expectMsg(GetNetworkStats)
+
+    f.router.reply( Some(new NetworkStats(1,2,capStat,cltvStat,feeBaseStat,feePropStat) ) )
```
araspitzu:
```suggestion
    f.router.reply(Some(new NetworkStats(1,2,capStat,cltvStat,feeBaseStat,feePropStat)))
```

n1bor:
Anything I need to do? Or just in the queue?

araspitzu:
> Anything I need to do? Or just in the queue?

Yes, could you please add another change to the PR? The change can also be just reordering the commit history (i.e squash + force push) but it's necessary for you to trigger the CI build. Semaphore CI doesn't work with all github accounts and we are whitelisting contributors to do so, once the CI checks are successful github should let me merge this PR. You might also need to sign your commits with PGP but i'm not sure about that since i can probably sign the merge commit on your behalf.

n1bor:
Did as requested - and looks like it worked.
But bit annoying that I can not see the semaphoreci web site. Get a 404 error...

t-bast:
I can't merge this because some of your commits were not signed.
Can you squash/rebase and push a **signed** commit?

n1bor:
Think I have done it! There is a green tick by the commit and the semaphore build.
Although I still have not access to see the semaphore stuff...

t-bast:
> Although I still have not access to see the semaphore stuff...

We had a chat with the semaphore support team, and it's expected: for now they only support "private" projects.
However they told me that support for public project (allowing anyone to see our builds) should ship in a couple of weeks.
So be patient, you will soon be able to see the build details ;)

t-bast:
It looks like there is a problem with your signature...
Github tells me: 

**Merge attempt failed**
Commits must have valid signatures.

Which is very weird because the UI does show the green `verified` tick...
Sorry for this, those requirements are there for a good reason but they can be annoying when first setting up. Let me investigate a bit.

t-bast:
According to Github documentation this should work. The only explanation I'd see is that the PR previously contained unsigned commits and Github doesn't handle that well.

I know it's annoying, but it will help us make our contribution process better in the future. @n1bor could you try closing this PR, making a new clean branch with a single signed commit with your changes and open a new PR? And if that doesn't work we'll get in touch with Github support.

t-bast:
Closed in favor of #1221

PR #1190, opened 2019-10-24
Add 'getnewaddress' API

### LINKED ISSUE #1006: [API] Get deposit address
The new API does not have a call to get a new deposit address.
This mean that if our client app need this, we need to pass bitcoind parameters to it while eclair already know all of it. Can you add this?

We need this for [btcpay integration](https://github.com/btcpayserver/BTCPayServer.Lightning/) that @Kukks  is working on.

araspitzu:
Add a simple API that calls `getfinaladdress` on EclairWallet, this can be used to obtain a new on-chain deposit address from your bitcoin-core wallet. Fixes #1006

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -80,6 +80,8 @@ trait Eclair {
 
   def receive(description: String, amount_opt: Option[MilliSatoshi], expire_opt: Option[Long], fallbackAddress_opt: Option[String], paymentPreimage_opt: Option[ByteVector32])(implicit timeout: Timeout): Future[PaymentRequest]
 
+  def depositAddress(): Future[String]
+
   def receivedInfo(paymentHash: ByteVector32)(implicit timeout: Timeout): Future[Option[IncomingPayment]]
 
   def send(externalId_opt: Option[String], recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest] = None, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
@@ -188,6 +190,8 @@ class EclairImpl(appKit: Kit) extends Eclair {
     fallbackAddress_opt.map { fa => fr.acinq.eclair.addressToPublicKeyScript(fa, appKit.nodeParams.chainHash) } // if it's not a bitcoin address throws an exception
     (appKit.paymentHandler ? ReceivePayment(description = description, amount_opt = amount_opt, expirySeconds_opt = expire_opt, fallbackAddress = fallbackAddress_opt, paymentPreimage = paymentPreimage_opt)).mapTo[PaymentRequest]
   }
+
+  override def depositAddress(): Future[String] = appKit.wallet.getFinalAddress
 
   override def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty)(implicit timeout: Timeout): Future[RouteResponse] = {
     (appKit.router ? RouteRequest(appKit.nodeParams.nodeId, targetNodeId, amount, assistedRoutes)).mapTo[RouteResponse]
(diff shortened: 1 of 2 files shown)


n1bor:
Maybe review/merge with #1189 as also API change.

araspitzu:
> Depending on the wallet implementation, a new address will be generated everytime this method is called. Probably not what we want?

Why not? Both backend provide a deposit address that be used to send funds to eclair (i.e eclair can use those funds to open channels), i don't see how generating a new address every time can be a problem, unless the application calling the API relies on the address to be the same across calls - which is probably wrong.

pm47:
It's not necessarily a problem, maybe just needs to be documented. I don't know how e.g. BTCPayServer would use it. What we don't want is having people generate thousands of addresses without being aware.

araspitzu:
Agreed that documenting this is very important, also what do you think of restricting it to bitcoin-backend only? Then we can make it clear in the doc that this is just an alias for bitcoin RPC `getnewaddress`, i'm not sure either how BTCPayServer would use it, maybe @NicolasDorier can clarify if they're going to rely on a specific behavior or they already assume it behaves like `getnewaddress`

NicolasDorier:
The semantic should be "Give me whatever address necessary so I can create a channel after"

eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -279,6 +279,9 @@ trait Service extends ExtraDirectives with Logging {
                         } ~
                         path("usablebalances") {
                           complete(eclairApi.usableBalances())
+                        } ~
+                        path("getdepositaddress"){
```
pm47: This name indicates that a new address will be generated every time, make it obvious that this is a simple pass-through to bitcoind, and doesn't add ambiguity by introducing  a new concept of `deposit` that's nowhere else in the code.

```suggestion
                        path("getnewaddress"){
```

araspitzu:
@pm47 what do you think of restricting the API usage to bitcoind-backend only? AFAIK the electrum backend is not supported on eclair-server but as a good API design we should make it hard for  developers to use the API in the wrong way (with electrum backend).

pm47:
> @pm47 what do you think of restricting the API usage to bitcoind-backend only? AFAIK the electrum backend is not supported on eclair-server but as a good API design we should make it hard for developers to use the API in the wrong way (with electrum backend).

I'm not too sure. On one hand it would create a special cases in our code for no real added value: if people run eclair with an electrum backend on server they have bigger issues (e.g. no access to the wallet). On the other hand, since we named it `getnewaddress` it clashes with the electrum implementation that uses `GetCurrentReceiveAddress`. Ok let's restrict it to bitcoind backend, as cleanly as possible.

Re: 7c4b30f you also need to update the `eclairApi.depositAddress`. We should avoid introducing new concepts like deposit (where is the withdrawal?) or things can get messy pretty quickly.

araspitzu:
0f1baba implements the requested changes. One last bikeshed comment: now that we called it `getnewaddress` feels a bit out of context and it's not immediately clear what address it is, i think we should either make it clear in the docs or have a self descriptive name.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -189,6 +192,14 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.paymentHandler ? ReceivePayment(description = description, amount_opt = amount_opt, expirySeconds_opt = expire_opt, fallbackAddress = fallbackAddress_opt, paymentPreimage = paymentPreimage_opt)).mapTo[PaymentRequest]
   }
 
+  override def newAddress(): Future[String] = {
+    if(appKit.nodeParams.watcherType == ELECTRUM){
```
pm47: Forbidding for electrum isn't the same as restricting to bitcoin core:

```scala
override def newAddress(): Future[String] = {
  appKit.wallet match {
    case w: BitcoinCoreWallet => w.getFinalAddress
    case _ => Future.failed(new IllegalArgumentException("this call is only available with a bitcoin core backend"))
  }
}
```

araspitzu:
Doc updated via cc98a560

NicolasDorier:
fuck yeah!!! Can we get a new docker soon? :D

araspitzu:
It's already in the _latest_ of https://hub.docker.com/r/acinq/eclair/builds ;)

NicolasDorier:
@araspitzu it works, when will you publish a new tag?

PR #1195, opened 2019-10-30
[WIP] Random scid

btcontract:
It's an attempt to implement https://github.com/lightningnetwork/lightning-rfc/pull/681, it is sufficient for my needs as is but not complete and overall has different goals mixed in, maybe parts of it will be of help.

What is does:
1. Implements random `scid` basics: feature bits, messages, assigning logic
- max random block height is set to `500000`
- the same `scid` is used for both sides of channel which is probably not compliant (https://github.com/lightningnetwork/lightning-rfc/pull/681#discussion_r340439270) but makes implementation easy (`Router` first receives `LocalChannelUpdate` and then remote `PeerRoutingMessage` with a same random `scid`)
- peers exchange `ChannelUpdate`s after assigning random `scid` to make sure `Router` has a refreshed `PrivateChannel`with both updates present.
2. Changes `UsableBalances` logic in `Relayer`, see https://gitter.im/ACINQ/developers?at=5db6cec8e886fb5aa20d6776 and https://gitter.im/ACINQ/developers?at=5db9756c14d55a3785a12f2b
3. Adds a new `createinvoicewithextrahops` API call which was a goal for all of this. This call creates an invoice with routing hints for local private channels which can receive a requested amount right now (have remote `ChannelUpdate`, have sufficient `availableBalanceForReceive`).


t-bast:
I think we can close this PR as rusty decided to retract the corresponding spec PR.
We will instead use blinded paths to achieve that result.

PR #1196, opened 2019-10-30
More metrics

pm47:
This adds metrics on bitcoin rpc calls and on scodec.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
@@ -16,9 +16,12 @@
 
 package fr.acinq.eclair.wire
 
-import fr.acinq.eclair.wire
+import fr.acinq.eclair.{KamonExt, wire}
 import fr.acinq.eclair.wire.CommonCodecs._
-import scodec.Codec
+import kamon.Kamon
+import kamon.tag.TagSet
+import scodec.bits.BitVector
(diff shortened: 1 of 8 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -286,4 +289,20 @@ object LightningMessageCodecs {
...
+      val begin = System.nanoTime()
+      val res = lightningMessageCodec.decode(bits)
+      val end = System.nanoTime()
+      res match {
+        case Attempt.Successful(decoded) => timer.withTag("type", decoded.value.getClass.getSimpleName).record(end - begin)
+        case Attempt.Failure(_) => timer.withTag("type", "unknown").record(end - begin)
+      }
+      res
+    }
+  )
```
sstone: LGTM but what is the performance cost of adding timers here ?

pm47: I don't know, the assumption is that the overhead of metrics is not significant compared to the code being measured.

According to the [doc](https://kamon.io/docs/latest/core/metrics/) I should probably statically define the metrics:

> When possible, define and refine your metrics up front, as a static member or instance members on the instrumented components of your services. This saves a bit of time in looking up instruments. If you can’t do so, rest assured that Kamon will always return the same instrument instance if requested with the same arguments (metric name and tags) from anywhere in your code.

PR #1197, opened 2019-10-31
Recovery tool

araspitzu:
Using the deterministic key paths (see #1097) the recovery tool provides the means to recover our main outputs when the user lost backups and all channel data, it only needs to remember the `nodeId` of the remote. 

The recovery can be triggered via API or manually by starting eclair with `-Declair.recovery-tool`, the API is also exposed internally so this can be reused in eclair-mobile.

Left to be done:
- [ ] Corner case: channel already closed and we're funder
- [ ] expose API
- [ ] handle channels with option_static_remotekey ?


araspitzu:
### Implementation details

The recovery tool extends the current `Peer` class in order to wait for the remote to send us a `channel_reestablish`,  this is useful in the case the user doesn't remember the channelId. We then use the channelId to lookup the funding transaction, since the channelId is constructed [using the funding txid and output](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#definition-of-channel_id) the recovery will try output indexes from 0 to 5.

After having verified that the funding transaction exists the recovery FSM will ask the remote peer to publish its latest commitment. Using backupless backups :tm: we reconstruct the channel basepoints and use the `remotePerCommitmentSecret` (received in the channel_reestablish) to compute the key necessary to spend our main P2WPKH output. 

The final stage of the recovery is to sign and publish the claim transaction, this transaction will spend the channel main output to an address of the configured on-chain wallet.

araspitzu:
#### Update
The recovery tool has been tested between eclair instances only, further tests are required. As the last commit the code touches some part of eclair-core and eclair-node, this is because it has been _wired_ to the internal eclair API for eclair-mobile integration and to the HTTP API for ease of usage. However it's very easy to extract everything into a standalone application, this would have a dependency on eclair-core but no further changes would be required to eclair-core.

n1bor:
Is there a strategy on this long term? The current "dump the database every time it changes" is not ideal!

One option is to move to a synchronous replicating DB as the back end. 
Although any sort of replicating db is going to hate the binary objects that the Channel Update commits consist of.

Other option to put the eclair.sql on a DRBD replicated block device? Although then do not have a "log device" so can not point in time recover if there is a corruption.

Another is to output synchronously some sort of log (that could be on a remove NFS server) each time a channel is updated. This then could be read by the recovery tool to make a full recovery - even if db was corrupted.

araspitzu:
@n1bor this PR isn't really about backup strategies but more on how to recover funds when you've lost you backup. For more about backup there are a few different improvements being discussed in https://gitter.im/ACINQ/developers.

Talkless:
@araspitzu How could I actually build use your branch for recovering lost wallet (seed and nodeId known)?

Tried building various ways:
```
mvn package -DskipTests
```
```
mvn package -pl eclair-node -am -DskipTests
```
```
mvn compile -pl eclair-core,eclair-node
```
But I don't see `eclair-node` binary (if it should be), or `eclair-node.sh`, only `.jar` files that fails with `no main manifest attribute, in ./eclair-node/target/eclair-node_2.11-0.3.3-SNAPSHOT.jar
`

t-bast:
Closing this inactive PR.

PR #1198, opened 2019-11-02
[WIP] Status by external

btcontract:
Judging by myself an API user would want a payment to be searchable by `external_id`, this commit enables that option.


btcontract:
This turned out to be half baked as is, I'll try to improve this to make external id much more usable.

PR #1200, opened 2019-11-04
Add synchronization when locking database connection

dpad85:
This pull request fixes an issue where creating concurrent database connection could deadlock.

### Context

An exclusive lock was added to `eclair.sqlite` to prevent running several instances of eclair using the same sqlite database. The code obtaining the exclusive lock is a succession of [three SQL queries](https://github.com/ACINQ/eclair/blob/synchronized-exclusive-lock/eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteUtils.scala#L102-L108), and is not thread-safe. If two threads try to obtain the lock at the same time, none of these locks will succeed, and subsequent locks will fail because the database is deadlocked.

This can happen on the eclair mobile application where several background jobs (that need exclusive lock on the `eclair.sqlite` db) may be launched at the exact same time by the OS due to battery optimization.

See: 
- https://github.com/ACINQ/eclair-mobile/issues/230
- https://github.com/ACINQ/eclair-mobile/issues/227

### Proposed change

This PR changes the `SqliteUtils.obtainExclusiveLock` method into a synchronized statement. Also, if the SQL connections creation fails, db connections will be closed if needed.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
@@ -21,6 +21,7 @@ import java.sql.{Connection, DriverManager}
 
 import fr.acinq.eclair.db.sqlite._
 import grizzled.slf4j.Logging
+import org.sqlite.SQLiteException
 
 trait Databases {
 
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -35,24 +37,39 @@ trait Databases {
...
+      SqliteUtils.obtainExclusiveLock(sqliteEclair) // there should only be one process writing to this file
+      logger.info("successful lock on eclair.sqlite")
+      databaseByConnections(sqliteAudit, sqliteNetwork, sqliteEclair)
+    } catch {
+      case t: Throwable => {
+        logger.error("could not create connection to sqlite databases: ", t)
+        if (sqliteEclair != null) sqliteEclair.close()
+        if (sqliteNetwork != null) sqliteNetwork.close()
+        if (sqliteAudit != null) sqliteAudit.close()
+        throw t
```
t-bast: Do the background jobs really need access to `eclair.sqlite`? Or only `network.sqlite` (in which case we shouldn't have any issue)?

t-bast: Do we want to add retries with exponential back-off here? Would this help with your use-case?

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteUtils.scala
```diff
@@ -99,7 +99,7 @@ object SqliteUtils {
    *
    * The lock will be kept until the database is closed, or if the locking mode is explicitly reset.
    */
-  def obtainExclusiveLock(sqlite: Connection) {
+  def obtainExclusiveLock(sqlite: Connection) = synchronized {
```
t-bast: This only fixes the issue if two calls to `obtainExclusiveLock` happen in the same process, but in the case of a background job running concurrently with the wallet app, this won't fix anything, right?
It's very surprising that sqlite doesn't have a way to atomically get a real exclusive lock or fail...

dpad85: This scenario is taken care of by the front application and should not happen. There's a logic that prevents background jobs to run concurrently with the front.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
dpad85: I don't think it would help.

dpad85: There are 2 background jobs, one syncing the routing table that needs access to `network.sqlite`, and one watching channels that needs access to the `eclair.sqlite` file.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteUtils.scala
t-bast: Ok in that case the `synchronized` statement should work (and it doesn't hurt to have it anyway).
I still think we can probably do better on the sqlite side with a better understanding of their locking mechanisms, but it can be done later.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -35,24 +37,39 @@ trait Databases {
 
   val pendingRelay: PendingRelayDb
 
-  def backup(file: File) : Unit
+  def backup(file: File): Unit
 }
 
-object Databases {
+object Databases extends Logging {
```
t-bast: It seems like we should try to avoid extending this `Logging` trait and instead implicitly pass a logger to the functions that need one. Can you try doing that?

pm47: Using `grizzled.slf4j.Logging` is indeed discouraged, but for code completely unrelated to actors I think it makes sense to keep using it. What we don't want is lose the context when we call code in static methods from within an actor.

PR #1203, opened 2019-11-08
Payment Request: speed up node id computation

sstone:
LN payment requests almost never explicitly include a node id, instead it is recomputed using key recovery. This PR calls native secp256k1 pubkey recovery  which is much faster than using Bouncy Castle (it makes a difference for mobile users), and removes the signature check that is useless when pubkey was recovered.

Even though no one AFAICT actually includes the node id field in their PR it is still something that we should implement (in another PR)


eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentRequestSpec.scala
```diff
@@ -19,7 +19,7 @@ package fr.acinq.eclair.payment
 import java.nio.ByteOrder
 
 import fr.acinq.bitcoin.Crypto.{PrivateKey, PublicKey}
-import fr.acinq.bitcoin.{Block, ByteVector32, Crypto, Protocol}
+import fr.acinq.bitcoin.{Block, ByteVector32, ByteVector64, Crypto, Protocol}
```
t-bast: nit: this doesn't seem to be needed

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -447,12 +447,9 @@ object PaymentRequest {
     val bolt11Data = Codecs.bolt11DataCodec.decode(data).require.value
     val signature = ByteVector64(bolt11Data.signature.take(64))
     val message: ByteVector = ByteVector.view(hrp.getBytes) ++ data.dropRight(520).toByteVector // we drop the sig bytes
-    val (pub1, pub2) = Crypto.recoverPublicKey(signature, Crypto.sha256(message))
     val recid = bolt11Data.signature.last
-    val pub = if (recid % 2 != 0) pub2 else pub1
+    val pub = Crypto.recoverPublicKey(signature, Crypto.sha256(message), recid)
```
t-bast: It seems to me that we still need to validate the signature...
If we don't we can just end up recovering a public key that will simply not match the recipient, and we won't know that the invoice is completely invalid.

sstone: The signature is always valid since we use key recovery. We could compare the recovered node id to a user-provided value but I'm not sure it's needed and it would be another PR

t-bast: That's true, in fact signature verification here doesn't make any sense.
Maybe it's worth adding a comment explaining that for future me?

PR #1212, opened 2019-11-14
Rework truncated integers codecs

t-bast:
The spec defines tu64 (and friends) without the length prefix.
Multi-part uses a tu64 without a length prefix inside the PaymentData record.

Our previous implementation only supported using tu64 alone in a TLV record.
We make this more flexible by separating the length encoding.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
@@ -19,6 +19,7 @@ package fr.acinq.eclair.wire
 import fr.acinq.eclair.UInt64.Conversions._
 import fr.acinq.eclair.wire.CommonCodecs._
 import fr.acinq.eclair.{MilliSatoshi, UInt64}
+import scodec.bits.ByteVector
 import scodec.codecs._
 import scodec.{Attempt, Codec, Err}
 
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
```diff
@@ -28,52 +29,80 @@ import scodec.{Attempt, Codec, Err}
...
-    .\(0x00) { case i if i < 0x01 => i }(variableSizeUInt64(0, 0x00))
-    .\(0x01) { case i if i < 0x0100 => i }(variableSizeUInt64(1, 0x01))
-    .\(0x02) { case i if i < 0x010000 => i }(variableSizeUInt64(2, 0x0100))
-    .\(0x03) { case i if i < 0x01000000 => i }(variableSizeUInt64(3, 0x010000))
-    .\(0x04) { case i if i < 0x0100000000L => i }(variableSizeUInt64(4, 0x01000000))
-    .\(0x05) { case i if i < 0x010000000000L => i }(variableSizeUInt64(5, 0x0100000000L))
-    .\(0x06) { case i if i < 0x01000000000000L => i }(variableSizeUInt64(6, 0x010000000000L))
-    .\(0x07) { case i if i < 0x0100000000000000L => i }(variableSizeUInt64(7, 0x01000000000000L))
-    .\(0x08) { case i if i <= UInt64.MaxValue => i }(variableSizeUInt64(8, 0x0100000000000000L))
+  val tu64: Codec[UInt64] = Codec(
```
pm47: I have this, but it's not much better:

```scala
val tu64: Codec[UInt64] = Codec(
    (u: UInt64) => {
      def encode(bits: Int) = u.toByteVector.takeRight(bits)
      val b = u match {
        case u if u < 0x01 => encode(0)
        case u if u < 0x0100 => encode(1)
        case u if u < 0x010000 => encode(2)
        case u if u < 0x01000000 => encode(3)
        case u if u < 0x0100000000L => encode(4)
        case u if u < 0x010000000000L => encode(5)
        case u if u < 0x01000000000000L => encode(6)
        case u if u < 0x0100000000000000L => encode(7)
        case u if u <= UInt64.MaxValue => encode(8)
      }
      Attempt.successful(b.bits)
    },
    b => {
      def decode(min: UInt64) = minimalvalue(uint64, min).decode(b.padLeft(64))
      b.length match {
        case l if l <= 0 => decode(UInt64(0x00))
        case l if l <= 8 => decode(UInt64(0x01))
        case l if l <= 16 => decode(UInt64(0x0100))
        case l if l <= 24 => decode(UInt64(0x010000))
        case l if l <= 32 => decode(UInt64(0x01000000))
        case l if l <= 40 => decode(UInt64(0x0100000000L))
        case l if l <= 48 => decode(UInt64(0x010000000000L))
        case l if l <= 56 => decode(UInt64(0x01000000000000L))
        case l if l <= 64 => decode(UInt64(0x0100000000000000L))
        case _ => Attempt.failure(Err(s"too many bytes to decode for truncated uint64 (${b.toHex})"))
      }
    })
```

t-bast: I'll at least apply these refactorings, it looks a bit nicer but I don't see a fundamental way of escaping this overall pattern...at some point we need to have conditional branches for each of the integer slice range to ensure minimal encoding :/

PR #1214, opened 2019-11-14
Hosted channels

btcontract:
I perfectly understand this could be the most ridiculous PR this project has ever seen but, before dismissing it right away please consider that it's just a tool which I believe can improve Eclair experience without involving trust. 

As a quick example, user can connect to their own Eclair node this way, the benefits are increased usability ("remote control" wallets become possible without using API) and privacy (full Eclair becomes a relay, not receive node from sender's point of view).

Spec: https://github.com/btcontract/hosted-channels-rfc

High level rationale: https://gist.github.com/btcontract/d4122a79911eef2620f16b3dfe2850a8

It's not exactly complete at this point, at the very least it should properly use https://github.com/lightningnetwork/lightning-rfc/pull/681 once that is merged so I guess what I'm asking for at this point is concept and code review (which is a lot to ask and probably won't happen, but just giving it a shot).


t-bast:
Thanks for raising this, we won't dismiss it right away ;)

It will take us some time to review the concept and spec (before we can even review the code thoroughly) and we have a lot on our plate with MPP, Trampoline and Phoenix right now so we can't really commit to a timeline, but we do want to eventually spend some time on analyzing the pros/cons of this solution compared to other proposals that have been floating around for a while.

One first general code comment though: it would be easier to accept this PR to the codebase if most of it was done in an external [plugin](https://github.com/acinq/eclair#Plugins) (and changes to eclair-core itself were minimized). I don't know how hard that would be, and it may require improving our plugin system to add new hooks and events, which I think would be beneficial for many use-cases. If you have time to investigate that, it would be great!

btcontract:
> One first general code comment though: it would be easier to accept this PR to the codebase if most of it was done in an external plugin

On one hand this seems doable because hosted channel, related data types and gateway actor are all live "aside" of original Eclair code, but there are some rather deep ways this blends into system, I think plugin can't substitute the following:

- `Peer` actor needs to know that certain kind of messages are to be sent to `HostedGateway` actor, not to `Peer`'s normal channels (those would be hosted channel messages and normal messages such as `UpdateAddHtlc` which have a hosted channel id).  

- `ChannelCommitments` is now a trait with `HostedCommitments` and `Commitments` classes descending from it. Actors such as `Relayer` and `Register` now rely on this to see no difference between hosted and normal channels.  

- `Switchboard` and `HTLCReaper` actors are now aware of hosted channels existence and take them into account when deciding which HTLCs to be removed on startup.

- `Service` and `Eclair` have new API endpoints.

t-bast:
It seems that all of those can still be solved by adding a layer of indirection in these places where you have to blend in, but I agree it may make the eclair-core code more complex and harder to grasp which we want to avoid...

I still think our plugin system could be made more flexible (and I believe it's the direction we need to take if we want to add complex features that are outside of the RFC) but it's likely to be a tough task. Listing the things you need but can't do inside a plugin is a good first step; then we need to make the right changes to make it possible.

However, it's true that in your case you're replacing something quite fundamental: the channel management. So that may need actual changes inside eclair-core if we want to allow that.

PR #1218, opened 2019-11-18
Improved tracing of single/multi-part payments

pm47:
This includes a bit of refactoring in `MultiPartPaymentLifecycle`. Note
that we can't use the `onTermination` handler to finish the spans,
because it is asynchronous and may not be called after a long time.
That's why we use a dedicated `myStop` function.

In Kamon 2.0, by default spans are automatically generated for tracked
actors, which we don't want because we define our own spans. That's why
there is an additional configuration in `application.conf`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
@@ -33,6 +33,8 @@ import fr.acinq.eclair.payment.send.PaymentLifecycle.SendPayment
 import fr.acinq.eclair.router._
 import fr.acinq.eclair.wire._
 import fr.acinq.eclair.{CltvExpiry, FSMDiagnosticActorLogging, Logs, LongToBtcAmount, MilliSatoshi, NodeParams, ShortChannelId, ToMilliSatoshiConversion}
+import kamon.Kamon
+import kamon.context.Context
 import scodec.bits.ByteVector
 
 import scala.annotation.tailrec
@@ -53,6 +55,12 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, 
   require(cfg.id == cfg.parentId, "multi-part payment cannot have a parent payment")
 
   val id = cfg.id
+
+  private val span = Kamon.spanBuilder("multi-part-payment")
+    .tag("parentPaymentId", cfg.parentId.toString)
+    .tag("paymentHash", cfg.paymentHash.toHex)
+    .tag("targetNodeId", cfg.targetNodeId.toString())
+    .start()
 
   startWith(WAIT_FOR_PAYMENT_REQUEST, WaitingForRequest)
 
@@ -84,7 +92,11 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, 
         goto(PAYMENT_ABORTED) using PaymentAborted(d.sender, d.request, LocalFailure(BalanceTooLow) :: Nil, Set.empty)
       } else {
         val pending = setFees(d.request.routeParams, payments, payments.size)
-        pending.foreach { case (childId, payment) => spawnChildPaymentFsm(childId) ! payment }
+        Kamon.runWithContextEntry(parentPaymentIdKey, cfg.parentId) {
+          Kamon.runWithSpan(span, finishSpan = true) {
+            pending.foreach { case (childId, payment) => spawnChildPaymentFsm(childId) ! payment }
+          }
+        }
         goto(PAYMENT_IN_PROGRESS) using PaymentProgress(d.sender, d.request, d.networkStats, channels.length, 0 msat, d.request.maxAttempts - 1, pending, Set.empty, Nil)
       }
   }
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -45,10 +47,26 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
+      .tag("paymentHash", cfg.paymentHash.toHex)
+      .tag("targetNodeId", cfg.targetNodeId.toString())
+      .start()
+  }
+
   startWith(WAITING_FOR_REQUEST, WaitingForRequest)
 
   when(WAITING_FOR_REQUEST) {
     case Event(c: SendPaymentToRoute, WaitingForRequest) =>
+      span.tag("amount", c.finalPayload.amount.toLong)
```
pm47: @t-bast we could avoid repetition here by making `SendPaymentToRoute` and `SendPayment` derive from a common trait. Curious why you didn't do it?

t-bast: I didn't change it because it was already in the code, with the two types separate, but that can probably be done yeah ;)

t-bast:
Damn looks like one of the MPP integration tests is randomly failing (probably because balances are too low).
I need to stress test it to find the right value that should be sent (or maybe add a first payment in the other direction to make sure there's balance available).

pm47:
Never reproduced locally FWIW (and I've run the tests more than a few times...).

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -85,7 +93,11 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
         goto(PAYMENT_ABORTED) using PaymentAborted(d.sender, d.request, d.failures :+ LocalFailure(new RuntimeException("balance is too low")), d.pending.keySet)
       } else {
         val pending = setFees(d.request.routeParams, payments, payments.size + d.pending.size)
-        pending.foreach { case (childId, payment) => spawnChildPaymentFsm(childId) ! payment }
+        Kamon.runWithContextEntry(parentPaymentIdKey, cfg.parentId) {
```
t-bast: I find it surprising that this wraps the `pending.foreach` instead of being inside it (creating one new span for each child payment). Shouldn't the `Kamon.runWithContextEntry` be inside the `pending.foreach` and wraps only the `spawnChildPaymentFsm`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -220,6 +236,8 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
 
 object MultiPartPaymentLifecycle {
 
+  val parentPaymentIdKey = Context.key[UUID]("parentPaymentId", UUID.fromString("00000000-0000-0000-0000-000000000000"))
```
t-bast: What's this 0 UUID? Do we need to keep a map of static assigned UUIDs somewhere?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -45,10 +47,26 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
   val id = cfg.id
   val paymentsDb = nodeParams.db.payments
 
+  private val span = Kamon.runWithContextEntry(MultiPartPaymentLifecycle.parentPaymentIdKey, cfg.parentId) {
+    val spanBuilder = if (Kamon.currentSpan().isEmpty) {
```
t-bast: Can't you look at whether you received a `parentPaymentId` from a parent actor instead of checking `currentSpan().isEmpty`?

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -49,6 +49,7 @@ import fr.acinq.eclair.transactions.Transactions.{HtlcSuccessTx, HtlcTimeoutTx}
 import fr.acinq.eclair.wire._
 import fr.acinq.eclair.{CltvExpiryDelta, Kit, LongToBtcAmount, MilliSatoshi, Setup, ShortChannelId, randomBytes32}
 import grizzled.slf4j.Logging
+import kamon.Kamon
```
t-bast: Do you need that?

eclair-node/src/main/resources/application.conf
```diff
@@ -2,6 +2,20 @@ eclair {
     enable-kamon = false
 }
 
+kamon.instrumentation.akka {
+  filters {
+    actors {
+      # Decides which actors generate Spans for the messages they process, given that there is already an ongoing trace
+      # in the Context of the processed message (i.e. there is a Sampled Span in the Context).
+      #
```
t-bast: Maybe explain here why we're disabling by default (what's in the PR description)?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
pm47: This doesn't create a span, it just sets the context, so that spans created by the children actors occur within that same context.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
pm47: Kamon just needs a default key for some reason, it doesn't really matter a lot because we set it when we need it.

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
pm47: Nope!

PR #1219, opened 2019-11-20
MPP routing improvements

t-bast:
I found a few things than can be improved in the MPP retry behavior.
Those will be particularly helpful for Trampoline-MPP to a remote peer (take into account the fact that a Trampoline node will likely have many channels).

### Improvement 1: sending to a direct peer should not retry

It doesn't make any sense to let the `PaymentLifecycle` retry when we're forcing the choice of the first hop and the recipient is on the other end of that channel.
This is fixed by setting `maxAttempts=1` when sending child payments via direct channels to the recipient.

### Improvement 2: only use public channels for remote recipients

When the recipient isn't a direct peer, it's very unlikely that choosing a route prefix with an unannounced channel will succeed. Most unannounced channels Endurance/Horizon has are with mobile wallets. We ignore them when sending to remote nodes.

### Improvement 3: handle RouteNotFound

We're choosing the first hop without looking at the graph. That means we may choose a first hop that can't yield a complete route to our recipient. When that happens, we should avoid retrying with that same hop.

If we have a lot of channels, it's likely that we'll often choose a bad first hop. Since we only retry 5 times, this will lead to many payment failures which is bad when we're relaying (Trampoline). So we add some randomness on whether to count a `LocalFailure` as a complete attempt or "refund" that attempt.

Note that this hack won't be necessary once we move the MPP logic inside the router: when that happens we will only select first hops that are on a potentially valid route to the recipient.

### Unrelated

I also fixed the random failure that was happening in one of the MPP IntegrationSpec test.
With the code that's running on master, the `PaymentInitiator` doesn't guarantee that it will send the `PaymentId` before payment events.
Now we guarantee that.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
@@ -32,7 +32,7 @@ import fr.acinq.eclair.payment.send.PaymentInitiator.SendPaymentConfig
 import fr.acinq.eclair.payment.send.PaymentLifecycle.SendPayment
 import fr.acinq.eclair.router._
 import fr.acinq.eclair.wire.{Onion, OnionRoutingPacket, OnionTlv, PaymentTimeout, UpdateAddHtlc}
-import fr.acinq.eclair.{CltvExpiry, FSMDiagnosticActorLogging, Logs, LongToBtcAmount, MilliSatoshi, NodeParams, ToMilliSatoshiConversion}
+import fr.acinq.eclair.{CltvExpiry, FSMDiagnosticActorLogging, Logs, LongToBtcAmount, MilliSatoshi, NodeParams, ShortChannelId, ToMilliSatoshiConversion}
 import scodec.bits.ByteVector
 
 import scala.annotation.tailrec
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -110,8 +112,30 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
       case None =>
         // Get updated local channels (will take into account the child payments that are in-flight).
         relayer ! GetOutgoingChannels()
-        val failedPayment = d.pending(pf.id)
-        goto(WAIT_FOR_CHANNEL_BALANCES) using d.copy(toSend = d.toSend + failedPayment.finalPayload.amount, pending = d.pending - pf.id, failures = d.failures ++ pf.failures)
+        val newState = {
```
pm47: WDYT?

```scala
val failedPayment = d.pending(pf.id)
val shouldBlacklist = shouldBlacklistChannel(pf)
val ignoreChannels1 = if (shouldBlacklist) d.ignoreChannels + getFirstHopShortChannelId(failedPayment) else d.ignoreChannels
val remainingAttempts1 = if (shouldBlacklist && Random.nextDouble() * math.log(d.channelsCount.get) > 2.0) {
  // When we have a lot of channels, many of them may end up being a bad route prefix for the destination we're
  // trying to reach. This is a cheap error that is detected quickly (RouteNotFound), so we don't want to count
  // it in our payment attempts to avoid failing too fast.
  // However we don't want to test all of our channels either which would be expensive, so we only probabilistically
  // count the failure in our payment attempts.
  // With the log-scale used, here are the probabilities:
  //  * 10 channels -> refund 13% of failures
  //  * 20 channels -> refund 32% of failures
  //  * 50 channels -> refund 50% of failures
  //  * 100 channels -> refund 56% of failures
  //  * 1000 channels -> refund 70% of failures
  // NB: this hack won't be necessary once multi-part is directly handled by the router.
  d.remainingAttempts
} else d.remainingAttempts
val d1 = d.copy(
toSend = d.toSend + failedPayment.finalPayload.amount,
pending = d.pending - pf.id,
failures = d.failures ++ pf.failures,
ignoreChannels = ignoreChannels1,
remainingAttempts = remainingAttempts1
)
goto(WAIT_FOR_CHANNEL_BALANCES) using d1
```

PR #1221, opened 2019-11-22
add networkstats API call to allow display of network stats

n1bor:
Try 2...


t-bast:
Thanks for taking the time to do this.
I'm very surprised, Github still won't allow me to merge because it still claims that signatures are invalid...
I'll take your commit and commit it myself to make it available on master and will start a thread with Github support.

PR #1222, opened 2019-11-25
Avoid incoherent payment request features.

t-bast:
MPP implies payment secret.
Payment secret feature implies that there actually is a payment secret in the invoice.

It's still a bit awkward to use because `paymentSecret` still has to be an `Option`.
It's safe to call `.get` when `features.allowPaymentSecret` but it's not enforced by the type system.
Can we make this better at the type level without making the `PaymentRequest` unnecessarily complex?

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
@@ -18,6 +18,7 @@ package fr.acinq.eclair.payment
 
 import fr.acinq.bitcoin.Crypto.{PrivateKey, PublicKey}
 import fr.acinq.bitcoin.{Base58, Base58Check, Bech32, Block, ByteVector32, ByteVector64, Crypto}
+import fr.acinq.eclair.Features._
 import fr.acinq.eclair.payment.PaymentRequest._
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -43,8 +44,11 @@ case class PaymentRequest(prefix: String, amount: Option[MilliSatoshi], timestam
 
   amount.foreach(a => require(a > 0.msat, s"amount is not valid"))
   require(tags.collect { case _: PaymentRequest.PaymentHash => }.size == 1, "there must be exactly one payment hash tag")
-  require(tags.collect { case _: PaymentRequest.PaymentSecret => }.size <= 1, "there must be at most one payment secret tag")
   require(tags.collect { case PaymentRequest.Description(_) | PaymentRequest.DescriptionHash(_) => }.size == 1, "there must be exactly one description tag or one description hash tag")
+  require(!features.allowMultiPart || features.allowPaymentSecret, "there must be a payment secret when using multi-part")
```
pm47: I think the following is less error-prone:

```scala
if (features.allowMultiPart) {
  require(features.allowPaymentSecret, "there must be a payment secret when using multi-part")
}
```

t-bast: Agreed, done.

PR #1224, opened 2019-11-28
Fix MPP post-restart HTLC clean-up

t-bast:
We previously had some logic where we would fail incoming HTLCs
for which we were the final recipient when a channel would come online.

That made sense when we didn't have MPP, but with MPP we cannot do that.
There is a risk that we would be failing HTLCs that are considered received by the MPP FSM.
Instead we need to use the CommandBuffer when we are the final recipient.
This way pending commands cannot be lost and HTLCs are cleaned-up on restart.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
@@ -26,9 +26,9 @@ import fr.acinq.eclair.NodeParams
 import fr.acinq.eclair.blockchain.EclairWallet
 import fr.acinq.eclair.channel.Helpers.Closing
 import fr.acinq.eclair.channel._
-import fr.acinq.eclair.db.PendingRelayDb
-import fr.acinq.eclair.payment.relay.{CommandBuffer, Origin}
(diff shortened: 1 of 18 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -229,25 +238,30 @@ class HtlcReaper extends Actor with ActorLogging {
...
+            // this htlc is cross signed in the current commitment, we can settle it
+            preimage match {
+              case Some(preimage) =>
+                log.info(s"fulfilling broken htlc=$htlc")
+                channel ! CMD_FULFILL_HTLC(htlc.id, preimage, commit = true)
+              case None =>
+                log.info(s"failing broken htlc=$htlc")
+                channel ! CMD_FAIL_HTLC(htlc.id, Right(TemporaryNodeFailure), commit = true)
+            }
             false // the channel may very well be disconnected before we sign (=ack) the fail, so we keep it for now
```
pm47:
```suggestion
            false // the channel may very well be disconnected before we sign (=ack) the fail/fulfill, so we keep it for now
```

PR #1225, opened 2019-11-28
High level payments overview method

dpad85:
This PR add a method to get a high level/high performance overview of the payments from the payment database.

The `PaymentDb` interface is impacted to add two objects: `PlainIncomingPayment` and `PlainOutgoingPayment`. These objects only contains the most basic data about the payments. 

Notably, the payment request attribute in these object is the raw serialized value stored in the DB which helps with performances on Android. Two methods have been added to the `PaymentRequest` object to allow for the fast reading of a serialized payment request's description and expiry.

This PR supercedes #1169

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
@@ -26,7 +26,9 @@ import fr.acinq.eclair.{MilliSatoshi, ShortChannelId}
 
 import scala.compat.Platform
 
-trait PaymentsDb extends IncomingPaymentsDb with OutgoingPaymentsDb
+trait PaymentsDb extends IncomingPaymentsDb with OutgoingPaymentsDb {
+  def listPaymentsOverview(limit: Int): Seq[PlainPayment]
+}
 
 trait IncomingPaymentsDb {
   /** Add a new expected incoming payment (not yet received). */
@@ -194,3 +196,36 @@ object FailureSummary {
     case UnreadableRemoteFailure(route) => FailureSummary(FailureType.UNREADABLE_REMOTE, "could not decrypt failure onion", route.map(h => HopSummary(h)).toList)
   }
 }
+
+/**
+ * Generic payment trait holding only the minimum information in the most plain type possible. Notably, payment request
+ * is kept as a String, because deserialization is costly.
+ * <p>
+ * This object should only be used for a high level snapshot of the payments stored in the payment database.
+ * <p>
+ * Payment status should be of the correct type, but may not contain all the required data (routes, failures...).
+ */
+sealed trait PlainPayment {
+  val paymentHash: ByteVector32
+  val paymentRequest: Option[String]
+  val finalAmount: Option[MilliSatoshi]
+  val createdAt: Long
+  val completedAt: Option[Long]
+}
+
+case class PlainIncomingPayment(paymentHash: ByteVector32,
+                                finalAmount: Option[MilliSatoshi],
+                                paymentRequest: Option[String],
+                                status: IncomingPaymentStatus,
+                                createdAt: Long,
+                                completedAt: Option[Long],
+                                expireAt: Option[Long]) extends PlainPayment
+
+case class PlainOutgoingPayment(parentId: Option[UUID],
+                                externalId: Option[String],
+                                paymentHash: ByteVector32,
+                                finalAmount: Option[MilliSatoshi],
+                                paymentRequest: Option[String],
+                                status: OutgoingPaymentStatus,
+                                createdAt: Long,
+                                completedAt: Option[Long]) extends PlainPayment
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -26,7 +26,9 @@ import fr.acinq.eclair.{MilliSatoshi, ShortChannelId}
 
 import scala.compat.Platform
 
-trait PaymentsDb extends IncomingPaymentsDb with OutgoingPaymentsDb
+trait PaymentsDb extends IncomingPaymentsDb with OutgoingPaymentsDb {
+  def listPaymentsOverview(limit: Int): Seq[PlainPayment]
```
t-bast: Why not put that in its own trait? 
```scala
trait PaymentsDb extends IncomingPaymentsDb with OutgoingPaymentsDb with PaymentOverviewDb

trait PaymentOverviewDb {
    def listPaymentsOverview(limit: Int): Seq[PlainPayment]
}
```

This way other components in eclair-core don't need to care about this payment overview and use either the `IncomingPaymentsDb` or `OutgoingPaymentsDb` trait, whereas mobile apps can leverage this new trait.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -150,30 +158,31 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
...
-          rs.getBitVectorOpt("failures").map(b => paymentFailuresCodec.decode(b) match {
-            case Attempt.Successful(failures) => failures.value
+        case Some(completedAt) => Some(OutgoingPaymentStatus.Failed(
+          failures.map(b => paymentFailuresCodec.decode(b) match {
+            case Attempt.Successful(f) => f.value
             case Attempt.Failure(_) => Nil
           }).getOrElse(Nil),
           completedAt
         ))
         // Else it's still pending.
```
t-bast: nit: that comment isn't up-to-date.
You should either remove that comment or change this method to always return an `OutgoingPaymentStatus` and use `Some(OutgoingPaymentStatus.Pending)` here instead of `None`.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -233,6 +242,7 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
       statement.executeUpdate()
     }
 
+
```
t-bast: nit: why the additional blank line?

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
t-bast: I think the better choice is to have that method always return an `OutgoingPaymentStatus` instead of an `Option` (otherwise you're duplicating the `getOrElse(OutgoingPaymentStatus.Pending)`

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -487,6 +487,47 @@ object PaymentRequest {
     )
   }
 
+  private def readBoltData(input: String): Bolt11Data = {
```
t-bast: Since you're still decoding the data part, I'm honestly unsure that this is faster than completely decoding the invoice (now that we removed signature validation from it). Do you have some performance measures on android?
I'll do some testing on a laptop and will post my numbers here (if this isn't more performant, it's less error-prone to simply decode the invoice).

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -487,6 +487,47 @@ object PaymentRequest {
...
+  }
+
+  /**
+   * Extracts the description from a serialized payment request that is **expected to be valid**.
+   * Throws an error if the payment request is not valid.
+   *
+   * @param input valid serialized payment request
+   * @return description as a String. If the description is a hash, returns the hash value as a String.
+   */
+  def fastReadDescription(input: String): String = {
```
t-bast: If you're calling both `fastReadDescription` and `fastHasExpired` for the same invoice, you'll be decoding the data twice...if that's the case these methods should take the decoded data as input instead of the string?

dpad85: These two methods are indeed called twice. 

However `fastHasExpired` is called very early, in `listPaymentsOverview` by `buildIncomingPaymentStatus` to get the status of an incoming payment. Note that it is also used by the regular `xxxIncomingPayment` methods so changing this method's signature would have some side-effects.

`fastReadDescription` is called at the last moment when actually displaying the description from the payment request (so it's only called for the currently visible payments, and not called for payments that are in the list but not displayed).

Also, AFAICT on Android the slow part in the `PaymentRequest.read` is not reading the bolt11 data but the recovering the public key.

t-bast: Ok makes sense, we can re-visit that later if we want to improve the perf even more :)

PR #1228, opened 2019-12-02
Handle chain re-org in balance computation

t-bast:
If a chain re-org happens and a new ShortChannelId is assigned,
the `Relayer` kept both entries (new and old).

This resulted in an incorrect balance because we effectively counted this channel twice.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
@@ -36,7 +36,7 @@ case class ChannelRestored(channel: ActorRef, peer: ActorRef, remoteNodeId: Publ
 
 case class ChannelIdAssigned(channel: ActorRef, remoteNodeId: PublicKey, temporaryChannelId: ByteVector32, channelId: ByteVector32) extends ChannelEvent
 
-case class ShortChannelIdAssigned(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId) extends ChannelEvent
+case class ShortChannelIdAssigned(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, previousShortChannelId: Option[ShortChannelId] = None) extends ChannelEvent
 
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
```diff
@@ -36,7 +36,7 @@ case class ChannelRestored(channel: ActorRef, peer: ActorRef, remoteNodeId: Publ
 
 case class ChannelIdAssigned(channel: ActorRef, remoteNodeId: PublicKey, temporaryChannelId: ByteVector32, channelId: ByteVector32) extends ChannelEvent
 
-case class ShortChannelIdAssigned(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId) extends ChannelEvent
+case class ShortChannelIdAssigned(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, previousShortChannelId: Option[ShortChannelId] = None) extends ChannelEvent
```
pm47: Why is `previousShortChannelId` optional if it can always be defined?

t-bast: It's only defined if you had a *different* previous shortChannelId.
Do you mean that we should instead always provide it, and have the same value in `shortChannelId` and `previousShortChannelId` when it doesn't change?
It feels a bit wasteful to me because it happens very infrequently, and would require changing a bit more code.

pm47: No I just missed the fact that you had defined a default `None` value. I tend to avoid that as much as possible because it is easier to estimate impacts on existing code.

t-bast: Got it, I can remove that default `None` and explicitly set it when this event is created, would that be ok?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -96,6 +97,17 @@ class Relayer(nodeParams: NodeParams, register: ActorRef, commandBuffer: ActorRe
       }
       context become main(channelUpdates1, node2channels)
 
+    case ShortChannelIdAssigned(_, channelId, shortChannelId, previousShortChannelId) =>
+      previousShortChannelId.foreach(previousShortChannelId => {
```
pm47: We should probably check that `shortChannelId != previousShortChannelId`.

PR #1237, opened 2019-12-06
Register: ShortChannelId can change because of re-orgs

t-bast:
I forgot to do this small clean-up when I did #1228

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
@@ -48,8 +48,13 @@ class Register extends Actor with ActorLogging {
     case ChannelIdAssigned(channel, remoteNodeId, temporaryChannelId, channelId) =>
       context become main(channels + (channelId -> channel) - temporaryChannelId, shortIds, channelsTo + (channelId -> remoteNodeId) - temporaryChannelId)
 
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
```diff
@@ -48,8 +48,13 @@ class Register extends Actor with ActorLogging {
     case ChannelIdAssigned(channel, remoteNodeId, temporaryChannelId, channelId) =>
       context become main(channels + (channelId -> channel) - temporaryChannelId, shortIds, channelsTo + (channelId -> remoteNodeId) - temporaryChannelId)
 
-    case ShortChannelIdAssigned(_, channelId, shortChannelId, _) =>
-      context become main(channels, shortIds + (shortChannelId -> channelId), channelsTo)
+    case ShortChannelIdAssigned(_, channelId, shortChannelId, previousShortChannelId) =>
```
pm47:
```suggestion
    case ShortChannelIdAssigned(_, channelId, shortChannelId, previousShortChannelId_opt) =>
```

to prevent shadowing

t-bast:
Thinking about it some more, maybe it's a feature that Register keeps the mapping `oldShortChannelId->channelId`. This doesn't waste a lot of RAM and may save us in some tricky re-org cases.

PR #1240, opened 2019-12-07
Postgres support

btcontract:
This adds `postgre` invariant for channels, peers and pending relay databases, no attempts to incorporate this into system are done here, I propose this PR to be used as basis when adding a full `postgre` support.

Other than `postgre` I've introduced a `HasHtlcId` trait for `CMD_FULFILL_HTLC, CMD_FAIL_HTLC and CMD_FAIL_MALFORMED_HTLC` which simplifies things a bit.

Notes about testing: looks like there is no way to get a throwaway in-memory db similar to how it's done for `sqlite` so for basic testing a real local `eclairtest` db is used. I don't yet know how to elegantly go about more advanced tests (such as those in `channel` package) because those rely on creating of thousands of in-memory databases. To convince myself `postgre` code works with all tests I did create thousands of real `postgre` databases but did not include any of that here because it's obviously not the right way to do it.


t-bast:
I think this PR contains two different things. The `HasHtlcId` refactoring and groundwork for PostgreSQL.
I like the `HasHtlcId` change and the small fixes/doc you introduced. Could you put that in its own PR (that we would integrate quickly)? And have the PostgreSQL work in a separate PR (that will need more work on our side to test/brainstorm the design)?

t-bast:
Regarding the testing, I think that as long as the PostgreSQL implementation of the DB traits is well tested, it doesn't need to be integrated in the other tests (they'll keep using the in-memory SQLite).

The goal of having those DB traits is that the rest of the application doesn't care about the implementation: so as long as all implementation (SQLite and PostgreSQL) are properly tested to implement the trait the same way, the rest of the application can pick the simplest one.

Christewart:
>Notes about testing: looks like there is no way to get a throwaway in-memory db similar to how it's done for sqlite so for basic testing a real local eclairtest db is used. I don't yet know how to elegantly go about more advanced tests (such as those in channel package) because those rely on creating of thousands of in-memory databases. To convince myself postgre code works with all tests I did create thousands of real postgre databases but did not include any of that here because it's obviously not the right way to do it.

You should create and destroy tables per unit tests in your fixtures

Here is some harness code we use over in bitcoin-s 

https://github.com/bitcoin-s/bitcoin-s/blob/master/testkit/src/main/scala/org/bitcoins/testkit/fixtures/BitcoinSFixture.scala#L25

`build` should be the logic to create the table, and `destroy` should be the logic to take down the table. 

You can see an example implementation here in bitcoin-s where we create a block header table and then destroy it after the test run.

https://github.com/bitcoin-s/bitcoin-s/blob/master/testkit/src/main/scala/org/bitcoins/testkit/chain/ChainUnitTest.scala#L149

Finally an example of how to use it in a test case: 

https://github.com/bitcoin-s/bitcoin-s/blob/master/chain-test/src/test/scala/org/bitcoins/chain/models/BlockHeaderDAOTest.scala#L16

https://github.com/bitcoin-s/bitcoin-s/blob/master/chain-test/src/test/scala/org/bitcoins/chain/models/BlockHeaderDAOTest.scala#L24

Christewart:
We also did a POC of doing this with eclair a few weeks back that can be found here: 

https://github.com/rorp/eclair/tree/postgresql

This manually re-writes all of the sql queries ( :cry: ) to make them compatible with postgres to avoid taking slick as a dependency.  I would prefer using Slick as I think it is the most scalable way to support [a bunch of different databases](https://scala-slick.org/doc/3.3.0/supported-databases.html) but I think other ACINQ developers aren't on board with that AFAIK.

btcontract:
> I would prefer using Slick as I think it is the most scalable way to support a bunch of different databases but I think other ACINQ developers aren't on board with that AFAIK.

I'd like to hear more about that if it's true. On mobile slick is definitely a bad option but for server setups I see no theoretical issues so far.

rorp:
I didn't use `slick` for my POC because I think it seems like overkill.

The database schema is quite simple, conceptually it's just a key-value store. A NoSQL database (like `DynamoDB`) could do the trick as well, but it's not supported by `slick`.

`slick` requires 3 more dependencies to achieve the same goal. But most of the users won't need anything else than `Sqlite`. Also that's 3 more potential attack vectors.

Personally, I'd like to see a pluggable DB interface in eclair, so the end users could drop a DB plugin jar-file into their class path and use whatever database they want. I think would require some restructuring, like extracting the traits from `fr.acinq.eclair.db` into a separate jar-file, etc.

eclair-core/pom.xml
```diff
@@ -194,6 +194,21 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.typesafe.slick</groupId>
+            <artifactId>slick-hikaricp_${scala.version.short}</artifactId>
+            <version>3.3.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <version>9.4-1206-jdbc42</version>
```
rorp: I believe the latest driver version is 42.2.9: https://jdbc.postgresql.org/

eclair-core/src/main/resources/reference.conf
```diff
@@ -163,6 +163,20 @@ eclair {
     port = 9051
     private-key-file = "tor.dat"
   }
+
+  postgre = {
+    connectionPool = "HikariCP"
+    dataSourceClass = "org.postgresql.ds.PGSimpleDataSource"
```
rorp: `PGSimpleDataSource` should not be used in production: https://jdbc.postgresql.org/documentation/publicapi/org/postgresql/ds/PGPoolingDataSource.html

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -43,6 +46,23 @@ object TestConstants {
...
+                                                  |    postgre = {
+                                                  |      connectionPool = "HikariCP"
+                                                  |      dataSourceClass = "org.postgresql.ds.PGSimpleDataSource"
+                                                  |      numThreads = 6
+                                                  |
+                                                  |      properties = {
+                                                  |        serverName = "localhost"
+                                                  |        portNumber = "5432"
+                                                  |        databaseName = "eclairtest"
+                                                  |        user = "postgres"
```
rorp: This file need editing for running tests. Any Idea how to make the tests working without changing the config and installing `postgresql`?

rorp:
I noticed that not all tables were ported to Postgres. What's the rationale for this?

rorp:
Also, I concur with @t-bast that it's not a really good idea to mix two big changes in one PR.

eclair-core/src/main/resources/reference.conf
btcontract: Can you elaborate on risks of using `PGSimpleDataSource`. Doc over there has a red warning, I'm not sure I understand what it means.

pm47:
https://github.com/ACINQ/eclair/pull/1249#issuecomment-564520827

t-bast:
@btcontract it seems that #1249 is almost ready.
Do you have time to review it? I'd like these two PRs to converge so that we only have one to review and integrate :)

btcontract:
Sadly, not at the moment. I'd also just rather close this one if no one minds since #1249 is much better developed.

t-bast:
> I'd also just rather close this one if no one minds since #1249 is much better developed.

SGTM, let's focus our postgres efforts on #1249

PR #1241, opened 2019-12-07
Minor: missing log category in peer

pm47:


araspitzu:
A rebase on master would include the fix for the flaky test from #1243

PR #1244, opened 2019-12-09
Add a configurable time-out to onchain fee provider requests

sstone:
We configure a timeout of 5 seconds, applicable to all fee providers. If a provider times out we switch to the next one in our list.
Our mobile app needs a feerate to start properly and currently waits too long when a fee provider is online but very slow to respond.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -31,6 +31,7 @@ eclair {
 
   min-feerate = 3 // minimum feerate in satoshis per byte
   smooth-feerate-window = 6 // 1 = no smoothing
+  feerate-provider-timeout = 5 seconds // max time we'll wait for answers from a fee provider before we fallback to the next one
 
   node-alias = "eclair"
   node-color = "49daaa"
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
@@ -26,7 +26,7 @@ import org.json4s.jackson.Serialization
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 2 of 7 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -31,6 +31,7 @@ eclair {
 
   min-feerate = 3 // minimum feerate in satoshis per byte
   smooth-feerate-window = 6 // 1 = no smoothing
+  feerate-provider-timeout = 1 second // max time we'll wait for answers from a fee provider before we fallback to the next one
```
pm47: Timeout is really `1 second`, not `5 seconds`.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
```diff
@@ -23,9 +23,10 @@ import org.json4s.DefaultFormats
 import org.json4s.JsonAST.{JInt, JValue}
 import org.json4s.jackson.Serialization
 
+import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 
-class BitgoFeeProvider(chainHash: ByteVector32)(implicit http: SttpBackend[Future, Nothing], ec: ExecutionContext) extends FeeProvider {
+class BitgoFeeProvider(chainHash: ByteVector32, readTimeOut: Duration = 5 seconds)(implicit http: SttpBackend[Future, Nothing], ec: ExecutionContext) extends FeeProvider {
```
pm47: Default values can be error-prone, as it is the case here.

eclair-core/src/main/resources/reference.conf
t-bast: I think we should set 5 seconds here and no default value in the object constructors

sstone: Agreed.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
sstone: Agreed, I've removed them

sstone:
I've added timeout tests (with a timeout of 1 ms to make sure it always fails) to the BigGo and Earn fee provider tests.

PR #1245, opened 2019-12-10
Add HasHtlcIdCommand

btcontract:
This is just `HasHtlcIdCommand` without `postgre`, as asked at https://github.com/ACINQ/eclair/pull/1240#issuecomment-563220559


eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -212,14 +212,14 @@ object Switchboard {
...
     // the command from the pending relay db.
-    val channel2Htlc: Set[(ByteVector32, Long)] =
-    channels
-      .flatMap(_.commitments.remoteCommit.spec.htlcs)
-      .filter(_.direction == OUT)
-      .map(htlc => (htlc.add.channelId, htlc.add.id))
-      .toSet
-
-    val pendingRelay: Set[(ByteVector32, Long)] = relayDb.listPendingRelay()
+    val channel2Htlc: Seq[(ByteVector32, Long)] = for {
```
t-bast: I preferred the previous code, why did you change that? And this still should be a set imo.

eclair-core/src/main/scala/fr/acinq/eclair/db/PendingRelayDb.scala
```diff
@@ -17,30 +17,28 @@
...
 
   def removePendingRelay(channelId: ByteVector32, htlcId: Long)
 
-  def listPendingRelay(channelId: ByteVector32): Seq[Command]
-
-  def listPendingRelay(): Set[(ByteVector32, Long)]
+  def listPendingRelay(channelId: ByteVector32): Seq[HasHtlcIdCommand]
 
-  def close(): Unit
+  def listPendingRelay(): Set[(ByteVector32, HasHtlcIdCommand)]
```
t-bast: Why do you change the signature? We currently only need `(ByteVector32, Long)` here which is more efficient to retrieve from the DB that decoding the command (especially since we don't use it).

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
btcontract: for aesthetic reasons :) somehow `for` seems more elegant to me, I'll change back.

eclair-core/src/main/scala/fr/acinq/eclair/db/PendingRelayDb.scala
btcontract: Returning `HasHtlcIdCommand` seems more general and can be used for other purposes in the future, also currently `listPendingRelay` is only used on startup once so efficiency does not seem that important. This is not a big deal though, I can change it back.

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
t-bast: Thanks

eclair-core/src/main/scala/fr/acinq/eclair/db/PendingRelayDb.scala
t-bast: To me this is YAGNI (you ain't gonna need it) :)
I would change that only when we actually need the whole command somewhere, especially since as you pointed out in other discussions we're reading from `pendingRelay` quite often (channel state changes) so we'd like it to be as lightweight as possible (even though it's not very lightweight).

t-bast:
Usual commit tidyness stuff: can you please:

* squash all the changes in a single commit
* rebase this single commit on master
* sign this commit
* force-push to this branch

Then we can merge this properly.

btcontract:
Did I do it right? (sorry, this is a part of git I'm not familiar with yet)

t-bast:
Unfortunately no :)
There are still 18 commits in the PR and none of them are signed.

Here are instructions on setting up signed commits: https://help.github.com/en/github/authenticating-to-github/signing-commits

To clean up your branch, here is what you should do (assuming that you configured acinq/eclair as upstream, see https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/configuring-a-remote-for-a-fork):

```sh
git checkout master
git fetch upstream
git merge upstream/master
git checkout has-htlc-id
git rebase -i master
```

This will open a summary file with all the commits.
Choose to squash them all into the first one, fix the conflicts, and that will produce a clean `has-htlc-id` branch with a single commit.

We need to put these instructions in a clean `Contributing.md` file, we'll get to that soon(-ish).

btcontract:
Looks like I have finally got it right, sorry for the trouble!

PR #1246, opened 2019-12-10
MPP: don't retry if failure comes from final recipient

t-bast:


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -249,11 +249,11 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
   }
 
   def handleChildFailure(pf: PaymentFailed, d: PaymentProgress): Option[PaymentAborted] = {
-    val paymentTimedOut = pf.failures.exists {
-      case f: RemoteFailure => f.e.failureMessage == PaymentTimeout
+    val recipientFailure = pf.failures.exists {
```
pm47: Would the intention be slightly more clear?

```scala
    if (pf.failures.collectFirst { case f: RemoteFailure if f.e.originNode == d.request.targetNodeId => f }.isDefined) {
      // final node error
      Some(PaymentAborted(d.sender, d.request, d.failures ++ pf.failures, d.pending.keySet - pf.id))
    } else if (d.remainingAttempts == 0) {
      // no more retries
      val failure = LocalFailure(RetryExhausted)
      Some(PaymentAborted(d.sender, d.request, d.failures ++ pf.failures :+ failure, d.pending.keySet - pf.id))
    } else {
      None
    }
```

t-bast: Changing to `collectFirst` makes it easier to understand.
I'd still put that result in a boolean variable instead of inline in the `if`.

PR #1247, opened 2019-12-10
Fix Sphinx privacy leak

t-bast:
This PR fixes two small issues:

* [Sphinx privacy leak](https://github.com/lightningnetwork/lightning-rfc/pull/697)
* [Recipient should not collect fees](https://github.com/lightningnetwork/lightning-rfc/pull/711)


eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -214,15 +214,14 @@ object Sphinx extends Logging {
      * @param onionPayloadFiller optional onion payload filler, needed only when you're constructing the last packet.
```
sstone: nit: doc for `packet` needs to be updated.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -214,15 +214,14 @@ object Sphinx extends Logging {
...
-    def wrap(payload: ByteVector, associatedData: ByteVector32, ephemeralPublicKey: PublicKey, sharedSecret: ByteVector32, packet: Option[wire.OnionRoutingPacket], onionPayloadFiller: ByteVector = ByteVector.empty): wire.OnionRoutingPacket = {
+    def wrap(payload: ByteVector, associatedData: ByteVector32, ephemeralPublicKey: PublicKey, sharedSecret: ByteVector32, packet: Either[ByteVector, wire.OnionRoutingPacket], onionPayloadFiller: ByteVector = ByteVector.empty): wire.OnionRoutingPacket = {
       require(payload.length <= PayloadLength - MacLength, s"packet payload cannot exceed ${PayloadLength - MacLength} bytes")
 
       val (currentMac, currentPayload): (ByteVector32, ByteVector) = packet match {
-        // Packet construction starts with an empty mac and payload.
-        case None => (ByteVector32.Zeroes, ByteVector.fill(PayloadLength)(0))
-        case Some(p) => (p.hmac, p.payload)
+        // Packet construction starts with an empty mac and random payload.
+        case Left(startingBytes) => (ByteVector32.Zeroes, startingBytes)
```
sstone: Are there constraints on the size of `startingBytes` ? (it looks like it should match the payload length ?)

t-bast: Indeed, I'll add a `require`

PR #1248, opened 2019-12-10
Use correct type for parameters in ExtendedBitcoinClient

araspitzu:
Replace the usage of `String` with the more precise `ByteVector32`

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
@@ -61,15 +61,9 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
 
   def signTransaction(tx: Transaction): Future[SignTransactionResponse] = signTransaction(Transaction.write(tx).toHex)
 
-  def getTransaction(txid: ByteVector32): Future[Transaction] = bitcoinClient.getTransaction(txid)
-
   def publishTransaction(tx: Transaction)(implicit ec: ExecutionContext): Future[String] = bitcoinClient.publishTransaction(tx)
 
-  def publishTransaction(hex: String)(implicit ec: ExecutionContext): Future[String] = publishTransaction(Transaction.read(hex))
-
   def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = rpcClient.invoke("lockunspent", true, outPoints.toList.map(outPoint => Utxo(outPoint.txid, outPoint.index))) collect { case JBool(result) => result }
-
-  def isTransactionOutputSpendable(txid: ByteVector32, outputIndex: Int, includeMempool: Boolean)(implicit ec: ExecutionContext): Future[Boolean] = bitcoinClient.isTransactionOutputSpendable(txid, outputIndex, includeMempool)
 
   override def getBalance: Future[Satoshi] = rpcClient.invoke("getbalance") collect { case JDecimal(balance) => Satoshi(balance.bigDecimal.scaleByPowerOfTen(8).longValue()) }
 
@@ -110,7 +104,7 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
     .map(_ => true) // if bitcoind says OK, then we consider the tx successfully published
     .recoverWith { case JsonRPCError(e) =>
     logger.warn(s"txid=${tx.txid} error=$e")
-    getTransaction(tx.txid).map(_ => true).recover { case _ => false } // if we get a parseable error from bitcoind AND the tx is NOT in the mempool/blockchain, then we consider that the tx was not published
+    bitcoinClient.getTransaction(tx.txid).map(_ => true).recover { case _ => false } // if we get a parseable error from bitcoind AND the tx is NOT in the mempool/blockchain, then we consider that the tx was not published
   }
     .recover { case _ => true } // in all other cases we consider that the tx has been published
 
@@ -118,7 +112,7 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
 
   override def doubleSpent(tx: Transaction): Future[Boolean] =
   for {
-    exists <- getTransaction(tx.txid)
+    exists <- bitcoinClient.getTransaction(tx.txid)
       .map(_ => true) // we have found the transaction
       .recover {
       case JsonRPCError(Error(_, message)) if message.contains("index") =>
@@ -134,7 +128,7 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
       // if the tx wasn't in the blockchain and one of it's input has been spent, it is double-spent
       // NB: we don't look in the mempool, so it means that we will only consider that the tx has been double-spent if
       // the overriding transaction has been confirmed at least once
-      Future.sequence(tx.txIn.map(txIn => isTransactionOutputSpendable(txIn.outPoint.txid, txIn.outPoint.index.toInt, includeMempool = false))).map(_.exists(_ == false))
+      Future.sequence(tx.txIn.map(txIn => bitcoinClient.isTransactionOutputSpendable(txIn.outPoint.txid, txIn.outPoint.index.toInt, includeMempool = false))).map(_.exists(_ == false))
     }
   } yield doublespent
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
@@ -19,15 +19,12 @@ package fr.acinq.eclair.blockchain.bitcoind
 import fr.acinq.bitcoin._
 import fr.acinq.eclair._
 import fr.acinq.eclair.blockchain._
-import fr.acinq.eclair.blockchain.bitcoind.rpc.{BitcoinJsonRPCClient, Error, JsonRPCError}
+import fr.acinq.eclair.blockchain.bitcoind.rpc.{BitcoinJsonRPCClient, Error, ExtendedBitcoinClient, JsonRPCError}
 import fr.acinq.eclair.transactions.Transactions
 import grizzled.slf4j.Logging
-import org.json4s.DefaultFormats
 import org.json4s.JsonAST._
-import org.json4s.jackson.Serialization
 import scodec.bits.ByteVector
 
-import scala.compat.Platform
 import scala.concurrent.{ExecutionContext, Future}
 
 /**
@@ -36,6 +33,8 @@ import scala.concurrent.{ExecutionContext, Future}
 class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionContext) extends EclairWallet with Logging {
 
   import BitcoinCoreWallet._
+
+  val bitcoinClient = new ExtendedBitcoinClient(rpcClient)
 
   def fundTransaction(hex: String, lockUnspents: Boolean, feeRatePerKw: Long): Future[FundTransactionResponse] = {
     val feeRatePerKB = BigDecimal(feerateKw2KB(feeRatePerKw))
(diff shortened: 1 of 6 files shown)


pm47:
You missed those ones:
https://github.com/ACINQ/eclair/blob/61efb22053c293a27e3b64acacbe0fb4aabdba63/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala#L73
https://github.com/ACINQ/eclair/blob/61efb22053c293a27e3b64acacbe0fb4aabdba63/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala#L148

Also maybe take the opportunity to homogeneize between `txId` and `txid`? I think the latter is what `bitcoin-lib` uses.

araspitzu:
Did miss them but they're out of scope for the PR. In fact i think we shouldn't just fix them in `BitcoinCoreWallet` but we should use `ExtendedBitcoinClient` , if you notice `isTransactionOutputSpendable` and it's implementation are duplicated between wallet and client. What do you think of using the `ExtendedBitcoinClient` inside `BitcoinCoreWallet`?

pm47:
I think the problem is that `ExtendedBitcoinClient` is bulky and I didn't want to make the small `BitcoinCoreWallet` depend on it.

How about making `ExtendedBitcoinClient` use the pimp pattern to extend `BitcoinJsonRPCClient`, the same way we made `ExtendedResultSet` extend `ResultSet` in order to do `rs.getByteVector` with the database? We could then move all the bitcoind-related functions of `BitcoinCoreWallet` to `ExtendedBitcoinClient` and that would address your concern.

pm47:
Wait I'm changing my mind. I don't think pimp pattern is a good idea because we are not extending `BitcoinJsonRPCClient` with other functions similar to the ones it already has, we are *adding* a set of functions that are much more higher level than what it currently does.

pm47:
> What do you think of using the ExtendedBitcoinClient inside BitcoinCoreWallet?

Yes let's do this. I think maybe as a next step we could consider making `ExtendedBitcoinClient` an object.

araspitzu:
> Wait I'm changing my mind. I don't think pimp pattern is a good idea because we are not extending `BitcoinJsonRPCClient` with other functions similar to the ones it already has, we are _adding_ a set of functions that are much more higher level than what it currently does.

:+1: I was also hesitant with making `ExtendedBitcoinClient` use the pimp pattern to extend `BitcoinJsonRPCClient`, it feels more natural that one uses the other but without extending it. Maybe we should take some time to clean up `ExtendedBitcoinClient` to be able to use it from `BitcoinCoreWallet`, i would like to remove the code duplication that we currently have. I'm in favour of doing the necessary cleansing here in this PR.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -70,7 +70,7 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
 
   def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = rpcClient.invoke("lockunspent", true, outPoints.toList.map(outPoint => Utxo(outPoint.txid.toString, outPoint.index))) collect { case JBool(result) => result }
 
-  def isTransactionOutputSpendable(txId: String, outputIndex: Int, includeMempool: Boolean)(implicit ec: ExecutionContext): Future[Boolean] = rpcClient.invoke("gettxout", txId, outputIndex, includeMempool) collect { case j => j != JNull }
+  def isTransactionOutputSpendable(txid: String, outputIndex: Int, includeMempool: Boolean)(implicit ec: ExecutionContext): Future[Boolean] = rpcClient.invoke("gettxout", txid, outputIndex, includeMempool) collect { case j => j != JNull }
```
pm47: Are you not fixing the type here as well? There is also the `Utxo` class.

araspitzu: Fixed, the `Utxo` class was a bit controversial because it's used to encapsulate parameters for the RPC call, using objects there requires to have a json serializer in scope.

araspitzu:
Latest commit 3410f46 s a small improvement in the `BitcoinCoreWallet` to remove duplicate code. I didn't migrate the `ExtendedBitcoinClient` to be an `object` because it makes it harder to test and breaks composability.

t-bast:
Is there something blocking this PR (apart from a rebase)?
It looks like a refactoring we may want to include before the next release.

araspitzu:
Rebased on master

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -19,7 +19,7 @@ package fr.acinq.eclair.blockchain.bitcoind
 import fr.acinq.bitcoin._
 import fr.acinq.eclair._
 import fr.acinq.eclair.blockchain._
-import fr.acinq.eclair.blockchain.bitcoind.rpc.{BitcoinJsonRPCClient, Error, JsonRPCError}
+import fr.acinq.eclair.blockchain.bitcoind.rpc.{BitcoinJsonRPCClient, Error, ExtendedBitcoinClient, JsonRPCError}
```
t-bast: nit: there are many unused imports now

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
```diff
@@ -18,18 +18,25 @@ package fr.acinq.eclair.blockchain.bitcoind.rpc
...
+import org.json4s.{CustomSerializer, DefaultFormats}
+import org.json4s.JsonAST.{JString, JValue}
 import org.json4s.jackson.Serialization
 
 import scala.concurrent.{ExecutionContext, Future}
 
 class BasicBitcoinJsonRPCClient(user: String, password: String, host: String = "127.0.0.1", port: Int = 8332, ssl: Boolean = false)(implicit http: SttpBackend[Future, Nothing]) extends BitcoinJsonRPCClient {
 
+  // necessary to properly serialize ByteVector32 into String readable by bitcoind
+  object ByteVector32Serializer extends CustomSerializer[ByteVector32](_ => ( {
```
t-bast: Do we have a test that verifies we correctly serialize in this client?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -34,81 +34,81 @@ class ExtendedBitcoinClient(val rpcClient: BitcoinJsonRPCClient) {
 
   implicit val formats = org.json4s.DefaultFormats
 
-  def getTxConfirmations(txId: String)(implicit ec: ExecutionContext): Future[Option[Int]] =
-    rpcClient.invoke("getrawtransaction", txId, 1) // we choose verbose output to get the number of confirmations
+  def getTxConfirmations(txid: ByteVector32)(implicit ec: ExecutionContext): Future[Option[Int]] =
+    rpcClient.invoke("getrawtransaction", txid.toHex, 1) // we choose verbose output to get the number of confirmations
```
t-bast: Instead of doing `txid.toHex`, can't you do the same as `BasicBitcoinJsonRPCClient` and add a serializer for ByteVector32? Maybe worth adding a `Serializers.scala` file in the `rpc` folder to share that.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
araspitzu: Yes, [this test](https://github.com/ACINQ/eclair/blob/5eaee701e65fb6407b849dee5cf9b0b084193ce6/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala#L169) uses the [rollback](https://github.com/ACINQ/eclair/pull/1248/files#diff-f0e4b1153995aaf8ef920c12f6c4c36bR73) which in turn uses `Utxo` with a `ByteVector32` in it, the test fails if we don't correctly serialize the bytevector.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
araspitzu: Since the serializer is declared in `BasicBitcoinJsonRPCClient` we can already pass `ByteVector32` to 'invoke', i applied the changes in the latest commit.

araspitzu:
For reviewers: the commit history of this PR is not great let me know if you want a squash/rebase.

t-bast:
>  let me know if you want a squash/rebase

Yes let's start fresh with a rebase (maybe once cltv-tie-breaker is merged since there will be conflicts).

araspitzu:
Rebased on master and squash/merged the commit history of this PR.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -133,7 +133,7 @@ class ExtendedBitcoinClient(val rpcClient: BitcoinJsonRPCClient) {
         Future.successful(tx.txid.toString())
       case e@JsonRPCError(Error(-25, _)) =>
         // "missing inputs (code: -25)" it may be that the tx has already been published and its output spent
-        getRawTransaction(tx.txid.toString()).map { case _ => tx.txid.toString() }.recoverWith { case _ => Future.failed[String](e) }
+        getRawTransaction(tx.txid).map { case _ => tx.txid.toString() }.recoverWith { case _ => Future.failed[String](e) }
```
t-bast: Why not `txid.toHex` to be consistent (the end result is the same since `toString()` internally does a `toHex`)?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -62,15 +61,15 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
...
+  def getTransaction(txid: ByteVector32): Future[Transaction] = bitcoinClient.getTransaction(txid)
 
-  def publishTransaction(tx: Transaction)(implicit ec: ExecutionContext): Future[String] = publishTransaction(Transaction.write(tx).toHex)
+  def publishTransaction(tx: Transaction)(implicit ec: ExecutionContext): Future[String] = bitcoinClient.publishTransaction(tx)
 
-  def publishTransaction(hex: String)(implicit ec: ExecutionContext): Future[String] = rpcClient.invoke("sendrawtransaction", hex) collect { case JString(txid) => txid }
+  def publishTransaction(hex: String)(implicit ec: ExecutionContext): Future[String] = publishTransaction(Transaction.read(hex))
 
-  def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = rpcClient.invoke("lockunspent", true, outPoints.toList.map(outPoint => Utxo(outPoint.txid.toString, outPoint.index))) collect { case JBool(result) => result }
+  def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = rpcClient.invoke("lockunspent", true, outPoints.toList.map(outPoint => Utxo(outPoint.txid, outPoint.index))) collect { case JBool(result) => result }
```
t-bast: It's a bit hard to understand how functionality is divided between `BitcoinCoreWallet` and `ExtendedBitcoinClient`.

The way I understand it, `ExtendedBitcoinClient` is just a thin typed wrapper around bitcoin-cli RPC, is that correct (with the exception of the kinda weird channel announcement validation)?
And `BitcoinCoreWallet` adds some lightning-specific functionality such as composing multiple calls to the `ExtendedBitcoinClient` to do things like creating the funding tx.

If that's indeed how it was meant to be split, it feels to me that `unlockOutpoints`, `getBalance` and `getNewAddress` should be provided by `ExtendedBitcoinClient`. I'm wondering if we shouldn't take this opportunity to provide cleaner interfaces, but I don't know that part of the codebase well so maybe it's not worth it?

araspitzu: That's correct, `ExtendedBitcoinClient` is a thin typed wrapper around `BitcoinJsonRPCClient` that exposes common functionalities used by other parts of the codebase, for example to get the tx confirmations we don't directly use the RPC client (and process the response) but we'd  reuse the code in `ExtendedBitcoinClient`. `BitcoinCoreWallet` is an implementation of `EclairWallet` interface with the bitcon-core backend, it naturally uses the `ExtendedBitcoinClient` for most operations ([example](https://github.com/ACINQ/eclair/pull/1248/files#diff-f0e4b1153995aaf8ef920c12f6c4c36bR64)) but it also contains lightning-specific logic for making the funding transaction etc. Having considered this, the `unlockOutpoints` seems a good fit for `BitcoinCoreWallet` because it's used only there and is specific to the bitcoin-core behavior, WDYT?

araspitzu:
In 380e4b3 i addressed some of your comments and i cleaned up `BitcoinCoreWallet` of some unused functions that were just an alias for `ExtendedBitcoinClient`.

Issue #1171, opened 2019-10-13
Should eclair exit on SQL Disk errors?


n1bor:
Got a disk write error. But eclair just carried on. I think it would be more useful to exit. As this is a dangerous state to be in - and ensures any monitoring picks up the issue.

Agree?

 org.sqlite.SQLiteException: [SQLITE_IOERR]  Some kind of disk I/O error occurred (disk I/O error)
        at org.sqlite.core.DB.newSQLException(DB.java:941)
        at org.sqlite.core.DB.newSQLException(DB.java:953)
        at org.sqlite.core.DB.execute(DB.java:857)
        at org.sqlite.core.DB.executeUpdate(DB.java:895)
        at org.sqlite.jdbc3.JDBC3PreparedStatement.executeUpdate(JDBC3PreparedStatement.java:102)
        at fr.acinq.eclair.db.sqlite.SqliteChannelsDb$$anonfun$removeChannel$1.apply(SqliteChannelsDb.scala:75)
        at fr.acinq.eclair.db.sqlite.SqliteChannelsDb$$anonfun$removeChannel$1.apply(SqliteChannelsDb.scala:73)
        at fr.acinq.eclair.db.sqlite.SqliteUtils$.using(SqliteUtils.scala:38)
        at fr.acinq.eclair.db.sqlite.SqliteChannelsDb.removeChannel(SqliteChannelsDb.scala:73)
        at fr.acinq.eclair.channel.Channel$$anonfun$19.applyOrElse(Channel.scala:1381)
        at fr.acinq.eclair.channel.Channel$$anonfun$19.applyOrElse(Channel.scala:1376)
        at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)

pm47:
Can you provide more logs right before and after this error?

n1bor:
Nothing after the message..
The below was before.

And repeated a LOT before I realised there was an issue.
File is on EFS on AWS and a reboot fixed it.

Although my node is now in a mess - as my eclair balance is much higher than it should be!
So need to trawl through the channels to see what is going on.

Other odd thing was after the restart I had over 1m requests in the queue to bitcoind which seemed odd.


[...]

PR #1249, opened 2019-12-10
Postgresql support

rorp:
The main goal of this PR is to improve fault tolerance of `eclair` nodes in enterprise settings, as well as to "standardize" backup procedures. As a bonus point it could simplify scaling out `eclair` nodes in the future.


pm47:
One of the main complications with using a remote db backend with eclair is how you prevent two instances of eclair running at the same time and using the same db, which would be catastrophic. Doable, but not trivial. There has to be a locking mechanism somewhere.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -94,30 +94,36 @@ class Setup(datadir: File,
 
   val database = db match {
     case Some(d) => d
-    case None => Databases.sqliteJDBC(chaindir)
+    case None =>
+      val dbConfig = config.getConfig("db")
+      dbConfig.getString("driver") match {
```
egp: perhaps add `.trim` and/or `.toLowerCase` before the match

rorp:
> There has to be a locking mechanism somewhere.
Exactly! That's why is just a POC.  We are in fact discussing the locking mechanism atm. Unfortunately, it's not that easy to implement it for PostgresSQL, comparing to the sqlite implementation. But we have some ideas, which I'm going to implement soon.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlUtils.scala
```diff
@@ -0,0 +1,141 @@
...
+  private def initializeLocksTable(implicit connection: Connection): Unit = {
+    using(connection.createStatement()) { statement =>
+      statement.executeUpdate(s"CREATE TABLE $LockTable (expires_at TIMESTAMP NOT NULL UNIQUE, instance VARCHAR NOT NULL UNIQUE)")
+    }
+  }
+
+  private def acquireExclusiveTableLock(lockTimeout: FiniteDuration)(implicit connection: Connection): Unit = {
+    using(connection.createStatement()) { statement =>
+      statement.executeUpdate(s"SET lock_timeout TO '${lockTimeout.toSeconds}s'")
+      statement.executeUpdate(s"LOCK TABLE $LockTable IN ACCESS EXCLUSIVE MODE")
```
Christewart: Documentation: https://www.postgresql.org/docs/11/sql-lock.html

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlUtils.scala
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+
+  private def currentLease(implicit connection: Connection): LockLease = {
```
Christewart: `Option[LockLease]` or maybe create an ADT ?

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlUtils.scala
```diff
@@ -0,0 +1,141 @@
...
+          logger.warn(s"Table ${LockTable} does not exist, trying to recreate it...")
+          initializeLocksTable(connection)
+          obtainExclusiveLock(instanceId, leaseDuration, lockTimeout, attempt + 1)
+        }
+    }
+  }
+
+  private def initializeLocksTable(implicit connection: Connection): Unit = {
+    using(connection.createStatement()) { statement =>
+      statement.executeUpdate(s"CREATE TABLE $LockTable (expires_at TIMESTAMP NOT NULL UNIQUE, instance VARCHAR NOT NULL UNIQUE)")
```
Christewart: Is there any way we can create an invaraint at the database level to make sure the `$LockTable` can only contain one row? https://stackoverflow.com/questions/25307244/how-to-allow-only-one-row-for-a-table

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlUtils.scala
rorp: It make the logic more complicated.

n1bor:
Few thoughts on this:

1. Not sure I agree with @pm47 concern about 2 instances connecting. Only going to use postgres in a server type environment with firewall rules, numerous controls, start/stop scripts etc... which would check. But added safety at the DB level would be good.
2. Brings up what to do if database goes down. My thinking is it should drop all peer connections and go into a watch only mode. In effect a watchtower for itself. So just listen to bitcoind for any tx's for past commitments - and publish commitments if needed. Raised this before as had a corruption of sqlite that caused a total mess as eclair carried on running but could not update the channels db. https://github.com/ACINQ/eclair/issues/1171
3. Thing should have option to run with non-critical tables remaining in sqlite - i.e. those you can recreate from gossip messages. As think most would run like this to reduce load on backend db.
4. Need a way to run all the tests against postgress. Even if only run manually on occasion - and to validate server install/version is OK.

But in general would be great. Am running a node on AWS and so with this could have a synchronous postgres replica in another data centre. So even if there was total loss of an AWS data centre I could (theoretically assume everyone else is not doing the same ;)) restart the node in perfect condition in a few mins - and not lose any channels or money.

And would mean we could move away from the "dump the whole database every 15mins" approach which is not very scale-able over time.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlChannelsDb.scala
```diff
@@ -0,0 +1,123 @@
...
+  val DB_NAME = "channels"
+  val CURRENT_VERSION = 2
+
+  withConnection { psql =>
+    using(psql.createStatement(), inTransaction = true) { statement =>
+      getVersion(statement, DB_NAME, CURRENT_VERSION) match {
+        case CURRENT_VERSION =>
+          statement.executeUpdate("CREATE TABLE IF NOT EXISTS local_channels (channel_id TEXT NOT NULL PRIMARY KEY, data BYTEA NOT NULL, is_closed BOOLEAN NOT NULL DEFAULT FALSE)")
+          statement.executeUpdate("CREATE TABLE IF NOT EXISTS htlc_infos (channel_id TEXT NOT NULL, commitment_number TEXT NOT NULL, payment_hash TEXT NOT NULL, cltv_expiry BIGINT NOT NULL, FOREIGN KEY(channel_id) REFERENCES local_channels(channel_id))")
+          statement.executeUpdate("CREATE INDEX IF NOT EXISTS htlc_infos_idx ON htlc_infos(channel_id, commitment_number)")
```
n1bor: Index on is_closed? As used for listLocalChannels.

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -170,7 +170,8 @@ class ApiServiceSpec extends FunSuite with ScalatestRouteTest with IdiomaticMock
       alias = "alice",
       chainHash = ByteVector32(hex"06226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf188910f"),
       blockHeight = 9999,
-      publicAddresses = NodeAddress.fromParts("localhost", 9731).get :: Nil
+      publicAddresses = NodeAddress.fromParts("localhost", 9731).get :: Nil,
+      instanceId = "instance"
```
n1bor: Could instance ID default to hash of the seed?

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -56,9 +59,59 @@ object TestConstants {
...
+    override def obtainLock(): Unit = ()
+  }
+
+  case object TestPsqlDatabases extends TestDatabases {
+    override val connection: Connection = psql()
+
+    import com.zaxxer.hikari.{HikariConfig, HikariDataSource}
+
+    val config = new HikariConfig
+    config.setJdbcUrl("jdbc:postgresql://localhost:5432/eclair")
```
n1bor: This another option?
https://dzone.com/articles/using-embedded-postgresql-databases-for-unit-testi

rorp:
> 1. Not sure I agree with @pm47 concern about 2 instances connecting. Only going to use postgres in a server type environment with firewall rules, numerous controls, start/stop scripts etc... which would check. But added safety at the DB level would be good.

Even though there are a bunch of security settings outside Eclair, there is always a way to shoot in the foot.

> 2. Brings up what to do if database goes down. My thinking is it should drop all peer connections and go into a watch only mode. In effect a watchtower for itself. So just listen to bitcoind for any tx's for past commitments - and publish commitments if needed. Raised this before as had a corruption of sqlite that caused a total mess as eclair carried on running but could not update the channels db. #1171

I think that's a generic problem unrelated to the database type. It should be addressed in a separate PR.

> 3. Thing should have option to run with non-critical tables remaining in sqlite - i.e. those you can recreate from gossip messages. As think most would run like this to reduce load on backend db.

We are going to do some performance testing to see how slow Postgres is comparing to Sqlite.

> 4. Need a way to run all the tests against postgress. Even if only run manually on occasion - and to validate server install/version is OK.

It's already there. Probably the naming of the tests were misleading, so I renamed them.

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
rorp: It's irrelevant for this test case. It checks if `getinfo` API call returns all fields with expected values.

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
rorp: Cool! I used `EmbeddedPG` for DB tests, so now there's no need to install actual Postgres to run tests.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -353,12 +352,47 @@ class Setup(datadir: File,
...
+      case Some(d) => d
+      case None =>
+        dbConfig.getString("driver") match {
+          case "sqlite" => Databases.sqliteJDBC(chaindir)
+          case "psql" =>
+            val psql = Databases.setupPsqlDatabases(dbConfig, { ex =>
+              logger.error("fatal error: Cannot obtain lock on the database.\n", ex)
+              sys.exit(-2)
+            })
+            if (LockType(dbConfig.getString("psql.lock-type")) == LockType.EXCLUSIVE) {
```
Christewart: Is `optimistic` locking not implemented?

rorp: It is.

eclair-core/src/test/scala/fr/acinq/eclair/db/SqliteAuditDbSpec.scala
```diff
@@ -35,362 +38,472 @@ import scala.util.Random
...
-      statement.executeUpdate("CREATE INDEX IF NOT EXISTS received_timestamp_idx ON received(timestamp)")
-      statement.executeUpdate("CREATE INDEX IF NOT EXISTS relayed_timestamp_idx ON relayed(timestamp)")
-      statement.executeUpdate("CREATE INDEX IF NOT EXISTS network_fees_timestamp_idx ON network_fees(timestamp)")
-      statement.executeUpdate("CREATE INDEX IF NOT EXISTS channel_events_timestamp_idx ON channel_events(timestamp)")
-      statement.executeUpdate("CREATE INDEX IF NOT EXISTS channel_errors_timestamp_idx ON channel_errors(timestamp)")
-    }
-
-    using(connection.createStatement()) { statement =>
-      assert(getVersion(statement, "audit", 4) == 3) // version 3 is deployed now
+    forAllDbs {
```
t-bast, line 286: Migration 3->4 doesn't need to be tested for psql once we remove that migration from `PsqlAuditDb` :wink:

eclair-core/src/test/scala/fr/acinq/eclair/db/SqlitePaymentsDbSpec.scala
```diff
@@ -37,383 +38,496 @@ class SqlitePaymentsDbSpec extends FunSuite {
...
-    assert(preMigrationDb.getOutgoingPayment(id1) === Some(ps1))
-    assert(preMigrationDb.listOutgoingPayments(parentId) === Seq(ps2, ps3))
-
-    val postMigrationDb = new SqlitePaymentsDb(connection)
-
-    using(connection.createStatement()) { statement =>
-      assert(getVersion(statement, "payments", 4) == 4) // version still to 4
+    forAllDbs {
+      case dbs: TestPsqlDatabases =>
+        import fr.acinq.eclair.db.psql.PsqlUtils._
```
t-bast, line 234: Same as audit DB, no need to test psql migration once we remove it

pom.xml
```diff
@@ -27,6 +27,7 @@
         <module>eclair-core</module>
         <module>eclair-node</module>
         <module>eclair-node-gui</module>
+        <module>sqlite2psql</module>
```
t-bast, line 30: I don't know if we want to have this tool in the eclair repository...

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteChannelsDb.scala
```diff
@@ -124,6 +124,18 @@ class SqliteChannelsDb(sqlite: Connection) extends ChannelsDb with Logging {
     }
   }
 
+  def listHtlcInfos(channelId: ByteVector32): Seq[(ByteVector32, CltvExpiry, Long)] = {
```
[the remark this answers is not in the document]
rorp: It's actually used in `SqliteToPgsql` along with other list/add methods in the other database. The reason why I created them is that the database classes are not simply CRUD, they also do some pre- and post-processing, and some of the add methods don't insert full information (e.g. `OutgoingPaymentsDb.addOutgoingPayment()`

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -431,6 +431,70 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
     }
   }
 
+  def listIncomingPayments(): Seq[(ByteVector32,ByteVector32,String,PaymentRequest,Option[MilliSatoshi],Long,Long,Option[Long])] = {
```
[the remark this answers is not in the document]
rorp: See the previous comment.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -69,23 +81,131 @@ object Databases extends Logging {
...
+    Databases.postgresJDBC(
+      database = database, host = host, port = port,
+      username = username, password = password,
+      poolProperties = properties,
+      databaseLeaseInterval = leaseInterval, lockTimeout = lockTimeout,
+      lockExceptionHandler = lockExceptionHandler, lockType = lockType, datadir = datadir
+    )
+  }
+
+  private def checkIfDatabaseUrlIsUnchanged(url: String, datadir: File ): Unit = {
```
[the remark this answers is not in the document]
rorp: So this is a safeguard agains pointing the node to a wrong database and losing the funds. In case of SQLite the seed and the database are stored in the same place so it's not an issue.

So this method prevents node from starting if the database url has been changed accidentally. If the node operator can remove `last_jdbcurl` file to force using the new database.

Probably it's not the best solution. As an alternative we can store node ID inside Postgres database and prevent the node from staring if it's public key does not match the node id in the database.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlAuditDb.scala
```diff
@@ -0,0 +1,337 @@
...
+          statement.setString(3, p.channelId.toHex)
+          statement.setString(4, p.direction)
+          statement.setString(5, p.relayType)
+          statement.setLong(6, e.timestamp)
+          statement.executeUpdate()
+        }
+      }
+    }
+
+  override def add(e: NetworkFeePaid): Unit = add(e, e.tx.txid)
```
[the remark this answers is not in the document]
rorp: That's another method used in `SqliteToPgsql`.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlPaymentsDb.scala
```diff
@@ -0,0 +1,475 @@
...
+        statement.setLong(7, sent.recipientAmount.toLong)
+        statement.setString(8, sent.recipientNodeId.value.toHex)
+        statement.setLong(9, sent.createdAt)
+        statement.setString(10, sent.paymentRequest.map(PaymentRequest.write).orNull)
+        statement.executeUpdate()
+      }
+    }
+  }
+
+  def addOutgoingPayment(sent: ((UUID, UUID,Option[String],ByteVector32,MilliSatoshi,PublicKey,Long,Option[PaymentRequest],Option[Long],Option[ByteVector32],Option[MilliSatoshi],Option[BitVector],Option[BitVector],String,MilliSatoshi))): Unit = {
```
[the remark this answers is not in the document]
rorp: Again, an utility method for `SqliteToPgsql`

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlPaymentsDb.scala
```diff
@@ -0,0 +1,475 @@
...
+        statement.setString(2, preimage.toHex)
+        statement.setString(3, paymentType)
+        statement.setString(4, PaymentRequest.write(pr))
+        statement.setLong(5, pr.timestamp.seconds.toMillis) // BOLT11 timestamp is in seconds
+        statement.setLong(6, (pr.timestamp + pr.expiry.getOrElse(PaymentRequest.DEFAULT_EXPIRY_SECONDS.toLong)).seconds.toMillis)
+        statement.executeUpdate()
+      }
+    }
+
+  def addIncomingPayment(payment: (ByteVector32,ByteVector32,String,PaymentRequest,Option[MilliSatoshi],Long,Long,Option[Long])): Unit = {
```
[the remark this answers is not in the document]
rorp: Ditto

pom.xml
rorp: We definitely need a SQLite -> PostgreSQL migration tool, otherwise there's no way to switch to Postgres for existing nodes.

I don't really like to have it as a separate project either. I'm open to suggestions how to package  it in a better way.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlUtils.scala
```diff
@@ -0,0 +1,278 @@
...
+    override def withLock[T](f: Connection => T)(implicit ds: DataSource): T = {
+      inTransaction { connection =>
+        val res = f(connection)
+        checkDatabaseOwnership(connection, instanceId, lockTimeout, lockExceptionHandler)
+        res
+      }
+    }
+  }
+
+  case class OptimisticLock(dataVersion: AtomicLong, lockExceptionHandler: LockExceptionHandler) extends DatabaseLock {
```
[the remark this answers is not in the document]
rorp: I removed it altogether

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -69,23 +81,131 @@ object Databases extends Logging {
...
+                   instanceId: String = ManagementFactory.getRuntimeMXBean().getName(),
+                   databaseLeaseInterval: FiniteDuration = 5.minutes,
+                   lockTimeout: FiniteDuration = 5.seconds,
+                   lockExceptionHandler: LockExceptionHandler = { _ => () },
+                   lockType: LockType = LockType.NONE, datadir: File = new File(File.separator + "tmp")): Databases = {
+    val url = s"jdbc:postgresql://${host}:${port}/${database}"
+
+    checkIfDatabaseUrlIsUnchanged(url, datadir)
+
+    implicit val lock: DatabaseLock = lockType match {
```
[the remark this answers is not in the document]
rorp: 'm not a big fan of using implicits either, but in this case they saved me a lot of typing.

n1bor:
Could migration be via a dump of the Channels rather than via the DB?
So you would have to do a clean shutdown with no pending HTLC's then dump / load the channels themselves as JSON?

t-bast:
> So you would have to do a clean shutdown with no pending HTLC's then dump / load the channels themselves as JSON?

What do you mean by "as JSON"? The issue is to correctly insert each value in the right column, respecting the DB type (types are a bit different between sqlite and psql). I don't know if something like that would work out of the box.

In this vein, the sqlite2psql tool could deal directly with SQL (instead of Scala classes) and directly list rows from sqlite and insert them into psql, but I don't know what type errors may happen (worth a try?).

n1bor:
I mean output using "eclair-cli channels" and de-serialise that into the Scala Classes then save to psql using the existing code.

t-bast:
> I mean output using "eclair-cli channels" and de-serialise that into the Scala Classes then save to psql using the existing code.

That can't be enough, it would be lacking a lot of internal information necessary for the DB to operate correctly.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlUtils.scala
```diff
@@ -0,0 +1,232 @@
...
+import scala.concurrent.duration.FiniteDuration
+import scala.util.{Failure, Success, Try}
+
+object PsqlUtils extends JdbcUtils with Logging {
+
+  val DataVersionTable = "data_version"
+
+  val OwnershipTable = "lease"
+
+  val lockTimeout = FiniteDuration(5, "s")
```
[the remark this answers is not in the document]
rorp: I think it's more than enough. The lease table will be locked for the duration of one SELECT and one INSERT/UPDATE statements. If they takes more than 5 seconds, there's definitely something wrong, so we should throw.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlNetworkDb.scala
```diff
@@ -0,0 +1,183 @@
...
+  val CURRENT_VERSION = 2
+
+  inTransaction { psql =>
+    using(psql.createStatement()) { statement =>
+      getVersion(statement, DB_NAME, CURRENT_VERSION) match {
+        case CURRENT_VERSION => () // nothing to do
+        case unknown => throw new IllegalArgumentException(s"unknown version $unknown for network db")
+      }
+      statement.executeUpdate("CREATE TABLE IF NOT EXISTS nodes (node_id TEXT NOT NULL PRIMARY KEY, data BYTEA NOT NULL)")
+      statement.executeUpdate("CREATE TABLE IF NOT EXISTS channels (short_channel_id BIGINT NOT NULL PRIMARY KEY, txid TEXT NOT NULL, channel_announcement BYTEA NOT NULL, capacity_sat BIGINT NOT NULL, channel_update_1 BYTEA NULL, channel_update_2 BYTEA NULL)")
```
araspitzu, line 45: Can `capacity_sat` be an `INTEGER` instead of `BIGINT`?

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlChannelsDb.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  val DB_NAME = "channels"
+  val CURRENT_VERSION = 2
+
+  inTransaction { psql =>
+    using(psql.createStatement()) { statement =>
+      getVersion(statement, DB_NAME, CURRENT_VERSION) match {
+        case CURRENT_VERSION =>
+          statement.executeUpdate("CREATE TABLE IF NOT EXISTS local_channels (channel_id TEXT NOT NULL PRIMARY KEY, data BYTEA NOT NULL, is_closed BOOLEAN NOT NULL DEFAULT FALSE)")
+          statement.executeUpdate("CREATE TABLE IF NOT EXISTS htlc_infos (channel_id TEXT NOT NULL, commitment_number TEXT NOT NULL, payment_hash TEXT NOT NULL, cltv_expiry BIGINT NOT NULL, FOREIGN KEY(channel_id) REFERENCES local_channels(channel_id))")
```
araspitzu, line 44: Do we really need to user `BIGINT` data type for `cltv_expiry`?

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -69,23 +82,122 @@ object Databases extends Logging {
...
     override val payments = new SqlitePaymentsDb(eclairJdbc)
     override val pendingRelay = new SqlitePendingRelayDb(eclairJdbc)
+    override def backup(backupFile: File): Unit = {
+      val tmpFile = new File(backupFile.getAbsolutePath.concat(".tmp"))
 
-    override def backup(file: File): Unit = {
       SqliteUtils.using(eclairJdbc.createStatement()) {
-        statement => {
-          statement.executeUpdate(s"backup to ${file.getAbsolutePath}")
+        statement =>  {
```
araspitzu, line 141: nit: whitespace

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -69,23 +82,122 @@ object Databases extends Logging {
...
     }
+  }
+
+  def postgresJDBC(database: String, host: String, port: Int,
+                   username: Option[String], password: Option[String],
+                   poolProperties: Map[String, Long],
+                   instanceId: String,
+                   databaseLeaseInterval: FiniteDuration,
+                   lockExceptionHandler: LockExceptionHandler = { _ => () },
+                   lockType: LockType = LockType.NONE, datadir: File = new File(File.separator + "tmp")): Databases = {
```
araspitzu, line 93: Can we remove the default for `datadir: File`? It's not used and could lead to potential problems if the `datadir` parameter is omitted by mistake.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -360,6 +367,38 @@ class Setup(datadir: File,
...
+            val psql = Databases.setupPsqlDatabases(dbConfig, instanceId, datadir, { ex =>
+              logger.error("fatal error: Cannot obtain lock on the database.\n", ex)
+              sys.exit(-2)
+            })
+            if (LockType(dbConfig.getString("psql.lock-type")) == LockType.OWNERSHIP_LEASE) {
+              val dbLockLeaseRenewInterval = dbConfig.getDuration("psql.ownership-lease.lease-renew-interval").toSeconds.seconds
+              val dbLockLeaseInterval = dbConfig.getDuration("psql.ownership-lease.lease-interval").toSeconds.seconds
+              if (dbLockLeaseInterval <= dbLockLeaseRenewInterval)
+                throw new RuntimeException("Invalid configuration: `db.psql.ownership-lease.lease-interval` must be greater than `db.psql.ownership-lease.lease-renew-interval`")
+              system.scheduler.schedule(dbLockLeaseRenewInterval, dbLockLeaseRenewInterval) {
```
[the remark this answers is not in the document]
t-bast: I think we should keep this one though, we kept it in other places of the code to make sure this also keeps working on akka 2.4 for eclair-mobile for now. However `Platform.currentTime` can definitely be replaced by `System.currentTimeMillis` which is also available on android.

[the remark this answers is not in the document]
araspitzu: okay, I thought we're going to remove the pgsql code when merging into the `android` branch, I'm fine leaving this here then.

n1bor:
I am a bit confused about instanceId. It is a random UUID that changes each time the server starts up.
Why would we not set it to hex(sha256(seed))? And store that in the DB, and checking it was correct on startup?
As that would protect against starting up with the wrong seed and connect to DB as well as protecting against multiple connections.

araspitzu:
> I am a bit confused about instanceId. It is a random UUID that changes each time the server starts up.
> Why would we not set it to hex(sha256(seed))? And store that in the DB, and checking it was correct on startup?
> As that would protect against starting up with the wrong seed and connect to DB as well as protecting against multiple connections.

It's fine to change the `instanceId` every time we start up, its purpose is to make sure that one and only one instance is connected to the database at all times. Check out [how db ownership is obtained](https://github.com/ACINQ/eclair/pull/1249/files#diff-aa65f6f5a27e62ab938f215a244237a4R134), there is a special table named `lease` in which we write the current instance id connected to the db, if you try to run another instance on the same db it will fail because it won't acquire the lock (different `instanceId`). The `lease` table also has a timeout for the lock in case one instance goes down you will be able to re-obtain the lock after the lease expired.

n1bor:
> It's fine to change the `instanceId` every time we start up, its purpose is to make sure that one and only one instance is connected to the database at all times. Check out [how db ownership is obtained](https://github.com/ACINQ/eclair/pull/1249/files#diff-aa65f6f5a27e62ab938f215a244237a4R134), there is a special table named `lease` in which we write the current instance id connected to the db, if you try to run another instance on the same db it will fail because it won't acquire the lock (different `instanceId`). The `lease` table also has a timeout for the lock in case one instance goes down you will be able to re-obtain the lock after the lease expired.

I get that. But that does not protect against connecting the wrong seed to the database if the other instance is also down. It should not be possible to connect to a database with the wrong seed as that would also create a mess. Maybe should be a seperate PR?

rorp:
> During testing I've encountered a small issue though: the first time i launched eclair with `psql` driver it couldn't find the `eclair` database in my postgre instance so i had to create it manually with`createdb eclair`, i think eclair should create the 'eclair' db if none exists.

I did it for purpose. This is why:

1. If you look at https://www.postgresql.org/docs/current/sql-createdatabase.html you can see that there are some `CREATE DATABASE` options that can't have reasonable default values (like `TABLESPACE`). It's for DBAs to decide which options they need to set.
2. From security perspective it's important for the Eclair database user to have the least amount of permissions. I wouldn't allow it to create databases. Also, I would split the single Eclair user into two separate users with different sets of permissions: one for migrations, the other for regular database operations.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlNetworkDb.scala
rorp: Scala's `Long` doen't fit Postgres' `INTEGER`.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlChannelsDb.scala
rorp: Ditto.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlUtils.scala
```diff
@@ -0,0 +1,232 @@
...
+
+  private def acquireExclusiveTableLock()(implicit connection: Connection): Unit = {
+    using(connection.createStatement()) {
+      statement =>
+        statement.executeUpdate(s"SET lock_timeout TO '${lockTimeout.toSeconds}s'")
+        statement.executeUpdate(s"LOCK TABLE $OwnershipTable IN ACCESS EXCLUSIVE MODE")
+    }
+  }
+
+  private def checkDatabaseOwnership(connection: Connection, instanceId: String, lockExceptionHandler: LockExceptionHandler): Unit = {
```
pm47, line 180: Can you add a comment somewhere explaining how the locking works. It isn't immediately obvious. Also, is this a standard technique, or completely tailored made to our use case?

eclair-core/src/main/resources/reference.conf
```diff
@@ -166,6 +166,28 @@ eclair {
...
+  db {
+    driver = "sqlite" // sqlite, psql
+    psql {
+      database = "eclair"
+      host = "localhost"
+      port = 5432
+      username = ""
+      password = ""
+      pool {
+        max-size = 10 // recommended value = number_of_cpu_cores * 2
```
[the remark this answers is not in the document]
rorp: Yes, it can. But `10` is a good default value for most of the cases.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlUtils.scala
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  case class OwnershipLeaseLock(instanceId: String, leaseDuration: FiniteDuration, lockExceptionHandler: LockExceptionHandler) extends DatabaseLock {
+    override def obtainExclusiveLock(implicit ds: DataSource): Unit =
+      obtainDatabaseOwnership(instanceId, leaseDuration)
+
+    override def withLock[T](f: Connection => T)(implicit ds: DataSource): T = {
+      inTransaction { connection =>
+        val res = f(connection)
+        checkDatabaseOwnership(connection, instanceId, lockExceptionHandler)
```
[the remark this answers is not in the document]
rorp: So that it doesn't need to hold a lock on `lease` table for the duration `f`.

eclair-core/src/main/scala/fr/acinq/eclair/db/psql/PsqlUtils.scala
rorp: It's a pretty much standard concept, but the implementation is tailored for this particular use case.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -69,23 +81,123 @@ object Databases extends Logging {
...
+                   lockExceptionHandler: LockExceptionHandler = { _ => () },
+                   lockType: LockType = LockType.NONE, datadir: File): Databases = {
+    val url = s"jdbc:postgresql://${host}:${port}/${database}"
+
+    checkIfDatabaseUrlIsUnchanged(url, datadir)
+
+    implicit val lock: DatabaseLock = lockType match {
+      case LockType.NONE => NoLock
+      case LockType.LEASE => LeaseLock(instanceId, databaseLeaseInterval, lockExceptionHandler)
+      case x@_ => throw new RuntimeException(s"Unknown postgres lock type: `$lockType`")
```
pm47, line 100:
```suggestion
      case _ => throw new RuntimeException(s"Unknown postgres lock type: `$lockType`")
```

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -83,7 +84,8 @@ case class NodeParams(keyManager: KeyManager,
                       routerConf: RouterConf,
                       socksProxy_opt: Option[Socks5ProxyParams],
                       maxPaymentAttempts: Int,
-                      enableTrampolinePayment: Boolean) {
+                      enableTrampolinePayment: Boolean,
+                      instanceId: UUID) {
```
pm47, line 88: I'd put this above in the arg list, maybe right after the `keyManager`. It also deserves a description.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -363,6 +370,38 @@ class Setup(datadir: File,
       None
     }
   }
+
+  private def initDatabase(dbConfig: Config, instanceId: UUID): Databases = {
```
pm47, line 374: This method probably belongs to Databases.scala.

eclair-core/src/test/scala/fr/acinq/eclair/db/BackupHandlerSpec.scala
```diff
@@ -23,6 +23,7 @@ import java.util.UUID
 import akka.actor.ActorSystem
 import akka.testkit.{TestKit, TestProbe}
 import fr.acinq.eclair.channel.ChannelPersisted
+import fr.acinq.eclair.db.Databases.CanBackup
```
pm47, line 26: Nit: there are unused imports in this file.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -37,11 +45,15 @@ trait Databases {
 
   val pendingRelay: PendingRelayDb
 
-  def backup(file: File): Unit
+  def obtainExclusiveLock(): Unit
```
pm47, line 48: The choice to make `obtainExclusiveLock` a method that can be implemented as a no-op, whereas `backup` is a separate trait is not very consistent. Also, changing `CanBackup`->`FileBackup` makes it clear that this is only one type of backup, there may be others. How about:

```scala
trait Databases {
  val network: NetworkDb
  val audit: AuditDb
  val channels: ChannelsDb
  val peers: PeersDb
  val payments: PaymentsDb
  val pendingRelay: PendingRelayDb
}

trait FileBackup { this: Databases =>
  def backup(file: File): Unit
}

case class SqliteDatabases(...) extends Databases with FileBackup {
  def backup(file: File): Unit = ???
}

case class PostgresDatabases(...) extends Databases {
  def obtainExclusiveLock(): Unit = ???
}
```

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -225,7 +284,8 @@ object TestConstants {
       ),
       socksProxy_opt = None,
       maxPaymentAttempts = 5,
-      enableTrampolinePayment = true
+      enableTrampolinePayment = true,
+      instanceId = UUID.fromString("01234567-0123-4567-89ab-0123456789ab")
```
pm47, line 288: It's a bit confusing that Alice and Bob share the same `instanceId`.

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -57,9 +62,62 @@ object TestConstants {
...
+    val connection: Connection
+    def network(): NetworkDb
+    def audit(): AuditDb
+    def channels(): ChannelsDb
+    def peers(): PeersDb
+    def payments(): PaymentsDb
+    def pendingRelay(): PendingRelayDb
+    def getVersion(statement: Statement, db_name: String, currentVersion: Int): Int
+    def close(): Unit
+  }
```
pm47, line 75: Why don't you use the `Databases` trait? You can mix-in the `close` method if needed.

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -46,19 +47,15 @@ import scala.util.{Failure, Success, Try}
   *
   * Constructor is private so users will have to use BackupHandler.props() which always specific a custom mailbox
   */
-class BackupHandler private(databases: Databases, backupFile: File, backupScript_opt: Option[String]) extends Actor with RequiresMessageQueue[BoundedMessageQueueSemantics] with ActorLogging {
+class BackupHandler private(databases: CanBackup, backupFile: File, backupScript_opt: Option[String]) extends Actor with RequiresMessageQueue[BoundedMessageQueueSemantics] with ActorLogging {
```
pm47, line 50: Let's make it more clear that this is only useful for one particular type of backup.:
```suggestion
class FileBackupHandler private(databases: CanBackup, backupFile: File, backupScript_opt: Option[String]) extends Actor with RequiresMessageQueue[BoundedMessageQueueSemantics] with ActorLogging {
```

eclair-core/src/main/scala/fr/acinq/eclair/db/BackupHandler.scala
```diff
@@ -46,19 +47,15 @@ import scala.util.{Failure, Success, Try}
...
 
   def receive = {
     case persisted: ChannelPersisted =>
       val start = System.currentTimeMillis()
-      val tmpFile = new File(backupFile.getAbsolutePath.concat(".tmp"))
-      databases.backup(tmpFile)
-      // this will throw an exception if it fails, which is possible if the backup file is not on the same filesystem
-      // as the temporary file
-      Files.move(tmpFile.toPath, backupFile.toPath, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.ATOMIC_MOVE)
+      databases.backup(backupFile)
```
pm47, line 58: I'd rather you revert the changes to those lines. The `databases.backup(backupFile` should only handle the dump to the given file, the rest (file replacement) is an implementation detail of the `FileBackupHandler` and would be the same if we supported, say, BerkeleyDb.

t-bast:
I've made some fixes to the `stats` DB call in https://github.com/ACINQ/eclair/pull/1470, I think you can easily replicate the changes in the postgres implementation.

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
rorp: 1.  I needed a sealed trait for exhaustive pattern matching
2. Some of the tests need a fresh connection every time `network`, `audit`, etc gets called, but they are defined as `val` in `Databases`
3. `Databases` require `obtainExclusiveLock()` to be implemented, but it's database specific and can't be tested in a generic tests.

rorp:
> I've made some fixes to the `stats` DB call in #1470, I think you can easily replicate the changes in the postgres implementation.

Done

pm47:
🎉

PR #1251, opened 2019-12-11
Allow Peer.Connect messages to force reconnection to remote.

### LINKED ISSUE #1242: Unable to force reconnection via /connect API
### Scenario
Eclair-server has a channel with a remote peer and when the peer goes offline eclair schedules some reconnection attempts. While the attempts are scheduled i'm unable to trigger a manual reconnection using the `/connect` API.

The expected behavior is to be able to manually initiate a reconnection overriding the ones scheduled by eclair.

Log output showing reconnection attempts scheduled by eclair
```
2019-12-09 12:12:31,495 INFO  f.a.e.c.TransportHandler CON n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - connection closed: PeerClosed
2019-12-09 12:12:31,499 INFO  fr.acinq.eclair.io.Peer n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - lost connection to 037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f
2019-12-09 12:12:35,821 INFO  fr.acinq.eclair.io.Peer n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - reconnecting to /127.0.0.1:9738 (next reconnection in 10 seconds)
2019-12-09 12:12:35,822 INFO  f.acinq.eclair.io.Client CON n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - connecting to 127.0.0.1:9738
2019-12-09 12:12:35,824 INFO  f.acinq.eclair.io.Client CON n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - connection failed to 127.0.0.1:9738
2019-12-09 12:12:45,842 INFO  fr.acinq.eclair.io.Peer n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - reconnecting to /127.0.0.1:9738 (next reconnection in 20 seconds)
2019-12-09 12:12:45,843 INFO  f.acinq.eclair.io.Client CON n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - connecting to 127.0.0.1:9738
2019-12-09 12:12:45,847 INFO  f.acinq.eclair.io.Client CON n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - connection failed to 127.0.0.1:9738
2019-12-09 12:13:05,863 INFO  fr.acinq.eclair.io.Peer n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - reconnecting to /127.0.0.1:9738 (next reconnection in 40 seconds)
2019-12-09 12:13:05,864 INFO  f.acinq.eclair.io.Client CON n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - connecting to 127.0.0.1:9738
2019-12-09 12:13:05,866 INFO  f.acinq.eclair.io.Client CON n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - connection failed to 127.0.0.1:9738
2019-12-09 12:13:45,882 INFO  fr.acinq.eclair.io.Peer n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - reconnecting to /127.0.0.1:9738 (next reconnection in 80 seconds)
2019-12-09 12:13:45,883 INFO  f.acinq.eclair.io.Client CON n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - connecting to 127.0.0.1:9738
2019-12-09 12:13:45,886 INFO  f.acinq.eclair.io.Client CON n:037820f011811bcb647a000120e85685dc9706f0afd5614d72086d29704400f63f - connection failed to 127.0.0.1:9738
```
When calling `/connect` the response is `"reconnection in progress"` and no manual reconnection is triggered.

araspitzu:
Currently the Peer class schedules reconnections to remote and while waiting for the reconnection trigger it ignores `Peer.Connect` messages, this PR fixes the behavior by allowing connect messages to initiate a connection even if tere is already a known address for it.

Fixes #1242

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -74,15 +74,9 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
           sender ! "no address found"
           stay
         case Some(address) =>
-          if (d.address_opt.contains(address)) {
-            // we already know this address, we'll reconnect automatically
-            sender ! "reconnection in progress"
-            stay
-          } else {
-            // we immediately process explicit connection requests to new addresses
-            context.actorOf(Client.props(nodeParams, authenticator, address, remoteNodeId, origin_opt = Some(sender())))
-            stay using d.copy(address_opt = Some(address))
-          }
+          // we immediately process explicit connection requests to new addresses
+          context.actorOf(Client.props(nodeParams, authenticator, address, remoteNodeId, origin_opt = Some(sender())))
+          stay using d.copy(address_opt = Some(address))
       }
 
     case Event(Reconnect, d: DisconnectedData) =>
--- a/eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
@@ -133,13 +133,21 @@ class PeerSpec extends TestkitBaseClass with StateTestsHelperMethods {
     awaitCond(peer.stateData.address_opt == Some(fakeIPAddress.socketAddress))
   }
 
-  test("ignore connect to same address") { f =>
+  test("connect forces reconnection even if there is already an address") { f =>
     import f._
     val probe = TestProbe()
-    val previouslyKnownAddress = new InetSocketAddress("1.2.3.4", 9735)
+    val mockServer = new ServerSocket(0, 1, InetAddress.getLocalHost) // port will be assigned automatically
+    val previouslyKnownAddress = new InetSocketAddress(mockServer.getInetAddress.getHostAddress, mockServer.getLocalPort)
+    awaitCond(peer.stateName == INSTANTIATING)
     probe.send(peer, Peer.Init(Some(previouslyKnownAddress), Set.empty))
-    probe.send(peer, Peer.Connect(NodeURI.parse("03933884aaf1d6b108397e5efe5c86bcf2d8ca8d2f700eda99db9214fc2712b134@1.2.3.4:9735")))
-    probe.expectMsg("reconnection in progress")
+    awaitCond(peer.stateName == DISCONNECTED)
+    probe.send(peer, Peer.Connect(NodeURI.parse(s"${randomKey.publicKey}@${previouslyKnownAddress.getHostName}:${previouslyKnownAddress.getPort}")))
+
+    // assert the peer established a connection with the mock server and moved to CONNECTED state
+    within(30 seconds) {
+      mockServer.accept()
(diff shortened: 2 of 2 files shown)


t-bast:
I'm not sure we want to do that, it seems to me that we should let the reconnection logic do its magic (and if there are issues, improve the reconnection logic).
Otherwise we're effectively losing the back-off mechanisms and risk being black-listed by our peers, don't we?

araspitzu:
The automatic reconnection logic is not affected by this PR. Assuming there is a scheduled reconnection the Peer class follows this behavior:
- receive `Peer.Connect` -> try to connect -> if successful abort scheduled reconnection
- receive `Peer.Connect` -> try to connect -> if unsuccessful do nothing

t-bast:
I think I'm missing something. The previous code prevents us from manually re-triggering a connection attempt because there's already one underway (with an exponential back-off mechanism).
What you're doing is removing that code, which means we can have multiple connection attempts at the same time (though the spec doesn't allow you to have multiple connections).
That effectively bypasses the exponential back-off mechanism, right?

araspitzu:
The reconnections are handled by `Reconnect` message and in the `Peer` class we use the akka scheduler to send it to ourself thus triggering reconnection attempts, manual connections instead are handled by the `Connect` message. When we're in state `DISCONNECTED` two different blocks of code are used to handle each case https://github.com/ACINQ/eclair/pull/1251/files#diff-51851274f6488b212bc8bfa2fa89797aR68-R82.

t-bast:
Ok, I don't know that part of the code enough, I need to dive into it more.
I'll let @pm47 review this to unblock.

pm47:
Current code prevents initiating reconnection to the exact same address we are already using. We will reconnect at most within an hour anyway, and that's if the remote peer has been down for a very long time.

The only use case I can think of for allowing to force a connection retry using the same address is if the user actually controls the remote peer and knows it is now up *and* the remote peer doesn't itself initiates a reconnection to the local node. That's really a corner case I think?

araspitzu:
True it's a bit of a corner case but it's also harmless to let the user make a reconnection attempt, in fact we already allow that _except_ if you're reconnecting to the same address. I still prefer to allow the user to make an attempt but if this change is too controversial i think we should document it thoroughly.

pm47:
It's not harmless actually. This restriction was added in #835 because it caused issues in eclair-mobile which was sending connection requests independently when we got offline (like an external reconnection mechanism).

araspitzu:
Okay then, since it's not harmless and fixes only a corner case i'm closing this.

PR #1252, opened 2019-12-12
Add contributing instructions

t-bast:
Add instructions for contributing and reporting security issues.
Greatly inspired by Bitcoin and LND.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,6 +142,8 @@ If you're using [IntelliJ](https://www.jetbrains.com/idea/), here are some usefu
 - Ctrl+Alt+L: format file (ensures consistency in the codebase)
 - Ctrl+Alt+o: optimize imports (removes unused imports)
 
+Note that we use IntelliJ's default formatting configuration for Scala to minimize conflicts.
+
 ### Contribution Checklist
 
 - The code being submitted is accompanied by tests which exercise both the positive and negative
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,7 @@ You can also use Github issues for [feature requests](https://github.com/acinq/e
 - [Deployable Lightning](https://github.com/ElementsProject/lightning/raw/master/doc/deployable-lightning.pdf)
 - [Understanding the Lightning Network](https://bitcoinmagazine.com/articles/understanding-the-lightning-network-part-building-a-bidirectional-payment-channel-1464710791)
 - [Lightning Network Specification](https://github.com/lightningnetwork/lightning-rfc)
+- [High Level Lightning Network Specification](https://medium.com/@rusty_lightning/the-bitcoin-lightning-spec-part-1-8-a7720fb1b4da)
 
 ## Recommended Skillset
 
@@ -71,7 +72,9 @@ The reason for this is two fold: it makes it easier for the reviewer to see what
 made between versions (since Github doesn't easily show prior versions) and it makes it easier on
 the PR author as they can set it to auto-squash the fixup commits on rebase.
 
-### Code Style
+It's recommended to take great care in writing tests and ensuring the entire test suite has a
+stable successful outcome; eclair uses continuous integration techniques and having a stable build
+helps the reviewers with their job.
 
 We don't have hard rules around code style, but we do avoid having too many conflicting styles;
 look around and make sure you code fits well with the rest of the codebase.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,139 @@
+# Contributing to Eclair
+
+Eclair welcomes contributions in the form of peer review, testing and patches.
+This document explains the practical process and guidelines for contributing.
+
+While developing a Lightning implementation is an exciting project that spans many domains
+(cryptography, peer-to-peer networking, databases, etc), contributors must keep in mind that this
+represents real money and introducing bugs or security vulnerabilities can have far more dire
+consequences than in typical projects. In the world of cryptocurrencies, even the smallest bug in
+the wrong area can cost users a significant amount of money.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,139 @@
...
+
+When adding a new feature, thought must be given to the long term technical debt and maintenance
+that feature may require after inclusion. Before proposing a new feature that will require
+maintenance, please consider if you are willing to maintain it (including bug fixing).
+
+When addressing pull request comments, we recommend using [fixup commits](https://robots.thoughtbot.com/autosquashing-git-commits).
+The reason for this is two fold: it makes it easier for the reviewer to see what changes have been
+made between versions (since Github doesn't easily show prior versions) and it makes it easier on
+the PR author as they can set it to auto-squash the fixup commits on rebase.
+
```
araspitzu: Maybe it goes without saying but it's helpful and _recommended_ to run all the test before opening a pull request and keep the continuous integration build as stable as possible.
```suggestion
It's recommended to take great care in writing test and ensuring the entire test suite has a stable successful outcome, eclair uses continuous integration techniques and having a stable build helps the reviewers with their job.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,139 @@
...
+You can also use Github issues for [feature requests](https://github.com/acinq/eclair/issues?q=is%3Aissue+label%3A"feature+request").
+
+## Recommended Reading
+
+- [Bitcoin Whitepaper](https://bitcoin.org/bitcoin.pdf)
+- [Lightning Network Whitepaper](https://lightning.network/lightning-network-paper.pdf)
+- [Deployable Lightning](https://github.com/ElementsProject/lightning/raw/master/doc/deployable-lightning.pdf)
+- [Understanding the Lightning Network](https://bitcoinmagazine.com/articles/understanding-the-lightning-network-part-building-a-bidirectional-payment-channel-1464710791)
+- [Lightning Network Specification](https://github.com/lightningnetwork/lightning-rfc)
+
```
araspitzu: When reading about the protocol i personally found very valuable this small series of BOLT protocol explanations https://medium.com/@rusty_lightning/the-bitcoin-lightning-spec-part-1-8-a7720fb1b4da might be worth adding them

t-bast: I wasn't lucky enough to stumble on that one when I started :D
Let's add it to help newcomers!

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+- Typically a hyphen or asterisk is used for the bullet, preceded by a
+  single space, with blank lines in between, but conventions vary here
+- Use a hanging indent
+```
+
+### Dependencies
+
+We try to minimize our dependencies (libraries and tools). Introducing new dependencies increases
+package size, attack surface and cognitive overhead.
+
```
sstone: We should specify that eclair also runs on Android which is still using Java 7, so new dependencies must also be compatible with Java 7.

CONTRIBUTING.md
```diff
@@ -0,0 +1,142 @@
...
+2. Create a topic branch
+3. Add commits
+4. Open a pull request
+
+### Pull Request Philosophy
+
+Pull requests should always be focused. For example, a pull request could add a feature, fix a bug,
+or refactor code; but not a mixture.
+Please also avoid super pull requests which attempt to do too much, are overly large, or overly
+complex as this makes review difficult.
```
sstone: I would also add something like "Pull request should always include a clean, detailed description of what they fix/improve, why, and how. Even if you think that it is obvious, don't be shy and add explicit details and explanations."

CONTRIBUTING.md
t-bast: Good point!

dpad85:
I have a general comment regarding the "code formatting in IntelliJ" section: we use the default Scala formatting code style from IntelliJ which is great. However if a contributor has modified this default setting, or use an old version of IntelliJ, `Ctrl+Alt+L` will not help.

It could be useful to export this code style setting and add it to the project. IntelliJ supports the `EditorConfig` [standard](https://editorconfig.org/) through a plugin. That being said, I'm not sure how fine the `.editorconfig` code style settings can be, and if we can achieve exactly what we want.

t-bast:
> I have a general comment regarding the "code formatting in IntelliJ" section

Good point, I don't think we need to go too far there. I'll add a comment saying that we're using the defaults. If that becomes a problem with too many contributors having different settings, then we'll revisit that and push a configuration to make sure everyone uses the same codestyle.

SECURITY.md
```diff
@@ -0,0 +1,15 @@
...
+To report security issues send an email to security@acinq.fr (not for support).
+
+The following keys may be used to communicate sensitive information to developers:
+
+| Name                | Fingerprint                                       |
+|---------------------|---------------------------------------------------|
+| Pierre-Marie Padiou | 6AA4 5A4C 209A 2D30 64CF 66BE E434 ED29 2E85 643A |
+| Fabrice Drouin      | C25A 288A 842E AF7A A5B5 303F 7A73 FE77 DE2C 4027 |
+
+You can import a key by running the following command with that individual’s fingerprint: `gpg --recv-keys "<fingerprint>"`.
```
araspitzu: In my case `gpg --recv-keys <fingerprint>` is not enough to import the keys because it fails to fetch the key for the fingerprint, adding the key server option fixed it:
`gpg --keyserver https://acinq.co/pgp/drouinf.asc --recv-keys C25A288A842EAF7AA5B5303F7A73FE77DE2C4027`

I couldn't find the key from @pm47 on `acinq.co/pgp` and using `https://api.github.com/users/pm47/gpg_keys` as key server doesn't work.

I propose to upload the keys and provide instructions to import the keys correctly.

t-bast: Good catch, I'll dive into that (sounds like a rabbit hole)

dpad85: > I couldn't find the key from @pm47 on `acinq.co/pgp` and using `https://api.github.com/users/pm47/gpg_keys` as key server doesn't work.

It's available on: https://acinq.co/pgp/padioupm.asc

PR #1253, opened 2019-12-12
Flat features

t-bast:
This PR implements https://github.com/lightningnetwork/lightning-rfc/pull/666

I removed the awkward `--allowMultiPart` API parameter to instead rely on `nodeParams.features`.
This feels cleaner and is how we'll integrate Trampoline support too.

Note that when we update our nodes, we'll need to update our `reference.conf`:

- change the `features` format
- update our `features` to advertise MPP support

Compatibility tests:

- With old eclair nodes, advertizing MPP and PaymentSecret: ok
- With c-lightning master: ok
- With lnd v0.8.2-beta: ok

Not done in this PR (and should probably be done at some point):

- real support for turning off features (e.g. if I turn off MPP, I should use a legacy PayFSM)
- better typing to indicate which feature bits go in Init vs ChannelAnn vs NodeAnn vs Invoice


eclair-core/src/main/resources/reference.conf
```diff
@@ -37,13 +37,11 @@ eclair {
   node-color = "49daaa"
 
   trampoline-payments-enable = false // TODO: @t-bast: once spec-ed this should use a global feature flag
-  global-features = "0200" // variable_length_onion
-  local-features = "088a" // initial_routing_sync + option_data_loss_protect + option_channel_range_queries + option_channel_range_queries_ex
+  features = "0a8a" // initial_routing_sync + option_data_loss_protect + option_channel_range_queries + option_channel_range_queries_ex + variable_length_onion
```
araspitzu: We could enhance the [`StartupSpec`](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/test/scala/fr/acinq/eclair/StartupSpec.scala#L31) configuration test to prevent people from using the old config keys.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
```diff
@@ -370,14 +370,16 @@ class PaymentPacketSpec extends FunSuite with BeforeAndAfterAll {
...
-    val Left(failure) = decrypt(UpdateAddHtlc(randomBytes32, 2, amount_bc, paymentHash, expiry_bc - CltvExpiryDelta(12), packet_c), priv_c.privateKey, TestConstants.globalFeatures)
+    val Left(failure) = decrypt(UpdateAddHtlc(randomBytes32, 2, amount_bc, paymentHash, expiry_bc - CltvExpiryDelta(12), packet_c), priv_c.privateKey, variableLengthOnionFeature)
     assert(failure === FinalIncorrectCltvExpiry(expiry_bc - CltvExpiryDelta(12)))
   }
 
 }
 
 object PaymentPacketSpec {
 
+  val variableLengthOnionFeature = hex"0200"
```
araspitzu: nit: this could be defined using the feature definition at `Features.VARIABLE_LENGTH_ONION_OPTIONAL`

eclair-core/src/main/resources/reference.conf
t-bast: Great idea, I'll do that!

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
t-bast: True, that would be more readable.

t-bast:
I took this opportunity to re-work `Features.scala` to properly type things.
We're now stricter on features dependencies, as described [here](https://github.com/lightningnetwork/lightning-rfc/pull/719).
Note that two tests are failing because the official spec test vectors haven't been updated yet.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -19,63 +19,123 @@ package fr.acinq.eclair
...
-  val TRAMPOLINE_PAYMENT_MANDATORY = 50
-  val TRAMPOLINE_PAYMENT_OPTIONAL = 51
+  case object TrampolinePayment extends Feature {
+    val rfcName = "trampoline_payment"
+    val mandatory = 50
+    val optional = 51
+  }
+
+  // Features may depend on other features, as specified in Bolt 9.
+  private val featuresDependency = Map(
```
pm47: :bowtie::bowtie::bowtie:

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -19,63 +19,123 @@ package fr.acinq.eclair
...
+sealed trait FeatureSupport
+
+// @formatter:off
+object FeatureSupport {
+  case object Mandatory extends FeatureSupport
+  case object Optional extends FeatureSupport
+}
+// @formatter:on
+
+sealed trait Feature {
```
pm47: Not too sure about this:

```suggestion
sealed trait Bolt9Feature {
```

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -19,63 +19,123 @@ package fr.acinq.eclair
...
+  val rfcName: String
+  val mandatory: Int
+  val optional: Int
+
+  override def toString = rfcName
+}
+
 object Features {
```
pm47:
```suggestion
object Bolt9Features {
```

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -19,63 +19,123 @@ package fr.acinq.eclair
...
+    PaymentSecret -> (VariableLengthOnion :: Nil),
+    BasicMultiPartPayment -> (PaymentSecret :: Nil),
+    TrampolinePayment -> (PaymentSecret :: Nil)
+  )
+
+  case class FeatureException(message: String) extends IllegalArgumentException(message)
+
+  def validateFeatureGraph(features: BitVector): Option[FeatureException] = featuresDependency.collectFirst {
+    case (feature, dependencies) if hasFeature(features, feature, None) && dependencies.exists(d => !hasFeature(features, d, None)) =>
+      FeatureException(s"${features.toBin} sets $feature but is missing a dependency (${dependencies.mkString(" or ")})")
```
pm47: A bit more expensive, but much more friendly:

```suggestion
      FeatureException(s"${features.toBin} sets $feature but is missing a dependency (${dependencies.filter(d => !hasFeature(features, d, None)).mkString(" and ")})")
```

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -170,11 +173,13 @@ object NodeParams {
     val nodeAlias = config.getString("node-alias")
     require(nodeAlias.getBytes("UTF-8").length <= 32, "invalid alias, too long (max allowed 32 bytes)")
 
-    val overrideFeatures: Map[PublicKey, (ByteVector, ByteVector)] = config.getConfigList("override-features").map { e =>
+    val features = ByteVector.fromValidHex(config.getString("features"))
+    Features.validateFeatureGraph(features).foreach(e => throw e)
```
pm47: I don't like that we  break from the `require` pattern, but can't see a better way of doing this right off the bat.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -137,19 +136,15 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
 
       log.info(s"peer is using globalFeatures=${remoteInit.globalFeatures.toBin} and localFeatures=${remoteInit.localFeatures.toBin}")
 
-      if (Features.areSupported(remoteInit.localFeatures)) {
+      if (Features.areSupported(remoteInit.features)) {
         d.origin_opt.foreach(origin => origin ! "connected")
 
-        import Features._
+        def hasLocalFeature(f: Feature) = Features.hasFeature(d.localInit.features, f, None)
```
```suggestion
        def localHasFeature(f: Feature) = Features.hasFeature(d.localInit.features, f, None)
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -137,19 +136,15 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
...
 
-        def hasLocalFeature(bit: Int) = Features.hasFeature(d.localInit.localFeatures, bit)
-
-        def hasRemoteFeature(bit: Int) = Features.hasFeature(remoteInit.localFeatures, bit)
-
-        val canUseChannelRangeQueries = (hasLocalFeature(CHANNEL_RANGE_QUERIES_BIT_OPTIONAL) || hasLocalFeature(CHANNEL_RANGE_QUERIES_BIT_MANDATORY)) && (hasRemoteFeature(CHANNEL_RANGE_QUERIES_BIT_OPTIONAL) || hasRemoteFeature(CHANNEL_RANGE_QUERIES_BIT_MANDATORY))
-
-        val canUseChannelRangeQueriesEx = (hasLocalFeature(CHANNEL_RANGE_QUERIES_EX_BIT_OPTIONAL) || hasLocalFeature(CHANNEL_RANGE_QUERIES_EX_BIT_MANDATORY)) && (hasRemoteFeature(CHANNEL_RANGE_QUERIES_EX_BIT_OPTIONAL) || hasRemoteFeature(CHANNEL_RANGE_QUERIES_EX_BIT_MANDATORY))
+        def hasRemoteFeature(f: Feature) = Features.hasFeature(remoteInit.features, f, None)
```
```suggestion
        def remoteHasFeature(f: Feature) = Features.hasFeature(remoteInit.features, f, None)
```

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -19,63 +19,123 @@ package fr.acinq.eclair
...
+object FeatureSupport {
+  case object Mandatory extends FeatureSupport
+  case object Optional extends FeatureSupport
+}
+// @formatter:on
+
+sealed trait Feature {
+  val rfcName: String
+  val mandatory: Int
+  val optional: Int
```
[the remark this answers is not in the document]
t-bast: Heh I hesitated...

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: Not sure either, it's a bit verbose...

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
t-bast: Yeah me too, I'll improve that.

t-bast:
I fixed the nits in `aee1472`.

> For Init I disagree with your approach of changing the type instead of the codec. If you figure it out we could probably do the same for LocalParams/NodeParams.

I'll investigate that.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: :sunglasses:

eclair-core/src/main/resources/reference.conf
```diff
@@ -37,14 +37,12 @@ eclair {
...
```
pm47: I wonder if we should use a completely different method, using `rfcName` field:

```
features {
  option_data_loss_protect = "mandatory"
  initial_routing_sync = "optional"
  gossip_queries = "mandatory"
  ...
} 
```

Less error prone and more readable, but also less flexible.

pm47: Would also affect the `override-features` attribute...

t-bast: I think it's a good idea if we really want to allow users to turn some features on and off.
The problem is there are quite a few features that we can't really turn off right now (some code paths will ignore the fact that a feature is turned off and still use it).
I think it makes sense to do that change once we've worked on making sure the code base really allows completely turning off all features.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -19,63 +19,115 @@ package fr.acinq.eclair
...
+      FeatureException(s"${features.toBin} sets $feature but is missing a dependency (${dependencies.filter(d => !hasFeature(features, d, None)).mkString(" and ")})")
+  }
+
+  def validateFeatureGraph(features: ByteVector): Option[FeatureException] = validateFeatureGraph(features.bits)
 
   // Note that BitVector indexes from left to right whereas the specification indexes from right to left.
   // This is why we have to reverse the bits to check if a feature is set.
 
-  def hasFeature(features: BitVector, bit: Int): Boolean = if (features.sizeLessThanOrEqual(bit)) false else features.reverse.get(bit)
+  private def hasFeature(features: BitVector, bit: Int): Boolean = if (features.sizeLessThanOrEqual(bit)) false else features.reverse.get(bit)
```
pm47:
```suggestion
  private def hasFeature(features: BitVector, bit: Int): Boolean = features.sizeGreaterThanOrEqual(bit) && features.reverse.get(bit)
```

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -19,63 +19,115 @@ package fr.acinq.eclair
...
 
-  def hasFeature(features: ByteVector, bit: Int): Boolean = hasFeature(features.bits, bit)
+  def hasFeature(features: BitVector, feature: Feature, support: Option[FeatureSupport]): Boolean = support match {
```
pm47: I think that's a rare case where I would put a default value for field `support`, because most of the time we just want to know if a feature is supported, without more details.

```suggestion
  def hasFeature(features: BitVector, feature: Feature, support: Option[FeatureSupport] = None): Boolean = support match {
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -16,22 +16,38 @@
...
 
-  val initCodec: Codec[Init] = (
-    ("globalFeatures" | varsizebinarydata) ::
-      ("localFeatures" | varsizebinarydata)).as[Init]
+  /** For historical reasons, features are divided into two feature bitmasks. We only send from the second one, but we allow receiving in both. */
+  val combinedFeaturesCodec: Codec[ByteVector] = (("globalFeatures" | varsizebinarydata) :: ("localFeatures" | varsizebinarydata)).exmap[ByteVector](
+    features => {
+      val f1 = features.head
+      val f2 = features.tail.head
+      val combinedFeatures = if (f1.length == f2.length) {
```
sstone: nitpick: instead of creating a codec for a list and implicitly using the fact that it contains 2 elements, we could use a codec for a `(ByteVector, ByteVector)` instead and make this explicit, something like:
```scala
val combinedFeaturesCodec: Codec[ByteVector] = (("globalFeatures" | varsizebinarydata) :: ("localFeatures" | varsizebinarydata)).as[(ByteVector, ByteVector)].xmap(
    { case (f1, f2) =>
      val combinedFeatures = if (f1.length == f2.length) {
        f1 | f2
      } else if (f1.length > f2.length) {
        f1 | f2.padLeft(f1.length)
      } else {
        f1.padLeft(f2.length) | f2
      }
      combinedFeatures
    },
    { features => (ByteVector.empty, features) }
  )
```

pm47:
```scala
  val combinedFeaturesCodec: Codec[ByteVector] = (("globalFeatures" | varsizebinarydata) :: ("localFeatures" | varsizebinarydata)).as[(ByteVector, ByteVector)].xmap(
    { case (f1, f2) =>
      val length = f1.size.max(f2.size)
      f1.padLeft(length) | f2.padLeft(length)
    },
    { features => (ByteVector.empty, features) }
  )
```

eclair-core/src/test/scala/fr/acinq/eclair/wire/LightningMessageCodecsSpec.scala
```diff
@@ -43,6 +43,24 @@ class LightningMessageCodecsSpec extends FunSuite {
...
+      (hex"00020200 0002088a", hex"0a8a", hex"0000 00020a8a"), // local and global - no conflict - same size
+      (hex"00020200 0003020002", hex"020202", hex"0000 0003020202"), // local and global - no conflict - different sizes
+      (hex"00020a02 0002088a", hex"0a8a", hex"0000 00020a8a"), // local and global - conflict - same size
+      (hex"00022200 000302aaa2", hex"02aaa2", hex"0000 000302aaa2") // local and global - conflict - different sizes
+    )
+
+    for ((bin, features, encoded) <- testCases) {
+      val init = initCodec.decode(bin.bits).require.value
+      assert(init.features === features)
+      assert(initCodec.encode(init).require.bytes === encoded)
```
pm47: We also need to make sure that we correctly decode migrated data:

```suggestion
      assert(initCodec.encode(init).require.bytes === encoded)
      assert(initCodec.decode(encoded.bits).require.value === init)
```

t-bast: Good point, adding that.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
t-bast: Thanks for the suggestions!

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
pm47: Maybe also `sed s/hasFeature(features, bit, None)/hasFeature(features, bit)/g` in the rest of the code?

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
pm47: Maybe also `sed s/hasFeature(features, bit, None)/hasFeature(features, bit)/g` in the rest of the code?

t-bast: Done!

PR #1254, opened 2019-12-13
Add networks to init message

t-bast:
Implement https://github.com/lightningnetwork/lightning-rfc/pull/682

This is a good opportunity to setup the infrastructure for a TLV stream in the Init message, which we can leverage for optional, un-speced features.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -100,8 +100,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
       log.debug(s"got authenticated connection to $remoteNodeId@${address.getHostString}:${address.getPort}")
       transport ! TransportHandler.Listener(self)
       context watch transport
-      val localInit = nodeParams.overrideFeatures.get(remoteNodeId) match {
-        case Some(f) => wire.Init(f, TlvStream(InitTlv.Networks(nodeParams.chainHash :: Nil)))
+      val localFeatures = nodeParams.overrideFeatures.get(remoteNodeId) match {
+        case Some(f) => f
         case None =>
           // Eclair-mobile thinks feature bit 15 (payment_secret) is gossip_queries_ex which creates issues, so we mask
           // off basic_mpp and payment_secret. As long as they're provided in the invoice it's not an issue.
@@ -116,9 +116,10 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
             // ... and leave the others untouched
             case (value, _) => value
           }).reverse.bytes.dropWhile(_ == 0)
-          wire.Init(tweakedFeatures, TlvStream(InitTlv.Networks(nodeParams.chainHash :: Nil)))
+          tweakedFeatures
       }
-      log.info(s"using features=${localInit.features.toBin}")
+      log.info(s"using features=${localFeatures.toBin}")
+      val localInit = wire.Init(localFeatures, TlvStream(InitTlv.Networks(nodeParams.chainHash :: Nil)))
       transport ! localInit
 
       val address_opt = if (outgoing) {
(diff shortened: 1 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
```diff
@@ -45,7 +45,9 @@ sealed trait HasChainHash extends LightningMessage { def chainHash: ByteVector32
 sealed trait UpdateMessage extends HtlcMessage // <- not in the spec
 // @formatter:on
 
-case class Init(features: ByteVector) extends SetupMessage
+case class Init(features: ByteVector, tlvs: TlvStream[InitTlv] = TlvStream.empty) extends SetupMessage {
+  val networks = tlvs.get[InitTlv.Networks].map(_.chainHashes).getOrElse(Nil)
```
pm47: In theory this should be an `Option[List]` imo, but in that particular case it makes sense to equal `Nil` and `None`.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -101,7 +101,7 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
       transport ! TransportHandler.Listener(self)
       context watch transport
       val localInit = nodeParams.overrideFeatures.get(remoteNodeId) match {
-        case Some(f) => wire.Init(f)
+        case Some(f) => wire.Init(f, TlvStream(InitTlv.Networks(nodeParams.chainHash :: Nil)))
```
pm47: We could factor the definition of `TlvStream(InitTlv.Networks(nodeParams.chainHash :: Nil))`.

Actually I think the pattern matching should only return `features` and we should build the `Init` after.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -148,9 +148,19 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
     case Event(remoteInit: wire.Init, d: InitializingData) =>
       d.transport ! TransportHandler.ReadAck(remoteInit)
 
-      log.info(s"peer is using features=${remoteInit.features.toBin}")
+      log.info(s"peer is using features=${remoteInit.features.toBin}, network=${remoteInit.networks}")
```
```suggestion
      log.info(s"peer is using features=${remoteInit.features.toBin}, networks=${remoteInit.networks.mkString(",")}")
```

eclair-core/src/test/scala/fr/acinq/eclair/wire/LightningMessageCodecsSpec.scala
```diff
@@ -44,21 +44,36 @@ class LightningMessageCodecsSpec extends FunSuite {
...
+    val chainHash2 = ByteVector32(hex"0202020202020202020202020202020202020202020202020202020202020202")
     val testCases = Seq(
-      (hex"0000 0000", hex"", hex"0000 0000"), // no features
-      (hex"0000 0002088a", hex"088a", hex"0000 0002088a"), // no global features
-      (hex"00020200 0000", hex"0200", hex"0000 00020200"), // no local features
-      (hex"00020200 0002088a", hex"0a8a", hex"0000 00020a8a"), // local and global - no conflict - same size
-      (hex"00020200 0003020002", hex"020202", hex"0000 0003020202"), // local and global - no conflict - different sizes
-      (hex"00020a02 0002088a", hex"0a8a", hex"0000 00020a8a"), // local and global - conflict - same size
-      (hex"00022200 000302aaa2", hex"02aaa2", hex"0000 000302aaa2") // local and global - conflict - different sizes
+      (hex"0000 0000", hex"", Nil, true, None), // no features
```
pm47: With that many elements in a tuple a case class could come in handy.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageTypes.scala
t-bast: Agreed, that was my reasoning too.

eclair-core/src/test/scala/fr/acinq/eclair/wire/LightningMessageCodecsSpec.scala
t-bast: :+1:

PR #1256, opened 2019-12-15
Provided solution to "non-segwit UTXO found."

dscotese:


README.md
```diff
@@ -48,7 +48,7 @@ You will find detailed guides and frequently asked questions there.
 :warning: Eclair requires Bitcoin Core 0.17.1 or higher. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
 
 Eclair needs a _synchronized_, _segwit-ready_, **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node.
-Eclair will use any BTC it finds in the Bitcoin Core wallet to fund any channels you choose to open. Eclair will return BTC from closed channels to this wallet.
+Eclair will use any BTC it finds in the Bitcoin Core wallet to fund any channels you choose to open. Eclair will return BTC from closed channels to this wallet. Eclair will stop with an error if the wallet contains any outputs that are not either `p2sh-segwit` addresses or `bech32` addresses. In this case, you must send all your bitcoins to a new `p2sh-segwit` or `bech32` address before starting Eclair.
```
t-bast: I think this is too verbose. The line below ("You can configure your Bitcoin Node to use either `p2sh-segwit` addresses or `bech32` addresses, Eclair is compatible with both modes.") already implies that only `p2sh-segwit` and `bech32` are supported.

If you'd like to add extra clarification, I would only add the following _after line 52_:
```
If your Bitcoin Core wallet has outputs that are neither `p2sh-segwit` or `bech32`, you must send them to a `p2sh-segwit` or `bech32` address.
```

dscotese: I agree. I made that change in my fork.  I'll see what "Resolve conversation does..."

PR #1258, opened 2019-12-15
Better Non-segwit UTXOs assert message.

dscotese:
The error message produced now provides the user with a solution.  Of course it's temporary until they receive another non-segwit UTXO.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -170,7 +170,7 @@ class Setup(datadir: File,
       assert(bitcoinVersion >= 170000, "Eclair requires Bitcoin Core 0.17.0 or higher")
       assert(chainHash == nodeParams.chainHash, s"chainHash mismatch (conf=${nodeParams.chainHash} != bitcoind=$chainHash)")
       if (chainHash != Block.RegtestGenesisBlock.hash) {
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -170,7 +170,7 @@ class Setup(datadir: File,
       assert(bitcoinVersion >= 170000, "Eclair requires Bitcoin Core 0.17.0 or higher")
       assert(chainHash == nodeParams.chainHash, s"chainHash mismatch (conf=${nodeParams.chainHash} != bitcoind=$chainHash)")
       if (chainHash != Block.RegtestGenesisBlock.hash) {
-        assert(unspentAddresses.forall(address => !isPay2PubkeyHash(address)), "Make sure that all your UTXOS are segwit UTXOS and not p2pkh (check out our README for more details)")
+        assert(unspentAddresses.forall(address => !isPay2PubkeyHash(address)), "To use the current bitcoin wallet, you must send the currently existing non-segwit UTXOs to a p2sh-segwit or bech32 address before starting Eclair. (check out our README for more details)")
```
t-bast: What about:

```
Your wallet contains non-segwit UTXOs (p2pkh). You must send those UTXOs to a p2sh-segwit or bech32 address to use Eclair (check out our README for more details).
```

dscotese: I like it, but in case p2pkh isn't the **only** UTXO that is non-segwit, I just removed "(p2pkh)". I guess a pull request is for a branch and not a specific commit, so the commits I just made to update it as you suggested (but without p2pkh) are part of this pull request.  Please correct my understanding if I'm wrong about that.

PR #1259, opened 2019-12-17
Update Maven (3.6.3)

t-bast:
This fixes the failing builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,12 +7,12 @@ RUN apk add --no-cache curl tar bash
 
 ARG MAVEN_VERSION=3.6.3
 ARG USER_HOME_DIR="/root"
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM adoptopenjdk/openjdk11:jdk-11.0.3_7-alpine as BUILD
 # this was meant to make the image smaller, but we use multi-stage build so we don't care
 RUN apk add --no-cache curl tar bash
 
-ARG MAVEN_VERSION=3.6.2
+ARG MAVEN_VERSION=3.6.3
 ARG USER_HOME_DIR="/root"
 ARG SHA=3fbc92d1961482d6fbd57fbf3dd6d27a4de70778528ee3fb44aa7d27eb32dfdc
```
sstone: Don't we need to also change this checksum ? (unless it's never checked)

t-bast: Good catch! I should test the Docker build properly, will do that.

PR #1261, opened 2019-12-19
Improve NodeRelayer errors

t-bast:
We add new errors that let senders know they need to raise the trampoline fee/ctlv.
When the error is downstream, we select the best error to forward.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
@@ -91,6 +91,7 @@ class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, c
           case Some(failure) =>
             log.warning(s"rejecting trampoline payment (amountIn=${upstream.amountIn} expiryIn=${upstream.expiryIn} amountOut=${nextPayload.amountToForward} expiryOut=${nextPayload.outgoingCltv} htlcCount=${parts.length} reason=$failure)")
             rejectPayment(upstream, Some(failure))
+            context become main(pendingIncoming - paymentHash, pendingOutgoing)
           case None =>
             log.info(s"relaying trampoline payment (amountIn=${upstream.amountIn} expiryIn=${upstream.expiryIn} amountOut=${nextPayload.amountToForward} expiryOut=${nextPayload.outgoingCltv} htlcCount=${parts.length})")
             val paymentId = relay(paymentHash, upstream, nextPayload, nextPacket)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
@@ -21,14 +21,15 @@ import java.util.UUID
 import akka.actor.{Actor, ActorRef, DiagnosticActorLogging, PoisonPill, Props}
 import akka.event.Logging.MDC
 import fr.acinq.bitcoin.ByteVector32
+import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.channel.{CMD_FAIL_HTLC, CMD_FULFILL_HTLC, Upstream}
+import fr.acinq.eclair.payment._
 import fr.acinq.eclair.payment.receive.MultiPartPaymentFSM
 import fr.acinq.eclair.payment.send.MultiPartPaymentLifecycle.SendMultiPartPayment
 import fr.acinq.eclair.payment.send.PaymentInitiator.SendPaymentConfig
 import fr.acinq.eclair.payment.send.PaymentLifecycle.SendPayment
 import fr.acinq.eclair.payment.send.{MultiPartPaymentLifecycle, PaymentLifecycle}
-import fr.acinq.eclair.payment.{IncomingPacket, PaymentFailed, PaymentSent, TrampolinePaymentRelayed}
-import fr.acinq.eclair.router.{RouteParams, Router}
+import fr.acinq.eclair.router.{RouteNotFound, RouteParams, Router}
 import fr.acinq.eclair.wire._
 import fr.acinq.eclair.{CltvExpiry, Logs, MilliSatoshi, NodeParams, nodeFee, randomBytes32}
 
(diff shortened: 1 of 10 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -240,4 +233,27 @@ object NodeRelayer {
...
+  private def translateError(failures: Seq[PaymentFailure], outgoingNodeId: PublicKey): Option[FailureMessage] = {
+    def tooManyRouteNotFound(failures: Seq[PaymentFailure]): Boolean = {
+      val routeNotFoundCount = failures.count(f => f == LocalFailure(RouteNotFound))
+      routeNotFoundCount > failures.length / 2
+    }
+
+    failures match {
+      case Nil => None
+      case LocalFailure(MultiPartPaymentLifecycle.BalanceTooLow) :: Nil => Some(TemporaryNodeFailure) // we don't have enough outgoing liquidity at the moment
+      case _ if tooManyRouteNotFound(failures) => Some(TrampolineFeeInsufficient) // if we couldn't find routes, it's likely that the fee/cltv was insufficient
```
dpad85: If I understand correctly, having more than 50% of `RouteNotFound` would mean that we should increase the trampoline fees for this payment to pass. How safe is that 50% estimate in terms of false positives?

t-bast: It's honestly impossible to say. But we need to impose some kind of threshold to get a short-term solution. We'll be able to greatly improve that and provide accurate retry fees/cltv once MPP is directly integrated inside the router, but until then we can only do heuristics.
Note that when that happens we can't know yet if what needs to be raised is fees, cltv or both. So for now Phoenix should slightly raise both (but for CLTV we can keep the upper bound described [here](https://github.com/ACINQ/phoenix/issues/1)).

t-bast: Another thing we could do that would greatly improve the Phoenix-to-non-Phoenix payments is to temporarily disable MPP between Acinq and the recipient for those.
Then there won't be any false positives.
I don't expect many people to do payments that really require MPP in the next couple of months, so that's probably the safe choice for now. I could do that in this PR, it's a simple change. WDYT?

t-bast:
I added the sender-side in `PaymentInitiator`. I hesitated to create a dedicated `TrampolinePaymentLifecycle` actor, but it felt a bit overkill at the moment.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -72,7 +72,7 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
       router ! FinalizeRoute(c.hops)
       if (cfg.storeInDb) {
         val targetNodeId = cfg.trampolineData.map(_.paymentRequest.nodeId).getOrElse(cfg.targetNodeId)
-        val finalAmount = c.finalPayload.amount - cfg.trampolineData.map(_.trampolineFees).getOrElse(0 msat)
+        val finalAmount = c.finalPayload.amount - cfg.trampolineData.map(_.trampolineAttempts.headOption.map(_._1).getOrElse(0 msat)).getOrElse(0 msat)
```
t-bast: Please ignore this hack. This is because we haven't update the DB schema yet and we needed to take trampoline fees into account in a hackish way. This will be fixed in the next trampoline PR which update the DB schema to clean things up.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -67,54 +69,93 @@ class PaymentInitiator(nodeParams: NodeParams, router: ActorRef, relayer: ActorR
     case r: SendTrampolinePaymentRequest =>
       val paymentId = UUID.randomUUID()
       sender ! paymentId
-      if (!r.paymentRequest.features.allowTrampoline && r.paymentRequest.amount.isEmpty) {
+      if (r.trampolineAttempts.isEmpty) {
```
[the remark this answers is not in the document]
t-bast: Thanks, I changed this way too many times, it was time to do a small clean-up

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -67,54 +69,93 @@ class PaymentInitiator(nodeParams: NodeParams, router: ActorRef, relayer: ActorR
...
-          OutgoingPacket.buildPacket(Sphinx.TrampolinePacket)(r.paymentRequest.paymentHash, trampolineRoute, finalPayload)
-        } else {
-          OutgoingPacket.buildTrampolineToLegacyPacket(r.paymentRequest, trampolineRoute, finalPayload)
-        }
-        // We generate a random secret for this payment to avoid leaking the invoice secret to the first trampoline node.
-        val trampolineSecret = randomBytes32
-        spawnMultiPartPaymentFsm(paymentCfg) forward SendMultiPartPayment(r.paymentRequest.paymentHash, trampolineSecret, r.trampolineNodeId, trampolineAmount, trampolineExpiry, 1, r.paymentRequest.routingInfo, r.routeParams, Seq(OnionTlv.TrampolineOnion(trampolineOnion.packet)))
+      r.trampolineAttempts match {
+        case Nil =>
+          sender ! PaymentFailed(paymentId, r.paymentRequest.paymentHash, LocalFailure(new IllegalArgumentException("trampoline fees and cltv expiry delta are missing")) :: Nil)
```
dpad85: `LocalFailure` accepts a generic `Throwable` for the failure details, however it would be useful to type this failure, like it was done with `RouteNotFound` or `BalanceTooLow`. This would let the front provide an accurate error message (incl. localization) with potentially a different flow depending on the error.

Ideally `LocalFailure` would force the throwable to be typed. There are several options to do that, but let's not derail this PR. 

In the meantime I think `LocalFailure` should contain a specific object as much as possible.

t-bast: Good idea, I'll create a specific object for that error and will improve the overall situation in the next trampoline PR.

t-bast: Done

perrystan1489:
I think I might can help with this

13202916251441262485:
في سبت، ٢١ أكتوبر، ٢٠٢٣ في ٩:٠٢ م، كتب Perrystan1489 <
***@***.***>:

PR #1262, opened 2019-12-22
Added some instructions for TOR auth

mikewerwin:
*.tor.auth defaults to password and cookie auth folks may not notice without log crawling.


t-bast:
I had to whitelist you for our CI builds, could you do a dummy push (e.g. amending your commit message) to trigger the build and unblock the merge?

PR #1264, opened 2019-12-23
Provide additional information to KeyManager.sign

devrandom:
This will allow an external signer to fully validate the transaction.

For example, providing the redeem scripts allows checking:

- the OP_CHECKSEQUENCEVERIFY delay
- that a remote transaction really sends funds to us
- etc.

For more background on the external signer strategy, see https://gitlab.com/ksedgwic/liposig/-/wikis/Lightning-Policy-Signing-PoC


t-bast:
Closing this stale PR. We should instead integrate with the lightning-signer project whenever it will make sense.

PR #1265, opened 2019-12-23
Remove repeated code

btcontract:
Just a small refactoring


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -92,6 +92,21 @@ object Channel {
   case class RemoteError(e: Error) extends ChannelError
   // @formatter:on
 
+  def ackPendingFailsAndFulfills(updates: List[UpdateMessage], relayer: ActorRef): Unit = updates.collect {
+    case u: UpdateFailMalformedHtlc => relayer ! CommandBuffer.CommandAck(u.channelId, u.id)
+    case u: UpdateFulfillHtlc => relayer ! CommandBuffer.CommandAck(u.channelId, u.id)
+    case u: UpdateFailHtlc => relayer ! CommandBuffer.CommandAck(u.channelId, u.id)
+  }
+
+  def failPending(adds: Traversable[UpdateAddHtlc], chanException: (Origin, UpdateAddHtlc) => ChannelException, onFound: (Origin, UpdateAddHtlc) => Unit, onNotFound: UpdateAddHtlc => Unit, relayer: ActorRef, commits: Commitments): Unit =
```
t-bast: This looks like JavaScript, it doesn't feel right :). I preferred the duplication in that case, it's more readable.

PR #1272, opened 2020-01-10
Restore missing 'db.close()' methods

sstone:
They are used by our mobile applications.


t-bast:
Can't we put that in a shared trait and add a comment saying that even though it doesn't seem used, it is by higher level components (such as eclair-mobile)?

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -401,4 +401,5 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
     }
   }
 
+  override def close(): Unit = sqlite.close()
```
t-bast: nit: this one doesn't have the comment "used by mobile apps"

PR #1273, opened 2020-01-10
Avoid sending PaymentSecret/MPP in Init

t-bast:
There is currently a backwards-compatibility issue with eclair-mobile.
Eclair-mobile mistakes feature bit 15 (payment_secret) for the
gossip_queries_ex prototype (which is incompatible with the spec-ed version).

To temporarily avoid this issue (until eclair-mobile is patched and all users have updated),
we never advertize those ambiguous bits in Init.
They're only really needed in the invoice so it's ok.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -33,7 +33,7 @@ import fr.acinq.eclair.wire._
 import fr.acinq.eclair.{wire, _}
 import kamon.Kamon
 import scodec.Attempt
-import scodec.bits.ByteVector
+import scodec.bits.{BitVector, ByteVector}
 
 import scala.compat.Platform
 import scala.concurrent.duration._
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -102,7 +102,14 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
       context watch transport
       val localInit = nodeParams.overrideFeatures.get(remoteNodeId) match {
         case Some(f) => wire.Init(f)
-        case None => wire.Init(nodeParams.features)
+        case None =>
+          // Eclair-mobile thinks feature bit 15 (payment_secret) is gossip_queries_ex which creates issues, so we mask
+          // off basic_mpp and payment_secret. As long as they're provided in the invoice it's not an issue.
+          // We use a long enough mask to account for future features.
+          // TODO: remove that once eclair-mobile is patched.
+          val featureMask = bin"111111111111111111111111111111000011111111111111".bytes
```
```scala
          val tweakedFeatures = BitVector.bits(nodeParams.features.bits.toIndexedSeq.zipWithIndex.map {
            // we disable those bits if they are set...
            case ((true, 14)) => false
            case ((true, 15)) => false
            case ((true, 16)) => false
            case ((true, 17)) => false
            // ... and leave the others untouched
            case (value, _) => value
          })
```

PR #1274, opened 2020-01-10
Fix how we split channel ids in channel queries

sstone:
We currently split in fixed-size chunks, which means that channel ids with the same block height may end up in different replies. Though this is not explicitly forbidden in the specs, it is implicitly required otherwise it becomes difficult to know when you've received all the replies you asked for and send another query.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -642,9 +642,7 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
           val shortChannelIds: SortedSet[ShortChannelId] = d.channels.keySet.filter(keep(firstBlockNum, numberOfBlocks, _))
           log.info("replying with {} items for range=({}, {})", shortChannelIds.size, firstBlockNum, numberOfBlocks)
           val chunks = Kamon.runWithSpan(Kamon.spanBuilder("split-channel-ids").start(), finishSpan = true) {
-            enforceMaximumSize(
-              split(shortChannelIds, firstBlockNum, numberOfBlocks, nodeParams.routerConf.channelRangeChunkSize),
-              Random.nextBoolean())
+            split(shortChannelIds, firstBlockNum, numberOfBlocks, nodeParams.routerConf.channelRangeChunkSize)
           }
 
           Kamon.runWithSpan(Kamon.spanBuilder("compute-timestamps-checksums").start(), finishSpan = true) {
@@ -1233,7 +1231,7 @@ object Router {
    *                              returned chunks may still contain more than `channelRangeChunkSize` elements
    * @return a list of short channel id chunks
    */
-  def split(shortChannelIds: SortedSet[ShortChannelId], firstBlockNum: Long, numberOfBlocks: Long, channelRangeChunkSize: Int): List[ShortChannelIdsChunk] = {
+  def split(shortChannelIds: SortedSet[ShortChannelId], firstBlockNum: Long, numberOfBlocks: Long, channelRangeChunkSize: Int, keepFirst: => Boolean = Random.nextBoolean()): List[ShortChannelIdsChunk] = {
     // see BOLT7: MUST encode a short_channel_id for every open channel it knows in blocks first_blocknum to first_blocknum plus number_of_blocks minus one
     val it = shortChannelIds.iterator.dropWhile(_.blockHeight < firstBlockNum).takeWhile(_.blockHeight < firstBlockNum + numberOfBlocks)
     if (it.isEmpty) {
@@ -1268,7 +1266,10 @@ object Router {
       }
 
       val first = it.next()
-      loop(first :: Nil, Nil)
+      val chunks = loop(first :: Nil, Nil)
+
+      // make sure that all our chunks match our max size policy
+      enforceMaximumSize(chunks, keepFirst)
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -1218,7 +1218,7 @@ object Router {
         // we use a random offset here, so even if shortChannelIds.size is much bigger than maximumSize (which should
         // not happen) peers will eventually receive info about all channels in this chunk
         val offset = Random.nextInt(shortChannelIds.size - maximumSize)
-        this.copy(shortChannelIds = this.shortChannelIds.drop(offset).take(maximumSize))
+        this.copy(shortChannelIds = this.shortChannelIds.slice(offset, offset + maximumSize))
       }
     }
   }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -71,6 +71,7 @@ case class RouterConf(randomizeRouteSelection: Boolean,
 
 // @formatter:off
 case class ChannelDesc(shortChannelId: ShortChannelId, a: PublicKey, b: PublicKey)
+
 case class PublicChannel(ann: ChannelAnnouncement, fundingTxid: ByteVector32, capacity: Satoshi, update_1_opt: Option[ChannelUpdate], update_2_opt: Option[ChannelUpdate]) {
   update_1_opt.foreach(u => assert(Announcements.isNode1(u.channelFlags)))
   update_2_opt.foreach(u => assert(!Announcements.isNode1(u.channelFlags)))
@@ -81,6 +82,7 @@ case class PublicChannel(ann: ChannelAnnouncement, fundingTxid: ByteVector32, ca
 
   def updateChannelUpdateSameSideAs(u: ChannelUpdate): PublicChannel = if (Announcements.isNode1(u.channelFlags)) copy(update_1_opt = Some(u)) else copy(update_2_opt = Some(u))
 }
+
 case class PrivateChannel(localNodeId: PublicKey, remoteNodeId: PublicKey, update_1_opt: Option[ChannelUpdate], update_2_opt: Option[ChannelUpdate]) {
   val (nodeId1, nodeId2) = if (Announcements.isNode1(localNodeId, remoteNodeId)) (localNodeId, remoteNodeId) else (remoteNodeId, localNodeId)
 
@@ -90,6 +92,7 @@ case class PrivateChannel(localNodeId: PublicKey, remoteNodeId: PublicKey, updat
 
   def updateChannelUpdateSameSideAs(u: ChannelUpdate): PrivateChannel = if (Announcements.isNode1(u.channelFlags)) copy(update_1_opt = Some(u)) else copy(update_2_opt = Some(u))
 }
+
 // @formatter:on
 
 case class AssistedChannel(extraHop: ExtraHop, nextNodeId: PublicKey, htlcMaximum: MilliSatoshi)
@@ -158,7 +161,9 @@ case class RouteResponse(hops: Seq[ChannelHop], ignoreNodes: Set[PublicKey], ign
 // @formatter:off
 /** This is used when we get a TemporaryChannelFailure, to give time for the channel to recover (note that exclusions are directed) */
 case class ExcludeChannel(desc: ChannelDesc)
+
 case class LiftChannelExclusion(desc: ChannelDesc)
+
 // @formatter:on
 
 case class SendChannelQuery(remoteNodeId: PublicKey, to: ActorRef, flags_opt: Option[QueryChannelRangeTlv])
@@ -194,11 +199,15 @@ case class Data(nodes: Map[PublicKey, NodeAnnouncement],
 
 // @formatter:off
 sealed trait State
+
 case object NORMAL extends State
 
 case object TickBroadcast
+
 case object TickPruneStaleChannels
+
 case object TickComputeNetworkStats
+
 // @formatter:on
 
 class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Promise[Done]] = None) extends FSMDiagnosticActorLogging[State, Data] {
@@ -1206,20 +1215,34 @@ object Router {
    * there could be several reply_channel_range messages for a single query
    */
   def split(shortChannelIds: SortedSet[ShortChannelId], channelRangeChunkSize: Int): List[ShortChannelIdsChunk] = {
-    // this algorithm can split blocks (meaning that we can in theory generate several replies with the same first_block/num_blocks
-    // and a different set of short_channel_ids) but it doesn't matter
     if (shortChannelIds.isEmpty) {
       List(ShortChannelIdsChunk(0, 0, List.empty))
     } else {
-      shortChannelIds
-        .grouped(channelRangeChunkSize)
-        .toList
-        .map { group =>
-          // NB: group is never empty
-          val firstBlock: Long = ShortChannelId.coordinates(group.head).blockHeight.toLong
-          val numBlocks: Long = ShortChannelId.coordinates(group.last).blockHeight.toLong - firstBlock + 1
-          ShortChannelIdsChunk(firstBlock, numBlocks, group.toList)
+
+      // we use an iterator for efficiency
+      val it = shortChannelIds.iterator
+
+      // we want to split ids in different chunks, with the following rules by order of priority
+      // ids that have the same block height must be grouped in the same chunk
+      // chunk should contain `channelRangeChunkSize` ids
+      @tailrec
+      def loop(currentHeight: Int, currentChunk: List[ShortChannelId], acc: List[ShortChannelIdsChunk]): List[ShortChannelIdsChunk] = {
+        if (it.hasNext) {
+          val id = it.next()
+          if (id.blockHeight == currentHeight)
+            loop(currentHeight, id :: currentChunk, acc) // same height => always add to the current chunk
+          else if (currentChunk.size < channelRangeChunkSize) // different height but we're under the size target => add to the current chunk
+            loop(id.blockHeight, id :: currentChunk, acc) // different height and over the size target => start a new chunk
+          else {
+            // we always prepend because it's more efficient so we have to reverse the current chunk
+            loop(id.blockHeight, id :: Nil, ShortChannelIdsChunk(currentChunk.last.blockHeight, currentChunk.head.blockHeight - currentChunk.last.blockHeight + 1, currentChunk.reverse) :: acc)
+          }
         }
+        else (ShortChannelIdsChunk(currentChunk.last.blockHeight, currentChunk.head.blockHeight - currentChunk.last.blockHeight + 1, currentChunk.reverse) :: acc).reverse
+      }
+
+      val first = it.next()
+      loop(ShortChannelId.coordinates(first).blockHeight, first :: Nil, Nil)
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -41,7 +41,7 @@ import scodec.bits.ByteVector
 import shapeless.HNil
 
 import scala.annotation.tailrec
-import scala.collection.immutable.SortedMap
+import scala.collection.immutable.{Queue, SortedMap}
 import scala.collection.{SortedSet, mutable}
 import scala.compat.Platform
 import scala.concurrent.duration._
@@ -642,10 +642,11 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
           val shortChannelIds: SortedSet[ShortChannelId] = d.channels.keySet.filter(keep(firstBlockNum, numberOfBlocks, _))
           log.info("replying with {} items for range=({}, {})", shortChannelIds.size, firstBlockNum, numberOfBlocks)
           val chunks = Kamon.runWithSpan(Kamon.spanBuilder("split-channel-ids").start(), finishSpan = true) {
-            split(shortChannelIds, firstBlockNum, numberOfBlocks, nodeParams.routerConf.channelRangeChunkSize)
+            enforceMaximumSize(
+              split(shortChannelIds, firstBlockNum, numberOfBlocks, nodeParams.routerConf.channelRangeChunkSize),
+              Random.nextBoolean())
           }
 
-          // we make sure that our replies cover the range specified in their queries
           Kamon.runWithSpan(Kamon.spanBuilder("compute-timestamps-checksums").start(), finishSpan = true) {
             chunks.foreach { chunk =>
               val (timestamps, checksums) = routingMessage.queryFlags_opt match {
@@ -944,6 +945,13 @@ object Router {
   val shortChannelIdKey = Context.key[ShortChannelId]("shortChannelId", ShortChannelId(0))
   val remoteNodeIdKey = Context.key[String]("remoteNodeId", "unknown")
 
+  // maximum number of ids we can keep in a single chunk and still have an encoded reply that is smaller than 65Kb
+  // please note that:
+  // - this is based on the worst case scenario where peer want timestamps and checksums and the reply is not compressed
+  // - the maximum number of public channels in a single block so far is less than 300, and the maximum number of tx per block
+  // almost never exceeds 2800 so this is not a real limitation yet
+  val MAXIMUM_CHUNK_SIZE = 3200
+
   def props(nodeParams: NodeParams, watcher: ActorRef, initialized: Option[Promise[Done]] = None) = Props(new Router(nodeParams, watcher, initialized))
 
   def toFakeUpdate(extraHop: ExtraHop, htlcMaximum: MilliSatoshi): ChannelUpdate = {
@@ -1015,7 +1023,7 @@ object Router {
    * Filters channels that we want to send to nodes asking for a channel range
    */
   def keep(firstBlockNum: Long, numberOfBlocks: Long, id: ShortChannelId): Boolean = {
-    val TxCoordinates(height, _, _) = ShortChannelId.coordinates(id)
+    val height = id.blockHeight
     height >= firstBlockNum && height < (firstBlockNum + numberOfBlocks)
   }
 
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1206,20 +1215,34 @@ object Router {
...
-          ShortChannelIdsChunk(firstBlock, numBlocks, group.toList)
+
+      // we use an iterator for efficiency
+      val it = shortChannelIds.iterator
+
+      // we want to split ids in different chunks, with the following rules by order of priority
+      // ids that have the same block height must be grouped in the same chunk
+      // chunk should contain `channelRangeChunkSize` ids
+      @tailrec
+      def loop(currentHeight: Int, currentChunk: List[ShortChannelId], acc: List[ShortChannelIdsChunk]): List[ShortChannelIdsChunk] = {
```
```suggestion
      def loop(currentChunk: List[ShortChannelId], acc: List[ShortChannelIdsChunk]): List[ShortChannelIdsChunk] = {
        val currentHeight = currentChunk.last.blockHeight
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -158,7 +161,9 @@ case class RouteResponse(hops: Seq[ChannelHop], ignoreNodes: Set[PublicKey], ign
 // @formatter:off
 /** This is used when we get a TemporaryChannelFailure, to give time for the channel to recover (note that exclusions are directed) */
 case class ExcludeChannel(desc: ChannelDesc)
+
```
pm47: there are some formatting glitches

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Also why not use a `Queue` intead of prepend + reverse?

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +128,23 @@ class ChannelRangeQueriesSpec extends FunSuite {
     assert(Router.computeFlag(channels)(ef.shortChannelId, None, None, false) === (INCLUDE_CHANNEL_ANNOUNCEMENT | INCLUDE_CHANNEL_UPDATE_1 | INCLUDE_CHANNEL_UPDATE_2))
     assert(Router.computeFlag(channels)(ef.shortChannelId, None, None, true) === (INCLUDE_CHANNEL_ANNOUNCEMENT | INCLUDE_CHANNEL_UPDATE_1 | INCLUDE_CHANNEL_UPDATE_2 | INCLUDE_NODE_ANNOUNCEMENT_1 | INCLUDE_NODE_ANNOUNCEMENT_2))
   }
+
+  test("split short channel ids correctly ") {
```
pm47: Why not start with a very simple test with tiny chunk sizes?

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +128,23 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+    def makeShortChannelIds(height: Int, count: Int): SortedSet[ShortChannelId] = SortedSet.empty[ShortChannelId] ++ (0 until count).map(c => ShortChannelId(height, c, 0))
+
+    val ids = makeShortChannelIds(42, 100) ++ makeShortChannelIds(43, 70) ++ makeShortChannelIds(44, 50) ++ makeShortChannelIds(45, 30) ++ makeShortChannelIds(50, 120)
+
+    // check that chunks do not overlap
+    def validate(chunks: List[ShortChannelIdsChunk]) : Boolean = chunks match {
```
pm47: this doesn't validate that we are not e.g. dropping or duplicating ids it seems

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +129,58 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+    }
+
+    // all ids in different blocks, chunk size == 2
+    {
+      val ids = List(ShortChannelId(1000, 0, 0), ShortChannelId(1001, 0, 0), ShortChannelId(1002, 0, 0), ShortChannelId(1003, 0, 0), ShortChannelId(1004, 0, 0), ShortChannelId(1005, 0, 0))
+      val chunks = Router.split(SortedSet.empty[ShortChannelId] ++ ids, 2)
+      assert(chunks == ShortChannelIdsChunk(1000, 2, List(ids(0), ids(1))) :: ShortChannelIdsChunk(1002, 2, List(ids(2), ids(3))) :: ShortChannelIdsChunk(1004, 2, List(ids(4), ids(5))) :: Nil)
+    }
+
+    // all ids in the same block
```
pm47: How about "all ids in same block, chunk size == 2"? This PR creates this very corner case we might as well test it?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
sstone: Because `Queue` is implemented with `List` so I don't think there would be a real gain here.

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
sstone: New extensive tests have been added

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1205,21 +1207,41 @@ object Router {
    * Have to split ids because otherwise message could be too big
    * there could be several reply_channel_range messages for a single query
    */
-  def split(shortChannelIds: SortedSet[ShortChannelId], channelRangeChunkSize: Int): List[ShortChannelIdsChunk] = {
-    // this algorithm can split blocks (meaning that we can in theory generate several replies with the same first_block/num_blocks
-    // and a different set of short_channel_ids) but it doesn't matter
-    if (shortChannelIds.isEmpty) {
-      List(ShortChannelIdsChunk(0, 0, List.empty))
+  def split(shortChannelIds: SortedSet[ShortChannelId], firstBlockNum: Long, numberOfBlocks: Long, channelRangeChunkSize: Int): List[ShortChannelIdsChunk] = {
+    val it = shortChannelIds.iterator.dropWhile(_.blockHeight < firstBlockNum).takeWhile(_.blockHeight < firstBlockNum + numberOfBlocks)
```
pm47: I like including RFC snippets for that kind of precise things, saves time for reviewers.

```suggestion
    // BOLT 7: MUST encode a short_channel_id [...] in blocks first_blocknum to first_blocknum + number_of_blocks - 1
    val it = shortChannelIds.iterator.dropWhile(_.blockHeight < firstBlockNum).takeWhile(_.blockHeight < firstBlockNum + numberOfBlocks)
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -642,8 +642,10 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
           val shortChannelIds: SortedSet[ShortChannelId] = d.channels.keySet.filter(keep(firstBlockNum, numberOfBlocks, _))
           log.info("replying with {} items for range=({}, {})", shortChannelIds.size, firstBlockNum, numberOfBlocks)
           val chunks = Kamon.runWithSpan(Kamon.spanBuilder("split-channel-ids").start(), finishSpan = true) {
-            split(shortChannelIds, nodeParams.routerConf.channelRangeChunkSize)
+            split(shortChannelIds, firstBlockNum, numberOfBlocks, nodeParams.routerConf.channelRangeChunkSize)
           }
+
+          // we make sure that our replies cover the range specified in their queries
```
t-bast: This comment seems out of place?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1014,7 +1016,7 @@ object Router {
    */
   def keep(firstBlockNum: Long, numberOfBlocks: Long, id: ShortChannelId): Boolean = {
     val TxCoordinates(height, _, _) = ShortChannelId.coordinates(id)
```
t-bast: nit: we could use the new `blockHeight` here instead of the whole `coordinates`

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +129,163 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+
+  def validate(chunk: ShortChannelIdsChunk) = {
+    require(!chunk.shortChannelIds.exists(!Router.keep(chunk.firstBlock, chunk.numBlocks, _)))
+  }
+
+  // check that chunks do not overlap and contain exactly the ids they were built from
+  def validate(ids: SortedSet[ShortChannelId], firstBlockNum: Long, numberOfBlocks: Long, chunks: List[ShortChannelIdsChunk]): Unit = {
+
+    @tailrec
+    def noOverloap(chunks: List[ShortChannelIdsChunk]): Boolean = chunks match {
```
t-bast: nit:

```suggestion
    def noOverlap(chunks: List[ShortChannelIdsChunk]): Boolean = chunks match {
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +129,163 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+      case Nil => true
+      case a :: b :: _ if b.firstBlock < a.firstBlock + a.numBlocks => false
+      case _ => noOverloap(chunks.tail)
+    }
+
+    // aggregate ids from all chunks, to check that they match our input ids exactly
+    val chunkIds = SortedSet.empty[ShortChannelId] ++ chunks.map(_.shortChannelIds).flatten.toSet
+    val expected = ids.filter(Router.keep(firstBlockNum, numberOfBlocks, _))
+
+    if (expected.isEmpty) require(chunks == List(ShortChannelIdsChunk(firstBlockNum, numberOfBlocks, Nil)))
```
t-bast: This looks redundant with the tests below

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +129,163 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+  test("split short channel ids correctly (basic tests") {
+    // no ids to split
+    {
+      val ids = Nil
+      val firstBlockNum = 10
+      val numberOfBlocks = 100
+      val chunks = Router.split(SortedSet.empty[ShortChannelId] ++ ids, firstBlockNum, numberOfBlocks, ids.size)
+      assert(chunks == ShortChannelIdsChunk(firstBlockNum, numberOfBlocks, Nil) :: Nil)
+    }
+    assert(Router.split(SortedSet.empty[ShortChannelId], 100, 1000, 100) == ShortChannelIdsChunk(100, 1000, Nil) :: Nil)
```
t-bast: This is redundant with the test above?

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +129,163 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+    val numberOfBlocks = 1000
+
+    validate(ids, firstBlockNum, numberOfBlocks, Router.split(ids, firstBlockNum, numberOfBlocks, 1))
+    validate(ids, firstBlockNum, numberOfBlocks, Router.split(ids, firstBlockNum, numberOfBlocks, 20))
+    validate(ids, firstBlockNum, numberOfBlocks, Router.split(ids, firstBlockNum, numberOfBlocks, 50))
+    validate(ids, firstBlockNum, numberOfBlocks, Router.split(ids, firstBlockNum, numberOfBlocks, 100))
+    validate(ids, firstBlockNum, numberOfBlocks, Router.split(ids, firstBlockNum, numberOfBlocks, 1000))
+  }
+
+  test("fuzzy test") {
```
t-bast: It's not really a fuzzy test without randomness...

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +129,163 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+    {
+      val ids = Nil
+      val firstBlockNum = 10
+      val numberOfBlocks = 100
+      val chunks = Router.split(SortedSet.empty[ShortChannelId] ++ ids, firstBlockNum, numberOfBlocks, ids.size)
+      assert(chunks == ShortChannelIdsChunk(firstBlockNum, numberOfBlocks, Nil) :: Nil)
+    }
+    assert(Router.split(SortedSet.empty[ShortChannelId], 100, 1000, 100) == ShortChannelIdsChunk(100, 1000, Nil) :: Nil)
+
+    // ids are all before the requested range
```
pm47:
```suggestion
    // ids are all after the requested range
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +129,163 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+    {
+      val ids = List(ShortChannelId(1000, 0, 0), ShortChannelId(1001, 0, 0), ShortChannelId(1002, 0, 0), ShortChannelId(1003, 0, 0), ShortChannelId(1004, 0, 0), ShortChannelId(1005, 0, 0))
+      val firstBlockNum = 10
+      val numberOfBlocks = 100
+      val chunks = Router.split(SortedSet.empty[ShortChannelId] ++ ids, firstBlockNum, numberOfBlocks, ids.size)
+      assert(chunks == ShortChannelIdsChunk(firstBlockNum, numberOfBlocks, Nil) :: Nil)
+    }
+
+    // ids are all after the requested range
```
pm47:
```suggestion
    // ids are all before the requested range
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +129,163 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+
+    if (expected.isEmpty) require(chunks == List(ShortChannelIdsChunk(firstBlockNum, numberOfBlocks, Nil)))
+    chunks.foreach(validate)
+    require(chunks.head.firstBlock == firstBlockNum)
+    require(chunks.last.firstBlock + chunks.last.numBlocks == firstBlockNum + numberOfBlocks)
+    require(chunkIds == expected)
+    require(noOverloap(chunks))
+  }
+
+  test("split short channel ids correctly (basic tests") {
```
pm47: Defining this helper method:
```scala
def id(blockHeight: Int, txIndex: Int = 0, outputIndex: Int = 0) = ShortChannelId(blockHeight, txIndex, outputIndex)
```

allows for significant less verbosity

pm47: NB: particularly because we can remove the `txIndex`/`outputIndex` when they are not overriden.

pm47: By the way I think we are missing tests splitting on `outputIndex`?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1205,21 +1207,41 @@ object Router {
    * Have to split ids because otherwise message could be too big
    * there could be several reply_channel_range messages for a single query
    */
-  def split(shortChannelIds: SortedSet[ShortChannelId], channelRangeChunkSize: Int): List[ShortChannelIdsChunk] = {
-    // this algorithm can split blocks (meaning that we can in theory generate several replies with the same first_block/num_blocks
-    // and a different set of short_channel_ids) but it doesn't matter
-    if (shortChannelIds.isEmpty) {
-      List(ShortChannelIdsChunk(0, 0, List.empty))
+  def split(shortChannelIds: SortedSet[ShortChannelId], firstBlockNum: Long, numberOfBlocks: Long, channelRangeChunkSize: Int): List[ShortChannelIdsChunk] = {
```
pm47: I think there should be scaladoc for the args, in order to precisely know what this method does (particularly for `channelRangeChunkSize`).

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +129,163 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+      assert(chunks == ShortChannelIdsChunk(firstBlockNum, numberOfBlocks, ids) :: Nil)
+    }
+
+    // all ids in the same block, chunk size == 2
+    {
+      val ids = List(ShortChannelId(1000, 0, 0), ShortChannelId(1000, 1, 0), ShortChannelId(1000, 2, 0), ShortChannelId(1000, 3, 0), ShortChannelId(1000, 4, 0), ShortChannelId(1000, 5, 0))
+      val firstBlockNum = 900
+      val numberOfBlocks = 200
+      val chunks = Router.split(SortedSet.empty[ShortChannelId] ++ ids, firstBlockNum, numberOfBlocks, 2)
+      assert(chunks == ShortChannelIdsChunk(firstBlockNum, numberOfBlocks, ids) :: Nil)
```
pm47: So, in this test, we are not respecting our own `channelRangeChunkSize` on purpose right?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1199,30 +1209,77 @@ object Router {
...
+
+      val first = it.next()
+      loop(first :: Nil, Nil)
     }
   }
 
+  /**
+   * Enforce max-size constraints for each chunk
+   * @param chunks list of short channel id chunks
+   * @param keepFirst flags that specifies wether to keep to first of last ids when there are too many of them
```
t-bast: nit:

```suggestion
   * @param keepFirst flags that specifies whether to keep first of last ids when there are too many of them
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
t-bast: I still think this statement is unnecessary: the following lines re-test that (even for the case where `expected.isEmpty`)

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +131,226 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+      } else {
+        outputIndex = outputIndex + 1
+      }
+      output += ShortChannelId(height, txIndex, outputIndex)
+    }
+    output.toList
+  }
+
+  def validate(chunk: ShortChannelIdsChunk) = {
+    require(!chunk.shortChannelIds.exists(!Router.keep(chunk.firstBlock, chunk.numBlocks, _)))
```
t-bast: nit: a bit more readable imo:

```suggestion
    require(chunk.shortChannelIds.forall(Router.keep(chunk.firstBlock, chunk.numBlocks, _)))
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +131,226 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+
+    @tailrec
+    def noOverlap(chunks: List[ShortChannelIdsChunk]): Boolean = chunks match {
+      case Nil => true
+      case a :: b :: _ if b.firstBlock < a.firstBlock + a.numBlocks => false
+      case _ => noOverlap(chunks.tail)
+    }
+
+    // aggregate ids from all chunks, to check that they match our input ids exactly
+    val chunkIds = SortedSet.empty[ShortChannelId] ++ chunks.map(_.shortChannelIds).flatten.toSet
```
t-bast: nit:

```suggestion
    val chunkIds = SortedSet.empty[ShortChannelId] ++ chunks.flatMap(_.shortChannelIds).toSet
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +131,226 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+    }
+
+    // all ids in the same block
+    {
+      val ids = makeShortChannelIds(1000, 100)
+      val firstBlockNum = 900
+      val numberOfBlocks = 200
+      val chunks = Router.split(SortedSet.empty[ShortChannelId] ++ ids, firstBlockNum, numberOfBlocks, 10)
+      assert(chunks == ShortChannelIdsChunk(firstBlockNum, numberOfBlocks, ids) :: Nil)
+    }
```
t-bast: I tried to add the following test and it doesn't pass. I think we should change the code to make it pass :)

```scala
    // don't start a new block in same chunk if it overflows chunk size
    {
      val ids = List(id(1000), id(1000, 1), id(1001), id(1001, 1))
      val firstBlockNum = 900
      val numberOfBlocks = 200
      val chunks = Router.split(SortedSet.empty[ShortChannelId] ++ ids, firstBlockNum, numberOfBlocks, 3)
      assert(chunks == ShortChannelIdsChunk(firstBlockNum, 100 + 1, List(ids(0), ids(1))) :: ShortChannelIdsChunk(1001, 1, List(ids(2), ids(3))) :: Nil)
    }
```

When we decide to start adding a new block to the current chunk, we should only do that if the whole block can fit in the chunk. Otherwise we should take the safer choice of starting a new chunk.
If we don't do that it's likely we'll run more often in the code that does the random split to avoid the 3200 elements limit which I think we should try to avoid (it's not really a problem to send a bit more chunks imo).

sstone: Yes the current "merge" algo is simplistic (but also very efficient). Being smarter would require to iterate twice and computing a `height -> count` map.
I propose instead to change the default `channel-range-chunk-size` to 1500 instead of 2500, it should reduces the probability of merging 2 chunks and ending up over the limit

t-bast: SGTM

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
sstone: Yes but I'd like to keep it as intent here is really obvious

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1199,30 +1207,83 @@ object Router {
...
+     *
+     * @param maximumSize maximum size of the short channel ids list
+     * @return a chunk with at most `maximumSize` ids
+     */
+    def enforceMaximumSize(maximumSize: Int) = {
+      if (shortChannelIds.size <= maximumSize) this else {
+        // we use a random offset here, so even if shortChannelIds.size is much bigger than maximumSize (which should
+        // not happen) peers will eventually receive info about all channels in this chunk
+        val offset = Random.nextInt(shortChannelIds.size - maximumSize)
+        this.copy(shortChannelIds = this.shortChannelIds.drop(offset).take(maximumSize))
```
```suggestion
        this.copy(shortChannelIds = this.shortChannelIds.slice(offset, offset + maximumSize))
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1199,30 +1207,83 @@ object Router {
...
+    /**
+     *
+     * @param maximumSize maximum size of the short channel ids list
+     * @return a chunk with at most `maximumSize` ids
+     */
+    def enforceMaximumSize(maximumSize: Int) = {
+      if (shortChannelIds.size <= maximumSize) this else {
+        // we use a random offset here, so even if shortChannelIds.size is much bigger than maximumSize (which should
+        // not happen) peers will eventually receive info about all channels in this chunk
+        val offset = Random.nextInt(shortChannelIds.size - maximumSize)
```
pm47: off-by-one bug (reproduced with test provided below)

```suggestion
        val offset = Random.nextInt(shortChannelIds.size - maximumSize + 1)
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +131,215 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+    val expected = ids.filter(Router.keep(firstBlockNum, numberOfBlocks, _))
+
+    if (expected.isEmpty) require(chunks == List(ShortChannelIdsChunk(firstBlockNum, numberOfBlocks, Nil)))
+    chunks.foreach(validate)
+    require(chunks.head.firstBlock == firstBlockNum)
+    require(chunks.last.firstBlock + chunks.last.numBlocks == firstBlockNum + numberOfBlocks)
+    require(chunkIds == expected)
+    require(noOverlap(chunks))
+  }
+
```
pm47:
```suggestion

    test("limit channel ids chunk size") {
    val ids = makeShortChannelIds(1, 3)
    val chunk = ShortChannelIdsChunk(0, 10, ids)

    val res1 = for (_ <- 0 until 100) yield chunk.enforceMaximumSize(1).shortChannelIds
    assert(res1.toSet == Set(List(ids(0)), List(ids(1)), List(ids(2))))

    val res2 = for (_ <- 0 until 100) yield chunk.enforceMaximumSize(2).shortChannelIds
    assert(res2.toSet == Set(List(ids(0), ids(1)), List(ids(1), ids(2))))

    val res3 = for (_ <- 0 until 100) yield chunk.enforceMaximumSize(3).shortChannelIds
    assert(res3.toSet == Set(List(ids(0), ids(1), ids(2))))
  }
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -127,4 +131,228 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+    val res1 = for (_ <- 0 until 100) yield chunk.enforceMaximumSize(1).shortChannelIds
+    assert(res1.toSet == Set(List(ids(0)), List(ids(1)), List(ids(2))))
+
+    val res2 = for (_ <- 0 until 100) yield chunk.enforceMaximumSize(2).shortChannelIds
+    assert(res2.toSet == Set(List(ids(0), ids(1)), List(ids(1), ids(2))))
+
+    val res3 = for (_ <- 0 until 100) yield chunk.enforceMaximumSize(3).shortChannelIds
+    assert(res3.toSet == Set(List(ids(0), ids(1), ids(2))))
+  }
+  test("split short channel ids correctly (basic tests") {
```
pm47:
```suggestion

  test("split short channel ids correctly (basic tests") {
```

PR #1275, opened 2020-01-10
Don't put the encoding type if there is no channel

pm47:
If the list of channel ids is empty, we don't encode anything (as
opposed as indicating an encoding type without any data).

This fixes a compatibility issue with lnd.

TODO: refine the spec

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/wire/ExtendedQueriesCodecsSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/wire/ExtendedQueriesCodecsSpec.scala
@@ -22,9 +22,29 @@ import fr.acinq.eclair.wire.LightningMessageCodecs._
 import fr.acinq.eclair.wire.ReplyChannelRangeTlv._
 import fr.acinq.eclair.{CltvExpiryDelta, LongToBtcAmount, ShortChannelId, UInt64}
 import org.scalatest.FunSuite
-import scodec.bits.ByteVector
+import scodec.bits.{ByteVector, _}
 
 class ExtendedQueriesCodecsSpec extends FunSuite {
+
+  test("encode a list of short channel ids") {
+    {
+      // empty list
+      val ids = EncodedShortChannelIds(EncodingType.UNCOMPRESSED, List.empty)
+      val encoded = encodedShortChannelIdsCodec.encode(ids).require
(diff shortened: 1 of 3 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/wire/ExtendedQueriesCodecsSpec.scala
```diff
@@ -22,9 +22,29 @@ import fr.acinq.eclair.wire.LightningMessageCodecs._
 import fr.acinq.eclair.wire.ReplyChannelRangeTlv._
 import fr.acinq.eclair.{CltvExpiryDelta, LongToBtcAmount, ShortChannelId, UInt64}
 import org.scalatest.FunSuite
-import scodec.bits.ByteVector
+import scodec.bits.{ByteVector, _}
```
t-bast:
```suggestion
import scodec.bits._
```

pm47: Fixed as part of my rebase

t-bast:
I think it's worth adding a test that verifies that we can still decode if it's encoded "the old way" (len = `0001` and encoding = `00` or `01` without any short channel ids afterwards)

pm47:
> I think it's worth adding a test that verifies that we can still decode if it's encoded "the old way" (len = 0001 and encoding = 00 or 01 without any short channel ids afterwards)

Good point!

t-bast:
Aaand it looks like c-lightning doesn't support that change either...we'll have to wait for Rusty to come back and ask him if they plan on doing that change or not.

pm47:
The more I think about it the less I like this change.

t-bast:
What @sstone suggests is to change our behavior to always use the uncompressed encoding if the list is empty: that avoids the lnd bug without doing this bigger encoding change that isn't supported by other implementations.

pm47:
> What @sstone suggests is to change our behavior to always use the uncompressed encoding if the list is empty

That's what I've done:
https://github.com/ACINQ/eclair/blob/948bb19a19b0ac58daa0d1fd7864cbd47b5c3e91/eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala#L237

pm47:
Superseded by #1276.

PR #1278, opened 2020-01-15
Turn off SSL when custom Electrum server address is a hidden service

CandleHater:
It would be great if I could connect the Eclair mobile wallet to my Tor only electrs server using Orbot. Sadly I can't because it requires a valid SSL certificate. Please make an exception if the custom Electrum server address is an onion domain.

As suggested from others in my initial PR https://github.com/ACINQ/eclair-mobile/pull/229, I created this PR to cover up eclair-core and not only eclair-mobile.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -182,15 +182,11 @@ class Setup(datadir: File,
           val host = config.getString("electrum.host")
           val port = config.getInt("electrum.port")
           val address = InetSocketAddress.createUnresolved(host, port)
-          val ssl = if (address.getHostName.endsWith(".onion")) {
-            // Tor already adds end-to-end encryption, adding TLS on top doesn't add anything
-            SSL.OFF
-          } else {
-            config.getString("electrum.ssl") match {
+          val ssl = config.getString("electrum.ssl") match {
+              case _ if address.getHostName.endsWith(".onion") => SSL.OFF // Tor already adds end-to-end encryption, adding TLS on top doesn't add anything
               case "off" => SSL.OFF
               case "loose" => SSL.LOOSE
               case _ => SSL.STRICT // strict mode is the default when we specify a custom electrum server, we don't want to be MITMed
-            }
           }
 
           logger.info(s"override electrum default with server=$address ssl=$ssl")
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -182,15 +182,15 @@ class Setup(datadir: File,
           val host = config.getString("electrum.host")
           val port = config.getInt("electrum.port")
           val address = InetSocketAddress.createUnresolved(host, port)
-          val ssl = config.getString("electrum.ssl") match {
-            case "off" => SSL.OFF
-            case "loose" => SSL.LOOSE
-            case _ => SSL.STRICT // strict mode is the default when we specify a custom electrum server, we don't want to be MITMed
-          }
-
-          // turn SSL verification off for a Tor hidden service
-          if (address.getHostName().endsWith(".onion")) {
-            ssl = SSL.OFF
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -181,12 +181,18 @@ class Setup(datadir: File,
...
+          val address = InetSocketAddress.createUnresolved(host, port)
           val ssl = config.getString("electrum.ssl") match {
             case "off" => SSL.OFF
             case "loose" => SSL.LOOSE
             case _ => SSL.STRICT // strict mode is the default when we specify a custom electrum server, we don't want to be MITMed
           }
-          val address = InetSocketAddress.createUnresolved(host, port)
+
+          // turn SSL verification off for a Tor hidden service
+          if (address.getHostName().endsWith(".onion")) {
```
t-bast: Here is what I suggest:

```scala
val host = config.getString("electrum.host")
val port = config.getInt("electrum.port")
val address = InetSocketAddress.createUnresolved(host, port)
val ssl = if (address.getHostName.endsWith(".onion")) {
  // Tor already adds end-to-end encryption, so adding TLS on top doesn't add anything (apart from showing a
  // valid certificate which only makes sense for websites or paid APIs where users verify the certificate).
  SSL.OFF
} else {
  config.getString("electrum.ssl") match {
    case "off" => SSL.OFF
    case "loose" => SSL.LOOSE
    case _ => SSL.STRICT // strict mode is the default when we specify a custom electrum server, we don't want to be MITMed
  }
}
logger.info(s"override electrum default with server=$address ssl=$ssl")
Set(ElectrumServerAddress(address, ssl))
```

CandleHater:
Thank you, I missed that. Made a new commit to address this issue.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
pm47: Another suggestion (more compact and carries the intent a bit better I think):

```scala
val ssl = config.getString("electrum.ssl") match {
  case _ if address.getHostName.endsWith(".onion") => SSL.OFF // Tor already adds end-to-end encryption, adding TLS on top doesn't add anything
  case "off" => SSL.OFF
  case "loose" => SSL.LOOSE
  case _ => SSL.STRICT // strict mode is the default when we specify a custom electrum server, we don't want to be MITMed
}
```

t-bast: Once we settle on the code, @sstone said he could test it E2E before we merge.

CandleHater: Thank you @pm47, great change! Just applied it https://github.com/ACINQ/eclair/pull/1278/commits/233bcdae3027e543e4b4f03afb8ac996b62c38dd.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -181,12 +181,14 @@ class Setup(datadir: File,
         case true =>
           val host = config.getString("electrum.host")
           val port = config.getInt("electrum.port")
+          val address = InetSocketAddress.createUnresolved(host, port)
           val ssl = config.getString("electrum.ssl") match {
-            case "off" => SSL.OFF
-            case "loose" => SSL.LOOSE
-            case _ => SSL.STRICT // strict mode is the default when we specify a custom electrum server, we don't want to be MITMed
+              case _ if address.getHostName.endsWith(".onion") => SSL.OFF // Tor already adds end-to-end encryption, adding TLS on top doesn't add anything
```
Kixunil: Shouldn't this be conditioned on `!"strict"`?

CandleHater: What do you mean, prevent ``SSL.OFF`` in case someone explicitly sets strict SSL verification while using an hidden service?

Kixunil: Yes, exactly. I think explicit setting should always override anything else.

PR #1279, opened 2020-01-16
Make sure that ids chunks strictly cover the requested range, without gaps

sstone:
lnd expects ids ranges in reply_channel_range messages to strictly follow each other, without gaps.
For example, using block heights and not ids, [1,2,4,5] would be split into (first=1, num=2, [1,2]) :: (first=3, num=2, [4, 5])

This is arguably a limitation of lnd (c-lightning does not requires this and it's not needed to properly process replies) but is easy to implement.


pm47:
NB: this needs to be added to the spec

PR #1280, opened 2020-01-16
Do not include optional timestamp/checksum TLV if they are empty

sstone:
This change fixes an interop issue with c-lighting, which does not handle `reply_channel_range` messages that include an option `timestamps` TLV that contains an empty array of timestamps with a `ZLIB` encoding` and fails with a `connection-level error, failing all channels! reason=reply_channel_range 0 timestamps when 0 scids?` error

It is related to https://github.com/lightningnetwork/lightning-rfc/pull/729

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -648,7 +648,7 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
           Kamon.runWithSpan(Kamon.spanBuilder("compute-timestamps-checksums").start(), finishSpan = true) {
             chunks.foreach { chunk =>
               val (timestamps, checksums) = routingMessage.queryFlags_opt match {
-                case Some(extension) if extension.wantChecksums | extension.wantTimestamps =>
+                case Some(extension) if chunk.shortChannelIds.nonEmpty && (extension.wantChecksums | extension.wantTimestamps) =>
                   // we always compute timestamps and checksums even if we don't need both, overhead is negligible
                   val (timestamps, checksums) = chunk.shortChannelIds.map(getChannelDigestInfo(d.channels)).unzip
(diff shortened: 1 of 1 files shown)


sstone:
I'll make it ready for review as soon as I've added proper tests

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1277,6 +1270,32 @@ object Router {
     }
   }
 
+  /**
+   * Compute extended channel information, which currently covers:
+   * - channel update timestamps
+   * - channel update checksums
+   * @param shortChannelIds list of short channel ids
+   * @param channels channel map
+   * @param queryFlags_opt query flag, which specifies what information has been request
```
t-bast: nit:

```suggestion
   * @param queryFlags_opt query flag, which specifies what information has been requested
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRangeQueriesSpec.scala
```diff
@@ -356,4 +360,37 @@ class ChannelRangeQueriesSpec extends FunSuite {
...
+          require(checksums.isEmpty)
+        case _ if ids.isEmpty =>
+          require(timestamps.isEmpty)
+          require(checksums.isEmpty)
+        case Some(extension) =>
+          if (extension.wantTimestamps) {
+            require(timestamps.map(_.timestamps.size) == Some(ids.size))
+            require(timestamps.map(_.encoding) == Some(encodingType))
+          }
+          if (extension.wantChecksums) require(checksums.map(_.checksums.size) == Some(ids.size))
```
t-bast: Maybe also check that if `!extension.wantTimestamps` timestamps aren't included, and if `!extension.wantChecksums` checksums aren't included?

sstone:
Test against c-lighting (at https://github.com/ElementsProject/lightning/commit/4be1868b8a603b1bd475400b57bc0531719872e2) and lnd (https://github.com/lightningnetwork/lnd/releases/tag/v0.9.0-beta-rc2)

pm47:
I think previous implementation made more sense. How can I tell now that my peer understood that I requested additional data, but just doesn't have any?

PR #1286, opened 2020-01-21
Upfront shutdown script

t-bast:
Extending all messages with a TLV stream (spec PR [here](https://github.com/lightningnetwork/lightning-rfc/pull/714)) requires making currently optional data to become mandatory.

We don't support `upfront_shutdown_script` yet which is going to be a problem if we want to extend `open_channel` and `accept_channel` with additional TLVs (which we already do for Phoenix).

This PR implements partial support for `upfront_shutdown_script`: we allow our peers to use that feature, but there's no hooks to use it ourselves. If our users want us to add support for specifying an `upfront_shutdown_script`, we can add it later.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
@@ -89,7 +89,8 @@ object LightningMessageCodecs {
       ("htlcBasepoint" | publicKey) ::
       ("firstPerCommitmentPoint" | publicKey) ::
       ("channelFlags" | byte) ::
-      ("upfront_shutdown_script" | upfrontShutdownScript)).as[OpenChannel]
+      ("upfront_shutdown_script" | upfrontShutdownScript) ::
+      ("tlvStream_opt" | optional(bitsRemaining, OpenTlv.openTlvCodec))).as[OpenChannel]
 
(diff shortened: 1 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -59,6 +59,17 @@ object LightningMessageCodecs {
       ("yourLastPerCommitmentSecret" | optional(bitsRemaining, privateKey)) ::
       ("myCurrentPerCommitmentPoint" | optional(bitsRemaining, publicKey))).as[ChannelReestablish]
 
+  // Legacy nodes are supposed to encode an upfront_shutdown_script only if both sides advertised support for option_upfront_shutdown_script.
```
sstone: I think that a node may include an optional shutdown script even if its peer does not support it:

>   - if both nodes advertised the `option_upfront_shutdown_script` feature:
>    - MUST include either a valid `shutdown_scriptpubkey` as required by `shutdown` >`scriptpubkey`, or a zero-length `shutdown_scriptpubkey`.
 > - otherwise:
  >  - MAY include a`shutdown_scriptpubkey`.

sstone:
I think there is something missing: if both peers support `option_upfront_shutdown_script` then we must include one (at least an empty one) ?

PR #1287, opened 2020-01-22
Trampoline/MPP DB changes

t-bast:
This PR refactors payment structures and DB schema to correctly handle Trampoline and MPP.
A notable change is the distinction between a target node and the final recipient.
To correctly report fees, we need to store information about the recipient, even though
we're sending a payment to the first trampoline node (our target node).

It can quickly become confusing because there are multiple amounts that appear during the payment:

* the final amount (amount from the invoice, what the final recipient should receive)
* to which we add trampoline fees to find the amount the first trampoline node needs to receive
* from which we can compute the amount of the first HTLC once we've found a route to the first trampoline node (and know the associated routing fees)

I tried to make the wording uniform across the codebase to make sure we don't get those mixed up.
Those changes allow us to clean multiple temporary hacks made to support Trampoline/MPP in the first version of Phoenix.

There were also multiple places where we weren't consistent when setting parentId vs id in payment events: this is now fixed.

If you think there are other columns that can be added/changed in the PaymentsDb or AuditDb, I think it's worth adding them now to minimize migrations. 

This is a big PR, but probably one of the most important ones for MPP/Trampoline support, so please take your time to review and challenge.
Even comments such as: "this part is a bit messy, please rework" are useful :).

In a follow-up PR, I will update the API to make it easier to work with MPP and Trampoline from the CLI.

E2E tests:

- [x] Test migrating the Endurance DB
- [x] Setup nodes, make various types of payments, then upgrade their DB (will be done once the PR is approved, just before merging)


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -35,19 +35,41 @@ sealed trait PaymentEvent {
...
-  val amount: MilliSatoshi = parts.map(_.amount).sum
-  val feesPaid: MilliSatoshi = parts.map(_.feesPaid).sum
+/**
+ * A payment was successfully sent and fulfilled.
+ *
+ * @param id              id of the whole payment attempt (if using multi-part, there will be multiple parts, each with
+ *                        a different id).
+ * @param paymentHash     payment hash.
+ * @param paymentPreimage payment preimage (proof of payment).
+ * @param finalAmount     amount that has been received by the final recipient.
```
pm47: Maybe`recipientAmount` / `recipientNodeId` or `finalAmount` / `finalNodeId` for consistency?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -35,19 +35,41 @@ sealed trait PaymentEvent {
...
+ * @param recipientNodeId id of the final recipient.
+ * @param parts           child payments (actual outgoing HTLCs).
+ */
+case class PaymentSent(id: UUID, paymentHash: ByteVector32, paymentPreimage: ByteVector32, finalAmount: MilliSatoshi, recipientNodeId: PublicKey, parts: Seq[PaymentSent.PartialPayment]) extends PaymentEvent {
+  require(parts.nonEmpty, "must have at least one sub-payment")
+  val amountWithFees: MilliSatoshi = parts.map(_.amountWithFees).sum
+  val feesPaid: MilliSatoshi = amountWithFees - finalAmount
```
pm47:
```suggestion
  val feesPaid: MilliSatoshi = amountWithFees - finalAmount // overall fees for this payment (routing + trampoline)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -35,19 +35,41 @@ sealed trait PaymentEvent {
...
+ *                        a different id).
+ * @param paymentHash     payment hash.
+ * @param paymentPreimage payment preimage (proof of payment).
+ * @param finalAmount     amount that has been received by the final recipient.
+ * @param recipientNodeId id of the final recipient.
+ * @param parts           child payments (actual outgoing HTLCs).
+ */
+case class PaymentSent(id: UUID, paymentHash: ByteVector32, paymentPreimage: ByteVector32, finalAmount: MilliSatoshi, recipientNodeId: PublicKey, parts: Seq[PaymentSent.PartialPayment]) extends PaymentEvent {
+  require(parts.nonEmpty, "must have at least one sub-payment")
+  val amountWithFees: MilliSatoshi = parts.map(_.amountWithFees).sum
```
pm47: Have you considered using `lazy val`s here?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -35,19 +35,41 @@ sealed trait PaymentEvent {
...
+  val feesPaid: MilliSatoshi = amountWithFees - finalAmount
+  val trampolineFees: MilliSatoshi = parts.map(_.amount).sum - finalAmount
+  val nonTrampolineFees: MilliSatoshi = parts.map(_.feesPaid).sum
```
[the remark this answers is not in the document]
t-bast: We could! It's probably more intuitive, I'll change that.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
t-bast: I hesitated, but since they don't take long to compute, don't use a lot of space, I felt it was a bit too early optimization.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -119,22 +127,26 @@ object IncomingPaymentStatus {
...
+ * @param recipientNodeId id of the final recipient.
+ * @param createdAt       absolute time in milli-seconds since UNIX epoch when the payment was created.
+ * @param paymentRequest  Bolt 11 payment request (if paying from an invoice).
+ * @param status          current status of the payment.
  */
 case class OutgoingPayment(id: UUID,
                            parentId: UUID,
                            externalId: Option[String],
                            paymentHash: ByteVector32,
+                           paymentType: Option[String],
```
pm47: This needs to be made mandatory imo, which may require a migration but makes future but sets the record clean.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -33,24 +33,29 @@ import scala.util.Try
...
+ * @param amount             amount to pay (empty string means no amount is specified)
+ * @param timestamp          request timestamp (UNIX format)
+ * @param nodeId             id of the node emitting the payment request
+ * @param tags               payment tags; must include a single PaymentHash tag and a single PaymentSecret tag.
+ * @param signature          request signature that will be checked against node id
+ * @param verifyFeatureGraph if false, doesn't validate that the feature bits are spec-compliant. This may be useful for
+ *                           backwards-compatibility with invoices that were generated before the spec made this mandatory.
  */
-case class PaymentRequest(prefix: String, amount: Option[MilliSatoshi], timestamp: Long, nodeId: PublicKey, tags: List[PaymentRequest.TaggedField], signature: ByteVector) {
+case class PaymentRequest(prefix: String, amount: Option[MilliSatoshi], timestamp: Long, nodeId: PublicKey, tags: List[PaymentRequest.TaggedField], signature: ByteVector)(verifyFeatureGraph: Boolean) {
```
pm47:
```suggestion
case class PaymentRequest(prefix: String, amount: Option[MilliSatoshi], timestamp: Long, nodeId: PublicKey, tags: List[PaymentRequest.TaggedField], signature: ByteVector) {
```

This is really a workaround for compatibility because we used to enable the MPP feature bit without the  `var_onion_optin`, which is now illegal. I would rather disable the feature bits dependency check for this particular feature, it has a low impact and saves us from circumvoluted changes in the codebase

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -37,34 +37,63 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
       case 2 =>
         logger.warn(s"migrating db $DB_NAME, found version=2 current=$CURRENT_VERSION")
         migration23(statement)
+        migration34(statement)
+        setVersion(statement, DB_NAME, CURRENT_VERSION)
+      case 3 =>
+        logger.warn(s"migrating db $DB_NAME, found version=3 current=$CURRENT_VERSION")
+        migration34(statement)
         setVersion(statement, DB_NAME, CURRENT_VERSION)
       case CURRENT_VERSION =>
```
dpad85: How about using triple quotes for this kind of queries?

Example:

```scala
statement.executeUpdate("""
  CREATE TABLE IF NOT EXISTS sent (
    amount_msat INTEGER NOT NULL, 
    fees_msat INTEGER NOT NULL, 
    final_amount_msat INTEGER NOT NULL,
    payment_id TEXT NOT NULL, -- renamed v4
    parent_payment_id TEXT NOT NULL, 
    payment_hash BLOB NOT NULL, 
    payment_preimage BLOB NOT NULL, 
    recipient_node_id BLOB NOT NULL,
    to_channel_id BLOB NOT NULL, 
    timestamp INTEGER NOT NULL
)""")
```

Of course it would make the database files feel more verbose, but it would also help understand what a table looks like right now, and how it has changed when doing a git `diff`.

pm47: FWIW, having it all in one line allows quick checking that the table declaration in the migration function is identical to the one in the "regular" path.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -35,19 +35,41 @@ sealed trait PaymentEvent {
...
+ * @param paymentPreimage payment preimage (proof of payment).
+ * @param finalAmount     amount that has been received by the final recipient.
+ * @param recipientNodeId id of the final recipient.
+ * @param parts           child payments (actual outgoing HTLCs).
+ */
+case class PaymentSent(id: UUID, paymentHash: ByteVector32, paymentPreimage: ByteVector32, finalAmount: MilliSatoshi, recipientNodeId: PublicKey, parts: Seq[PaymentSent.PartialPayment]) extends PaymentEvent {
+  require(parts.nonEmpty, "must have at least one sub-payment")
+  val amountWithFees: MilliSatoshi = parts.map(_.amountWithFees).sum
+  val feesPaid: MilliSatoshi = amountWithFees - finalAmount
+  val trampolineFees: MilliSatoshi = parts.map(_.amount).sum - finalAmount
```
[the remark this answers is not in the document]
t-bast: Not exactly, those are the fees for all trampoline nodes.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -57,14 +79,27 @@ case class PaymentFailed(id: UUID, paymentHash: ByteVector32, failures: Seq[Paym
...
 
-case class TrampolinePaymentRelayed(amountIn: MilliSatoshi, amountOut: MilliSatoshi, paymentHash: ByteVector32, toNodeId: PublicKey, fromChannelIds: Seq[ByteVector32], toChannelIds: Seq[ByteVector32], timestamp: Long = Platform.currentTime) extends PaymentRelayed
+case class TrampolinePaymentRelayed(paymentHash: ByteVector32, incoming: PaymentRelayed.Incoming, outgoing: PaymentRelayed.Outgoing, timestamp: Long = Platform.currentTime) extends PaymentRelayed {
+  override val amountIn = incoming.map(_.amount).sum
+  override val amountOut = outgoing.map(_.amount).sum
+}
+
+object PaymentRelayed {
+
+  case class Part(amount: MilliSatoshi, channelId: ByteVector32)
```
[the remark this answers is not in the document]
t-bast: Hopefully long enough :D

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: I don't have a strong opinion on that.
The way I saw it was that the code in the `CURRENT_VERSION` didn't need to mention what changed: in most cases you're only interested in what the schema is now.
If you're interested in what the schema was before, then you look at migration code and tests.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
t-bast: Fixed in `18a34ae`

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
t-bast: Fixed in `3b6f468`

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -210,13 +229,15 @@ trait PaymentsOverviewDb {
...
   val paymentHash: ByteVector32
+  val paymentType: Option[String]
   val paymentRequest: Option[String]
   val finalAmount: Option[MilliSatoshi]
   val createdAt: Long
   val completedAt: Option[Long]
 }
 
 case class PlainIncomingPayment(paymentHash: ByteVector32,
+                                paymentType: Option[String],
```
pm47: couldn't we make `paymentType` required and manage compatibility within the db migration script (e.g. based on the description)?

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -175,61 +208,94 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
+            rs.getByteVector32("payment_hash"),
+            rs.getByteVector32("payment_preimage"),
+            MilliSatoshi(rs.getLong("recipient_amount_msat")),
+            PublicKey(rs.getByteVector("recipient_node_id")),
+            Seq(part))
+        }
+        sentByParentId = sentByParentId + (parentId -> sent)
       }
-      q
+      sentByParentId.values.toSeq.sortBy(_.timestamp)
```
pm47: Wonder if using two collections would help (one to index by parent id, one to manage ordering).

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -175,61 +208,94 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
+          rs.getString("direction"),
+          rs.getString("relay_type"),
+          rs.getLong("timestamp"))
+        relayedByHash = relayedByHash + (paymentHash -> (relayedByHash.getOrElse(paymentHash, Nil) :+ part))
       }
-      q
+      relayedByHash.map {
+        case (paymentHash, parts) => parts.head.relayType match {
+          case "channel" => parts.foldLeft(ChannelPaymentRelayed(0 msat, 0 msat, paymentHash, ByteVector32.Zeroes, ByteVector32.Zeroes, parts.head.timestamp)) {
+            case (e, part) => if (part.direction == "IN") {
```
pm47:
```scala
(part.direction: @unchecked) match {
  case "IN" => ...
  case "OUT" => ...
}
```

would make sure something breaks if `part.direction` is neither of `IN`/`OUT`

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -166,12 +166,12 @@ class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, c
 
   override def mdc(currentMessage: Any): MDC = {
     val paymentHash_opt = currentMessage match {
-      case IncomingPacket.NodeRelayPacket(add, _, _, _) => Some(add.paymentHash)
-      case MultiPartPaymentFSM.MultiPartHtlcFailed(paymentHash, _, _) => Some(paymentHash)
-      case MultiPartPaymentFSM.MultiPartHtlcSucceeded(paymentHash, _) => Some(paymentHash)
-      case MultiPartPaymentFSM.ExtraHtlcReceived(paymentHash, _, _) => Some(paymentHash)
-      case PaymentFailed(_, paymentHash, _, _) => Some(paymentHash)
-      case PaymentSent(_, paymentHash, _, _) => Some(paymentHash)
+      case m: IncomingPacket.NodeRelayPacket => Some(m.add.paymentHash)
```
pm47: This pattern is much better to easily handle changes in case class signatures.

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -119,22 +127,26 @@ object IncomingPaymentStatus {
...
- * @param targetNodeId   node ID of the payment recipient.
- * @param createdAt      absolute time in milli-seconds since UNIX epoch when the payment was created.
- * @param paymentRequest Bolt 11 payment request (if paying from an invoice).
- * @param status         current status of the payment.
+ * @param id              internal payment identifier.
+ * @param parentId        internal identifier of a parent payment, or [[id]] if single-part payment.
+ * @param externalId      external payment identifier: lets lightning applications reconcile payments with their own db.
+ * @param paymentHash     payment_hash.
+ * @param paymentType     distinguish different payment types (standard, swaps, etc).
+ * @param amount          amount that will be received by the target node.
```
pm47:
```suggestion
 * @param amount          amount that will be received by the target node, will be different from recipientAmount for trampoline payments.
```

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -130,23 +163,23 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
-        case ChannelPaymentRelayed(_, _, _, fromChannelId, toChannelId, _) =>
-          statement.setBytes(4, fromChannelId.toArray)
-          statement.setBytes(5, toChannelId.toArray)
-        case TrampolinePaymentRelayed(_, _, _, _, fromChannelIds, toChannelIds, _) =>
-          // TODO: @t-bast: we should change the DB schema to allow accurate Trampoline reporting
-          statement.setBytes(4, fromChannelIds.head.toArray)
-          statement.setBytes(5, toChannelIds.head.toArray)
+  override def add(e: PaymentRelayed): Unit = {
+    val payments = e match {
+      case ChannelPaymentRelayed(amountIn, amountOut, _, fromChannelId, toChannelId, ts) => Seq(RelayedPart(fromChannelId, amountIn, "IN", "channel", ts), RelayedPart(toChannelId, amountOut, "OUT", "channel", ts))
```
pm47:
```suggestion
      case ChannelPaymentRelayed(amountIn, amountOut, _, fromChannelId, toChannelId, ts) => Seq(RelayedPart(fromChannelId, amountIn, "IN", "channel", ts), RelayedPart(toChannelId, amountOut, "OUT", "channel", ts)) // non-trampoline relayed payments have one input and one output
```

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -130,23 +163,23 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
+      case TrampolinePaymentRelayed(_, incoming, outgoing, ts) => incoming.map(i => RelayedPart(i.channelId, i.amount, "IN", "trampoline", ts)) ++ outgoing.map(o => RelayedPart(o.channelId, o.amount, "OUT", "trampoline", ts))
```
pm47:
```suggestion
      case TrampolinePaymentRelayed(_, incoming, outgoing, ts) => incoming.map(i => RelayedPart(i.channelId, i.amount, "IN", "trampoline", ts)) ++ outgoing.map(o => RelayedPart(o.channelId, o.amount, "OUT", "trampoline", ts)) // trampoline relayed payments do AMP aggregation and may have m incoming parts and n outgoing parts.
```

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -130,23 +163,23 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
+    }
+    for (p <- payments) {
+      using(sqlite.prepareStatement("INSERT INTO relayed VALUES (?, ?, ?, ?, ?, ?)")) { statement =>
```
pm47: This could be batched? Probably too early to optimize

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -175,61 +208,94 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
-            rs.getByteVector32("to_channel_id"),
-            None, // we don't store the route
-            rs.getLong("timestamp"))))
+        val parentId = UUID.fromString(rs.getString("parent_payment_id"))
+        val part = PaymentSent.PartialPayment(
+          UUID.fromString(rs.getString("payment_id")),
+          MilliSatoshi(rs.getLong("amount_msat")),
+          MilliSatoshi(rs.getLong("fees_msat")),
+          rs.getByteVector32("to_channel_id"),
+          None, // we don't store the route
```
pm47:
```suggestion
          None, // we don't store the route in the audit db
```

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -37,34 +37,63 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
+        statement.executeUpdate("CREATE TABLE IF NOT EXISTS sent (amount_msat INTEGER NOT NULL, fees_msat INTEGER NOT NULL, recipient_amount_msat INTEGER NOT NULL, payment_id TEXT NOT NULL, parent_payment_id TEXT NOT NULL, payment_hash BLOB NOT NULL, payment_preimage BLOB NOT NULL, recipient_node_id BLOB NOT NULL, to_channel_id BLOB NOT NULL, timestamp INTEGER NOT NULL)")
         statement.executeUpdate("CREATE TABLE IF NOT EXISTS received (amount_msat INTEGER NOT NULL, payment_hash BLOB NOT NULL, from_channel_id BLOB NOT NULL, timestamp INTEGER NOT NULL)")
-        statement.executeUpdate("CREATE TABLE IF NOT EXISTS relayed (amount_in_msat INTEGER NOT NULL, amount_out_msat INTEGER NOT NULL, payment_hash BLOB NOT NULL, from_channel_id BLOB NOT NULL, to_channel_id BLOB NOT NULL, timestamp INTEGER NOT NULL)")
+        statement.executeUpdate("CREATE TABLE IF NOT EXISTS relayed (payment_hash BLOB NOT NULL, amount_msat INTEGER NOT NULL, channel_id BLOB NOT NULL, direction TEXT NOT NULL, relay_type TEXT NOT NULL, timestamp INTEGER NOT NULL)")
```
pm47: Should `direction` and `relay_type` be constrained values?

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -175,61 +208,94 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
...
-          toChannelId = rs.getByteVector32("to_channel_id"),
-          timestamp = rs.getLong("timestamp"))
+        val paymentHash = rs.getByteVector32("payment_hash")
+        val part = RelayedPart(
+          rs.getByteVector32("channel_id"),
+          MilliSatoshi(rs.getLong("amount_msat")),
+          rs.getString("direction"),
+          rs.getString("relay_type"),
+          rs.getLong("timestamp"))
+        relayedByHash = relayedByHash + (paymentHash -> (relayedByHash.getOrElse(paymentHash, Nil) :+ part))
```
pm47: There is also this pattern:

```scala
var relayedByHash = Map.empty[ByteVector32, Seq[RelayedPart]].withDefaultValue(Nil)
...
relayedByHash = relayedByHash + (paymentHash -> (relayedByHash(paymentHash) :+ part))
```

But I'm not sure I like it better

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
pm47: Other approach (will also break if there is an unknown relay type):

```scala
      relayedByHash.map {
        case (paymentHash, parts) => (parts.head.relayType: @unchecked) match {
          case "channel" =>
            parts.foldLeft(ChannelPaymentRelayed(0 msat, 0 msat, paymentHash, ByteVector32.Zeroes, ByteVector32.Zeroes, parts.head.timestamp)) {
              case (e, part) if part.direction == "IN" => e.copy(amountIn = part.amount, fromChannelId = part.channelId)
              case (e, part) if part.direction == "OUT" => e.copy(amountOut = part.amount, toChannelId = part.channelId)
            }
          case "trampoline" =>
            parts.foldLeft(TrampolinePaymentRelayed(paymentHash, Nil, Nil, parts.head.timestamp)) {
              case (e, part) if part.direction == "IN" => e.copy(incoming = e.incoming :+ PaymentRelayed.Part(part.amount, part.channelId))
              case (e, part) if part.direction == "OUT" => e.copy(outgoing = e.outgoing :+ PaymentRelayed.Part(part.amount, part.channelId))
            }
        }
      }.toSeq.sortBy(_.timestamp)
```

eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
t-bast: Oups I missed it here, paymentType is required and the migration is there.

t-bast: Fixed in `375b342`

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: Sorting is considered a "for-free" primitive, I think it's less expensive to sort than to duplicate the amount of memory used. If we see a bottleneck there we can start experimenting.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: I hesitated, but indeed I thought it was too early :)

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: I think having them constrained in the code is enough (as long as we handle unknown cases with pattern matching), it makes it easier to change without needing a DB migration.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: I don't have a strong opinion, whatever you feel is better!

pm47: Let's leave it like that

PR #1290, opened 2020-01-23
UpfrontShutdownScript field

t-bast:
We don't implement the `upfront_shutdown_script` feature yet.
However we update our encoding to always specify its field and support reading it.
This allows extending `OpenChannel`/`AcceptChannel` with tlv streams while staying compliant with the spec and other implementations (lnd just started to add the `upfront_shutdown_script` field to their `OpenChannel` message, even if we don't activate the feature bit).

There is one caveat: Phoenix shipped with a version that's incompatible because it interprets the `upfront_shutdown_script` fields as a TLV stream.
So we use a workaround to identify unpatched Phoenix versions and send them the old encoding.
This PR can be applied and deployed safely today.
Once Phoenix is upgraded with that code in the next release, and enough users have updated, we can remove that ugly switch and forget this ever happened :).

Tested against:

* [x] Eclair 0.3.2: ok
* [x] C-lightning master: ok
* [x] lnd 0.9.0-beta: ok 

TODO:

* [ ] Create an issue to remind ourselves to remove that code once Phoenix is updated

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
@@ -20,9 +20,9 @@ import fr.acinq.eclair.wire.CommonCodecs._
 import fr.acinq.eclair.{KamonExt, wire}
 import kamon.Kamon
 import kamon.tag.TagSet
-import scodec.bits.{BitVector, ByteVector}
+import scodec.bits.{BitVector, ByteVector, HexStringSyntax}
 import scodec.codecs._
-import scodec.{Attempt, Codec}
+import scodec.{Attempt, Codec, DecodeResult}
 
 /**
  * Created by PM on 15/11/2016.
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -59,7 +59,25 @@ object LightningMessageCodecs {
...
-  val openChannelCodec: Codec[OpenChannel] = (
+  // Legacy nodes may encode an upfront_shutdown_script even if we didn't advertise support for option_upfront_shutdown_script.
+  // To allow extending all messages with TLV streams, the upfront_shutdown_script was made mandatory in https://github.com/lightningnetwork/lightning-rfc/pull/714.
+  // This codec decodes both legacy and new versions, while always encoding with an upfront_shutdown_script (of length 0 if none actually provided).
+  private val shutdownScriptGuard = Codec[Boolean](
+    (included: Boolean) => if (included) Attempt.Successful(BitVector.empty) else Attempt.Successful(hex"0000".bits),
+    // Bolt 2 specifies that upfront_shutdown_scripts must be P2PKH/P2SH or segwit-v0 P2WPK/P2WSH.
+    // The length of such scripts will always start with 0x00, so we use that to discriminate whether we're decoding an
+    // upfront_shutdown_script or a tlv stream.
+    (b: BitVector) => Attempt.successful(DecodeResult(b.nonEmpty && b.startsWith(hex"00".bits), b))
```
pm47: How about something more explicit:

```scala
    // Bolt 2 specifies that upfront_shutdown_scripts must be P2PKH/P2SH or segwit-v0 P2WPK/P2WSH.
    // Those have predictable lengths, which we use to discriminate whether we're decoding an
    // upfront_shutdown_script or a tlv stream.
    (b: BitVector) => Attempt.successful(DecodeResult(b.startsWith(hex"0016".bits) || b.startsWith(hex"0017".bits) || b.startsWith(hex"0019".bits) || b.startsWith(hex"0022".bits), b))
```

PR #1292, opened 2020-01-24
Channel queries: Encode empty arrays as UNCOMPRESSED

sstone:
c-lightning fails to decode empty arrays of scids or timestamps with an encoding type set to COMPRESSED_ZLIB.
The spec is not specific enough on whether this is valid or not, so we'll set the encoding type of empty arrays to UNCOMPRESSED.

This is PR supersedes #1280


sstone:
@cdecker this fixes an interop issue between eclair and c-lightning (see https://github.com/ElementsProject/lightning/issues/3425 and https://github.com/lightningnetwork/lightning-rfc/pull/729)

cdecker:
Great, thanks for following up on this @sstone and @t-bast :+1:

PR #1293, opened 2020-01-27
Fix availableForSend/Receive

t-bast:
There was a rounding issue with the availableForSend/Receive calculation.
Because CommitTx fee and Htlc fee were computed separately,
but each was individually rounded down to Satoshis, we could
end up with an off-by-one error.

This resulted in an incapacity to send/receive the maximum amount available.

We now allow computing fees in msat, which removes rounding issues.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
@@ -109,15 +109,16 @@ object Transactions {
   val mainPenaltyWeight = 484
   val htlcPenaltyWeight = 578 // based on spending an HTLC-Success output (would be 571 with HTLC-Timeout)
 
-  def weight2fee(feeratePerKw: Long, weight: Int) = Satoshi((feeratePerKw * weight) / 1000)
+  def weight2feeMsat(feeratePerKw: Long, weight: Int) = MilliSatoshi(feeratePerKw * weight)
+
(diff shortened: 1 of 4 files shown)


t-bast:
Great, it looks like the fuzz test has found something :)
I'm investigating that (it may be an issue with the test itself).

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -142,15 +143,23 @@ object Transactions {
...
+  def htlcOutputFee(feeratePerKw: Long): MilliSatoshi = weight2feeMsat(feeratePerKw, htlcOutputWeight)
 
-  def commitTxFee(dustLimit: Satoshi, spec: CommitmentSpec): Satoshi = {
+  /**
+   * While fees are generally computed in Satoshis (since this is the smallest on-chain unit), it may be useful in some
+   * cases to calculate it in MilliSatoshi to avoid rounding issues.
+   * If you are adding multiple fees together for example, you should always add them in MilliSatoshi and then round
+   * down to Satoshi.
+   */
+  def commitTxFeeMsat(dustLimit: Satoshi, spec: CommitmentSpec): MilliSatoshi = {
```
pm47: Since we have a nice typing mechanism which removes all ambiguity, I would have changed the signature of `commitTxFee` to return `MilliSatoshi`, and let callers truncate to satoshis when they need it.

Same comment for `weight2fee`.

t-bast: I did that at first, but it was impacting too much code that really only wanted to deal with sats (all the parts related to actual on-chain transactions related to funding/closing)...

PR #1294, opened 2020-01-28
Fix ChannelUpdate rebroadcast

t-bast:
Comparing with the router ActorRef simply didn't work.
The reason is probably because Peers receive the router's supervisor ref,
which doesn't match what `self` is inside `Router`.

Checking that the origin was the router felt brittle anyway.
Our own gossip doesn't have an origin; signaling that with an empty
origin is what makes the most sense.


t-bast:
Waiting for a confirmation that this fixes the issue for our gitter users and then I'll merge.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -406,10 +406,12 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
                     awaiting = awaiting1)
                   // we only reprocess updates and nodes if validation succeeded
                   val d2 = reprocessUpdates.foldLeft(d1) {
-                    case (d, (u, origins)) => origins.foldLeft(d) { case (d, origin) => handle(u, origin, d) } // we reprocess the same channel_update for every origin (to preserve origin information)
+                    case (d, (u, origins)) if origins.isEmpty => handle(u, None, d)
```
pm47: We may want to use a proper typing for origin, instead of using an ActorRef, this would remove the ambiguity here.

t-bast: Done. Feels cleaner!

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -171,9 +171,18 @@ case object GetRoutingState
...
 
-case class Stash(updates: Map[ChannelUpdate, Set[ActorRef]], nodes: Map[NodeAnnouncement, Set[ActorRef]])
+// @formatter:off
+sealed trait GossipOrigin {
+  val actor: ActorRef
+}
+/** Gossip that we received from a remote peer. */
+case class RemoteGossip(actor: ActorRef) extends GossipOrigin
+/** Gossip that was generated by our node. */
+case class LocalGossip(actor: ActorRef) extends GossipOrigin
```
pm47: Why not:

 ```scala
case object LocalGossip extends GossipOrigin
```

File: eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala [hunk 3]
```diff
@@ -378,8 +378,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
       /**
        * Send and count in a single iteration
        */
-      def sendAndCount(msgs: Map[_ <: RoutingMessage, Set[ActorRef]]): Int = msgs.foldLeft(0) {
-        case (count, (_, origins)) if origins.contains(self) =>
+      def sendAndCount(msgs: Map[_ <: RoutingMessage, Set[GossipOrigin]]): Int = msgs.foldLeft(0) {
+        case (count, (_, origins)) if origins.map(_.actor).contains(self) =>
```
```suggestion
        case (count, (_, origins)) if origins.contains(RemoteGossip(self)) =>
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -602,11 +602,11 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
    *         - true if there is a filter and msg has no timestamp, or has one that matches the filter
    *         - false otherwise
    */
-  def timestampInRange(msg: RoutingMessage, origins: Set[ActorRef], gossipTimestampFilter_opt: Option[GossipTimestampFilter]): Boolean = {
+  def timestampInRange(msg: RoutingMessage, origins: Set[GossipOrigin], gossipTimestampFilter_opt: Option[GossipTimestampFilter]): Boolean = {
     // For our own gossip, we should ignore the peer's timestamp filter.
     val isOurGossip = msg match {
       case n: NodeAnnouncement if n.nodeId == nodeParams.nodeId => true
-      case _ if origins.isEmpty || origins == Set(router) => true // if gossip doesn't come from another peer, it's ours
+      case _ if origins.exists(_.isInstanceOf[LocalGossip]) => true
```
pm47: With the suggestion below this would become:

```suggestion
      case _ if origins.contains(LocalGossip) => true
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -171,9 +171,18 @@ case object GetRoutingState
 
 case class RoutingState(channels: Iterable[PublicChannel], nodes: Iterable[NodeAnnouncement])
 
-case class Stash(updates: Map[ChannelUpdate, Set[ActorRef]], nodes: Map[NodeAnnouncement, Set[ActorRef]])
+// @formatter:off
+sealed trait GossipOrigin {
```
pm47: I don't think we should make `actor` an attribute of `GossipOrigin`. Also, from what I gather the `RemoteGossip.actor` is a `Peer` whereas `LocalGossip.actor` is a `Channel`.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -171,9 +171,18 @@ case object GetRoutingState
...
+  val actor: ActorRef
+}
+/** Gossip that we received from a remote peer. */
+case class RemoteGossip(actor: ActorRef) extends GossipOrigin
```
```suggestion
case class RemoteGossip(peer: ActorRef) extends GossipOrigin
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: I hesitated, I figured it would be better to keep the reference of the (local) sender in case we wanted to send it an ACK or something, but it's probably YAGNI so we can remove it and re-add it later if we need it.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -171,9 +171,21 @@ case object GetRoutingState
 
 case class RoutingState(channels: Iterable[PublicChannel], nodes: Iterable[NodeAnnouncement])
 
-case class Stash(updates: Map[ChannelUpdate, Set[ActorRef]], nodes: Map[NodeAnnouncement, Set[ActorRef]])
+// @formatter:off
+sealed trait GossipOrigin {
+  def getPeer: Option[ActorRef] = this match {
```
pm47: I'd rather use pattern matching where needed.

t-bast: Sounds good, done.

PR #1295, opened 2020-01-28
Support deterministic build of eclair-core artifact on ubuntu

araspitzu:
This PR adds support for deterministic builds of the eclair-core artifact by updating the necessary maven plugins (maven-source-plugin and maven-jar-plugin) to version 3.2.0 which support deterministic builds and hardcoding the property `project.build.outputTimestamp` to a fixed date (this can be updated at each release) . The deterministic build only works with the submodule `eclair-core` due to some quirks with the maven-capsule plugin, also the build is *not* deterministic across different OSs because of the way maven packages files with system-dependent newline markers. The SHA256 of eclair core at commit 93c2683 is: `db3064fc4fc9e20c141770294186fbb82b50190ffc13f6bcc09379db2fb33b2b`. 

This is a first step toward deterministic builds for the mobile apps.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -11,7 +11,7 @@
 Eclair supports deterministic builds for the eclair-core submodule, this is the 'core' of the eclair application
 and its artifact can be deterministically built achieving byte-to-byte equality for each build. To build the exact 
 same artifacts that we release, you must use the build environment (OS, JDK, maven...) that we specify in our 
-release notes, to check if your artifact is correct please check out the release page for the latest published release.
+release notes.
 
 To build the project and run the tests, simply run:
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -10,10 +10,10 @@
 
 Eclair supports deterministic builds for the eclair-core submodule, this is the 'core' of the eclair application
 and its artifact can be deterministically built achieving byte-to-byte equality for each build. This process helps
-to ensure that the artifact you're building is exactly the same as the one published in the release, please note
-that the deterministic build is limited to eclair-core and is also OS dependent, only unix based builds might be
-published in the release page. To check if your artifact is correct please check out the release page for the latest 
-published release.
+to ensure the artifact you're building is exactly the same as the one published in the release, please note
+that the deterministic build is limited to eclair-core and is OS dependent, to build the artifact deterministically
+you must use Ubuntu 18.10, JDK11.0.4 and maven 3.6.0. To check if your artifact is correct please check out the release 
+page for the latest published release.
 
 To build the project and run the tests, simply run:
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -9,11 +9,9 @@
 ## Build
 
 Eclair supports deterministic builds for the eclair-core submodule, this is the 'core' of the eclair application
-and its artifact can be deterministically built achieving byte-to-byte equality for each build. This process helps
-to ensure the artifact you're building is exactly the same as the one published in the release, please note
-that the deterministic build is limited to eclair-core and is OS dependent, to build the artifact deterministically
-you must use Ubuntu 18.10, JDK11.0.4 and maven 3.6.0. To check if your artifact is correct please check out the release 
-page for the latest published release.
+and its artifact can be deterministically built achieving byte-to-byte equality for each build. To build the exact 
+same artifacts that we release, you must use the build environment (OS, JDK, maven...) that we specify in our 
+release notes, to check if your artifact is correct please check out the release page for the latest published release.
 
 To build the project and run the tests, simply run:
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -8,6 +8,13 @@
 
 ## Build
 
+Eclair supports deterministic builds for the eclair-core submodule, this is the 'core' of the eclair application
+and its artifact can be deterministically built achieving byte-to-byte equality for each build. This process helps
+to ensure that the artifact you're building is exactly the same as the one published in the release, please note
+that the deterministic build is limited to eclair-core and is also OS dependent, only unix based builds might be
+published in the release page. To check if your artifact is correct please check out the release page for the latest 
+published release.
+
 To build the project and run the tests, simply run:
 
 ```shell
--- a/eclair-core/pom.xml
+++ b/eclair-core/pom.xml
@@ -69,21 +69,6 @@
                     </archive>
                 </configuration>
             </plugin>
-            <plugin>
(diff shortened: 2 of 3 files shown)


t-bast:
Ave you tried that: https://maven.apache.org/guides/mini/guide-reproducible-builds.html
It seems that maven 3.6.3 has almost everything needed for reproducible builds out of the box (which would be better than another plugin). But I'm not sure it works :)

araspitzu:
It didn't work because there are some other non-reproducible bits in the MANIFEST file inside the generated jar, it *might* work by changing and/or updating other plugins like the maven-jar-plugin but i couldn't get it working and it felt like a rabbit hole, on the other hand maven-reproducible-build works out of the box :) (but unfortunately is yet another plugin to our build)

araspitzu:
Not sure if we want to add a section in the README.md  for the reproducible builds or we just add it to the release page.

t-bast:
> Not sure if we want to add a section in the README.md for the reproducible builds or we just add it to the release page.

I think it's worth adding to the main README so that people find it easily.
We need to specify that it only works on linux, and maybe other querks too.

pm47:
> It didn't work because there are some other non-reproducible bits in the MANIFEST file inside the generated jar, it might work by changing and/or updating other plugins like the maven-jar-plugin

The doc says to `upgrade your plugins to reproducible version, particularly maven-source-plugin, maven-jar-plugin and maven-assembly-plugin to version 3.2.0 minimum`, have you tried that? Agree with the rabbit hole risk, I'd just like to make sure that we have fully tried the simpler solution.

>>  Not sure if we want to add a section in the README.md for the reproducible builds or we just add it to the release page.

> I think it's worth adding to the main README so that people find it easily.
We need to specify that it only works on linux, and maybe other querks too.

How about BUILD.md? And a standard paragraph in each release description (like the part on signature)?

pm47:
NB, seems the last pain point (update of `project.build.outputTimestamp`) has been fixed two months ago: https://issues.apache.org/jira/browse/MRELEASE-1029. Seems really tempting. @sstone are you using the release plugin for this project btw?

araspitzu:
> > It didn't work because there are some other non-reproducible bits in the MANIFEST file inside the generated jar, it might work by changing and/or updating other plugins like the maven-jar-plugin
> 
> The doc says to `upgrade your plugins to reproducible version, particularly maven-source-plugin, maven-jar-plugin and maven-assembly-plugin to version 3.2.0 minimum`, have you tried that? Agree with the rabbit hole risk, I'd just like to make sure that we have fully tried the simpler solution.
> 
> > > Not sure if we want to add a section in the README.md for the reproducible builds or we just add it to the release page.
> 
> > I think it's worth adding to the main README so that people find it easily.
> > We need to specify that it only works on linux, and maybe other querks too.
> 
> How about BUILD.md? And a standard paragraph in each release description (like the part on signature)?

Alright, so i just gave another try to the non-plugin based approach and..it worked, i had to update *both* the maven-source-plugin and maven-jar-plugin to 3.2.0, i think in my earlier attempt i missed the maven-source-plugin and deemed the approach unfeasible too early. As you point out the `project.build.outputTimestamp` could be updated by the release phase of maven but i'm not sure if we use it during our release process, for the moment i'm hardcoding it to a fixed date. I'm also adding a small paragraph in BUILD.md to mention the eclair-core artifact build is deterministic.

sstone:
We don't use the maven release plugin yet (testing and tagging the release is done manually)

sstone:
@araspitzu Did you run tests with different versions of JDK11 and different Linux distros ? 
I think that it's fine to restrict deterministic builds to specific operating systems and JDKs but it should be documented too.

araspitzu:
> @araspitzu Did you run tests with different versions of JDK11 and different Linux distros ?
> I think that it's fine to restrict deterministic builds to specific operating systems and JDKs but it should be documented too.

I didn't try with different distros but I tried with different versions of ubuntu (18.10 vs 17.10) and different minor versions of JDK (JDK11.0.4 vs JDK11.0.3), the resulting artifact was identical. If we're okay restricting the deterministic build to the latest ubuntu version i'll add this to the paragraph   in the BUILD.md.

BUILD.md
```diff
@@ -10,10 +10,10 @@
 
 Eclair supports deterministic builds for the eclair-core submodule, this is the 'core' of the eclair application
 and its artifact can be deterministically built achieving byte-to-byte equality for each build. This process helps
-to ensure that the artifact you're building is exactly the same as the one published in the release, please note
-that the deterministic build is limited to eclair-core and is also OS dependent, only unix based builds might be
-published in the release page. To check if your artifact is correct please check out the release page for the latest 
-published release.
+to ensure the artifact you're building is exactly the same as the one published in the release, please note
+that the deterministic build is limited to eclair-core and is OS dependent, to build the artifact deterministically
+you must use Ubuntu 18.10, JDK11.0.4 and maven 3.6.0. To check if your artifact is correct please check out the release
```
sstone: These versions will change over time (maven 3.6.0 files are not available anymore for example, and JDK11.0.6 has been released a few weeks ago). I propose instead: 
`...byte-to-byte equality for each build. To build the exact same artifacts that we release, you must use the build environment (OS, JDK, maven...) that we specify in our release notes.`

BUILD.md
```diff
@@ -8,6 +8,11 @@
 
 ## Build
 
+Eclair supports deterministic builds for the eclair-core submodule, this is the 'core' of the eclair application
+and its artifact can be deterministically built achieving byte-to-byte equality for each build. To build the exact 
+same artifacts that we release, you must use the build environment (OS, JDK, maven...) that we specify in our 
+release notes, to check if your artifact is correct please check out the release page for the latest published release.
```
sstone: nit: I would remove the last part of the sentence and stop at `that we specify in our ￼release notes.`

sstone:
Can you please update the hash of the eclair-core jar that should be built ? thanks

n1bor:
This one builds to the same hash:
eb3091401b8f86ed7c23328acb069b17a0805a4aadeb852bef86c3ab08f591a1  eclair-core_2.11-0.3.3.jar

But as mentioned somewhere node and node-gui jar is not.. which is very annoying as that is the one I actually run...

sstone:
> But as mentioned somewhere node and node-gui jar is not.. which is very annoying as that is the one I actually run...

Yes it's mentioned in the PR's description :)
You can manually unzip  the eclair-node jar (or eclair-node-gui) and verify the checksum of the included eclair-core jar, but it's tedious. We're working on finding another way of packaging our server apps.

PR #1298, opened 2020-01-30
Electrum: allow watcher to watch for mempool transactions

sstone:
Watcher now handles WatchConfirmed watches where min depth is set to 0: the watch event will sent when the tx enters the mempool

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
@@ -117,6 +117,7 @@ class ElectrumWatcher(blockCount: AtomicLong, client: ActorRef) extends Actor wi
 
     case ElectrumClient.GetScriptHashHistoryResponse(_, history) =>
       // we retrieve the transaction before checking watches
+      // NB: height=-1 means that the tx is unconfirmed and at least one of its inputs is also unconfirmed. we need to take them into consideration if we want to handle unconfirmed txes (which is the case for turbo channels)
       history.filter(_.height >= -1).foreach { item => client ! ElectrumClient.GetTransaction(item.tx_hash, Some(item)) }
 
     case ElectrumClient.GetTransactionResponse(tx, Some(item: ElectrumClient.TransactionHistoryItem)) =>
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
@@ -233,7 +233,11 @@ object ElectrumWatcher {
   def makeDummyShortChannelId(txid: ByteVector32): (Int, Int) = {
     // we use a height of 0
     // - to make sure that the tx will be marked as "confirmed"
-    // - to identify scids linked to 0-conf channels
+    // - to easily identify scids linked to 0-conf channels
+    //
+    // this gives us a probability of collisions of 0.1% for 5 0-conf channels and 1% for 20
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
```diff
@@ -117,7 +117,7 @@ class ElectrumWatcher(blockCount: AtomicLong, client: ActorRef) extends Actor wi
 
     case ElectrumClient.GetScriptHashHistoryResponse(_, history) =>
       // we retrieve the transaction before checking watches
-      history.filter(_.height >= 0).foreach { item => client ! ElectrumClient.GetTransaction(item.tx_hash, Some(item)) }
+      history.filter(_.height >= -1).foreach { item => client ! ElectrumClient.GetTransaction(item.tx_hash, Some(item)) }
```
pm47:
```suggestion
      // NB: height=-1 means that the tx is unconfirmed and at least one of its inputs is also unconfirmed. we need to take them into consideration if we want to handle unconfirmed txes (which is the case for turbo channels)
      history.filter(_.height >= -1).foreach { item => client ! ElectrumClient.GetTransaction(item.tx_hash, Some(item)) }
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
```diff
@@ -133,8 +133,13 @@ class ElectrumWatcher(blockCount: AtomicLong, client: ActorRef) extends Actor wi
           channel ! WatchEventSpentBasic(event)
           Some(w)
       }).flatten
+      // special case for mempool watches (min depth = 0)
```
pm47:
```scala
      // this is for WatchConfirmed
      val watchConfirmedTriggered = watches.collect {
          case w@WatchConfirmed(channel, txid, _, minDepth, BITCOIN_FUNDING_DEPTHOK) if txid == tx.txid && minDepth == 0 =>
            // special case for mempool watches (min depth = 0)
            channel ! WatchEventConfirmed(BITCOIN_FUNDING_DEPTHOK, 0, 0, tx)
            Some(w)
          case WatchConfirmed(_, txid, _, minDepth, _) if txid == tx.txid && minDepth > 0=>
            // min depth > 0 here
            val txheight = item.height
            val confirmations = height - txheight + 1
            log.info(s"txid=$txid was confirmed at height=$txheight and now has confirmations=$confirmations (currentHeight=$height)")
            if (confirmations >= minDepth) {
              // we need to get the tx position in the block
              client ! ElectrumClient.GetMerkle(txid, txheight, Some(tx))
            }
            None
        }.flatten
```

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcherSpec.scala
```diff
@@ -123,6 +123,122 @@ class ElectrumWatcherSpec extends TestKit(ActorSystem("test")) with FunSuiteLike
...
+    probe.send(bitcoincli, BitcoinReq("sendrawtransaction", tx2.toString()))
+    probe.expectMsgType[JValue]
+
+    val confirmed = listener.expectMsgType[WatchEventConfirmed](20 seconds)
+    assert(confirmed.tx.txid === tx2.txid)
+    system.stop(watcher)
+  }
+
+  test("compute dummy height/txindex for 0-conf channels") {
+    val buffer = new Array[Byte](32)
```
pm47: What are we testing exactly here?

```scala
    import fr.acinq.eclair.randomBytes32
    val dummies = (0 to 1000).map { _ =>
      ElectrumWatcher.makeDummyShortChannelId(randomBytes32)
    } toSet

    assert(dummies.size >= 1000 - 10)
```

PR #1301, opened 2020-02-03
Notify change in balance when updating network fee

pm47:
As a funder, when we send an `update_fee`, the channel fails to notify
other parts of the code (e.g. the relayer), causing "insufficient funds"
errors down the road.


pm47:
@t-bast not sure what's wrong with the integration tests

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -751,6 +749,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
             // if we have newly acknowledged changes let's sign them
             self ! CMD_SIGN
           }
+          if (commitments1.availableBalanceForSend != commitments1.availableBalanceForSend) {
```
t-bast:
```suggestion
          if (d.commitments.availableBalanceForSend != commitments1.availableBalanceForSend) {
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
```diff
@@ -53,7 +53,9 @@ case class ChannelErrorOccurred(channel: ActorRef, channelId: ByteVector32, remo
 case class NetworkFeePaid(channel: ActorRef, remoteNodeId: PublicKey, channelId: ByteVector32, tx: Transaction, fee: Satoshi, txType: String) extends ChannelEvent
 
 // NB: this event is only sent when the channel is available
-case class AvailableBalanceChanged(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, localBalance: MilliSatoshi, commitments: Commitments) extends ChannelEvent
+case class AvailableBalanceChanged(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, commitments: Commitments) extends ChannelEvent {
+  def localBalance: MilliSatoshi = commitments.remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(commitments.remoteCommit).spec.toRemote
```
t-bast: I would completely remove this `localBalance` field and the corresponding DB table (which is probably very noisy).

t-bast:
We probably need to add this check in all outgoing update_* commands...

PR #1302, opened 2020-02-04
Better Bolt 11 parsing spec compliance

t-bast:
Add more Bolt 11 spec tests.
Ignore hashed fields with invalid length as the spec requires:

```
A reader:
    * MUST skip over unknown fields, OR an f field with unknown version, OR p, h, s or n fields that do NOT have data_lengths of 52, 52, 52 or 53, respectively.
```

See https://github.com/lightningnetwork/lightning-rfc/pull/699 and https://github.com/lightningnetwork/lightning-rfc/pull/736.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
@@ -396,20 +396,23 @@ object PaymentRequest {
 
     val paymentHashCodec: Codec[PaymentHashTag] = discriminatorWithDefault(
       discriminated.by(dataLengthCodec)
-        // 260 is 256 bits padded to the next multiple of 5 (because of Bech32)
-        .typecase(260, dataCodecLengthDiscriminated(260, bytes32.as[PaymentHash])),
+        // Bolt11: MUST skip over (...) p, h, s or n fields that do NOT have data_lengths of 52, 52, 52 or 53, respectively.
+        // NB: Bech32 encodes 5 bits per character.
+        .typecase(52 * 5, dataCodecLengthDiscriminated(52 * 5, bytes32.as[PaymentHash])),
       bits.as[InvalidPaymentHash].upcast[PaymentHashTag])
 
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -377,41 +390,58 @@ object PaymentRequest {
...
+
+    val paymentSecretCodec: Codec[PaymentSecretTag] = discriminatorWithDefault(
+      discriminated.by(dataLengthCodec)
+        .typecase(260, dataCodecLengthDiscriminated(260, bytes32.as[PaymentSecret])),
+      bits.as[InvalidPaymentSecret].upcast[PaymentSecretTag])
+
+    val descriptionHashCodec: Codec[DescriptionHashTag] = discriminatorWithDefault(
+      discriminated.by(dataLengthCodec)
+        .typecase(260, dataCodecLengthDiscriminated(260, bytes32.as[DescriptionHash])),
+      bits.as[InvalidDescriptionHash].upcast[DescriptionHashTag])
```
sstone: A comment that explains where `260` comes from (maybe a quote from BOLT11) may help here

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -158,11 +158,12 @@ object PaymentRequest {
...
   sealed trait TaggedField
 
-  sealed trait UnknownTaggedField extends TaggedField
+  sealed trait UnknownTaggedField extends TaggedField {
+    def data: BitVector
+  }
 
   // @formatter:off
   case class UnknownTag0(data: BitVector) extends UnknownTaggedField
-  case class UnknownTag1(data: BitVector) extends UnknownTaggedField
```
pm47: Nice catch

pm47:
Here is my take: https://github.com/ACINQ/eclair/tree/ignore-fields-payment-req

t-bast:
Closing in favor of #1308

PR #1303, opened 2020-02-05
Better logging for transport handler

pm47:
We currently rely on `require`, which is convenient, but doesn't allow
fine-grained exception control.

Also, in case of errors, logging is done at the supervisor level, where
we lose the remote `node_id`.

Instead, we type some crypto-related errors and log them in the
`TransportHandler`, which already has the correct MDC.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
@@ -144,7 +144,7 @@ class TransportHandler[T: ClassTag](keyPair: KeyPair, rs: Option[ByteVector], co
                 case (reader1, message, None) => {
                   // we're still in the middle of the handshake process and the other end must first received our next
                   // message before they can reply
-                  require(remainder.isEmpty, "unexpected additional data received during handshake")
+                  if (remainder.nonEmpty) throw UnexpectedDataDuringHandshake(ByteVector(remainder))
                   connection ! Tcp.Write(buf(TransportHandler.prefix +: message))
                   stay using HandshakeData(reader1, remainder)
                 }
@@ -313,6 +313,7 @@ object TransportHandler {
   val prefix: Byte = 0x00
 
   case class InvalidTransportPrefix(buffer: ByteVector) extends RuntimeException(s"invalid transport prefix first64=${buffer.take(64).toHex}")
+  case class UnexpectedDataDuringHandshake(buffer: ByteVector) extends RuntimeException(s"unexpected additional data received during handshake first64=${buffer.take(64).toHex}")
 
   val prologue = ByteVector.view("lightning".getBytes("UTF-8"))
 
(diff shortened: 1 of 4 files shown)


t-bast:
I fixed the flaky audit DB test in https://github.com/ACINQ/eclair/pull/1304

eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
```diff
@@ -118,162 +119,191 @@ class TransportHandler[T: ClassTag](keyPair: KeyPair, rs: Option[ByteVector], co
...
+              context.parent ! HandshakeCompleted(connection, self, remoteNodeId)
+              val nextStateData = WaitingForListenerData(Encryptor(ExtendedCipherState(enc, ck)), Decryptor(ExtendedCipherState(dec, ck), ciphertextLength = None, remainder))
+              goto(WaitingForListener) using nextStateData
+
+            case (writer, _, None) => {
+              writer.write(ByteVector.empty) match {
+                case (reader1, message, None) => {
+                  // we're still in the middle of the handshake process and the other end must first received our next
+                  // message before they can reply
+                  require(remainder.isEmpty, "unexpected additional data received during handshake")
```
t-bast: Another require to remove! Death to them all!

eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
```diff
@@ -118,162 +119,191 @@ class TransportHandler[T: ClassTag](keyPair: KeyPair, rs: Option[ByteVector], co
...
+          case _ => log.warning(s"unhandled message $msg in state ${d.getClass.getSimpleName}")
+        }
+        stay
+    }
   }
 
-  override def aroundPostStop(): Unit = connection ! Tcp.Close // attempts to gracefully close the connection when dying
+  override def aroundPostStop(): Unit
+
+  = connection ! Tcp.Close // attempts to gracefully close the connection when dying
```
t-bast: Weird line changes in this function...

eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
```diff
@@ -118,162 +119,191 @@ class TransportHandler[T: ClassTag](keyPair: KeyPair, rs: Option[ByteVector], co
...
 
   initialize()
 
-  override def mdc(currentMessage: Any): MDC = {
+  override def mdc(currentMessage: Any): MDC
+
```
t-bast: Same, weird line change, formatter is going crazy

eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
pm47: Good catch!

PR #1306, opened 2020-02-06
Node relayer faster fulfill

t-bast:
These two commits are best reviewed individually.

`06faa52` is a small refactoring of payment errors to make it simpler for Phoenix/Eclair-Mobile to classify errors.

`2e48792` is an improvement for trampoline HTLC fulfills.
We were previously waiting for the whole downstream payment to be settled (all individual HTLCs).
We can do better and fulfill upstream as soon as we get the preimage (which only needs one downstream fulfill).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
@@ -20,8 +20,8 @@ import java.util.UUID
 
 import akka.actor.{Actor, ActorRef, DiagnosticActorLogging, PoisonPill, Props}
 import akka.event.Logging.MDC
-import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.bitcoin.{ByteVector32, Crypto}
 import fr.acinq.eclair.channel.{CMD_FAIL_HTLC, CMD_FULFILL_HTLC, Upstream}
 import fr.acinq.eclair.payment._
 import fr.acinq.eclair.payment.receive.MultiPartPaymentFSM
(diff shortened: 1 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -200,8 +219,10 @@ object NodeRelayer {
    *
    * @param upstream    complete HTLC set received.
    * @param nextPayload relay instructions.
+   * @param paymentId   id of the outgoing payment.
+   * @param settled     true if we already settled the payment upstream.
```
pm47: I think this name creates confusion with an on-chain settlement. How about:
```suggestion
   * @param fulfilledUpstream     true if we already settled the payment upstream.
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -50,31 +50,36 @@ class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, c
 
   override def receive: Receive = main(Map.empty, Map.empty)
 
-  def main(pendingIncoming: Map[ByteVector32, PendingRelay], pendingOutgoing: Map[UUID, PendingResult]): Receive = {
+  def main(pendingIncoming: Map[ByteVector32, PendingRelay], pendingOutgoing: Map[ByteVector32, PendingResult]): Receive = {
```
pm47: Love the symmetry

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -95,26 +100,40 @@ class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, c
...
-          val outgoing = parts.map(part => PaymentRelayed.Part(part.amountWithFees, part.toChannelId))
-          context.system.eventStream.publish(TrampolinePaymentRelayed(paymentHash, incoming, outgoing))
-      }
-      context become main(pendingIncoming, pendingOutgoing - id)
-
-    case PaymentFailed(id, _, failures, _) =>
-      log.debug("trampoline payment failed")
-      pendingOutgoing.get(id).foreach { case PendingResult(upstream, nextPayload) => rejectPayment(upstream, translateError(failures, nextPayload.outgoingNodeId)) }
-      context become main(pendingIncoming, pendingOutgoing - id)
+      // We may have already fulfilled upstream, but we can now emit an accurate relayed event and clean-up resources.
```
pm47: I would have factored the whole fulfillment process and done it the same way in both handlers, because a payment is really sent when we forward the preimage upstream. No real reasons to distinguish I think.

PR #1307, opened 2020-02-07
Rework packaging

araspitzu:
This PR reworks the way we package eclair replacing the unmaintained maven-capsule plugin with  maven-assembly. The output of the packaging is now different and consist in a `zip` archive containing a folder with launcher scripts (windows + unix) and another folder with our artifacts and their dependencies (jar packed and with signatures). The result of the build for both `eclair-node` and `eclair-node-gui` is now deterministic.

SHA256 of `eclair-node-0.3.4-SNAPSHOT-c4f7cbd-bin.zip`:
`4be4e290e364c67b59a1ee7797b66c9ab59a8c06ec5370abbdea438030f5b3ec`

- [x] Fix docker image
- [x] Check java agent usage and inclusion
- [x] Update readme and instructions


pm47:
It would be interesting to do a quick comparison which a few popular jvm projects and find out what is the standard way of dealing with this.

pm47:
Btw where are those scripts from?

araspitzu:
@pm47 zipping and using two folders `lib/` and `bin/` seem to be pretty standard, for example gradle distribution plugin does [the same](https://docs.gradle.org/current/userguide/application_plugin.html?_ga=2.108101888.304935308.1581005172-1693878526.1580131293#sec:the_distribution). The launcher scripts are taken from [sbt native packager](https://github.com/sbt/sbt-native-packager/tree/master/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts) , these are actively maintained and proven to work :) . About the scripts, they have been distributed under BSD-2 license so i included the license header in the header of the file.

pm47:
@araspitzu yes just took a look at https://sbt-native-packager.readthedocs.io and this seems to be heading to the right direction. I know that @sstone was leaning towards that layout too.

eclair-node/pom.xml
```diff
@@ -49,25 +49,24 @@
...
                         </goals>
-                        <configuration>
-                            <appClass>fr.acinq.eclair.Boot</appClass>
-                            <type>fat</type>
-                            <fileName>${project.name}-${project.version}</fileName>
-                            <fileDesc>-${git.commit.id.abbrev}</fileDesc>
-                            <manifest>
-                                <entry>
-                                    <key>Java-Agents</key>
-                                    <value>kanela-agent-1.0.1.jar</value>
```
pm47: How will agents be dealt with?

araspitzu: Good catch i have to re-include it in the jar manifest, note that the [launcher](https://github.com/ACINQ/eclair/pull/1307/files#diff-282a1410413f84c78de1260977c66120R162) has support for java agents (although it needs to be updated)

eclair-node-gui/src/main/resources/eclair-node-gui.sh
```diff
@@ -0,0 +1,361 @@
...
+#
+# Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+# Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#!/usr/bin/env bash
```
sstone: Will this work if the user's shell is not bash (zsh is very popular for example ) ?

araspitzu: I don't think so, while it may work with other shells the script is designed to work with bash. However there is some level of interoperability between shells, I just tried the script with `#!/usr/bin/sh` (which is `dash` shell in my system) and it does work so we could recommend `bash` as default shell but suggest it may work with others too. Note that if we want to use an `alpine` linux image for our docker we should use an `ash` shell script, for which there is already a  [template](https://github.com/sbt/sbt-native-packager/tree/master/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts)

eclair-node/pom.xml
araspitzu: At 106224c the kanela agent is disabled by default but can be activated using the `-with-kanela` cmd line option when launching eclair-node or eclair-node-gui.

pm47:
I'd remove the `_2.11` part from the zips.

araspitzu:
I think we should add the `eclair-cli` in the `/bin` folder for ease of usage, in a normal situation the user deploying eclair wants to use the cli as well, WDYT?

sstone:
@araspitzu yes it makes sense

eclair-node-gui/src/main/resources/eclair-node-gui.sh
```diff
@@ -0,0 +1,362 @@
...
+    exit 1
+  else
+    local major=$(echo "$java_version" | cut -d'.' -f1)
+    if [[ "$major" -eq "1" ]]; then
+     local major=$(echo "$java_version" | cut -d'.' -f2)
+    fi
+    if [[ "$major" -lt "6" ]]; then
+      echo
+      echo The java installation you have is not up to date
+      echo $app_name requires at least version 1.6+, you have
```
dpad85: > at least version 1.6+

Should be `at least version 1.8+`

eclair-node-gui/src/main/resources/eclair-node-gui.sh
```diff
@@ -0,0 +1,362 @@
...
+
+# Now check to see if it's a good enough version
+# TODO - Check to see if we have a configured default java version, otherwise use 1.6
+java_version_check() {
+  readonly java_version=$("$java_cmd" -version 2>&1 | awk -F '"' '/version/ {print $2}')
+  if [[ "$java_version" == "" ]]; then
+    echo
+    echo No java installations was detected.
+    echo Please go to http://www.java.com/getjava/ and download
+    echo
```
dpad85: > Please go to http://www.java.com/getjava/ and download

Should point to: https://adoptopenjdk.net/?variant=openjdk11&jvmVariant=hotspot

eclair-node/src/main/resources/eclair-node.bat
```diff
@@ -0,0 +1,192 @@
...
+
+if "%JAVAOK%"=="false" (
+  echo.
+  echo A Java JDK is not installed or can't be found.
+  if not "%JAVA_HOME%"=="" (
+    echo JAVA_HOME = "%JAVA_HOME%"
+  )
+  echo.
+  echo Please go to
+  echo   http://www.oracle.com/technetwork/java/javase/downloads/index.html
```
dpad85: We should point to https://adoptopenjdk.net/?variant=openjdk11&jvmVariant=hotspot instead.

eclair-node/src/main/resources/eclair-node.sh
```diff
@@ -0,0 +1,362 @@
...
+
+# Now check to see if it's a good enough version
+# TODO - Check to see if we have a configured default java version, otherwise use 1.6
+java_version_check() {
+  readonly java_version=$("$java_cmd" -version 2>&1 | awk -F '"' '/version/ {print $2}')
+  if [[ "$java_version" == "" ]]; then
+    echo
+    echo No java installations was detected.
+    echo Please go to http://www.java.com/getjava/ and download
+    echo
```
dpad85: Should point to https://adoptopenjdk.net/?variant=openjdk11&jvmVariant=hotspot

eclair-node/src/main/resources/eclair-node.sh
```diff
@@ -0,0 +1,362 @@
...
+    exit 1
+  else
+    local major=$(echo "$java_version" | cut -d'.' -f1)
+    if [[ "$major" -eq "1" ]]; then
+     local major=$(echo "$java_version" | cut -d'.' -f2)
+    fi
+    if [[ "$major" -lt "6" ]]; then
```
dpad85: We need at least java 8, java 11 recommended.

eclair-node/pom.xml
```diff
@@ -49,25 +49,25 @@
...
             </plugin>
             <plugin>
-                <groupId>com.github.chrisdchristo</groupId>
-                <artifactId>capsule-maven-plugin</artifactId>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <version>3.2.0</version>
+                <configuration>
+                    <finalName>${project.name}-${project.version}-${git.commit.id.abbrev}</finalName>
+                    <descriptors>
+                        <descriptor>modules/eclair-node.xml</descriptor>
```
pm47: Nit, could also be `${project.name}.xml` but I don't see the reason to provide too much cardinality.

```suggestion
                        <descriptor>modules/assembly.xml</descriptor>
```

PR #1310, opened 2020-02-09
Add public channel hints to invoices

btcontract:
This adds a `withExtraHops=1/0` optional parameter to `createinvoice` API call which would enrich invoice with routing hints for up to 4 smallest public channels which can receive a requested amount right now.

Reasons this might be useful:
- constrained devices may start pruning a graph as it grows and the easiest strategy would be to exclude smaller channels first so unless hinted not all public channels may be visible.

- otherwise this may work as a cue to sender about which channels to use first by, say, temporary elevating their score in a local graph. This may work especially well if receiver has, say, hundreds of public channels but only small number of those can actually receive a given amount right now.

This was successfully tested manually on testnet, any ideas about how to construct a proper automatic test? What confuses me is this involves three subsystems at once (API, router, relayer).


t-bast:
We probably already have similar code in one of the branches used by eclair-mobile / phoenix.
I'll let @dpad85 or @pm47 chime in on the best way to integrate that in our master branch (we had already discussed this internally).

dpad85:
> We probably already have similar code in one of the branches used by eclair-mobile / phoenix (for private channel hints, which could be useful as well).

Invoice are manually generated by eclair-mobile, and we also manually add the routing hints so it could be interesting (for this app) to use a core's API call instead.

btcontract:
Another point to add is that current local channel balances could be remotely probed this way by requesting a series of invoices with different amounts, so this is a trade-off between better privacy and better delivery chances.

OTOH this is already possible by doing a series of to-self payments through particular channels of particular node.

t-bast:
My end goal is to remove that entirely and replace route hints with a rendezvous onion. But I must admit that we probably still want something before that dream comes true.

I would leave the router out of this. In my opinion, you should change the `OutgoingChannel` class. I would add to this class the remote's channel update. This way you only need to call `relayer ! GetOutgoingChannels()` and you have all the information you need to create the routing hints.

btcontract:
I don't see any simple way to get an update into `OutgoingChannel` as it should be a remote, not local update, only router has those. Am I missing something here?

t-bast:
You're right that this needs adding some hooks before that can be done (we need some kind of `LocalChannelUpdate` to be triggered when we receive an update from the remote side).
I'll be working on the `Router` so I'll add that soon, maybe it's worth waiting a bit before we can do clean invoice hints (no need for you to do this refactoring as I'll be changing some things in that area too).

PR #1311, opened 2020-02-10
make Commitments method return Try[]

### LINKED ISSUE #1305: Make Commitments.{sendX, receiveX} methods return a Try
Instead of throwing exceptions. That's an easy change which would simplify `Channel` a bit.

pm47:
Currently, those methods throw exceptions, and we rely on `Channel` to
call them within a `Try(...)`. It puts more burden on `Channel` and
isn't very functional.

Some methods were returning an `Either[]`, which seem to play the role
of a `Try` but isn't used. It seems the idea was to not fail the channel
upon receiving a `fulfill`/`fail` for an unknown htlc, but it is not
fully wired, and isn't compliant (BOLT 2):

> A receiving node:
> if the id does not correspond to an HTLC in its current commitment
transaction:
>    MUST fail the channel.

For signature-related methods, I went with the minimal change of
encapsulating portions of the code inside a `Try {...}` to minimize risk
of regression. We could also make `CommitmentSpec` methods return
`Try[]` but I suspect that would be more complicated with little
benefit.

Note that if some part isn't covered by the `Try` and ends up throwing
an exception, that will be covered by the `handleException` handler of
`Channel`.

Fixes #1305.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -615,23 +615,22 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       handleCommandError(AddHtlcFailed(d.channelId, c.paymentHash, error, origin(c), Some(d.channelUpdate), Some(c)), c)
 
     case Event(c: CMD_ADD_HTLC, d: DATA_NORMAL) =>
-      Try(Commitments.sendAdd(d.commitments, c, origin(c), nodeParams.currentBlockHeight)) match {
-        case Success(Right((commitments1, add))) =>
+      Commitments.sendAdd(d.commitments, c, origin(c), nodeParams.currentBlockHeight) match {
+        case Success((commitments1, add)) =>
           if (c.commit) self ! CMD_SIGN
           context.system.eventStream.publish(AvailableBalanceChanged(self, d.channelId, d.shortChannelId, commitments1))
           handleCommandSuccess(sender, d.copy(commitments = commitments1)) sending add
-        case Success(Left(error)) => handleCommandError(AddHtlcFailed(d.channelId, c.paymentHash, error, origin(c), Some(d.channelUpdate), Some(c)), c)
         case Failure(cause) => handleCommandError(AddHtlcFailed(d.channelId, c.paymentHash, cause, origin(c), Some(d.channelUpdate), Some(c)), c)
       }
 
     case Event(add: UpdateAddHtlc, d: DATA_NORMAL) =>
-      Try(Commitments.receiveAdd(d.commitments, add)) match {
+      Commitments.receiveAdd(d.commitments, add) match {
         case Success(commitments1) => stay using d.copy(commitments = commitments1)
         case Failure(cause) => handleLocalError(cause, d, Some(add))
       }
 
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -643,17 +642,16 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
 
     case Event(fulfill: UpdateFulfillHtlc, d: DATA_NORMAL) =>
-      Try(Commitments.receiveFulfill(d.commitments, fulfill)) match {
-        case Success(Right((commitments1, origin, htlc))) =>
+      Commitments.receiveFulfill(d.commitments, fulfill) match {
+        case Success((commitments1, origin, htlc)) =>
           // we forward preimages as soon as possible to the upstream channel because it allows us to pull funds
           relayer ! Relayer.ForwardFulfill(fulfill, origin, htlc)
           stay using d.copy(commitments = commitments1)
-        case Success(Left(_)) => stay
```
pm47: @t-bast please double-check this thx.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -293,100 +295,100 @@ object Commitments {
...
+      }
     }
   }
 
-  def receiveFail(commitments: Commitments, fail: UpdateFailHtlc): Either[Commitments, (Commitments, Origin, UpdateAddHtlc)] =
+  def receiveFail(commitments: Commitments, fail: UpdateFailHtlc): Try[(Commitments, Origin, UpdateAddHtlc)] =
     getHtlcCrossSigned(commitments, OUT, fail.id) match {
-      case Some(htlc) => Right((addRemoteProposal(commitments, fail), commitments.originChannels(fail.id), htlc))
-      case None => throw UnknownHtlcId(commitments.channelId, fail.id)
+      case Some(htlc) => Success((addRemoteProposal(commitments, fail), commitments.originChannels(fail.id), htlc))
```
t-bast: Can `commitments.originChannels` not contain `fail.id`?
If that happens this will throw inside a Success, so we're not honoring the signature (still throwing an exception instead of wrapping it inside a `Failure`).
This happens in several places.

t-bast: It looks ok to me in the places where we're doing it, but worth another close look.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: This looks fine, the code for `receiveFulfill` never returned a `Left`, so it looks like this was dead code, right?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: > Can commitments.originChannels not contain fail.id?

It should, but we could add:

```suggestion
      case Some(htlc) if commitments.originChannels.contains(fail.id) => Success((addRemoteProposal(commitments, fail), commitments.originChannels(fail.id), htlc))
```

Feels a bit fragile tbh, it is really an issue with the model I think.

PR #1312, opened 2020-02-10
Payment metrics

t-bast:
Consolidate and add payment metrics.
This can be reviewed commit by commit, each commit is scoped to a specific area of the codebase.

That will allow building many graphs that will be helpful when doing A/B testing with path-finding algorithm changes, and will allow easy alerting on many payment KPIs.

Please bikeshed the names, as changing those will require updating all the graphs that rely on them, so it's better to get them right from the start (as much as possible).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/PimpKamon.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/PimpKamon.scala
@@ -18,6 +18,7 @@ package fr.acinq.eclair
 
 import fr.acinq.eclair.payment.{LocalFailure, PaymentFailure, RemoteFailure, UnreadableRemoteFailure}
 import kamon.Kamon
+import kamon.metric.Timer
 import kamon.tag.TagSet
 import kamon.trace.Span
 
@@ -31,6 +32,15 @@ object KamonExt {
       f
     } finally {
       timer.stop()
+    }
+  }
+
+  def time[T](timer: Timer)(f: => T): T = {
+    val started = timer.start()
+    try {
+      f
+    } finally {
+      started.stop()
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/PimpKamon.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/PimpKamon.scala
@@ -25,7 +25,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 object KamonExt {
 
-  def time[T](name: String, tags: TagSet = TagSet.Empty)(f: => T) = {
+  def time[T](name: String, tags: TagSet = TagSet.Empty)(f: => T): T = {
     val timer = Kamon.timer(name).withTags(tags).start()
     try {
       f
@@ -37,7 +37,7 @@ object KamonExt {
   def timeFuture[T](name: String, tags: TagSet = TagSet.Empty)(f: => Future[T])(implicit ec: ExecutionContext): Future[T] = {
     val timer = Kamon.timer(name).withTags(tags).start()
     val res = f
-    res onComplete { case _ => timer.stop }
+    res onComplete (_ => timer.stop)
     res
   }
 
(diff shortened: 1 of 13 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartPaymentFSM.scala
```diff
@@ -98,6 +104,7 @@ class MultiPartPaymentFSM(nodeParams: NodeParams, paymentHash: ByteVector32, tot
         case PaymentSucceeded(parts) =>
           // We expect the parent actor to send us a PoisonPill after receiving this message.
           parent ! MultiPartHtlcSucceeded(paymentHash, parts)
+          Metrics.ReceivedPaymentDuration.withTag(Tags.Success, value = true).record(Platform.currentTime - start, TimeUnit.MILLISECONDS)
```
pm47: This could be factored in another `onTransition`, not sure if worth it.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -121,7 +126,14 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
     case Event(fail: UpdateFailHtlc, WaitingForComplete(s, c, _, failures, sharedSecrets, ignoreNodes, ignoreChannels, hops)) =>
-      Sphinx.FailurePacket.decrypt(fail.reason, sharedSecrets) match {
+      val tryDecrypt = Sphinx.FailurePacket.decrypt(fail.reason, sharedSecrets)
+      tryDecrypt match {
+        case Success(e) =>
+          Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType.get(RemoteFailure(Nil, e))).increment()
+        case Failure(_) =>
+          Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType.get(UnreadableRemoteFailure(Nil))).increment()
+      }
+      tryDecrypt match {
```
pm47: Without intermediate variable (and because I wanted to try multi-line suggestions):

```suggestion
      (Sphinx.FailurePacket.decrypt(fail.reason, sharedSecrets) match {
        case s@Success(e) =>
          Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType.get(RemoteFailure(Nil, e))).increment()
          s
        case f@Failure(_) =>
          Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType.get(UnreadableRemoteFailure(Nil))).increment()
          f
      }) match {
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -1397,6 +1399,11 @@ object Router {
 
     if (localNodeId == targetNodeId) throw CannotRouteToSelf
 
+    val timer = Metrics.FindRouteDuration
```
pm47: Why don't you use `KamonExt.time()`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartPaymentFSM.scala
t-bast: I think it would use more code without adding much value, I'll leave it unchanged for now.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Done!

t-bast:
NB: I'll do the rebase/merge after #1324 since it's going to land soon and will create conflicts.

t-bast:
NB: I'll do the rebase/merge after #1324 since it's going to land soon and will create conflicts.

t-bast:
NB: I'll do the rebase/merge after #1324 since it's going to land soon and will create conflicts.

eclair-core/src/main/scala/fr/acinq/eclair/PimpKamon.scala
```diff
@@ -34,10 +35,19 @@ object KamonExt {
     }
   }
 
+  def time[T](timer: Timer)(f: => T): T = {
+    val started = timer.start()
```
pm47: I think `started` is a better name but you should have changed the method above too for consistency

PR #1313, opened 2020-02-10
Avoid illegal reflective operation during startup

araspitzu:
~~This PR rewords the way we load plugins providing a stricter and safer way to load them. We now require plugins to exhibit a MANIFEST file with some well defined entries:~~
~~- `Main-Class` to identify the plugin entry point ( class implementing `Plugin`)~~
~~- `X-Eclair-Supported-Versions` contains a list of all supported eclair version~~

~~The `Main-Class` is mainly used to load the plugin without scanning using `org.clapper.classutils`, this allows us to avoid an illegal reflective call during startup. The list of supported version allows us to check if the plugin we're loading supports the current version of eclair, this prevents from running outdated (and incompatible) plugins.~~

This PR slightly reworks the way we load plugins to avoid doing an illegal reflective call. Instead of using org.clapper.classutils to scan for classes (and perform the illegal reflective call) we now require the plugin to tell us which class is implementing the Plugin interface via MANIFEST file.

### DIFF
--- a/eclair-node/src/main/scala/fr/acinq/eclair/Plugin.scala
+++ b/eclair-node/src/main/scala/fr/acinq/eclair/Plugin.scala
@@ -40,24 +40,21 @@ object Plugin extends Logging {
     * @return
     */
   def loadPlugins(jars: Seq[File]): Seq[Plugin] = {
-    val urls = jars.flatMap(f =>
-      getOrNone(new URL(s"jar:file:${f.getCanonicalPath}!/").openConnection().asInstanceOf[JarURLConnection], s"unable to load plugin file:${f.getAbsolutePath} ")
-    )
+    val urls = jars.flatMap(openJar)
     val loader = new URLClassLoader(urls.map(_.getJarFileURL).toArray, ClassLoader.getSystemClassLoader)
     val pluginClasses = urls
-        .map(_.getMainAttributes.getValue("Main-Class"))
-        .flatMap(classFQDN => getOrNone[Class[_]](loader.loadClass(classFQDN), s"error loading $classFQDN "))
-        .flatMap(c => getOrNone[Plugin](c.getDeclaredConstructor().newInstance().asInstanceOf[Plugin], s"${c.getCanonicalName} does not implement $Plugin "))
+        .flatMap(j => Option(j.getMainAttributes.getValue("Main-Class")))
+        .flatMap(classFQDN => Try[Class[_]](loader.loadClass(classFQDN)).toOption)
+        .flatMap(c => Try(c.getDeclaredConstructor().newInstance().asInstanceOf[Plugin]).toOption)
 
     logger.info(s"loading ${pluginClasses.size} plugins")
     pluginClasses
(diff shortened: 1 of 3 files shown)


eclair-node/src/main/scala/fr/acinq/eclair/Plugin.scala
```diff
@@ -29,17 +31,48 @@ trait Plugin {
...
+    pluginClasses
+  }
+
+  def getOrNone[T](f: => T, errMsg: String): Option[T] = Try(f) match {
+    case Failure(exception) =>
+      logger.error(errMsg, exception)
+      None
+    case Success(value) => Some(value)
+  }
+
```
pm47: Minor refactoring. It feels weird to define a generic `getOrNone`:

```suggestion
    val eclairVersion = Option(getClass.getPackage.getImplementationVersion) // will be None if this is a snapshot

    val urls = jars.flatMap(openJar)
    val loader = new URLClassLoader(urls.map(_.getJarFileURL).toArray, ClassLoader.getSystemClassLoader)

    val (supported, unsupported) = urls.partition { jar =>
      val supportedVersions = Option(jar.getMainAttributes.getValue("X-Eclair-Supported-Versions")).map(_.split(",")).toSeq.flatten
      val versionOk = eclairVersion.isEmpty || supportedVersions.contains(eclairVersion.get)
      val mainClassOk = Option(jar.getMainAttributes.getValue("Main-Class")).isDefined
      versionOk && mainClassOk
    }

    unsupported.foreach(jar => logger.warn(s"found unsupported plugin ${jar.getJarFile.getName}"))
    supported.foreach(jar => logger.info(s"loading plugin ${jar.getJarFile.getName}"))

    val pluginClasses = supported
      .map(_.getMainAttributes.getValue("Main-Class"))
      .map(loader.loadClass)
      .map(_.getDeclaredConstructor().newInstance().asInstanceOf[Plugin])

    pluginClasses
  }

  def openJar(jar: File): Option[JarURLConnection] =
    Try(new URL(s"jar:file:${jar.getCanonicalPath}!/").openConnection().asInstanceOf[JarURLConnection]) match {
      case Success(url) => Some(url)
      case Failure(t) => logger.error(s"unable to load plugin file:${jar.getAbsolutePath} ", t); None
    }

```

eclair-node/src/main/scala/fr/acinq/eclair/Plugin.scala
```diff
@@ -29,17 +31,53 @@ trait Plugin {
 
 }
 
-object Plugin {
+object Plugin extends Logging {
+
+  val versionRegex = """^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)([-SNAPSHOT]*)""".r
```
pm47: This is a bit fragile, and doesn't match the version 0.3 that we have released last year.

araspitzu:
The PR has been updated to not apply any change regarding the version checks of the plugins.

PR #1315, opened 2020-02-11
Electrum: add support for socks5 proxies

sstone:
We use the same socks5 proxy that is optionally enabled in the configuration and is typically used to connect to LN nodes running as TOR hidden services.

This will allow eclair:
- to connect to .onion Electrum servers running behind TOR
- to connect to regular Electrum servers through TOR


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumClient.scala
```diff
@@ -93,10 +96,12 @@ class ElectrumClient(serverAddress: InetSocketAddress, ssl: SSL)(implicit val ec
       ch.pipeline.addLast(new JsonRPCRequestEncoder)
       // error handler
       ch.pipeline.addLast(new ExceptionHandler)
+      socksProxy_opt.foreach(params => ch.pipeline().addFirst(new Socks5ProxyHandler(params.address)))
     }
   })
 
   // Start the client.
+  socksProxy_opt.foreach(params => b.resolver(NoopAddressResolverGroup.INSTANCE))
```
t-bast: Can you explain that? Maybe a comment? Browsing to it in code doesn't reveal anything very useful...

sstone:
Yes, with a manual test using Tor's default proxy (just setting `eclair.socks5.enabled` to `true` should be enough)

Issue #1314, opened 2020-02-11
Feature request: Use getzmqnotifications for the ZMQ addresses


Thoragh:
I think it would be nice if eclair used the getzmqnotifications RPC method in bitcoin to get the ZMQ addresses. It would mean users has less things to write in the `eclair.conf` file.

t-bast:
Interesting, that's something that could be tried. This is quite a crucial part of the application so that change would need to be thoroughly tested.
We're open to external contributions if you want to do that!

pm47:
Really like the idea.

sstone:
What are the limitations of using this rpc call ?

- What does it return ? Is it just whatever is configured in bitcoin.conf or is it smarter (as in: don't return something when it there was a zmq error)
- How does it work when the bitcoin node is not on the same machine as the LN node and is accessed through its IP address ?
- How does it work when the bitcoin node is not on the same machine and is accessed with SSH and port forwarding ? (this is a very common setup for testing and afaik the recommended way of accessing remote bitcoin node)

t-bast:
@Thoragh could you help answer sstone's questions?

Thoragh:
>     * What does it return ? Is it just whatever is configured in bitcoin.conf or is it smarter (as in: don't return something when it there was a zmq error)
The rpc will only return something it could successfully bind all addresses. The addresses are returned in the same way they are writen in bitcoin.conf.
>     * How does it work when the bitcoin node is not on the same machine as the LN node and is accessed through its IP address ?
If the the address in bitcoin.conf was set to for example tcp://192.168.0.123:29000 and the machines are in the same network it should work. If 0.0.0.0 was used it will not work.
>     * How does it work when the bitcoin node is not on the same machine and is accessed with SSH and port forwarding ? (this is a very common setup for testing and afaik the recommended way of accessing remote bitcoin node)
If the ports are the same for example `ssh -L 29000:127.0.0.1:29000 user@host` it should work I think.

There are cases where using the rpc call wont work so the current zmqblock and zmqtx config options need to stay. The default fallback should in my opinion be to use the rpc call if the user has not specified the zmq addresses.

t-bast:
What's the status of this?
To be honest I think the additional code required for this is not worth the effort, especially since the fallback needs to stay.

t-bast:
Closing this issue. Since the fallback will always be needed, I don't think this would be that much of an improvement considering the extra code it adds.

PR #1316, opened 2020-02-11
Use 'getzmqnotifications' to get ZMQ addresses from bitcoind

araspitzu:
Replaces the configuration keys for the ZMQ with an RPC call to obtain the addresses from bitcoind, see #1314

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -52,7 +52,7 @@ import fr.acinq.eclair.tor.TorProtocolHandler.OnionServiceVersion
 import fr.acinq.eclair.tor.{Controller, TorProtocolHandler}
 import fr.acinq.eclair.wire.NodeAddress
 import grizzled.slf4j.Logging
-import org.json4s.JsonAST.JArray
+import org.json4s.JsonAST.{JArray, JString}
 import scodec.bits.ByteVector
 
 import scala.concurrent._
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -247,11 +247,21 @@ class Setup(datadir: File,
...
+          case _ => (None, None)
+        }
+        case _ => Future.successful((None, None))
+      }
 
       watcher = bitcoin match {
         case Bitcoind(bitcoinClient) =>
-          system.actorOf(SimpleSupervisor.props(Props(new ZMQActor(config.getString("bitcoind.zmqblock"), Some(zmqBlockConnected))), "zmqblock", SupervisorStrategy.Restart))
-          system.actorOf(SimpleSupervisor.props(Props(new ZMQActor(config.getString("bitcoind.zmqtx"), Some(zmqTxConnected))), "zmqtx", SupervisorStrategy.Restart))
+          system.actorOf(SimpleSupervisor.props(Props(new ZMQActor(zmqBlocks_opt.get, Some(zmqBlockConnected))), "zmqblock", SupervisorStrategy.Restart))
```
pm47: This looks like it will throw cryptic exceptions when bitcoin core is started with zmq disabled.

araspitzu: At 3d37e08 we check beforehand to avoid the cryptic exception in case the addresses are missing

sstone: using `.get` on an `Option` is almost always a code smell (I don't think there is another instance of this in our code ?)

sstone:
This feels a bit rushed, see my comments on #1314. Before we implement changes we should first ask ourselves "why" and "how" ?. For example, I use SSH/port forwarding a lot to test against remote bitcoin nodes, I'm pretty sure that it will not work with your current changes.
=> Can we really use the values returned by `getzmqnotifications` or do we just use them to check that zmq notifications are available (it's a common issue with users who build bitcoin core without zmq support) but keep reading values from our own configuration file ?

araspitzu:
> This feels a bit rushed, see my comments on #1314. Before we implement changes we should first ask ourselves "why" and "how" ?. For example, I use SSH/port forwarding a lot to test against remote bitcoin nodes, I'm pretty sure that it will not work with your current changes.
> => Can we really use the values returned by `getzmqnotifications` or do we just use them to check that zmq notifications are available (it's a common issue with users who build bitcoin core without zmq support) but keep reading values from our own configuration file ?

Yes it was a bit rushed, I liked the idea and I thought it's a low hanging fruit but we need to make sure it works we all the setup/scenario we want. I did some testing and the values we get from `getzmqnotifications` are just those in the conf file, if the user doesn't set `#zmqpubrawblock=` then the result of the API call contains only the entry for the rawtx:
```
[
  {
    "type": "pubrawtx",
    "address": "tcp://127.0.0.1:29002",
    "hwm": 1000
  }
]
```
I agree it needs more testing especially the port forwarding setup.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
araspitzu: I'll have a look on how to improve that. Yes it is a bit of a code smell but I think it's acceptable only if there is a `nonEmpty` check very close by (because of readability), unfortunately we do have several instances of `.get` in the codebase.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -247,11 +247,24 @@ class Setup(datadir: File,
           feeratesRetrieved.trySuccess(Done)
       })
       _ <- feeratesRetrieved.future
+      (zmqBlocks_opt, zmqTxs_opt) <- bitcoin match {
+        case Bitcoind(client) => client.invoke("getzmqnotifications").map {
```
pm47: This should be done before, along with all other bitcoind-related validations. See:

https://github.com/ACINQ/eclair/blob/use_getzmqnotification-pm/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala

araspitzu:
> This feels a bit rushed, see my comments on #1314. Before we implement changes we should first ask ourselves "why" and "how" ?. For example, I use SSH/port forwarding a lot to test against remote bitcoin nodes, I'm pretty sure that it will not work with your current changes.
> => Can we really use the values returned by `getzmqnotifications` or do we just use them to check that zmq notifications are available (it's a common issue with users who build bitcoin core without zmq support) but keep reading values from our own configuration file ?

It returns the socket address for the ZMQ notifications if their socket was successfully bound. There are some differences between 0.18 and 0.19 where with 0.18 the API doesn't return an address if it there were errors and 0.19 doesn't start bitcoind at all if there are errors binding to the selected socket. 

SSH tunneling works as it is completely transparent for the application, however `getzmqnotifications`
 will return the socket address that is used on the remote machine so users will need to tune their 
 tunneling accordingly, which means they must use the same local port as in the remote server. I think this makes it a bit inconvenient in complex setups where to connect to multiple remote nodes one has to select different local port for each of them.

Another major, and blocking, issue is that using a wildcard address such as `zmqpubrawtx=tcp://0.0.0.0:28333` in your bitcoin.conf results in `getzmqnotifications` outputting the wildcard address and that is just not usable (notably it breaks most docker and docker-compose builds).

araspitzu:
Considering that we can't rely only on `getzmqnotifications` we must keep the fallback configuration keys; however this API could be used to do a first attempt to connect and this might help in some setups to reduce (albeit by little) the amount of configuration needed for eclair (we still need to mention the fallback in the README though)

PR #1317, opened 2020-02-11
Always re-send ChannelUpdate to private channels

btcontract:
Current approach seems to be too brittle as `ChannelUpdate` is only sent once but may not reach a remote peer due to disconnect (this did happen to me). 

This is what other implementations (LND at least) do already, I suppose that traffic overhead of this would be minimal. If it is a concern though then perhaps a following smarter strategy could work: send `ChannelUpdate` to private channel on each reconnect until there was an incoming payment attempt (respected commit number updated in commitments?).


eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -2384,12 +2384,12 @@ class NormalStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
...
     sender.send(alice, WatchEventConfirmed(BITCOIN_FUNDING_DEEPLYBURIED, 400000, 42, null))
     val update1a = alice2bob.expectMsgType[ChannelUpdate]
-    assert(Announcements.isEnabled(update1a.channelFlags) == true)
+    assert(Announcements.isEnabled(update1a.channelFlags))
 
     // actual test starts here
     sender.send(alice, INPUT_DISCONNECTED)
     awaitCond(alice.stateName == OFFLINE)
-    alice2bob.expectNoMsg(1 second)
+    alice2bob.expectMsgType[ChannelUpdate]
```
t-bast: That sounds undesirable: alice is going offline, why would she re-broadcast her channel update?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -2409,9 +2409,7 @@ class NormalStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     sender.send(alice, INPUT_DISCONNECTED)
     assert(relayerA.expectMsgType[Status.Failure].cause.asInstanceOf[AddHtlcFailed].paymentHash === htlc1.paymentHash)
     assert(relayerA.expectMsgType[Status.Failure].cause.asInstanceOf[AddHtlcFailed].paymentHash === htlc2.paymentHash)
-    val update2a = alice2bob.expectMsgType[ChannelUpdate]
-    assert(channelUpdateListener.expectMsgType[LocalChannelUpdate].channelUpdate === update2a)
-    assert(!Announcements.isEnabled(update2a.channelFlags))
```
btcontract: Looks like this was catching a superfluous `ChannelUpdate` on entering into `OFFLINE` state, it's not sent now.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1678,16 +1678,15 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-        case (d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && !d1.buried && d2.buried =>
-          // for a private channel, when the tx was just buried we need to send the channel_update to our peer (even if it didn't change)
+      (state, nextState, stateData, nextStateData) match {
+        // if channel is private, we send the channel_update directly to remote on each reconnect
+        // they need it "to learn the other end's forwarding parameters" (BOLT 7)
+        case (NORMAL, NORMAL, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && !d1.buried && d2.buried =>
           forwarder ! d2.channelUpdate
-        case (d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && d1.channelUpdate != d2.channelUpdate && d2.buried =>
-          // otherwise, we only send it when it is different, and tx is already buried
+        case (OFFLINE, NORMAL, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && d2.buried =>
```
t-bast: I believe you still need the other case: when you're in state normal and update your fee policy, you do want to re-broadcast the new `channel_update`.

I'm also not sure the case you added will ever happen, because you're using `d1: DATA_NORMAL` from state `OFFLINE`...that requires more tests but doesn't look correct.

btcontract: Unless I miss something `DATA_NORMAL` is the one which present in `OFFLINE` state and `OFFLINE -> NORMAL` transition?

t-bast: Woops you're totally right, I'll have another look at that.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1678,16 +1678,15 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
 
-      // if channel is private, we send the channel_update directly to remote
-      // they need it "to learn the other end's forwarding parameters" (BOLT 7)
-      (stateData, nextStateData) match {
-        case (d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && !d1.buried && d2.buried =>
-          // for a private channel, when the tx was just buried we need to send the channel_update to our peer (even if it didn't change)
+      (state, nextState, stateData, nextStateData) match {
+        // if channel is private, we send the channel_update directly to remote on each reconnect
+        // they need it "to learn the other end's forwarding parameters" (BOLT 7)
+        case (NORMAL, NORMAL, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && (!d1.buried && d2.buried || d1.channelUpdate != d2.channelUpdate) =>
```
t-bast: I think that's incorrect. You still always want to wait for `d2.buried` before sending a `channel_update` (which isn't the case here if `d1.channelUpdate != d2.channelUpdate`.

I'm a bit weary of matching now on states + data, there may be a case we're losing compared to the previous code.

Here is my take:

```scala
// if channel is private, we send the channel_update directly to remote
// they need it "to learn the other end's forwarding parameters" (BOLT 7)
  case (_, _, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && !d1.buried && d2.buried =>
    // for a private channel, when the tx was just buried we need to send the channel_update to our peer (even if it didn't change)
    forwarder ! d2.channelUpdate
  case (OFFLINE, NORMAL, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && d2.buried =>
    // otherwise if we're coming back online, we rebroadcast the latest channel_update
    // this makes sure that if the channel_update was missed, we have a chance to re-send it
    forwarder ! d2.channelUpdate
  case (_, _, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && d1.channelUpdate != d2.channelUpdate && d2.buried =>
    // otherwise, we only send it when it is different, and tx is already buried
    forwarder ! d2.channelUpdate
  case _ => ()
}
```

I believe it's safer as it keeps the existing behavior and only adds a new case when coming back online. WDYT?

btcontract: > I think that's incorrect. 

It's less code and not sure it's strictly incorrect: `NORMAL_DATA` is present so at least default 3 confirmations are there, occasional premature update at this point won't add too much risk.

But I'm also totally fine with using your code.

PR #1318, opened 2020-02-11
Fix random trampoline test failure

t-bast:
Writing the PaymentRelayed event to the DB is not atomic.
Each part is written independently of the others (one row per part).
This is fine as nothing relies on this event being written atomically.
However tests were expecting that and we would observe rare test failures on travis.


araspitzu:
It looks correct to me, because of the non-atomicity of writing `PaymentRelayed` we could end up reading the object before all its parts were persisted, the extra check `relayed.head.amountOut >= amount` ensures that all the parts were actually written to the db because if it was otherwise the `amountOut` would be less than the expected amount.

t-bast:
@araspitzu could you approve to unblock the merge? Thanks!!

PR #1320, opened 2020-02-12
Add PaymentsDb helper to extract success

t-bast:
Since we internally retry failed payment attempts, listing from the DB may yield some rows that are failed (even when the payment succeeded).

We now provide a helper to extract only the successful parts with the relevant fee information.


t-bast:
Closing as this should probably not be done in eclair-core, but rather in Phoenix or Eclair-Mobile.

PR #1321, opened 2020-02-12
Channel sends typed responses

t-bast:
Instead of sending strings, channel now sends typed responses.
This is more future-proof when we want to add data to those responses.

type-all-the-things

TODO:

- [x] E2E tests to verify nothing funky is happening.


t-bast:
Looks like I have to do some json serialization magic to make it smooth in the CLI, will add that in a separate commit.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -144,19 +145,19 @@ class EclairImpl(appKit: Kit) extends Eclair {
       pushMsat = pushAmount_opt.getOrElse(0 msat),
       fundingTxFeeratePerKw_opt = fundingFeerateSatByte_opt.map(feerateByte2Kw),
       channelFlags = flags_opt.map(_.toByte),
-      timeout_opt = Some(openTimeout))).mapTo[String]
+      timeout_opt = Some(openTimeout))).mapTo[ChannelCommandResponse].map(_.toString)
```
pm47: You could return a `ChannelCommandResponse` and define a json deserializer that does a `toString`. Not sure if that'd be better.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -84,13 +83,20 @@ object Channel {
...
 
   // @formatter:off
-  sealed trait ChannelError
-  case class LocalError(t: Throwable) extends ChannelError
-  case class RemoteError(e: Error) extends ChannelError
+  sealed trait ChannelOpenError
+  case class LocalError(t: Throwable) extends ChannelOpenError
+  case class RemoteError(e: Error) extends ChannelOpenError
+
+  sealed trait ChannelCommandResponse
```
pm47: Commands are defined in `ChannelTypes.scala` and responses in `Channel.scala`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1733,7 +1739,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   /**
    * This function is used to return feedback to user at channel opening
    */
-  def replyToUser(message: Either[Channel.ChannelError, String]): Unit = {
+  def replyToUser(message: Either[ChannelOpenError, ChannelCommandResponse]): Unit = {
```
pm47: Maybe this method should be renamed to make it clear this is used at the channel opening.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: It probably makes sense to return a `ChannelCommandResponse` and push the conversion to string (which is lossy) closer to the boundary of the program.

pm47: Yes. The disadvantage being that there is a symmetry problem given that this function doesn't take a `Command`. I believe we are in bikeshed territory.

t-bast: True, but I think it's quite ok there :)

pm47:
Needs merge/rebase

PR #1323, opened 2020-02-17
Wumbo support

araspitzu:
Add support for the upcoming feature `option_support_large_channel` to create channels with large funding amount. To make sure we safely handle the funding transaction we use the following formula to compute the `min_depth`: `(((15 * funding_amount) / 6.25).ceil + 1)` this ensures that the cumulative block reward that will be issued while we wait for the channel *largely exceeds* the funding amount. There is also a new config key `max-funding-satoshis` that can be used to limit the funding amount even above the wumbo limit. Because in a wumbo world channels can be much bigger than the current size the routing heuristics have been updated as well: we now consider a 100k sat funding size to be "small" and a 1BTC channel size to be "large" (i'd like to have some feedback on those numbers).


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -351,8 +351,8 @@ object Graph {
     val BLOCK_TIME_TWO_MONTHS = 8640
 
     // Low/High bound for channel capacity
-    val CAPACITY_CHANNEL_LOW = (1000 sat).toMilliSatoshi
-    val CAPACITY_CHANNEL_HIGH = Channel.MAX_FUNDING.toMilliSatoshi
+    val CAPACITY_CHANNEL_LOW = Btc(0.001).toMilliSatoshi
```
t-bast: Honestly I wouldn't change this right now.
I'd first wait to see wumbo channels appear in the network, then run path-finding tests on the network DB for this change before merging it.

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -204,8 +204,8 @@ class IntegrationSpec extends TestKit(ActorSystem("test")) with BitcoindService
     connect(nodes("C"), nodes("F3"), 5000000 sat, 0 msat)
     connect(nodes("C"), nodes("F4"), 5000000 sat, 0 msat)
     connect(nodes("C"), nodes("F5"), 5000000 sat, 0 msat)
-    connect(nodes("B"), nodes("G"), 16000000 sat, 0 msat)
-    connect(nodes("G"), nodes("C"), 16000000 sat, 0 msat)
+    connect(nodes("B"), nodes("G"), 46700000 sat, 0 msat)
```
t-bast: I would test the wumbo channel on an Fn channel instead.
Changing the capacity of those is probably messing with the MPP tests (that now may become single-part since there's a lot of capacity). That may be an indication that the MPP/Trampoline tests should be more assertive that multiple parts are actually used...

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -430,13 +430,13 @@ class RouteCalculationSpec extends FunSuite with ParallelTestExecution {
     val extraHop4 = ExtraHop(d, ShortChannelId(4), 50.sat.toMilliSatoshi, 0, CltvExpiryDelta(42))
     val extraHops = extraHop1 :: extraHop2 :: extraHop3 :: extraHop4 :: Nil
 
-    val amount = 900 sat // below RoutingHeuristics.CAPACITY_CHANNEL_LOW
+    val amount = 90000 sat // below RoutingHeuristics.CAPACITY_CHANNEL_LOW
```
t-bast: As said in my other comment, I think it's a bit premature...

pm47:
+1 for scaling the `min_depth` according to the size of the channel, I think it's a must have.

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
araspitzu: Right! This has been addressed in b70726d, we now create only 1 wumbo channel between C-F1 and we also send a large HTLC (0.4BTC) through it.

araspitzu:
In 9ca1fb9 we compute the `min_depth` according to the size of the channel, the heuristics followed is to make sure the cumulative block reward exceeds the funding (plus one block for extra security). We use the max between the computed `min_depth` and the user configured `mindepth-blocks`, unfortunately it's limited to mainnet and a bit hard to test.

araspitzu:
061220b adds the new config key `eclair.max-funding-satoshis` to set the max allowed funding size, which can be below or over the current limit. Together with the adjustable `min_depth` this should allow for enough fine grained control of the channel size and security parameters.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: I think it doesn't hurt, the values here are used only to compute the heuristic score to make more informed decision in path finding, whatever value we choose here it never excludes the channel from the result set. There is also the argument that currently we consider a small channel a size of 1000sat which is really tiny and realistically not usable from a routing PoV, it seems to me that small channels in the network are slowly disappearing and updating this value here can help to speed up the process because they will be less likely to be selected.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -308,14 +309,22 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
...
-      log.info(s"requesting a new channel with fundingSatoshis=${c.fundingSatoshis}, pushMsat=${c.pushMsat} and fundingFeeratePerByte=${c.fundingTxFeeratePerKw_opt} temporaryChannelId=$temporaryChannelId localParams=$localParams")
-      channel ! INPUT_INIT_FUNDER(temporaryChannelId, c.fundingSatoshis, c.pushMsat, channelFeeratePerKw, fundingTxFeeratePerKw, localParams, d.transport, d.remoteInit, c.channelFlags.getOrElse(nodeParams.channelFlags), ChannelVersion.STANDARD)
-      stay using d.copy(channels = d.channels + (TemporaryChannelId(temporaryChannelId) -> channel))
+      if (c.fundingSatoshis > Channel.MAX_FUNDING && !Features.hasFeature(nodeParams.features, Wumbo)) {
+        sender() ! s"fundingSatoshis=${c.fundingSatoshis} is too big, you must enable wumbo to use funding above ${Channel.MAX_FUNDING}"
+        stay()
+      } else if (c.fundingSatoshis > nodeParams.maxFundingSatoshis) {
+        sender() ! s"fundingSatoshis=${c.fundingSatoshis} is too big for the current settings, increase 'eclair.max-funding-satoshis'"
```
t-bast: nit:
```suggestion
        sender ! s"fundingSatoshis=${c.fundingSatoshis} is too big for the current settings, increase 'eclair.max-funding-satoshis' in your eclair.conf"
```

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -107,6 +107,7 @@ object TestConstants {
       paymentRequestExpiry = 1 hour,
       multiPartPaymentExpiry = 30 seconds,
       minFundingSatoshis = 1000 sat,
+      maxFundingSatoshis = Btc(1),
```
t-bast: Maybe also activate wumbo in Alice and bob's features if you put a wumbo value in here?
Or keep `16777216 sat` here to preserve existing tests, and only override it in wumbo-specific tests?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2250,6 +2250,24 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+  /**
+    * Returns the amount of confirmation needed to safely handle the funding transaction,
+    * we makes sure the cumulative block reward exceeds the channel size.
+    *
+    * @param fundingSatoshis funding amount of the channel
+    * @return number of confirmation needed
+    */
+  def minDepthForFunding(fundingSatoshis:Satoshi): Long = fundingSatoshis match {
+    // if this is not a wumbo channel or we're not on mainnet use the default
+    case f if f <= MAX_FUNDING || nodeParams.chainHash != Block.LivenetGenesisBlock.hash =>
```
t-bast: nit: I think the check on the `chainHash` is unnecessary, IMO we do want to be able to test the scaling logic on `regtest`/`testnet`.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
```diff
@@ -43,19 +45,33 @@ class WaitForAcceptChannelStateSpec extends TestkitBaseClass with StateTestsHelp
...
+
     import setup._
     val channelVersion = ChannelVersion.STANDARD
     val (aliceParams, bobParams) = (Alice.channelParams, Bob.channelParams)
     val aliceInit = Init(aliceParams.features)
     val bobInit = Init(bobParams.features)
     within(30 seconds) {
-      alice ! INPUT_INIT_FUNDER(ByteVector32.Zeroes, TestConstants.fundingSatoshis, TestConstants.pushMsat, TestConstants.feeratePerKw, TestConstants.feeratePerKw, aliceParams, alice2bob.ref, bobInit, ChannelFlags.Empty, channelVersion)
-      bob ! INPUT_INIT_FUNDEE(ByteVector32.Zeroes, bobParams, bob2alice.ref, aliceInit)
+      if(test.tags.contains("wumbo")){
```
t-bast: nit: you could factor those two (probably lacks an explicit conversion Btc->sat):
```scala
val fundingAmount = if(test.tags.contains("wumbo")) Btc(50) else TestConstants.fundingSatoshis
alice ! INPUT_INIT_FUNDER(ByteVector32.Zeroes, fundingAmount, TestConstants.pushMsat, TestConstants.feeratePerKw, TestConstants.feeratePerKw, aliceParams, alice2bob.ref, bobInit, ChannelFlags.Empty, channelVersion)
bob ! INPUT_INIT_FUNDEE(ByteVector32.Zeroes, bobParams, bob2alice.ref, aliceInit)
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForOpenChannelStateSpec.scala
```diff
@@ -36,7 +37,10 @@ class WaitForOpenChannelStateSpec extends TestkitBaseClass with StateTestsHelper
   case class FixtureParam(bob: TestFSMRef[State, Data, Channel], alice2bob: TestProbe, bob2alice: TestProbe, bob2blockchain: TestProbe)
 
   override def withFixture(test: OneArgTest): Outcome = {
-    val setup = init()
+    val setup = test.tags.toList match {
```
t-bast: nit: why not the simpler `if (test.tags.contains("wumbo"))`?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForOpenChannelStateSpec.scala
```diff
@@ -88,6 +92,15 @@ class WaitForOpenChannelStateSpec extends TestkitBaseClass with StateTestsHelper
     awaitCond(bob.stateName == CLOSED)
   }
 
+  test("recv OpenChannel (fundingSatoshis > max-funding-satoshis)", Tag("wumbo")) { f =>
+    import f._
+    val open = alice2bob.expectMsgType[OpenChannel]
+    val highFundingSat = Btc(2).toSatoshi // bob has max-funding-satoshis = 1BTC
```
t-bast: Maybe more future-proof:
```suggestion
    val highFundingSat = Bob.nodeParams.maxFundingSatoshis + Btc(1)
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: Allright then! But do you have a way of running tests on the real network? Changing the path-finding heuristics should (IMO) always be backed by hard evidence that we're not making things worse for real users...

t-bast:
Once all the current comments are resolved, I really think we need to add validation on the fundee side. If Mallory opens a 50 BTC channel to me with a confirmation of 6 blocks, then sends a lot of HTLCs that I forward, then double-spend the channel, I will have lost all the forwarded amount.

IMO the easiest solution is to delay our own `funding_confirmed` by the `minFundingDepth` that we calculate. This is easier than negotiating by rejecting `open_channel` requests that set a depth too low, because there's no built-in mechanism in the spec to reject while saying "I would be ok if minDepth was XXX".

eclair-core/src/main/resources/reference.conf
```diff
@@ -109,6 +109,7 @@ eclair {
   payment-request-expiry = 1 hour // default expiry for payment requests generated by this node
   multi-part-payment-expiry = 60 seconds // default expiry for receiving all parts of a multi-part payment
   min-funding-satoshis = 100000
+  max-funding-satoshis = 100000000 // 1BTC
```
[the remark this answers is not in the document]
araspitzu: The idea was to have wumbo as first limit and another max-funding-satoshis with a relatively high limit in case wumbo is enabled. But i like this solution too, we default to the wumbo limit and force the user to increase the max funding accordingly, also the default is also used in the test now (as of a65c090) .

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2250,6 +2250,24 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+      nodeParams.minDepthBlocks
+    case f if f > MAX_FUNDING =>
+      // pre halving each block yields 12.5BTC, post 6.25BTC
+      val blockReward = if(nodeParams.currentBlockHeight <= 630000) 12.5 else 6.25
+      val blocksToReachFunding = ((f.toBtc.toDouble / blockReward).ceil + 1).toInt
```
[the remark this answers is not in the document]
araspitzu: I like this! I was a bit hesitant to introduce a scaling factor but a more conservative approach was needed, i used you suggestion of scaling_factor=15, we can also extract it into a configuration key if it's too long for the users.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
araspitzu: It was used to make a correct guess on the actual block reward but since we're hard-coding it we don't need the `chainHash` anymore.

araspitzu:
> Once all the current comments are resolved, I really think we need to add validation on the fundee side. If Mallory opens a 50 BTC channel to me with a confirmation of 6 blocks, then sends a lot of HTLCs that I forward, then double-spend the channel, I will have lost all the forwarded amount.
> 
> IMO the easiest solution is to delay our own `funding_confirmed` by the `minFundingDepth` that we calculate. This is easier than negotiating by rejecting `open_channel` requests that set a depth too low, because there's no built-in mechanism in the spec to reject while saying "I would be ok if minDepth was XXX".

I agree we need to take care of the fundee side but I'm not sure about delaying the `funding_confirmed` message, it's a good solution but isn't it enough to reject with a human readable error containing our parameters for `min_depth`?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: I do :) , i use [this](https://github.com/ACINQ/eclair/blob/graph_fuzzy_testing/eclair-core/src/test/scala/fr/acinq/eclair/router/FuzzyGraphSpec.scala#L88) to load any networkDB and run some path-finding on it, I'll report the results here.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForOpenChannelStateSpec.scala
```diff
@@ -177,6 +193,15 @@ class WaitForOpenChannelStateSpec extends TestkitBaseClass with StateTestsHelper
     awaitCond(bob.stateName == CLOSED)
   }
 
+  test("recv OpenChannel (wumbo size)", Tag("wumbo"), Tag("max-funding-satoshis")) { f =>
+    import f._
+    val open = alice2bob.expectMsgType[OpenChannel]
+    val highFundingSat = Btc(1).toSatoshi
+    bob ! open.copy(fundingSatoshis = highFundingSat)
+    bob2alice.expectMsgType[AcceptChannel]
```
t-bast: I think we're not yet testing that when we accept a wumbo channel, we correctly set the `min-depth` in our accept message to something that scaled on the amount of the channel.

Maybe here we should use something bigger than 1 BTC (that should have a min-depth > 6) and assert that the `AcceptChannel` sent by Bob correctly sets that `min-depth`?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
araspitzu: Here we go: 

The simulations have been run on a up-to-date mainnet db, with a sample size of 1000 randomly chosen nodes used to run 500 path-finds. The test has been repeated with different amounts to evaluate how the heuristic works with different payment size (recall the new wumbo heuristics slightly privileges larger channel) and with a fixed starting point (ACINQ node). The test are reproducible running `FuzzyGraphSpec` in this [branch](https://github.com/ACINQ/eclair/tree/graph_fuzzy_testing), the results are printed in 3 different files and include the actual route that was found. Note that it's okay for the test to not find a route between the nodes, they have been randomly chosen and there might not be a route connecting them.

| TEST                                                 | routes found(master) | routes found(wumbo) | different routes found |
|------------------------------------------------------|----------------------|---------------------|------------------------|
| 1000sat payment, 2 random nodes       | 242                  | 242                 | 58                     |
| 100000sat payment, 2 random nodes     | 164                  | 164                 | 40                     |
| 1000sat payment, ACINQ -> random node | 443                  | 443                 | 55                     |

There are zero cases where the new heuristics would fail in finding a route that was previously found, where the new heuristics finds a different route that is on average shorter than the one found on master (without the wumbo heuristic)

t-bast: Great thanks, let's keep that change then

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForOpenChannelStateSpec.scala
araspitzu: Uhm I believe that's done in [`WaitForAcceptChannelStateSpec`](https://github.com/ACINQ/eclair/pull/1323/files#diff-42debadc7a9bf325bd7ceaa8545be205R162)? Here we do test the usage of `min_depth` in `AcceptChannel` and raw invocation of `minDepthForFunding`

t-bast: Not really because this test expects the result to be the default conf value of 6.
We can't know that the scaling factor was applied and then rounded to 6 because it was lower than that.
If you change that test to use a funding of 5 BTC (and thus min-depth of 20) then that would work.

t-bast: But I agree that the test you're pointing to is likely the best place to test this.

araspitzu: Okay :), i was using 1BTC in the test to check the "maxing" with the configured value, i changed it now to 5BTC and added the test for the configured value right below it (note that by our formula 5BTC -> 13 confs). Do you want me to squash the commit history of this PR for easier review? It has reached a whopping 10 commits already..

eclair-core/src/main/resources/reference.conf
```diff
@@ -109,6 +109,7 @@ eclair {
   payment-request-expiry = 1 hour // default expiry for payment requests generated by this node
   multi-part-payment-expiry = 60 seconds // default expiry for receiving all parts of a multi-part payment
   min-funding-satoshis = 100000
+  max-funding-satoshis = 16777216 // to open channels larger than 16777216 you must enable the large_channel_support feature in 'eclair.features'
```
```suggestion
  max-funding-satoshis = 16777215 // to open channels larger than 16777216 you must enable the large_channel_support feature in 'eclair.features'
```

Off-by-one error?

araspitzu:
> I've just tested this E2E on regtest and we correctly reject wumbo channels when we haven't activated it ourselves or haven't set the proper funding limits in the configuration.
> 
> However the confirmation scaling doesn't work at all. For a 4 BTC channels, we announce a `min_depth` of 11 blocks (which is correct), but after only 6 blocks both sides send their `funding_locked` (!!!).
> I believe we're not taking into account at all the computed `min_depth`, or it's replaced somewhere with a hard-coded 6 (or the value taken from `nodeParams`).
> 
> Could you investigate this? Ideally a test that reproes this, then a fix?
> 
> It couldn't be that simple :D

Confirmed and now fixed. I forgot to adjust the `min_depth` used in the WatchConfirmed and it was using the configured default, this led to send the `funding_locked` earlier than expected. I've added a test for this in `IntegrationTest` , we now open a 5BTC channel and wait 13 confirmations checking we don't transition to NORMAL before that.

araspitzu:
For reviewers, there is a `FUNDING_TIMEOUT_FUNDEE` hardcoded to 5 days which puts a limit on how much we can wait, if the timeout expires we abort the channel opening. Should we adjust the timeout according to our min_depth calculation or we leave it? 5 days seems to be quite long and we only timeout if the channel size is above 300BTC (and that's a lot).

eclair-core/src/main/resources/reference.conf
araspitzu: Good catch! We wanted the max-funding to be equal to the channel limit which is 2^24 but when validating the parameters we make sure the channel size is LTE than the max-funding. Thus a user with wumbo **disabled** can open up to `16777215`, with wumbo **enabled** up to `16777216` by default because of the max-funding limit.

t-bast:
> Should we adjust the timeout according to our min_depth calculation or we leave it?

No I think 5 days of waiting for a channel to open is enough :D
Especially since both sides would need to state that they are ok with a funding amount > 300 BTC, which is honestly quite crazy.

We just need to double-check that the fundee also verifies that the amount is less than its configured `max-funding-satoshis`.

araspitzu:
> Hum I think we're still using `nodeParams.minDepth` in some `WatchConfirmed` in `Channel.scala`.
> For example lines 485, 1267 and 1457 and 2012.
> Are you sure some of those don't need to use the scaled min-depth?

They don't because when we're funder we don't care about double spends and use the the default `min_depth`. Except line 1457 which was readding the watcher on channel reconnection and i forgot to adjust to the new `min_depth` calculation. I double checked other places in Channel where this could happen and there are none, i also expandeded the integration test to simulate a disconnection to make sure we always use the right min_depth.

t-bast:
> They don't because when we're funder we don't care about double spends and use the the default min_depth

That's true, but what bugs me is that the spec for `funding_locked` says that the sender `MUST wait until the funding transaction has reached minimum_depth before sending this message.`
There is no requirement on the receiver that says it should close/forget the channel if the funder sends a `funding_locked` too early though, so it's probably ok to keep on doing what we're doing.

I think the cleanest way would be to save the `minimum_depth` from `accept_channel` in the channel data and use that. But unfortunately we're not saving it currently, and we should wait for the re-vamp of the channel data encoding (that fixes length and byte alignment) before adding new data there.

I'll do an E2E test and will ACK the latest state of the PR if everything goes well, good job ;)

araspitzu:
Rebased on master

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -308,14 +309,22 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
...
-      val fundingTxFeeratePerKw = c.fundingTxFeeratePerKw_opt.getOrElse(nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(target = nodeParams.onChainFeeConf.feeTargets.fundingBlockTarget))
-      log.info(s"requesting a new channel with fundingSatoshis=${c.fundingSatoshis}, pushMsat=${c.pushMsat} and fundingFeeratePerByte=${c.fundingTxFeeratePerKw_opt} temporaryChannelId=$temporaryChannelId localParams=$localParams")
-      channel ! INPUT_INIT_FUNDER(temporaryChannelId, c.fundingSatoshis, c.pushMsat, channelFeeratePerKw, fundingTxFeeratePerKw, localParams, d.transport, d.remoteInit, c.channelFlags.getOrElse(nodeParams.channelFlags), ChannelVersion.STANDARD)
-      stay using d.copy(channels = d.channels + (TemporaryChannelId(temporaryChannelId) -> channel))
+      if (c.fundingSatoshis >= Channel.MAX_FUNDING && !Features.hasFeature(nodeParams.features, Wumbo)) {
+        sender ! s"fundingSatoshis=${c.fundingSatoshis} is too big, you must enable large channels support in 'eclair.features' to use funding above ${Channel.MAX_FUNDING} (see eclair.conf)"
```
t-bast: I think that you should not be sending a string here, but instead a `Status.Failure(new IllegalArgumentException("your current string"))`.

It may even be better to properly type that error? If you go that way, it will probably need to correctly override toString and may need some JsonSerializer magic to look nice in the CLI?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -308,14 +309,22 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
...
+        stay
+      } else if (c.fundingSatoshis > nodeParams.maxFundingSatoshis) {
+        sender ! s"fundingSatoshis=${c.fundingSatoshis} is too big for the current settings, increase 'eclair.max-funding-satoshis' (see eclair.conf)"
```
t-bast: Same here.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
araspitzu: Sending strings was consistent with the rest of the code in `Peer` but indeed very brittle, commit 457ab1c returns `Status.Failure` with a `RuntimeException` inside. I did not change the JSON serializer/cli because it's already handled by [Throwable serializer](https://github.com/ACINQ/eclair/blob/master/eclair-node/src/main/scala/fr/acinq/eclair/api/JsonSerializers.scala#L192) and the reply from the cli looks like:
```
andrea@andrea-tkp:~/workspace/eclair$ eclair-core/eclair-cli -a localhost:8086 -p foobar open --nodeId=038746cc812250549ba04d5f12d1452aa0c35b48ad37ec17141feb8c407c530409 --fundingSatoshis=50000000
fundingSatoshis=Satoshi(50000000) is too big, you must enable large channels support in 'eclair.features' to use funding above Satoshi(16777216) (see eclair.conf)
```

raw API response:
```json
{
  "error": "fundingSatoshis=Satoshi(400000000) is too big, you must enable large channels support in 'eclair.features' to use funding above Satoshi(16777216) (see eclair.conf)"
}
```

I think we should change other responses too but this should go in a separate PR.

t-bast: Sounds good, we can change that in another PR, it's a broader scope than wumbo.

araspitzu:
> I don't understand why both `push` builds failed, it's surprising

Turns out that under certain circumstances the integration test for the wumbo channel test failed around the reconnection attempt, this happened because one of the two peers might still be _disconnecting_. This caused the peers to move to state OFFLINE after the first iteration of `awaitCond` and never reach the desired state, the fix is to trigger a reconnection at each iteration of the `awaitCond` (done in e4dba1b)

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -260,6 +262,85 @@ class IntegrationSpec extends TestKit(ActorSystem("test")) with BitcoindService
...
+    sender.send(funder.register, Forward(channelId, CMD_GETSTATE))
+    assert(sender.expectMsgType[State] == WAIT_FOR_FUNDING_LOCKED)
+
+    // simulate a disconnection
+    sender.send(funder.switchboard, Peer.Disconnect(fundee.nodeParams.nodeId))
+    assert(sender.expectMsgType[String] == "disconnecting")
+
+    awaitCond({
+      sender.send(fundee.register, Forward(channelId, CMD_GETSTATE))
+      Try(sender.expectMsgType[State]) == Success(OFFLINE)
```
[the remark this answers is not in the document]
t-bast: Wasn't that the real reason of the random failure in the tests?
Because you would start trying to reconnect while fundee may have not been offline yet?
If that's the case it seems you won't need the `awaitCond` below, it's surprising that we need it if peers are both correctly offline when we send the reconnect.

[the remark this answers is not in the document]
araspitzu: Right the result of the first check was thrown away and the `Try` were redundant, i force-pushed a fix. Yes the reason it failed is that we begun the reconnection attempt while the peer wasn't offline yet, note that we check the channel STATE not the peer STATE and the 2 peers have more than one channel between them. So after the first `Peer.Connect` was dismissed with `"already connected"` we didn't try again. The `awaitCond` is handy to repeat the connection attempt, note that there might be one superfluous attempt due to the transition OFFLINE -> SYNCING -> WAIT_FOR_FUNDING_CONFIRMED; the alternative is to wait until the `Peer` is in state OFFLINE but feels a bit cumbersome.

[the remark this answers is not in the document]
araspitzu: FYI [this](https://github.com/araspitzu/eclair/commit/d6c371feb15bd497417802739a0c91e3ad8dd220) is the test with peer disconnection, we still have an `awaitCond` on the peer STATE :/

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -20,7 +20,7 @@ import akka.actor.{ActorRef, FSM, OneForOneStrategy, Props, Status, SupervisorSt
 import akka.event.Logging.MDC
 import akka.pattern.pipe
 import fr.acinq.bitcoin.Crypto.{PrivateKey, PublicKey, sha256}
-import fr.acinq.bitcoin.{ByteVector32, OutPoint, Satoshi, Script, ScriptFlags, Transaction}
+import fr.acinq.bitcoin.{Block, ByteVector32, OutPoint, Satoshi, Script, ScriptFlags, Transaction}
```
pm47: Unused

```suggestion
import fr.acinq.bitcoin.{ByteVector32, OutPoint, Satoshi, Script, ScriptFlags, Transaction}
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -436,8 +436,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           context.system.eventStream.publish(ChannelSignatureReceived(self, commitments))
           // NB: we don't send a ChannelSignatureSent for the first commit
           log.info(s"waiting for them to publish the funding tx for channelId=$channelId fundingTxid=${commitInput.outPoint.txid}")
+          val watchMinDepth = minDepthForFunding(fundingAmount)
```
```suggestion
          val fundingMinDepth = minDepthForFunding(fundingAmount)
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2233,6 +2240,22 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+    * Returns the number of confirmations needed to safely handle the funding transaction,
+    * we make sure the cumulative block reward largely exceeds the channel size.
+    *
+    * @param fundingSatoshis funding amount of the channel
+    * @return number of confirmations needed
+    */
+  def minDepthForFunding(fundingSatoshis: Satoshi): Long = fundingSatoshis match {
+    case funding if funding <= MAX_FUNDING => nodeParams.minDepthBlocks
+    case funding if funding > MAX_FUNDING =>
+      val blockReward = 6.25
```
```suggestion
      val blockReward = 6.25 // this is true as of ~May 2020, and will be too conservative after 2024
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
```diff
@@ -145,6 +157,21 @@ class WaitForAcceptChannelStateSpec extends TestkitBaseClass with StateTestsHelp
...
   }
 
+  test("recv AcceptChannel (wumbo size channel)", Tag("wumbo"), Tag("min-depth"), Tag("max-funding-size"))  { f =>
+    import f._
+    val accept = bob2alice.expectMsgType[AcceptChannel]
+    assert(accept.minimumDepth == 13) // with wumbo tag we use fundingSatoshis=5BTC
+    bob2alice.forward(alice, accept)
+    awaitCond(alice.stateName == WAIT_FOR_FUNDING_INTERNAL)
+
+    assert(alice.underlyingActor.minDepthForFunding(Btc(1)) == 6)  // 4 conf would be enough but we use min-depth=6
```
pm47: This should be moved to `HelpersSpec.scala`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2233,6 +2240,22 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
   }
 
+  /**
+    * Returns the number of confirmations needed to safely handle the funding transaction,
+    * we make sure the cumulative block reward largely exceeds the channel size.
+    *
+    * @param fundingSatoshis funding amount of the channel
+    * @return number of confirmations needed
+    */
+  def minDepthForFunding(fundingSatoshis: Satoshi): Long = fundingSatoshis match {
```
pm47: This static method should be in `Helpers.scala` right before `validateParamsFundee`.

araspitzu:
@pm47 ACKd and implemented all the feedback in f7c97e2 also rebased on master.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
```diff
@@ -147,6 +159,14 @@ class WaitForAcceptChannelStateSpec extends TestkitBaseClass with StateTestsHelp
     awaitCond(alice.stateName == CLOSED)
   }
 
+  test("recv AcceptChannel (wumbo size channel)", Tag("wumbo"), Tag("min-depth"), Tag("max-funding-size"))  { f =>
```
pm47: The `min-depth` tag isn't used anymore, and tags should describe what it does (example: the tag "mainnet" isn't called "chainhash")

```suggestion
  test("recv AcceptChannel (wumbo size channel)", Tag("wumbo"), Tag("high-max-funding-size"))  { f =>
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
araspitzu: This has actually become 
```
// this is true as of ~May 2020, but will be too large after 2024
```
Since the block reward estimate will be too large after 2024 and our min_depth calculations will have to consider this.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
araspitzu: I squashed and rebased the commits to include this nit as well in the fixup.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -312,6 +312,9 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
       if (c.fundingSatoshis >= Channel.MAX_FUNDING && !Features.hasFeature(nodeParams.features, Wumbo)) {
         sender ! Status.Failure(new RuntimeException(s"fundingSatoshis=${c.fundingSatoshis} is too big, you must enable large channels support in 'eclair.features' to use funding above ${Channel.MAX_FUNDING} (see eclair.conf)"))
         stay
+      } else if(c.fundingSatoshis >= Channel.MAX_FUNDING && !Features.hasFeature(d.remoteInit.features, Wumbo)) {
```
pm47: seems to not be formatter

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -312,6 +312,9 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
...
+        sender ! Status.Failure(new RuntimeException(s"fundingSatoshis=${c.fundingSatoshis} is too big, the remote peer doesn't support wumbo."))
```
```suggestion
        sender ! Status.Failure(new RuntimeException(s"fundingSatoshis=${c.fundingSatoshis} is too big, the remote peer doesn't support wumbo"))
```

PR #1324, opened 2020-02-17
Refactor MultiPartPaymentFSM

t-bast:
Add an abstraction of a PaymentPart.
Remove unnecessary intermediary case classes.
This allows extending how payment parts can be received.
It's not limited to HTLCs (could be swaps, pay-to-opens, etc).

@araspitzu you also mentioned a missing abstraction necessary for hodl invoice, can you detail? Maybe this is the right PR to also add something like this.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
@@ -17,7 +17,7 @@
 package fr.acinq.eclair.payment.receive
 
 import akka.actor.Actor.Receive
-import akka.actor.{ActorContext, ActorRef, PoisonPill, Status}
+import akka.actor.{ActorContext, ActorRef, ActorSystem, PoisonPill, Status}
 import akka.event.{DiagnosticLoggingAdapter, LoggingAdapter}
 import fr.acinq.bitcoin.{ByteVector32, Crypto}
 import fr.acinq.eclair.channel.{CMD_FAIL_HTLC, CMD_FULFILL_HTLC, Channel, ChannelCommandResponse}
@@ -48,10 +48,50 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
    */
   def doHandle(paymentHash: ByteVector32): Boolean = true
 
+  /** Can be overridden to do custom pre-processing on successfully received payments. */
+  def preFulfill(success: MultiPartPaymentFSM.MultiPartPaymentSucceeded)(implicit log: LoggingAdapter): MultiPartPaymentFSM.MultiPartPaymentSucceeded = success
+
+  /** Fulfill all parts of an incoming payment. */
+  def doFulfill(system: ActorSystem, preimage: ByteVector32, succeeded: MultiPartPaymentFSM.MultiPartPaymentSucceeded)(implicit log: LoggingAdapter): PaymentReceived = {
+    val received = PaymentReceived(succeeded.paymentHash, succeeded.parts.map {
+      case p: MultiPartPaymentFSM.HtlcPart => PaymentReceived.PartialPayment(p.amount, p.htlc.channelId)
+    })
+    // The first thing we do is store the payment. This allows us to reconcile pending HTLCs after a restart.
+    db.receiveIncomingPayment(succeeded.paymentHash, received.amount, received.timestamp)
+    succeeded.parts.collect {
+      case p: MultiPartPaymentFSM.HtlcPart => commandBuffer ! CommandBuffer.CommandSend(p.htlc.channelId, CMD_FULFILL_HTLC(p.htlc.id, preimage, commit = true))
+    }
+    system.eventStream.publish(received)
+    received
+  }
+
+  /** Can be overridden to do custom post-processing on successfully received payments. */
+  def postFulfill(paymentReceived: PaymentReceived)(implicit log: LoggingAdapter): Unit = ()
+
+  /** Fail all parts of an incoming payment. */
+  def doFail(failed: MultiPartPaymentFSM.MultiPartPaymentFailed): Unit = failed.parts.collect {
+    case p: MultiPartPaymentFSM.HtlcPart => commandBuffer ! CommandBuffer.CommandSend(p.htlc.channelId, CMD_FAIL_HTLC(p.htlc.id, Right(failed.failure), commit = true))
+  }
+
   /**
-   * Can be overridden to do custom processing on successfully received payments.
+   * Handle an extraneous payment received after we reached a final state (succeeded or failed).
+   * This is a spec violation by the sender, but it's a risk-less one: only the sender may lose money because of this.
    */
-  def onSuccess(paymentReceived: PaymentReceived)(implicit log: LoggingAdapter): Unit = ()
+  def doHandleExtraPayment(system: ActorSystem, e: MultiPartPaymentFSM.ExtraPaymentReceived): Unit = e.failure match {
+    case Some(failure) => e.payment match {
+      case p: MultiPartPaymentFSM.HtlcPart => commandBuffer ! CommandBuffer.CommandSend(p.htlc.channelId, CMD_FAIL_HTLC(p.htlc.id, Right(failure), commit = true))
+    }
+    case None => e.payment match {
+      // NB: this case shouldn't happen unless the sender violated the spec, so it's ok that we take a slightly more
+      // expensive code path by fetching the preimage from DB.
+      case p: MultiPartPaymentFSM.HtlcPart => db.getIncomingPayment(e.paymentHash).foreach(record => {
+        commandBuffer ! CommandBuffer.CommandSend(p.htlc.channelId, CMD_FULFILL_HTLC(p.htlc.id, record.paymentPreimage, commit = true))
+        val received = PaymentReceived(e.paymentHash, PaymentReceived.PartialPayment(p.amount, p.htlc.channelId) :: Nil)
+        db.receiveIncomingPayment(e.paymentHash, p.amount, received.timestamp)
+        system.eventStream.publish(received)
+      })
+    }
+  }
 
   override def handle(implicit ctx: ActorContext, log: DiagnosticLoggingAdapter): Receive = {
     case ReceivePayment(amount_opt, desc, expirySeconds_opt, extraHops, fallbackAddress_opt, paymentPreimage_opt, paymentType) =>
@@ -69,7 +109,7 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
           Some(PaymentRequest.Features(f1 ++ f2 ++ f3: _*))
         }
         val paymentRequest = PaymentRequest(nodeParams.chainHash, amount_opt, paymentHash, nodeParams.privateKey, desc, fallbackAddress_opt, expirySeconds = Some(expirySeconds), extraHops = extraHops, features = features)
-        log.debug(s"generated payment request={} from amount={}", PaymentRequest.write(paymentRequest), amount_opt)
+        log.debug("generated payment request={} from amount={}", PaymentRequest.write(paymentRequest), amount_opt)
         db.addIncomingPayment(paymentRequest, paymentPreimage, paymentType)
         paymentRequest
       } match {
@@ -84,13 +124,13 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
             case Some(cmdFail) =>
               commandBuffer ! CommandBuffer.CommandSend(p.add.channelId, cmdFail)
             case None =>
-              log.info(s"received payment for amount=${p.add.amountMsat} totalAmount=${p.payload.totalAmount}")
+              log.info("received payment for amount={} totalAmount={}", p.add.amountMsat, p.payload.totalAmount)
               pendingPayments.get(p.add.paymentHash) match {
                 case Some((_, handler)) =>
-                  handler ! MultiPartPaymentFSM.MultiPartHtlc(p.payload.totalAmount, p.add)
+                  handler ! MultiPartPaymentFSM.HtlcPart(p.payload.totalAmount, p.add)
                 case None =>
                   val handler = ctx.actorOf(MultiPartPaymentFSM.props(nodeParams, p.add.paymentHash, p.payload.totalAmount, ctx.self))
-                  handler ! MultiPartPaymentFSM.MultiPartHtlc(p.payload.totalAmount, p.add)
+                  handler ! MultiPartPaymentFSM.HtlcPart(p.payload.totalAmount, p.add)
                   pendingPayments = pendingPayments + (p.add.paymentHash -> (record.paymentPreimage, handler))
               }
           }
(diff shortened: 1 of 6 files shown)


araspitzu:
@t-bast yes the missing abstraction is [this private local map](https://github.com/ACINQ/eclair/pull/1324/files#diff-5f64796fe03a014868e8897eaac2e9e5R43) that must be updated after each step of the payment handler. An hypothetical plugin developer that wants to inject a behavior is forced to intercept and handle *all* messages to the handler, without doing so the `pendingPayments` map would be left in a wrong state. For a concrete example: doing the hodl-invoice plugin i wanted to intercept only [MultiPartPaymentSucceeded](https://github.com/ACINQ/eclair/pull/1324/files#diff-5f64796fe03a014868e8897eaac2e9e5R111) but from the subclass i can't access `pendingPayments` to get the preimage and later remove the entry for the payment. A possible solution for this is to bring up `pendingPayments` to the handler interface `ReceiveHandler` and expose getter/setter(s) to interact with it, subclasses can override these methods and/or use them to update the map (see [commit](https://github.com/araspitzu/eclair/commit/0d83f78a1e3e49e6fd580013409872034f524e07)).

t-bast:
I really don't like exposing the `pendingPayments` map. This is private state, if modified in unexpected ways it may break many invariants. It should really stay hidden inside the `MultiPartHandler`.

It seems to me that the abstraction you're missing is the fulfillment (where you would want to avoid fulfilling HTLCs instantly and instead wait for some user input). What about simply adding a `doFulfill` method on `MultiPartHandler`? This way your plugin's payment handler could simply override that to avoid fulfilling right away.

pm47:
> This way your plugin's payment handler could simply override that to avoid fulfilling right away.

A somewhat hackish way of doing that is to override `ChargesDb.receiveIncomingPayment`. That's what we do for Strike (the real one).

araspitzu:
Yeah exposing `pendingPayments` is hackish and a bit error prone. OTOH exposing a `doFulfill` seems just enough for my use case but not a generic solution, if a plugin developer wants to intercept the behavior of failed payments we'd need to add a `doFail` as well.  More in general I think we can have a more plugin friendly handler architecture if we expose `do<Operation>` stile functions but they change depending on the handler (thinking about `PaymentHandler` vs `PaymentInitiator`).

t-bast:
I believe offering hook for `doFulfill` and `doFail` is enough for the `MultiPartHandler`.
Plugins that want more capabilities can always register a completely different `PaymentHandler` than `MultiPartHandler`, right?

I'll start with this `doFail` and `doFulfill` abstractions and we'll improve incrementally from there.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -100,41 +125,41 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
         }
       }
 
-    case MultiPartPaymentFSM.MultiPartHtlcFailed(paymentHash, failure, parts) if doHandle(paymentHash) =>
+    case e@MultiPartPaymentFSM.MultiPartPaymentFailed(paymentHash, failure, parts) if doHandle(paymentHash) =>
       Logs.withMdc(log)(Logs.mdc(paymentHash_opt = Some(paymentHash))) {
-        log.warning(s"payment with paidAmount=${parts.map(_.payment.amount).sum} failed ($failure)")
+        log.warning(s"payment with paidAmount=${parts.map(_.amount).sum} failed ($failure)")
```
pm47: As a best practice we should use less string interpolation for logs and more `"{}", myvar`

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -100,41 +125,41 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
...
-          case Some(failure) => commandBuffer ! CommandBuffer.CommandSend(p.payment.fromChannelId, CMD_FAIL_HTLC(p.htlcId, Right(failure), commit = true))
-          // NB: this case shouldn't happen unless the sender violated the spec, so it's ok that we take a slightly more
-          // expensive code path by fetching the preimage from DB.
-          case None => db.getIncomingPayment(paymentHash).foreach(record => {
-            commandBuffer ! CommandBuffer.CommandSend(p.payment.fromChannelId, CMD_FULFILL_HTLC(p.htlcId, record.paymentPreimage, commit = true))
-            db.receiveIncomingPayment(paymentHash, p.payment.amount, p.payment.timestamp)
-            ctx.system.eventStream.publish(PaymentReceived(paymentHash, p.payment :: Nil))
-          })
+          case Some(failure) => p match {
+            case p: MultiPartPaymentFSM.HtlcPart => commandBuffer ! CommandBuffer.CommandSend(p.htlc.channelId, CMD_FAIL_HTLC(p.htlc.id, Right(failure), commit = true))
```
pm47: I'm not sure the `doFulfill`/`doFail` methods are useful if we can't factor that in.

t-bast: I'm not sure, this code has to be there to handle unexpected cases, but that is a spec violation.
I believe most use-cases can simply ignore that technicality.
But we probably need real use-cases to figure that case-by-case.

araspitzu:
The `doFulfill/doFail` are enough for the plugin use case, these are the most important operations going on in the payment handler and if a plugin developer wants to have more control it can always register a completely new handler. I don't think we should expose procotol level details to the plugin developer.

t-bast:
Actually it doesn't hurt to extract the extra HTLC handling to a dedicated function. This way plugins can choose to either handle it themselves or let the default behavior. I'll do that in the next commit. Good to know that this will unblock a few useful plugins!

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: I think it's especially useful for `log.debug` and `log.info` that may be turned off, for `warn` and `error` which you should always keep this isn't an issue. I'll update this file though.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartPaymentFSM.scala
```diff
@@ -48,14 +47,13 @@ class MultiPartPaymentFSM(nodeParams: NodeParams, paymentHash: ByteVector32, tot
...
-      val updatedParts = d.parts :+ pp
-      if (totalAmount != totalAmount2) {
-        log.warning(s"multi-part payment total amount mismatch: previously $totalAmount, now $totalAmount2")
-        goto(PAYMENT_FAILED) using PaymentFailed(IncorrectOrUnknownPaymentDetails(totalAmount2, nodeParams.currentBlockHeight), updatedParts)
-      } else if (d.paidAmount + htlc.amountMsat >= totalAmount) {
+    case Event(part: PaymentPart, d: WaitingForHtlc) =>
+      require(part.paymentHash == paymentHash, s"invalid payment hash (expected $paymentHash, received ${part.paymentHash}")
+      val updatedParts = d.parts :+ part
+      if (totalAmount != part.totalAmount) {
+        log.warning(s"multi-part payment total amount mismatch: previously $totalAmount, now ${part.totalAmount}")
```
araspitzu: nit: this can be migrated to not use the string interpolation

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartPaymentFSM.scala
```diff
@@ -66,20 +64,20 @@ class MultiPartPaymentFSM(nodeParams: NodeParams, paymentHash: ByteVector32, tot
     // A sender should not send us additional htlcs for that payment once we've already reached the total amount.
     // However if that happens the only rational choice is to fulfill it, because the pre-image has been released so
     // intermediate nodes will be able to fulfill that htlc anyway. This is a harmless spec violation.
-    case Event(MultiPartHtlc(_, htlc), _) =>
-      require(htlc.paymentHash == paymentHash, s"invalid payment hash (expected $paymentHash, received ${htlc.paymentHash}")
-      log.info(s"received extraneous htlc amountMsat=${htlc.amountMsat}")
-      parent ! ExtraHtlcReceived(paymentHash, PendingPayment(htlc.id, PartialPayment(htlc.amountMsat, htlc.channelId)), None)
+    case Event(part: PaymentPart, _) =>
+      require(part.paymentHash == paymentHash, s"invalid payment hash (expected $paymentHash, received ${part.paymentHash}")
```
araspitzu: nit: same as above, this can be migrated to not use string interpolation

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartPaymentFSM.scala
```diff
@@ -66,20 +64,20 @@ class MultiPartPaymentFSM(nodeParams: NodeParams, paymentHash: ByteVector32, tot
...
+      log.info(s"received extraneous payment part with amount=${part.amount}")
```
araspitzu: same

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartPaymentFSM.scala
```diff
@@ -66,20 +64,20 @@ class MultiPartPaymentFSM(nodeParams: NodeParams, paymentHash: ByteVector32, tot
...
   when(PAYMENT_FAILED) {
     // If we receive htlcs after the multi-part payment has expired, we must fail them.
     // The LocalPaymentHandler will create a new instance of MultiPartPaymentHandler to handle a new attempt.
-    case Event(MultiPartHtlc(_, htlc), PaymentFailed(failure, _)) =>
-      require(htlc.paymentHash == paymentHash, s"invalid payment hash (expected $paymentHash, received ${htlc.paymentHash}")
-      log.info(s"received extraneous htlc for payment hash $paymentHash")
-      parent ! ExtraHtlcReceived(paymentHash, PendingPayment(htlc.id, PartialPayment(htlc.amountMsat, htlc.channelId)), Some(failure))
+    case Event(part: PaymentPart, PaymentFailed(failure, _)) =>
+      require(part.paymentHash == paymentHash, s"invalid payment hash (expected $paymentHash, received ${part.paymentHash}")
+      log.info(s"received extraneous payment part for payment hash $paymentHash")
```
araspitzu: nit: string interpolation

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -67,28 +67,28 @@ class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, c
...
 
     // We always fail extraneous HTLCs. They are a spec violation from the sender, but harmless in the relay case.
     // By failing them fast (before the payment has reached the final recipient) there's a good chance the sender won't lose any money.
-    case MultiPartPaymentFSM.ExtraHtlcReceived(_, p, failure) => rejectHtlc(p.htlcId, p.payment.fromChannelId, p.payment.amount, failure)
+    case MultiPartPaymentFSM.ExtraPaymentReceived(_, p: MultiPartPaymentFSM.HtlcPart, failure) => rejectHtlc(p.htlc.id, p.htlc.channelId, p.amount, failure)
 
-    case MultiPartPaymentFSM.MultiPartHtlcFailed(paymentHash, failure, parts) =>
-      log.warning(s"could not relay payment (paidAmount=${parts.map(_.payment.amount).sum} failure=$failure)")
+    case MultiPartPaymentFSM.MultiPartPaymentFailed(paymentHash, failure, parts) =>
+      log.warning(s"could not relay payment (paidAmount=${parts.map(_.amount).sum} failure=$failure)")
```
araspitzu: nit: string interpolation

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -48,6 +48,52 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
    */
   def doHandle(paymentHash: ByteVector32): Boolean = true
 
+  /**
+   * Fulfill all parts of an incoming payment.
+   * Can be overridden for a more fine-grained control of when and how to fulfill.
+   */
+  def doFulfill(ctx: ActorContext, preimage: ByteVector32, e: MultiPartPaymentFSM.MultiPartPaymentSucceeded)(implicit log: LoggingAdapter): Unit = {
```
araspitzu: The current changes are enough for the hodl plugin invoices and likely for most use cases of the incoming payments. I highlighted here a change that could simplify even more the life of the plugin developers:
```suggestion
  def doFulfill(system: ActorSystem, preimage: ByteVector32, e: MultiPartPaymentFSM.MultiPartPaymentSucceeded)(implicit log: LoggingAdapter): Unit = {
```
This change lets the plugin developer avoid to reimplement `doFulfill` entirely, instead it can use `super.doFulfill` when it wants to fulfill the HTLC. That's necessary because a subclass of `MultiPartHandler` is not an actor and doesn't have an actor context but it does have the actor system which is enough to access the event stream. I'm not strongly in favor of this change but is be a nice-to-have.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartPaymentFSM.scala
t-bast: I can't do that in `require`, only in `log`. And it only works when there's at most 4 arguments so there are two string interpolations that I left.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: Done, it was an easy low hanging fruit.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -48,6 +48,52 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
    */
   def doHandle(paymentHash: ByteVector32): Boolean = true
 
+  /**
+   * Fulfill all parts of an incoming payment.
+   * Can be overridden for a more fine-grained control of when and how to fulfill.
+   */
+  def doFulfill(system: ActorSystem, preimage: ByteVector32, e: MultiPartPaymentFSM.MultiPartPaymentSucceeded)(implicit log: LoggingAdapter): Unit = {
```
pm47: I'm not comfortable with letting plugin writers override that kind of tricky method as a recommended practice.

How about we let those methods as-is, but don't recommend overriding them, and go akka-style by adding a  `preFulfill` and `postFulfill` methods (the latter replacing `onSuccess`)?

t-bast: I'll try that.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1324/commits/7f4545660b41d286a12c2a383f8b9fb23947801d

t-bast:
Addressed all comments, rebased and squashed.
Please do a last pass and we should be good to go!

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -100,42 +140,28 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
...
-        val received = PaymentReceived(paymentHash, parts.map(_.payment))
-        log.info(s"received complete payment for amount=${received.amount}")
-        // The first thing we do is store the payment. This allows us to reconcile pending HTLCs after a restart.
-        db.receiveIncomingPayment(paymentHash, received.amount, received.timestamp)
+        log.info("received complete payment for amount={}", parts.map(_.amount).sum)
         pendingPayments.get(paymentHash).foreach {
           case (preimage: ByteVector32, handler: ActorRef) =>
             handler ! PoisonPill
-            parts.foreach(p => commandBuffer ! CommandBuffer.CommandSend(p.payment.fromChannelId, CMD_FULFILL_HTLC(p.htlcId, preimage, commit = true)))
+            postFulfill(doFulfill(ctx.system, preimage, preFulfill(s)))
```
pm47: nice one liner 🤓

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -48,10 +48,50 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
    */
   def doHandle(paymentHash: ByteVector32): Boolean = true
 
+  /** Can be overridden to do custom pre-processing on successfully received payments. */
+  def preFulfill(success: MultiPartPaymentFSM.MultiPartPaymentSucceeded)(implicit log: LoggingAdapter): MultiPartPaymentFSM.MultiPartPaymentSucceeded = success
```
pm47: What's the point of making this method return the `MultiPartPaymentFSM.MultiPartPaymentSucceeded`? Is it that we want the user to be able to alter this object and what would be an example of that?

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: Pipe-lining FTW :)

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: Yes I figured that if this method doesn't return anything, then it cannot influence the `doFulfill` in any way so it becomes completely useless.

A plugin could modify the `PaymentPart`s to enrich them with more data from an external DB (not sure yet if this is useful). I was wondering if we should take this further and provide the `preFulfill` with the ability to delay the `doFulfill` (if for example `preFulfill` asks the user for some confirmation and wants to wait for its input without blocking the actor).
Could we return a Future to make this possible? Or another synchronization object?

t-bast: I made `DoFulfill` a message instead of a function in https://github.com/ACINQ/eclair/pull/1324/commits/d18a5bb2072c3419eb0fe8ada8f513c18b073093.
This allows a plugin to override `preFulfill` to arbitrarily delay or change what needs to be fulfilled, or even fetch a preimage from an external source.
I think we can start with that, and then evolve the abstraction as we build plugins and see what's missing.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -49,9 +49,40 @@ class MultiPartHandler(nodeParams: NodeParams, db: IncomingPaymentsDb, commandBu
   def doHandle(paymentHash: ByteVector32): Boolean = true
 
   /**
-   * Can be overridden to do custom processing on successfully received payments.
+   * Can be overridden to do custom pre-processing on successfully received payments.
+   * Must send a [[DoFulfill]] message for the payment to actually be fulfilled upstream.
    */
-  def onSuccess(paymentReceived: PaymentReceived)(implicit log: LoggingAdapter): Unit = ()
+  def preFulfill(ctx: ActorContext, preimage: ByteVector32, success: MultiPartPaymentFSM.MultiPartPaymentSucceeded)(implicit log: LoggingAdapter): Unit = {
```
pm47: More consistent with existing code (line 148 needs to be updated too):

```suggestion
  def preFulfill(preimage: ByteVector32, success: MultiPartPaymentFSM.MultiPartPaymentSucceeded)(implicit ctx: ActorContext, log: LoggingAdapter): Unit = {
```

pm47: Alternative proposition that makes it easy to implement the hold-invoice use case (cc @araspitzu) and doesn't rely on the user to send the `DoFulfill`.

```scala
  /**
   * Can be overridden to do custom pre-processing on successfully received payments.
   * The fulfill won't be attempted until the future completes successfully.
   */
  def preFulfill(success: MultiPartPaymentFSM.MultiPartPaymentSucceeded)(implicit ctx: ActorContext, log: LoggingAdapter): Future[Unit] = Future.successful()

...

    case s@MultiPartPaymentFSM.MultiPartPaymentSucceeded(paymentHash, parts) if doHandle(paymentHash) =>
      Logs.withMdc(log)(Logs.mdc(paymentHash_opt = Some(paymentHash))) {
        log.info("received complete payment for amount={}", parts.map(_.amount).sum)
        pendingPayments.get(paymentHash).foreach {
          case (preimage: ByteVector32, handler: ActorRef) =>
            handler ! PoisonPill
            preFulfill(s).map(_ => ctx.self ! DoFulfill(preimage, s))(ctx.dispatcher)
        }
        pendingPayments = pendingPayments - paymentHash
      }
```

t-bast: Great, the future feels much better!

PR #1325, opened 2020-02-17
SendToRoute with routing hints

t-bast:
SendToRoute previously didn't accept invoice routing hints.
This was a limitation when paying a wallet invoice.
Invoice hints are now correctly taken into account.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -148,7 +148,7 @@ case class RouteRequest(source: PublicKey,
                         ignoreChannels: Set[ChannelDesc] = Set.empty,
                         routeParams: Option[RouteParams] = None)
 
-case class FinalizeRoute(hops: Seq[PublicKey])
+case class FinalizeRoute(hops: Seq[PublicKey], assistedRoutes: Seq[Seq[ExtraHop]] = Nil)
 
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -533,9 +533,12 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
       sender ! d
       stay
 
-    case Event(FinalizeRoute(partialHops), d) =>
+    case Event(FinalizeRoute(partialHops, assistedRoutes), d) =>
+      val assistedChannels: Map[ShortChannelId, AssistedChannel] = assistedRoutes.flatMap(toAssistedChannels(_, partialHops.last, 0 msat)).toMap
```
araspitzu: This is more a remark than a request for change: using `0 msat` here causes the fake channel update to have `htlc_maximum_msat` equal to `RoutingHeuristics.CAPACITY_CHANNEL_LOW`. This is not a problem but it can create routes that are technically unusable because the last hop max htlc is smaller than the amount we want to send, but when using send to route we don't really check that anymore. Maybe a small explanatory comment here would be good.

t-bast: True that it's a bit subtle, I'll add a comment :)

PR #1326, opened 2020-02-18
Allow receiving non-standard payments

t-bast:


eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -169,7 +169,8 @@ object MultiPartHandler {
                             expirySeconds_opt: Option[Long] = None,
                             extraHops: List[List[ExtraHop]] = Nil,
                             fallbackAddress: Option[String] = None,
-                            paymentPreimage: Option[ByteVector32] = None)
+                            paymentPreimage: Option[ByteVector32] = None,
+                            paymentType: String = PaymentType.Standard)
```
pm47: How about we remove the default value? Default value creates bugs, this is an example of it 😈

PR #1327, opened 2020-02-20
Notify shutdown

dscotese:
travis.yml updated to use a working URL for Maven. The call to exit() in FxApp is now preceded by a log entry.


dscotese:
It is the github website that added newlines at the ends of the files.  The page you linked to doesn't describe how to avoid that, but I found an issue with "Atom" which is a text editor provided by Github and which has a project called "github :octocat: Git and GitHub integration for Atom" (see https://github.com/atom/atom/issues/4741).  This suggests to me that the Contributions Doc should indicate that files cannot be edited using the github website because it will always add a newline to the end of each file when it is saved (See https://github.com/ACINQ/eclair/commit/a4282cc43c792d8e4b56091bbb98a02219a7e848 for an example showing that my attempt to remove it was unsuccessful).  I am now trying to figure out how to undo the edit on my laptop and then update from github and then edit out the two extra newlines and add that to the PR.  I'm stuck on reverting the change on my laptop.

dscotese:
I couldn't figure out now to update this pull request with the corrected files, so I'm closing this in favor of #1329

PR #1331, opened 2020-02-24
Allow funder to dip into reserve if increased fee

t-bast:
See https://github.com/lightningnetwork/lightning-rfc/issues/728 and https://github.com/lightningnetwork/lightning-rfc/pull/750

Allow a funder to dip into its channel reserve to pay the increased commit tx fee for one incoming HTLC.
This prevents being stuck with an unusable channel.

This is an alternative to #1319. I am waiting for feedback from the other implementations to see which of the two solutions we choose to merge.


t-bast:
We agreed during the spec meeting to use the additional reserve instead of dipping into the existing reserve, so I'm closing this PR in favor of #1319.

PR #1332, opened 2020-02-26
Load plugins from config file

araspitzu:
~~This PR changes the way plugins are passed to the eclair application, instead of specifying the plugin filename as a command line parameter we now require all the plugins to be placed in a `plugin/` folder inside the eclair `<datadir>`.  I've updated the README but i think we might need a separate file for plugins instruction.~~

This PR improves plugin loading by allowing the user to specify the full path for a plugin file via configuration, this is handy for users who cannot/don't want to deal with command line options. The new configuration key is an array of strings containing the full path for the plugin file(s).

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,10 +195,16 @@ For advanced usage, Eclair supports plugins written in Scala, Java, or any JVM-c
 A valid plugin is a jar that contains an implementation of the [Plugin](eclair-node/src/main/scala/fr/acinq/eclair/Plugin.scala) interface, and 
 a manifest entry for `Main-Class` with the FQDN of the implementation.
 
-Here is how to run Eclair with plugins:
+To run eclair with plugins you can either specify the plugin files via command line or configuration.
 
+Using command line:
 ```shell
 eclair-node-<version>-<commit_id>/bin/eclair-node.sh <plugin1.jar> <plugin2.jar> <...>
+```
+
+Using configuration:
+```
+eclair.plugin-paths = [ "/my/plugin/full/path.jar", "/my/plugin/full/path1.jar" ]
 ```
 
 ## Testnet usage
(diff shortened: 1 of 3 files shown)


pm47:
Doesn't it feels a bit strange to put _code_ in a _data_ directory?

araspitzu:
> Doesn't it feels a bit strange to put _code_ in a _data_ directory?
> 
> edit: Actually it definitely feels wrong

It seems to me better suited than the alternative which is the unzipped eclair folder (where we have `/bin` and `/lib`). If we put the plugins in the unzipped folder at each deploy the user would overwrite it and there is some copy/pasting to be done to preserve the plugin files. OTOH the _datadir_ is always available and is preserved across deploys. I choose the `datadir` mostly for being a lot more practical than the `/lib` or `/bin` folder.

sstone:
I see the point about not deploying code in the data directory but I really don't like having users add things to the eclair app folder. 
What about adding a new `plugins` option to `eclair.conf` that specifies a list of .jar files( and could also be defined with a `-D` java environment variable) ?

araspitzu:
The main point was to make it easier for the users to activate plugins and having a `/plugin` folder would allow them to just "drop" the plugin files without further ado. I do reckon there are valid arguments against using both the `datadir` and the eclair app dir, and i think if we can't settle on the usage of a new directory we should settle with @sstone 's suggestions which is strictly better than what we have now (recall that a windows GUI user has to manually alter its `eclair-node-gui.bat` to use plugins). 

Here some examples of other projects that preferred to store plugins under their datadir instead of the app directory:
- [Intellij idea](https://www.jetbrains.com/help/idea/2019.3/tuning-the-ide.html?_ga=2.253281417.1221530418.1582887193-570356072.1581608926#plugins-directory)
- [c-lightning](https://github.com/lightningd/plugins#automatic-plugin-initialization)

araspitzu:
After some offline discussion it has been decided to go with the middle ground of loading the plugin files via configuration, commit ff83bfe implements the new changes and it's rebased on the latest master.

t-bast:
Closing this stale PR.

PR #1333, opened 2020-02-26
Make UpfrontShutdownScript a TLV record

t-bast:
We make some previously optional fields mandatory in `ChannelReestablish`, `OpenChannel` and `AcceptChannel`. This allows us to extend all messages with an optional TLV stream.

See https://github.com/lightningnetwork/lightning-rfc/pull/714 for more details.

We're moving `upfront_shutdown_script` to a TLV field: this is compatible at the byte level and provides more flexibility in the future (to either include or exlude it). For now we always include an empty one, for backwards-compatibility.

This allowed removing the ugly hack in the codecs that was made specifically for Phoenix. In the branch we deploy to the node connected to Phoenix wallets, we should simply remove the `UpfrontShutdownScript(ByteVector.empy)` from the TLV stream and we will be backwards-compatible with older Phoenix.

E2E tests:

- [x] eclair-v0.3.3
- [x] lnd v0.9.1-beta.rc1
- [x] c-lightning v0.8.1


eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelTlv.scala
```diff
@@ -0,0 +1,59 @@
...
+object OpenChannelTlv {
+
+  import ChannelTlv._
+
+  case class Placeholder(b: ByteVector) extends OpenChannelTlv
+
+  val openTlvCodec: Codec[TlvStream[OpenChannelTlv]] = tlvStream(discriminated[OpenChannelTlv].by(varint)
+    .typecase(UInt64(0), variableSizeBytesLong(varintoverflow, bytes).as[UpfrontShutdownScript])
+    .typecase(UInt64(65717), variableSizeBytesLong(varintoverflow, bytes).as[Placeholder])
+  )
```
pm47:
```suggestion
  )
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -56,33 +56,10 @@ object LightningMessageCodecs {
...
-    // stream.
-    (b: BitVector) => Attempt.successful(DecodeResult(b.startsWith(hex"00".bits), b))
-  )
-
-  private def emptyToNone(script: Option[ByteVector]): Option[ByteVector] = script match {
-    case Some(s) if s.nonEmpty => script
-    case _ => None
-  }
+      ("yourLastPerCommitmentSecret" | privateKey) ::
+      ("myCurrentPerCommitmentPoint" | publicKey)).as[ChannelReestablish]
```
pm47: Phew! nice to get rid of that

PR #1334, opened 2020-02-27
Allow users to have multiple wallets in bitcoind

### LINKED ISSUE #1330: fatal error: bitcoind must have wallet support enabled when Bitcoin Core have multiple wallets
What behaviour did you expect?
Eclair should start up.

What was the actual behaviour (provide screenshots if the issue is GUI-related)?
When starting up eclair for the first time, eclair fails on startup with the following error message : fatal error: bitcoind must have wallet support enabled.

How reliably can you reproduce the issue, what are the steps to do so?
Just run `java -jar` with the .jar eclair file

What version of Eclair are you using, where did you get it (website, self-compiled, etc)?
Downloaded 3.3 version from here

What environment are you observing the error on (OS, JDK version)?
Linux Celeri 4.15.0-48-generic #51-Ubuntu SMP Wed Apr 3 08:28:49 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

Any extra information that might be useful in the debugging process.
The cause is that I have multiple wallets on Bitcoin Core, so every call needs `--rpcwallet=` option, which apparently eclair is not doing when it calls `getbalance`
I guess the best way to solve the problem would be to reorganise my wallets and only have the default one, but I wonder if there's a quick fix that would allow me to keep my wallets the way it is now.
Anyway I think this is a problem that should be taken care of at some point, at least by documenting the fact that it's not possible to run eclair with a multi wallets Bitcoin Core, or by modifying the way eclair makes its bitcoin-cli calls accordingly (maybe systematically add `--rpcwallet=` could work?).

eclair.log : 
2020-02-21 23:06:04,238 INFO  fr.acinq.eclair.Setup - using chain=mainnet chainHash=6fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d6190000000000
2020-02-21 23:06:06,056 ERROR fr.acinq.eclair.Boot$ - fatal error: bitcoind must have wallet support enabledfr.acinq.eclair.BitcoinWalletDisabledException$: bitcoind must have wallet support enabled
        at fr.acinq.eclair.BitcoinWalletDisabledException$.<clinit>(Setup.scala)
        at fr.acinq.eclair.Setup$$anonfun$10$$anonfun$apply$1.applyOrElse(Setup.scala:150)
        at fr.acinq.eclair.Setup$$anonfun$10$$anonfun$apply$1.applyOrElse(Setup.scala:150)
        at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
        at scala.util.Failure$$anonfun$recover$1.apply(Try.scala:216)
        at scala.util.Try$.apply(Try.scala:192)
        at scala.util.Failure.recover(Try.scala:216)
        at scala.concurrent.Future$$anonfun$recover$1.apply(Future.scala:326)
        at scala.concurrent.Future$$anonfun$recover$1.apply(Future.scala:326)
        at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:36)
        at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
        at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
        at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.pollAndExecAll(ForkJoinPool.java:1253)
        at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1346)
        at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
        at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)

araspitzu:
In this PR we force the usage of the default bitcoind wallet (named ""). This is necessary to allow the user to have multiple bitcoind wallets because when there are more than one we need to specify which wallet to use in the RPC calls. Fixes #1330.

Note for reviewers: the extra path in the bitcoin RPC URL identifies the wallet to be used and can be used in all the others RPC too. By default bitcoind has an unnamed wallet `""` and eclair always expect to have that, users can add new wallets but should never rename or remove the default wallet (maybe we can add a few lines in the README for this).

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ You will find detailed guides and frequently asked questions there.
 :warning: Eclair requires Bitcoin Core 0.17.1 or higher. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
 
 Eclair needs a _synchronized_, _segwit-ready_, **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node.
-Eclair will use any BTC it finds in the Bitcoin Core wallet to fund any channels you choose to open. Eclair will return BTC from closed channels to this wallet. 
+Eclair will use the default unnamed wallet in Bitcoin Core, you can have more than one wallet but make sure there is always the default wallet available.
+Any BTC found in the wallet can be used to fund the channels you choose to open and the BTC from closed channels will return to this wallet. 
 You can configure your Bitcoin Node to use either `p2sh-segwit` addresses or `bech32` addresses, Eclair is compatible with both modes.
 If your Bitcoin Core wallet has "non-segwit UTXOs" (outputs that are neither `p2sh-segwit` or `bech32`), you must send them to a `p2sh-segwit` or `bech32` address.
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ You will find detailed guides and frequently asked questions there.
 :warning: Eclair requires Bitcoin Core 0.17.1 or higher. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
 
 Eclair needs a _synchronized_, _segwit-ready_, **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node.
-Eclair will use the default unnamed wallet in Bitcoin Core, you can have more than one wallet but make sure there is always the default wallet available.
+Eclair will use any BTC it finds in the default Bitcoin Core wallet to fund any channels you choose to open. Eclair will return BTC from closed channels to this wallet. You can have multiple Bitcoin Core wallets but make sure that the default one is always available.
 Any BTC found in the wallet can be used to fund the channels you choose to open and the BTC from closed channels will return to this wallet. 
 You can configure your Bitcoin Node to use either `p2sh-segwit` addresses or `bech32` addresses, Eclair is compatible with both modes.
 If your Bitcoin Core wallet has "non-segwit UTXOs" (outputs that are neither `p2sh-segwit` or `bech32`), you must send them to a `p2sh-segwit` or `bech32` address.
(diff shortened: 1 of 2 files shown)


sstone:
I was able to reproduce #1330 on testnet and check that this PR fixes it, but as mentioned it's worth adding a few lines to our documentation.

t-bast:
Open question: if we do that, why don't we allow the user to specify which wallet to use?
They could create a wallet for lightning, provide its name in `eclair.conf` and we'd use that.
And by default if they don't provide anything, we'd use the unnamed one.

araspitzu:
> Open question: if we do that, why don't we allow the user to specify which wallet to use?
> They could create a wallet for lightning, provide its name in `eclair.conf` and we'd use that.
> And by default if they don't provide anything, we'd use the unnamed one.

Why not :) as long as we can provide additional features without damaging the 0-conf setup i think we can add it.

araspitzu:
Note for reviewers: at commit 44e009e this PR reads the bitcoin core wallet name from the configuration. Under some circumstances this introduces a new way the users can hurt themselves: between the [`fundTransaction`](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala#L40) and [`rollback`](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala#L118) operation we lock and unlock outpoints, if a user inadvertently restarts eclair before the rollback with a different `walletname` than it won't work and the outputs won't be unlocked. I think it's an acceptable risk (perhaps we can add a mention in the readme) because in general the users should not change bitcoind/wallet while there are pending operations in eclair.

araspitzu:
On a second thought there are some other cases where the users can hurt themselves and even lose funds (think of swapping wallet and shutdown scripts usage), as a precaution i  reverted the changes to implement only the original scope of this PR.

README.md
```diff
@@ -48,7 +48,8 @@ You will find detailed guides and frequently asked questions there.
 :warning: Eclair requires Bitcoin Core 0.17.1 or higher. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
 
 Eclair needs a _synchronized_, _segwit-ready_, **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node.
-Eclair will use any BTC it finds in the Bitcoin Core wallet to fund any channels you choose to open. Eclair will return BTC from closed channels to this wallet. 
+Eclair will use the default unnamed wallet in Bitcoin Core, you can have more than one wallet but make sure there is always the default wallet available.
```
sstone:
```suggestion
Eclair will use any BTC it finds in the default Bitcoin Core wallet to fund any channels you choose to open. Eclair will return BTC from closed channels to this wallet. You can have multiple Bitcoin Core wallets but make sure that the default one is always available.
```

PR #1339, opened 2020-03-04
Harden requirements on htlc-minimum-msat

t-bast:
We were allowing users to set htlc-minimum-msat to 0, which directly
contradicts the fact that we must never send an HTLC for 0 msat.
We now explicitly disallow that behavior: the minimum is 1 msat.

In case the remote side of a channel had set its htlc-minimum-msat to 0,
we would forward HTLC with a value of 0 msat if a sender crafted such a
payment. The spec disallows that, so we now explicitly check for that
lower bound.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -162,7 +162,7 @@ object NodeParams {
     }
 
     val htlcMinimum = MilliSatoshi(config.getInt("htlc-minimum-msat"))
-    require(htlcMinimum > 0.msat, "htlc-minimum-msat must not be 0")
+    require(htlcMinimum > 0.msat, "htlc-minimum-msat must be strictly greater than 0")
 
     val maxAcceptedHtlcs = config.getInt("max-accepted-htlcs")
     require(maxAcceptedHtlcs <= Channel.MAX_ACCEPTED_HTLCS, s"max-accepted-htlcs must be lower than ${Channel.MAX_ACCEPTED_HTLCS}")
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -161,6 +161,9 @@ object NodeParams {
       require(dustLimitSatoshis >= Channel.MIN_DUSTLIMIT, s"dust limit must be greater than ${Channel.MIN_DUSTLIMIT}")
     }
 
+    val htlcMinimum = MilliSatoshi(config.getInt("htlc-minimum-msat"))
+    require(htlcMinimum > 0.msat, "htlc-minimum-msat must not be 0")
```
pm47: nit

```suggestion
    require(htlcMinimum > 0.msat, "htlc-minimum-msat must be greater than 0")
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -169,8 +169,8 @@ object Commitments {
       return Failure(ExpiryTooBig(commitments.channelId, maximum = maxExpiry, actual = cmd.cltvExpiry, blockCount = blockHeight))
     }
 
-    if (cmd.amount < commitments.remoteParams.htlcMinimum) {
-      return Failure(HtlcValueTooSmall(commitments.channelId, minimum = commitments.remoteParams.htlcMinimum, actual = cmd.amount))
+    if (cmd.amount < commitments.remoteParams.htlcMinimum.max(1 msat)) {
```
pm47: Why do we need this if we have a hard requirement in the settings?

t-bast: Because we can't ensure that our remote has that requirement as well ;)
It seems that c-lightning's default value here is 0, but we still don't want to forward them 0-value HTLCs.

PR #1340, opened 2020-03-04
Send ping if no message received from peer for 30s

pm47:
As long as we receive messages from our peer, we consider it is
online and don't send ping requests. If we don't hear from the
peer, we send pings and expect timely answers, otherwise we'll
close the connection.

This is implemented by scheduling a ping request every 30 seconds,
and pushing it back every time we receive a message from the peer.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -192,7 +192,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
         // we will delay all rebroadcasts with this value in order to prevent herd effects (each peer has a different delay)
         val rebroadcastDelay = Random.nextInt(nodeParams.routerConf.routerBroadcastInterval.toSeconds.toInt).seconds
         log.info(s"rebroadcast will be delayed by $rebroadcastDelay")
-        goto(CONNECTED) using ConnectedData(d.address_opt, d.transport, d.localInit, remoteInit, d.channels.map { case (k: ChannelId, v) => (k, v) }, rebroadcastDelay) forMax (30 seconds) // forMax will trigger a StateTimeout
+        scheduleNextPing()
+        goto(CONNECTED) using ConnectedData(d.address_opt, d.transport, d.localInit, remoteInit, d.channels.map { case (k: ChannelId, v) => (k, v) }, rebroadcastDelay)
       }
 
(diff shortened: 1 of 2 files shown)


pm47:
NB: I highly recommend selecting "Hide whitespace changes" for the review

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -389,6 +389,21 @@ class PeerSpec extends TestkitBaseClass with StateTestsHelperMethods {
     assert(transport.expectMsgType[Pong].data.size === ping.pongLength)
   }
 
+  test("send a ping if no message received for 30s") { f =>
+    import f._
+    connect(remoteNodeId, authenticator, watcher, router, relayer, connection, transport, peer)
+    // we make the transport send a message, this will delay the sending of a ping
+    val dummy = updates.head
+    for (_ <- 1 to 5) {
```
t-bast: IIUC the reason you're making many messages at an interval < 30 seconds is to verify that those messages correctly delay the ping sent by the peer.
I think there should be a better comment to explain that, my first reaction was "why do you do that loop?".

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -192,7 +192,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
         // we will delay all rebroadcasts with this value in order to prevent herd effects (each peer has a different delay)
         val rebroadcastDelay = Random.nextInt(nodeParams.routerConf.routerBroadcastInterval.toSeconds.toInt).seconds
         log.info(s"rebroadcast will be delayed by $rebroadcastDelay")
-        goto(CONNECTED) using ConnectedData(d.address_opt, d.transport, d.localInit, remoteInit, d.channels.map { case (k: ChannelId, v) => (k, v) }, rebroadcastDelay) forMax (30 seconds) // forMax will trigger a StateTimeout
+        scheduleNextPing()
```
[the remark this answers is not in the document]
pm47: Nice catch, I indeed changed my mind and moved it to the `onTransition` which I thought was safer.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
pm47: Right, the intent is definitely not obvious.

PR #1341, opened 2020-03-05
Kanela-agent 1.0.1->1.0.5

pm47:
NB: using `addJava "-javaagent:$lib_dir/kanela-agent-*.jar"` in scripts
to make them update proof doesn't work unfortunately.


araspitzu:
Too bad the wildcard doesn't work, changes ACK.

PR #1342, opened 2020-03-06
Re-send ChannelUpdate on reconnection (vol 2)

pm47:
Previous implementation in #1317 wasn't working because of a bug in the
transition. Added a test and fixed it.

cc @t-bast @btcontract

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
@@ -573,7 +573,7 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     bob2alice.forward(alice)
     alice2bob.forward(bob)
 
-    //  point the channel still isn't deeply buried: channel_update isn't sent again
+    //  at this point the channel still isn't deeply buried: channel_update isn't sent again
     alice2bob.expectNoMsg()
     bob2alice.expectNoMsg()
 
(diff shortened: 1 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
```diff
@@ -534,4 +534,69 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
...
+    awaitCond(alice.stateName == OFFLINE)
+    awaitCond(bob.stateName == OFFLINE)
+    sender.send(alice, INPUT_RECONNECTED(alice2bob.ref, aliceInit, bobInit))
+    sender.send(bob, INPUT_RECONNECTED(bob2alice.ref, bobInit, aliceInit))
+    alice2bob.expectMsgType[ChannelReestablish]
+    bob2alice.expectMsgType[ChannelReestablish]
+    bob2alice.forward(alice)
+    alice2bob.forward(bob)
+
+    //  point the channel still isn't deeply buried: channel_update isn't sent again
```
t-bast: nit:
```suggestion
    //  at this point the channel still isn't deeply buried: channel_update isn't sent again
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
```diff
@@ -534,4 +534,69 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
...
+    alice2bob.expectNoMsg()
+    bob2alice.expectNoMsg()
+
+    // funding tx gets 6 confirmations, channel is private so there is no announcement sigs
```
t-bast: You should have an `alice2bob.expectNoMsg()` to verify that no announcement sig is sent, don't you?

pm47: We receive `ChannelUpdate` instead.

t-bast: Isn't there a way to expect no message of type X?

PR #1345, opened 2020-03-10
Router property-based testing

t-bast:
Add property-based testing and start migrating some of our old-school fuzz tests.

By default our property-based tests run with a very small factor (10) so that they're fast (but they won't find many issues). This way `mvn test` is still fast, but still runs those tests (which is important to ensure they're not completely broken).

I defined a new pipeline for the CI, that runs only the fuzz tests over a lot more samples.
This pipeline is configured to run every night at 1am UTC (this is configured directly on the semaphore UI). Here is one example run of the nightly build: https://acinq.semaphoreci.com/workflows/b9eb193d-2f1b-43b3-af58-99eaa7a6daac

This allowed me to add some tests on the router that generate graphs with certain properties. This is an important first step before we start experimenting with changes to the path-finding algorithm.

### DIFF
--- a/.semaphore/fuzzing.yml
+++ b/.semaphore/fuzzing.yml
@@ -1,0 +1,25 @@
+version: v1.0
+name: Fuzz Tests
+agent:
+  machine:
+    type: e1-standard-4
+    os_image: ubuntu1804
+execution_time_limit:
+  minutes: 15
+
+blocks:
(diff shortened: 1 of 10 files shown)


t-bast:
I don't think this is a priority, closing for now to reduce noise.

PR #1347, opened 2020-03-11
Split the `Peer` in two

pm47:
Peer has been split in two and now handles only channel related stuff.

A new PeerConnection class is in charge of managing the BOLT 1 part
(init handshake, pings) and has the same lifetime as the underlying
connection.

Two possible next steps:
- have `TransportHandler` be a child of `PeerConnection`; this is easy
for outgoing connection, but not straightforward for incoming
connections because we don't know what the remote node id is before the
crypto hasndshake. We could work around that by either (1) making the
`remoteNodeId` an attribute of the state of `PeerConnection` instead of
a constructor argument (edit: done with 6b4d580) or (2) read the first few bytes of the incoming
connection before the crypto handshake to know the node id before it is
authenticated.
- put the reconnection logic outside of the `Peer` to simplify it even
further (done with 91a9b25).


eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -0,0 +1,486 @@
...
+
+  val IGNORE_NETWORK_ANNOUNCEMENTS_PERIOD = 5 minutes
+
+  def props(nodeParams: NodeParams, switchboard: ActorRef, router: ActorRef) = Props(new PeerConnection(nodeParams, switchboard, router))
+
+  // @formatter:off
+
+  sealed trait Data
+  sealed trait HasTransport extends Data { def transport: ActorRef }
+  case class Nothing() extends Data
```
t-bast: nit:

```suggestion
  case object Nothing extends Data
```

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -0,0 +1,486 @@
...
+  case class ConnectionReady(peerConnection: ActorRef, remoteNodeId: PublicKey, address: InetSocketAddress, outgoing: Boolean, localInit: wire.Init, remoteInit: wire.Init)
+
+  case class DelayedRebroadcast(rebroadcast: Rebroadcast)
+
+  sealed trait BadMessage
+  case class InvalidSignature(r: RoutingMessage) extends BadMessage
+  case class InvalidAnnouncement(c: ChannelAnnouncement) extends BadMessage
+  case class ChannelClosed(c: ChannelAnnouncement) extends BadMessage
+
+  case class Behavior(fundingTxAlreadySpentCount: Int = 0, fundingTxNotFoundCount: Int = 0, ignoreNetworkAnnouncement: Boolean = false)
```
t-bast: `fundingTxNotFoundCount` seems unused, remove?

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
```diff
@@ -0,0 +1,345 @@
...
+
+    val peerConnection: TestFSMRef[PeerConnection.State, PeerConnection.Data, PeerConnection] = TestFSMRef(new PeerConnection(aliceParams, switchboard.ref, router.ref))
+    withFixture(test.toNoArgTest(FixtureParam(remoteNodeId, switchboard, router, connection, transport, peerConnection, peer)))
+  }
+
+  def connect(remoteNodeId: PublicKey, switchboard: TestProbe, router: TestProbe, connection: TestProbe, transport: TestProbe, peerConnection: TestFSMRef[PeerConnection.State, PeerConnection.Data, PeerConnection], peer: TestProbe, remoteInit: wire.Init = wire.Init(Bob.nodeParams.features), expectSync: Boolean = false): Unit = {
+    // let's simulate a connection
+    val probe = TestProbe()
+    probe.send(peerConnection, PeerConnection.PendingAuth(connection.ref, Some(remoteNodeId), address, origin_opt = None, transport_opt = Some(transport.ref)))
+    transport.send(peerConnection, TransportHandler.HandshakeCompleted(remoteNodeId))
```
t-bast: Verify that we sent an `Authenticated` to the switchboard?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -620,31 +314,6 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
...
-    // Otherwise we check if this message has a timestamp that matches the timestamp filter.
-    val matchesFilter = (msg, gossipTimestampFilter_opt) match {
-      case (_, None) => false // BOLT 7: A node which wants any gossip messages would have to send this, otherwise [...] no gossip messages would be received.
-      case (hasTs: HasTimestamp, Some(GossipTimestampFilter(_, firstTimestamp, timestampRange))) => hasTs.timestamp >= firstTimestamp && hasTs.timestamp <= firstTimestamp + timestampRange
-      case _ => true // if there is a filter and message doesn't have a timestamp (e.g. channel_announcement), then we send it
-    }
-    isOurGossip || matchesFilter
-  }
-
   // a failing channel won't be restarted, it should handle its states
```
t-bast: nit: the comment isn't completely accurate anymore, we also have some `Client` children (and we should also let them stop themselves without restarting them).

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -0,0 +1,486 @@
...
+        val nodesSent = sendAndCount(rebroadcast.nodes)
+
+        if (channelsSent > 0 || updatesSent > 0 || nodesSent > 0) {
+          log.info(s"sent announcements: channels={} updates={} nodes={}", channelsSent, updatesSent, nodesSent)
+        }
+        stay
+
+      case Event(msg: GossipTimestampFilter, d: ConnectedData) =>
+        // special case: time range filters are peer specific and must not be sent to the router
+        sender ! TransportHandler.ReadAck(msg)
```
sstone: Use of `sender` vs `d.transport` is not consistent here

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -0,0 +1,486 @@
...
+      case Event(msg: wire.RoutingMessage, d: ConnectedData) =>
+        msg match {
+          case _: AnnouncementSignatures =>
+            // this is actually for the channel
+            d.transport ! TransportHandler.ReadAck(msg)
+            d.peer ! msg
+            stay
+          case _: ChannelAnnouncement | _: ChannelUpdate | _: NodeAnnouncement if d.behavior.ignoreNetworkAnnouncement =>
+            // this peer is currently under embargo!
+            sender ! TransportHandler.ReadAck(msg)
```
sstone: Same as above: use of `sender` vs `d.transport` is not consistent here

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -0,0 +1,486 @@
...
+   * This is implemented by scheduling a ping request every 30 seconds, and pushing it back every time we receive a
+   * message from the peer.
+   *
+   */
+  def heartbeat(s: StateFunction): StateFunction = {
+    case event if s.isDefinedAt(event) =>
+      event match {
+        case Event(_: LightningMessage, d: ConnectedData) if sender == d.transport =>
+          // this is a message from the peer, he's alive, we can delay the next ping
+          scheduleNextPing()
```
sstone: `scheduleNextPing()` is only called when we receive a LightningMessage -> if we never received anything once we're in `CONNECTED` state we'll never send a `Ping`

pm47:
TODO: add diagnostics logging

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -85,12 +82,10 @@ class Switchboard(nodeParams: NodeParams, authenticator: ActorRef, watcher: Acto
         case None => sender ! Status.Failure(new RuntimeException("no connection to peer"))
       }
 
-    case auth@Authenticator.Authenticated(_, _, remoteNodeId, _, _, _) =>
+    case authenticated: io.PeerConnection.Authenticated =>
```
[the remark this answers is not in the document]
pm47: I think an auto refactor is causing those

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -784,6 +787,22 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
...
+        .sliding(100, 100)
+        .map(NodesDiscovered)
+        .foreach(listener ! _)
+      d.channels
+        .values
+        .map(pc => SingleChannelDiscovered(pc.ann, pc.capacity, pc.update_1_opt, pc.update_2_opt))
+        .sliding(100, 100)
+        .map(ChannelsDiscovered)
+        .foreach(listener ! _)
+      listener ! "done"
```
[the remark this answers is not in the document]
pm47: Went with a dedicated `object`

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -784,6 +787,22 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
       self ! progress
       stay using d.copy(sync = sync1)
 
+    case Event(RegisterToChanges(listener), d) =>
```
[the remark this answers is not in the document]
pm47: There is some overlap, but the use case is a bit different. I went with:
```
case object GetRoutingState
case class GetRoutingStateStreaming(subscribeToChanges: Boolean)
```

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
pm47: Good catch. This is handled in the `InvalidAnnouncement` case, where we disconnect immediately. We could keep the count because disconnection is asynchronous, but it seems useless.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: It wasn't accurate before, we were already creating `Client` children.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
pm47: Great catch, and missing a test on that. Will add a timeout on the `Init` too.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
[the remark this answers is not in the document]
t-bast: Allright, sounds good for the `GetRoutingStateStreaming`.
I still find the `subscribeToChanges` weird though.
Can you explain why it's the sender wouldn't want to directly register to the event himself?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -161,25 +162,27 @@ case class ExcludeChannel(desc: ChannelDesc)
...
 /** Gossip that we received from a remote peer. */
-case class RemoteGossip(peer: ActorRef) extends GossipOrigin
+case class RemoteGossip(peerConnection: ActorRef, nodeId: PublicKey) extends GossipOrigin
 /** Gossip that was generated by our node. */
 case object LocalGossip extends GossipOrigin
 
 case class Stash(updates: Map[ChannelUpdate, Set[GossipOrigin]], nodes: Map[NodeAnnouncement, Set[GossipOrigin]])
 case class Rebroadcast(channels: Map[ChannelAnnouncement, Set[GossipOrigin]], updates: Map[ChannelUpdate, Set[GossipOrigin]], nodes: Map[NodeAnnouncement, Set[GossipOrigin]])
+
+case class PeerRoutingMessage(origin: RemoteGossip, message: RoutingMessage)
```
pm47: Not entirely sure with the moving/redefinition of `PeerRoutingMessage`, let me know what you think.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -463,10 +466,21 @@ object PeerConnection {
 
   case class DelayedRebroadcast(rebroadcast: Rebroadcast)
 
-  sealed trait BadMessage
-  case class InvalidSignature(r: RoutingMessage) extends BadMessage
-  case class InvalidAnnouncement(c: ChannelAnnouncement) extends BadMessage
-  case class ChannelClosed(c: ChannelAnnouncement) extends BadMessage
+  object Gossip {
```
t-bast: I would even move that to its own separate file (in the router package), where we can start adding utility functions in the future. I would move `GossipOrigin` and some other classes that are currently directly defined in the router package (in `Router.scala`) inside this new `Gossip.scala` file.

pm47:
> Why not start doing that in a separate PR? Let's finalize and merge the split between peer and peer connections and work on the router/gossip refactoring separately to avoid monster PRs and rebase/merge hell?

Agreed, have already reverted the commit 5f483ee and will add it to another PR, which also includes breaking up the router.

> On the gossip refactoring, I would take this opportunity to go even further and start moving as much as possible in that new Gossip object (try to break the monster router in smaller pieces, at least in terms of objects and functions - it might be easier to said than done though).

Agreed too, see above.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -96,7 +97,34 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, router: ActorRef
       require(remoteNodeId == remoteNodeId1, s"invalid nodeid: $remoteNodeId != $remoteNodeId1")
       log.debug(s"got authenticated connection to $remoteNodeId@${address.getHostString}:${address.getPort}")
 
-      context watch peerConnection
+      // We want to log all incoming/outgoing messages to/from this peer. Logging incoming messages is easy (they all go
+      // through this actor), but for outgoing messages it's a bit trickier because channels directly send messages to
+      // the connection. That's why we use this pass-through actor that just logs messages and forward them.
+      val logConnection = context.actorOf(Props(new Actor {
```
t-bast: It feels like we could generalize this. Maybe creating a `LightningMessageLogger` actor that wraps a child actor and takes some extra configuration for some MDC values that should be fixed (`remoteNodeId` in this case)?

It feels like `Peer` could also be wrapped in such a logging actor instead of defining almost the same thing inside an `aroundReceive`.

Not entirely sure it would work, just my 2 cents.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -96,7 +97,34 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, router: ActorRef
...
+        // we use this to log raw messages coming in and out of the peer
+        val logMsgOut = new BusLogging(context.system.eventStream, "", classOf[Peer.MessageLogs], context.system.asInstanceOf[ExtendedActorSystem].logFilter) with DiagnosticLoggingAdapter
+        context watch peerConnection
+        override def receive: Receive = {
+          case Terminated(_) =>
+            context stop self
+          case msg =>
+            msg match {
+              case _: LightningMessage =>
+                logMsgOut.mdc(mdc(msg))
```
t-bast: Is the inner call to `mdc(msg)` really safe (since DiagnosticActorLogging explicitly mentions that it's not thread-safe)? It feels like the Peer's `mdc` function can be executing in two different threads at the same time if we're processing messages in both the `Peer` and `PeerConnection` concurrently (which we will).

And how would that captured lambda even work when a Peer and its PeerConnection end up on different hosts?

eclair-core/src/test/scala/fr/acinq/eclair/router/RoutingSyncSpec.scala
```diff
@@ -302,12 +304,9 @@ object RoutingSyncSpec {
     outputIndex <- 0 to 1
   } yield ShortChannelId(block, txindex, outputIndex)).foldLeft(SortedSet.empty[ShortChannelId])(_ + _)
 
-  // this map will store private keys so that we can sign new announcements at will
-  val pub2priv: mutable.Map[PublicKey, PrivateKey] = mutable.HashMap.empty
```
t-bast: Why not use a `concurrent.Map`?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: Setting/clearing the MDC isn't thread because it uses thread-local storage, but the `def mdc(currentMessage: Any): MDC` method is threadsafe, it's a simple stateless method that just returns a new immutable Map.

pm47: > And how would that captured lambda even work when a Peer and its PeerConnection end up on different hosts?

Not sure what you mean

eclair-core/src/test/scala/fr/acinq/eclair/router/RoutingSyncSpec.scala
pm47: I started with that, but I've done to much actor programming and can't get myself to share state anymore...

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: Ok I got confused there thinking it would somehow be stateful and would ruin everything. This can probably be ignored, your answers make sense.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -37,7 +37,25 @@ import scala.concurrent.duration._
 import scala.util.Random
 
 /**
- * Created by PM on 26/08/2016.
+ * This actor represents a protocol-level connection to another lightning node. As soon as a TCP connection is established
+ * with a peer, a new [[PeerConnection]] will be spawned and the raw connection actor will be passed to it. This is done by
+ * the [[Peer]] or a [[Client]] depending on whether it is an incoming our outgoing connection.
```
t-bast: nit:
```suggestion
 * the [[Peer]] or a [[Client]] depending on whether it is an incoming or outgoing connection.
```

pm47:
Phew, took 91a9b25 took more time than I thought.

@sstone @t-bast interested in your thoughts regarding the handling of user-requested connection requests (`Peer.Connect`) vs automated reconnections (handled by the `ReconnectionTask` actor). They are really independent and I could have kept the processing of `Peer.Connect` entirely within the `Peer` actor. I moved it to `ReconnectionTask` in order to factor the calls to `Client`, but it's a bit clunky and some parts still remain in the `Peer` (e.g. rejection in state `CONNECTED`). On the plus side, the `Peer` is really simple now.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -34,75 +34,45 @@ import fr.acinq.eclair.{wire, _}
...
- * The [[Peer] actor is mostly in charge of managing channels, stays alive when the peer is disconnected, and only dies
+ * The [[Peer]] actor is mostly in charge of managing channels, stays alive when the peer is disconnected, and only dies
  * when there are no more channels.
  *
  * Everytime a new connection is established, it is sent to the [[Peer]] and replaces the previous one.
  *
  * Created by PM on 26/08/2016.
  */
-class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, router: ActorRef, watcher: ActorRef, relayer: ActorRef, paymentHandler: ActorRef, wallet: EclairWallet) extends FSMDiagnosticActorLogging[Peer.State, Peer.Data] {
+class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, switchboard: ActorRef, router: ActorRef, watcher: ActorRef, relayer: ActorRef, paymentHandler: ActorRef, wallet: EclairWallet) extends FSMDiagnosticActorLogging[Peer.State, Peer.Data] {
```
t-bast: Why do we need to add the switchboard as dependency? Can't we assume that `context.parent` will always be the switchboard? Or is it because it's more future-proof and easier to test (which would be a completely valid argument)?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -34,75 +34,45 @@ import fr.acinq.eclair.{wire, _}
...
 
   import Peer._
 
+  val reconnectionTask: ActorRef = context.actorOf(ReconnectionTask.props(nodeParams, remoteNodeId, switchboard, router), "reconnection-task")
+  // we don't use FSM.SubscribeTransitionCallBack(reconnectionTask) because we don't the reconnectionTask to miss the first transition
```
t-bast: Can you clarify? It's the initial going to `INSTANTIATING` that we don't want to miss?

nit: I think this sentence is missing "want":

```suggestion
  // we don't use FSM.SubscribeTransitionCallBack(reconnectionTask) because we don't want the reconnectionTask to miss the first transition
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -34,75 +34,45 @@ import fr.acinq.eclair.{wire, _}
...
-          stay using d.copy(nextReconnectionDelay = nextReconnectionDelay(d.nextReconnectionDelay, nodeParams.maxReconnectInterval))
-      }
+    case Event(p: Peer.Connect, _) =>
+      reconnectionTask forward p
+      stay
 
     case Event(PeerConnection.ConnectionReady(peerConnection, remoteNodeId1, address, outgoing, localInit, remoteInit), d: DisconnectedData) =>
       require(remoteNodeId == remoteNodeId1, s"invalid nodeid: $remoteNodeId != $remoteNodeId1")
-      log.debug(s"got authenticated connection to $remoteNodeId@${address.getHostString}:${address.getPort}")
+      log.debug(s"got authenticated connection to address ${address.getHostString}:${address.getPort}")
```
t-bast: nit:
```suggestion
      log.debug("got authenticated connection to address {}:{}", address.getHostString, address.getPort)
```

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
```diff
@@ -0,0 +1,189 @@
...
+  // @formatter:off
+  sealed trait Data
+  case object Nothing extends Data
+  case class ConnectingData(connection: ActorRef, to: InetSocketAddress, nextReconnectionDelay: FiniteDuration) extends Data
+  case class WaitingData(nextReconnectionDelay: FiniteDuration) extends Data
+  // @formatter:on
+
+  def getPeerAddressFromDb(peersDb: PeersDb, networkDb: NetworkDb, remoteNodeId: PublicKey): Option[InetSocketAddress] = {
+    peersDb.getPeer(remoteNodeId) // TODO should we start with the network db which may be more up to date? or rotate?
+      .orElse(networkDb.getNode(remoteNodeId).flatMap(_.addresses.headOption)) // TODO gets the first of the list, improve selection?
```
t-bast: It would be interesting to grab the peer and network DBs from our node, and compare the addresses in both to see what discrepancies there are. It feels to me that if you're online most of the time, the peer DB should be more up-to-date than the network DB (provided that when our peer connects to us from a new address, we correctly save that address to the peer DB).

Doing a one-time analysis on those DBs will help us see what happens in the real world and should let us address these TODOs.

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
```diff
@@ -0,0 +1,189 @@
...
+      .map(_.socketAddress)
+  }
+
+  def hostAndPort2InetSocketAddress(hostAndPort: HostAndPort): InetSocketAddress = new InetSocketAddress(hostAndPort.getHost, hostAndPort.getPort)
+
+  /**
+   * This helps preventing peers reconnection loops due to synchronization of reconnection attempts.
```
t-bast: nit:
```suggestion
   * This helps prevent peers reconnection loops due to synchronization of reconnection attempts.
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: It's easier to test, I got bitten with the `context.parent` and preferred to make things simpler.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: Here is what happens when you send an `FSM.SubscribeTransitionCallBack` to an FSM:

```scala
    case SubscribeTransitionCallBack(actorRef) ⇒
      // TODO Use context.watch(actor) and receive Terminated(actor) to clean up list
      listeners.add(actorRef)
      // send current state back as reference point
      actorRef ! CurrentState(self, currentState.stateName)
```

If we make the `ReconnectionTask` subscribe, or even have the `Peer` subscribe its child actor by sending a message to itself, there would be a race between `FSM.SubscribeTransitionCallBack` and `Peer.Init` (and potentially other messages like a `Peer.Connect`). This would complicate the `ReconnectionTask` because the first transition would be unpredictable. Sure, we would be able to sort things out by looking at the `CurrentState` but it's just more work.

Making the `Peer` synchronously register the `ReconnectionTask` as a listener removes the race condition. The `ReconnectionTask` can now safely assume that it will always receive the same initial transitions (`INSTANTIATING`->`INSTANTIATING` and then `INSTANTIATING`->`DISCONNECTED`). We can then get rid of the `CurrentState` handling.

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
pm47: Indeed, but if the peer changes ip the network db will be more accurate. There is some material for a future PR... 

> (provided that when our peer connects to us from a new address, we correctly save that address to the peer DB)

https://github.com/ACINQ/eclair/blob/split-peer/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala#L107-L111

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
```diff
@@ -0,0 +1,189 @@
...
+        val firstNextReconnectionDelay = previousState match {
+          case Peer.INSTANTIATING =>
+            // When restarting, we will ~immediately reconnect, but then:
+            // - we don't want all the subsequent reconnection attempts to be synchronized (herd effect)
+            // - we don't want to go through the exponential backoff delay, because we were offline, not them, so there is no
+            // reason to eagerly retry
+            // That's why we set the next reconnection delay to a random value between MAX_RECONNECT_INTERVAL/2 and MAX_RECONNECT_INTERVAL.
+            nodeParams.maxReconnectInterval.minus(Random.nextInt(nodeParams.maxReconnectInterval.toSeconds.toInt / 2).seconds)
+          case Peer.CONNECTED =>
+            log.info("peer is disconnected")
```
t-bast: It feels like this log is out of place, maybe put it in the Peer's transition from CONNECTED -> DISCONNECTED? Same for the other `log.info` about transitions.
It feels like you used these log lines to verify the state transitions of the reconnectTask, right? In that case you could instead add a `log.debug` on all state transitions of this actor?

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
```diff
@@ -0,0 +1,189 @@
...
+    case Event(Peer.Connect(_, hostAndPort_opt), _) =>
+      // this happens completely independently of the automated reconnection process
+      // if we are already connecting/connected, the peer will kill any duplicate connections
+      hostAndPort_opt
+        .map(hostAndPort2InetSocketAddress)
+        .orElse(getPeerAddressFromDb(nodeParams.db.peers, nodeParams.db.network, remoteNodeId)) match {
+        case Some(address) => connect(address, Some(sender))
+        case None => sender ! "no address found"
+      }
+      stay
```
t-bast: Shouldn't we go to CONNECTING? If we do that you probably wouldn't need the `case Event(_: Terminated, _)` in the `whenUnhandled` (or am I missing something?).

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -146,421 +129,143 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
...
-      case Event(Terminated(actor), d: ConnectedData) if d.channels.values.toSet.contains(actor) =>
-        // we will have at most 2 ids: a TemporaryChannelId and a FinalChannelId
-        val channelIds = d.channels.filter(_._2 == actor).keys
-        log.info(s"channel closed: channelId=${channelIds.mkString("/")}")
-        if (d.channels.values.toSet - actor == Set.empty) {
-          log.info(s"that was the last open channel, closing the connection")
-          d.transport ! PoisonPill
-        }
-        stay using d.copy(channels = d.channels -- channelIds)
+    case Event(Disconnect(nodeId), d: ConnectedData) if nodeId == remoteNodeId =>
```
t-bast: It looks like if we manually disconnect via this event, the reconnectionTask will still reconnect us afterwards :confounded: 
Maybe we need a field in `DisconnectedData` to mention that it's disconnected from a user request and should stay disconnected until a `Peer.Connect`?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -146,421 +129,143 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: A
...
+      // we will have at most 2 ids: a TemporaryChannelId and a FinalChannelId
+      val channelIds = d.channels.filter(_._2 == actor).keys
+      log.info(s"channel closed: channelId=${channelIds.mkString("/")}")
+      if (d.channels.values.toSet - actor == Set.empty) {
+        log.info(s"that was the last open channel, closing the connection")
+        d.peerConnection ! PoisonPill
+      }
+      stay using d.copy(channels = d.channels -- channelIds)
+
+    case Event(connectionReady: PeerConnection.ConnectionReady, d: ConnectedData) =>
```
t-bast: Hah this is one is painful...once we go to DISCONNECTED the reconnectTask will also try to reconnect for no reason.
Do you think the randomization in the reconnectTask is enough, or should we explicitly tell the reconnectTask to NOT reconnect (which may lead to way too many particular cases to handle :/)?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -28,109 +29,91 @@ import fr.acinq.eclair.Features.Wumbo
...
+ *
  * Created by PM on 26/08/2016.
  */
-class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, authenticator: ActorRef, watcher: ActorRef, router: ActorRef, relayer: ActorRef, paymentHandler: ActorRef, wallet: EclairWallet) extends FSMDiagnosticActorLogging[Peer.State, Peer.Data] {
+class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, switchboard: ActorRef, router: ActorRef, watcher: ActorRef, relayer: ActorRef, paymentHandler: ActorRef, wallet: EclairWallet) extends FSMDiagnosticActorLogging[Peer.State, Peer.Data] {
 
   import Peer._
 
-  startWith(INSTANTIATING, Nothing())
+  val reconnectionTask: ActorRef = context.actorOf(ReconnectionTask.props(nodeParams, remoteNodeId, switchboard, router), "reconnection-task")
```
t-bast: We're not watching the `reconnectionTask`, so what happens if it crashes? Shouldn't we restart it?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: Ok, that makes sense. Maybe we can make the comment clearer, something like: `we register the reconnection task to our transitions ourselves; if we let the child actor register itself, there is a race condition and it may miss the first transition.`?

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
t-bast: > Indeed, but if the peer changes ip the network db will be more accurate

It depends...if our remote changes IP while it's connected, it will kill the TCP session but the remote should reconnect to us with the new IP very quickly. The `node_announcement` with its new IP will probably take longer to reach us. It's just a supposition though, looking at what's inside our DB with timestamps will probably help us learn how it behaves in the real world!

pm47: > the remote should reconnect to us with the new IP very quickly

We only store the ip info for outgoing connections, because most of the time the peer isn't reachable.

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
pm47: It's there because I didn't want to print "peer is disconnected" on startup. Those logs should be debug anyway. An equivalent log already exists in `Peer`.

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
pm47: No, that's the tricky part (see my comment https://github.com/ACINQ/eclair/pull/1347#issuecomment-600803658). The user-requested connect is treated independently of the regular automated connection loop. Feels a bit hackish at first but it makes sense.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: > It looks like if we manually disconnect via this event, the reconnectionTask will still reconnect us afterwards

Yes, but that will happen only if we have existing channels and I don't know what the use case would be for not reconnecting in that case. FWIW `Disconnect` is mostly useful for tests and debugging, and we can also disable automated reconnection altogether with the `auto-reconnect` config parameter (which we use in tests).

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: The randomization is there for precisely that purpose, not sure what we can do more. What if we refrain from reconnecting and our peer does the same?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: Restarting it won't be enough as we are making assumptions on the initial case. I think not reconnecting is relatively harmless and I would just let it stop. We would still be reachable but would not attempt to make connections.

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
t-bast: Oh yeah true, that makes sense

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: That's true, it's probably an edge-case we can safely ignore

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: Yeah this is probably working in practice, it's just really hard to evaluate by just looking at the code...but we'll learn as we watch what happens on our node

t-bast:
I'm not a big fan of d299666ddf14d3a2199cd2dc53477e5b1170bc96
If the automatic timer is being annoying, maybe what you want is to just tweak the nodeParams to raise the `nodeParams.maxReconnectInterval`? It would be less invasive.

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
t-bast: Ok, I get it, this is indeed a bit subtle, maybe worth a comment above that `stay`?

pm47:
> I'm not a big fan of d299666
If the automatic timer is being annoying, maybe what you want is to just tweak the nodeParams to raise the nodeParams.maxReconnectInterval? It would be less invasive.

Neither am I. Increasing `nodeParams.maxReconnectInterval` would not completely remove the race...

pm47:
>> I'm not a big fan of d299666
    If the automatic timer is being annoying, maybe what you want is to just tweak the nodeParams to raise the nodeParams.maxReconnectInterval? It would be less invasive.

>Neither am I. Increasing nodeParams.maxReconnectInterval would not completely remove the race...


Ok, how about:
```diff
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala b/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
index bb4d6b430..798dc0e25 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
@@ -54,7 +54,7 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey, switchbo
   when(CONNECTING) {
     case Event(Terminated(actor), d: ConnectingData) if actor == d.connection =>
       log.info(s"connection failed, next reconnection in ${d.nextReconnectionDelay.toSeconds} seconds")
-      setTimer(RECONNECT_TIMER, if (bypassTimer) 'dummy else TickReconnect, d.nextReconnectionDelay, repeat = false)
+      setReconnectTimer(d.nextReconnectionDelay)
       goto(WAITING) using WaitingData(nextReconnectionDelay(d.nextReconnectionDelay, nodeParams.maxReconnectInterval))

     case Event(FSM.Transition(_, Peer.DISCONNECTED, Peer.CONNECTED), _) =>
@@ -105,7 +105,7 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey, switchbo
             log.info("peer is disconnected")
             nextReconnectionDelay(initialDelay, nodeParams.maxReconnectInterval)
         }
-        setTimer(RECONNECT_TIMER, if (bypassTimer) 'dummy else TickReconnect, initialDelay, repeat = false)
+        setReconnectTimer(initialDelay)
         goto(WAITING) using WaitingData(firstNextReconnectionDelay)
       } else {
         stay
@@ -135,6 +135,10 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey, switchbo
       stay
   }

+  private def setReconnectTimer(delay: FiniteDuration) = if (!bypassTimer) {
+    setTimer(RECONNECT_TIMER, TickReconnect, delay, repeat = false)
+  }
+
   private def connect(address: InetSocketAddress, origin_opt: Option[ActorRef]): ActorRef = {
     log.info(s"connecting to $address")
     val connection = context.actorOf(Client.props(nodeParams, switchboard, router, address, remoteNodeId, origin_opt = origin_opt))
```

t-bast:
It's slightly better, I'm still not a fan of having an internal variable only for tests, but at least it's quite isolated.

pm47:
New idea with overriding a method instead of using a variable. It keeps the original class mostly untouched (just make a method `protected`):

```scala
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala b/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
index bb4d6b430..6cf89a5e0 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
@@ -46,15 +46,12 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey, switchbo
 
   import ReconnectionTask._
 
-  // this is used in tests to control transitions
-  var bypassTimer = false
-
   startWith(IDLE, Nothing)
 
   when(CONNECTING) {
     case Event(Terminated(actor), d: ConnectingData) if actor == d.connection =>
       log.info(s"connection failed, next reconnection in ${d.nextReconnectionDelay.toSeconds} seconds")
-      setTimer(RECONNECT_TIMER, if (bypassTimer) 'dummy else TickReconnect, d.nextReconnectionDelay, repeat = false)
+      setReconnectTimer(d.nextReconnectionDelay)
       goto(WAITING) using WaitingData(nextReconnectionDelay(d.nextReconnectionDelay, nodeParams.maxReconnectInterval))
 
     case Event(FSM.Transition(_, Peer.DISCONNECTED, Peer.CONNECTED), _) =>
@@ -105,7 +102,7 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey, switchbo
             log.info("peer is disconnected")
             nextReconnectionDelay(initialDelay, nodeParams.maxReconnectInterval)
         }
-        setTimer(RECONNECT_TIMER, if (bypassTimer) 'dummy else TickReconnect, initialDelay, repeat = false)
+        setReconnectTimer(initialDelay)
         goto(WAITING) using WaitingData(firstNextReconnectionDelay)
       } else {
         stay
@@ -135,6 +132,8 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey, switchbo
       stay
   }
 
+  protected def setReconnectTimer(delay: FiniteDuration) = setTimer(RECONNECT_TIMER, TickReconnect, delay, repeat = false)
+
   private def connect(address: InetSocketAddress, origin_opt: Option[ActorRef]): ActorRef = {
     log.info(s"connecting to $address")
     val connection = context.actorOf(Client.props(nodeParams, switchboard, router, address, remoteNodeId, origin_opt = origin_opt))
diff --git a/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala b/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
index 24ef10084..06d0c9826 100644
--- a/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
@@ -49,7 +49,13 @@ class ReconnectionTaskSpec extends TestkitBaseClass with StateTestsHelperMethods
       aliceParams.db.network.addNode(bobAnnouncement)
     }
 
-    val reconnectionTask: TestFSMRef[ReconnectionTask.State, ReconnectionTask.Data, ReconnectionTask] = TestFSMRef(new ReconnectionTask(aliceParams, remoteNodeId, TestProbe().ref, TestProbe().ref))
+    val reconnectionTask: TestFSMRef[ReconnectionTask.State, ReconnectionTask.Data, ReconnectionTask] = if (test.tags.contains("disable_tick_reconnect")) {
+      TestFSMRef(new ReconnectionTask(aliceParams, remoteNodeId, TestProbe().ref, TestProbe().ref) {
+        override protected def setReconnectTimer(delay: FiniteDuration): Unit = () // allows control of transitions in test
+      })
+    } else {
+      TestFSMRef(new ReconnectionTask(aliceParams, remoteNodeId, TestProbe().ref, TestProbe().ref))
+    }
 
     val monitor = TestProbe()
     reconnectionTask ! SubscribeTransitionCallBack(monitor.ref)
@@ -89,11 +95,9 @@ class ReconnectionTaskSpec extends TestkitBaseClass with StateTestsHelperMethods
     assert(expectedNextReconnectionDelayInterval contains connectingData.nextReconnectionDelay.toSeconds) // we only reconnect once
   }
 
-  test("reconnect with increasing delays", Tag("auto_reconnect")) { f =>
+  test("reconnect with increasing delays", Tag("auto_reconnect"), Tag("disable_tick_reconnect")) { f =>
     import f._
 
-    reconnectionTask.underlyingActor.bypassTimer = true
-
     val probe = TestProbe()
     val peer = TestProbe()
     nodeParams.db.peers.addOrUpdatePeer(remoteNodeId, NodeAddress.fromParts("localhost", 42).get)

```

pm47:
@t-bast this time I got it right with 8942b7c82

t-bast:
Looks good, I only have two remaining comments on things that I feel can be improved, and then we can go to testing on Endurance and merge if everything goes well:

- https://github.com/ACINQ/eclair/pull/1347#discussion_r395161186
- https://github.com/ACINQ/eclair/pull/1347#discussion_r393630326

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: The problem we have here is that incoming messages are received by the `Peer`, but outgoing messages (mostly sent by `Channel`) are directly sent to the `PeerConnection`, thus bypassing the `Peer`. I don't see how your proposal addressed this issue.

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
pm47: The comment is there already (lines 124,125), not sure what I can add to make it clearer

t-bast: I would add that this explains why we don't modify our state (it's far from obvious IMO).
Something like:
```
// manual connection requests happen completely independently of the automated reconnection process;
// we initiate a connection but don't modify our state.
// if we are already connecting/connected, the peer will kill any duplicate connections
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: Yes but that's completely symmetrical: the code for `logMsgIn` here is very close to the code for `logMsgOut`.
It feels to me that this `logMsgOut` is just the same of `logMsgIn`, but from the perspective of the `PeerConnection`, isn't it?

sstone: There should be an explicit comment that explains that `logConnection` is to be used as `remote` whenever we send a `connected` notification.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -0,0 +1,519 @@
...
+      goto(BEFORE_INIT) using BeforeInitData(remoteNodeId, d.pendingAuth, d.transport)
+
+    case Event(AuthTimeout, _) =>
+      log.warning(s"authentication timed out after ${nodeParams.authTimeout}")
+      stop(FSM.Normal)
+  }
+
+  when(BEFORE_INIT) {
+    case Event(InitializeConnection(peer), d: BeforeInitData) =>
+      d.transport ! TransportHandler.Listener(self)
```
sstone: Here we're relying on `switchboard` replying back with an `InitializeConnection` message, shouldn't we use a timeout for this state ?

pm47: I've added state timeout only for adversarial scenarii. In this case if because of an internal problem we get stuck, then we can let our peer deal with that right? Otherwise we would want to be defensive we need a *lot* of timeouts all over the codebase?

sstone: True, but having to split the connection workflow here (because for incoming connections switchboard does not have a peer at this point) does not feel 100% right

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
pm47: Done with 5106463

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: Tried a different strategy with 647461f, wdyt?

t-bast: I like https://github.com/ACINQ/eclair/commit/647461f9edc3b2b75bb61f597cfa3533a1024e12 better than the logging actor, I think it's clearer. I don't think the forwarder was really needed, the changes in `Channel.scala` aren't very scary. It's a bit easier to follow the flow if we know that everything goes through the peer IMO.

Maybe we can use an explicit type instead of a `Tuple` in the channel's `send`, and a couple refactoring in that commit and it sounds good to me.

pm47: Okay, let's see what @sstone thinks.

It does feel better, even if it messes up the tests a little bit.

sstone: It looks much better indeed but I had the same reaction as @t-bast : using an Option[ActorRef ] to store the channel's active connection feels a bit more natural

pm47: Wdyt of my answer?

> I think deadletters is better here, because this is what will happen in the normal case when a connection dies. Allows us to only have 2 different states.

pm47:
Squashed

PR #1349, opened 2020-03-17
Improve AddHtlcFailed

t-bast:
I recommend reading each commit separately.
Each commit message should explain the problem it's trying to fix.


pm47:
Even before reviewing this, the core issue with `AddHtlcFailed` is that it is imo wrongly used in the closing scenario. Initially, its goal was to handle cases where the outgoing htlc wouldn't exist, meaning that we didn't have an identifier. I wonder if we should add another separate class that is used specifically for the closing scenario, where the identifier is required.

Another lead could be to send a `CMD_FAIL_HTLC` directly from the downstream actor? Or maybe send a `Relayer.ForwardLocalFail` which would be consistent with the success case.

t-bast:
I amended the last commit; please read its commit message and let me know if that sounds right.
I went with nesting `AddHtlcFailed` below a more general `UpdateFailLocalHtlc` trait (that intentionally feels like `UpdateFailHtlc`: we need to behave as if we received an `UpdateFailHtlc`).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -602,7 +601,6 @@ object Helpers {
       val channelKeyPath = keyManager.channelKeyPath(localParams, channelVersion)
       val localHtlcPubkey = Generators.derivePubKey(keyManager.htlcPoint(channelKeyPath).publicKey, remoteCommit.remotePerCommitmentPoint)
       val remoteHtlcPubkey = Generators.derivePubKey(remoteParams.htlcBasepoint, remoteCommit.remotePerCommitmentPoint)
-      val localPerCommitmentPoint = keyManager.commitmentPoint(channelKeyPath, commitments.localCommit.index.toInt)
```
t-bast: Is it expected that this value was unused? Looks like a check was removed at some point?

pm47: Has been added by @sstone in 24dadff6253278662d95bd1bc4871cddff706fe4 when reworking the key manager and used by another variable `localRevocationPubKey` which was unused. I then cleaned up `localRevocationPubKey` in be4cf6b6940c0a8d95988420c2c24217954d106d without noticing that it was the only variable using `localPerCommitmentPoint `, and there we are...

pm47:
> I amended the last commit; please read its commit message and let me know if that sounds right.
I went with nesting AddHtlcFailed below a more general UpdateFailLocalHtlc trait (that intentionally feels like UpdateFailHtlc: we need to behave as if we received an UpdateFailHtlc).

I think we need to go deeper (said with the voice of Eddy Murphy in Beverly Hills Cop). It should probably be closer to a `CMD_*` message than an `Update*` message, since it's generated locally. Looking at the fulfill scenario, we see here that it is also hackish:

https://github.com/ACINQ/eclair/blob/2df07277bb06d81236875a35b03662d89192ecbc/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala#L846-L847

I *think* the reason we do it that way is that we don't know within that method what the upstream channel is, which prevents us from building a `CMD_FULFILL_HTLC`/`CMD_FAIL_HTLC` message. But we do have the origin one step higher in the channel:

https://github.com/ACINQ/eclair/blob/2df07277bb06d81236875a35b03662d89192ecbc/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1264-L1265

The more I think about it, the more I think we need a `ForwardLocalFulfill`/`ForwardLocalFail` (or `Local`->`Onchain`?), defined in the relayer next to the other `Forward*` as:

```scala
case class ForwardOnchainFulfill(preimage: ByteVector32, to: Origin, htlc: UpdateAddHtlc) extends ForwardMessage
case class ForwardOnchainFail(cause: ChannelException, to: Origin, htlc: UpdateAddHtlc) extends ForwardMessage
```

Those would then be converted to `CMD_FAIL/FULFILL_HTLC` by the relayer just like the other `Forward*` handlers and everything would be nice and consistent.

I think the `AddHtlcFailed` class has nothing to do with the resolution of htlcs on-chain and should be left alone; it was a mistake to use it in the first place.

WDYT? I can submit a PR if you want.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Ok so we can remove it safely :)

t-bast:
Got it, I think that makes sense. We're currently trying to hide the fact that some local fail/fulfills happen on-chain because we think they need to be handled the same way as remote fail/fulfills, but that's not always true.

I'll amend the last commit today to add `ForwardOnchainFail`/`ForwardOnchainFulfill` and we'll see how deep that needs to go.

t-bast:
@pm47 done, I feel it's much better indeed! Let me know what you think of the two latest commits (I think the first 3 are quite non-controversial, but please review them as well :grin:). Maybe we can *still go deeper*, let me know if you have ideas (or maybe we can keep it at that for now).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -222,11 +210,13 @@ object Relayer extends Logging {
...
   sealed trait ForwardMessage
   case class ForwardAdd(add: UpdateAddHtlc, previousFailures: Seq[AddHtlcFailed] = Seq.empty) extends ForwardMessage
-  case class ForwardFulfill(fulfill: UpdateFulfillHtlc, to: Origin, htlc: UpdateAddHtlc) extends ForwardMessage
-  case class ForwardOnChainFulfill(preimage: ByteVector32, to: Origin, htlc: UpdateAddHtlc) extends ForwardMessage
-  case class ForwardFail(fail: UpdateFailHtlc, to: Origin, htlc: UpdateAddHtlc) extends ForwardMessage
-  case class ForwardFailMalformed(fail: UpdateFailMalformedHtlc, to: Origin, htlc: UpdateAddHtlc) extends ForwardMessage
-  case class ForwardOnChainFail(to: Origin, htlc: UpdateAddHtlc) extends ForwardMessage
+  sealed trait ForwardFulfill extends ForwardMessage { val fulfill: UpdateFulfillHtlc; val to: Origin; val htlc: UpdateAddHtlc }
+  case class ForwardRemoteFulfill(fulfill: UpdateFulfillHtlc, to: Origin, htlc: UpdateAddHtlc) extends ForwardFulfill
+  case class ForwardOnChainFulfill(preimage: ByteVector32, to: Origin, htlc: UpdateAddHtlc) extends ForwardFulfill { override val fulfill  = UpdateFulfillHtlc(htlc.channelId, htlc.id, preimage) }
```
pm47: why do you generate a fake `fulfill` instead of building a `CMD_FULFILL_HTLC` like in the failure case?

pm47: Gave it a try, seems to work well (NB: needs to also update `PaymentLifecycle`):

```diff
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
index aedd85664..838bba6a0 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
@@ -157,9 +157,9 @@ class Relayer(nodeParams: NodeParams, router: ActorRef, register: ActorRef, comm

     case ff: ForwardFulfill => ff.to match {
       case Origin.Local(_, None) => postRestartCleaner forward ff
-      case Origin.Local(_, Some(sender)) => sender ! ff.fulfill
+      case Origin.Local(_, Some(sender)) => sender ! ff.preimage
       case Origin.Relayed(originChannelId, originHtlcId, amountIn, amountOut) =>
-        val cmd = CMD_FULFILL_HTLC(originHtlcId, ff.fulfill.paymentPreimage, commit = true)
+        val cmd = CMD_FULFILL_HTLC(originHtlcId, ff.preimage, commit = true)
         commandBuffer ! CommandBuffer.CommandSend(originChannelId, cmd)
         context.system.eventStream.publish(ChannelPaymentRelayed(amountIn, amountOut, ff.htlc.paymentHash, originChannelId, ff.htlc.channelId))
       case Origin.TrampolineRelayed(_, None) => postRestartCleaner forward ff
@@ -210,9 +210,9 @@ object Relayer extends Logging {
   // @formatter:off
   sealed trait ForwardMessage
   case class ForwardAdd(add: UpdateAddHtlc, previousFailures: Seq[AddHtlcFailed] = Seq.empty) extends ForwardMessage
-  sealed trait ForwardFulfill extends ForwardMessage { val fulfill: UpdateFulfillHtlc; val to: Origin; val htlc: UpdateAddHtlc }
-  case class ForwardRemoteFulfill(fulfill: UpdateFulfillHtlc, to: Origin, htlc: UpdateAddHtlc) extends ForwardFulfill
-  case class ForwardOnChainFulfill(preimage: ByteVector32, to: Origin, htlc: UpdateAddHtlc) extends ForwardFulfill { override val fulfill  = UpdateFulfillHtlc(htlc.channelId, htlc.id, preimage) }
+  sealed trait ForwardFulfill extends ForwardMessage { val preimage: ByteVector32; val to: Origin; val htlc: UpdateAddHtlc }
+  case class ForwardRemoteFulfill(fulfill: UpdateFulfillHtlc, to: Origin, htlc: UpdateAddHtlc) extends ForwardFulfill { override val preimage = fulfill.paymentPreimage }
+  case class ForwardOnChainFulfill(preimage: ByteVector32, to: Origin, htlc: UpdateAddHtlc) extends ForwardFulfill
   sealed trait ForwardFail extends ForwardMessage { val to: Origin; val htlc: UpdateAddHtlc }
   case class ForwardRemoteFail(fail: UpdateFailHtlc, to: Origin, htlc: UpdateAddHtlc) extends ForwardFail
   case class ForwardRemoteFailMalformed(fail: UpdateFailMalformedHtlc, to: Origin, htlc: UpdateAddHtlc) extends ForwardFail
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -147,50 +147,39 @@ class Relayer(nodeParams: NodeParams, router: ActorRef, register: ActorRef, comm
...
+      case Origin.Relayed(originChannelId, originHtlcId, _, _) =>
+        Metrics.recordPaymentRelayFailed(Tags.FailureType.Remote, Tags.RelayType.Channel)
+        val cmd = ff match {
+          case ForwardRemoteFail(fail, _, _) => CMD_FAIL_HTLC(originHtlcId, Left(fail.reason), commit = true)
+          case ForwardRemoteFailMalformed(fail, _, _) => CMD_FAIL_MALFORMED_HTLC(originHtlcId, fail.onionHash, fail.failureCode, commit = true)
+          case _: ForwardOnChainFail => CMD_FAIL_HTLC(originHtlcId, Right(PermanentChannelFailure), commit = true)
+        }
+        commandBuffer ! CommandBuffer.CommandSend(originChannelId, cmd)
+      case Origin.TrampolineRelayed(_, None) => postRestartCleaner forward ff
+      case Origin.TrampolineRelayed(_, Some(paymentSender)) => paymentSender ! translateFail(ff)
```
pm47:
```suggestion
      case Origin.TrampolineRelayed(_, Some(_)) => nodeRelayer ! translateFail(ff)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -147,50 +147,39 @@ class Relayer(nodeParams: NodeParams, router: ActorRef, register: ActorRef, comm
...
-          context.system.eventStream.publish(ChannelPaymentRelayed(amountIn, amountOut, add.paymentHash, originChannelId, fulfill.channelId))
-        case Origin.TrampolineRelayed(_, None) => postRestartCleaner forward ff
-        case Origin.TrampolineRelayed(_, Some(_)) => nodeRelayer forward ff
-      }
+    case Status.Failure(addFailed: AddHtlcFailed) => addFailed.origin match {
+      case Origin.Local(id, None) => log.error(s"received unexpected add failed with no sender (paymentId=$id)")
+      case Origin.Local(_, Some(sender)) => sender ! Status.Failure(addFailed)
+      case _: Origin.Relayed => channelRelayer forward Status.Failure(addFailed)
+      case Origin.TrampolineRelayed(htlcs, None) => log.error(s"received unexpected add failed with no sender (upstream=${htlcs.mkString(", ")}")
+      case Origin.TrampolineRelayed(_, Some(paymentSender)) => paymentSender ! Status.Failure(addFailed)
```
pm47:
```suggestion
      case Origin.TrampolineRelayed(_, Some(_)) => nodeRelayer ! Status.Failure(addFailed)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
pm47: Ate my own dogfood and implemented it in e042660da. For even more consistency we could also forward the raw `ForwardFail` to payment senders in the failure case, but that would mean slightly more work in their handling. Trade-off...

t-bast: Heh this was exactly my first implementation, but then I changed it because I preferred keeping the payment lifecycle as untouched as possible and containing the change in the relayer (because pay FSM is already doing a lot, and relayer is simpler).
I had the content of `translateFail` inside a handler for `ForwardFail` in the pay FSM.

If you think that's better I can do that, let me know!

pm47: I think it is a better because building a fake `UpdateFulfillHtlc` is bad. It's like a trap for future maintainers. The payment lifecycle is only lightly touched, I had a formatting error in my commit, fixed in a566a27ec.

t-bast: Allright let me amend the last commit then.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1349/commits/7d8020c2a3e772e4158bf3c4ba449e282292ddaf

no need to go deeper

PR #1351, opened 2020-03-20
Support composable behavior on payment initiator

araspitzu:
This PR refactors the PaymentInitiator to support the composition of handlers, inspired by the `ReceiveHandler`, in order to allow plugin developers to register their own handler and provide custom behavior for outgoing payments.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -33,7 +33,7 @@ import fr.acinq.eclair.io.{NodeURI, Peer}
 import fr.acinq.eclair.payment._
 import fr.acinq.eclair.payment.receive.MultiPartHandler.ReceivePayment
 import fr.acinq.eclair.payment.relay.Relayer.{GetOutgoingChannels, OutgoingChannels, UsableBalance}
-import fr.acinq.eclair.payment.send.PaymentInitiator.{SendPaymentRequest, SendPaymentToRouteRequest, SendPaymentToRouteResponse}
+import fr.acinq.eclair.payment.send.PaymentInitiatorHandler.{SendPaymentRequest, SendPaymentToRouteRequest, SendPaymentToRouteResponse}
 import fr.acinq.eclair.router._
 import fr.acinq.eclair.wire.{ChannelAnnouncement, ChannelUpdate, NodeAddress, NodeAnnouncement}
 import scodec.bits.ByteVector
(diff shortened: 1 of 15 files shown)


t-bast:
Can you describe the motivation behind the change in more details?
What kind of concrete scenario become possible that weren't possible before?

araspitzu:
> Can you describe the motivation behind the change in more details?
> What kind of concrete scenario become possible that weren't possible before?

The motivation is that in eclair master is not possible for plugin developers to plug into the outgoing payment mechanism, the proposed changes introduce a composable behavior to be "stacked"  in the payment initiator making it possible for plugins to control it. 
This opens up some new scenarios: writing a custom `PaymentInitiatorHandler` it's possible to use your own router for route calculation (perhaps using a different path-finding algo) and it's possible to craft specific onion packets for outgoing payments, specifically this allows for spontaneous payments and to expose custom TLV records (i.e using HTLCs for message transport) to the API.

t-bast:
TBH I'm not convinced by the motivation. This makes the code more complex and payment initiation is already quite complex because of the many layers it goes through (`PaymentInitiator -> MultiPartLifecycle -> PayLifecycle`). I'm not saying we shouldn't do this change, but I think it deserves a better motivation to justify it (don't hesitate to provide more scenario to convince me).

> use your own router for route calculation (perhaps using a different path-finding algo)

If someone has an interesting path-finding algorithm that's different from ours, I think it would be better to encourage them to integrate it directly into our codebase. We could then let users choose what algorithm to use via config. It feels too involved honestly to change something like this in a plugin (and the plugin would probably be broken after each of our releases because of internal changes). If people simply want to experiment, then they can do their calculations completely outside of `eclair` and use `sendtoroute`. 

> craft specific onion packets

I think a better way to allow that is to allow users to provide these custom fields as input, and we just carry them around all the way to the destination. That's something we should support at some point, I don't think it needs such a drastic change on the `PaymentInitiator`.

araspitzu:
The motivations are indeed broader than the few that i wrote, it's meant to give full control of the outgoing payment lifecycle to the plugin developer. I used a structure similar to the `PaymentHandler` to not introduce an even bigger change, in fact the proposed changes modify very little of the `PaymentInitiator` (it's mostly using the `ActorContext` to provide a sender actor). It's true that a theoretical plugin would need to be updated at each release but that is true for the `PaymentHandler` as well which is designed with a pluggable architecture in mind. My personal motivation was to make a plugin for spontaneous payments, I like your suggestion of providing custom fields as input and to be honest i've left that as a plan-B if this PR was going to be rejected :).

t-bast:
> The motivations are indeed broader than the few that i wrote, it's meant to give full control of the outgoing payment lifecycle to the plugin developer.

If we don't have enough concrete motivating examples for that, I believe we shouldn't do it. Why add more code and complexity to maintain if we aren't going to need it?
I really don't think we should give "full control of the outgoing payment lifecycle" to plugin developers, that feels really dangerous (and nobody has requested that yet)...

Your current use-case calls for a much simpler solution, so I'd go in that direction instead :)
We should have a look at what c-lightning and lnd do, but I think they have something very simple where you simply provide these additional fields already encoded (binary blob) and the core includes it in the onion tlvs.

araspitzu:
Alright, let's go for the simpler solution then :)

PR #1352, opened 2020-03-20
Move router handlers to separate files

pm47:
This build on #1347.

The goal is to simplify the router by moving handlers to dedicated stateless functions in separated files, for syncing, announcement validation, route calculation, etc. That's quite a big change, but it's very repetitive.

Also, all gossip messages are now acknowledged by the router with a `GossipDecision` reply.


eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculationHandlers.scala
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2019 ACINQ SAS
```
dpad85: Much nitpick: you can use a dynamic year in Intellij.

pm47: Doen with e646c4c .

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculationHandlers.scala
```diff
@@ -0,0 +1,68 @@
...
+    val assistedChannels: Map[ShortChannelId, AssistedChannel] = assistedRoutes.flatMap(toAssistedChannels(_, target, amount)).toMap
+    val extraEdges = assistedChannels.values.map(ac => GraphEdge(ChannelDesc(ac.extraHop.shortChannelId, ac.extraHop.nodeId, ac.nextNodeId), toFakeUpdate(ac.extraHop, ac.htlcMaximum))).toSet
+    val ignoredEdges = ignoreChannels ++ d.excludedChannels
+    val defaultRouteParams: RouteParams = getDefaultRouteParams(routerConf)
+    val params = routeParams.getOrElse(defaultRouteParams)
+    val routesToFind = if (params.randomize) DEFAULT_ROUTES_COUNT else 1
+
+    log.info(s"finding a route $source->$target with assistedChannels={} ignoreNodes={} ignoreChannels={} excludedChannels={}", assistedChannels.keys.mkString(","), ignoreNodes.map(_.value).mkString(","), ignoreChannels.mkString(","), d.excludedChannels.mkString(","))
+    log.info(s"finding a route with randomize={} params={}", routesToFind > 1, params)
+    findRoute(d.graph, source, target, amount, numRoutes = routesToFind, extraEdges = extraEdges, ignoredEdges = ignoredEdges, ignoredVertices = ignoreNodes, routeParams = params, currentBlockHeight)
```
araspitzu, line 62: I think it would be great if `findRoute` wouldn't throw an exception when the route is not found, we could have a simple hierarchy like:
```
sealed trait FindRouteResult
case class RouteFound(route: Seq[ChannelHop]) extends FindRouteResult
case object RouteNotFound extends FindRouteResult // or case class RouteNotFound(t: Throwable)
```

t-bast: I think this should be in a separate PR though (and I'll be working a lot on the path-finding soon, so I'll likely be doing that).
The current PR only moves things around (which is already a lot of changes), I think it's better to avoid mixing it with behavior changes.

eclair-core/src/main/scala/fr/acinq/eclair/router/SyncHandlers.scala
```diff
@@ -0,0 +1,203 @@
...
+          )
+        }
+
+        // we update our sync data to this node (there may be multiple channel range responses and we can only query one set of ids at a time)
+        val replies = shortChannelIdAndFlags
+          .grouped(routerConf.channelQueryChunkSize)
+          .map(buildQuery)
+          .toList
+
+        val (sync1, replynow_opt) = addToSync(d.sync, origin.nodeId, replies)
```
t-bast, line 132: Shouldn't we also move `addToSync` inside this file?

eclair-core/src/main/scala/fr/acinq/eclair/router/SyncHandlers.scala
```diff
@@ -0,0 +1,203 @@
...
+
+    Kamon.runWithContextEntry(remoteNodeIdKey, origin.nodeId.toString) {
+      Kamon.runWithSpan(Kamon.spanBuilder("reply-channel-range").start(), finishSpan = true) {
+
+        @tailrec
+        def loop(ids: List[ShortChannelId], timestamps: List[ReplyChannelRangeTlv.Timestamps], checksums: List[ReplyChannelRangeTlv.Checksums], acc: List[ShortChannelIdAndFlag] = List.empty[ShortChannelIdAndFlag]): List[ShortChannelIdAndFlag] = {
+          ids match {
+            case Nil => acc.reverse
+            case head :: tail =>
+              val flag = computeFlag(d.channels)(head, timestamps.headOption, checksums.headOption, routerConf.requestNodeAnnouncements)
```
t-bast, line 92: We should move `computeFlag` inside this file as well.

eclair-core/src/main/scala/fr/acinq/eclair/router/SyncHandlers.scala
```diff
@@ -0,0 +1,203 @@
...
+        // keep channel ids that are in [firstBlockNum, firstBlockNum + numberOfBlocks]
+        val shortChannelIds: SortedSet[ShortChannelId] = channels.keySet.filter(keep(q.firstBlockNum, q.numberOfBlocks, _))
+        log.info("replying with {} items for range=({}, {})", shortChannelIds.size, q.firstBlockNum, q.numberOfBlocks)
+        val chunks = Kamon.runWithSpan(Kamon.spanBuilder("split-channel-ids").start(), finishSpan = true) {
+          split(shortChannelIds, q.firstBlockNum, q.numberOfBlocks, routerConf.channelRangeChunkSize)
+        }
+
+        Kamon.runWithSpan(Kamon.spanBuilder("compute-timestamps-checksums").start(), finishSpan = true) {
+          chunks.foreach { chunk =>
+            val reply = Router.buildReplyChannelRange(chunk, q.chainHash, routerConf.encodingType, q.queryFlags_opt, channels)
```
t-bast, line 72: buildReplyChannelRange should be in this file too.

eclair-core/src/main/scala/fr/acinq/eclair/router/ValidationHandlers.scala
```diff
@@ -0,0 +1,442 @@
...
+            // channel isn't announced and we never heard of it (maybe it is a private channel or maybe it is a public channel that doesn't yet have 6 confirmations)
+            // let's create a corresponding private channel and process the channel_update
+            log.debug("adding unannounced local channel to remote={} shortChannelId={}", lcu.remoteNodeId, shortChannelId)
+            val d1 = d.copy(privateChannels = d.privateChannels + (shortChannelId -> PrivateChannel(localNodeId, lcu.remoteNodeId, None, None)))
+            handleChannelUpdate(d1, db, routerConf, Set(LocalGossip), u)
+        }
+    }
+  }
+
+  def handleLocalChannelDown(d: Data, localNodeId: PublicKey, lcd: LocalChannelDown)(implicit ctx: ActorContext, log: LoggingAdapter): Data = {
```
t-bast, line 419: This one doesn't need `ActorContext`

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -165,7 +162,7 @@ case class LiftChannelExclusion(desc: ChannelDesc)
 // @formatter:on
 
 // @formatter:off
-case class SendChannelQuery(remoteNodeId: PublicKey, to: ActorRef, flags_opt: Option[QueryChannelRangeTlv])
+case class SendChannelQuery(chainHash: ByteVector32, remoteNodeId: PublicKey, to: ActorRef, flags_opt: Option[QueryChannelRangeTlv])
```
t-bast, line 165: I think it would be nice to move all these objects that are currently defined directly in the router namespace to the objects they really belong to

t-bast: And there are also some helper functions that are still in the `Router` object that can move to the handler file they're used in.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculationHandlers.scala
pm47: Agree, let's do this in another PR.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculationHandlers.scala
```diff
@@ -0,0 +1,68 @@
...
+
+package fr.acinq.eclair.router
+
+import akka.actor.{ActorContext, ActorRef, Status}
+import akka.event.LoggingAdapter
+import fr.acinq.eclair.{ShortChannelId, _}
+import fr.acinq.eclair.router.Graph.GraphStructure.{DirectedGraph, GraphEdge}
+import fr.acinq.eclair.router.Router._
+
+object RouteCalculationHandlers {
```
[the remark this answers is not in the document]
pm47: Done with dc79e8d.

eclair-core/src/main/scala/fr/acinq/eclair/router/SyncHandlers.scala
pm47: Done with dc79e8d.

eclair-core/src/main/scala/fr/acinq/eclair/router/SyncHandlers.scala
pm47: Done with dc79e8d.

eclair-core/src/main/scala/fr/acinq/eclair/router/SyncHandlers.scala
pm47: Done with dc79e8d.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculationHandlers.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  def finalizeRoute(d: Data, fr: FinalizeRoute)(implicit ctx: ActorContext, log: LoggingAdapter): Data = {
+    implicit val sender: ActorRef = ctx.self // necessary to preserve origin when sending messages to other actors
+    import fr.{assistedRoutes, hops => partialHops}
```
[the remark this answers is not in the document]
pm47: Done with ad47ec4.

eclair-core/src/test/scala/fr/acinq/eclair/router/BaseRouterSpec.scala
```diff
@@ -100,29 +100,30 @@ abstract class BaseRouterSpec extends TestkitBaseClass {
 
 
       // let's we set up the router
```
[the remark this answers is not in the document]
pm47: Done with ad47ec4.

pm47:
I've moved all static methods to respective handlers file, and also took the opportunity to move router-related types to the `Router` object.

As this touches a lot of files and may generate conflicts, I suggest we merge this rather sooner than later.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -351,6 +352,7 @@ class PeerConnection(nodeParams: NodeParams, switchboard: ActorRef, router: Acto
               setTimer(ResumeAnnouncements.toString, ResumeAnnouncements, IGNORE_NETWORK_ANNOUNCEMENTS_PERIOD, repeat = false)
               d.behavior.copy(fundingTxAlreadySpentCount = d.behavior.fundingTxAlreadySpentCount + 1, ignoreNetworkAnnouncement = true)
             }
+          case _ => d.behavior // other rejections are not considered punishable offenses
```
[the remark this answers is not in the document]
pm47: But we're not doing the exact same action. Another way would be to just replace the `case _ =>` catch-all by more precise matchers.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculationHandlers.scala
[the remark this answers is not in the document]
pm47: True, but that wouldn't be consistent with the rest of the handlers. I can split this one in two and have it call a pure function, WDYT?

eclair-core/src/main/scala/fr/acinq/eclair/router/ValidationHandlers.scala
pm47: Fixed with fe2294f.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Done with ee49982.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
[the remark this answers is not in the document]
t-bast: But couldn't this catch-all be expressed by a specific trait defined in the router?

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculationHandlers.scala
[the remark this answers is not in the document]
t-bast: I don't think this is worth it then. We'll see in future refactorings.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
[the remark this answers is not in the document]
pm47: Only if the trait meant "should *not* lead to a connection close.

eclair-core/src/main/scala/fr/acinq/eclair/router/ValidationHandlers.scala
```diff
@@ -0,0 +1,448 @@
...
+import fr.acinq.eclair.router.Monitoring.Metrics
+import fr.acinq.eclair.router.Router._
+import fr.acinq.eclair.transactions.Scripts
+import fr.acinq.eclair.wire._
+import fr.acinq.eclair.{Logs, NodeParams, ShortChannelId, TxCoordinates}
+import kamon.Kamon
+
+object ValidationHandlers {
+
+  def sendDecision(peerConnection: ActorRef, decision: GossipDecision): Unit = {
```
t-bast, line 36: Doesn't this one need an implicit actor context and to define the `sender`?

pm47: Great catch. Actually this should be enough:
```suggestion
  def sendDecision(peerConnection: ActorRef, decision: GossipDecision)(implicit sender: ActorRef): Unit = {
```

t-bast: Maybe we should test this. An easy way to test this would be to add a check to existing tests that the `sender` they see if the router?

pm47: I did, with 5ca43a6, and found another bug 🙃 . I should test all the rejection cases but can't do it rn.

t-bast: Sounds good, I like the current state of the PR.
Once we add some more tests it should be good to go from my point of view.

eclair-core/src/main/scala/fr/acinq/eclair/router/Monitoring.scala
```diff
@@ -16,14 +16,27 @@
...
 
 object Monitoring {
 
   object Metrics {
     val FindRouteDuration = Kamon.timer("router.find-route.duration", "Path-finding duration")
     val RouteLength = Kamon.histogram("router.find-route.length", "Path-finding result length")
+
+    private val GossipResult = Kamon.counter("router.gossip.result")
+
+    // @formatter:off
```
t-bast, line 32: nit: why the formatter:off (with weird formatting)?
What about:
```scala
def gossipResult(decision: GossipDecision): Counter = decision match {
  case _: GossipDecision.Accepted => GossipResult.withTag("result", "accepted")
  case rejected: GossipDecision.Rejected => GossipResult.withTag("result", "rejected").withTag("reason", getSimpleClassName(rejected))
}
```

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -16,12 +16,15 @@
 
 package fr.acinq.eclair.router
 
+
```
t-bast, line 19: supernit: empty line

eclair-core/src/main/scala/fr/acinq/eclair/router/ValidationHandlers.scala
pm47: Phew, done with c738659. No new bug found.

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
pm47: Done with fa23cc0.

eclair-core/src/main/scala/fr/acinq/eclair/router/Monitoring.scala
pm47: Done with fa23cc0.

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -185,4 +185,9 @@ package object eclair {
     // @formatter:on
   }
 
+  /**
+   * Apparently .getClass.getSimpleName can crashes java 8 with a "Malformed class name" error
```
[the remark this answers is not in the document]
pm47: Done with fa23cc0.

pm47:
Rebase, waiting for review from @sstone.

t-bast:
I completely agree with @sstone, that was exactly my reaction too :)
It's a good first step towards that direction, and doing everything in one step would have been too much.

PR #1355, opened 2020-03-25
Add wallet balance metric

t-bast:
Add a metric to track wallet balance.
Add a few channel metrics.
Tweak RPC metrics.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
@@ -28,6 +28,7 @@ import fr.acinq.eclair.blockchain._
 import fr.acinq.eclair.blockchain.bitcoind.rpc.ExtendedBitcoinClient
 import fr.acinq.eclair.channel.BITCOIN_PARENT_TX_CONFIRMED
 import fr.acinq.eclair.transactions.Scripts
+import org.json4s.JsonAST.JDecimal
 import scodec.bits.ByteVector
 
 import scala.collection.{Set, SortedMap}
@@ -84,6 +85,9 @@ class ZmqWatcher(blockCount: AtomicLong, client: ExtendedBitcoinClient)(implicit
           log.debug("setting blockCount={}", count)
           blockCount.set(count)
           context.system.eventStream.publish(CurrentBlockCount(count))
+      }
+      client.rpcClient.invoke("getbalance").collect {
+        case JDecimal(balance) => Metrics.BitcoinBalance.withoutTags().update(balance.doubleValue() * 1000)
       }
       // TODO: beware of the herd effect
       KamonExt.timeFuture(Metrics.NewBlockCheckConfirmedDuration.withoutTags()) {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
@@ -87,6 +87,7 @@ class ZmqWatcher(blockCount: AtomicLong, client: ExtendedBitcoinClient)(implicit
           context.system.eventStream.publish(CurrentBlockCount(count))
       }
       client.rpcClient.invoke("getbalance").collect {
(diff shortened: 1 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -85,6 +86,9 @@ class ZmqWatcher(blockCount: AtomicLong, client: ExtendedBitcoinClient)(implicit
           blockCount.set(count)
           context.system.eventStream.publish(CurrentBlockCount(count))
       }
+      client.rpcClient.invoke("getbalance").collect {
+        case JDecimal(balance) => Metrics.BitcoinBalance.withoutTags().update(balance.doubleValue() * 1000)
```
pm47: Why do you do `* 1000` ?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Monitoring.scala
```diff
@@ -21,6 +21,7 @@ import kamon.Kamon
 object Monitoring {
 
   object Metrics {
+    val ChannelCount = Kamon.gauge("channels.count")
```
pm47: Nit: `ChannelsCount` or `ChannelCount`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Monitoring.scala
```diff
@@ -29,13 +30,19 @@ object Monitoring {
     val Event = "event"
     val Fatal = "fatal"
     val Origin = "origin"
+    val Status = "status"
```
pm47:
```suggestion
    val State = "state"
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Monitoring.scala
```diff
@@ -29,13 +30,19 @@ object Monitoring {
...
 
     object Events {
       val Created = "created"
       val Closing = "closing"
       val Closed = "closed"
     }
 
+    object Statuses {
```
pm47: Isn't that duplicated with `Events`? I think `States` is the proper name.

```suggestion
    object States {
```

pm47:
Also, this needs a rebase

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: Because Kamon does a `floor`, so if your balance is `4.99 BTC` it will show you only `4 BTC`.
A balance difference of `0.99 BTC` is a big change that we don't want to miss :)

It's worth a comment though, and maybe I can rename the metric to `MilliBitcoinBalance` or something like that?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Monitoring.scala
t-bast: I'll align with peer (and potentially update the metrics that don't follow the same pattern).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Monitoring.scala
t-bast: > Isn't that duplicated with Events?

It's somewhat but not completely duplicated, as `events` don't exactly map to channel states (since `created` isn't really a channel state). I think we'll need to keep both

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
pm47: Maybe

```scala
val BitcoinBalance = Kamon.gauge("bitcoin.balance", "Bitcoin balance (mBTC)")
```

t-bast: Done, it's a pity Kamon doesn't let us define our own measurement units (it only supports standard units like time units, file size units and number of instances of something).

PR #736, opened 2018-10-17
Support for Tor onion services

rorp:

PR #1356, opened 2020-03-25
fixup! Support for Tor onion services (#736)

pm47:


pm47:
> It's a bit worrisome that we're lacking tests and didn't catch that earlier

We're indeed missing automated tests with an actual tor instance, there is only a manual test. On the bright side, the impact was only in logging.

PR #1357, opened 2020-03-25
 fixup! Split the `Peer` in two (#1347) 

pm47:


eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
```diff
@@ -21,8 +21,10 @@ import kamon.Kamon
 object Monitoring {
 
   object Metrics {
-    val PeerConnections = Kamon.counter("peers.connecting.count")
-    val ConnectedPeers = Kamon.gauge("peers.connected").withoutTags()
+    val PeerConnectionConnecting = Kamon.counter("peerconnection.connecting")
+    val PeerConnectionConnected = Kamon.gauge("peerconnection.connected")
+
+    val PeerConnected = Kamon.gauge("peer.connected")
```
t-bast: Why did you change the name from `peers.connected` to `peer.connected`? Plural feels better here IMO

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
```diff
@@ -21,8 +21,10 @@ import kamon.Kamon
 object Monitoring {
 
   object Metrics {
-    val PeerConnections = Kamon.counter("peers.connecting.count")
-    val ConnectedPeers = Kamon.gauge("peers.connected").withoutTags()
+    val PeerConnectionConnecting = Kamon.counter("peerconnection.connecting")
+    val PeerConnectionConnected = Kamon.gauge("peerconnection.connected")
```
t-bast: Are you using this one to monitor for discrepancies with `PeerConnected`?

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
```diff
@@ -21,8 +21,10 @@ import kamon.Kamon
 object Monitoring {
 
   object Metrics {
-    val PeerConnections = Kamon.counter("peers.connecting.count")
-    val ConnectedPeers = Kamon.gauge("peers.connected").withoutTags()
+    val PeerConnectionConnecting = Kamon.counter("peerconnection.connecting")
```
t-bast: I wouldn't not explicitly mention `peerconnection` instead of `peer` here, it's quite an implementation detail that people looking at the monitoring graphs don't even need to know about. This is really a peer connecting, so `peers.connecting.count` seemed better IMHO.

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
```diff
@@ -21,8 +21,10 @@ import kamon.Kamon
 object Monitoring {
 
   object Metrics {
```
t-bast: We could probably also instrument the reconnection tasks? Maybe a counter for successful/failed reconnections?

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -456,7 +467,7 @@ object PeerConnection {
   sealed trait Data
   sealed trait HasTransport extends Data { def transport: ActorRef }
   case object Nothing extends Data
-  case class AuthenticatingData(pendingAuth: PendingAuth, transport: ActorRef) extends Data
+  case class AuthenticatingData(pendingAuth: PendingAuth, transport: ActorRef) extends Data with HasTransport
```
t-bast: nit: `HasTransport` already extends `Data`, so you don't need `extends Data with HasTransport` (and same below).

eclair-core/src/main/scala/fr/acinq/eclair/io/Server.scala
```diff
@@ -55,7 +55,6 @@ class Server(nodeParams: NodeParams, switchboard: ActorRef, router: ActorRef, ad
   def listening(listener: ActorRef): Receive = {
     case Connected(remote, _) =>
       log.info(s"connected to $remote")
-      Metrics.PeerConnections.withTag(Tags.ConnectionState, Tags.ConnectionStates.Connected).increment()
```
t-bast: Good catch!

sstone: nit: the `import` director for Metrics can be removed now

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
sstone: Yes but this one provides information about the actual connection state (authenticating/authenticated/...)  and is "richer" than the peer metrics which is just off/on.

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
sstone: Agreed, what is the point of this specific gauge ?

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
t-bast: Yes but it's mostly about the naming, I would have just named this `peers.connecting.count`. The ops looking at the graph doesn't necessarily need to know that `peer` and `peerconnection` are different things. But it's probably nit.

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
pm47: > Plural feels better here IMO

Agree, it was to maintain consistency with `PeerConnectionConnected`.

Another idea I had was to use inner objects:

```scala
object PeerConnection {
      val Connecting = Kamon.counter("peerconnection.connecting")
      val Connected = Kamon.gauge("peerconnection.connected")
    }

    object Peer {
      val Connected = Kamon.gauge("peer.connected")
    }
```

But I refrained from doing that because you didn't use that pattern for other metrics.

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
pm47: Our code guarantees that we have one single instance of `Peer` for a given `remoteNodeId`, but we can have several `PeerConnection` to the same `remoteNodeId` even if that should be a transient state. This is particularly worh monitoring in a distributed deployment.

Does that make sense?

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
pm47: @t-bast The problem with your suggestion is that we also need to track the state "connected" in the `PeerConnection` as I explained https://github.com/ACINQ/eclair/pull/1357#discussion_r398743209.

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
pm47: Oh yeah

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
pm47: I know, that's because I have used the cake pattern too much.

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
t-bast: That's only an argument for keeping the peerconnection gauge, not really for the naming of this metric. But anyway this is nit, not worth wasting time bikeshedding.

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
t-bast: Agreed, it's worth verifying that we don't constantly have much more `peerconnection.connected` than `peer.connected`

eclair-core/src/main/scala/fr/acinq/eclair/io/Monitoring.scala
t-bast: I think the name of the objects are fine, it's just the name of the metric (`peerconnection.*` and `peer.*`) where I would have used plural (`peerconnections.*` and `peers.*`). Not a big issue though.

pm47:
What do you guys think of removing the `Authenticated`/`InitializeConnection` between the `PeerConnection` and the `Switchboard`?

Meaning going from:

```
 PeerConnection        Switchboard              Peer
       |                    |                    |
       |   Authenticated    |                    |
       |------------------->|                    |
       |                    |                    |
       |   InitializeConn   |                    |
       |<-------------------|                    |
       |                    |                    |
       |                                         |
       |           LightningMessage              |
       |---------------------------------------->|
       |---------------------------------------->|
       |---------------------------------------->|
       |                                         |
       |                                         |
```

to:

```
 PeerConnection        Switchboard              Peer
       |                    |                    |
       |  LightningMessage  |                    |
       |------------------->|  LightningMessage  |
       |                    |------------------->|
       |  LightningMessage  |                    |
       |------------------->|  LightningMessage  |
       |                    |------------------->|
       |  LightningMessage  |                    |
       |------------------->|  LightningMessage  |
       |                    |------------------->|
```

Tradeoff being that the flow is simpler, but all messages have to go through the `Switchboard`, meaning that they also need to be encapsulated in an envelope message containing the `remoteNodeId`.

t-bast:
I prefer how it is now. It's really not much code and avoid too much forwarding and encapsulating.
The switchboard acts as a manager during the authentication and then gets out of the way, I think this is its job.

t-bast:
Apart from that, I'm in favor on merging this PR once the `HasTransport` thing is settled to get this fixup in eclair-0.3.4!

pm47:
> I prefer how it is now. It's really not much code and avoid too much forwarding and encapsulating.
The switchboard acts as a manager during the authentication and then gets out of the way, I think this is its job.

Gave it a try with 27c7e681b.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
pm47: Done with df54c287d

t-bast: What does that cake pattern bring here (apart from being less explicit than `sealed trait HasTransport extends Data`)?

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -69,10 +69,15 @@ class Switchboard(nodeParams: NodeParams, router: ActorRef, watcher: ActorRef, r
...
-      // if this is an incoming connection, we might not yet have created the peer
-      val peer = createOrGetPeer(authenticated.remoteNodeId, offlineChannels = Set.empty)
-      authenticated.peerConnection ! PeerConnection.InitializeConnection(peer)
+    case p: PeerConnection.ConnectionReady =>
+      // we create a peer if it doesn't exist
+      val peer = createOrGetPeer(p.remoteNodeId, offlineChannels = Set.empty)
+      peer forward p
+
+    case p: Peer.PeerMessage =>
+      // we create a peer if it doesn't exist
```
t-bast: We should never have to create one here, right?
Are there unwanted side-effects if we do (some assumption each of the FSMs make about the other FSMs)?
Shouldn't we reject those messages if we don't have a peer actor to handle it?

It feels to me that the peer FSM already have to know about the other FSMs and assume some of their behavior, and this change is making it slightly worse...

pm47: > We should never have to create one here, right?

Correct, this is only needed in the handler for `PeerConnection.ConnectionReady`.

> It feels to me that the peer FSMs already have to know about the other FSMs and assume some of their behavior, and this change is making it slightly worse...

There has to be an interface contract, even if it less explicit with (untyped) actors. If anything the interface is a bit simpler now. But let's revert this and we can revisit later.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
pm47: Not much: it conveys the idea that `HasTransport` is just a behavior and not a definition, which is closer to how I see it.

t-bast: Got it, I'm not very familiar with that pattern

pm47: In other words you have to use `Data with HasTransport`, you can't use `HasTransport` by itself.

PR #1360, opened 2020-03-31
 fixup! Use CLTV as tie-breaker for offered htlc output sorting (#790) 

pm47:
This is a follow-up to #790 with more refactoring.


eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -62,54 +61,62 @@ object CommitmentSpec {
...
+      case IncomingHtlc(add) => spec.copy(toRemote = spec.toRemote - add.amountMsat, htlcs = spec.htlcs + directedHtlc)
     }
   }
 
   // OUT means we are sending an UpdateFulfillHtlc message which means that we are fulfilling an HTLC that they sent
-  def fulfillHtlc(spec: CommitmentSpec, direction: Direction, htlcId: Long): CommitmentSpec = {
-    spec.findHtlcById(htlcId, direction.opposite) match {
-      case Some(htlc) if direction == OUT => spec.copy(toLocal = spec.toLocal + htlc.add.amountMsat, htlcs = spec.htlcs - htlc)
-      case Some(htlc) if direction == IN => spec.copy(toRemote = spec.toRemote + htlc.add.amountMsat, htlcs = spec.htlcs - htlc)
+  def fulfillOutgoingHtlc(spec: CommitmentSpec, htlcId: Long): CommitmentSpec = {
```
t-bast: I think the naming becomes counter-intuitive with this refactoring.
It made sense to say `fulfillHtlc(spec, OUT, id)` to mean that you were the one sending the fulfill (and `fulfillHtlc(spec, IN, id)` when you were the one receiving the fulfill).

Now it's really weird because we're saying `fulfillOutgoingHtlc` for an incoming HTLC, and `fulfillIncomingHtlc` for an outgoing HTLC...I would reverse that: `fulfillOutgoingHtlc` should fulfill an HTLC you sent, and `fulfillIncomingHtlc` should fulfill one you received.

It's the same for the `failXXX` functions below.

araspitzu:
I wonder if we should make a type alias or case class for `Set[DirectedHtlc]`. Since we often run filter/collect functions on it we could simplify the code by exposing our own `filterIncoming` and `filterOutgoing` but maybe it's an overkill.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -720,7 +720,7 @@ class NormalStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     // bob replies immediately with a signature
     bob2alice.expectMsgType[CommitSig]
 
-    awaitCond(bob.stateData.asInstanceOf[DATA_NORMAL].commitments.localCommit.spec.htlcs.exists(h => h.add.id == htlc.id && h.direction == IN))
+    awaitCond(bob.stateData.asInstanceOf[DATA_NORMAL].commitments.localCommit.spec.htlcs.exists { case IncomingHtlc(add) => add.id == htlc.id; case _ => false })
```
t-bast: nit: not a big fan of the one-liner pattern matching

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
```diff
@@ -150,6 +145,37 @@ class ChannelCodecsSpec extends FunSuite {
...
+        ByteVector32(hex"dac3bc8b2e16fdf2db3e627483bc395c701c1fc96ace53e4ebc54150e807921d"))
+    )
+    val remaining = bin"0000000"
```
t-bast: ??

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -38,22 +32,27 @@ object CommitmentOutput {
...
-  def direction: Direction
   val add: UpdateAddHtlc
   def opposite: DirectedHtlc = this match {
     case IncomingHtlc(_) => OutgoingHtlc(add)
     case OutgoingHtlc(_) => IncomingHtlc(add)
   }
+  def direction: String = this match {
+    case IncomingHtlc(_) => "IN"
+    case OutgoingHtlc(_) => "OUT"
+  }
```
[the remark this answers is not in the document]
pm47: Or:

```scala
def incoming: PartialFunction[DirectedHtlc, UpdateAddHtlc] = { case h: IncomingHtlc => h.add }
def outgoing: PartialFunction[DirectedHtlc, UpdateAddHtlc] = { case h: OutgoingHtlc => h.add }

val outgoingHtlcsBefore = reduced.htlcs.collect { case IncomingHtlc(add) => add  }
val outgoingHtlcsAfter = reduced.htlcs.collect(incoming)
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -835,7 +835,7 @@ object Helpers {
         // if an outgoing htlc is in the remote commitment, then:
         // - either it is in the local commitment (it was never fulfilled)
         // - or we have already received the fulfill and forwarded it upstream
-        val outgoingHtlcs = localCommit.spec.htlcs.filter(_.direction == OUT).map(_.add)
+        val outgoingHtlcs = localCommit.spec.htlcs.collect { case OutgoingHtlc(add) => add }
```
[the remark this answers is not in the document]
pm47: See also https://github.com/ACINQ/eclair/pull/1360#discussion_r400828790

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
pm47: Agree, this file is a mess, surely we can do better.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
pm47: See the commit message, we need to test the remainder of a codec.

Here, because we use the 1-bit `bool` codec to indicate the direction and our input data is aligned on bytes, we will end up with a 7-bits remainder.

t-bast: Got it, probably worth a comment in the code, right above this `remaining`?

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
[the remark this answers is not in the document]
araspitzu: This looks nice and it's readable even when composed with more `filter` functions, ACK for me.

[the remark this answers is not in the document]
t-bast: I like it, but I think it's also worth having `def incomingHtlcs(): Seq[UpdateAddHtlc]`, there are so many places where we would just end up doing `spec.htlcs.collect(incoming).map(_.add)` whereas we could do `spec.incomingHtlcs`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -835,10 +836,10 @@ object Helpers {
         // if an outgoing htlc is in the remote commitment, then:
         // - either it is in the local commitment (it was never fulfilled)
         // - or we have already received the fulfill and forwarded it upstream
-        val outgoingHtlcs = localCommit.spec.htlcs.filter(_.direction == OUT).map(_.add)
-        outgoingHtlcs.collect {
-          case add if add.paymentHash == sha256(paymentPreimage) => (add, paymentPreimage)
-        }
+        localCommit.spec.htlcs
+          .collect {
```
t-bast: very nit: I'd put `.collect` on the above line

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
pm47: Fixed with 988d2ac.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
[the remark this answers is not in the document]
pm47: Fixed with 988d2ac.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -864,14 +865,14 @@ object Helpers {
...
-        (localCommit.spec.htlcs.filter(_.direction == OUT) -- Transactions.trimOfferedHtlcs(localDustLimit, localCommit.spec)).map(_.add)
+        (localCommit.spec.htlcs.collect(outgoing) -- Transactions.trimOfferedHtlcs(localDustLimit, localCommit.spec).map(_.add))
       } else {
         // maybe this is a timeout tx, in that case we can resolve and fail the corresponding htlc
         tx.txIn.map(_.witness match {
           case ScriptWitness(Seq(ByteVector.empty, remoteSig, localSig, ByteVector.empty, htlcOfferedScript)) =>
             val paymentHash160 = htlcOfferedScript.slice(109, 109 + 20)
             log.info(s"extracted paymentHash160=$paymentHash160 from tx=$tx (htlc-timeout)")
-            localCommit.spec.htlcs.filter(_.direction == OUT).map(_.add).filter(add => ripemd160(add.paymentHash) == paymentHash160)
+            localCommit.spec.htlcs.collect { case OutgoingHtlc(add) if ripemd160(add.paymentHash) == paymentHash160 => add }
```
t-bast: nit: could be `.collect(outgoing).filter(add => ripemd160(add.paymentHash) == paymentHash160)` to look like we did in other places (or not if you prefer this)

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
pm47: Done with 7809e58.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: Done with fa2a134.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: Done with fa2a134.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -80,19 +80,19 @@ class PostRestartHtlcCleanerSpec extends TestkitBaseClass {
...
-      buildHtlc(0, channelId_ab_2, randomBytes32, IN), // not relayed
-      buildHtlc(1, channelId_ab_2, randomBytes32, OUT),
-      buildHtlc(2, channelId_ab_2, trampolineRelayedPaymentHash, IN),
-      buildHtlc(3, channelId_ab_2, randomBytes32, OUT),
-      buildHtlc(4, channelId_ab_2, randomBytes32, IN) // not relayed
+      buildHtlcIn(0, channelId_ab_2, randomBytes32), // not relayed
+      buildHtlcOut(1, channelId_ab_2, randomBytes32),
+      buildHtlcIn(2, channelId_ab_2, trampolineRelayedPaymentHash),
+      buildHtlcOut(3, channelId_ab_2, randomBytes32),
+      buildHtlcIn(4, channelId_ab_2, randomBytes32) // not relayed
```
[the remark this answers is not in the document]
pm47: Done with 4b1a70b.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
```diff
@@ -162,7 +162,7 @@ class ChannelCodecsSpec extends FunSuite {
...
         ByteVector32(hex"dac3bc8b2e16fdf2db3e627483bc395c701c1fc96ace53e4ebc54150e807921d"))
     )
-    val remaining = bin"0000000"
+    val remaining = bin"0000000" // 7 bits remainder because the direction is encoded en 1 bit and we are dealing with bytes
```
t-bast:
```suggestion
    val remaining = bin"0000000" // 7 bits remainder because the direction is encoded with 1 bit and we are dealing with bytes
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -886,14 +886,14 @@ object Helpers {
...
-        (remoteCommit.spec.htlcs.filter(_.direction == IN) -- Transactions.trimReceivedHtlcs(remoteDustLimit, remoteCommit.spec)).map(_.add)
+        (remoteCommit.spec.htlcs.collect(incoming) -- Transactions.trimReceivedHtlcs(remoteDustLimit, remoteCommit.spec).map(_.add))
       } else {
         // maybe this is a timeout tx, in that case we can resolve and fail the corresponding htlc
         tx.txIn.map(_.witness match {
           case ScriptWitness(Seq(remoteSig, ByteVector.empty, htlcReceivedScript)) =>
             val paymentHash160 = htlcReceivedScript.slice(69, 69 + 20)
             log.info(s"extracted paymentHash160=$paymentHash160 from tx=$tx (claim-htlc-timeout)")
-            remoteCommit.spec.htlcs.filter(_.direction == IN).map(_.add).filter(add => ripemd160(add.paymentHash) == paymentHash160)
+            remoteCommit.spec.htlcs.collect { case IncomingHtlc(add) if ripemd160(add.paymentHash) == paymentHash160 => add }
```
araspitzu: nit: this could also use `.collect(incoming)` as above

pm47: Done with cc438497f.

pm47:
I'm not sure, tbh I started this just with 617e2e6b481ef95962d5766532eae8520b93ea25 in mind and pulled the thread... I think if we don't have a better idea right now, let's merge this now, we can always improve later.

t-bast:
My proposal is to simply switch the names between the two functions, what do you think?

- `fulfillOutgoingHtlc` is what you call when the remote sent you an `UpdateFulfillHtlc`
- `fulfillIncomingHtlc` is what you call when you are about to send an `UpdateFulfillHtlc`

That sounds simple enough and coherent, is there an issue with that?

pm47:
> My proposal is to simply switch the names between the two functions, what do you think?

That sounds entirely reasonable, done with 33977e53b.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -62,54 +66,62 @@ object CommitmentSpec {
...
+      case IncomingHtlc(add) => spec.copy(toRemote = spec.toRemote - add.amountMsat, htlcs = spec.htlcs + directedHtlc)
     }
   }
 
   // OUT means we are sending an UpdateFulfillHtlc message which means that we are fulfilling an HTLC that they sent
-  def fulfillHtlc(spec: CommitmentSpec, direction: Direction, htlcId: Long): CommitmentSpec = {
-    spec.findHtlcById(htlcId, direction.opposite) match {
-      case Some(htlc) if direction == OUT => spec.copy(toLocal = spec.toLocal + htlc.add.amountMsat, htlcs = spec.htlcs - htlc)
-      case Some(htlc) if direction == IN => spec.copy(toRemote = spec.toRemote + htlc.add.amountMsat, htlcs = spec.htlcs - htlc)
+  def fulfillIncomingHtlc(spec: CommitmentSpec, htlcId: Long): CommitmentSpec = {
```
t-bast: Much better! Just need to remove the obsolete comment above and line 91

PR #1361, opened 2020-03-31
Separate configuration for peer-connection

pm47:
The goal is not having to bring a whole `NodeParams` to open a connection to a peer.

Another approach would have been to define `PeerConnectionConf` as a trait and to make `NodeParams` extend this trait, we had already considered doing this for router conf.

I'm still unsure what to do with the `features`. Should that be in the
`peer-connection` section, or is it more general? Same for `chain_hash`. I added a `def features = peerConnectionConf.features` to minimize the impact in the codebase.

The `override-features` is completely broken, it is only used when exchanging init messages, but elsewhere we use the regular `features` for all nodes.


eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -460,7 +462,19 @@ object PeerConnection {
...
+                                authTimeout: FiniteDuration,
+                                initTimeout: FiniteDuration,
+                                pingInterval: FiniteDuration,
+                                pingTimeout: FiniteDuration,
+                                pingDisconnect: Boolean,
+                                features: ByteVector,
+                                overrideFeatures: Map[PublicKey, ByteVector],
+                                syncWhitelist: Set[PublicKey],
+                                maxRebroadcastDelay: FiniteDuration
+                               )
```
araspitzu, line 477: nit: indentation

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -272,6 +271,18 @@ object NodeParams {
...
+        chainHash = chainHash,
+        authTimeout = FiniteDuration(config.getDuration("peer-connection.auth-timeout").getSeconds, TimeUnit.SECONDS),
+        initTimeout = FiniteDuration(config.getDuration("peer-connection.init-timeout").getSeconds, TimeUnit.SECONDS),
+        pingInterval = FiniteDuration(config.getDuration("peer-connection.ping-interval").getSeconds, TimeUnit.SECONDS),
+        pingTimeout = FiniteDuration(config.getDuration("peer-connection.ping-timeout").getSeconds, TimeUnit.SECONDS),
+        pingDisconnect = config.getBoolean("peer-connection.ping-disconnect"),
+        features = features,
+        overrideFeatures = overrideFeatures,
+        syncWhitelist = syncWhitelist,
+        maxRebroadcastDelay = FiniteDuration(config.getDuration("router.broadcast-interval").getSeconds, TimeUnit.SECONDS) // it makes sense to use the same value here
```
araspitzu, line 284: Would be nice to have a comment saying why we use a config key for the router here, or maybe a comment on the config file.

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -76,6 +77,8 @@ class Switchboard(nodeParams: NodeParams, router: ActorRef, watcher: ActorRef, r
 
     case 'peers => sender ! context.children
 
+    case GetRouterPeerConf => sender ! RouterPeerConf(nodeParams.routerConf, nodeParams.peerConnectionConf)
```
araspitzu, line 80: This seems to be unused, and what's the purpose exactly? Seems that any "entity" (i.e plugin) that can send akka messages to the switchboard probably have a `NodeParams` available and doesn't need that.

eclair-core/src/main/resources/reference.conf
```diff
@@ -99,11 +92,23 @@ eclair {
...
-  ping-timeout = 10 seconds // will disconnect if peer takes longer than that to respond
-  ping-disconnect = true // disconnect if no answer to our pings
+  peer-connection {
+    auth-timeout = 10 seconds // will disconnect if connection authentication doesn't happen within that timeframe
+    init-timeout = 10 seconds // will disconnect if initialization doesn't happen within that timeframe
+    ping-interval = 30 seconds
+    ping-timeout = 10 seconds // will disconnect if peer takes longer than that to respond
+    ping-disconnect = true // disconnect if no answer to our pings
+
+    features = "0a8a" // initial_routing_sync + option_data_loss_protect + option_channel_range_queries + option_channel_range_queries_ex + variable_length_onion
```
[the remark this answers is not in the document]
pm47: We would still need to pass them to the `PeerConnection` somehow. Maybe in the constructor? This would put the resolution (`features`/`overrideFeatures`) outside of the scope of `PeerConnection`, which tbh may be the proper way to do it. But by doing so, we move further from our standard pattern of defining actors `class ActorName(nodeParams: nodeParams, actor1: ActorRef, actor2: ActorRef...)`.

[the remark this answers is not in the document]
pm47: That's where the `trait` approach could come in handy too.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
pm47: it is correctly indented?

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
pm47: I gather that `// it makes sense to use the same value here` isn't satisfactory.

How about:

```suggestion
        maxRebroadcastDelay = FiniteDuration(config.getDuration("router.broadcast-interval").getSeconds, TimeUnit.SECONDS) // it makes sense to not delay rebroadcast by more than the rebroadcast period
```

araspitzu: Yes that's much better IMHO

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
araspitzu: Well we usually don't put the parenthesis on a new line, there are several instances of this in the codebase: [RouterConf](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala#L74), [RouteRequest](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala#L153), [SendPaymentConfig](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala#L309), and many others. But to be fair there are also a minority of case classes that are ~~wrongly~~ differently indented.

pm47: Good point, consistency FTW! Fixed with 884efe7e9.

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
araspitzu: Yes I was thinking to resolve `features/overrideFeatures` outside `PeerConnection` after all it's a node-wide setting that affects many parts other than the `Peer` but i see the point that would break our class defining pattern. What if we leave `features` inside the `PeerConnectionConf` object but outside the `peer-connection` config block? It's a bit of an exception but lets us avoid doing more complex things with the trait approach and we don't break of pattern.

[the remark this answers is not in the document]
t-bast: I agree that `features` is a node-wide parameter, so imo it shouldn't be inside the peer conf. I think the trait approach would be nice, the only painful part is figuring out how to mix-in some common fields that would be at the top-level of the conf is "sub-traits" (like providing features along with a `PeerParams`).

pm47:
Superseded by #1548.

PR #1364, opened 2020-04-02
Fixup funder fee buffer

t-bast:
We were previously only counting the additional HTLC at twice
the current feerate.

An HTLC in isolation doesn't make much sense: the feerate
applies to the whole commit tx. Our fee buffer needs to account
for a x2 feerate increase on the commit tx + an additional htlc.

Note that this introduces another subtlety. The commit tx fee at twice the
current feerate may actually be lower than the commit tx fee at the current
feerate (if the commit tx contains many htlcs that are only slightly above
the trim threshold).

See https://github.com/lightningnetwork/lightning-rfc/pull/740#discussion_r399740407 for context.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -101,6 +101,32 @@ case class Commitments(channelVersion: ChannelVersion,
 
   val announceChannel: Boolean = (channelFlags & 0x01) != 0
 
+  // NB: when computing availableBalanceForSend and availableBalanceForReceive, the funder keeps an extra buffer on top
+  // of its usual channel reserve to avoid getting channels stuck in case the on-chain feerate increases (see
+  // https://github.com/lightningnetwork/lightning-rfc/issues/728 for details).
+  //
+  // This extra buffer (which we call "funder fee buffer") is calculated as follows:
+  //  1) Simulate a x2 feerate increase and compute the corresponding commit tx fee (note that it may trim some HTLCs)
+  //  2) Add the cost of adding a new untrimmed HTLC at that increased feerate. This ensures that we'll be able to
+  //     actually use the channel to add new HTLCs if the feerate doubles.
+  //
+  // If for example the current feerate is 1000 sat/kw, the dust limit 546 sat, and we have 3 pending outgoing HTLCs for
+  // respectively 1250 sat, 2000 sat and 2500 sat.
+  // commit tx fee = commitWeight * feerate + 3 * htlcOutputWeight * feerate = 724 * 1000 + 3 * 172 * 1000 = 1240 sat
+  // To calculate the funder fee buffer, we first double the feerate and calculate the corresponding commit tx fee.
+  // By doubling the feerate, the first HTLC becomes trimmed so the result is: 724 * 2000 + 2 * 172 * 2000 = 2136 sat
+  // We then add the additional fee for a potential new untrimmed HTLC: 172 * 2000 = 344 sat
+  // The funder fee buffer is 2136 + 344 = 2480 sat
+  //
+  // If there are many pending HTLCs that are only slightly above the trim threshold, the funder fee buffer may be
+  // smaller than the current commit tx fee because those HTLCs will be trimmed and the commit tx weight will decrease.
+  // For example if we have 10 outgoing HTLCs of 1250 sat:
+  //  - commit tx fee = 724 * 1000 + 10 * 172 * 1000 = 2444 sat
+  //  - commit tx fee at twice the feerate = 724 * 2000 = 1448 sat (all HTLCs have been trimmed)
+  //  - cost of an additional untrimmed HTLC = 172 * 2000 = 344 sat
+  //  - funder fee buffer = 1448 + 344 = 1792 sat
+  // In that case the current commit tx fee is higher than the funder fee buffer and will dominate the balance restrictions.
+
   lazy val availableBalanceForSend: MilliSatoshi = {
     // we need to base the next current commitment on the last sig we sent, even if we didn't yet receive their revocation
     val remoteCommit1 = remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit)
@@ -109,16 +135,19 @@ case class Commitments(channelVersion: ChannelVersion,
     if (localParams.isFunder) {
       // The funder always pays the on-chain fees, so we must subtract that from the amount we can send.
       val commitFees = commitTxFeeMsat(remoteParams.dustLimit, reduced, commitmentFormat)
-      // the funder needs to keep an extra reserve to be able to handle fee increase without getting the channel stuck
-      // (see https://github.com/lightningnetwork/lightning-rfc/issues/728)
-      val funderFeeReserve = htlcOutputFee(reduced.feeratePerKw * 2, commitmentFormat)
-      val htlcFees = htlcOutputFee(reduced.feeratePerKw, commitmentFormat)
-      if (balanceNoFees - commitFees < offeredHtlcTrimThreshold(remoteParams.dustLimit, reduced, commitmentFormat)) {
+      // the funder needs to keep a "funder fee buffer" (see explanation above)
+      val funderFeeBuffer = commitTxFeeMsat(remoteParams.dustLimit, reduced.copy(feeratePerKw = reduced.feeratePerKw * 2), commitmentFormat) + htlcOutputFee(reduced.feeratePerKw * 2, commitmentFormat)
+      val amountToReserve = commitFees.max(funderFeeBuffer)
+      if (balanceNoFees - amountToReserve < offeredHtlcTrimThreshold(remoteParams.dustLimit, reduced, commitmentFormat)) {
         // htlc will be trimmed
-        (balanceNoFees - commitFees - funderFeeReserve).max(0 msat)
+        (balanceNoFees - amountToReserve).max(0 msat)
       } else {
         // htlc will have an output in the commitment tx, so there will be additional fees.
-        (balanceNoFees - commitFees - funderFeeReserve - htlcFees).max(0 msat)
+        val commitFees1 = commitFees + htlcOutputFee(reduced.feeratePerKw, commitmentFormat)
+        // we take the additional fees for that htlc output into account in the fee buffer at a x2 feerate increase
+        val funderFeeBuffer1 = funderFeeBuffer + htlcOutputFee(reduced.feeratePerKw * 2, commitmentFormat)
+        val amountToReserve1 = commitFees1.max(funderFeeBuffer1)
+        (balanceNoFees - amountToReserve1).max(0 msat)
       }
     } else {
       // The fundee doesn't pay on-chain fees.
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -197,10 +199,12 @@ object Commitments {
...
-    // (see https://github.com/lightningnetwork/lightning-rfc/issues/728)
-    val funderFeeReserve = htlcOutputFee(2 * reduced.feeratePerKw)
-    val missingForSender = reduced.toRemote - commitments1.remoteParams.channelReserve - (if (commitments1.localParams.isFunder) fees + funderFeeReserve else 0.msat)
+    // the funder needs to keep an extra buffer to be able to handle a x2 feerate increase and an additional htlc to avoid
+    // getting the channel stuck (see https://github.com/lightningnetwork/lightning-rfc/issues/728).
+    val funderFeeBuffer = commitTxFeeMsat(commitments1.remoteParams.dustLimit, reduced.copy(feeratePerKw = 2 * reduced.feeratePerKw)) + htlcOutputFee(2 * reduced.feeratePerKw)
+    // NB: increasing the feerate can actually remove htlcs from the commit tx (if they fall below the trim threshold)
+    // which may result in a lower commit tx fee; this is why we take the max of the two.
+    val missingForSender = reduced.toRemote - commitments1.remoteParams.channelReserve - (if (commitments1.localParams.isFunder) fees.max(funderFeeBuffer.truncateToSatoshi) else 0.sat)
```
SomberNight, line 207: note that this is not what c-lightning has implemented: https://github.com/ElementsProject/lightning/blob/af7e87930878117a0633bb7a8cd4488b8e9c5b6d/channeld/full_channel.c#L430-L440

they trim htlcs at the current feerate, and then double the feerate and not trim again

t-bast: And what do you think of it? I think their choice is a bit weird and the additional reserve it ends up adding may be too big, but that's just my opinion.

SomberNight: If we were to plot the commitment tx fees as a function of feerate, because of trimming, I expect it would be a spiky curve. This PR atm (sort of) takes the max of two samplings, while c-lightning takes a conservative upper bound.
As feerates change, the max of two samplings might be lower than the actual commitment tx fees, and potentially the issue the additional reserve is trying to fix could resurface (i.e. the fee spike buffer that is supposed to handle fee spikes up to 2x of current feerate might not be sufficient to handle all feerates in between). Though admittedly I did not look at specific numbers so perhaps this is highly unlikely in practice.

My personal opinion is that it might make more sense to take an upper bound of the curve, as c-lightning does, as then that should cover all possible feerates up to the chosen parameter (of "2x").

t-bast: > If we were to plot the commitment tx fees as a function of feerate, because of trimming, I expect it would be a spiky curve.

Yes exactly, and that can be surprise at first!

> the issue the additional reserve is trying to fix could resurface (i.e. the fee spike buffer that is supposed to handle fee spikes up to 2x of current feerate might not be sufficient to handle all feerates in between).

Yes but only temporarily, because the HTLCs you have in the commit tx will eventually settle.
Once they settle that gives you money back and unblocks the channel. And I think that's an important insight.

Taking the conservative upper bound also works, but it means there's more of your channel balance that you just can't use, which I want to avoid unless absolutely necessary. It's already hard to tell users "you put X mBTC in this channel but in fact what you'll be able to use is less than that, but trust us, this is necessary", so let's not add more to that if we can avoid it. Does that make sense?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -98,16 +98,17 @@ case class Commitments(channelVersion: ChannelVersion,
...
       val commitFees = commitTxFeeMsat(remoteParams.dustLimit, reduced)
-      // the funder needs to keep an extra reserve to be able to handle fee increase without getting the channel stuck
-      // (see https://github.com/lightningnetwork/lightning-rfc/issues/728)
-      val funderFeeReserve = htlcOutputFee(2 * reduced.feeratePerKw)
-      val htlcFees = htlcOutputFee(reduced.feeratePerKw)
-      if (balanceNoFees - commitFees < offeredHtlcTrimThreshold(remoteParams.dustLimit, reduced)) {
+      // the funder needs to keep an extra buffer to be able to handle a x2 feerate increase and an additional htlc to avoid
+      // getting the channel stuck (see https://github.com/lightningnetwork/lightning-rfc/issues/728).
+      val funderFeeBuffer = commitTxFeeMsat(remoteParams.dustLimit, reduced.copy(feeratePerKw = 2 * reduced.feeratePerKw)) + htlcOutputFee(2 * reduced.feeratePerKw)
+      if (balanceNoFees - commitFees.max(funderFeeBuffer) < offeredHtlcTrimThreshold(remoteParams.dustLimit, reduced)) {
```
pm47, line 104: Nit: maybe put `commitFees.max(funderFeeBuffer)` in a named variable for better readibility?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -98,16 +98,17 @@ case class Commitments(channelVersion: ChannelVersion,
...
         // htlc will be trimmed
-        (balanceNoFees - commitFees - funderFeeReserve).max(0 msat)
+        (balanceNoFees - commitFees.max(funderFeeBuffer)).max(0 msat)
       } else {
         // htlc will have an output in the commitment tx, so there will be additional fees.
-        (balanceNoFees - commitFees - funderFeeReserve - htlcFees).max(0 msat)
+        val commitFees1 = commitFees + htlcOutputFee(reduced.feeratePerKw)
+        val funderFeeBuffer1 = funderFeeBuffer + htlcOutputFee(2 * reduced.feeratePerKw)
```
pm47, line 110: You are adding an `htlcOutputFee` to `funderBufferFee` which already had an additional `htlcOutputFee`?

t-bast: This is damn confusing. An additional comment may help (I'm adding it).

In `funderFeeBuffer` we take into account the requirement to handle a *future* additional HTLC at `2 * feeratePerKw`.
In this `else` branch, we are adding a *present* HTLC into the commitment, so we need to take its fee into account (just as we add it to `commitFees` to obtain `commitFees1`). And since the fee buffer uses `2 * feeratePerKw`, this is why we use that value inside `htlcOutputFee`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1364/commits/d125826e282c8fb9d33330f8bd220fc8135dc370

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1364/commits/fbe4529c97a9c2b6c1056f49fe76551009708b94

t-bast:
What should we do about this PR @pm47?

We haven't noticed the stuck channel issue at all for months, which means the previous fix is enough to mitigate it.
Merging this PR would make us "more" spec-compliant but wastes a bit more liquidity than what we currently have, so I'm not sure we should merge unless users heavily request that.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -107,16 +107,20 @@ case class Commitments(channelVersion: ChannelVersion,
...
       val commitFees = commitTxFeeMsat(remoteParams.dustLimit, reduced)
-      // the funder needs to keep an extra reserve to be able to handle fee increase without getting the channel stuck
-      // (see https://github.com/lightningnetwork/lightning-rfc/issues/728)
-      val funderFeeReserve = htlcOutputFee(2 * reduced.feeratePerKw)
-      val htlcFees = htlcOutputFee(reduced.feeratePerKw)
-      if (balanceNoFees - commitFees < offeredHtlcTrimThreshold(remoteParams.dustLimit, reduced)) {
+      // the funder needs to keep an extra buffer to be able to handle a x2 feerate increase and an additional htlc to avoid
+      // getting the channel stuck (see https://github.com/lightningnetwork/lightning-rfc/issues/728).
+      val funderFeeBuffer = commitTxFeeMsat(remoteParams.dustLimit, reduced.copy(feeratePerKw = 2 * reduced.feeratePerKw)) + htlcOutputFee(2 * reduced.feeratePerKw)
+      val amountToReserve = commitFees.max(funderFeeBuffer)
```
pm47, line 113: How can `amountToReserve` not equal `funderFeeBuffer`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -107,16 +107,20 @@ case class Commitments(channelVersion: ChannelVersion,
...
         // htlc will be trimmed
-        (balanceNoFees - commitFees - funderFeeReserve).max(0 msat)
+        (balanceNoFees - amountToReserve).max(0 msat)
       } else {
         // htlc will have an output in the commitment tx, so there will be additional fees.
-        (balanceNoFees - commitFees - funderFeeReserve - htlcFees).max(0 msat)
+        val commitFees1 = commitFees + htlcOutputFee(reduced.feeratePerKw)
+        // we take that htlc output into account in the fee buffer at a x2 feerate increase
+        val funderFeeBuffer1 = funderFeeBuffer + htlcOutputFee(2 * reduced.feeratePerKw)
+        val amountToReserve1 = commitFees1.max(funderFeeBuffer1)
```
pm47, line 122: Same comment as above.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: I still don't get it. If we are in the *trimmed* case, then why do we include an `htlcOutputFee` in `funderFeeBuffer`?

t-bast: In this `else` branch we are in the *untrimmed* case, so we take the current htlc into account.
In `funderFeeBuffer` we kept some room for a *future, potential* untrimmed HTLC at twice the current feerate (this is really just keeping a buffer for an unknown future).
Is it a bit more clear?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: Because of the trim limit (damn this is tricky)...
Imagine you have a commit tx with many HTLCs in it that are slightly above the trim limit when using `reduced.feeratePerKw`.
When you use `2 * reduced.feeratePerKw` all these HTLCs become trimmed: the feerate is bigger, but the commit tx becomes smaller. If the number of pending HTLCs was high enough, the total fee may end up smaller than before.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: Ok I think I'm understanding your comment a bit better.

In all cases (both trimmed and untrimmed), we need to reserve the cost of adding a future untrimmed HTLC at twice the feerate (which we do when we calculate `funderFeeBuffer`).

If after that, we still have enough balance to send untrimmed HTLCs (which is what happens in this `else` branch), we need to take into account the cost of that untrimmed HTLC. This is why we add the htlc fee to `commitFees` to get `commitFees1`. Similarly, we need to add this HTLC to the funder fee buffer, but at twice the feerate, which is how we calculate `funderFeeBuffer1`. It may look like we already accounted for it when calculating `funderFeeBuffer`, but this is not the same HTLC! In `funderFeeBuffer` we added the cost for an HTLC in the future, whereas in `funderFeeBuffer1` we need to add the cost for an HTLC now (`availableBalanceForSend` considers that you're going to try to send that HTLC now).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: Added an example that showcases this in the latest commit.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: I get it now. Thanks!

pm47 (approved):

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -101,6 +101,32 @@ case class Commitments(channelVersion: ChannelVersion,
...
+  // The funder fee buffer is 2136 + 344 = 2480 sat
+  //
+  // If there are many pending HTLCs that are only slightly above the trim threshold, the funder fee buffer may be
+  // smaller than the current commit tx fee because those HTLCs will be trimmed and the commit tx weight will decrease.
+  // For example if we have 10 outgoing HTLCs of 1250 sat:
+  //  - commit tx fee = 724 * 1000 + 10 * 172 * 1000 = 2444 sat
+  //  - commit tx fee at twice the feerate = 724 * 2000 = 1448 sat (all HTLCs have been trimmed)
+  //  - cost of an additional untrimmed HTLC = 172 * 2000 = 344 sat
+  //  - funder fee buffer = 1448 + 344 = 1792 sat
+  // In that case the current commit tx fee is higher than the funder fee buffer and will dominate the balance restrictions.
```
ziggie1984, line 128: I think a fee-buffer can act in two different directions, meaning that in case of a down-spike former dust htlcs can use this buffer as well. So I would argue to not really dynamically simulate a 2x increase in the feerate and based of this calculate the feebuffer but rather just take the worst case scenario and account for the double fee at the current feerate (thats basically a 2x simulation without taking any trimmed htlcs into account) ? Why not consider tripped htlcs? Because trimmed htlcs are accounted for fees anyways. So in your above example when all 10 htlcs are trimmed, we have a final fee of
 0.724 * 2000 + 10 * 1250 =  13950 sasts, so we could just keep the double fee as a buffer because this would have the advantage, if there are already trimmed away htlcs on the commitment, the feerate could also be decreased in an event of a fee-dump ? wdyt ?

PR #1366, opened 2020-04-02
Refactor transaction generation errors

t-bast:
Transaction generation functions used to throw exceptions.
We have a perfectly good `TxGenerationSkipped` type to express potential errors,
so these functions should return an Either to make the contract explicit.

I also did some formatting clean-up in `Helpers`.

dream-exceptions

If that covid confinement keeps going for too long, I'm pretty sure I'll end up spending more time finding the right memes to go in my PR descriptions than doing the actual code...

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -262,7 +262,7 @@ object Helpers {
      *
      * @return (localSpec, localTx, remoteSpec, remoteTx, fundingTxOutput)
      */
-    def makeFirstCommitTxs(keyManager: KeyManager, channelVersion: ChannelVersion, temporaryChannelId: ByteVector32, localParams: LocalParams, remoteParams: RemoteParams, fundingAmount: Satoshi, pushMsat: MilliSatoshi, initialFeeratePerKw: Long, fundingTxHash: ByteVector32, fundingTxOutputIndex: Int, remoteFirstPerCommitmentPoint: PublicKey, maxFeerateMismatch: Double): (CommitmentSpec, CommitTx, CommitmentSpec, CommitTx) = {
+    def makeFirstCommitTxs(keyManager: KeyManager, channelVersion: ChannelVersion, temporaryChannelId: ByteVector32, localParams: LocalParams, remoteParams: RemoteParams, fundingAmount: Satoshi, pushMsat: MilliSatoshi, initialFeeratePerKw: Long, fundingTxHash: ByteVector32, fundingTxOutputIndex: Int, remoteFirstPerCommitmentPoint: PublicKey): (CommitmentSpec, CommitTx, CommitmentSpec, CommitTx) = {
       val toLocalMsat = if (localParams.isFunder) fundingAmount.toMilliSatoshi - pushMsat else pushMsat
       val toRemoteMsat = if (localParams.isFunder) pushMsat else fundingAmount.toMilliSatoshi - pushMsat
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -48,7 +48,6 @@ object Helpers {
   /**
    * Depending on the state, returns the current temporaryChannelId or channelId
    *
-   * @param stateData
    * @return the long identifier of the channel
    */
   def getChannelId(stateData: Data): ByteVector32 = stateData match {
@@ -63,9 +62,6 @@ object Helpers {
 
   /**
    * We update local/global features at reconnection
-   *
-   * @param data
-   * @return
    */
   def updateFeatures(data: HasCommitments, localInit: Init, remoteInit: Init): HasCommitments = {
     val commitments1 = data.commitments.copy(
@@ -86,7 +82,6 @@ object Helpers {
    * Returns the number of confirmations needed to safely handle the funding transaction,
    * we make sure the cumulative block reward largely exceeds the channel size.
    *
-   * @param nodeParams
    * @param fundingSatoshis funding amount of the channel
    * @return number of confirmations needed
    */
@@ -183,7 +178,6 @@ object Helpers {
    *
    * If current update more than [[Channel.REFRESH_CHANNEL_UPDATE_INTERVAL]] old then the delay will be zero.
    *
-   * @param currentUpdateTimestamp
    * @return the delay until the next update
    */
   def nextChannelUpdateRefresh(currentUpdateTimestamp: Long)(implicit log: DiagnosticLoggingAdapter): FiniteDuration = {
@@ -196,7 +190,6 @@ object Helpers {
   }
 
   /**
-   *
    * @param referenceFeePerKw reference fee rate per kiloweight
    * @param currentFeePerKw   current fee rate per kiloweight
    * @return the "normalized" difference between i.e local and remote fee rate: |reference - current| / avg(current, reference)
@@ -208,7 +201,6 @@ object Helpers {
     feeRateMismatch(networkFeeratePerKw, commitmentFeeratePerKw) > updateFeeMinDiffRatio
 
   /**
-   *
    * @param referenceFeePerKw       reference fee rate per kiloweight
    * @param currentFeePerKw         current fee rate per kiloweight
    * @param maxFeerateMismatchRatio maximum fee rate mismatch ratio
@@ -219,7 +211,6 @@ object Helpers {
     feeRateMismatch(referenceFeePerKw, currentFeePerKw) > maxFeerateMismatchRatio
 
   /**
-   *
    * @param remoteFeeratePerKw remote fee rate per kiloweight
    * @return true if the remote fee rate is too small
    */
@@ -227,7 +218,7 @@ object Helpers {
     remoteFeeratePerKw < fr.acinq.eclair.MinimumFeeratePerKw
   }
 
-  def makeAnnouncementSignatures(nodeParams: NodeParams, commitments: Commitments, shortChannelId: ShortChannelId) = {
+  def makeAnnouncementSignatures(nodeParams: NodeParams, commitments: Commitments, shortChannelId: ShortChannelId): AnnouncementSignatures = {
     val features = ByteVector.empty // empty features for now
     val fundingPubKey = nodeParams.keyManager.fundingPublicKey(commitments.localParams.fundingKeyPath)
     val (localNodeSig, localBitcoinSig) = nodeParams.keyManager.signChannelAnnouncement(fundingPubKey.path, nodeParams.chainHash, shortChannelId, commitments.remoteParams.nodeId, commitments.remoteParams.fundingPubKey, features)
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -511,16 +492,16 @@ object Helpers {
       }
       val (closingTx, closingSigned) = makeClosingTx(keyManager, commitments, localScriptPubkey, remoteScriptPubkey, remoteClosingFee)
       val signedClosingTx = Transactions.addSigs(closingTx, keyManager.fundingPublicKey(commitments.localParams.fundingKeyPath).publicKey, remoteParams.fundingPubKey, closingSigned.signature, remoteClosingSig)
-      Transactions.checkSpendable(signedClosingTx).map(x => signedClosingTx.tx).recover { case _ => throw InvalidCloseSignature(commitments.channelId, signedClosingTx.tx) }
+      Transactions.checkSpendable(signedClosingTx).map(_ => signedClosingTx.tx).recover { case _ => throw InvalidCloseSignature(commitments.channelId, signedClosingTx.tx) }
     }
 
-    def generateTx(desc: String)(attempt: Try[TransactionWithInputInfo])(implicit log: LoggingAdapter): Option[TransactionWithInputInfo] = {
+    def generateTx(desc: String)(attempt: Try[Either[TxGenerationSkipped, TransactionWithInputInfo]])(implicit log: LoggingAdapter): Option[TransactionWithInputInfo] = {
```
t-bast, line 498: I guess the main reason why tx generation was throwing an exception is because of this function.
We really really want to make sure our tx generation doesn't fail with an unhandled exception, so it kind of makes sense to wrap all of them in a `Try` like we currently do.

But what exactly may throw here? The tx signing? If that's the case we should probably go deeper and mark those functions as return a `Try`?

araspitzu: > But what exactly may throw here? The tx signing? If that's the case we should probably go deeper and mark those functions as return a Try?

I think this is the right direction but we won't be able to completely get rid of the `Try`  until we address all the possible reasons for why something might throw an exception down there. If we can catch in most of the errors (and translate them to `TxGenerationSkipped`) then the signing part should be the only one left to be done inside the `Try`. This is also a bit dangerous because if we overlook some code that can still throw exception we might have a failing tx generation because of an unhandled exception.

t-bast: > until we address all the possible reasons for why something might throw an exception down there.

That's exactly the goal ;)
We are fortunate enough to be using a functional language, where exceptions shouldn't be hidden below us. We should be able to tell whether some code we're calling is pure or not, and may throw exceptions or not. Otherwise we're not doing better than JS...

I don't think it's acceptable for code that handles users' money to say "I'm not exactly sure what this code does so I'll wrap it in a `Try` to be safe". I think we should know exactly what code may throw an exception that we want to catch, and use the type system to provide accurate function signatures. Maybe I'm too optimistic, but if we could get there our codebase would feel very very safe.

araspitzu: I agree on the principle and i think we can leverage the type system to do better. However scala is not a pure functional language and it still has ugly things like `null`s and NPE (which are absent in functional programming), so for example [here](https://github.com/ACINQ/eclair/pull/1366/files#diff-aeb98076f262543b70a86753bd3719adR519) if we pass a `null` commitTx parameter it will throw a NPE in the row below and if not caught by an external `Try..match` this can be fatal. 
Having said that i think we **can** work out all the possible exceptions that can be thrown by our critical code and wrap them into a `TxGenerationSkipped`, as you said you probably we still need to wrap the signing methods around a `Try` but that should be all of the exception-catching.

t-bast: > scala is not a pure functional language

True, maybe we can only get this far with Scala...when Haskell??

pm47: We are also using Akka, with its "Let it crash" philosophy. In general we do not even try to catch exceptions (which is a lost cause as @araspitzu explained). We do have a [`handleExceptions`](https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L2254-L2261) in Channels.scala, but it's merely to react to the inevitable exceptions, not to proactively handle them.

In this particular piece of code, which is very sensitive, we have been extra-cautious and have been using `Try`s as an added layer of defense against unexpected behavior. The reason is mainly that the underlying signature code contains JNI calls to libsecp256k1.

I would be fine with having `generateTx` return an `Either`, provided that exceptions are catched within the method, and converted to a `Left`.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1366/commits/a816c2fac0299f36df4964461f2b8ed10b611b30, I think it's nice to push this `Try` inside `generateTx` instead of repeating it every time we call `generateTx`.

t-bast:
@pm47 https://github.com/ACINQ/eclair/pull/1366/commits/b0e97818816f91fef3cdc019f6f294c70361ce50 removes an unused maxFeerateMismatch when creating the first commit tx. Is there a bug here? Should we actually use this field and reject channel opens when we have a big feerate mismatch?

pm47:
> @pm47 b0e9781 removes an unused maxFeerateMismatch when creating the first commit tx. Is there a bug here? Should we actually use this field and reject channel opens when we have a big feerate mismatch?

This dates back from this ancient commit: https://github.com/ACINQ/eclair/commit/875dc04d3904d44e7bb10bc30d53434f1d81e9a6#diff-5fa783866be9932536bd1825c7998eb7 where we moved the feerate validation earlier in the process: https://github.com/ACINQ/eclair/blob/ec7750211f5963f439a7997ce69907939b061935/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala#L137-L142

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -511,16 +492,19 @@ object Helpers {
       }
       val (closingTx, closingSigned) = makeClosingTx(keyManager, commitments, localScriptPubkey, remoteScriptPubkey, remoteClosingFee)
       val signedClosingTx = Transactions.addSigs(closingTx, keyManager.fundingPublicKey(commitments.localParams.fundingKeyPath).publicKey, remoteParams.fundingPubKey, closingSigned.signature, remoteClosingSig)
-      Transactions.checkSpendable(signedClosingTx).map(x => signedClosingTx.tx).recover { case _ => throw InvalidCloseSignature(commitments.channelId, signedClosingTx.tx) }
+      Transactions.checkSpendable(signedClosingTx).map(_ => signedClosingTx.tx).recover { case _ => throw InvalidCloseSignature(commitments.channelId, signedClosingTx.tx) }
```
pm47, line 495: Alternative without `throw`:

```suggestion
      Transactions
        .checkSpendable(signedClosingTx)
        .transform(_ => Success(signedClosingTx.tx), _ => Failure(InvalidCloseSignature(commitments.channelId, signedClosingTx.tx)))
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -511,16 +492,19 @@ object Helpers {
...
     }
 
-    def generateTx(desc: String)(attempt: Try[TransactionWithInputInfo])(implicit log: LoggingAdapter): Option[TransactionWithInputInfo] = {
-      attempt match {
-        case Success(txinfo) =>
+    /** Wraps transaction generation in a Try and filters failures to avoid one transaction negatively impacting a whole commitment. */
+    private def generateTx(desc: String)(attempt: () => Either[TxGenerationSkipped, TransactionWithInputInfo])(implicit log: LoggingAdapter): Option[TransactionWithInputInfo] = {
```
pm47, line 499: Do we really need to make `attempt` a function?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -700,7 +677,7 @@ object Helpers {
       require(tx.txIn.size == 1, "commitment tx should have 1 input")
       //val fundingPubKey = commitments.localParams.fundingPubKey(keyManager)
       val channelKeyPath = keyManager.channelKeyPath(localParams, channelVersion)
-      val obscuredTxNumber = Transactions.decodeTxNumber(tx.txIn(0).sequence, tx.lockTime)
+      val obscuredTxNumber = Transactions.decodeTxNumber(tx.txIn.head.sequence, tx.lockTime)
```
pm47, line 680: Nit: Line 678 can be removed.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: `match`/`case` would work too.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: If we want to be able to catch exceptions it may throw, I think we do. Do you see an alternative?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1366/commits/46e0a10dca00a7cc99bc633ed27f556edaa1f924

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: I see. No I don't.

pm47: Actually yes:
```suggestion
    private def generateTx(desc: String)(attempt: => Either[TxGenerationSkipped, TransactionWithInputInfo])(implicit log: LoggingAdapter): Option[TransactionWithInputInfo] = {
```

Allows you to call it that way:
```scala
      val mainDelayedTx = generateTx("main-delayed-output") {
        Transactions.makeClaimDelayedOutputTx(tx, localParams.dustLimit, localRevocationPubkey, remoteParams.toSelfDelay, localDelayedPubkey, localParams.defaultFinalScriptPubKey, feeratePerKwDelayed).right.map(claimDelayed => {
          val sig = keyManager.sign(claimDelayed, keyManager.delayedPaymentPoint(channelKeyPath), localPerCommitmentPoint)
          Transactions.addSigs(claimDelayed, sig)
        })
      }
```
Instead of:
```scala
      val mainDelayedTx = generateTx("main-delayed-output")(() =>
        Transactions.makeClaimDelayedOutputTx(tx, localParams.dustLimit, localRevocationPubkey, remoteParams.toSelfDelay, localDelayedPubkey, localParams.defaultFinalScriptPubKey, feeratePerKwDelayed).right.map(claimDelayed => {
          val sig = keyManager.sign(claimDelayed, keyManager.delayedPaymentPoint(channelKeyPath), localPerCommitmentPoint)
          Transactions.addSigs(claimDelayed, sig)
        }))
```

pm47: Done with 48db1d51a.

PR #1367, opened 2020-04-03
Support additional TLV records in SendPayentRequest

araspitzu:
This PR adds the possibility to attach extra TLV records when sending a payment, the feature is meant to be used by plugin developers and experimentally via API (although this PR is missing the API part). It will allow to support features like `keysend` (invoiceless payments) and/or custom messages attached as TLV record to final onion packet.
I've chosen to not support this feature in trampoline/mpp payments for the moment because i'm not aware of any custom protocol built on top them. 
Users willing to experiment with this feature must take care of properly sorting the extra TLV records.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
@@ -54,13 +54,14 @@ class PaymentInitiator(nodeParams: NodeParams, router: ActorRef, relayer: ActorR
         case Some(invoice) if invoice.features.allowMultiPart && Features.hasFeature(nodeParams.features, Features.BasicMultiPartPayment) =>
           invoice.paymentSecret match {
             case Some(paymentSecret) =>
-              spawnMultiPartPaymentFsm(paymentCfg) forward SendMultiPartPayment(paymentSecret, r.recipientNodeId, r.recipientAmount, finalExpiry, r.maxAttempts, r.assistedRoutes, r.routeParams)
+              spawnMultiPartPaymentFsm(paymentCfg) forward SendMultiPartPayment(paymentSecret, r.recipientNodeId, r.recipientAmount, finalExpiry, r.maxAttempts, r.assistedRoutes, r.routeParams, userCustomTlvs = r.userCustomTlvs)
             case None =>
               sender ! PaymentFailed(paymentId, r.paymentHash, LocalFailure(PaymentSecretMissing) :: Nil)
           }
         case _ =>
-          // NB: we only generate legacy payment onions for now for maximum compatibility.
-          spawnPaymentFsm(paymentCfg) forward SendPayment(r.recipientNodeId, FinalLegacyPayload(r.recipientAmount, finalExpiry), r.maxAttempts, r.assistedRoutes, r.routeParams)
+          val paymentSecret = r.paymentRequest.flatMap(_.paymentSecret)
+          val finalPayload = Onion.createSinglePartPayload(r.recipientAmount, finalExpiry, paymentSecret, r.userCustomTlvs)
+          spawnPaymentFsm(paymentCfg) forward SendPayment(r.recipientNodeId, finalPayload, r.maxAttempts, r.assistedRoutes, r.routeParams)
       }
 
     case r: SendTrampolinePaymentRequest =>
@@ -201,6 +202,7 @@ object PaymentInitiator {
    * @param externalId       (optional) externally-controlled identifier (to reconcile between application DB and eclair DB).
    * @param assistedRoutes   (optional) routing hints (usually from a Bolt 11 invoice).
    * @param routeParams      (optional) parameters to fine-tune the routing algorithm.
+   * @param userCustomTlvs   (optional) user-defined custom tlvs that will be added to the onion sent to the target node.
    */
   case class SendPaymentRequest(recipientAmount: MilliSatoshi,
                                 paymentHash: ByteVector32,
@@ -210,7 +212,8 @@ object PaymentInitiator {
                                 paymentRequest: Option[PaymentRequest] = None,
                                 externalId: Option[String] = None,
                                 assistedRoutes: Seq[Seq[ExtraHop]] = Nil,
-                                routeParams: Option[RouteParams] = None) {
+                                routeParams: Option[RouteParams] = None,
+                                userCustomTlvs: Seq[GenericTlv] = Nil) {
     // We add one block in order to not have our htlcs fail when a new block has just been found.
     def finalExpiry(currentBlockHeight: Long) = finalExpiryDelta.toCltvExpiry(currentBlockHeight + 1)
   }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
@@ -276,17 +276,14 @@ object Onion {
     NodeRelayPayload(TlvStream(tlvs2))
   }
 
-  def createSinglePartPayload(amount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: Option[ByteVector32] = None): FinalPayload = paymentSecret match {
-    // We try to use the legacy format as much as possible for maximum compatibility, but when we have a payment secret we need to use TLV to include it.
-    case Some(paymentSecret) => FinalTlvPayload(TlvStream(AmountToForward(amount), OutgoingCltv(expiry), PaymentData(paymentSecret, amount)))
+  def createSinglePartPayload(amount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: Option[ByteVector32] = None, userCustomTlvs: Seq[GenericTlv] = Nil): FinalPayload = paymentSecret match {
+    case Some(paymentSecret) => FinalTlvPayload(TlvStream(Seq(AmountToForward(amount), OutgoingCltv(expiry), PaymentData(paymentSecret, amount)), userCustomTlvs))
+    case None if userCustomTlvs.nonEmpty => FinalTlvPayload(TlvStream(Seq(AmountToForward(amount), OutgoingCltv(expiry)), userCustomTlvs))
     case None => FinalLegacyPayload(amount, expiry)
   }
 
-  def createMultiPartPayload(amount: MilliSatoshi, totalAmount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: ByteVector32): FinalPayload =
-    FinalTlvPayload(TlvStream(AmountToForward(amount), OutgoingCltv(expiry), PaymentData(paymentSecret, totalAmount)))
-
-  def createMultiPartPayload(amount: MilliSatoshi, totalAmount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: ByteVector32, additionalTlvs: Seq[OnionTlv]): FinalPayload =
-    FinalTlvPayload(TlvStream(AmountToForward(amount) +: OutgoingCltv(expiry) +: PaymentData(paymentSecret, totalAmount) +: additionalTlvs))
+  def createMultiPartPayload(amount: MilliSatoshi, totalAmount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: ByteVector32, additionalTlvs: Seq[OnionTlv] = Nil, userCustomTlvs: Seq[GenericTlv] = Nil): FinalPayload =
+    FinalTlvPayload(TlvStream(AmountToForward(amount) +: OutgoingCltv(expiry) +: PaymentData(paymentSecret, totalAmount) +: additionalTlvs, userCustomTlvs))
 
   /** Create a trampoline outer payload. */
   def createTrampolinePayload(amount: MilliSatoshi, totalAmount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: ByteVector32, trampolinePacket: OnionRoutingPacket): FinalPayload = {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
@@ -27,7 +27,7 @@ import scodec.{Attempt, Codec, Err}
  * Created by t-bast on 20/06/2019.
  */
 object TlvCodecs {
-
+  // high range types are greater than or equal 2^16, see https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#type-length-value-format
   private val TLV_TYPE_HIGH_RANGE = 65536
 
   /**
@@ -112,10 +112,7 @@ object TlvCodecs {
     }
   }
 
-  private val genericTlv: Codec[GenericTlv] = (("tag" | varint) :: variableSizeBytesLong(varintoverflow, bytes)).as[GenericTlv].exmap(
-    validateGenericTlv, // we must reject incoming even unknown tlv records
-    g => Attempt.Successful(g) // but we allow outgoing even unknown tlv records (so that users can provide experimental tlv records)
-  )
+  private val genericTlv: Codec[GenericTlv] = (("tag" | varint) :: variableSizeBytesLong(varintoverflow, bytes)).as[GenericTlv].exmap(validateGenericTlv, validateGenericTlv)
 
   private def tag[T <: Tlv](codec: DiscriminatorCodec[T, UInt64], record: Either[GenericTlv, T]): UInt64 = record match {
     case Left(generic) => generic.tag
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
@@ -27,6 +27,9 @@ import scodec.{Attempt, Codec, Err}
  * Created by t-bast on 20/06/2019.
  */
 object TlvCodecs {
+
+  private val TLV_TYPE_HIGH_RANGE = 65536
+
   /**
    * Truncated uint64 (0 to 8 bytes unsigned integer).
    * The encoder minimally-encodes every value, and the decoder verifies that values are minimally-encoded.
@@ -102,7 +105,7 @@ object TlvCodecs {
   val ltu16: Codec[Int] = variableSizeBytes(uint8, tu16)
 
   private def validateGenericTlv(g: GenericTlv): Attempt[GenericTlv] = {
-    if (g.tag.toBigInt % 2 == 0) {
+    if (g.tag < TLV_TYPE_HIGH_RANGE && g.tag.toBigInt % 2 == 0) {
       Attempt.Failure(Err("unknown even tlv type"))
     } else {
       Attempt.Successful(g)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
@@ -26,9 +26,7 @@ import scodec.{Attempt, Codec, Err}
 /**
  * Created by t-bast on 20/06/2019.
  */
-
 object TlvCodecs {
-
   /**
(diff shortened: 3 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -293,6 +293,9 @@ object Onion {
     FinalTlvPayload(TlvStream(AmountToForward(amount), OutgoingCltv(expiry), PaymentData(paymentSecret, totalAmount), TrampolineOnion(trampolinePacket)))
   }
 
+  def createCustomRecordsFinalPayload(amount: MilliSatoshi, expiry: CltvExpiry, additionalTlvRecords: Seq[GenericTlv] = Seq.empty): FinalPayload =
```
t-bast, line 296: I'm not sure we need a new function, I would just add the `additionalTlvRecords: Seq[GenericTlv] = Seq.empty` parameter to `createSinglePartPayload` and the two `createMultiPartPayload`.
For instance it makes sure we provide a `paymentSecret` which is really important even without MPP.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -210,7 +215,8 @@ object PaymentInitiator {
                                 paymentRequest: Option[PaymentRequest] = None,
                                 externalId: Option[String] = None,
                                 assistedRoutes: Seq[Seq[ExtraHop]] = Nil,
-                                routeParams: Option[RouteParams] = None) {
+                                routeParams: Option[RouteParams] = None,
+                                tlvRecords: Seq[GenericTlv] = Seq.empty) {
```
t-bast, line 219: I would make the name even more explicit, what about `userCustomRecords`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -59,8 +59,12 @@ class PaymentInitiator(nodeParams: NodeParams, router: ActorRef, relayer: ActorR
               sender ! PaymentFailed(paymentId, r.paymentHash, LocalFailure(PaymentSecretMissing) :: Nil)
           }
         case _ =>
-          // NB: we only generate legacy payment onions for now for maximum compatibility.
-          spawnPaymentFsm(paymentCfg) forward SendPayment(r.recipientNodeId, FinalLegacyPayload(r.recipientAmount, finalExpiry), r.maxAttempts, r.assistedRoutes, r.routeParams)
+          val finalPayload = if (r.tlvRecords.isEmpty) {
+            FinalLegacyPayload(r.recipientAmount, finalExpiry)
```
t-bast, line 63: I think it's worth updating this line to use the `createSinglePartPayload` function.
This way you can directly provide the additional records without the need for an `if/else`.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentInitiatorSpec.scala
```diff
@@ -69,6 +70,19 @@ class PaymentInitiatorSpec extends TestKit(ActorSystem("test")) with fixture.Fun
     withFixture(test.toNoArgTest(FixtureParam(nodeParams, initiator, payFsm, multiPartPayFsm, sender, eventListener)))
   }
 
+  test("send TLV payment with custom TLV records to node") { f =>
```
t-bast, line 73: nit: would better fit with the rest of the file:
```suggestion
  test("forward payment with user custom tlv records") { f =>
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
araspitzu: Okay, this is now using the new `createSinglePartPayload` and the decision between legacy/tlv final payload is made over there, if you don't pass a `payment_secret` it will use the legacy format. As discussed offline this created a bit of confusion for my use case (`keysend` plugin) because it turned out that LND can accept a keysend payment without `payment_secret` (and without the `final_amount` too :warning:) so plugin developers that want to attach custom records to their payment MUST attach a payment secret too.

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
araspitzu: :+1:  refactored and added the extra param to both methods

araspitzu:
I've added an extra test for the encoding/decoding of a TLV based FinalPayload with some custom records. 

There is another incompatibility that i found when developing `keysend` against LND, they have a [different interpretation](https://github.com/lightningnetwork/lnd/blob/master/record/custom_records.go) of the [spec](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#type-length-value-format) by which the "custom records" (as in those with type > 2^16) don't have be checked against the even/odd rule, in fact the keysend record type is defined as _even_ and is not usable in eclair at the moment.

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -276,17 +276,14 @@ object Onion {
...
+    case Some(paymentSecret) => FinalTlvPayload(TlvStream(records = Seq(AmountToForward(amount), OutgoingCltv(expiry), PaymentData(paymentSecret, amount)), unknown = additionalTlvRecords))
     case None => FinalLegacyPayload(amount, expiry)
   }
 
-  def createMultiPartPayload(amount: MilliSatoshi, totalAmount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: ByteVector32): FinalPayload =
-    FinalTlvPayload(TlvStream(AmountToForward(amount), OutgoingCltv(expiry), PaymentData(paymentSecret, totalAmount)))
-
-  def createMultiPartPayload(amount: MilliSatoshi, totalAmount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: ByteVector32, additionalTlvs: Seq[OnionTlv]): FinalPayload =
-    FinalTlvPayload(TlvStream(AmountToForward(amount) +: OutgoingCltv(expiry) +: PaymentData(paymentSecret, totalAmount) +: additionalTlvs))
+  def createMultiPartPayload(amount: MilliSatoshi, totalAmount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: ByteVector32, additionalOnionTlvs: Seq[OnionTlv] = Seq.empty, additionalTlvRecords: Seq[GenericTlv] = Seq.empty): FinalPayload =
```
t-bast, line 285: nit:
```suggestion
  def createMultiPartPayload(amount: MilliSatoshi, totalAmount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: ByteVector32, additionalTlvs: Seq[OnionTlv] = Seq.empty, customTlvs: Seq[GenericTlv] = Seq.empty): FinalPayload =
```

eclair-core/src/test/scala/fr/acinq/eclair/wire/OnionCodecsSpec.scala
```diff
@@ -168,7 +168,8 @@ class OnionCodecsSpec extends FunSuite {
...
+      TlvStream[OnionTlv](Seq(AmountToForward(561 msat), OutgoingCltv(CltvExpiry(42))), Seq(GenericTlv(UInt64.Conversions.longToUint64(5482373485L), hex"16c7ec71663784ff100b6eface1e60a97b92ea9d18b8ece5e558586bc7453828"))) -> hex"31 02020231 04012aff0000000146c6616d20 16c7ec71663784ff100b6eface1e60a97b92ea9d18b8ece5e558586bc7453828"
```
t-bast, line 172: Do you really need `UInt64.Conversions.longToUint64(5482373485L)`? Doesn't `UInt64(5482373485L)` work out of the box?

t-bast:
> they have a different interpretation of the spec by which the "custom records" (as in those with type > 2^16) don't have be checked against the even/odd rule, in fact the keysend record type is defined as even and is not usable in eclair at the moment.

I think it makes sense to have the keysend record be even. If you include it, you really need the receiver to understand it (otherwise your payment won't work).

Are you saying that eclair fails to include this record because we don't allow writing an unknown even type? Or are you saying that eclair will not allow receiving that keysend record (which is the right behavior)?

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -276,17 +276,14 @@ object Onion {
     NodeRelayPayload(TlvStream(tlvs2))
   }
 
-  def createSinglePartPayload(amount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: Option[ByteVector32] = None): FinalPayload = paymentSecret match {
+  def createSinglePartPayload(amount: MilliSatoshi, expiry: CltvExpiry, paymentSecret: Option[ByteVector32] = None, additionalTlvRecords: Seq[GenericTlv] = Seq.empty): FinalPayload = paymentSecret match {
```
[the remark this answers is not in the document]
araspitzu: Gotcha, i named them `customTlvRecords` wherever they're used and added a comment explaining they come from actual user input.

araspitzu:
I'm still having some troubles while choosing the right format for the final payload, with the [current logic](https://github.com/ACINQ/eclair/pull/1367/files#diff-17b76fbfdb1b6f0a578859b0fe5a82f7R281) we generate the TLV payload if there is a payment secret or there are user defined TLV records, however when paying an invoice with optional MPP (but we have MPP disabled) we should still generate a legacy payload but this is not the case because the invoice contains the payment secret. 

If we match only on the user defined TLV records and append the optional payment secret as TLV record we break the node relayer.

araspitzu:
>Are you saying that eclair fails to include this record because we don't allow writing an unknown even type? Or are you saying that eclair will not allow receiving that keysend record (which is the right behavior)

Yes eclair fails to produce the record because we don't allow writing an unknown even type, is this behavior correct? (I imagined we only want that when reading to make sure we understand the payment data)

t-bast:
> Yes eclair fails to produce the record because we don't allow writing an unknown even type, is this behavior correct?

Ok that's what I feared, I think for `unknown` records we should allow even types to go through.
Let me have a look at it to see how painful that would be.

araspitzu:
> Ok that's what I feared, I think for unknown records we should allow even types to go through.
Let me have a look at it to see how painful that would be.

So it's okay to allow unknown records to have an even type? If yes then something like [this](https://github.com/araspitzu/eclair/commit/e42ee507d306471e876148612a688f78ecd437a9) should do the trick

t-bast:
Here is my proposal: https://github.com/ACINQ/eclair/commit/4c265293d3ae9d8ad83c2d6152ca1606330e55c0

I changed the codecs to allow *encoding* unknown even tlvs, but still reject *decoding* such types.
I also did some clean-up/harmonization on the field names.

araspitzu:
> Here is my proposal: [4c26529](https://github.com/ACINQ/eclair/commit/4c265293d3ae9d8ad83c2d6152ca1606330e55c0)
> 
> I changed the codecs to allow _encoding_ unknown even tlvs, but still reject _decoding_ such types.
> I also did some clean-up/harmonization on the field names.
> 
> That means user will be able to use eclair to send payments to endpoints that support `keysend`, but not to receive. If we want receive support, that needs to be implemented inside eclair and the type needs to be explicitly added. Maybe we can allow decoding unknown even types in the high range as you suggest?

ACK,i was just merging your proposed change with mines. I think it would be nice if we allow decoding unkown even types in the high range (especially for keysend) and the [spec](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#type-length-value-format) is not very clear here, in fact there is one interpretation by which the `tlv_records` must adhere the odd/even rule but there is no mention of the rule for the custom records (so they don't support the odd/even rule?).  I'll push a change with your proposal + the decoding of unknown even types if in the high range.

araspitzu:
Latest commit allows to encode unknown even types but not to decode them. I gave a few shots to allow plugins to export a list of whitelisted even TLV types in [217862b](https://github.com/araspitzu/eclair/commit/217862bdb23a6f1f0985f900482c2a14753f13af) although i'm not very satisfied with it. We can have a new plugin hook that is used in Boot/Setup to inject a list of `UInt64`, however all our codecs are declared in singleton `object`s and it's impossible to inject anything into it. Possible ways are making it an instance (trait/class) *or* having the whitelist as implicit parameter, both of them are very impacting on the codebase and i'm not sure we want to go ahead with that. 

Some possible alternatives:
- Allow decoding even unknown types in the *high range*
- Adding the keysend record on the receiving side in eclair (after all it might make it into the spec)
- Not allowing to decode unknown even types at all: this would prevent from writing a keysend plugin 
for the receiving side

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentInitiatorSpec.scala
```diff
@@ -105,14 +120,14 @@ class PaymentInitiatorSpec extends TestKit(ActorSystem("test")) with fixture.Fun
...
-  test("forward legacy payment when multi-part deactivated", Tag("mpp_disabled")) { f =>
+  test("forward single-part payment when multi-part deactivated", Tag("mpp_disabled")) { f =>
     import f._
     val pr = PaymentRequest(Block.LivenetGenesisBlock.hash, Some(finalAmount), paymentHash, randomKey, "Some MPP invoice", features = Some(Features(VariableLengthOnion.optional, PaymentSecret.optional, BasicMultiPartPayment.optional)))
     val req = SendPaymentRequest(finalAmount, paymentHash, c, 1, CltvExpiryDelta(42), Some(pr))
     sender.send(initiator, req)
     val id = sender.expectMsgType[UUID]
     payFsm.expectMsg(SendPaymentConfig(id, id, None, paymentHash, finalAmount, c, Upstream.Local(id), Some(pr), storeInDb = true, publishEvent = true, Nil))
-    payFsm.expectMsg(SendPayment(c, FinalLegacyPayload(finalAmount, req.finalExpiry(nodeParams.currentBlockHeight)), 1))
+    payFsm.expectMsg(SendPayment(c, FinalTlvPayload(TlvStream(OnionTlv.AmountToForward(finalAmount), OnionTlv.OutgoingCltv(req.finalExpiry(nodeParams.currentBlockHeight)), OnionTlv.PaymentData(pr.paymentSecret.get, finalAmount))), 1))
```
araspitzu, line 130: I'm a bit hesitant here because with the proposed changes we're not using the legacy format anymore, can this be a problem? Note that in this test the recipient supports the `payment_secret` in the invoice so we know it can read a TLV final payload.

t-bast: I think this is fine. As long as for non-MPP cases we still use the legacy format we're good.

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
```diff
@@ -26,9 +26,10 @@ import scodec.{Attempt, Codec, Err}
 /**
  * Created by t-bast on 20/06/2019.
  */
-
 object TlvCodecs {
 
+  private val TLV_TYPE_HIGH_RANGE = 65536
```
t-bast, line 31: Maybe a comment pointing to the spec?

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
```diff
@@ -104,14 +105,17 @@ object TlvCodecs {
...
-    if (g.tag.toBigInt % 2 == 0) {
+    if (g.tag < TLV_TYPE_HIGH_RANGE && g.tag.toBigInt % 2 == 0) {
       Attempt.Failure(Err("unknown even tlv type"))
     } else {
       Attempt.Successful(g)
     }
   }
 
-  private val genericTlv: Codec[GenericTlv] = (("tag" | varint) :: variableSizeBytesLong(varintoverflow, bytes)).as[GenericTlv].exmap(validateGenericTlv, validateGenericTlv)
+  private val genericTlv: Codec[GenericTlv] = (("tag" | varint) :: variableSizeBytesLong(varintoverflow, bytes)).as[GenericTlv].exmap(
```
t-bast, line 115: Now that we chose to allow everything in the range > 2^16, I think we should keep the stricter version for tlv < 2^16.
So I'd keep the previous code here, WDYT?

eclair-core/src/test/scala/fr/acinq/eclair/wire/TlvCodecsSpec.scala
```diff
@@ -308,11 +306,25 @@ class TlvCodecsSpec extends FunSuite {
...
+
+    assert(testTlvStreamCodec.decode(testTlvStreamCodec.encode(lowRangeOdd).require).isSuccessful)
+    assert(testTlvStreamCodec.decode(testTlvStreamCodec.encode(lowRangeEven).require).isFailure)
+    assert(testTlvStreamCodec.decode(testTlvStreamCodec.encode(highRangeOdd).require).isSuccessful)
+    assert(testTlvStreamCodec.decode(testTlvStreamCodec.encode(highRangeEven).require).isSuccessful)
+  }
+
   test("encode invalid tlv stream") {
     val testCases = Seq(
-      // Unknown even type.
```
t-bast, line 313: I think you can restore those (once we reject encoding tlv types that are not in the > 2^16 range - see my other comment)

eclair-core/src/main/scala/fr/acinq/eclair/wire/TlvCodecs.scala
araspitzu: Agreed. It was a workaround  to play nice with plugins but since we now allow encoding/decoding even types above the high range it feels correct to *disallow* encoding them if they're not in the high range. Changes for this are in 8e5d22b

eclair-core/src/test/scala/fr/acinq/eclair/wire/TlvCodecsSpec.scala
```diff
@@ -308,6 +306,16 @@ class TlvCodecsSpec extends FunSuite {
...
   }
 
+  test("encode/decode even generic tlv types above high range") {
+    val lowRangeEven = TlvStream[TestTlv](records = Nil, unknown = Seq(GenericTlv(124, hex"2a")))
+    val highRangeEven = TlvStream[TestTlv](records = Nil, unknown = Seq(GenericTlv(67876545678L, hex"2b")))
+
+    assert(testTlvStreamCodec.encode(lowRangeEven).isFailure)
+    assert(testTlvStreamCodec.encode(highRangeEven).isSuccessful)
+    assert(testTlvStreamCodec.decode(hex"7c 01 2a".toBitVector).isFailure) // lowRangeEven
+    assert(testTlvStreamCodec.decode(testTlvStreamCodec.encode(highRangeEven).require.bytes.toBitVector).isSuccessful)
```
t-bast, line 316: I don't think you need the bytes <-> bits back and forth:

```suggestion
    assert(testTlvStreamCodec.decode(testTlvStreamCodec.encode(highRangeEven).require).isSuccessful)
```

eclair-core/src/test/scala/fr/acinq/eclair/wire/TlvCodecsSpec.scala
```diff
@@ -308,6 +306,16 @@ class TlvCodecsSpec extends FunSuite {
     assert(lengthPrefixedTestTlvStreamCodec.encode(stream).require.toByteVector === hex"0f 01012a 0b012b 0d012a fd00fe02002a")
   }
 
+  test("encode/decode even generic tlv types above high range") {
```
t-bast, line 309: The spec calls those "custom records", I think we should align with that.

```suggestion
  test("encode/decode custom even tlv records") {
```

araspitzu: Done in 73b82dd

eclair-core/src/test/scala/fr/acinq/eclair/wire/TlvCodecsSpec.scala
araspitzu: Done in 73b82dd

PR #1369, opened 2020-04-06
Better tracking of channels count metric

t-bast:
Track all channel states (cardinality shouldn't be too high).
Prevent duplicate transitions that messed up the gauge.

Also the `withoutTags` in the ctor wasn't such a good idea.
It's better to explicitly track each state with its own tag.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1682,8 +1682,6 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     case WAIT_FOR_INIT_INTERNAL -> WAIT_FOR_INIT_INTERNAL => () // called at channel initialization
     case state -> nextState =>
       if (state != nextState) {
-        if (state != WAIT_FOR_INIT_INTERNAL) Metrics.ChannelsCount.withTag(Tags.State, state.toString).decrement()
-        if (nextState != WAIT_FOR_INIT_INTERNAL) Metrics.ChannelsCount.withTag(Tags.State, nextState.toString).increment()
         context.system.eventStream.publish(ChannelStateChanged(self, peer, remoteNodeId, state, nextState, nextStateData))
       }
       if (nextState == CLOSED) {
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1685,13 +1682,14 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     case WAIT_FOR_INIT_INTERNAL -> WAIT_FOR_INIT_INTERNAL => () // called at channel initialization
     case state -> nextState =>
       if (state != nextState) {
+        if (state != WAIT_FOR_INIT_INTERNAL) Metrics.ChannelsCount.withTag(Tags.State, state.toString).decrement()
```
pm47, line 1685: Why use a dedicate `onTransition`? And why not count the initial state like every other?

```scala
  onTransition {
    case WAIT_FOR_INIT_INTERNAL -> WAIT_FOR_INIT_INTERNAL => Metrics.ChannelsCount.withTag(Tags.State, WAIT_FOR_INIT_INTERNAL.toString).increment()
    case state -> nextState if state != nextState =>
      Metrics.ChannelsCount.withTag(Tags.State, state.toString).decrement()
      Metrics.ChannelsCount.withTag(Tags.State, nextState.toString).increment()
  }
```

t-bast: You mean "why not", right?
We can use a dedicated `onTransition` if you prefer.
However we can't accurately count the `WAIT_FOR_INIT_INTERNAL`. Your proposal will increment it many more times than decrement it. It's quite painful to do correctly and I don't think it's very useful, which is why I'm only tracking after-init states.

pm47: Ok, how about:

```scala
  onTransition {
    case state -> nextState if state != nextState =>
      if (state != WAIT_FOR_INIT_INTERNAL) Metrics.ChannelsCount.withTag(Tags.State, state.toString).decrement()
      if (nextState != WAIT_FOR_INIT_INTERNAL) Metrics.ChannelsCount.withTag(Tags.State, nextState.toString).increment()
  }
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1369/commits/ef48c934bd40778ec3aabcaf825f630f2efd5b49

Issue #1205, opened 2019-11-10
Repeated failure and disconnects from one peer.


n1bor:
Every few mins I get the below entry in the logs....

I assume this is the other node misbehaving?
Looks like the sort of thing we need to black list them for?

```
21:53:39.255 DEBUG akka://eclair-node/system/IO-TCP/selectors/$a/0 akka.io.TcpListener - New connection accepted
21:53:39.257 INFO  akka://eclair-node/user/$l/server fr.acinq.eclair.io.Server - connected to /92.53.65.100:44926
21:53:39.258 DEBUG akka://eclair-node/user/$j/authenticator fr.acinq.eclair.io.Authenticator - authenticating connection to 92.53.65.100:44926 (pending=0 handlers=12)
21:53:39.259 INFO  akka://eclair-node/user/$j/authenticator/$cd f.a.eclair.crypto.TransportHandler - connection closed: PeerClosed
21:53:40.415 DEBUG akka://eclair-node/system/IO-TCP/selectors/$a/0 akka.io.TcpListener - New connection accepted
21:53:40.416 INFO  akka://eclair-node/user/$l/server fr.acinq.eclair.io.Server - connected to /92.53.65.100:45026
21:53:40.416 DEBUG akka://eclair-node/user/$j/authenticator fr.acinq.eclair.io.Authenticator - authenticating connection to 92.53.65.100:45026 (pending=0 handlers=12)
21:53:40.553 INFO  akka://eclair-node/user/$j/authenticator fr.acinq.eclair.io.Authenticator - connection authenticated with 0375d873f2ad784e004c9675ceeb302b6424559fc9e84d7a535a0c010ac55e3ec3@92.53.65.100:45026 direction=incoming
21:53:40.553 INFO  akka://eclair-node/user/$k/switchboard fr.acinq.eclair.io.Switchboard - creating new peer current=13
21:53:40.554 DEBUG akka://eclair-node/user/$k/switchboard/peer-0375d873f2ad784e004c9675ceeb302b6424559fc9e84d7a535a0c010ac55e3ec3 fr.acinq.eclair.io.Peer - got authenticated connection to 0375d873f2ad784e004c9675ceeb302b6424559fc9e84d7a535a0c010ac55e3ec3@92.53.65.100:45026
[...]

PR #1371, opened 2020-04-09
Do not reconnect immediately if we keep getting disconnected

pm47:
If our latest successful connection attempt was less
than 30 seconds ago, we pick up the exponential
back-off retry delay where we left it. The
goal is to address cases where the reconnection
is successful, but we are disconnected right away.


n1bor:
fixes this one too #1205 - that was closed...

PR #1372, opened 2020-04-09
Gossip metrics

t-bast:
Adding metrics for channel queries and gossip KPIs.

@sstone @pm47 this is a good opportunity to add other metrics/logs if there are things that you wanted to have while you were investigating an issue and were missing. Please let me know if I can add other metrics related to gossip, or logs that may be useful for eclair-mobile/phoenix.

Don't hesitate to bikeshed the naming too if you don't like it, it will be a bit painful to change metric names once we have graphs/alerts we rely on.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Monitoring.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Monitoring.scala
@@ -17,15 +17,39 @@
 package fr.acinq.eclair.router
 
 import fr.acinq.eclair.router.Router.GossipDecision
+import fr.acinq.eclair.wire.ChannelUpdate
 import fr.acinq.eclair.{LongToBtcAmount, MilliSatoshi, getSimpleClassName}
 import kamon.Kamon
-import kamon.metric.Counter
+import kamon.metric.{Counter, MeasurementUnit}
 
 object Monitoring {
 
   object Metrics {
     val FindRouteDuration = Kamon.timer("router.find-route.duration", "Path-finding duration")
     val RouteLength = Kamon.histogram("router.find-route.length", "Path-finding result length")
+
+    val QueryChannelRangeBlocksCount = Kamon.histogram("router.gossip.query-channel-range.blocks-count", "Number of blocks requested in query-channel-range")
+    val QueryChannelRangeRepliesCount = Kamon.histogram("router.gossip.query-channel-range.replies-count", "Number of reply-channel-range replies sent")
+
+    val ReplyChannelRangeIdsCount = Kamon.histogram("router.gossip.reply-channel-range.ids-count", "Number of short channel ids in reply-channel-range")
+    val ReplyChannelRangeBlocksCount = Kamon.histogram("router.gossip.reply-channel-range.blocks-count", "Number of blocks in reply-channel-range")
+    val NewChannelAnnouncements = Kamon.histogram("router.gossip.reply-channel-range.new-channel-announcements", "Number of new channel announcements discovered in reply-channel-range")
+    val NewChannelUpdates = Kamon.histogram("router.gossip.reply-channel-range.new-channel-updates", "Number of new channel updates discovered in reply-channel-range")
+
+    val QueryShortChannelIdsNodesCount = Kamon.histogram("router.gossip.query-short-channel-ids.node-announcements", "Number of node announcements sent in response to a query-short-channel-ids")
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Monitoring.scala
```diff
@@ -17,16 +17,40 @@
...
-import kamon.metric.Counter
+import kamon.metric.{Counter, MeasurementUnit}
 
 object Monitoring {
 
   object Metrics {
     val FindRouteDuration = Kamon.timer("router.find-route.duration", "Path-finding duration")
     val RouteLength = Kamon.histogram("router.find-route.length", "Path-finding result length")
 
+    val QueryChannelRangeBlocksCount = Kamon.histogram("router.gossip.query-channel-range.blocks-count", "Number of blocks requested in query-channel-range")
```
pm47, line 31: What is the rationale behind the trailing `-count`? Why does e.g. `NewChannelAnnouncements` not have it?

t-bast: True, I can remove those, they're implicit

eclair-core/src/main/scala/fr/acinq/eclair/router/Monitoring.scala
```diff
@@ -17,16 +17,40 @@
...
+    val QueryChannelRangeRepliesCount = Kamon.histogram("router.gossip.query-channel-range.replies-count", "Number of reply-channel-range replies sent")
+
+    val ReplyChannelRangeIdsCount = Kamon.histogram("router.gossip.reply-channel-range.ids-count", "Number of short channel ids in reply-channel-range")
+    val ReplyChannelRangeBlocksCount = Kamon.histogram("router.gossip.reply-channel-range.blocks-count", "Number of blocks in reply-channel-range")
+    val NewChannelAnnouncements = Kamon.histogram("router.gossip.reply-channel-range.new-channel-announcements", "Number of new channel announcements discovered in reply-channel-range")
```
pm47, line 36: The naming is a bit too generic, shouldn't that be:
```suggestion
    val ReplyChannelRangeNewChannelAnnouncements = Kamon.histogram("router.gossip.reply-channel-range.new-channel-announcements", "Number of new channel announcements discovered in reply-channel-range")
```

pm47: Maybe we could use `object` instead of very long names? E.g. `ReplyChannelRange.NewChannelAnnouncements`.

eclair-core/src/main/scala/fr/acinq/eclair/router/Monitoring.scala
```diff
@@ -17,16 +17,40 @@
...
+import kamon.metric.{Counter, MeasurementUnit}
 
 object Monitoring {
 
   object Metrics {
     val FindRouteDuration = Kamon.timer("router.find-route.duration", "Path-finding duration")
     val RouteLength = Kamon.histogram("router.find-route.length", "Path-finding result length")
 
+    val QueryChannelRangeBlocksCount = Kamon.histogram("router.gossip.query-channel-range.blocks-count", "Number of blocks requested in query-channel-range")
+    val QueryChannelRangeRepliesCount = Kamon.histogram("router.gossip.query-channel-range.replies-count", "Number of reply-channel-range replies sent")
```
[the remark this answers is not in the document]
t-bast: Do you want to rename this one `ReplyChannelRangeCount`? It kinda loses the information that it's the number of replies generated by a single incoming query, but as long as the comment here explains it...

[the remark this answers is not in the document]
pm47: Sorry to not be helping much, I just feel that something is a bit awkward in the naming but have trouble coming up with something better.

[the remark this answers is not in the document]
t-bast: Once inside a dedicated object, I think it's better. See https://github.com/ACINQ/eclair/pull/1372/commits/3e5a7b082198cc817452737f29a0662e1499e571

eclair-core/src/main/scala/fr/acinq/eclair/router/Monitoring.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1372/commits/3e5a7b082198cc817452737f29a0662e1499e571

PR #1376, opened 2020-04-14
Revert Electrum Tor disable SSL

t-bast:
Revert #1278. The correct way to handle that is to add more ways for
users to choose their SSL behavior in phoenix/eclair-mobile.

In the long run it will be via more checkboxes / a better configuration
panel, but at first it can simply be moving #1278's logic directly in both
Phoenix and eclair-mobile (which is done in https://github.com/ACINQ/phoenix/pull/53 and https://github.com/ACINQ/eclair-mobile/pull/242).


dpad85:
It does make sense to move this logic to the front mobile apps, so that the front can change it according to its needs. It will probably mean having the exact same duplicated logic in eclair-mobile/phoenix, but that's not a problem.

PR #1377, opened 2020-04-15
Add logback config for eclair-node tests

pm47:
Otherwise we use the default logback.xml file which writes to a
rolling file in the user directory.

NB: when enabling console logging, there are a lot of WARN outputs. Maybe something to look into @araspitzu?
```shell
10:37:33.657 WARN  ActorSystemImpl - withRequestTimeout was used in route however no request-timeout is set!
10:37:33.657 WARN  ActorSystemImpl - withRequestTimeoutResponse was used in route however no request-timeout is set!
10:37:33.683 WARN  ActorSystemImpl - withRequestTimeout was used in route however no request-timeout is set!
10:37:33.683 WARN  ActorSystemImpl - withRequestTimeoutResponse was used in route however no request-timeout is set!         
```

### DIFF
--- a/eclair-node/src/test/resources/logback-test.xml
+++ b/eclair-node/src/test/resources/logback-test.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2019 ACINQ SAS
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


araspitzu:
The reason is that when we test the `ServiceSpec` we use `route.TestKit` instead of starting an actual HTTP server, this is the correct choice but there is a known issue with timeouts and akka-http testing https://github.com/akka/akka-http/issues/952

pm47:
Fair enough, merging

PR #1379, opened 2020-04-15
Use sender instead of providing actor refs

pm47:
In some messages, we provide a reference to the sending actor byt adding `self` in the message, but we could use `sender` instead.


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -201,14 +201,14 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
     case Event(n: NodeAnnouncement, d: Data) =>
       stay using Validation.handleNodeAnnouncement(d, nodeParams.db.network, Set(LocalGossip), n)
 
-    case Event(PeerRoutingMessage(peerConnection, remoteNodeId, n: NodeAnnouncement), d: Data) =>
-      stay using Validation.handleNodeAnnouncement(d, nodeParams.db.network, Set(RemoteGossip(peerConnection, remoteNodeId)), n)
+    case Event(PeerRoutingMessage(remoteNodeId, n: NodeAnnouncement), d: Data) =>
+      stay using Validation.handleNodeAnnouncement(d, nodeParams.db.network, Set(RemoteGossip(sender, remoteNodeId)), n)
```
t-bast, line 205: Don't you still have an issue with `RemoteGossip` containing an actor ref?

pm47: Maybe in the future, but not now because we are only exchanging `PeerRoutingMessage` between `PeerConnection` and `Router` actors. The router then uses `GossipOrigin` internally, and with `Peer` but those are local.

PR #1380, opened 2020-04-16
Support bitcoin-0.19.1

araspitzu:
Update our regtests to use bitcoin-0.19.1, the most notable changes in this release are about fees (especially the CPFP-carve out rule for the dual-anchor commitment format). There are also some RPC changes: the `fundrawtransaction` RPC does not fail when providing a too little feerate but instead bitcoin will fail when broadcasting the transaction with `sendrawtransaction`, the error message is also changed.

I replaced the URLs for our bitcoin-core test instance with bitcoincore.org which is quite faster than bitcoin.org

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ You will find detailed guides and frequently asked questions there.
 
 ### Configuring Bitcoin Core
 
-:warning: Eclair requires Bitcoin Core 0.17.1 or higher. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
+:warning: Eclair requires Bitcoin Core 0.18.1 or 0.19.1. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
 
 Eclair needs a _synchronized_, _segwit-ready_, **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node.
 Eclair will use any BTC it finds in the default Bitcoin Core wallet to fund any channels you choose to open. Eclair will return BTC from closed channels to this wallet. You can have multiple Bitcoin Core wallets but make sure that the default one is always available.
--- a/eclair-core/pom.xml
+++ b/eclair-core/pom.xml
@@ -79,10 +79,9 @@
                 <activeByDefault>true</activeByDefault>
             </activation>
             <properties>
-                <bitcoind.url>https://bitcoin.org/bin/bitcoin-core-0.18.1/bitcoin-0.18.1-x86_64-linux-gnu.tar.gz
-                </bitcoind.url>
-                <bitcoind.md5>d3159a28702ca0cba2e0459e83219dfb</bitcoind.md5>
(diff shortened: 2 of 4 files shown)


sstone:
I think that the impact of recent fee policy changes should be documented in this PR: do we need to change how we create funding transactions ? can funding transactions be created but not published ? Do users need to take any action ?

araspitzu:
Right, so here is the gist of the fee policy changes in the new bitcoin release:
- [CPFP carve-out](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.19.0.1.md#mempool-and-transaction-relay): allows for one extra ancestor in the CPFP "package" of transactions, this relaxes the mempool acceptance constraints to allow the dual-anchor design. We don't have to take action for this.
-  [`-maxtxfee`](https://github.com/bitcoin/bitcoin/pull/16257) parameter in `fundrawtransaction`: this is a hard limit set by default to 0.1BTC to the maximum fee payable by a transaction created via RPC, we don't have to take action for this as our transactions are (hopefully) paying less than 0.1BTC, although it's worth to bear this parameter in mind.

One notable change for eclair are the RPCs `fundrawtransction` and `sendrawtransaction`, it's a bit undocumented in the release notes but their behavior has slightly changed. `fundrawtransaction` does **not** fail anymore if the feerate is too low, instead bitcoin-core will fail to broadcast it (with `sendrawtransaction`) if it doesn't meet the criteria to enter the mempool, this means that transactions can be created but not published (the min-fee parameters are unchanged in this release and it's still 253 sat/kw). Users don't have to take any action nor we need to change the way the transactions are created because we correctly handle failure to publish transactions.

t-bast:
Our README still mentions that we need at least bitcoin core 0.17.1, do we want to update that to a more recent version? And our wiki?

pm47:
@sstone please confirm that you have checked the hashes.

pm47:
> Our README still mentions that we need at least bitcoin core 0.17.1, do we want to update that to a more recent version? And our wiki?

Good point, we should recommend what we test.

araspitzu:
> > Our README still mentions that we need at least bitcoin core 0.17.1, do we want to update that to a more recent version? And our wiki?
> 
> Good point, we should recommend what we test.

Do we want to drop support to 0.17.1? I've been using 0.18 for quite some time in my personal instance, i suggest to update the README saying we support 0.18 and 0.19 but we recommend 0.18, WDYT?

pm47:
> Do we want to drop support to 0.17.1? I've been using 0.18 for quite some time in my personal instance, i suggest to update the README saying we support 0.18 and 0.19 but we recommend 0.18, WDYT?

But we could be breaking compatibility with 0.17.1 and 0.18 tomorrow without noticing, because we are not testing them. I have no strong opinion on this though.

sstone:
@pm47 Yes I've checked all hashes
Good point about the readme which I think should be updated to 0.18.1 and higher (I'm  in favor of dropping support for older versions)

araspitzu:
I updated the README to support explicitly 0.18.1 and 0.19.1, so when 0.20 is out we don't get caught off-guard :), both supported versions are actively tested either by running eclair on mainnet with 0.18.1 or using the 0.19.1 it in the tests.

PR #1382, opened 2020-04-16
Router channel balance

t-bast:
This is a first step towards more changes to the router/path-finding.
It makes sense for the router to know about channel balances when looking for routes, especially for MPP.

This PR only adds a listener on `AvailableBalanceChanged` and tracks local channel balance in the router's internal maps. The next step will be to add this information to the graph edges. It will also be a good occasion to use the channel `capacity` instead of `htlcMaximumMsat` inside edges; we'll be able to provide more information to the path-finding algorithm.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -19,7 +19,7 @@ package fr.acinq.eclair.router
 import akka.actor.{ActorContext, ActorRef, Status}
 import akka.event.LoggingAdapter
 import fr.acinq.bitcoin.Crypto.PublicKey
-import fr.acinq.bitcoin.{ByteVector32, ByteVector64}
+import fr.acinq.bitcoin.{ByteVector32, ByteVector64, Satoshi}
 import fr.acinq.eclair.payment.PaymentRequest.ExtraHop
 import fr.acinq.eclair.router.Graph.GraphStructure.DirectedGraph.graphEdgeToHop
 import fr.acinq.eclair.router.Graph.GraphStructure.{DirectedGraph, GraphEdge}
@@ -38,14 +38,16 @@ object RouteCalculation {
   def finalizeRoute(d: Data, fr: FinalizeRoute)(implicit ctx: ActorContext, log: LoggingAdapter): Data = {
     implicit val sender: ActorRef = ctx.self // necessary to preserve origin when sending messages to other actors
 
-    // NB: using a capacity of 0 msat will impact the path-finding algorithm. However here we don't run any path-finding, so it's ok.
-    val assistedChannels: Map[ShortChannelId, AssistedChannel] = fr.assistedRoutes.flatMap(toAssistedChannels(_, fr.hops.last, 0 msat)).toMap
-    val extraEdges = assistedChannels.values.map(ac => GraphEdge(ChannelDesc(ac.extraHop.shortChannelId, ac.extraHop.nodeId, ac.nextNodeId), toFakeUpdate(ac.extraHop, ac.htlcMaximum))).toSet
+    val assistedChannels: Map[ShortChannelId, AssistedChannel] = fr.assistedRoutes.flatMap(toAssistedChannels(_, fr.hops.last, fr.amount)).toMap
+    val extraEdges = assistedChannels.values.map(ac =>
+      GraphEdge(ChannelDesc(ac.extraHop.shortChannelId, ac.extraHop.nodeId, ac.nextNodeId), toFakeUpdate(ac.extraHop, ac.htlcMaximum), htlcMaxToCapacity(ac.htlcMaximum), Some(ac.htlcMaximum))
+    ).toSet
     val g = extraEdges.foldLeft(d.graph) { case (g: DirectedGraph, e: GraphEdge) => g.addEdge(e) }
     // split into sublists [(a,b),(b,c), ...] then get the edges between each of those pairs
     fr.hops.sliding(2).map { case List(v1, v2) => g.getEdgesBetween(v1, v2) }.toList match {
       case edges if edges.nonEmpty && edges.forall(_.nonEmpty) =>
-        val selectedEdges = edges.map(_.maxBy(_.update.htlcMaximumMsat.getOrElse(0 msat))) // select the largest edge
+        // select the largest edge (using balance when available, otherwise capacity).
+        val selectedEdges = edges.map(es => es.maxBy(e => e.balance_opt.getOrElse(e.capacity.toMilliSatoshi)))
         val hops = selectedEdges.map(d => ChannelHop(d.desc.a, d.desc.b, d.update))
         ctx.sender ! RouteResponse(hops, Set.empty, Set.empty)
       case _ => // some nodes in the supplied route aren't connected in our graph
@@ -56,21 +58,22 @@ object RouteCalculation {
 
   def handleRouteRequest(d: Data, routerConf: RouterConf, currentBlockHeight: Long, r: RouteRequest)(implicit ctx: ActorContext, log: LoggingAdapter): Data = {
     implicit val sender: ActorRef = ctx.self // necessary to preserve origin when sending messages to other actors
-    import r._
 
     // we convert extra routing info provided in the payment request to fake channel_update
     // it takes precedence over all other channel_updates we know
-    val assistedChannels: Map[ShortChannelId, AssistedChannel] = assistedRoutes.flatMap(toAssistedChannels(_, target, amount)).toMap
-    val extraEdges = assistedChannels.values.map(ac => GraphEdge(ChannelDesc(ac.extraHop.shortChannelId, ac.extraHop.nodeId, ac.nextNodeId), toFakeUpdate(ac.extraHop, ac.htlcMaximum))).toSet
-    val ignoredEdges = ignoreChannels ++ d.excludedChannels
+    val assistedChannels: Map[ShortChannelId, AssistedChannel] = r.assistedRoutes.flatMap(toAssistedChannels(_, r.target, r.amount)).toMap
+    val extraEdges = assistedChannels.values.map(ac =>
+      GraphEdge(ChannelDesc(ac.extraHop.shortChannelId, ac.extraHop.nodeId, ac.nextNodeId), toFakeUpdate(ac.extraHop, ac.htlcMaximum), htlcMaxToCapacity(ac.htlcMaximum), Some(ac.htlcMaximum))
+    ).toSet
+    val ignoredEdges = r.ignoreChannels ++ d.excludedChannels
     val defaultRouteParams: RouteParams = getDefaultRouteParams(routerConf)
-    val params = routeParams.getOrElse(defaultRouteParams)
+    val params = r.routeParams.getOrElse(defaultRouteParams)
     val routesToFind = if (params.randomize) DEFAULT_ROUTES_COUNT else 1
-
-    log.info(s"finding a route $source->$target with assistedChannels={} ignoreNodes={} ignoreChannels={} excludedChannels={}", assistedChannels.keys.mkString(","), ignoreNodes.map(_.value).mkString(","), ignoreChannels.mkString(","), d.excludedChannels.mkString(","))
+    
+    log.info(s"finding a route ${r.source}->${r.target} with assistedChannels={} ignoreNodes={} ignoreChannels={} excludedChannels={}", assistedChannels.keys.mkString(","), r.ignoreNodes.map(_.value).mkString(","), r.ignoreChannels.mkString(","), d.excludedChannels.mkString(","))
     log.info(s"finding a route with randomize={} params={}", routesToFind > 1, params)
-    findRoute(d.graph, source, target, amount, numRoutes = routesToFind, extraEdges = extraEdges, ignoredEdges = ignoredEdges, ignoredVertices = ignoreNodes, routeParams = params, currentBlockHeight)
-      .map(r => ctx.sender ! RouteResponse(r, ignoreNodes, ignoreChannels))
+    findRoute(d.graph, r.source, r.target, r.amount, numRoutes = routesToFind, extraEdges = extraEdges, ignoredEdges = ignoredEdges, ignoredVertices = r.ignoreNodes, routeParams = params, currentBlockHeight)
+      .map(route => ctx.sender ! RouteResponse(route, r.ignoreNodes, r.ignoreChannels))
       .recover { case t => ctx.sender ! Status.Failure(t) }
     d
   }
@@ -95,6 +98,9 @@ object RouteCalculation {
         (nextAmount, acs + (extraHop.shortChannelId -> AssistedChannel(extraHop, nextNodeId, nextAmount)))
     }._2
   }
+
+  /** Bolt 11 routing hints don't include the channel's capacity, so we round up the maximum htlc amount. */
+  def htlcMaxToCapacity(htlcMaximum: MilliSatoshi): Satoshi = htlcMaximum.truncateToSatoshi + 1.sat
 
   /**
    * This method is used after a payment failed, and we want to exclude some nodes that we know are failing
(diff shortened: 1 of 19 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -38,14 +38,16 @@ object RouteCalculation {
...
+    val extraEdges = assistedChannels.values.map(ac =>
+      GraphEdge(ChannelDesc(ac.extraHop.shortChannelId, ac.extraHop.nodeId, ac.nextNodeId), toFakeUpdate(ac.extraHop, ac.htlcMaximum), htlcMaxToCapacity(ac.htlcMaximum), Some(ac.htlcMaximum))
+    ).toSet
     val g = extraEdges.foldLeft(d.graph) { case (g: DirectedGraph, e: GraphEdge) => g.addEdge(e) }
     // split into sublists [(a,b),(b,c), ...] then get the edges between each of those pairs
     fr.hops.sliding(2).map { case List(v1, v2) => g.getEdgesBetween(v1, v2) }.toList match {
       case edges if edges.nonEmpty && edges.forall(_.nonEmpty) =>
-        val selectedEdges = edges.map(_.maxBy(_.update.htlcMaximumMsat.getOrElse(0 msat))) // select the largest edge
+        // select the largest edge (using balance when available, otherwise capacity).
+        val selectedEdges = edges.map(es => es.maxBy(e => e.balance_opt.getOrElse(e.capacity.toMilliSatoshi)))
```
araspitzu, line 50: Is there the chance that we mix the largest by `capacity` and by `balance`? The `balance` should be available on all local channels but if we end up having a `None` then we'd effectively compare balance and capacity which is wrong, 
for example in the following situation we'd pick the wrong channel:
1. local_channel_AB: _toLocal_: 1000 sat, _toRemote_: 50000 sat, _capacity_: 50000 sat
2. local_channel_CB: _toLocal_: 0 sat, _toRemote_: 60000 sat,  _capacity_: 60000 sat

Maybe we can add a test to make sure this doesn't happen.

t-bast: There is a small window where a race is possible and the router still has a `None` for some of the local channels (after a node restart). But I don't think it's an issue. We've lived this far with only looking at the capacity without too many issues. Looking at the balance when available is already a net improvement over that. And if that payment attempt fails, the retry will very likely not hit the race condition and will have all balances available.

Note that in your example, the previous code would have failed to route all the time. Now it has a chance to route once the router has the balance information available. It's just a matter of waiting a bit to have the information propagate. I don't think it's worth adding more complexity here for such an edge case.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -275,12 +275,18 @@ object Validation {
       // related channel is already known (note: this means no related channel_update is in the stash)
       val publicChannel = true
       val pc = d.channels(u.shortChannelId)
+      val pc1 = update match {
+        case Left(lcu) => pc.updateChannelUpdateSameSideAs(u).updateBalances(lcu.commitments)
+        case Right(_) => pc.updateChannelUpdateSameSideAs(u)
+      }
       val desc = getDesc(u, pc.ann)
```
araspitzu, line 282:
```suggestion
      val desc = getDesc(u, pc1.ann)
```

And all the instances below, we should be using `pc1` and not `pc` anymore.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -328,7 +334,11 @@ object Validation {
     } else if (d.privateChannels.contains(u.shortChannelId)) {
       val publicChannel = false
       val pc = d.privateChannels(u.shortChannelId)
-      val desc = if (Announcements.isNode1(u.channelFlags)) ChannelDesc(u.shortChannelId, pc.nodeId1, pc.nodeId2) else ChannelDesc(u.shortChannelId, pc.nodeId2, pc.nodeId1)
+      val pc1 = update match {
+        case Left(lcu) => pc.updateChannelUpdateSameSideAs(u).updateBalances(lcu.commitments)
+        case Right(_) => pc.updateChannelUpdateSameSideAs(u)
+      }
+      val desc = getDesc(u, pc)
```
araspitzu, line 341: same as above, we should be using `pc1`

t-bast:
> Should we add a test to check that we update the capacity of PublicChannels once we get the validation result?

Good idea, I'll check that.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -274,24 +278,33 @@ object Router {
 
   // @formatter:off
   case class ChannelDesc(shortChannelId: ShortChannelId, a: PublicKey, b: PublicKey)
-  case class PublicChannel(ann: ChannelAnnouncement, fundingTxid: ByteVector32, capacity: Satoshi, update_1_opt: Option[ChannelUpdate], update_2_opt: Option[ChannelUpdate]) {
+  case class PublicChannel(ann: ChannelAnnouncement, fundingTxid: ByteVector32, capacity: Satoshi, balance_1_opt: Option[MilliSatoshi], update_1_opt: Option[ChannelUpdate], balance_2_opt: Option[MilliSatoshi], update_2_opt: Option[ChannelUpdate]) {
```
pm47, line 281: My gut feeling is that adding those two arguments next to the existing ones doesn't feel right. The `PublicChannel` class holds *verified*, *static* information (expecially `ChannelAnnouncement` and `fundingTxid`, but also `update_X_opt` contain the timestamp at which they are valid). Current balances are pretty different and are more akin to a metadata, they are "always late", not updated the same way, etc. Furthermore, `update_X_opt` are optional because they may not exist, whereas balances always exist, but we just may not know them.

How about regrouping those meta information in a self-contained attribute to make that more clear? The code could be reused for public and private channels too.
```suggestion
  case class PublicChannel(ann: ChannelAnnouncement, fundingTxid: ByteVector32, capacity: Satoshi, update_1_opt: Option[ChannelUpdate], update_2_opt: Option[ChannelUpdate], meta: Option[ChannelMeta]) {
```

We could use this meta to store more data, information about past successes, and so on. The more complex our route finding algo is, the more data it will probably need.

Also, we may only know the balances of a channel for local channels right, which doesn't appear in the model. Not sure if that's a good thing or not.

t-bast: > How about regrouping those meta information in a self-contained attribute to make that more clear?

Good idea, I'll update that, this will allow more extensibility in the future.

> Also, we may only know the balances of a channel for local channels right, which doesn't appear in the model. Not sure if that's a good thing or not.

I hesitated to make `PublicChannel` a trait, and have a `LocalPublicChannel` and `RemotePublicChannel`. But I abandoned that because we could have balance information for non-local channels. If you deploy multiple nodes in the network, it may be (in the far future) useful to synchronize the balances between your nodes to improve your path-finding heuristics. So I prefer the solution of adding a `meta` that may contain balances to all channels, regardless of local/remote.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
t-bast: Actually no, on the contrary, the usage below needs to use `pc` because it checks the previous state compare to the update we're receiving.
For the line you suggest it ends up being the same, but I think using `pc` makes more sense.

t-bast: On the contrary, the usage below needs to use `pc` because it checks the previous state compare to the update we're receiving.
For the line you suggest it ends up being the same, but I think using `pc` makes more sense.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1382/commits/4a43bd8568434bf32d9f83a0fdf5ad8177b60986

t-bast:
> Should we add a test to check that we update the capacity of PublicChannels once we get the validation result?

It is done in RouterSpec's "update local channels balances" test.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
araspitzu: I see, i was confused by the naming, i think it's not clear because we usually use this pattern to update an object and use the updated instance from there on in the code. We should rename those `pc` and `pc1` to something more readable, but probably in another PR.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -258,8 +257,9 @@ object Validation {
     }
   }
 
-  def handleChannelUpdate(d: Data, db: NetworkDb, routerConf: RouterConf, origins: Set[GossipOrigin], u: ChannelUpdate, wasStashed: Boolean = false)(implicit ctx: ActorContext, log: LoggingAdapter): Data = {
+  def handleChannelUpdate(d: Data, db: NetworkDb, routerConf: RouterConf, origins: Set[GossipOrigin], update: Either[LocalChannelUpdate, ChannelUpdate], wasStashed: Boolean = false)(implicit ctx: ActorContext, log: LoggingAdapter): Data = {
```
pm47, line 260: This change create potential inconsistency with the `origins` field unfortunately.

`update == Left(_)` <=> `origins == Set(LocalGossip)`

And also confusion between `handleChannelUpdate` vs `handleLocalChannelUpdate`.

t-bast: Yes for a `LocalChannelUpdate` you need to set origins to `Local`, but that doesn't feel too dangerous, the risk of an inconsistency appearing here is quite low.

Do you think we should merge `handleChannelUpdate` and `handleLocalChannelUpdate` in a single function (at the risk of having a huge function)?

t-bast: Actually it's quite a mess to group everything in a single method, because `handleLocalChannelUpdate` may apply some changes to the `Data` before calling `handleChannelUpdate`...

pm47: Changing the signature to:
```suggestion
  def handleChannelUpdate(d: Data, db: NetworkDb, routerConf: RouterConf, update: Either[LocalChannelUpdate, (Set[RemoteGossip], ChannelUpdate)], wasStashed: Boolean = false)(implicit ctx: ActorContext, log: LoggingAdapter): Data = {
```
would solve the inconsistency, not sure if it is better overall though.

pm47: > Do you think we should merge handleChannelUpdate and handleLocalChannelUpdate in a single function (at the risk of having a huge function)?

No I don't think so

t-bast: Let me know what you think of https://github.com/ACINQ/eclair/pull/1382/commits/cd37fa1ab42e04d78ca183931ee6c1ed818d10be

t-bast: We can't directly use a `Set[RemoteGossip]`, we have to keep a `Set[GossipOrigin]` for `channel_update`s we receive via payment failures (they are remote, but we don't have an origin for them). Maybe in that case we could use an empty set instead, but that's something I'd like to keep for another PR as it's getting away from the core of this PR.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -347,15 +357,19 @@ object Validation {
...
         ctx.system.eventStream.publish(ChannelUpdatesReceived(u :: Nil))
         // we also need to update the graph
-        val graph1 = d.graph.removeEdge(desc).addEdge(desc, u)
-        d.copy(privateChannels = d.privateChannels + (u.shortChannelId -> pc.updateChannelUpdateSameSideAs(u)), graph = graph1)
+        val graph1 = if (Announcements.isEnabled(u.channelFlags)) {
+          d.graph.removeEdge(desc).addEdge(desc, u, pc1.capacity, pc1.getBalanceSameSideAs(u))
+        } else {
+          d.graph.removeEdge(desc)
+        }
+        d.copy(privateChannels = d.privateChannels + (u.shortChannelId -> pc1), graph = graph1)
```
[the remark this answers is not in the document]
t-bast: Yep, probably not impactful because there aren't many private channels used for relaying, but in the future we may use them more

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -258,29 +257,36 @@ object Validation {
...
+        (rcu.channelUpdate, rcu.origins, remote)
     }
     if (d.channels.contains(u.shortChannelId)) {
       // related channel is already known (note: this means no related channel_update is in the stash)
       val publicChannel = true
       val pc = d.channels(u.shortChannelId)
+      val pc1 = update match {
+        case Left(lcu) => pc.updateChannelUpdateSameSideAs(u).updateBalances(lcu.commitments)
+        case Right(_) => pc.updateChannelUpdateSameSideAs(u)
+      }
```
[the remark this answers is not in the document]
t-bast: We can't do that, we need to use the "old" pc in most places.
I put the computation of `pc1` here to avoid duplication, but I can move it to a helper function which will probably be clearer.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -258,29 +257,36 @@ object Validation {
...
+      case Left(lcu) => (lcu.channelUpdate, Set[GossipOrigin](LocalGossip), Nil)
+      case Right(rcu) =>
+        val remote = rcu.origins.collect {
+          case RemoteGossip(peerConnection, _) if wasStashed => peerConnection
+          case RemoteGossip(peerConnection, _) =>
+            peerConnection ! TransportHandler.ReadAck(rcu.channelUpdate)
+            log.debug("received channel update for shortChannelId={}", rcu.channelUpdate.shortChannelId)
+            peerConnection
+        }.toSeq
+        (rcu.channelUpdate, rcu.origins, remote)
```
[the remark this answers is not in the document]
t-bast: I'll try that!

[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1382/commits/b940a675c019abe022d28572bfd3e0b61dc7cf70, it's better!

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1382/commits/b940a675c019abe022d28572bfd3e0b61dc7cf70, I think it's clearer

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -257,55 +261,61 @@ object Validation {
...
+  private def applyChannelUpdate(pc: PublicChannel, update: Either[LocalChannelUpdate, RemoteChannelUpdate]): PublicChannel = update match {
+    case Left(lcu) => pc.updateChannelUpdateSameSideAs(lcu.channelUpdate).updateBalances(lcu.commitments)
+    case Right(rcu) => pc.updateChannelUpdateSameSideAs(rcu.channelUpdate)
+  }
+
+  private def applyChannelUpdate(pc: PrivateChannel, update: Either[LocalChannelUpdate, RemoteChannelUpdate]): PrivateChannel = update match {
+    case Left(lcu) => pc.updateChannelUpdateSameSideAs(lcu.channelUpdate).updateBalances(lcu.commitments)
+    case Right(rcu) => pc.updateChannelUpdateSameSideAs(rcu.channelUpdate)
+  }
+
```
pm47, line 273: Shouldn't those be part of `PublicChannel`/`PrivateChannel`?

t-bast: I hesitated, we can.

PR #1383, opened 2020-04-16
Send `update_fee` on reconnection

### LINKED ISSUE #1381: Funder should refresh the `update_fee` on reconnection
Funder is responsible for keeping the channel feerate in sync with the bitcoin network by sending `update_fee` from time to time. Our current policy is to do this when channels receive the `CurrentFeerates` event, which is sent every 10 minutes.

This works well for always-connected peers, but not for mostly-offline peers like mobile nodes. On reconnection, funder should check if the channel feerate is off (with `Helpers.shouldUpdateFee`) and send an `update_fee` if needed.

t-bast:
We update transaction fees at every block (ie every 10 minutes). While this works well when the remote peer is a node that's online for more than 10 minutes, it's an issue for mobile wallets that usually come online for a few minutes and then disconnect.

We want to make sure we send these wallet peers an update_fee when one is needed, so we now check for feerate updates on reconnection.

Fixes #1381.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1545,8 +1545,20 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
                 ()
             }
           }
+
           // we will re-enable the channel after some delay to prevent flappy updates in case the connection is unstable
           setTimer(Reconnected.toString, BroadcastChannelUpdate(Reconnected), 10 seconds, repeat = false)
+
+          // We usually handle feerate updates once per block (~10 minutes), but when our remote is a mobile wallet that
+          // only briefly connects and then disconnects, we may never have the opportunity to send our `update_fee`, so
+          // we send it (if needed) when reconnected.
+          if (d.commitments.localParams.isFunder) {
+            val currentFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw
+            val networkFeeratePerKw = nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(nodeParams.onChainFeeConf.feeTargets.commitmentBlockTarget)
+            if (Helpers.shouldUpdateFee(currentFeeratePerKw, networkFeeratePerKw, nodeParams.onChainFeeConf.updateFeeMinDiffRatio)) {
+              self ! CMD_UPDATE_FEE(networkFeeratePerKw, commit = true)
+            }
+          }
 
           goto(NORMAL) using d.copy(commitments = commitments1) sending sendQueue
       }
--- a/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
@@ -488,7 +488,7 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     alice2blockchain.expectMsg(PublishAsap(aliceCommitTx))
   }
 
-  test("handle feerate changes while offline (don't close on mismatch, update on reconnect)", Tag("disable-offline-mismatch")) { f =>
+  test("handle feerate changes while offline (don't close on mismatch)", Tag("disable-offline-mismatch")) { f =>
     import f._
     val sender = TestProbe()
 
@@ -504,8 +504,28 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     val highFeerate = FeeratesPerKw.single(tooHighFeeratePerKw)
 
     // this time Alice will ignore feerate changes for the offline channel
-    alice.underlyingActor.nodeParams.onChainFeeConf.feeEstimator.asInstanceOf[TestFeeEstimator].setFeerate(highFeerate)
     sender.send(alice, CurrentFeerates(highFeerate))
+    alice2blockchain.expectNoMsg()
+    alice2bob.expectNoMsg()
+  }
+
+  test("handle feerate changes while offline (update at reconnection)") { f =>
+    import f._
+    val sender = TestProbe()
+
+    // we simulate a disconnection
+    sender.send(alice, INPUT_DISCONNECTED)
+    sender.send(bob, INPUT_DISCONNECTED)
+    awaitCond(alice.stateName == OFFLINE)
+    awaitCond(bob.stateName == OFFLINE)
+
+    val localFeeratePerKw = alice.stateData.asInstanceOf[DATA_NORMAL].commitments.localCommit.spec.feeratePerKw
+    val networkFeeratePerKw = 2 * localFeeratePerKw
+    val networkFeerate = FeeratesPerKw.single(networkFeeratePerKw)
+
+    // Alice ignores feerate changes while offline
+    alice.underlyingActor.nodeParams.onChainFeeConf.feeEstimator.asInstanceOf[TestFeeEstimator].setFeerate(networkFeerate)
+    sender.send(alice, CurrentFeerates(networkFeerate))
     alice2blockchain.expectNoMsg()
     alice2bob.expectNoMsg()
 
@@ -520,7 +540,7 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     bob2alice.forward(alice)
 
     alice2bob.expectMsgType[FundingLocked] // since the channel's commitment hasn't been updated, we re-send funding_locked
-    alice2bob.expectMsg(UpdateFee(channelId(alice), tooHighFeeratePerKw))
+    alice2bob.expectMsg(UpdateFee(channelId(alice), networkFeeratePerKw))
   }
 
   test("handle feerate changes while offline (fundee scenario)") { f =>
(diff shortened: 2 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -941,6 +941,16 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       goto(NORMAL) using d.copy(channelUpdate = channelUpdate) storing() replying ChannelCommandResponse.Ok
 
     case Event(BroadcastChannelUpdate(reason), d: DATA_NORMAL) =>
+      // We usually handle feerate updates once per block (~10 minutes), but when our remote is a mobile wallet that only
+      // briefly connects and then disconnects, we may never have the opportunity to send our `update_fee`, so we always
+      // send it (if needed) when reconnected.
+      if (reason == Reconnected && d.commitments.localParams.isFunder) {
+        val currentFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw
```
araspitzu, line 948: Why not reusing `handleCurrentFeerate`? It could become:
```
val currentFeeRate = CurrentFeerates(nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(nodeParams.onChainFeeConf.feeTargets.commitmentBlockTarget)
handleCurrentFeerate(currentFeeRate, d)
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
```diff
@@ -508,9 +504,23 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     val highFeerate = FeeratesPerKw.single(tooHighFeeratePerKw)
 
     // this time Alice will ignore feerate changes for the offline channel
```
araspitzu, line 506: nit: both the name of the test and this comment are outdated now, i think either we split the test or update the name to something more descriptive.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: This code unfortunately doesn't work, you can't build back the `CurrentFeerates` that easily, it would require ugly code to call `feeEstimator` for every block target...
And `handleCurrentFeerate` does more than what we do, it handles fundee side and does a state transition which we shouldn't do in that specific case.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
t-bast: Right, the comment should be updated to say that we only handle feerate change when reconnecting, thanks

t-bast:
> The approach is correct but I'm not 100% sure we should use BroadcastChannelUpdate, isn't this event fired every REFRESH_CHANNEL_UPDATE_INTERVAL = 10 days? (Recall we want to re-send the update_fee on channel reestablish)

Look at the handler for `ChannelReestablish`, that's where we create this event. We add a delay of 10 seconds after reconnection to avoid flappyness, which is something I think we want to re-use for the `update_fee`: this is why I put it here. It's true that this event is more general, the alternative takes more code (we'd need to define a new event for `BroadcastFeeUpdate`), but if you strongly prefer separating these two I can do that.

araspitzu:
> > The approach is correct but I'm not 100% sure we should use BroadcastChannelUpdate, isn't this event fired every REFRESH_CHANNEL_UPDATE_INTERVAL = 10 days? (Recall we want to re-send the update_fee on channel reestablish)
> 
> Look at the handler for `ChannelReestablish`, that's where we create this event. We add a delay of 10 seconds after reconnection to avoid flappyness, which is something I think we want to re-use for the `update_fee`: this is why I put it here. It's true that this event is more general, the alternative takes more code (we'd need to define a new event for `BroadcastFeeUpdate`), but if you strongly prefer separating these two I can do that.

Right, this looks correct to me and i'm fine with reusing `BroadcastChannelUpdate` with the nice comment you put on it. One possible issues we might have is if the mobile app user tries to send the payment before the 10s  (so before we send out the `update_fee`).

t-bast:
> One possible issues we might have is if the mobile app user tries to send the payment before the 10s (so before we send out the update_fee).

This isn't necessarily an issue. The `update_fee` isn't meant to be blocking, there can be concurrent payments happening. At worst the `update_fee` will be deferred if the additional HTLC makes the new fee unapplicable, but once the HTLC settles it means the funder has more balance on his side so it's very likely that next time it will be able to send out the `update_fee`, even if more HTLCs are added to the commit tx.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -941,6 +941,16 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+        val networkFeeratePerKw = nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(nodeParams.onChainFeeConf.feeTargets.commitmentBlockTarget)
```
pm47, line 949: Why not use this code in the `when(SYNCING)` handler right before going to normal?

t-bast: For the same reason that delay was added for sending the `channel_update`, to avoid flappy connections. But in the case of `update_fee` maybe it's better to send it directly at the end of the `channel_reestablish` in syncing?

pm47: I think it's better to send it right away, because it can be viewed as maintaining the commitment tx (just like re-sending sig/rev). 

Note that https://github.com/ACINQ/eclair/pull/1371 addressed the outgoing flappy connections issue, still need to address the incoming ones (with some kind of banning probably).

t-bast: I'll move this to the end of the `channel_reestablish` handshake then.

t-bast: Done.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1545,9 +1545,21 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
                 ()
             }
           }
+
           // we will re-enable the channel after some delay to prevent flappy updates in case the connection is unstable
           setTimer(Reconnected.toString, BroadcastChannelUpdate(Reconnected), 10 seconds, repeat = false)
 
+          // We usually handle feerate updates once per block (~10 minutes), but when our remote is a mobile wallet that
```
pm47, line 1552: What do you think? A bit simpler to reason about no?

t-bast: Yep looks good and it's not much code.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1545,9 +1545,21 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+          // only briefly connects and then disconnects, we may never have the opportunity to send our `update_fee`, so
+          // we send it (if needed) when reconnected.
+          if (d.commitments.localParams.isFunder) {
+            val currentFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw
```
pm47, line 1556: We should look at the remote commit, because we may already have a pending `update_fee`, and it will first appear in the remote commitment. Btw this means that `handleCurrentFeerate` is bugged too.

```suggestion
            val currentFeeratePerKw = d.commitments.remoteCommit.spec.feeratePerKw
```

pm47: Actually, we really care about our commitment, so shouldn't we also take the local commitment into consideration? Damn...

t-bast: I think we really want it to be applied to our local commit. I don't think it matters much if we end up sending a duplicate `update_fee`, the spec doesn't prevent that so better be safe here and look at our commit?

pm47: I was afraid of creating a loop but we can only having one pending sig so I guess we're good.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
```diff
@@ -492,7 +488,7 @@ class OfflineStateSpec extends TestkitBaseClass with StateTestsHelperMethods {
     alice2blockchain.expectMsg(PublishAsap(aliceCommitTx))
   }
 
-  test("handle feerate changes while offline (don't close on mismatch)", Tag("disable-offline-mismatch")) { f =>
+  test("handle feerate changes while offline (don't close on mismatch, update on reconnect)", Tag("disable-offline-mismatch")) { f =>
```
pm47, line 491: The test is good, but you have kind of hacked an existing unrelated one. I think this should be a separate test, a bit like "re-send channel_update at reconnection for private channels".

t-bast: Good point, done.

PR #1385, opened 2020-04-20
Use human readable features in configuration

### LINKED ISSUE #1359: Human-readable features in reference.conf
It would  be great if we could define features in a human-readable way in `reference.conf`, something like:

```
features = {
  option_data_loss_protect = mandatory
  gossip_queries = optional
  payment_secret = mandatory
}
```

We would probably need that kind of thing (didn't go too far in prototyping):

```scala
object Features {
  object Resolution {
    private val mandatory = "mandatory"
    private val optional = "optional"
    val resolveName = Map(
      (OptionDataLossProtect.rfcName, mandatory) -> OptionDataLossProtect.mandatory,
      (OptionDataLossProtect.rfcName, optional) -> OptionDataLossProtect.optional,
      (InitialRoutingSync.rfcName, optional) -> InitialRoutingSync.optional,
      (ChannelRangeQueries.rfcName, mandatory) -> ChannelRangeQueries.mandatory,
      (ChannelRangeQueries.rfcName, optional) -> ChannelRangeQueries.optional,
      (VariableLengthOnion.rfcName, mandatory) -> VariableLengthOnion.mandatory,
     ...
    )
  }
}
```

araspitzu:
Allow users to specify the desired features in human readable way, instead of the byte vector. When parsing from config we ignore unknown features but we crash if there is a known feature with unknown support, we also pretty print the features when using `/getinfo` API or when parsing a BOLT11 invoice.

Fixes #1359


eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -154,4 +161,67 @@ object Features {
    */
   def areSupported(features: ByteVector): Boolean = areSupported(features.bits)
 
+  object Resolution {
```
t-bast, line 164: I think the functions/contracts in this object are a bit confusing.
What do you think of the following alternative proposal:

```scala
object Resolution {
    
    private val knownFeatures: Seq[Feature] = Seq(
      OptionDataLossProtect,
      InitialRoutingSync,
      ChannelRangeQueries,
      VariableLengthOnion,
      ChannelRangeQueriesExtended,
      PaymentSecret,
      BasicMultiPartPayment,
      Wumbo,
      TrampolinePayment
    )

    case class ActivatedFeature(feature: Feature, support: FeatureSupport)

    def fromConfiguration(config: Config): Seq[ActivatedFeature] = ???
    
    def toByteVector(features: Seq[ActivatedFeature]): ByteVector = ???

    def fromByteVector(features: ByteVector): Seq[ActivatedFeature] = ???

}
```

The `ActivatedFeature` would replace the `HumanReadableFeature` in `Eclair.scala`.
We could maybe even create `case class ActivatedFeatures(features: ActivatedFeature*)` and put `fromByteVector/toByteVector` as methods on this class? I think that would make the contracts clearer.

araspitzu: Currently the role of `HumanReadableFeature` is to help render the features as JSON, notice how the fields `name` and `support` end up serialized as JSON and that's what i wanted to achieve, it makes sense for anyone consuming the API to have that kind of object in the response.

I like the suggestion of introducing `ActivatedFeature` but i think it's even better if we take it one step further and incapsulate them inside a `Features` object, this would be carried around in the codebase instead of the ByteVector features.

```
case class Features(features: Seq[ActivatedFeature]){
  def hasFeature(f: Feature): Boolean
  def toByteVector: ByteVector
  def apply(raw: ByteVector): Features
}
```
WDYT?

araspitzu:
> This is going to be a nice improvement!
> I'd like to agree on a clearer interface/functions before reviewing the tests, see my comment below.
> I think you're also missing config migration (in `makeNodeParams`) and corresponding tests?

Gotcha, i added the config migration + test in the latest commit

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: > how the fields name and support end up serialized as JSON

But that's true as well with `ActivatedFeature`, right? The only difference is that the JSON is `{ feature: data_loss_protect, support: mandatory }` instead of `name`?

> I like the suggestion of introducing ActivatedFeature but i think it's even better if we take it one step further and incapsulate them inside a Features object, this would be carried around in the codebase instead of the ByteVector features.

I really like that! That means `NodeParams` would directly contain this `Features` object instead of a `ByteVector`, it would be really nice.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -154,4 +161,67 @@ object Features {
...
+    private val knownFeatures: Seq[Feature] = Seq(
```
pm47, line 165: How about using some reflection to guard us against forgetting new future features? How ugly would it look like?

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -154,4 +161,67 @@ object Features {
...
+      }
+    }
+
+    /** tries to extract the given feature name from the config, if successful returns its feature support */
+    def getFeature(config: Config, name: String): Option[FeatureSupport] = {
+      if(!config.hasPath(s"features.$name")) None
+      else {
+        config.getString(s"features.$name") match {
+          case "mandatory" => Some(Mandatory)
+          case "optional" => Some(Optional)
```
pm47, line 195: A bit over the top?
```suggestion
          case support if support == Mandatory.toString => Some(Mandatory)
          case support if support == Optional.toString => Some(Optional)
```

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
pm47: Let's see what the `ActivatedFeature` idea lead us to, but imo this `Resolution` class is a *lot* of code for what it's doing.

araspitzu:
Latest commit 1dd4ba5 introduces `ActivatedFeature` and `InactiveFeature`, the former is used to store the known features and the latter those that we don't know. It's necessary to store the unknown features as well because we want to validate the remote peer's features and doing something like:
```
 case class Features(...) {
    require(noUnknownEvenFeatures)
 }
```
seems quite ugly because it would fail when parsing the `Init` from remote and it's preferable to use the existing validation methods. 

I addressed most of the feedback in the last commit including the json serialization (now `Features` has its own serializer) and refactored the `Resolution` object into something smaller inside `Features`. 

I also took the chance to rename the invoice features to `PaymentRequestFeatures` as the naming caused a bit of confusion. Note that i added the `Features` inside `wire.Init` but not in all the other wire messages, let me know if we want to propagate this down to all the other messages (even though there isn't a big advantage in doing so).

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -147,6 +147,10 @@ object NodeParams {
       case (old, new_) => require(!config.hasPath(old), s"configuration key '$old' has been replaced by '$new_'")
     }
 
+    // since v0.3.5 features cannot be a byte vector (hex string)
```
t-bast, line 150:
```suggestion
    // since v0.4.1 features cannot be a byte vector (hex string)
```

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -147,6 +147,10 @@ object NodeParams {
...
+    val isFeatureByteVector = config.getValue("features").valueType() == ConfigValueType.STRING
+    require(!isFeatureByteVector, "configuration key 'features' have moved from bytevector to human readable (ex: 'feature-name' = optional/mandatory)")
```
t-bast, line 152: nit:

```suggestion
    require(config.getValue("features").valueType() != ConfigValueType.STRING, "configuration key 'features' have moved from bytevector to human readable (ex: 'feature-name' = optional/mandatory)")
```

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
```diff
@@ -191,19 +193,19 @@ class PeerConnectionSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike wi
...
     import f._
-    val remoteInit = wire.Init(bin"0000001010000000".bytes) // bob supports channel range queries and variable length onion
+    val remoteInit = wire.Init(Features(Set(ActivatedFeature(ChannelRangeQueries, Optional), ActivatedFeature(VariableLengthOnion, Optional))))
     connect(remoteNodeId, switchboard, router, connection, transport, peerConnection, peer, remoteInit, expectSync = true)
   }
 
   test("don't sync if whitelist doesn't contain peer", Tag("sync-whitelist-random")) { f =>
     import f._
-    val remoteInit = wire.Init(bin"0000001010000000".bytes) // bob supports channel range queries
+    val remoteInit = wire.Init(Features(bin"0000001010000000")) // bob supports channel range queries
```
t-bast, line 208: I think it's not this test's responsibility to test the feature parsing from bits, better make it more readable:

```suggestion
    val remoteInit = wire.Init(Features(Set(ActivatedFeature(ChannelRangeQueries, Optional))))
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartHandlerSpec.scala
```diff
@@ -43,9 +46,19 @@ import scala.concurrent.duration._
 
 class MultiPartHandlerSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike {
 
+  val features = Features(Set(
```
t-bast, line 49: nit:

```suggestion
  val featuresWithMpp = Features(Set(
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentInitiatorSpec.scala
```diff
@@ -50,8 +51,17 @@ class PaymentInitiatorSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike
...
   case class FixtureParam(nodeParams: NodeParams, initiator: TestActorRef[PaymentInitiator], payFsm: TestProbe, multiPartPayFsm: TestProbe, sender: TestProbe, eventListener: TestProbe)
 
+  val defaultTestFeatures = Features(Set(
+    ActivatedFeature(InitialRoutingSync, Optional),
+    ActivatedFeature(OptionDataLossProtect, Optional),
+    ActivatedFeature(ChannelRangeQueries, Optional),
+    ActivatedFeature(ChannelRangeQueriesExtended, Optional),
+    ActivatedFeature(VariableLengthOnion, Optional)))
+
+  val featuresWithMpp = Features(defaultTestFeatures.activated + ActivatedFeature(BasicMultiPartPayment, Optional))
```
t-bast, line 61:
```suggestion
  val featuresWithMpp = Features(defaultTestFeatures.activated + ActivatedFeature(PaymentSecret, Optional) + ActivatedFeature(BasicMultiPartPayment, Optional))
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
```diff
@@ -47,6 +48,8 @@ class PaymentPacketSpec extends AnyFunSuite with BeforeAndAfterAll {
 
   implicit val log: akka.event.LoggingAdapter = akka.event.NoLogging
 
+  val variableLengthOnionFeature = Features(Set(ActivatedFeature(VariableLengthOnion, Optional)))
```
t-bast, line 51: This should be in the object below, it's currently duplicated (see line 385)

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -105,7 +106,7 @@ class PeerConnection(nodeParams: NodeParams, switchboard: ActorRef, router: Acto
           // off basic_mpp and payment_secret. As long as they're provided in the invoice it's not an issue.
           // We use a long enough mask to account for future features.
           // TODO: remove that once eclair-mobile is patched.
-          val tweakedFeatures = BitVector.bits(nodeParams.features.bits.reverse.toIndexedSeq.zipWithIndex.map {
+          val tweakedFeatures = BitVector.bits(nodeParams.features.toByteVector.bits.reverse.toIndexedSeq.zipWithIndex.map {
```
t-bast, line 109: You can now make this code a lot more readable. Instead of modifying a `ByteVector`, just modify the `Features` object to remove the `basic_mpp` and `payment_secret` features from the set of activated features.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -38,11 +40,87 @@ sealed trait Feature {
...
+    case FeatureSupport.Mandatory => mandatory
+    case FeatureSupport.Optional => optional
+  }
+
   override def toString = rfcName
 }
 
+case class ActivatedFeature(feature: Feature, support: FeatureSupport)
+
+case class InactiveFeature(bitIndex: Int)
```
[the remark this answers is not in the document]
araspitzu: Went for `UnknownFeature` in 8840426

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
araspitzu: Not sure about this, isn't it too big to be inline?

t-bast: It's only the text that is long, not really the equality check, so it doesn't shock me. But if you prefer the current version don't hesitate to ignore!

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
araspitzu: Yes, this is done in 962d2a5 and it's been extracted into a function.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
araspitzu: Not too ugly but we'd have to add a new dependency for `scala-reflect`:
```scala
    import scala.reflect.runtime.universe._
    import scala.reflect.runtime.{ universe => runtime }
    val mirror = runtime.runtimeMirror(ClassLoader.getSystemClassLoader)
    val subclasses = typeOf[Feature].typeSymbol.asClass.knownDirectSubclasses
    lazy val knownFeatures = subclasses.map({ desc =>
      val mod = mirror.staticModule(desc.asClass.fullName)
      mirror.reflectModule(mod).instance
    })
```

araspitzu: As a compromise i added the test to guard us against forgetting new features but without using reflection at runtime. b2441ed

eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
```diff
@@ -73,17 +73,17 @@ object Announcements {
...
+    val witness = nodeAnnouncementWitnessEncode(timestamp, nodeSecret.publicKey, color, alias, features.toByteVector, nodeAddresses, unknownFields = ByteVector.empty)
     val sig = Crypto.sign(witness, nodeSecret)
     NodeAnnouncement(
       signature = sig,
       timestamp = timestamp,
       nodeId = nodeSecret.publicKey,
       rgbColor = color,
       alias = alias,
-      features = features,
+      features = features.toByteVector,
```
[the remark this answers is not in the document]
araspitzu: I wasn't sure if we wanted to type all of them so i started with the `Init`, with the latest changes (eab0654) i typed the remaining `ChannelAnnouncement` and `NodeAnnouncement`

PR #1386, opened 2020-04-21
fixup! Use sender instead of providing actor refs (#1379)

pm47:
When switching to a new connection while already connected, peer
immediately kills the current connection and sends back the
`PeerConnection.ConnectionReady` to itself. Since #1379, the sender of
this message is assumed to be the `PeerConnection` actor. If peer
doesn't preserve the sender by using a `forward` instead of a `tell`, it
will assume that itself is the `PeerConnection`, which will break
everything.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
@@ -16,9 +16,8 @@
 
 package fr.acinq.eclair.io
 
-import java.net.{InetAddress, InetSocketAddress, ServerSocket}
+import java.net.{InetAddress, ServerSocket}
 
-import akka.actor.FSM.{CurrentState, SubscribeTransitionCallBack, Transition}
 import akka.actor.Status.Failure
 import akka.testkit.{TestFSMRef, TestProbe}
 import com.google.common.net.HostAndPort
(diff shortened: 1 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -137,6 +136,31 @@ class PeerSpec extends TestkitBaseClass with StateTestsHelperMethods {
     probe.expectMsg("disconnecting")
   }
 
+  test("handle new connection in state CONNECTED") { f =>
```
t-bast, line 139: Nice test!

pm47: I think I can actually improve it a little bit.

pm47: Done with b25a040. Testing only via the deatch watch was pretty indirect.

pm47: Actually that's the reason why I had to do two rounds (because I needed to check that `peerConnection2` was killed). I could get rid of `peerConnection3` now. Well, let's leave it that way.

t-bast: I agree, better to keep `peerConnection3` it makes it even more explicit.

PR #1387, opened 2020-04-22
Test preimage extraction from mempool

t-bast:
Instead of waiting for htlc-success txs to be confirmed, eclair also looks at mempool txs to detect preimages as soon as possible.

This has been the case for a very long time, but our integration tests didn't showcase this correctly.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/WatcherSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/WatcherSpec.scala
@@ -16,28 +16,96 @@
 
 package fr.acinq.eclair.blockchain
 
-import fr.acinq.bitcoin.Transaction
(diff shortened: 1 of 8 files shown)


pm47:
We should test both cases: a tx may appear in a block that we haven't seen in the mempool.

t-bast:
> We should test both cases: a tx that we haven't seen in the mempool may appear in a block.

Indeed. But I think we won't be able to test that in the IntegrationSpec.
I'll check what we already have in the non E2E tests, we probably already test that.

t-bast:
I added tests (and test helpers) to `ZmqWatcher` in https://github.com/ACINQ/eclair/pull/1387/commits/09a308d6bc9a471bbfc618ec447c3618f04a1d89. The `ClosingStateSpec` already has tests for preimage extraction from `WatchEventSpent`s so we should be covering most of it now.

PR #1389, opened 2020-04-22
Prepare upgrade to scala 2.13 and akka 2.6

pm47:
We do all the changes that will be required and are already possible to
minimize the diff:
- update dependencies
- `'something` -> `Symbol("something")`
- `BigDecimal.xValue()` -> `BigDecimal.xValue`
- `Map.filterKeys` -> `Map.filterKeys.toMap` (same for `Map.mapValues`)
- `def myMethod(...)` -> `def myMethod(...): Unit`


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2287,7 +2287,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
   case class MyState(state: FSM.State[fr.acinq.eclair.channel.State, Data]) {
 
-    def storing(): FSM.State[fr.acinq.eclair.channel.State, Data] = {
+    def storing(unused: Unit = ()): FSM.State[fr.acinq.eclair.channel.State, Data] = {
```
t-bast, line 2290: This one is weird, any idea why that is needed?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -722,7 +722,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         case Failure(cause) => handleLocalError(cause, d, Some(fee))
       }
 
-    case Event(c@CMD_SIGN, d: DATA_NORMAL) =>
+    case Event(c: Command, d: DATA_NORMAL) if c == CMD_SIGN =>
```
t-bast, line 725: The following doesn't work?
```suggestion
    case Event(CMD_SIGN, d: DATA_NORMAL) =>
```

eclair-core/src/test/scala/fr/acinq/eclair/router/BaseRouterSpec.scala
```diff
@@ -103,7 +103,8 @@ abstract class BaseRouterSpec extends TestkitBaseClass {
       // let's set up the router
       val peerConnection = TestProbe()
       val watcher = TestProbe()
-      val nodeParams = Alice.nodeParams
+      import com.softwaremill.quicklens._
+      val nodeParams = Alice.nodeParams.modify(_.routerConf.routerBroadcastInterval).setTo(1 day) // "disable" auto rebroadcast
```
t-bast, line 107: Why is that change done here?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: That would force us to duplicate `CMD_SIGN` on line 753:

```scala
  case Failure(cause) => handleCommandError(cause, CMD_SIGN)
```

eclair-core/src/test/scala/fr/acinq/eclair/router/BaseRouterSpec.scala
pm47: I don't know if tests run faster or in a different order with scala 2.13/akka 2.6, but a race condition appeared.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47:
```
Error:(1992, 34) can't supply unit value with infix notation because nullary method storing: ()akka.actor.FSM.State[fr.acinq.eclair.channel.State,fr.acinq.eclair.channel.Data] takes no arguments; use dotted invocation instead: Channel.this.state2mystate(Channel.this.goto(CLOSING).using(nextData)).storing()
    goto(CLOSING) using nextData storing() calling (doPublish(closingTx))
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -760,7 +760,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
     case Event(commit: CommitSig, d: DATA_NORMAL) =>
       Commitments.receiveCommit(d.commitments, commit, keyManager) match {
-        case Success((commitments1, revocation)) =>
+        case Success((commitments1: Commitments, revocation)) =>
```
pm47, line 763: @t-bast this is frankly even weirder

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/GUIUpdater.scala
```diff
@@ -214,15 +214,15 @@ class GUIUpdater(mainController: MainController) extends Actor with ActorLogging
       log.debug(s"payment sent with h=${p.paymentHash}, amount=${p.recipientAmount}, fees=${p.feesPaid}")
       val message = CoinUtils.formatAmountInUnit(p.amountWithFees, FxApp.getUnit, withUnit = true)
       mainController.handlers.notification("Payment Sent", message, NOTIFICATION_SUCCESS)
-      runInGuiThread(() => mainController.paymentSentList.prepend(PaymentSentRecord(p, LocalDateTime.now())))
+      runInGuiThread(() => mainController.paymentSentList.asScala.prepend(PaymentSentRecord(p, LocalDateTime.now())))
```
pm47, line 217: Is the prepend really necessary @dpad85?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Got it, so it's to maintain the way we call that with a space, sounds ok

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: True but I think it makes the pattern matching more obvious. And inside this handler you have to know that what you're handling is `CMD_SIGN` so I don't find it too shocking

eclair-core/src/test/scala/fr/acinq/eclair/router/BaseRouterSpec.scala
t-bast: Damn, faster tests, I'm all in for that! But when will I have time for coffee breaks now?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Yeah I really don't get why it thinks it's not obvious that this is a `Commitments`...

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: I disagree, duplicating `CMD_SIGN` can only lead ton inconsistency. Let me try to find a better way.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Nevermind, this isn't necessary, removed with 8c6302a.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: I can't find better. Here is my argument for keeping it that way:
- in my proposal, if someone accidentally removes `if c == CMD_SIGN`, then tests will fail because the pattern will match all commands
- in your proposal, if someone accidentally copy-pastes `handleCommandError(cause, CMD_SIGN)`, then compilation and tests will run just fine

t-bast: > if someone accidentally copy-pastes handleCommandError(cause, CMD_SIGN), then compilation and tests will run just fine

That's true, ideally a test should catch that but we can't test everything...

t-bast: https://github.com/scala/bug/issues/11574

pm47: Awesome find! "variable-binding pattern" is the key word I was looking for. hasn't been added to the release note AFAICT.

pm47: Done with a7ddc2fd7.

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/GUIUpdater.scala
dpad85: No it's not necessary and only affects the order of the payments sent (or relayed/received) in the activity tab in the gui. With `prepend`, new events are on top in the lists.

pm47: Done with 263d96ae1.

pm47:
rebased

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1063,7 +1063,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         case Failure(cause) => handleLocalError(cause, d, Some(fee))
       }
 
-    case Event(c@CMD_SIGN, d: DATA_SHUTDOWN) =>
+    case Event(c: Command, d: DATA_SHUTDOWN) if c == CMD_SIGN =>
```
t-bast, line 1066: I think we can do `case Event(c: CMD_SIGN.type, d: DATA_SHUTDOWN)` even in Scala 2.11, why not apply this commit to this PR?

pm47: Fixed with 53a10ff.

PR #1390, opened 2020-04-22
Update to scala 2.13 and akka 2.6 (incremental)

pm47:
This builds on #1389, so the changeset is minimal (disregard the first commit).

Main changes:
- relaxed compiler parameters to minimize impact (e.g. allow
deprecated features)
- `scala.collection.JavaConverters` -> `scala.jdk.CollectionConverters`
- `MultiMap` -> `MultiDict`

Compilation is 25% faster on my machine, compiler is a bit more strict
(it found an "invalid comparison" bug).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
 dist: trusty
 language: scala
 scala:
- - 2.11.12
+ - 2.13.1
 env:
  - export LD_LIBRARY_PATH=/usr/local/lib
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
 dist: trusty
 language: scala
 scala:
- - 2.13.1
+ - 2.13.2
 env:
  - export LD_LIBRARY_PATH=/usr/local/lib
 before_install:
--- a/eclair-core/pom.xml
+++ b/eclair-core/pom.xml
@@ -20,11 +20,11 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>fr.acinq.eclair</groupId>
-        <artifactId>eclair_2.11</artifactId>
+        <artifactId>eclair_2.13</artifactId>
         <version>0.3.5-SNAPSHOT</version>
     </parent>
 
-    <artifactId>eclair-core_2.11</artifactId>
+    <artifactId>eclair-core_2.13</artifactId>
     <packaging>jar</packaging>
 
     <name>${project.artifactId}</name>
@@ -115,6 +115,11 @@
     </profiles>
 
     <dependencies>
+        <dependency>
+            <groupId>org.scala-lang.modules</groupId>
+            <artifactId>scala-collection-contrib_${scala.version.short}</artifactId>
+            <version>0.2.1</version>
+        </dependency>
         <!-- AKKA -->
         <dependency>
             <groupId>com.typesafe.akka</groupId>
@@ -170,7 +175,7 @@
         <dependency>
             <groupId>org.scodec</groupId>
             <artifactId>scodec-core_${scala.version.short}</artifactId>
-            <version>1.11.2</version>
+            <version>1.11.7</version>
         </dependency>
         <dependency>
             <groupId>commons-codec</groupId>
--- a/eclair-core/src/main/scala/fr/acinq/eclair/package.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/package.scala
@@ -174,6 +174,7 @@ package object eclair {
     override def toFloat(x: MilliSatoshi): Float = x.toLong
     override def toDouble(x: MilliSatoshi): Double = x.toLong
     override def compare(x: MilliSatoshi, y: MilliSatoshi): Int = x.compare(y)
+    override def parseString(str: String): Option[MilliSatoshi] = ???
     // @formatter:on
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
@@ -150,7 +150,7 @@ object ChannelRelayer {
       case Some(nextNodeId) =>
         log.debug(s"next hop for htlc #{} is nodeId={}", add.id, nextNodeId)
         // then we retrieve all known channels to this node
-        val allChannels = node2channels.getOrElse(nextNodeId, Set.empty[ShortChannelId])
+        val allChannels = node2channels.get(nextNodeId)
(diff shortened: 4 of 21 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -150,7 +150,7 @@ object ChannelRelayer {
       case Some(nextNodeId) =>
         log.debug(s"next hop for htlc #{} is nodeId={}", add.id, nextNodeId)
         // then we retrieve all known channels to this node
-        val allChannels = node2channels.getOrElse(nextNodeId, Set.empty[ShortChannelId])
+        val allChannels = node2channels.get(nextNodeId)
```
t-bast, line 153: Isn't that a regression if `nextNodeId` isn't found?

pom.xml
```diff
@@ -64,12 +64,12 @@
...
-        <scala.version.short>2.11</scala.version.short>
-        <akka.version>2.4.20</akka.version>
-        <akka.http.version>10.0.11</akka.http.version>
+        <scala.version>2.13.1</scala.version>
+        <scala.version.short>2.13</scala.version.short>
+        <akka.version>2.6.4</akka.version>
+        <akka.http.version>10.1.11</akka.http.version>
         <sttp.version>1.7.2</sttp.version>
-        <bitcoinlib.version>0.17</bitcoinlib.version>
+        <bitcoinlib.version>0.18-SNAPSHOT</bitcoinlib.version>
```
t-bast, line 72: Placeholder comment so that we don't forget to update that to a release version of bitcoinlib

pom.xml
```diff
@@ -132,16 +132,12 @@
                 <version>3.4.2</version>
                 <configuration>
                     <args combine.children="append">
-                        <arg>-deprecation</arg>
+                        <!--arg>-Xlint:deprecation</arg-->
```
t-bast, line 135: Maybe create a tracking issue to revert that back in the future (once all warnings are fixed)?

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -174,6 +174,7 @@ package object eclair {
     override def toFloat(x: MilliSatoshi): Float = x.toLong
     override def toDouble(x: MilliSatoshi): Double = x.toLong
     override def compare(x: MilliSatoshi, y: MilliSatoshi): Int = x.compare(y)
+    override def parseString(str: String): Option[MilliSatoshi] = ???
```
pm47, line 177: I don't think it's a problem to leave this unimplemented.

eclair-core/pom.xml
```diff
@@ -115,6 +115,11 @@
     </profiles>
 
     <dependencies>
+        <dependency>
+            <groupId>org.scala-lang.modules</groupId>
+            <artifactId>scala-collection-contrib_${scala.version.short}</artifactId>
+            <version>0.2.1</version>
+        </dependency>
```
pm47, line 122: This is required for MultiDict because MultiMap has been removed.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
pm47: Note that `node2channels` is now a `MultiDicts`. Here is the signature of this method:

```scala
  def get(key: K): Set[V] = sets.get(key).getOrElse(Set.empty)
```

pom.xml
pm47: The reason why I'm not fixing the warning is that it would break source compatibility with scala 2.11.

t-bast: Agreed, but we probably want to get rid of it "at some point" (once the `android` branch is gone), just want to make sure we don't forget it (we're not looking at the pom very often to be honest)

pm47: > we're not looking at the pom very often to be honest

Alright. I wouldn't worry about it though, my OCD is already in overdrive with those:

image

pm47:
Rebased.

pom.xml
pm47: Agreed. @sstone 🙏

sstone: Yes I'll add a comment when it's available on maven central

.travis.yml
```diff
@@ -4,7 +4,7 @@ services:
 dist: trusty
 language: scala
 scala:
- - 2.11.12
+ - 2.13.1
```
t-bast, line 7: Why not even 2.13.2? #reckless?

eclair-node/pom.xml
```diff
@@ -137,6 +137,12 @@
             <version>${akka.http.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
t-bast, line 140: Why is this new dependency needed?

pom.xml
sstone: bitcoin-lib_2.13 0.18 is now available on maven central

.travis.yml
pm47: Nice catch. Main pom.xml was also using 2.13.1. Fixed in ed5f7c707.

eclair-node/pom.xml
pm47: It's required at runtime during tests by akka http, for some reason. If omitted I get this:

```
 Cause: java.lang.NoClassDefFoundError: akka/stream/testkit/TestSubscriber$
  at akka.http.scaladsl.testkit.WSProbe$$anon$1.<init>(WSProbe.scala:107)
  at akka.http.scaladsl.testkit.WSProbe$.apply(WSProbe.scala:106)
  at fr.acinq.eclair.api.ApiServiceSpec.$anonfun$new$58(ApiServiceSpec.scala:469)
  at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
  at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
  at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
  at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
  at org.scalatest.Transformer.apply(Transformer.scala:22)
  at org.scalatest.Transformer.apply(Transformer.scala:20)
  at org.scalatest.funsuite.AnyFunSuiteLike$$anon$1.apply(AnyFunSuiteLike.scala:189)
```

pm47:
Rebased.

pom.xml
pm47: Done with https://github.com/ACINQ/eclair/issues/1393.

PR #1391, opened 2020-04-24
 Revert "Use sender instead of providing actor refs (#1379)"

pm47:
It was error-prone (cf bff833b1b42c9c18a96e40fe130c6ab2aefce555) and not a good practice according to @sstone.

Commit 2ad91b2 is optional, let me know what you think.


t-bast:
I like the revert, but not sure about https://github.com/ACINQ/eclair/pull/1391/commits/2ad91b22a7218fe4838a926806a1acba71cdd82d
While it simplifies a bit the code, `PeerRoutingMessage`s are not necessarily gossip, so if we want to make those evolve independently (add new fields) we'll likely need to revert that.
Not a strong NACK, but not a strong ACK either :)

pm47:
> I like the revert, but not sure about 2ad91b2
While it simplifies a bit the code, PeerRoutingMessages are not necessarily gossip, so if we want to make those evolve independently (add new fields) we'll likely need to revert that.
Not a strong NACK, but not a strong ACK either :)

I reverted with 13ea541 (insert 'yo_dawg.jpg') but kept the `LightningMessage`->`RoutingMessage` signature change.

PR #1394, opened 2020-04-27
Downgrade to scala 2.13.1

pm47:
Latest scala version has a bug [1] that breaks build on windows.

[1] https://github.com/scala/bug/issues/11955


pm47:
Not sure how I got the branch name *that* wrong.

PR #1395, opened 2020-04-27
Use channel balance in path-finding

t-bast:
This is a step towards more flexibility for path-finding in the router.
I started wanting to simply add support for channel balance, but as I dived into the code I couldn't help myself and refactored, renamed, improved documentation and fixed some weirdness that I stumbled upon.
I hope the resulting code is easier to read and reason about, but I'm obviously biased so let me know how I can improve.

I split that work into three commits; it may help to review commit-by-commit, or to look at the resulting code instead of looking at the diff.

If you're not familiar with Yen's algorithm, [the example on the wikipedia page](https://en.wikipedia.org/wiki/Yen%27s_algorithm#Example) should make it easy to understand.

## Taking channel capacity into account

The [first commit](https://github.com/ACINQ/eclair/commit/54af0b3544ecf09324b2e3eca046492ce84e23ae) is quite small.

It updates the path finding algorithm to use channel capacity instead of htlcMaximumMsat.
It also takes into account channel balance when available and excludes channels that don't have enough funds to relay the payment.

This change also fixes an off-by-one error in weight computation: we were incorrectly applying a channel's fee to the amount that needs to be relayed through that channel (whereas this is instead what the node needs to receive to collect enough fee *before* relaying).

## Refactoring the graph file

In the [second commit](https://github.com/ACINQ/eclair/commit/9d655c550a916606daf5d2cf5e9ec452ab287e8c), I heavily refactor the graph file without introducing functional changes (except for very minor tweaks). This is the one that's somewhat hard to review, but I hope the result makes it easier to understand for people who don't know the code.

## Simplify path-finding implementations

In the [third commit](https://github.com/ACINQ/eclair/commit/da73e819e86071b1ceff74ee906f67b767651b17), I fix a couple confusing steps in the implementation of Yen's algorithm.

The first one was the computation of the `edgesToIgnore` and the specific handling of the case i = 0. This specific case wasn't needed and made the code a bit hard to read.

The second one was the weight provided to dijkstra for spur paths.
The weight of the root path was applied to the target node.
It was probably an attempt to take into account the fact that dijkstra wasn't computing a complete path and that fees may not match, but it couldn't really work.
I removed that and added a fee check at the end of the path-finding.

## Tests

I did an apple-to-apple comparison between the new and old implementation, by selecting two nodes at random from the mainnet graph and trying to find 3 routes (Yen's algorithm) between them for different amounts. When the results of the old and new algorithm didn't match I manually checked the results.

The new algorithm consistently finds more routes and cheaper ones. The route prefix are more diverse, which is good as well (especially for MPP). The only cases where the old implementation was finding routes that the new implementation didn't find are cases where that route was in fact invalid, because one of the nodes was missing an `htlcMaximumMsat` and we didn't correctly take into account its capacity.

I also run performance benchmarks between the two algorithms, and the new one is consistently 25% faster on my machine (when looking for 3 routes).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -210,14 +210,12 @@ object Graph {
       if (current.key != sourceNode) {
         val currentWeight = bestWeights.get(current.key) // NB: there is always an entry for the current in the 'weights' map
         // build the neighbors with optional extra edges
-        val currentNeighbors = extraEdges.isEmpty match {
-          case true => g.getIncomingEdgesOf(current.key)
-          case false =>
-            val extraNeighbors = extraEdges.filter(_.desc.b == current.key)
-            // the resulting set must have only one element per shortChannelId; we prioritize extra edges
-            g.getIncomingEdgesOf(current.key).filterNot(e => extraNeighbors.exists(_.desc.shortChannelId == e.desc.shortChannelId)) ++ extraNeighbors
+        val neighborEdges = {
+          val extraNeighbors = extraEdges.filter(_.desc.b == current.key)
+          // the resulting set must have only one element per shortChannelId; we prioritize extra edges
+          g.getIncomingEdgesOf(current.key).filterNot(e => extraNeighbors.exists(_.desc.shortChannelId == e.desc.shortChannelId)) ++ extraNeighbors
         }
-        currentNeighbors.foreach { edge =>
+        neighborEdges.foreach { edge =>
           val neighbor = edge.desc.a
           // NB: this contains the amount (including fees) that will need to be sent to `neighbor`, but the amount that
           // will be relayed through that edge is the one in `currentWeight`.
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -98,61 +98,63 @@ object Graph {
                         boundaries: RichWeight => Boolean): Seq[WeightedPath] = {
     // find the shortest path (k = 0)
     val targetWeight = RichWeight(amount, 0, CltvExpiryDelta(0), 0)
-    dijkstraShortestPath(graph, sourceNode, sourceNode, targetNode, ignoredEdges, ignoredVertices, extraEdges, targetWeight, boundaries, currentBlockHeight, wr) match {
-      case Nil => Seq.empty // if we can't even find a single path, avoid returning a Seq(Seq.empty)
-      case shortestPath =>
-        var allSpurPathsFound = false
-        val shortestPaths = new mutable.ArrayDeque[WeightedPath]
-        shortestPaths += WeightedPath(shortestPath, pathWeight(sourceNode, shortestPath, amount, currentBlockHeight, wr))
-        // stores the candidates for the k-th shortest path, sorted by path cost
-        val candidates = new mutable.PriorityQueue[WeightedPath]
-        // main loop
-        for (k <- 1 until pathsToFind) {
-          if (!allSpurPathsFound) {
-            val prevShortestPath = shortestPaths(k - 1).path
-            // for every edge in the path, we will try to find a different path after that edge
-            for (i <- prevShortestPath.indices) {
-              // select the spur node as the i-th element of the previous shortest path
-              val spurNode = prevShortestPath(i).desc.a
-              // select the sub-path from the source to the spur node
-              val rootPathEdges = prevShortestPath.take(i)
-              // we ignore all the paths that we have already fully explored in previous iterations
-              // if for example the spur node is B, and we already found shortest paths starting with A-B-C and A-B-D,
-              // we want to ignore the B-C and B-D edges
-              //          +-- C -- [...]
-              //          |
-              // A -- B --+-- D -- [...]
-              //          |
-              //          +-- E -- [...]
-              val alreadyExploredEdges = shortestPaths.collect { case p if p.path.take(i) == rootPathEdges => p.path(i).desc }.toSet
-              // we also want to ignore any link that can lead back to the previous node (we only want to go forward)
-              val returningEdges = rootPathEdges.lastOption.map(last => graph.getEdgesBetween(last.desc.b, last.desc.a).map(_.desc).toSet).getOrElse(Set.empty)
-              // find the "spur" path, a sub-path going from the spur node to the target avoiding previously found sub-paths
-              val spurPath = dijkstraShortestPath(graph, sourceNode, spurNode, targetNode, ignoredEdges ++ alreadyExploredEdges ++ returningEdges, ignoredVertices, extraEdges, targetWeight, boundaries, currentBlockHeight, wr)
-              if (spurPath.nonEmpty) {
-                // candidate k-shortest path is made of the root path and the new spur path, but the cost of the spur
-                // path is likely higher than previous shortest paths, so we need to validate that the root path can
-                // relay the increased amount.
-                val completePath = rootPathEdges ++ spurPath
-                val candidatePath = WeightedPath(completePath, pathWeight(sourceNode, completePath, amount, currentBlockHeight, wr))
-                if (boundaries(candidatePath.weight) && !shortestPaths.contains(candidatePath) && !candidates.exists(_ == candidatePath) && validatePath(completePath, amount)) {
-                  candidates.enqueue(candidatePath)
-                }
-              }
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -91,209 +96,183 @@ object Graph {
...
-              // update the visiting tree
-              prev.put(neighbor, edge)
-
-              // update the queue
-              vertexQueue.insert(WeightedNode(neighbor, newMinimumKnownWeight)) // O(1)
-
-              // update the minimum known distance array
-              weight.put(neighbor, newMinimumKnownWeight)
+  private def dijkstraShortestPath(g: DirectedGraph,
+                                   sender: PublicKey,
```
araspitzu, line 176: Why do we need this? Dijkstra computes paths from `source -> target` and we shouldn't need anything else.
When we are computing the shortest path then `sender == sourceNode` and we don't need the sender
When computing partial paths then `sender != sourceNode` but we're indeed looking for a path that does **not** go through the sender, hence we never fall into the case of applying 0-fee for the channel because its our own.
WDYT?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -91,209 +96,183 @@ object Graph {
...
+              // we don't use "getOrDefault" because it is not available in JDK7
+              val previousNeighborWeight = bestWeights.get(neighbor) match {
+                case null => RichWeight(MilliSatoshi(Long.MaxValue), Int.MaxValue, CltvExpiryDelta(Int.MaxValue), Double.MaxValue)
+                case w => w
+              }
+              // if this path between neighbor and the target has a shorter distance than previously known, we select it
+              if (neighborWeight.weight < previousNeighborWeight.weight) {
+                // update the best edge for this vertex
+                bestEdges.put(neighbor, edge)
+                // continue exploring from this updated node
```
araspitzu, line 236: Disagree :) , adding the new weight for this node won't necessarily result in continuing the exploration of the search space from it, the next iteration will pop off the queue (which is actually a fibonacci heap) the node with the smallest distance from the target and it might not be `neighbor`

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: This is to replace the `isNeighborTarget` and `isPartial` parameters that were previously used. I found them confusing and there was one place where they were incorrectly used...

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: Good catch, that comment isn't accurate

t-bast:
> I'm not sure about the refactoring done in 9d655c5 when this code was written i chose a slightly more imperative style because i think it's more readable and easier to understand when comparing it to pseudocode, i think the proposed refactoring loses a bit of readability but i'm not strongly opposed to it.

That's a reasonable argument indeed, the thing that lead me to that refactoring was mostly the `return`, because they have a quite surprising and not-so-expected behavior. I wanted to get rid of them and that slowly led me to that. But I don't mind going back to something more imperative, let's see what @sstone prefers to settle it!

> It's a way to make sure we don't find spur paths that exceeds the parameters for path-finding, i think with the proposed changes there is a regression: the proposed change might find an (invalid) spur path that is outside the boundaries parameters (i.e too hi CLTV) because we don't consider the previous part of the route (the root path); however the current code could find an alternative spur path that is coherent with the boundaries because it has knowledge of the total CLTV of the route (the total CLTV of the root path + the CLTV of the spur path).

This is handled by the added call to `validatePath`. Applying the weight like was done before was on the contrary incorrect because of %-based fees. In a route `A -> B -> C -> D -> E` you can't use the weight of `A -> B -> C` and apply it to `E` to find `C -> D -> E`, the percentage fee will make you reject some valid spur paths for no reason. I think it's better to do a "normal" dijkstra to find `C -> D -> E` with the right final amount, then validate that concatenating both results in a valid path (and I find that the code is clearer).

However you're right that we may want `validatePath` to check the boundaries as well, I should add that.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: It's true that the dijkstra calls for spur paths may implicitly not need this, but I think it's much simpler to provide it so that it's obvious that `addEdgeWeight` and `pathWeight` do the right thing and exclude the fee for the first hop (and only that first hop).

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1395/commits/702493f329b4f60800fb20054bd835df40b16264

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -302,56 +283,75 @@ object Graph {
       val ageFactor = normalize(channelBlockHeight, min = currentBlockHeight - BLOCK_TIME_TWO_MONTHS, max = currentBlockHeight)
 
       // Every edge is weighted by channel capacity, larger channels add less weight
-      val edgeMaxCapacity = edge.update.htlcMaximumMsat.getOrElse(CAPACITY_CHANNEL_LOW)
+      val edgeMaxCapacity = edge.capacity.toMilliSatoshi
```
sstone, line 286: Good catch!

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -293,7 +293,10 @@ object Validation {
       } else if (pc.getChannelUpdateSameSideAs(u).exists(_.timestamp >= u.timestamp)) {
         log.debug("ignoring {} (duplicate)", u)
         sendDecision(origins, GossipDecision.Duplicate(u))
-        d
+        // NB: we update the graph because the balances may have changed even if the channel_update is the same.
```
sstone, line 296: But only it is a local channel update ?

t-bast: If hesitated to add the check, you're right that it would be cleaner, I'll update that.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -171,129 +158,123 @@ object Graph {
...
-      val current = vertexQueue.deleteMin().getKey // O(log(n))
-
+    while (!toExplore.isEmpty && !targetFound) {
+      // node with the smallest distance from the target
+      val current = toExplore.deleteMin().getKey // O(log(n))
       if (current.key != sourceNode) {
-
+        val currentWeight = bestWeights.get(current.key) // NB: there is always an entry for the current in the 'weights' map
         // build the neighbors with optional extra edges
         val currentNeighbors = extraEdges.isEmpty match {
```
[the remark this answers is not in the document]
t-bast: Good catch, it's going to be easier to read without the match!

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1395/commits/1d0bb60c23f6aa70f970adbaa1ab8358395d55b9

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1395/commits/1d0bb60c23f6aa70f970adbaa1ab8358395d55b9

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -171,129 +158,123 @@ object Graph {
...
-
-      // node with the smallest distance from the source
-      val current = vertexQueue.deleteMin().getKey // O(log(n))
-
+    while (!toExplore.isEmpty && !targetFound) {
+      // node with the smallest distance from the target
+      val current = toExplore.deleteMin().getKey // O(log(n))
       if (current.key != sourceNode) {
-
+        val currentWeight = bestWeights.get(current.key) // NB: there is always an entry for the current in the 'weights' map
```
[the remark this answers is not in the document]
t-bast: Thanks, done in https://github.com/ACINQ/eclair/pull/1395/commits/272359b5765181d799d217536b6078fb2dbe9fc3

Issue #1396, opened 2020-04-27
Invalid comparison in `ElectrumWallet`


pm47:
https://github.com/ACINQ/eclair/blob/22d476774d7145150a6275acf7cd9385dcae9177/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWallet.scala#L210

PR #1398, opened 2020-04-28
Fix issue #1396 (invalid header comparison)

sstone:


eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWalletSimulatedClientSpec.scala
```diff
@@ -89,30 +101,24 @@ class ElectrumWalletSimulatedClientSpec extends TestKit(ActorSystem("test")) wit
...
-
   test("tell wallet is ready when a new block comes in, even if nothing else has changed") {
+    val f = fixture
+    import f._
+
     val last = wallet.stateData.blockchain.tip
     assert(last.header == headers.last)
     val header = makeHeader(last.header)
-    headers = headers :+ header
+    //headers = headers :+ header
```
t-bast, line 111: Hmm, shouldn't this be uncommented?

t-bast:
Don't we also need to apply this on master?

pm47:
> Don't we also need to apply this on master?

Definitely!

sstone:
The faulty header comparison as already been fixed on master (as part of the switch to scala 2.13). I'm not even sure we should keep the Electrum code in master now ?

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWalletSimulatedClientSpec.scala
sstone: This was useless, I've removed it now

t-bast:
> I'm not even sure we should keep the Electrum code in master now ?

Good point, in the long term we probably want to remove it (the `Setup` code will be simpler, the configuration too, and we'll be making the codebase smaller) but I think for now it's safer to keep it until we delete the android branch?

PR #1399, opened 2020-04-29
Disable backup handler in tests

t-bast:
We've been witnessing random test suites freezes (since ages).
We've observed that when these freezes happen, there are usually a lot of "too many open files" errors raised by the OS.

The backup handler is a likely culprit as the IntegrationSpec is running multiple nodes and exchanging HTLCs at a fast rate.

At least it won't hurt disabling it in tests, and will speed up the test suite.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -21,6 +21,9 @@ blocks:
       jobs:
         - name: Build & Test
           commands:
+            # configure file limits
+            - ulimit -S -n 1024000
+            - echo "fs.file-max = 1024000" | sudo tee -a /etc/sysctl.conf
             - sem-version java 11
             - checkout
             - cache restore maven
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -15,6 +15,7 @@ eclair {
     password = "" // password for basic auth, must be non empty if json-rpc api is enabled
   }
 
+  enable-db-backup = true // enable the automatic sqlite db backup; do not change this unless you know what you are doing
   // override this with a script/exe that will be called everytime a new database backup has been created
   # backup-notify-script = "/absolute/path/to/script.sh"
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -66,6 +66,7 @@ import scala.concurrent.duration._
  * @param datadir          directory where eclair-core will write/read its data.
  * @param overrideDefaults use this parameter to programmatically override the node configuration .
  * @param seed_opt         optional seed, if set eclair will use it instead of generating one and won't create a seed.dat file.
+ * @param db               optional databases to use, if not set eclair will create the necessary databases
  */
 class Setup(datadir: File,
             overrideDefaults: Config = ConfigFactory.empty(),
@@ -117,8 +118,10 @@ class Setup(datadir: File,
   val feeratesPerKw = new AtomicReference[FeeratesPerKw](null)
 
   val feeEstimator = new FeeEstimator {
+    // @formatter:off
     override def getFeeratePerKb(target: Int): Long = feeratesPerKB.get().feePerBlock(target)
     override def getFeeratePerKw(target: Int): Long = feeratesPerKw.get().feePerBlock(target)
+    // @formatter:on
   }
 
   val nodeParams = NodeParams.makeNodeParams(config, keyManager, initTor(), database, blockCount, feeEstimator)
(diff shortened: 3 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -66,11 +66,14 @@ import scala.concurrent.duration._
...
  * @param seed_opt         optional seed, if set eclair will use it instead of generating one and won't create a seed.dat file.
+ * @param db               optional databases to use, if not set eclair will create the necessary databases
+ * @param disableBackup    disable automatic database backup (to speed up tests)
  */
 class Setup(datadir: File,
             overrideDefaults: Config = ConfigFactory.empty(),
             seed_opt: Option[ByteVector] = None,
-            db: Option[Databases] = None)(implicit system: ActorSystem) extends Logging {
+            db: Option[Databases] = None,
+            disableBackup: Boolean = false)(implicit system: ActorSystem) extends Logging {
```
pm47, line 76: Why not make that a configuration parameter?

t-bast: That was my first idea, but in fact we really don't want users shooting themselves in the foot by disabling backup...that's why I only put it in the code itself, but if you think it's better in the conf I can do that.

t-bast:
Allright one of the test runs failed, so it's clearly not the only culprit...
Back to digging into files...

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
pm47: I think it's better for consistency, with a "*do not change this unless you know what you are doing*" comment. There are other parameters that can be used as footgun

eclair-core/src/main/resources/reference.conf
```diff
@@ -15,6 +15,7 @@ eclair {
     password = "" // password for basic auth, must be non empty if json-rpc api is enabled
   }
 
+  disable-db-backup = false // disable the automatic sqlite db backup; do not change this unless you know what you are doing
```
pm47, line 18: Did you do the double negative on purpose? I think it's better to avoid it

```suggestion
  enable-db-backup = true // enable the automatic sqlite db backup; *do not change this unless you know what you are doing*
```

t-bast: I wanted to make it more explicit that the default behavior is to have it `enabled`, and we only provide an option to `disable` it. But it's nit, I don't have a strong opinion.

pm47: Yeah I figured you had that in mind. But even knowing it, I had to think twice before deciding whether backups would be on or off by default. I think the other way around is better.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -270,17 +273,20 @@ class Setup(datadir: File,
...
-          nodeParams.db,
-          new File(chaindir, "eclair.sqlite.bak"),
-          if (config.hasPath("backup-notify-script")) Some(config.getString("backup-notify-script")) else None
-        ), "backuphandler", SupervisorStrategy.Resume))
+      backupHandler = if (config.getBoolean("enable-db-backup")) {
+        system.actorOf(SimpleSupervisor.props(
+          BackupHandler.props(
+            nodeParams.db,
+            new File(chaindir, "eclair.sqlite.bak"),
+            if (config.hasPath("backup-notify-script")) Some(config.getString("backup-notify-script")) else None
```
pm47, line 284: Have you noticed that for the `backup-notify-script` parameter (which is also sensitive), we chose to not document it at all? Could have used the same pattern for `enable-db-backup`. And in that case, using the negative would make more sense...

.semaphore/semaphore.yml
```diff
@@ -21,6 +21,9 @@ blocks:
       jobs:
         - name: Build & Test
           commands:
+            # configure file limits
+            - ulimit -S -n 1024000
+            - echo "fs.file-max = 1024000" | sudo tee -a /etc/sysctl.conf
```
pm47, line 26: Now that we have fixed the test issues in #1400 I wonder if this is really necessary. If it is, why do the test pass? Shouldn't the "too many open files" error completely break the tests?

t-bast: > Shouldn't the "too many open files" error completely break the tests?

It looks like it does, but it makes our test suite hang so we don't cleanly witness it (because we kill the build after 15 minutes to avoid wasting resources).

> I wonder if this is really necessary

I can't be entirely sure, but since I added that I haven't repro-ed the hang on semaphore (I restarted the test suite several times), so it looks like it happens less often (hopefully it doesn't happen anymore). At least it can't hurt to raise this limit (the ulimit is already at 1024000 by default in semaphore, but it wasn't configured in sysctl.conf).

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
t-bast: It is documented in `reference.conf`: `// override this with a script/exe that will be called everytime a new database backup has been created`...
I don't have a strong opinion honestly, let me know what you prefer :)

pm47: Right 🤦‍♂️ . Nevermind.

.semaphore/semaphore.yml
pm47: Ok, so this fixes the intermittent hanging while #1400 introduced then fixed a permanent hanging.

PR #1400, opened 2020-04-29
Delegate client creation to a specialized actor

pm47:
This allows decoupling the reconnection task from the actual client
creation.

First 4 commits are the core of this PR. The rest is bugfix and somewhat unrelated improvements in tests. Maybe I should split this PR in two.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -114,34 +114,34 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
         d.peerConnection ! PoisonPill
         stay
 
-    case Event(err: wire.Error, d: ConnectedData) =>
-      // error messages are a bit special because they can contain either temporaryChannelId or channelId (see BOLT 1)
-      d.channels.get(FinalChannelId(err.channelId)).orElse(d.channels.get(TemporaryChannelId(err.channelId))) match {
-        case Some(channel) => channel forward err
-        case None => () // let's not create a ping-pong of error messages here
-      }
-      stay
+      case Event(err: wire.Error, d: ConnectedData) =>
+        // error messages are a bit special because they can contain either temporaryChannelId or channelId (see BOLT 1)
+        d.channels.get(FinalChannelId(err.channelId)).orElse(d.channels.get(TemporaryChannelId(err.channelId))) match {
+          case Some(channel) => channel forward err
+          case None => () // let's not create a ping-pong of error messages here
+        }
+        stay
 
-    case Event(c: Peer.OpenChannel, d: ConnectedData) =>
-      if (c.fundingSatoshis >= Channel.MAX_FUNDING && !Features.hasFeature(nodeParams.features, Wumbo)) {
-        sender ! Status.Failure(new RuntimeException(s"fundingSatoshis=${c.fundingSatoshis} is too big, you must enable large channels support in 'eclair.features' to use funding above ${Channel.MAX_FUNDING} (see eclair.conf)"))
-        stay
-      } else if (c.fundingSatoshis >= Channel.MAX_FUNDING && !Features.hasFeature(d.remoteInit.features, Wumbo)) {
-        sender ! Status.Failure(new RuntimeException(s"fundingSatoshis=${c.fundingSatoshis} is too big, the remote peer doesn't support wumbo"))
-        stay
-      } else if (c.fundingSatoshis > nodeParams.maxFundingSatoshis) {
-        sender ! Status.Failure(new RuntimeException(s"fundingSatoshis=${c.fundingSatoshis} is too big for the current settings, increase 'eclair.max-funding-satoshis' (see eclair.conf)"))
-        stay
-      } else {
-        val (channel, localParams) = createNewChannel(nodeParams, funder = true, c.fundingSatoshis, origin_opt = Some(sender))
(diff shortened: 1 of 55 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -44,15 +44,10 @@ import scodec.bits.ByteVector
  *
  * Created by PM on 26/08/2016.
  */
-class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, switchboard: ActorRef, router: ActorRef, watcher: ActorRef, relayer: ActorRef, wallet: EclairWallet) extends FSMDiagnosticActorLogging[Peer.State, Peer.Data] {
+class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRef, relayer: ActorRef, wallet: EclairWallet) extends FSMDiagnosticActorLogging[Peer.State, Peer.Data] {
 
   import Peer._
 
-  val reconnectionTask: ActorRef = context.actorOf(ReconnectionTask.props(nodeParams, remoteNodeId, switchboard, router), "reconnection-task")
-  // we register the reconnection task to our transitions ourselves; if we let the child actor register itself, there is
```
t-bast, line 52: What happened to that comment? Why isn't it an issue anymore?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -39,7 +39,8 @@ import fr.acinq.eclair.transactions.DirectedHtlc.{incoming, outgoing}
 import fr.acinq.eclair.transactions.Transactions
 import fr.acinq.eclair.transactions.Transactions.{HtlcSuccessTx, htlcSuccessWeight, htlcTimeoutWeight, weight2fee}
 import fr.acinq.eclair.wire.{AnnouncementSignatures, ChannelUpdate, ClosingSigned, CommitSig, Error, FailureMessageCodecs, PermanentChannelFailure, RevokeAndAck, Shutdown, UpdateAddHtlc, UpdateFailHtlc, UpdateFailMalformedHtlc, UpdateFee, UpdateFulfillHtlc}
-import fr.acinq.eclair.{TestConstants, TestkitBaseClass, randomBytes32, _}
+import fr.acinq.eclair.{TestConstants, TestKitBaseClass, randomBytes32, _}
```
t-bast, line 42: nit:
```suggestion
import fr.acinq.eclair._
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -239,6 +234,12 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, switchboard: Act
     case Event(_: Channel.OutgoingMessage, _) => stay // we got disconnected or reconnected and this message was for the previous connection
   }
 
+  private val reconnectionTask = context.actorOf(ReconnectionTask.props(nodeParams, remoteNodeId), "reconnection-task")
+
+  onTransition {
+    case _ -> DISCONNECTED => reconnectionTask ! Peer.Transition(stateData, nextStateData)
```
t-bast, line 240: But we're also interested in transition to the `CONNECTED` state, aren't we?

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
```diff
@@ -44,19 +44,19 @@ import scala.util.Random
...
 
   when(CONNECTING) {
-    case Event(Terminated(actor), d: ConnectingData) if actor == d.connection =>
+    case Event(Status.Failure(_: Client.ConnectionFailed), d: ConnectingData) =>
       log.info(s"connection failed, next reconnection in ${d.nextReconnectionDelay.toSeconds} seconds")
       setReconnectTimer(d.nextReconnectionDelay)
       goto(WAITING) using WaitingData(nextReconnectionDelay(d.nextReconnectionDelay, nodeParams.maxReconnectInterval))
 
-    case Event(FSM.Transition(_, Peer.DISCONNECTED, Peer.CONNECTED), d) =>
+    case Event(Peer.Transition(_, _: Peer.ConnectedData), d) =>
```
t-bast, line 59: I don't understand how we can receive this event now...reconnection task doesn't listen to the peer's FSM transitions anymore, and the peer only explicitly sends a transition to the reconnection task when going to `DISCONNECTED`...

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: This isn't an issue anymore, because `Peer` actively sends the event on transition, instead of the child asynchronously registering to state transitions.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: Dammit. This hasn't been caught by unit tests because we are mocking the peer with proper behavior.

eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
pm47: You're right, we aren't.

pm47: Fixed with 9c6eed3.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: Fixed with 9c6eed3.

pm47:
Rebased.

PR #1401, opened 2020-04-30
Fix BitcoinCoreWalletSpec

t-bast:
This test was randomly failing.
There may be a rounding error or a bit of randomness somewhere in Bitcoin Core 0.19.1, sometimes a feerate right below the limit is accepted into the mempool.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
@@ -155,11 +155,12 @@ class BitcoinCoreWalletSpec extends TestKit(ActorSystem("test")) with BitcoindSe
 
     val fundingTxes = for (_ <- 0 to 3) yield {
       val pubkeyScript = Script.write(Script.pay2wsh(Scripts.multiSig2of2(randomKey.publicKey, randomKey.publicKey)))
-      wallet.makeFundingTx(pubkeyScript, MilliBtc(50), 249).pipeTo(sender.ref) // create a tx with an invalid feerate (too little)
+      wallet.makeFundingTx(pubkeyScript, MilliBtc(50), 200).pipeTo(sender.ref) // create a tx with an invalid feerate (too little)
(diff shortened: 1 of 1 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -155,11 +155,12 @@ class BitcoinCoreWalletSpec extends TestKit(ActorSystem("test")) with BitcoindSe
 
     val fundingTxes = for (_ <- 0 to 3) yield {
       val pubkeyScript = Script.write(Script.pay2wsh(Scripts.multiSig2of2(randomKey.publicKey, randomKey.publicKey)))
-      wallet.makeFundingTx(pubkeyScript, MilliBtc(50), 249).pipeTo(sender.ref) // create a tx with an invalid feerate (too little)
+      wallet.makeFundingTx(pubkeyScript, MilliBtc(50), 200).pipeTo(sender.ref) // create a tx with an invalid feerate (too little)
       val belowFeeFundingTx = sender.expectMsgType[MakeFundingTxResponse].fundingTx
       wallet.publishTransaction(belowFeeFundingTx).pipeTo(sender.ref) // try publishing the tx
-      assert(sender.expectMsgType[Failure].cause.asInstanceOf[JsonRPCError].error.message.contains("min relay fee not met, 152 < 153 (code 66)"))
-      wallet.rollback(belowFeeFundingTx) // rollback the locked outputs
+      assert(sender.expectMsgType[Failure].cause.asInstanceOf[JsonRPCError].error.message.contains("min relay fee not met, 122 < 153 (code 66)"))
```
sstone, line 161: I'm not sure that will always work since the pb was that sometimes the min relay fee was 152 instead of 153 ?

```suggestion
      assert(sender.expectMsgType[Failure].cause.asInstanceOf[JsonRPCError].error.message.contains("min relay fee not met"))
```

t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1401/commits/7f88fce49e769620f3aa93495a3a389f21709e25

PR #1402, opened 2020-04-30
Monitoring documentation

t-bast:
The first commit moves some existing documentation pages to the `docs` folder.
The second commit adds documentation for monitoring with Kamon.


t-bast:
> I would leave the BUILD.md at the root of the project

Done in https://github.com/ACINQ/eclair/pull/1402/commits/867fe3cdbd33ac439f62cfeece83cc7f415b19f0

I think having this monitoring documentation, even if incomplete, will help users move away from the GUI which we'll remove in the next release. We can also slowly migrate documentation out of the wiki (since no-one seems to be looking at it) inside this `docs` folder in the future.

PR #1404, opened 2020-05-04
Unlock transaction inputs if tx cannot be published

sstone:
In some cases, funding a tx will work but publishing it may fail (because mempool fees are not met for example). In that case we need to make sure that the tx inputs, which were locked by `fundrawtransaction`, are unlocked.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
@@ -66,15 +66,21 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
 
   def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = {
     // we unlock utxos one by one and not as a list as it would fail at the first utxo that is not actually lock and the rest would not be processed
-    val futures = outPoints.toList.map(outPoint => Utxo(outPoint.txid, outPoint.index)).map(utxo => {
-      val f = rpcClient.invoke("lockunspent", true, List(utxo))
-      f.recover { case t: Throwable =>
-        logger.warn(s"Cannot unlock utxo=$utxo", t)
-        false
-      }
-      f
-    })
-    val future = Future.sequence(futures).map(_ collect { case JBool(result) => result })
+    val futures = outPoints
+      .map(outPoint => Utxo(outPoint.txid, outPoint.index))
+      .map(utxo => rpcClient
+        .invoke("lockunspent", true, List(utxo))
+        .mapTo[JBool]
+        .transformWith {
+          case Success(JBool(result)) => Future.successful(result)
+          case Failure(JsonRPCError(error)) if error.message.contains("expected locked output") =>
+            Future.successful(true) // we consider that the outpoint was successfully unlocked (since it was not locked to begin with)
+          case Failure(t) =>
+            logger.warn(s"Cannot unlock utxo=$utxo", t)
+            Future.successful(false)
+        })
+    val future = Future.sequence(futures)
+    // return true if all outpoints were unlocked false otherwise
     future.map(_.forall(b => b))
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
@@ -133,7 +133,7 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
       logger.warn(s"txid=${tx.txid} error=$e")
       bitcoinClient.getTransaction(tx.txid).transformWith {
         case Success(_) => Future.successful(true) // tx is in the mempool, we consider that it was published
-        case Failure(_) => rollback(tx).transform { case _ => Success(false) } // we use transform here as unlocking utxos that are not locked will fail
+        case Failure(_) => rollback(tx).transform { case _ => Success(false) } // we use transform here because we want to return false in all cases even if rollback fails
       }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
@@ -26,6 +26,7 @@ import org.json4s.JsonAST._
 import scodec.bits.ByteVector
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
 
 /**
  * Created by PM on 06/07/2017.
@@ -63,7 +64,19 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
 
   def publishTransaction(tx: Transaction)(implicit ec: ExecutionContext): Future[String] = bitcoinClient.publishTransaction(tx)
 
-  def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = rpcClient.invoke("lockunspent", true, outPoints.toList.map(outPoint => Utxo(outPoint.txid, outPoint.index))) collect { case JBool(result) => result }
+  def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = {
+    // we unlock utxos one by one and not as a list as it would fail at the first utxo that is not actually lock and the rest would not be processed
+    val futures = outPoints.toList.map(outPoint => Utxo(outPoint.txid, outPoint.index)).map(utxo => {
+      val f = rpcClient.invoke("lockunspent", true, List(utxo))
+      f.recover { case t: Throwable =>
+        logger.warn(s"Cannot unlock utxo=$utxo", t)
+        false
+      }
+      f
+    })
+    val future = Future.sequence(futures).map(_ collect { case JBool(result) => result })
+    future.map(_.forall(b => b))
+  }
 
   override def getBalance: Future[Satoshi] = rpcClient.invoke("getbalance") collect { case JDecimal(balance) => Satoshi(balance.bigDecimal.scaleByPowerOfTen(8).longValue) }
 
@@ -103,13 +116,15 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
     } yield MakeFundingTxResponse(fundingTx, outputIndex, fee)
   }
 
-  override def commit(tx: Transaction): Future[Boolean] = publishTransaction(tx)
-    .map(_ => true) // if bitcoind says OK, then we consider the tx successfully published
-    .recoverWith { case JsonRPCError(e) =>
+  override def commit(tx: Transaction): Future[Boolean] = publishTransaction(tx).transformWith {
+    case Success(_) => Future.successful(true)
+    case Failure(e) =>
       logger.warn(s"txid=${tx.txid} error=$e")
-      bitcoinClient.getTransaction(tx.txid).map(_ => true).recover { case _ => false } // if we get a parseable error from bitcoind AND the tx is NOT in the mempool/blockchain, then we consider that the tx was not published
-    }
-    .recover { case _ => true } // in all other cases we consider that the tx has been published
+      bitcoinClient.getTransaction(tx.txid).transformWith {
+        case Success(_) => Future.successful(true) // tx is in the mempool, we consider that it was published
+        case Failure(_) => rollback(tx).transform { case _ => Success(false) } // we use transform here as unlocking utxos that are not locked will fail
+      }
+  }
 
   override def rollback(tx: Transaction): Future[Boolean] = unlockOutpoints(tx.txIn.map(_.outPoint)) // we unlock all utxos used by the tx
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
@@ -64,6 +64,11 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
 
   def publishTransaction(tx: Transaction)(implicit ec: ExecutionContext): Future[String] = bitcoinClient.publishTransaction(tx)
 
+  /**
+   *
+   * @param outPoints outpoints to unlock
+   * @return true if all outpoints were successfully unlocked, false otherwise
+   */
   def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = {
     // we unlock utxos one by one and not as a list as it would fail at the first utxo that is not actually lock and the rest would not be processed
     val futures = outPoints
--- a/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
@@ -163,6 +163,7 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
       // and try to unlock all outpoints: it should work too
       wallet.rollback(fundingTx).pipeTo(sender.ref)
       assert(sender.expectMsgType[Boolean])
+      assert(getLocks(sender) isEmpty)
     }
   }
 
--- a/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
@@ -156,9 +156,8 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
       // unlock the first 2 outpoints
       val tx1 = fundingTx.copy(txIn = fundingTx.txIn.take(2))
       wallet.rollback(tx1).pipeTo(sender.ref)
-      val check = sender.expectMsgType[Boolean]
+      assert(sender.expectMsgType[Boolean])
       assert(getLocks(sender) == fundingTx.txIn.drop(2).map(_.outPoint).toSet)
-      assert(check)
 
       // and try to unlock all outpoints: it should work too
       wallet.rollback(fundingTx).pipeTo(sender.ref)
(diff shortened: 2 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -68,6 +68,63 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
...
+
+    // publishing fundingTx will fail as its first 2 inputs are already spent by tx above in the mempool
+    wallet.commit(fundingTx).pipeTo(sender.ref)
+    val result = sender.expectMsgType[Boolean]
+    assert(!result)
+
+    // and all locked inputs should now be unlocked
+    awaitCond({
+      val locks = getLocks()
+      locks intersect expectedLocks isEmpty
```
t-bast, line 124: Why can't we directly test `locks.isEmpty`?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -63,7 +64,19 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
 
   def publishTransaction(tx: Transaction)(implicit ec: ExecutionContext): Future[String] = bitcoinClient.publishTransaction(tx)
 
-  def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = rpcClient.invoke("lockunspent", true, outPoints.toList.map(outPoint => Utxo(outPoint.txid, outPoint.index))) collect { case JBool(result) => result }
+  def unlockOutpoints(outPoints: Seq[OutPoint])(implicit ec: ExecutionContext): Future[Boolean] = {
+    // we unlock utxos one by one and not as a list as it would fail at the first utxo that is not actually lock and the rest would not be processed
```
t-bast, line 68: That's a bit sad, maybe it's worth adding a feature request to Bitcoin Core (support skipping over already unlocked utxos)?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -63,7 +64,19 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
...
+    val futures = outPoints.toList.map(outPoint => Utxo(outPoint.txid, outPoint.index)).map(utxo => {
+      val f = rpcClient.invoke("lockunspent", true, List(utxo))
+      f.recover { case t: Throwable =>
+        logger.warn(s"Cannot unlock utxo=$utxo", t)
+        false
+      }
+      f
+    })
+    val future = Future.sequence(futures).map(_ collect { case JBool(result) => result })
+    future.map(_.forall(b => b))
```
t-bast, line 78: I don't know if that's really what we should return. If we failed to unlock a utxo because it was already unlocked, from eclair's point of view it should be a success: the utxo is unlocked which is what we wanted, right?
But maybe the code for that behavior is hard to get right?

sstone: Yes basically the only case for which we know what happened is when everything worked.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
sstone: Yes I had the same idea. What we do here is not that inefficient because RPC requests are batched, and in most cases transactions will have very few inputs, but it would save some internal lock/unlock cycles.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
t-bast: There is no way to list utxos and know whether they're locked/unlocked?
Or no way to interpret the error message when trying to unlock and detect the error "can't unlock, it's already unlocked"?

pm47: I think what's missing here is a method comment explaining what the return value will be.

How about transforming "already unlocked" errors to success and leaving the rest as-is.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -103,13 +116,15 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
     } yield MakeFundingTxResponse(fundingTx, outputIndex, fee)
   }
 
-  override def commit(tx: Transaction): Future[Boolean] = publishTransaction(tx)
-    .map(_ => true) // if bitcoind says OK, then we consider the tx successfully published
-    .recoverWith { case JsonRPCError(e) =>
+  override def commit(tx: Transaction): Future[Boolean] = publishTransaction(tx).transformWith {
```
pm47, line 119: `transformWith` looks better than what we had before, but I would have made a minimal fix for this sensitive piece of code. Maybe all things considered it's better to make it more readable. We really really do not want to return `false` if there is the slightest chance that the tx has been published.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -66,15 +66,21 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
...
+        .transformWith {
+          case Success(JBool(result)) => Future.successful(result)
+          case Failure(JsonRPCError(error)) if error.message.contains("expected locked output") =>
+            Future.successful(true) // we consider that the outpoint was successfully unlocked (since it was not locked to begin with)
+          case Failure(t) =>
+            logger.warn(s"Cannot unlock utxo=$utxo", t)
+            Future.successful(false)
+        })
+    val future = Future.sequence(futures)
+    // return true if all outpoints were unlocked false otherwise
```
pm47, line 83: Sorry I wasn't clear. I meant it should be scaladoc, so that the contract interface is well-known.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -63,7 +64,25 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
...
+    // we unlock utxos one by one and not as a list as it would fail at the first utxo that is not actually lock and the rest would not be processed
+    val futures = outPoints
+      .map(outPoint => Utxo(outPoint.txid, outPoint.index))
+      .map(utxo => rpcClient
+        .invoke("lockunspent", true, List(utxo))
+        .mapTo[JBool]
+        .transformWith {
+          case Success(JBool(result)) => Future.successful(result)
+          case Failure(JsonRPCError(error)) if error.message.contains("expected locked output") =>
+            Future.successful(true) // we consider that the outpoint was successfully unlocked (since it was not locked to begin with)
```
pm47, line 77: I think that's way better, but it creates adherence to the current bitcoind version, which mandates a non-reg test. A very basic one where we manually lock and lock different sets of utxos (all locked, and a mix locked+unlocked).

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
sstone: We return `false` only if `publish` failed AND the tx is not in the mempool, as before, but in that case we also rollback its inputs. I tried to minimize the diff here but it was ugly.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
sstone: Ok. Parsing bitcoin RPC error messages is brittle (they may change even between minor releases it it bit us a few times before) but this one seems fairly stable, done

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -68,6 +68,105 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
...
+    assert(!result)
+
+    // and all locked inputs should now be unlocked
+    awaitCond({
+      val locks = getLocks(sender)
+      locks isEmpty
+    }, max = 10 seconds, interval = 1 second)
+  }
+
+  test("unlock outpoints correcly") {
```
pm47, line 128: I would have done it as an even simpler test in `ExtendedBitcoinClientSpec` but why not.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -68,6 +68,105 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
...
+
+    {
+      // test #1: unlock outpoints that are actually locked
+      // create a huge tx so we make sure it has > 1 inputs
+      wallet.makeFundingTx(pubkeyScript, Btc(250), 1000).pipeTo(sender.ref)
+      val MakeFundingTxResponse(fundingTx, outputIndex, _) = sender.expectMsgType[MakeFundingTxResponse]
+      assert(getLocks(sender) == fundingTx.txIn.map(_.outPoint).toSet)
+      wallet.rollback(fundingTx).pipeTo(sender.ref)
+      val check = sender.expectMsgType[Boolean]
+      assert(check)
```
pm47, line 148: Nit
```suggestion
      assert(sender.expectMsgType[Boolean])
```

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -68,6 +68,105 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
...
+    val wallet = new BitcoinCoreWallet(bitcoinClient)
+
+    val sender = TestProbe()
+    val pubkeyScript = Script.write(Script.pay2wsh(Scripts.multiSig2of2(randomKey.publicKey, randomKey.publicKey)))
+
+    {
+      // test #1: unlock outpoints that are actually locked
+      // create a huge tx so we make sure it has > 1 inputs
+      wallet.makeFundingTx(pubkeyScript, Btc(250), 1000).pipeTo(sender.ref)
+      val MakeFundingTxResponse(fundingTx, outputIndex, _) = sender.expectMsgType[MakeFundingTxResponse]
```
pm47, line 144: Why not add `assert`s on the number of inputs since the test depends on that.
```suggestion
      // create a huge tx so we make sure it has > 1 inputs
      wallet.makeFundingTx(pubkeyScript, Btc(250), 1000).pipeTo(sender.ref)
      val MakeFundingTxResponse(fundingTx, outputIndex, _) = sender.expectMsgType[MakeFundingTxResponse]
      assert(fundingTx.txIn.size > 2)
```

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -68,6 +68,105 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
...
+      assert(getLocks(sender) == fundingTx.txIn.map(_.outPoint).toSet)
+      wallet.rollback(fundingTx).pipeTo(sender.ref)
+      val check = sender.expectMsgType[Boolean]
+      assert(check)
+    }
+    {
+      // test #2: some outpoints are locked, some are unlocked
+      wallet.makeFundingTx(pubkeyScript, Btc(250), 1000).pipeTo(sender.ref)
+      val MakeFundingTxResponse(fundingTx, outputIndex, _) = sender.expectMsgType[MakeFundingTxResponse]
```
pm47, line 153:
```suggestion
      wallet.makeFundingTx(pubkeyScript, Btc(250), 1000).pipeTo(sender.ref)
      val MakeFundingTxResponse(fundingTx, outputIndex, _) = sender.expectMsgType[MakeFundingTxResponse]
      assert(fundingTx.txIn.size > 2)
```

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -68,6 +68,105 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
...
+      assert(getLocks(sender) == fundingTx.txIn.map(_.outPoint).toSet)
+
+      // unlock the first 2 outpoints
+      val tx1 = fundingTx.copy(txIn = fundingTx.txIn.take(2))
+      wallet.rollback(tx1).pipeTo(sender.ref)
+      val check = sender.expectMsgType[Boolean]
+      assert(getLocks(sender) == fundingTx.txIn.drop(2).map(_.outPoint).toSet)
+      assert(check)
```
pm47, line 161:
```suggestion
      assert(sender.expectMsgType[Boolean])
      assert(getLocks(sender) == fundingTx.txIn.drop(2).map(_.outPoint).toSet)
```

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -68,6 +68,105 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
...
+
+      // and try to unlock all outpoints: it should work too
+      wallet.rollback(fundingTx).pipeTo(sender.ref)
+      val check1 = sender.expectMsgType[Boolean]
+      assert(check1)
```
pm47, line 166:
```suggestion
      assert(sender.expectMsgType[Boolean])
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -103,13 +127,15 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
...
+  override def commit(tx: Transaction): Future[Boolean] = publishTransaction(tx).transformWith {
+    case Success(_) => Future.successful(true)
+    case Failure(e) =>
       logger.warn(s"txid=${tx.txid} error=$e")
-      bitcoinClient.getTransaction(tx.txid).map(_ => true).recover { case _ => false } // if we get a parseable error from bitcoind AND the tx is NOT in the mempool/blockchain, then we consider that the tx was not published
-    }
-    .recover { case _ => true } // in all other cases we consider that the tx has been published
+      bitcoinClient.getTransaction(tx.txid).transformWith {
+        case Success(_) => Future.successful(true) // tx is in the mempool, we consider that it was published
+        case Failure(_) => rollback(tx).transform { case _ => Success(false) } // we use transform here as unlocking utxos that are not locked will fail
```
t-bast, line 136: The comment isn't accurate anymore, is it?

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
pm47: My bad, I missed that `unlockOutpoints` was defined in `BitcoinCoreWallet`.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
sstone: I thought it was not needed since here our wallet only has 50 btc coinbase txs and needs 6 of them to create a 250 btc tx +change but ok

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -103,13 +127,15 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
...
+  override def commit(tx: Transaction): Future[Boolean] = publishTransaction(tx).transformWith {
+    case Success(_) => Future.successful(true)
+    case Failure(e) =>
       logger.warn(s"txid=${tx.txid} error=$e")
-      bitcoinClient.getTransaction(tx.txid).map(_ => true).recover { case _ => false } // if we get a parseable error from bitcoind AND the tx is NOT in the mempool/blockchain, then we consider that the tx was not published
-    }
-    .recover { case _ => true } // in all other cases we consider that the tx has been published
+      bitcoinClient.getTransaction(tx.txid).transformWith {
+        case Success(_) => Future.successful(true) // tx is in the mempool, we consider that it was published
+        case Failure(_) => rollback(tx).transform { case _ => Success(false) } // we use transform here because we want to return false in all cases even if rollback fails
```
t-bast, line 136: nit:

```suggestion
        case Failure(_) => rollback(tx).transform { case _ => Success(false) } // we use transform here because we want to return false in all cases even if rollback succeeds
```

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
```diff
@@ -68,6 +68,104 @@ class BitcoinCoreWalletSpec extends TestKitBaseClass with BitcoindService with A
...
+
+      // unlock the first 2 outpoints
+      val tx1 = fundingTx.copy(txIn = fundingTx.txIn.take(2))
+      wallet.rollback(tx1).pipeTo(sender.ref)
+      assert(sender.expectMsgType[Boolean])
+      assert(getLocks(sender) == fundingTx.txIn.drop(2).map(_.outPoint).toSet)
+
+      // and try to unlock all outpoints: it should work too
+      wallet.rollback(fundingTx).pipeTo(sender.ref)
+      assert(sender.expectMsgType[Boolean])
```
t-bast, line 165: Maybe add an `assert(getLocks(sender).isEmpty)`?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
sstone: No it's really what I meant, to mark the difference with `map`

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWalletSpec.scala
sstone: Done

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
t-bast: Ok got it, regardless of whether rollback success or fails we want to return a `Success(false)`, that makes sense

PR #1405, opened 2020-05-04
Update README

t-bast:
Remove mentions of the demo GUI.


README.md
```diff
@@ -36,7 +32,7 @@ For more information please visit the [API documentation website](https://acinq.
 
 ## Documentation
 
-Please visit our [wiki](https://github.com/acinq/eclair/wiki) to find detailed instructions on how to configure your
+Please visit our [docs](https://github.com/lightningnetwork/lightning-rfc/issues/769) and [wiki](https://github.com/acinq/eclair/wiki) to find detailed instructions on how to configure your
```
pm47, line 35: Copy paste error?

t-bast: Wow oh yeah that went very wrong

pm47: Is there a reason why you are not using a relative link?

t-bast: I think you can't have a relative link to a folder, only to a file, can you?

t-bast: But I can add an index/readme file inside the `docs` folder and point to that if you prefer

pm47: Apparently you can (see https://github.com/ACINQ/eclair/tree/readme-relative-folder)

t-bast: Nice!

PR #1407, opened 2020-05-05
Minor: catch harmless unhandled events

pm47:
This prevents unnecessary warnings to clog up the logs.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
@@ -121,6 +121,10 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey) extends 
   }
 
   whenUnhandled {
+    case Event("connected", _) => stay
+
+    case Event(Status.Failure(_: Client.ConnectionFailed), _) => stay
+
     case Event(TickReconnect, _) => stay
 
     case Event(Peer.Connect(_, hostAndPort_opt), _) =>
--- a/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
@@ -33,7 +33,7 @@ import scala.concurrent.duration._
 
 class ReconnectionTaskSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with StateTestsHelperMethods {
 
-  val fakeIPAddress = NodeAddress.fromParts("localhost", 42000).get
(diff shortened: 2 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
```diff
@@ -121,6 +121,10 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey) extends
   }
 
   whenUnhandled {
+    case Event("connected", _) => stay
+
+    case Event(Status.Failure(_: Client.ConnectionFailed), _) => stay
```
t-bast, line 126: In which case do we receive this event outside of the `CONNECTING` state?

pm47: If there are two concurrent connections (one inbound the other outbound), the inbound connection works and the outbound connection fails. Happens all the time if a peer doesn't maintain an accurate public address e.g.

pm47:
What the hell? How are tests failing.
```
- reconnect with increasing delays *** FAILED ***
  scala.MatchError: TransitionWithData(CONNECTING,WAITING,ConnectingData(/127.0.0.1:42000,1600 milliseconds),WaitingData(3200 milliseconds)) (of class fr.acinq.eclair.io.ReconnectionTaskSpec$TransitionWithData)
  at fr.acinq.eclair.io.ReconnectionTaskSpec.$anonfun$new$5(ReconnectionTaskSpec.scala:150)
  at org.scalatest.FixtureTestSuite$TestFunAndConfigMap.$anonfun$apply$1(FixtureTestSuite.scala:275)
  at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
  at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
  at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
  at org.scalatest.FixtureTestSuite$TestFunAndConfigMap.apply(FixtureTestSuite.scala:275)
  at org.scalatest.FixtureTestSuite$OneArgTest$$anon$1.apply(FixtureTestSuite.scala:223)
  at org.scalatest.TestSuite.withFixture(TestSuite.scala:196)
  at org.scalatest.TestSuite.withFixture$(TestSuite.scala:195)
  at fr.acinq.eclair.TestKitBaseClass.withFixture(TestkitBaseClass.scala:27)
```

t-bast:
I've seen this test randomly fail once too last week (it's unrelated to this PR)

pm47:
I think I get it: changing the fake ip address from `1.2.3.4:42000` to `localhost:42000` makes the dummy connection fail much faster, creating a race in the test.

PR #1408, opened 2020-05-05
Release 0.4

sstone:


eclair-core/pom.xml
```diff
@@ -21,7 +21,7 @@
     <parent>
         <groupId>fr.acinq.eclair</groupId>
         <artifactId>eclair_2.13</artifactId>
-        <version>0.3.5-SNAPSHOT</version>
+        <version>4.0</version>
```
dpad85: I think you meant 0.4.0?

pm47: Or 0.4 if we want to stay consistent with what we did for 0.3: https://github.com/ACINQ/eclair/blob/v0.3/pom.xml

PR #1413, opened 2020-05-07
mvn add note about clean argument

alevchuk:
when rebuild on different branches, build errors will occur if files are not cleaned

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -15,8 +15,10 @@ that we release, you must use the build environment (OS, JDK, maven...) that we 
 To build the project and run the tests, simply run:
 
 ```shell
-mvn clean package
+mvn package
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -18,6 +18,7 @@ To build the project and run the tests, simply run:
 mvn package
 ```
 
+NB: if the build fails, you may need to clean previously built artifacts with the `mvn clean` commands.
```
t-bast, line 21: I needed to whitelist you for the build to pass, to trigger a build you'll need to push another commit so here is a tiny suggestion.

```suggestion
NB: if the build fails, you may need to clean previously built artifacts with the `mvn clean` command.
```

PR #1415, opened 2020-05-11
Feerate provider improvements

t-bast:
- Use ConstantFeeProvider only on regtest: it doesn't make sense to fallback to hardcoded feerates on testnet/mainnet, better to stick with the one we previously used
- Fail bootstrap if we cannot retrieve feerates: this will prevent some unnecessary channel closing when starting eclair-mobile with a crappy connection
- Increase feerate providers timeout: 5 seconds was a bit aggressive for mobile connections

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -32,7 +32,7 @@ eclair {
 
   min-feerate = 3 // minimum feerate in satoshis per byte
   smooth-feerate-window = 6 // 1 = no smoothing
-  feerate-provider-timeout = 5 seconds // max time we'll wait for answers from a fee provider before we fallback to the next one
+  feerate-provider-timeout = 15 seconds // max time we'll wait for answers from a fee provider before we fallback to the next one
 
   node-alias = "eclair"
   node-color = "49daaa"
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -32,7 +32,7 @@ eclair {
 
   min-feerate = 3 // minimum feerate in satoshis per byte
   smooth-feerate-window = 6 // 1 = no smoothing
-  feerate-provider-timeout = 5 seconds // max time we'll wait for answers from a fee provider before we fallback to the next one
+  feerate-provider-timeout = 15 seconds // max time we'll wait for answers from a fee provider before we fallback to the next one
```
dpad85, line 35: I think we should leave it to 5 seconds, it seems reasonable for server environments. 

Of course for mobile apps, that can be increased to 15 seconds but that change should be done in the application's own configuration file.

t-bast: If you prefer, I checked and currently eclair-mobile / phoenix don't increase these values. I can open a separate PR on eclair-mobile / phoenix to do the increase, how does that sound?

PR #1417, opened 2020-05-12
Compute max fee before route calculation

t-bast:
Move the maximum fee computation outside of `findRoute`: this should be done earlier in the payment pipeline if we want to allow accurate fee control for MPP retries.

Right now MPP uses approximations when retrying which can lead to payments that exceed the maximum configured fees. This is a first step towards ensuring that this situation cannot happen anymore.

The way it will work for MPP split v2 is that the MultiPartPaymentLifecycle will compute the total fee budget for that payment request, and when partial payments fail it will be able to compute the available fee budget by subtracting the fees of the pending parts.

I also took this opportunity to slightly refactor `findRoute` to return all the routes it finds instead of selecting only the first one. This way `findRoute` can be used as a primitive by a future `findMultiPartRoute`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -29,6 +29,7 @@ import fr.acinq.eclair.router.Router._
 import fr.acinq.eclair.wire.ChannelUpdate
 import fr.acinq.eclair.{ShortChannelId, _}
 
+import scala.annotation.tailrec
 import scala.concurrent.duration._
 import scala.util.{Failure, Random, Success, Try}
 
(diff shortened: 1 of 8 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -146,63 +154,68 @@ object RouteCalculation {
...
-          case _ => foundRoutes
-        }
-        // At this point 'routes' cannot be empty
-        val randomizedRoutes = if (routeParams.randomize) Random.shuffle(routes) else routes
-        val route = randomizedRoutes.head.path.map(graphEdgeToHop)
-        Metrics.RouteLength.withTag(Tags.Amount, Tags.amountBucket(amount)).record(route.length)
-        Route(amount, route)
+    val boundaries: RichWeight => Boolean = { weight => feeOk(weight.cost - amount) && lengthOk(weight.length) && cltvOk(weight.cltv) }
+
+    val foundRoutes = Graph.yenKshortestPaths(g, localNodeId, targetNodeId, amount, ignoredEdges, ignoredVertices, extraEdges, numRoutes, routeParams.ratios, currentBlockHeight, boundaries)
```
pm47, line 204: Probably better to type this explicitely, because your implementation later assumes that the collection order is maintained (`++` concatenation).

```suggestion
    val foundRoutes: Seq[WeightedPath] = Graph.yenKshortestPaths(g, localNodeId, targetNodeId, amount, ignoredEdges, ignoredVertices, extraEdges, numRoutes, routeParams.ratios, currentBlockHeight, boundaries)
```

PR #1419, opened 2020-05-13
Atomically switch to new connection

pm47:
When a new connection happens while already connected, the `Peer` will
switch to the new connection.

In the current implementation, upon receiving a `ConnectionReady`, the
`Peer` will kill the current connection, then sends back the
`ConnectionReady` message to itself. In the meantime, the
`PeerConnection` will happily forward any incoming messages to the `Peer`.

This opens up to a race in the `Peer`'s mailbox between the
`ConnectionReady` message, and any incoming messages from the new
connection. If the latter win, they will get dropped because the `Peer`
is in state `DISCONNECTED`. Typically those are `ChannelReestablish`
messages and channels will get stuck in state `SYNCING`.

This PR make the `Peer` atomically switch to the new connection, without
going back to the `DISCONNECTED` state. As a result, we now have a
`CONNECTED`->`CONNECTED` transition.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -66,22 +66,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
       reconnectionTask forward p
       stay
 
-    case Event(PeerConnection.ConnectionReady(peerConnection, remoteNodeId1, address, outgoing, localInit, remoteInit), d: DisconnectedData) =>
-      require(remoteNodeId == remoteNodeId1, s"invalid nodeid: $remoteNodeId != $remoteNodeId1")
-      log.debug("got authenticated connection to address {}:{}", address.getHostString, address.getPort)
-
-      context watch peerConnection
-
-      if (outgoing) {
-        // we store the node address upon successful outgoing connection, so we can reconnect later
-        // any previous address is overwritten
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -210,8 +196,7 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
         context unwatch d.peerConnection
         d.peerConnection ! PoisonPill
         d.channels.values.toSet[ActorRef].foreach(_ ! INPUT_DISCONNECTED) // we deduplicate with toSet because there might be two entries per channel (tmp id and final id)
```
t-bast, line 198: I'm wondering if this is the best way to handle connection switching for the underlying channels.
Currently it means the channel will go offline, cancelling incoming HTLCs and disabling the channel (with a new `channel_update`), only to re-enable it right after.

Shouldn't we go one step further and have the channel FSM handle `INPUT_RESTORED` in normal states (SYNCING, NORMAL, ...)? We wouldn't send an `INPUT_DISCONNECTED`, and the channel would simply update its `peerConnection` reference.

But maybe it's a bit ugly to have this logic bleed into the channel (and maybe we don't care much because this case happens mostly for private channels with mobile nodes)?

pm47: We have to do it because there really is a disconnection. We could hide it but our peer will not, he will cancel htlcs and send us `channel_reestablish`. That's a protocol level thing.

t-bast: Ah right, this makes sense.

PR #1420, opened 2020-05-13
Use actor system as conf container

pm47:
This seems to be recommended practice:
https://doc.akka.io/docs/akka/current/general/actor-systems.html#configuration-container.

As a nice side effect, it allows putting akka-related config in eclair.conf.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -180,3 +180,21 @@ eclair {
     type = PinnedDispatcher
   }
 }
+
+akka {
+  io {
+    tcp {
+      # The maximum number of bytes delivered by a `Received` message. Before
+      # more data is read from the network the connection actor will try to
+      # do other work.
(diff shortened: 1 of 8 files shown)


araspitzu:
Changes ACK but I have a question: since we can put akka-related config in the reference.conf how about we merge it with the [application.conf](https://github.com/ACINQ/eclair/blob/master/eclair-node/src/main/resources/application.conf) ?

pm47:
> Changes ACK but I have a question: since we can put akka-related config in the reference.conf how about we merge it with the application.conf ?

They are already merged at runtime by the typesafe config library. We put as much as possible in the `reference.conf` library. IMO the metrics and logging should stay in the `application.conf` because it's application-level stuff (as opposed to library-level). I think we should probably move `akka.io.tcp.max-received-message-size` and `akka.http.client.parsing.max-content-length` to the `reference.conf`.

Issue #1416, opened 2020-05-12
eclair-node.sh shebang broken


ysangkok:
The shebang does not appear in the first line.

This means e.g. that eclair-node.sh can't be executed without specifying the interpreter on zsh.

If I try this is what I get:
`
/eclair-node.sh: 92: Syntax error: redirection unexpected`

Tested on the release of version 0.4.

t-bast:
@araspitzu do you know if this is expected? These scripts come from the maven packager, right?

araspitzu:
The scripts come from [sbt-native-packager](https://www.scala-sbt.org/sbt-native-packager) and they expect to be run by the bash interpreter (although other shells work too), @ysangkok have you tried launching it with bash?
```
bash ./eclair-node.sh
```

ysangkok:
Yeah, I tried that and I get a Future timeout error (Probably because I am using Java 8). So I think using bash explicitly helps. But the shebang should be fixed.

t-bast:
Fixed by #1423

PR #1423, opened 2020-05-14
Use shebang in first line of launcher scripts

### LINKED ISSUE #1422: The shebang is misplaced in eclair-node.sh
In eclair-node.sh v0.4 `#!/usr/bin/env bash` is not in the first line. That means that it gets executed with a default shell. The problem is that in some systems the default shell is not `bash`, so the script fails with syntax errors, like:
```
eclair-node.sh: 92: eclair-node.sh: Syntax error: redirection unexpected
```

araspitzu:
Update the unix launcher scripts to use the _shebang_ in the first line of the file, as expected by most unix-based systems. Fixes #1422 and #1416


t-bast:
@rorp can you test that?

Issue #1421, opened 2020-05-13
eclair-node.sh fails on OSX Mojave


rorp:
I've just downloaded eclair v0.4 from https://github.com/ACINQ/eclair/releases/download/v0.4/eclair-node-0.4-69c538e-bin.zip

The startup script fails with this error:
```bash
bash-3.2$ bin/eclair-node.sh
bin/eclair-node.sh: line 349: -4: substring expression < 0
Error: Could not find or load main class fr.acinq.eclair.Boot
Caused by: java.lang.ClassNotFoundException: fr.acinq.eclair.Boot
```

My OSX version is 10.14.6
```
bash-3.2$ uname -rsv
Darwin 18.7.0 Darwin Kernel Version 18.7.0: Mon Feb 10 21:08:45 PST 2020; root:xnu-4903.278.28~1/RELEASE_X86_64
```

bash version
```
bash-3.2$ /usr/bin/env bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
```

t-bast:
That looks bad, did it work with eclair 0.3.4? I don't think we changed those scripts, but updating the mvn packager versions may have something to do with it.

t-bast:
Seems to be the following lines:

```sh
declare -r real_script_path="$(realpath "$0")"
declare -r app_home="$(realpath "$(dirname "$real_script_path")")"
declare -r lib_dir="$(realpath "${app_home::-4}/lib")" # {app_home::-4} transforms ../bin in ../
```

Can you insert some prints and share the output to figure out what could go wrong?

PR #1424, opened 2020-05-14
Fix eclair-node.sh for OSX

rorp:
Fixes https://github.com/ACINQ/eclair/issues/1421


t-bast:
> There was no change in `eclair.bat`

Of course that had no impact, my bad :facepalm: 
Thanks for testing anyway!

PR #608, opened 2018-05-24
Set a minimum feerate-per-kw of 253 (fixes #602)

sstone:
We currently use feerate-per-kw = feerate-per-kb / 4, which gives 250 for 1000 sat/kb, which can result in transactions that are below the minimum relay fee of 1000 sat/kb.
This is because bitcoin core checks minimum relay fees using a transaction's virtual size (and not its actual size in bytes).
So we want :
    fee > 1000 * virtual size
    feerate-per-kw * weight > 1000 * (3 * weight / 4)
    feerate_per-kw > 250 + 3000 / (4 * weight)
    
With a conservative minimum weight of 400, we get a minimum feerate_per-kw of 253

PR #1425, opened 2020-05-15
eclair-core: correct a comment wrt feerate floor

darosior:
Just a nit that https://bitcoin.stackexchange.com/questions/95926/why-is-a-minrelaytxfee-of-1000-sat-byte-not-equivalent-to-250-sat-kw brought up to my attention.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/package.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/package.scala
@@ -74,16 +74,20 @@ package object eclair {
   def feerateKw2Byte(feeratePerKw: Long): Long = feeratePerKw / 250
 
   /**
-   * why 253 and not 250 since feerate-per-kw is feerate-per-kb / 250 and the minimum relay fee rate is 1000 satoshi/Kb ?
+   * Why 253 and not 250 since feerate-per-kw should be feerate-per-kvb / 4 and the minimum relay fee rate is 1000 satoshi/Kvb ?
    *
-   * because bitcoin core uses neither the actual tx size in bytes or the tx weight to check fees, but a "virtual size"
-   * which is (3 * weight) / 4 ...
-   * so we want :
-   * fee > 1000 * virtual size
-   * feerate-per-kw * weight > 1000 * (3 * weight / 4)
-   * feerate_per-kw > 250 + 3000 / (4 * weight)
-   * with a conservative minimum weight of 400, we get a minimum feerate_per-kw of 253
+   * Because the virtual transaction size is, as described in bip-0141, rounded *up*.
+   * Bitcoin-core uses this virtual size for its mempool logic, which ends up rounding *down* the feerate.
+   * We have `vbyte = (weight + 3) // 4`, we want :
+   *     fee // kvbyte >= 1000
+   *     fee * 1000 // vbyte >= 1000
+   *     fee * 1000 // ((weight + 3) // 4) >= 1000
+   *     fee * 1000 // (weight + 3) >= 250
+   *     (feerate * weight) // 1000 * 1000 // (weight + 3) >= 250
    *
+   * The spurious `// 1000` for the last line is because we want a feerate in *kW* (not
+   * in W, which would give us something like 0.25).
+   * With a conservative minimum weight of 400, we get a minimum feerate per kW of 253,
    * see https://github.com/ElementsProject/lightning/pull/1251
    **/
   val MinimumFeeratePerKw = 253


eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -74,15 +74,10 @@ package object eclair {
...
    *
-   * because bitcoin core uses neither the actual tx size in bytes or the tx weight to check fees, but a "virtual size"
-   * which is (3 * weight) / 4 ...
-   * so we want :
-   * fee > 1000 * virtual size
-   * feerate-per-kw * weight > 1000 * (3 * weight / 4)
-   * feerate_per-kw > 250 + 3000 / (4 * weight)
-   * with a conservative minimum weight of 400, we get a minimum feerate_per-kw of 253
+   * because bitcoin core uses neither the actual tx size in bytes nor the tx weight to check fees, but a "virtual size"
+   * which is (weight + 3) / 4 ...
```
sstone, line 80: lgtm, except I'd rather keep the explanation for why 253 was selected, why remove it ?

darosior: I explained it in the commit message :

> The calculus below was then based upon this mistake so I removed it completely.

With the corrected formula, the computation doesn't hold anymore :/

sstone: > The calculus below was then based upon this mistake so I removed it completely.

No it wasn't, the explanation and the computation are correct, there was just a typo, imho it's better to keep them and fix it.

```
fee > 1000 * virtual size	
feerate-per-kw * weight > 1000 * (3 * weight / 4) <-- typo here, should be (3 + weight) / 4 
feerate_per-kw > 250 + 3000 / (4 * weight)	<-- but not here
with a conservative minimum weight of 400, we get a minimum feerate_per-kw of 253
```

darosior: Ok, I overlooked it. Thanks !

darosior:
Ok, re-added the calculus. However I still can't get the rationale of it ...
You state :
> with a conservative minimum weight of 400, we get a minimum feerate_per-kw of 253

But :
```
feerate_per-kw > 250 + 3000 / (4 * weight)
feerate_per-kw > 251 # For weight = 400 (251.875 without rounding)
```

t-bast:
I agree with @darosior, I find the same result...
We'd need a weight of `250` to have `feerate_per_kw > 253` :confused:

sstone:
We want:
```
feerate_per-kw > 250 + 3000 / (4 * weight)
feerate_per-kw > 252 (we round up not down !)
```
So 253 is the minimum value we can actually use

t-bast:
Probably worth clarifying the calculation a bit more then and we should be good to go ;)

darosior:
> (we round up not down !)

Why do we ? `bitcoind` does not. (Sorry to insist, I just want to be sure to understand the rationale behind this)

sstone:
> > (we round up not down !)
> 
> Why do we ? `bitcoind` does not. (Sorry to insist, I just want to be sure to understand the rationale behind this)

That's not what I understood at the time: `bitcoind` checked the actual fee (in sats) against `feerate * size / 1000` so you need to round up to be safe. I you disagree please provide specific details and examples with actual values. 

The point here is to express as simply as possible the mismatch between what we use in LN (tx weight) and what bitcoind use (tx vsize) to compute fees, why it lead to errors in all LN implementations because the fees we used were too low, and how we all came up with basically the same fix at roughly the same time (AFAIK lnd and c-lightning also use a minimum fee rate of 253).

darosior:
>  I you disagree please provide specific details and examples with actual values.

....

> The point here is to express as simply as possible the mismatch between what we use in LN (tx weight) and what bitcoind use (tx vsize) to compute fees, why it lead to errors in all LN implementations because the fees we used were too low, and how we all came up with basically the same fix at roughly the same time (AFAIK lnd and c-lightning also use a minimum fee rate of 253).

I know about the point, and I don't question the `253` value *at all*. Just wanted to understand the rationale behind *your* specific calculus in the comment.

sstone:
> Just wanted to understand the rationale behind _your_ specific calculus in the comment.

See comment above (`feerate_per-kw > 252 (we round up not down !`).

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -74,16 +74,15 @@ package object eclair {
...
-   * because bitcoin core uses neither the actual tx size in bytes or the tx weight to check fees, but a "virtual size"
-   * which is (3 * weight) / 4 ...
+   * because bitcoin core uses neither the actual tx size in bytes nor the tx weight to check fees, but a "virtual size"
+   * which is (weight + 3) / 4 ...
    * so we want :
    * fee > 1000 * virtual size
-   * feerate-per-kw * weight > 1000 * (3 * weight / 4)
+   * feerate-per-kw * weight > 1000 * (3 + weight) / 4
    * feerate_per-kw > 250 + 3000 / (4 * weight)
    * with a conservative minimum weight of 400, we get a minimum feerate_per-kw of 253
```
t-bast, line 85: What about this to finish clarifying?

```suggestion
   * feerate-per-kw * weight > 1000 * (3 + weight) / 4
   * feerate_per-kw > 250 + 3000 / (4 * weight)
   * with a conservative minimum weight of 400, we get feerate_per-kw > 252 (because we round up),
   which leads to a minimum feerate_per-kw of 253
```

sstone: @darosior do you think it is better this way ? If so you can apply the suggestion and we'll merge the PR. Thanks!

darosior: @sstone I stopped to argue because it's really a nit. But here it is anyway.

I [already said](https://github.com/ACINQ/eclair/pull/1425#issuecomment-630021388) that I don't understand the reason for the round up.
This **is not what bitcoind does** (actually, if it did you would have `247` as default here !) and it seems to just fit well to justify your computation.
I think this is wrong, and I don't want to commit this under my name. That said maybe **I am** wrong but you didn't give me any rationale for this apart from "[We round up because we round up](https://github.com/ACINQ/eclair/pull/1425#issuecomment-630170178)".

You want :
```
// The * 1000 is here because the right side constant is in sat / *kilo*vbyte
// The // is a division rounded *down*
fee * 1000 // vbyte >= 1000
fee * 1000 // ((weight + 3) * 4) >= 1000
fee * 1000 // (weight + 3) >= 250
(feerate * weight) * 1000 // (weight + 3) >= 250
```
Which is only satisfied, if `weight == 400`, by `feerate >= 253` :
```python
>>> weight = 400
>>> feerate = 250
>>> (feerate * weight) // 1000 * 1000 // (weight +3)
248
>>> feerate = 252
>>> (feerate * weight) // 1000 * 1000 // (weight +3)
248
>>> feerate = 253
>>> (feerate * weight) // 1000 * 1000 // (weight +3)
250
```

t-bast: The way I see it there is just this simple inequality we want to satisfy:

```
feerate_per-kw > 250 + 3000 / (4 * weight)
```

Which, for `weight = 400`, amounts to:

```
feerate_per-kw > 251,875
```

If we want to be safe, we should assume Bitcoin Core rounds up to 252 (and if they round down, we just took extra precautions and our tx is above the treshold, so great). And if we want to be extra safe, we should assume that they check `>` instead of `>=`, hence the `253`.

Does that make sense?

t-bast: Oh are you saying that we're computing this backwards in that explanation? And what counts is going from feerate to fee in Bitcoin Core to get something that's >= 250?

darosior: > If we want to be safe, we should assume Bitcoin Core rounds up to 252 (and if they round down, we just took extra precautions and our tx is above the treshold, so great)

Yeah, absoutely I understand for the code taking some breathin room is a good idea (hence `>=` --> `>`). Just for the explanation it's... just wrong :/

> Oh are you saying that we're computing this backwards in that explanation?

Yep. I think the original computation is based on Rusty's explanation ([the commit that introduced it reference his PR](https://github.com/ACINQ/eclair/pull/608/commits/0b7d23cf96bc88c1b7819d49d0f69ba561dadf49)).
Unfortunately he made a typo ([this line](https://github.com/ElementsProject/lightning/pull/1251/files#diff-20e1124cb17d3a9a6bba3458c5061986R283), which is [now corrected](https://github.com/ElementsProject/lightning/blob/8d98b99bde55ad0844761b584311543c9051b6f7/bitcoin/feerate.h#L19)) and that's why I think the comment here needs some more assumptions to still barely hold.

> And what counts is going from feerate to fee in Bitcoin Core to get something that's >= 250?

Yep :-)
[Here is](https://bitcoin.stackexchange.com/a/95927/101498) what `bitcoind` does for the rationale behind my calculus.

> If that's the case, even if it's a nit I think it's worth clarifying.

I think too, does my explanation above fit ?

t-bast: Great, I finally fully understand this!
I think your explanation does make sense, it's clear to me now.
Can you apply it?

darosior: Sure, done

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -74,16 +74,17 @@ package object eclair {
...
-   * so we want :
-   * fee > 1000 * virtual size
-   * feerate-per-kw * weight > 1000 * (3 * weight / 4)
-   * feerate_per-kw > 250 + 3000 / (4 * weight)
-   * with a conservative minimum weight of 400, we get a minimum feerate_per-kw of 253
+   * Because the virtual transaction size is, as described in bip-0141, rounded *up*. Bitcoin-core uses this virtual size for its
+   * mempool logic, which ends up rounding *down* the feerate.
+   * We have vbyte = (weight + 3) // 4, we want :
+   *     fee * 1000 // vbyte >= 1000
+   *     fee * 1000 // ((weight + 3) * 4) >= 1000
```
t-bast, line 83: I think the parenthesis aren't correctly placed. Shouldn't this be:

```suggestion
   *     (fee * 1000 // (weight + 3)) * 4 >= 1000
```

Or 

```suggestion
   *     4 * fee * 1000 // (weight + 3) >= 1000
```

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -74,16 +74,17 @@ package object eclair {
...
+   *     fee * 1000 // (weight + 3) >= 250
+   *     (feerate * weight) * 1000 // (weight + 3) >= 250
    *
+   * With a conservative minimum weight of 400, we get a minimum feerate_per-kw of 253
```
t-bast, line 87: It's true that if we follow the computation now, it looks like we should select 252...
Maybe change this sentence to something like:

```suggestion
   * With a conservative minimum weight of 400, we get a minimum feerate_per-kw of 252, to which we add 1 for extra safety
```

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
darosior: Argh it's because I made a typo, it should be 
```suggestion
   *     fee * 1000 // ((weight + 3) / 4) >= 1000
```

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
darosior: And I made YA typo, will re-push a more carefully written patch ^^

darosior:
Corrected two typos, added an additional detail first line and an additional explanation for the last line. Now I think it's correct and we have :
```python
>>> weight = 400
>>> feerate = 250
>>> (feerate * weight) // 1000 * 1000 // (weight +3) >= 250
False
>>> feerate = 251
>>> (feerate * weight) // 1000 * 1000 // (weight +3) >= 250
False
>>> feerate = 252
>>> (feerate * weight) // 1000 * 1000 // (weight +3) >= 250
False
>>> feerate = 253
>>> (feerate * weight) // 1000 * 1000 // (weight +3) >= 250
True
```

sstone:
@dariosor you skipped over https://github.com/ACINQ/eclair/pull/1425#issuecomment-630055474 which explains why we disagree: my assumption was that bitcoin core is comparing the actual tx fees to a minimum value set  to `vsize   * 1000 / 1000`. Since in LN the actual fee that we pay are computed as `(feerate-per-kw  * weight) / 1000`, rounded down, we need to round up when computing `feerate_per-kw > 250 + 3000 / (4 * weight)` (not saying it's correct but that is the rationale I used at the time, if I made the same typo as Rusty then it's pure bad luck). 

Your explanation is clear enough, but why do you start with` bitcoind needs (worst-case): fee * 1000 / (weight + 3) >= 250` which seems to imply that bitcoin core is checking the fee rate, and that it should be > 250 ?

darosior:
> @dariosor you skipped over #1425 (comment) which explains why we disagree: my assumption was that bitcoin core is comparing the actual tx fees to a minimum value set to vsize * 1000 / 1000. 

You [are](https://github.com/bitcoin/bitcoin/blob/520e435b5e56dec7e86b0e5dbc50e3906e586a82/src/validation.cpp#L528) [right](https://github.com/bitcoin/bitcoin/blob/520e435b5e56dec7e86b0e5dbc50e3906e586a82/src/policy/feerate.cpp#L23).

> Since in LN the actual fee that we pay are computed as (feerate-per-kw * weight) / 1000, rounded down, we need to round up when computing feerate_per-kw > 250 + 3000 / (4 * weight) (not saying it's correct but that is the rationale I used at the time, if I made the same typo as Rusty then it's pure bad luck).

Ok, I agree on this but was disagreeing on the fact that you round up *twice* :
```python
>>> weight = 400
>>> 250 + 3000 // (4 * weight)
251
```

> Your explanation is clear enough, but why do you start with bitcoind needs (worst-case): fee * 1000 / (weight + 3) >= 250 

I .. [don't](https://github.com/ACINQ/eclair/pull/1425/commits/e9af84d65d1a10b4ab3fed3ad0b88be107b072db) ?
```
   * Because the virtual transaction size is, as described in bip-0141, rounded *up*.
   * which is (3 * weight) / 4 ...
   * Bitcoin-core uses this virtual size for its mempool logic, which ends up rounding *down* the feerate.
   * so we want :
   * We have `vbyte = (weight + 3) // 4`, we want :
   * fee > 1000 * virtual size
   *     fee // kvbyte >= 1000
```

t-bast:
Superceded by https://github.com/ACINQ/eclair/pull/1504

PR #1428, opened 2020-05-18
Path finding standard collections

t-bast:
I've been doing a lot of performance tests on the mainnet graph for the MPP split v2 work, and was wondering if the Scala 2.13 collections performance improvement was noticeable.

Good news: it is! On my laptop, switching to Scala's `HashMap` and `PriorityQueue` yields a consistent 10% performance increase compared to `java.util.HashMap` and the `FibonacciHeap` we were using.

However, merging this would make the merge more painful for android, as on android we still need to use the old collections. So I split the PR into two commits:

- the first one makes the merge to android *easier*, because `ArrayDeque` was not available on Scala 2.11 -> we should definitely apply this one (it's not a hot code-path, the standard `Queue` will not be a bottleneck)
- the second one makes the merge *harder* but will very likely improve performance on servers: I don't know if we should apply it and have a commit that reverses it on the `android` branch, or wait and apply it once we don't need the `android` branch anymore?

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -104,8 +104,8 @@ object Graph {
     }
 
     var allSpurPathsFound = false
-    val shortestPaths = new mutable.ArrayDeque[WeightedPath]
-    shortestPaths += WeightedPath(shortestPath, pathWeight(sourceNode, shortestPath, amount, currentBlockHeight, wr))
+    val shortestPaths = new mutable.Queue[WeightedPath]
+    shortestPaths.enqueue(WeightedPath(shortestPath, pathWeight(sourceNode, shortestPath, amount, currentBlockHeight, wr)))
     // stores the candidates for the k-th shortest path, sorted by path cost
     val candidates = new mutable.PriorityQueue[WeightedPath]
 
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -195,20 +195,21 @@ object Graph {
     // conservative estimation to avoid over-allocating memory: this is not the actual optimal size for the maps,
     // because in the worst case scenario we will insert all the vertices.
     val initialSize = 100
-    val bestWeights = new java.util.HashMap[PublicKey, RichWeight](initialSize)
-    val bestEdges = new java.util.HashMap[PublicKey, GraphEdge](initialSize)
-    val toExplore = new org.jheaps.tree.SimpleFibonacciHeap[WeightedNode, Short](NodeComparator)
+    val bestWeights = mutable.HashMap.newBuilder[PublicKey, RichWeight](initialSize, 0.75).result()
```
dpad85, line 198: I think it would be useful to provide the rationale for this 0.75 parameter in a comment.

dpad85:
The merge into android will not be big problem, we'll just have to not accept the changes from 8fa7740 when merging.

Also AFAIK if we merge then revert 8fa7740, we will have to repeat the revert every time master is merge into android, so it would be better to take time when doing the first merge into android, carefully reject changes from 8fa7740 then we should be good.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: True, it's mandatory to specify it when we specify the initial size, this is simply the default value. I'll update that.

PR #1430, opened 2020-05-18
Fix race condition on early connection failure

pm47:
Both `Client` and `TransportHandler` were watching the connection actor,
which resulted in undeterministic behavior during termination of
`PeerConnection`.

We now always return a message when a connection fails during
authentication.

Took the opportunity to add more typing (insert
deathtoallthestring.jpg).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
@@ -24,7 +24,6 @@ import akka.io.Tcp.SO.KeepAlive
 import akka.io.{IO, Tcp}
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.Logs.LogCategory
-import fr.acinq.eclair.io.Client.ConnectionFailed
 import fr.acinq.eclair.tor.Socks5Connection.{Socks5Connect, Socks5Connected, Socks5Error}
 import fr.acinq.eclair.tor.{Socks5Connection, Socks5ProxyParams}
 import fr.acinq.eclair.{Logs, NodeParams}
@@ -60,7 +59,7 @@ class Client(nodeParams: NodeParams, switchboard: ActorRef, router: ActorRef, re
     case Tcp.CommandFailed(c: Tcp.Connect) =>
       val peerOrProxyAddress = c.remoteAddress
       log.info(s"connection failed to ${str(peerOrProxyAddress)}")
-      origin_opt.foreach(_ ! Status.Failure(ConnectionFailed(remoteAddress)))
+      origin_opt.foreach(_ ! PeerConnection.ConnectionResult.ConnectionFailed(remoteAddress))
       context stop self
 
     case Tcp.Connected(peerOrProxyAddress, _) =>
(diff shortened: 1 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
```diff
@@ -112,24 +115,28 @@ class Client(nodeParams: NodeParams, switchboard: ActorRef, router: ActorRef, re
       SupervisorStrategy.Stop
   }
 
+
+  override def postStop(): Unit = {
```
t-bast, line 119: Do you want to keep that?

eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
```diff
@@ -75,32 +74,36 @@ class Client(nodeParams: NodeParams, switchboard: ActorRef, router: ActorRef, re
...
             case Tcp.CommandFailed(_: Socks5Connect) =>
               log.info(s"connection failed to ${str(remoteAddress)} via SOCKS5 ${str(proxyAddress)}")
-              origin_opt.foreach(_ ! Status.Failure(ConnectionFailed(remoteAddress)))
+              origin_opt.foreach(_ ! PeerConnection.ConnectionResult.ConnectionFailed(remoteAddress))
               context stop self
             case Socks5Connected(_) =>
               log.info(s"connected to ${str(remoteAddress)} via SOCKS5 proxy ${str(proxyAddress)}")
-              auth(proxy)
-              context become connected(proxy)
+              context unwatch proxy
```
t-bast, line 81: What happens to the proxy then? We're now calling `auth` on `connection` instead of `proxy`, is that ok?

eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
pm47: Done with 3e6a4b4.

eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
pm47: Oops, fixed with a113b8c.

We are obviously missing a test, but it is easier said than done. Would need to mock a socks5 proxy?

t-bast: It's not easy, if we simply mock there's a risk our mock deviates from what the real proxy does and our test isn't accurate...it's enough of an edge case that I think we're ok with what we have for now

t-bast: Maybe we should deploy an Acinq node behind Tor, with a small amount of funds, just to make sure we test this E2E often enough?

pm47: We do, both on testnet and mainnet. But our peers can initiate connections too, so this bugs is not easy to detect. It also means it doesn't have a lot of impact.

pm47:
> maybe worth a manual test for Tor if that's not too painful?

will monitor on next deployment

PR #1431, opened 2020-05-19
Correctly set new channel balance

t-bast:
We were incorrectly resetting channel balances to `None` when announcing a new channel.


eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -125,7 +125,8 @@ object Validation {
                     val nodeAnn = Announcements.makeNodeAnnouncement(nodeParams.privateKey, nodeParams.alias, nodeParams.color, nodeParams.publicAddresses, nodeParams.features)
                     ctx.self ! nodeAnn
                   }
-                  Some(PublicChannel(c, tx.txid, capacity, None, None, None))
+                  val channelMeta_opt = d0.privateChannels.get(c.shortChannelId).map(_.meta)
```
pm47, line 128:
```suggestion
                  // public channels that haven't yet been announced are considered as private channels
                  val channelMeta_opt = d0.privateChannels.get(c.shortChannelId).map(_.meta)
```

Issue #1434, opened 2020-05-25
Configuration values that are stored as maps in configuration files are merged, not overwritten.


sstone:
`Features` are represented as a map in our configuration files, which means that the corresponding configuration object defined in `application.conf` does not overwrite the "base" definition found in `reference.conf`, as it would if it was a string or number, but is merged with it (see https://github.com/lightbend/config/blob/c17d58e09af11f9e2a9ddae2eccca18f5f0f01f6/HOCON.md#duplicate-keys-and-object-merging).

With the following configuration files:
```
reference.conf:
  features {
    initial_routing_sync = optional
    option_data_loss_protect = optional
    gossip_queries = optional
    gossip_queries_ex = optional
    var_onion_optin = optional
  }

application.conf:
  features {
    option_data_loss_protect = mandatory
    var_onion_optin = optional
    payment_secret = optional
    basic_mpp = optional
  }
```

The final configuration, as read by the application, will be equivalent to:
```
  features {
    initial_routing_sync = optional
    option_data_loss_protect = mandatory
    gossip_queries = optional
    gossip_queries_ex = optional
    var_onion_optin = optional
    payment_secret = optional
    basic_mpp = optional
  }
```

There are several potential fixes:

- configuration keys that are represented by maps could be removed from eclair-core's `reference.conf` file, so there's nothing to merge with.
- a `null` value could be inserted in `application,.conf` to discard values defined in `reference.conf`:
```
application.conf:

  features : null
  features {
    option_data_loss_protect = mandatory
    var_onion_optin = optional
    payment_secret = optional
    basic_mpp = optional
  }
``` 
- or we could use an another representation for map-like fields such as `Features`

t-bast:
Fixed by #1435

PR #1435, opened 2020-05-25
Do not merge user provided features with defaults

araspitzu:
~~Currently we always merge the features defined by the user in `eclair.conf` with our `reference.conf` file, this led to some unexpected behaviors as described in #1434.
The PR corrects this behavior by making sure we never merge our default features with the user defined features, if the user provides some value then we use only that and if it doesn't we fallback to our hard-coded default features. Note that this makes impossible for a user to completely disable all features (i.e setting `features { }`) but i think it's okay because it doesn't affect any actual use-case.~~

Because of #1434 we need a way to let the user disable the default features. This PR explicitly allows to disable any feature from the configuration by setting `disabled` near the feature definition, all features are defined in the reference conf even if not active by default.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -46,10 +46,12 @@ eclair {
     gossip_queries = optional
     gossip_queries_ex = optional
     var_onion_optin = optional
+    option_static_remotekey = disabled
     payment_secret = optional
     basic_mpp = optional
     option_support_large_channel = optional
     trampoline_payment = disabled
+    keysend = disabled
   }
   override-features = [ // optional per-node features
     #  {
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -49,6 +49,7 @@ eclair {
     payment_secret = optional
     basic_mpp = optional
     option_support_large_channel = optional
+    trampoline_payment = disabled
   }
   override-features = [ // optional per-node features
     #  {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -133,6 +133,7 @@ object Features {
       config.getString(s"features.$name") match {
         case support if support == Mandatory.toString => Some(Mandatory)
         case support if support == Optional.toString => Some(Optional)
+        case support if support == "disabled" => None
         case wrongSupport => throw new IllegalArgumentException(s"Wrong support specified ($wrongSupport)")
       }
     }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -94,6 +94,11 @@ case class Features(activated: Set[ActivatedFeature], unknown: Set[UnknownFeatur
     )
   }
 
+  override def toString: String = {
+    val a = activated.map(f => f.feature.rfcName + ":" + f.support).mkString(",")
+    val u = unknown.map(_.bitIndex).mkString(",")
+    s"features=$a" + (if (unknown.nonEmpty) s" (unknown=$u)" else "")
+  }
 }
 
 object Features {
--- a/eclair-core/src/test/scala/fr/acinq/eclair/FeaturesSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/FeaturesSpec.scala
@@ -244,6 +244,26 @@ class FeaturesSpec extends AnyFunSuite {
 
       assertThrows[RuntimeException](fromConfiguration(confWithUnknownSupport))
     }
+
+    {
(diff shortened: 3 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -92,6 +93,14 @@ case class NodeParams(keyManager: KeyManager,
...
 object NodeParams {
 
+  val defaultFeatures = ConfigFactory.parseMap(Map(
+    s"features.${InitialRoutingSync.rfcName}" -> "optional",
+    s"features.${OptionDataLossProtect.rfcName}" -> "optional",
+    s"features.${ChannelRangeQueries.rfcName}" -> "optional",
+    s"features.${ChannelRangeQueriesExtended.rfcName}" -> "optional",
+    s"features.${VariableLengthOnion.rfcName}" -> "optional"
+  ).asJava)
+
```
sstone, line 103: I don't understand why we need to hard-code this ?
Instead we could define application features in the relevant `application.conf` (eclair-node, eclair-mobile, phoenix), and use a specific configuration file for testing.

eclair-core/src/test/scala/fr/acinq/eclair/StartupSpec.scala
```diff
@@ -82,6 +82,26 @@ class StartupSpec extends AnyFunSuite {
     assert(nodeParamsAttempt.failed.get.getMessage == "requirement failed: configuration key 'features' have moved from bytevector to human readable (ex: 'feature-name' = optional/mandatory)")
   }
 
+  test("NodeParams should not merge user defined features with default features") {
```
sstone, line 85: Testing should instead be done in eclair-node, and it would check that only features that are defined in its `application.conf` file are actually activated. This is not the case now and a failing test could be a good first commit for this PR.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
araspitzu: Because any configuration that we put in `application.conf` will be merged with the user defined configuration in `eclair.conf`. Sure we can have a different `application.conf` for eclair-mobile/phoenix but for eclair-node we still need to make sure that we don't merge the user defined features with our default. 

For example a user placing the following config block in its `eclair.conf` expects to activate only `option_support_large_channel` and no other options, this is not the case at the moment because we merge with our default features and result in a quite different feature set.
```
  features {
    option_support_large_channel = optional
  }
```

pm47:
Rebased, and added a commit to improve the `toString`:

```
features=initial_routing_sync:optional,option_data_loss_protect:optional,var_onion_optin:mandatory
features=gossip_queries_ex:optional,basic_mpp:optional (unknown=5)
features= (unknown=25,28,30)
```

eclair-core/src/main/resources/reference.conf
```diff
@@ -49,6 +49,7 @@ eclair {
     payment_secret = optional
     basic_mpp = optional
     option_support_large_channel = optional
+    trampoline_payment = disabled
```
t-bast, line 52: Aren't we also missing `option_static_remotekey = disabled` and `keysend = disabled` if we need to specify all features?

pm47: This shows the ambiguity that "not defined by default" == "defined as disabled". 🤷‍♂️

t-bast: I tested and it doesn't look like we need to explicitly disable not-defined-by-default features...
So we should either:

- remove that line and keep `option_static_remotekey` and `keysend` absent as well
- or include them all?

I slightly prefer the second option where the `reference.conf` lists all supported features, with some of them disabled.
It makes it easier for users to figure out what features they can enable/disable without having to look at the actual code.

pm47: > I tested and it doesn't look like we need to explicitly disable not-defined-by-default features...

Indeed, I see the `disabled` one as an example on how to explicitly disable other features that are enable by default.

> I slightly prefer the second option where the reference.conf lists all supported features

I bet new features will be forgotten...

The more I think about it, the more I think we should just rely on the `null` trick as laid out in the [HOCON doc](https://github.com/lightbend/config/blob/master/HOCON.md#duplicate-keys-and-object-merging).

t-bast: > Indeed, I see the disabled one as an example on how to explicitly disable other features that are enabled by default.

Got it, that makes sense.

> I bet new features will be forgotten...

Clearly :), but it's not an issue! We can add them when we realize we forgot some. I still think it's slightly better for users to have them all listed here, they clearly see what's disabled by default (experimental or obsolete).

> The more I think about it, the more I think we should just rely on the null trick

Meh. I prefer explicitly disabling stuff! It would be a problem if we had like 100 features, but given our rate of adding new feature bits we should be ok :)

pm47: Fair enough, done in deca797b3.

PR #1439, opened 2020-05-27
Multipart FSM v2

t-bast:
This is based on #1427, so please ignore the first commits.

mpp-fsm-v2

We re-work the `MultiPartPaymentLifecycle` state machine to leverage the new MPP split algorithm exposed by the router. This greatly simplifies the FSM, removes a few hacks that were added to the `PaymentLifecycle`, and should make MPP split better for both mobile and server nodes.

It also allows re-enabling the use of MPP for trampoline relays to non-trampoline nodes (yay!).
I have done a lot of regtest testing to simulate Phoenix/Eclair-Mobile scenario, but nothing beats the real thing. Once this is approved, we should create testnet builds for Phoenix and Eclair-Mobile and extensively test that everything works.

Since it was already a lot of changes, I refrained from refactoring the `RouteRequest` part. We can improve this later by cleaning up some types / creating nice type hierarchies.

NB: the failing test is because of https://github.com/ACINQ/eclair/issues/1434

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
@@ -23,7 +23,7 @@ import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.MilliSatoshi
 import fr.acinq.eclair.crypto.Sphinx
 import fr.acinq.eclair.router.Announcements
-import fr.acinq.eclair.router.Router.{ChannelDesc, ChannelHop, Hop}
+import fr.acinq.eclair.router.Router.{ChannelDesc, ChannelHop, Hop, Ignore}
 import fr.acinq.eclair.wire.Node
 
 /**
@@ -162,43 +162,43 @@ object PaymentFailure {
       .isDefined
 
   /** Update the set of nodes and channels to ignore in retries depending on the failure we received. */
-  def updateIgnored(failure: PaymentFailure, ignoreNodes: Set[PublicKey], ignoreChannels: Set[ChannelDesc]): (Set[PublicKey], Set[ChannelDesc]) = failure match {
+  def updateIgnored(failure: PaymentFailure, ignore: Ignore): Ignore = failure match {
     case RemoteFailure(hops, Sphinx.DecryptedFailurePacket(nodeId, _)) if nodeId == hops.last.nextNodeId =>
       // The failure came from the final recipient: the payment should be aborted without penalizing anyone in the route.
-      (ignoreNodes, ignoreChannels)
+      ignore
     case RemoteFailure(_, Sphinx.DecryptedFailurePacket(nodeId, _: Node)) =>
-      (ignoreNodes + nodeId, ignoreChannels)
+      ignore + nodeId
     case RemoteFailure(_, Sphinx.DecryptedFailurePacket(nodeId, failureMessage: Update)) =>
       if (Announcements.checkSig(failureMessage.update, nodeId)) {
         // We were using an outdated channel update, we should retry with the new one and nobody should be penalized.
-        (ignoreNodes, ignoreChannels)
+        ignore
       } else {
         // This node is fishy, it gave us a bad signature, so let's filter it out.
-        (ignoreNodes + nodeId, ignoreChannels)
+        ignore + nodeId
       }
     case RemoteFailure(hops, Sphinx.DecryptedFailurePacket(nodeId, _)) =>
       // Let's ignore the channel outgoing from nodeId.
       hops.collectFirst {
         case hop: ChannelHop if hop.nodeId == nodeId => ChannelDesc(hop.lastUpdate.shortChannelId, hop.nodeId, hop.nextNodeId)
       } match {
-        case Some(faultyChannel) => (ignoreNodes, ignoreChannels + faultyChannel)
-        case None => (ignoreNodes, ignoreChannels)
+        case Some(faultyChannel) => ignore + faultyChannel
+        case None => ignore
       }
     case UnreadableRemoteFailure(hops) =>
       // We don't know which node is sending garbage, let's blacklist all nodes except the one we are directly connected to and the final recipient.
-      val blacklist = hops.map(_.nextNodeId).drop(1).dropRight(1)
-      (ignoreNodes ++ blacklist, ignoreChannels)
+      val blacklist = hops.map(_.nextNodeId).drop(1).dropRight(1).toSet
+      ignore ++ blacklist
     case LocalFailure(hops, _) => hops.headOption match {
       case Some(hop: ChannelHop) =>
         val faultyChannel = ChannelDesc(hop.lastUpdate.shortChannelId, hop.nodeId, hop.nextNodeId)
-        (ignoreNodes, ignoreChannels + faultyChannel)
-      case _ => (ignoreNodes, ignoreChannels)
+        ignore + faultyChannel
+      case _ => ignore
     }
   }
 
   /** Update the set of nodes and channels to ignore in retries depending on the failures we received. */
-  def updateIgnored(failures: Seq[PaymentFailure], ignoreNodes: Set[PublicKey], ignoreChannels: Set[ChannelDesc]): (Set[PublicKey], Set[ChannelDesc]) = {
-    failures.foldLeft((ignoreNodes, ignoreChannels)) { case ((nodes, channels), failure) => updateIgnored(failure, nodes, channels) }
+  def updateIgnored(failures: Seq[PaymentFailure], ignore: Ignore): Ignore = {
+    failures.foldLeft(ignore) { case (current, failure) => updateIgnored(failure, current) }
   }
 
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
@@ -29,11 +29,11 @@ import fr.acinq.eclair.payment.receive.MultiPartPaymentFSM
 import fr.acinq.eclair.payment.send.MultiPartPaymentLifecycle.SendMultiPartPayment
 import fr.acinq.eclair.payment.send.PaymentInitiator.SendPaymentConfig
 import fr.acinq.eclair.payment.send.PaymentLifecycle.SendPayment
-import fr.acinq.eclair.payment.send.{MultiPartPaymentLifecycle, PaymentError, PaymentLifecycle}
+import fr.acinq.eclair.payment.send.{MultiPartPaymentLifecycle, PaymentLifecycle}
 import fr.acinq.eclair.router.Router.RouteParams
-import fr.acinq.eclair.router.{RouteCalculation, RouteNotFound}
+import fr.acinq.eclair.router.{BalanceTooLow, RouteCalculation, RouteNotFound}
 import fr.acinq.eclair.wire._
-import fr.acinq.eclair.{CltvExpiry, Logs, MilliSatoshi, NodeParams, nodeFee, randomBytes32}
+import fr.acinq.eclair.{CltvExpiry, Features, Logs, MilliSatoshi, NodeParams, nodeFee, randomBytes32}
 
 import scala.collection.immutable.Queue
 
@@ -46,7 +46,7 @@ import scala.collection.immutable.Queue
  * It aggregates incoming HTLCs (in case multi-part was used upstream) and then forwards the requested amount (using the
  * router to find a route to the remote node and potentially splitting the payment using multi-part).
  */
-class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, commandBuffer: ActorRef, register: ActorRef) extends Actor with DiagnosticActorLogging {
+class NodeRelayer(nodeParams: NodeParams, router: ActorRef, commandBuffer: ActorRef, register: ActorRef) extends Actor with DiagnosticActorLogging {
 
   import NodeRelayer._
 
@@ -147,7 +147,7 @@ class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, c
 
   def spawnOutgoingPayFSM(cfg: SendPaymentConfig, multiPart: Boolean): ActorRef = {
     if (multiPart) {
-      context.actorOf(MultiPartPaymentLifecycle.props(nodeParams, cfg, relayer, router, register))
+      context.actorOf(MultiPartPaymentLifecycle.props(nodeParams, cfg, router, register))
     } else {
       context.actorOf(PaymentLifecycle.props(nodeParams, cfg, router, register))
     }
@@ -158,14 +158,19 @@ class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, c
     val paymentCfg = SendPaymentConfig(paymentId, paymentId, None, paymentHash, payloadOut.amountToForward, payloadOut.outgoingNodeId, upstream, None, storeInDb = false, publishEvent = false, Nil)
     val routeParams = computeRouteParams(nodeParams, upstream.amountIn, upstream.expiryIn, payloadOut.amountToForward, payloadOut.outgoingCltv)
     payloadOut.invoiceFeatures match {
-      case Some(_) =>
-        log.debug("relaying trampoline payment to non-trampoline recipient")
+      case Some(features) =>
         val routingHints = payloadOut.invoiceRoutingInfo.map(_.map(_.toSeq).toSeq).getOrElse(Nil)
-        // TODO: @t-bast: MPP is disabled for trampoline to non-trampoline payments until we improve the splitting algorithm for nodes with a lot of channels.
-        val payFSM = spawnOutgoingPayFSM(paymentCfg, multiPart = false)
-        val finalPayload = Onion.createSinglePartPayload(payloadOut.amountToForward, payloadOut.outgoingCltv, payloadOut.paymentSecret)
-        val payment = SendPayment(payloadOut.outgoingNodeId, finalPayload, nodeParams.maxPaymentAttempts, routingHints, Some(routeParams))
-        payFSM ! payment
+        payloadOut.paymentSecret match {
+          case Some(paymentSecret) if Features(features).hasFeature(Features.BasicMultiPartPayment) =>
+            log.debug("relaying trampoline payment to non-trampoline recipient using MPP")
+            val payment = SendMultiPartPayment(paymentSecret, payloadOut.outgoingNodeId, payloadOut.amountToForward, payloadOut.outgoingCltv, nodeParams.maxPaymentAttempts, routingHints, Some(routeParams))
+            spawnOutgoingPayFSM(paymentCfg, multiPart = true) ! payment
+          case _ =>
+            log.debug("relaying trampoline payment to non-trampoline recipient without MPP")
+            val finalPayload = Onion.createSinglePartPayload(payloadOut.amountToForward, payloadOut.outgoingCltv, payloadOut.paymentSecret)
+            val payment = SendPayment(payloadOut.outgoingNodeId, finalPayload, nodeParams.maxPaymentAttempts, routingHints, Some(routeParams))
+            spawnOutgoingPayFSM(paymentCfg, multiPart = false) ! payment
+        }
       case None =>
         log.debug("relaying trampoline payment to next trampoline node")
         val payFSM = spawnOutgoingPayFSM(paymentCfg, multiPart = true)
@@ -208,7 +213,7 @@ class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, c
 
 object NodeRelayer {
 
-  def props(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, commandBuffer: ActorRef, register: ActorRef) = Props(new NodeRelayer(nodeParams, relayer, router, commandBuffer, register))
+  def props(nodeParams: NodeParams, router: ActorRef, commandBuffer: ActorRef, register: ActorRef) = Props(new NodeRelayer(nodeParams, router, commandBuffer, register))
 
   /**
    * We start by aggregating an incoming HTLC set. Once we received the whole set, we will compute a route to the next
@@ -259,15 +264,11 @@ object NodeRelayer {
    * should return upstream.
    */
   private def translateError(failures: Seq[PaymentFailure], outgoingNodeId: PublicKey): Option[FailureMessage] = {
-    def tooManyRouteNotFound(failures: Seq[PaymentFailure]): Boolean = {
-      val routeNotFoundCount = failures.collect { case f@LocalFailure(_, RouteNotFound) => f }.length
-      routeNotFoundCount > failures.length / 2
-    }
-
+    val routeNotFound = failures.collectFirst { case f@LocalFailure(_, RouteNotFound) => f }.nonEmpty
     failures match {
       case Nil => None
-      case LocalFailure(_, PaymentError.BalanceTooLow) :: Nil => Some(TemporaryNodeFailure) // we don't have enough outgoing liquidity at the moment
-      case _ if tooManyRouteNotFound(failures) => Some(TrampolineFeeInsufficient) // if we couldn't find routes, it's likely that the fee/cltv was insufficient
+      case LocalFailure(_, BalanceTooLow) :: Nil => Some(TemporaryNodeFailure) // we don't have enough outgoing liquidity at the moment
+      case _ if routeNotFound => Some(TrampolineFeeInsufficient) // if we couldn't find routes, it's likely that the fee/cltv was insufficient
       case _ =>
         // Otherwise, we try to find a downstream error that we could decrypt.
         val outgoingNodeFailure = failures.collectFirst { case RemoteFailure(_, e) if e.originNode == outgoingNodeId => e.failureMessage }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
@@ -157,6 +157,7 @@ class NodeRelayer(nodeParams: NodeParams, router: ActorRef, commandBuffer: Actor
     val paymentId = UUID.randomUUID()
     val paymentCfg = SendPaymentConfig(paymentId, paymentId, None, paymentHash, payloadOut.amountToForward, payloadOut.outgoingNodeId, upstream, None, storeInDb = false, publishEvent = false, Nil)
     val routeParams = computeRouteParams(nodeParams, upstream.amountIn, upstream.expiryIn, payloadOut.amountToForward, payloadOut.outgoingCltv)
+    // If invoice features are provided in the onion, the sender is asking us to relay to a non-trampoline recipient.
     payloadOut.invoiceFeatures match {
       case Some(features) =>
         val routingHints = payloadOut.invoiceRoutingInfo.map(_.map(_.toSeq).toSeq).getOrElse(Nil)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
@@ -70,7 +70,7 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, 
       val routeParams = r.getRouteParams(nodeParams, randomize = false) // we don't randomize the first attempt, regardless of configuration choices
       val maxFee = routeParams.getMaxFee(r.totalAmount)
       log.debug("sending {} with maximum fee {}", r.totalAmount, maxFee)
-      val d = PaymentProgress(sender, r, r.maxAttempts, Map.empty, Set.empty, Set.empty, Nil)
+      val d = PaymentProgress(sender, r, r.maxAttempts, Map.empty, Ignore.empty, Nil)
       router ! createRouteRequest(nodeParams, r.totalAmount, maxFee, routeParams, d)
       goto(WAIT_FOR_ROUTES) using d
   }
@@ -99,16 +99,16 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, 
 
     case Event(Status.Failure(t), d: PaymentProgress) =>
       log.warning("router error: {}", t.getMessage)
-      if (d.ignoreChannels.nonEmpty) {
+      if (d.ignore.channels.nonEmpty) {
         // Channels are mostly ignored for temporary reasons, likely because they didn't have enough balance to forward
         // the payment. When we're retrying an MPP split, it may make sense to retry those ignored channels because with
         // a different split, they may have enough balance to forward the payment.
         val (toSend, maxFee) = remainingToSend(nodeParams, d.request, d.pending.values)
         retriedFailedChannels = true
-        log.debug("retry sending {} with maximum fee {} without ignoring channels ({})", toSend, maxFee, d.ignoreChannels.map(_.shortChannelId).mkString(","))
+        log.debug("retry sending {} with maximum fee {} without ignoring channels ({})", toSend, maxFee, d.ignore.channels.map(_.shortChannelId).mkString(","))
         val routeParams = d.request.getRouteParams(nodeParams, randomize = true) // we randomize route selection when we retry
-        router ! createRouteRequest(nodeParams, toSend, maxFee, routeParams, d).copy(ignoreChannels = Set.empty)
-        stay using d.copy(remainingAttempts = (d.remainingAttempts - 1).max(0), ignoreChannels = Set.empty)
+        router ! createRouteRequest(nodeParams, toSend, maxFee, routeParams, d).copy(ignore = d.ignore.emptyChannels())
+        stay using d.copy(remainingAttempts = (d.remainingAttempts - 1).max(0), ignore = d.ignore.emptyChannels())
       } else {
         val failure = LocalFailure(Nil, t)
         Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType(failure)).increment()
@@ -119,8 +119,8 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, 
       if (isFinalRecipientFailure(pf, d)) {
         goto(PAYMENT_ABORTED) using PaymentAborted(d.sender, d.request, d.failures ++ pf.failures, d.pending.keySet - pf.id)
       } else {
-        val (ignoreNodes, ignoreChannels) = PaymentFailure.updateIgnored(pf.failures, d.ignoreNodes, d.ignoreChannels)
-        stay using d.copy(pending = d.pending - pf.id, ignoreNodes = ignoreNodes, ignoreChannels = ignoreChannels, failures = d.failures ++ pf.failures)
+        val ignore1 = PaymentFailure.updateIgnored(pf.failures, d.ignore)
+        stay using d.copy(pending = d.pending - pf.id, ignore = ignore1, failures = d.failures ++ pf.failures)
       }
 
     // The recipient released the preimage without receiving the full payment amount.
@@ -140,12 +140,12 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, 
         Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType(failure)).increment()
         goto(PAYMENT_ABORTED) using PaymentAborted(d.sender, d.request, d.failures ++ pf.failures :+ failure, d.pending.keySet - pf.id)
       } else {
-        val (ignoreNodes, ignoreChannels) = PaymentFailure.updateIgnored(pf.failures, d.ignoreNodes, d.ignoreChannels)
+        val ignore1 = PaymentFailure.updateIgnored(pf.failures, d.ignore)
         val stillPending = d.pending - pf.id
         val (toSend, maxFee) = remainingToSend(nodeParams, d.request, stillPending.values)
         log.debug("child payment failed, retry sending {} with maximum fee {}", toSend, maxFee)
         val routeParams = d.request.getRouteParams(nodeParams, randomize = true) // we randomize route selection when we retry
-        val d1 = d.copy(pending = stillPending, ignoreNodes = ignoreNodes, ignoreChannels = ignoreChannels, failures = d.failures ++ pf.failures)
+        val d1 = d.copy(pending = stillPending, ignore = ignore1, failures = d.failures ++ pf.failures)
         router ! createRouteRequest(nodeParams, toSend, maxFee, routeParams, d1)
         goto(WAIT_FOR_ROUTES) using d1
       }
@@ -331,15 +331,14 @@ object MultiPartPaymentLifecycle {
    * @param request           payment request containing the total amount to send.
    * @param remainingAttempts remaining attempts (after child payments fail).
    * @param pending           pending child payments (payment sent, we are waiting for a fulfill or a failure).
-   * @param ignoreChannels    channels that should be ignored (previously returned a permanent error).
+   * @param ignore            channels and nodes that should be ignored (previously returned a permanent error).
    * @param failures          previous child payment failures.
    */
   case class PaymentProgress(sender: ActorRef,
                              request: SendMultiPartPayment,
                              remainingAttempts: Int,
                              pending: Map[UUID, Route],
-                             ignoreNodes: Set[PublicKey],
-                             ignoreChannels: Set[ChannelDesc],
+                             ignore: Ignore,
                              failures: Seq[PaymentFailure]) extends Data
 
   /**
@@ -373,8 +372,7 @@ object MultiPartPaymentLifecycle {
       toSend,
       maxFee,
       d.request.assistedRoutes,
-      d.ignoreNodes,
-      d.ignoreChannels,
+      d.ignore,
       Some(routeParams),
       allowMultiPart = true,
       d.pending.values.toSeq)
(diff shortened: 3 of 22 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -312,12 +307,15 @@ object PaymentLifecycle {
...
    * Send a payment to a pre-defined route without running the path-finding algorithm.
    *
-   * @param hops         payment route to use.
+   * @param route        payment route to use.
    * @param finalPayload onion payload for the target node.
    */
-  case class SendPaymentToRoute(hops: Seq[PublicKey], finalPayload: FinalPayload, assistedRoutes: Seq[Seq[ExtraHop]] = Nil) {
-    require(hops.nonEmpty, s"payment route must not be empty")
-    val targetNodeId = hops.last
+  case class SendPaymentToRoute(route: Either[Seq[PublicKey], Route], finalPayload: FinalPayload, assistedRoutes: Seq[Seq[ExtraHop]] = Nil) {
```
t-bast, line 313: If you dislike the `Either`, I could also make `SendPaymentToRoute` a trait with two distinct child case classes, I don't know if it's better.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -216,18 +198,21 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
+
+    case Event(_: RouteResponse, _) => stay
+    case Event(_: Status.Failure, _) => stay
   }
 
   onTransition {
     case _ -> PAYMENT_ABORTED => nextStateData match {
-      case d: PaymentAborted if d.pending.isEmpty =>
-        myStop(d.sender, Left(PaymentFailed(id, paymentHash, d.failures)))
+      case d: PaymentAborted if d.pending.isEmpty => self ! Stop
```
t-bast, line 210: I introduced this instead of directly calling `myStop`, because when calling `myStop` directly, the FSM doesn't terminate during the tests. It feels like calling `FSM.stop` inside a transition has a slightly different result than doing it from a known state. I haven't investigated that much, but if someone has seen this behavior before I'm interested in what they learnt.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -51,7 +45,7 @@ import scala.util.Random
  * Sender for a multi-part payment (see https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md#basic-multi-part-payments).
  * The payment will be split into multiple sub-payments that will be sent in parallel.
  */
-class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, relayer: ActorRef, router: ActorRef, register: ActorRef) extends FSMDiagnosticActorLogging[MultiPartPaymentLifecycle.State, MultiPartPaymentLifecycle.Data] {
+class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: ActorRef, register: ActorRef) extends FSMDiagnosticActorLogging[MultiPartPaymentLifecycle.State, MultiPartPaymentLifecycle.Data] {
```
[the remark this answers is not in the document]
t-bast: Yes but there's a subtlety there because of trampoline (that probably needs to be better documented). The `SendPaymentConfig` contains the final recipient (from the Bolt 11 invoice). When we're using trampoline, the `SendMultiPartPayment` does *not* contain the final recipient, but rather the trampoline node (which is a final recipient of the first part of the route).

We need both to record accurate information about the payment in the DB and in events, which is why they are separate. This trampoline subtlety is quite hard to express easily...

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -72,108 +66,91 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
-        log.warning(s"cannot send ${d.request.totalAmount} with our current balance")
-        Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType(LocalFailure(Nil, PaymentError.BalanceTooLow)))
-        goto(PAYMENT_ABORTED) using PaymentAborted(d.sender, d.request, LocalFailure(Nil, PaymentError.BalanceTooLow) :: Nil, Set.empty)
-      } else {
-        val pending = setFees(d.request.routeParams, payments, payments.size)
+  when(WAIT_FOR_ROUTES) {
+    case Event(RouteResponse(routes), d: PaymentProgress) =>
+      log.info("{} routes found (attempt={}/{})", routes.length, d.request.maxAttempts - d.remainingAttempts + 1, d.request.maxAttempts)
+      val (toSend, maxFee) = remainingToSend(nodeParams, d.request, d.pending.values)
+      if (routes.map(_.amount).sum == toSend) {
```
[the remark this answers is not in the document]
t-bast: This part is a bit subtle, this is explained in the `else`.
The router response is valid and correctly matches the request we sent it.
But while we were waiting for routes, another child payment failed, increase the value of `toSend`. So the routes we now receive don't cover everything we want to send, which is why we discard these routes and send a new request to the router.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -72,108 +66,91 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
+        val childPayments = routes.map(route => (UUID.randomUUID(), route)).toMap
         Kamon.runWithContextEntry(parentPaymentIdKey, cfg.parentId) {
           Kamon.runWithSpan(span, finishSpan = true) {
-            pending.foreach { case (childId, payment) => spawnChildPaymentFsm(childId) ! payment }
+            childPayments.foreach { case (childId, route) => spawnChildPaymentFsm(childId) ! createChildPayment(route, d.request) }
           }
         }
-        goto(PAYMENT_IN_PROGRESS) using PaymentProgress(d.sender, d.request, d.networkStats, channels.length, 0 msat, d.request.maxAttempts - 1, pending, Set.empty, Nil)
+        goto(PAYMENT_IN_PROGRESS) using d.copy(remainingAttempts = (d.remainingAttempts - 1).max(0), pending = d.pending ++ childPayments)
```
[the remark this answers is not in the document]
t-bast: Yes, since the HTLC fan-out can vary greatly (because we can either split in many parts or not), the way I see it each attempt should map to a run of the path-finding algorithm to try a different set of routes (with a small subtlety when we discard routes because of a racing failure and don't want to count that as an attempt).

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -72,108 +66,91 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
-          d.remainingAttempts
-        }
-        goto(RETRY_WITH_UPDATED_BALANCES) using d.copy(toSend = d.toSend + failedPayment.finalPayload.amount, pending = d.pending - pf.id, failures = d.failures ++ pf.failures, ignoreChannels = ignoreChannels, remainingAttempts = remainingAttempts)
-    }
+    case Event(Status.Failure(t), d: PaymentProgress) =>
+      log.warning("router error: {}", t.getMessage)
+      if (d.ignoreChannels.nonEmpty) {
+        // Channels are mostly ignored for temporary reasons, likely because they didn't have enough balance to forward
+        // the payment. When we're retrying an MPP split, it may make sense to retry those ignored channels because with
+        // a different split, they may have enough balance to forward the payment.
```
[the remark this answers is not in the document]
t-bast: What do you mean by `loops`? We're consuming a payment attempt so we're making progress (in the FSM sense).

We may want to be more precise and decorate with failures, it's worth digging into. But it should be rather an improvement in a future PR based on what we'd see happening IRL I think.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -72,108 +66,91 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
+        val failure = LocalFailure(Nil, t)
+        Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType(failure)).increment()
+        goto(PAYMENT_ABORTED) using PaymentAborted(d.sender, d.request, d.failures :+ failure, d.pending.keySet)
+      }
 
+    case Event(pf: PaymentFailed, d: PaymentProgress) =>
+      if (isFinalRecipientFailure(pf, d)) {
+        goto(PAYMENT_ABORTED) using PaymentAborted(d.sender, d.request, d.failures ++ pf.failures, d.pending.keySet - pf.id)
+      } else {
+        val (ignoreNodes, ignoreChannels) = PaymentFailure.updateIgnored(pf.failures, d.ignoreNodes, d.ignoreChannels)
```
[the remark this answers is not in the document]
t-bast: Oh yeah, I've been really tempted with that :)

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
pm47: Right.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
pm47: Yes a loop, not an *infinite* loop. I mean retrying the same channels over and over.

Suppose there is only one route, which fails because of a `TemporaryChannelFailure` somewhere. Won't we keep retrying this route until we have `maxAttempts`?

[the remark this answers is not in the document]
t-bast: Got it. Yes in your example we'd be retrying that route instead of instantly failing, that's true.
But there's a possibility that one of the retries works, because a completely unrelated HTLC at the failing node may have shifted the balance in the meantime, so it could be worth it.

And note as well that the retry randomizes the path-finding algorithm, which means we may split the amounts differently (in the case where there's more than one route) so the temporary failure may not apply to the newly split amounts.

But you're right that maybe we'd need a special case here: if we found a single route (which we can test by checking `pending.isEmpty` we may want to fail directly instead of retrying. How does that sound?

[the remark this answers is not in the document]
pm47: > But there's a possibility that one of the retries works, because a completely unrelated HTLC at the failing node may have shifted the balance in the meantime, so it could be worth it.

FWIW, in our toolbox we also have the `ExcludeChannel` command which bans a channel for 1 minute (by default). I don't think assuming the balance will change in our favor in the few seconds it takes for a payment to complete is very reasonable, but the randomization makes sense. Maybe the errors should actually be handled by the routing algorithm in a black box? It would be able to make subtle trade offs like this one without spoiling the payment lifecycle.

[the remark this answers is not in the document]
t-bast: I don't know, what I like with the current code is that `RouteCalculation` just operates on a graph and sets of ignored/additional edges/vertices, no need for business logic around HTLCs. I think it's a good componentisation to have the business logic of choosing which nodes/channels to ignore based on failures done somewhere else.

I agree though that it's potentially unlikely to have the failing channel work when retrying instantly, so in the single route case it's probably worth aborting instantaneously. But in the case where we have multiple routes and randomization, I think this "retry without ignoring channels" makes sense.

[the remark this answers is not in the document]
pm47: > I don't know, what I like with the current code is that RouteCalculation just operates on a graph and sets of ignored/additional edges/vertices, no need for business logic around HTLCs. I think it's a good componentisation to have the business logic of choosing which nodes/channels to ignore based on failures done somewhere else.

Fair enough.

> I agree though that it's potentially unlikely to have the failing channel work when retrying instantly, so in the single route case it's probably worth aborting instantaneously. But in the case where we have multiple routes and randomization, I think this "retry without ignoring channels" makes sense.

Alright, I guess that's one of the things we'll need to monitor closely at the beginning.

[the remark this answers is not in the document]
t-bast: I'll see if I can add a metric that tracks the success rate of this mechanism, that should reveal interesting data.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/4065eaea3ecf2dae22449be6d9e6fcc51fd27f01

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -385,160 +367,20 @@ object MultiPartPaymentLifecycle {
...
-    def split(remaining: MilliSatoshi, payments: Seq[SendPayment], channels: Seq[OutgoingChannel], splitInsideChannel: (MilliSatoshi, OutgoingChannel) => Seq[SendPayment]): Seq[SendPayment] = channels match {
-      case Nil => payments
-      case _ if remaining == 0.msat => payments
-      case _ if remaining < 0.msat => throw new RuntimeException(s"payment splitting error: remaining amount must not be negative ($remaining): sending $toSend to ${request.targetNodeId} with local channels=${localChannels.map(_.toUsableBalance)}, current channels=${channels.map(_.toUsableBalance)}, network=${networkStats.map(_.capacity)}, fees=($maxFeeBase, $maxFeePct)")
-      case channel :: rest if channel.commitments.availableBalanceForSend == 0.msat => split(remaining, payments, rest, splitInsideChannel)
-      case channel :: rest =>
-        val childPayments = splitInsideChannel(remaining, channel)
-        split(remaining - childPayments.map(_.finalPayload.amount).sum, payments ++ childPayments, rest, splitInsideChannel)
-    }
+  private def isFinalRecipientFailure(pf: PaymentFailed, d: PaymentProgress): Boolean = pf.failures.collectFirst {
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/60c8ba41aaf1b6e494567c79bc4048092eb07f97

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -72,108 +66,91 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
-      if (remaining > 0.msat) {
-        log.warning(s"cannot send ${d.request.totalAmount} with our current balance")
-        Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType(LocalFailure(Nil, PaymentError.BalanceTooLow)))
-        goto(PAYMENT_ABORTED) using PaymentAborted(d.sender, d.request, LocalFailure(Nil, PaymentError.BalanceTooLow) :: Nil, Set.empty)
-      } else {
-        val pending = setFees(d.request.routeParams, payments, payments.size)
+  when(WAIT_FOR_ROUTES) {
+    case Event(RouteResponse(routes), d: PaymentProgress) =>
+      log.info("{} routes found (attempt={}/{})", routes.length, d.request.maxAttempts - d.remainingAttempts + 1, d.request.maxAttempts)
+      val (toSend, maxFee) = remainingToSend(nodeParams, d.request, d.pending.values)
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/60c8ba41aaf1b6e494567c79bc4048092eb07f97

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -72,108 +66,91 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
 
   when(WAIT_FOR_PAYMENT_REQUEST) {
     case Event(r: SendMultiPartPayment, _) =>
-      router ! GetNetworkStats
-      goto(WAIT_FOR_NETWORK_STATS) using WaitingForNetworkStats(sender, r)
+      val routeParams = r.getRouteParams(nodeParams, randomize = false) // we don't randomize the first attempt, regardless of configuration choices
+      val maxFee = routeParams.getMaxFee(r.totalAmount)
+      log.debug("sending {} with maximum fee {}", r.totalAmount, maxFee)
+      val d = PaymentProgress(sender, r, r.maxAttempts, Map.empty, Set.empty, Set.empty, Nil)
+      router ! d.createRouteRequest(nodeParams, r.totalAmount, maxFee, routeParams)
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/60c8ba41aaf1b6e494567c79bc4048092eb07f97

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -264,19 +249,6 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
-      val failure = LocalFailure(Nil, PaymentError.RetryExhausted)
-      Metrics.PaymentError.withTag(Tags.Failure, Tags.FailureType(failure))
-      Some(PaymentAborted(d.sender, d.request, d.failures ++ pf.failures :+ failure, d.pending.keySet - pf.id))
-    } else {
-      None
-    }
-  }
-
   override def mdc(currentMessage: Any): MDC = {
     Logs.mdc(category_opt = Some(Logs.LogCategory.PAYMENT), parentPaymentId_opt = Some(cfg.parentId), paymentId_opt = Some(id), paymentHash_opt = Some(paymentHash))
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/60c8ba41aaf1b6e494567c79bc4048092eb07f97

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
t-bast: I added a metric in https://github.com/ACINQ/eclair/pull/1439/commits/60c8ba41aaf1b6e494567c79bc4048092eb07f97 to track whether this is useful or not.

I kept the retry in all cases, here is why (the sample scenario I had in mind when I added this heuristic). Imagine we have two routes to the destination: `A -> B -> D` and `A -> C -> D` and `A -> B -> D` has the capacity to send the whole payment, but in fact the balance `B -> D` is too low for that. We'll send along `A -> B -> D` and then ignore `B -> D`. If `A -> C -> D` has capacity for only half the amount, the router will return `RouteNotFound`. But if we retry without ignoring the channel `B -> D`, since the retry is randomized, it will send a smaller amount on the `A -> B -> D` route and the remainder on the `A -> C -> D` route which will work (hopefully).

Obviously this is very theoretic, we'll see in practice how this behaves. I wasn't sure what the best metric was to track that, I used a counter but if you think something else is more suited let me know.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -374,19 +374,34 @@ object Router {
     }
   }
 
+  case class Ignore(nodes: Set[PublicKey], channels: Set[ChannelDesc]) {
+    // @formatter:off
+    def +(ignoreNode: PublicKey): Ignore = copy(nodes = nodes + ignoreNode)
+    def ++(ignoreNodes: Set[PublicKey]): Ignore = copy(nodes = nodes ++ ignoreNodes)
+    def +(ignoreChannel: ChannelDesc): Ignore = copy(channels = channels + ignoreChannel)
```
pm47, line 381: I see someone is having fun with operators!

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -162,43 +162,43 @@ object PaymentFailure {
       .isDefined
 
   /** Update the set of nodes and channels to ignore in retries depending on the failure we received. */
-  def updateIgnored(failure: PaymentFailure, ignoreNodes: Set[PublicKey], ignoreChannels: Set[ChannelDesc]): (Set[PublicKey], Set[ChannelDesc]) = failure match {
+  def updateIgnored(failure: PaymentFailure, ignore: Ignore): Ignore = failure match {
```
pm47, line 165: *much* more readable with the `Ignore` object

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -158,14 +158,19 @@ class NodeRelayer(nodeParams: NodeParams, relayer: ActorRef, router: ActorRef, c
...
-      case Some(_) =>
-        log.debug("relaying trampoline payment to non-trampoline recipient")
+      case Some(features) =>
         val routingHints = payloadOut.invoiceRoutingInfo.map(_.map(_.toSeq).toSeq).getOrElse(Nil)
-        // TODO: @t-bast: MPP is disabled for trampoline to non-trampoline payments until we improve the splitting algorithm for nodes with a lot of channels.
-        val payFSM = spawnOutgoingPayFSM(paymentCfg, multiPart = false)
-        val finalPayload = Onion.createSinglePartPayload(payloadOut.amountToForward, payloadOut.outgoingCltv, payloadOut.paymentSecret)
-        val payment = SendPayment(payloadOut.outgoingNodeId, finalPayload, nodeParams.maxPaymentAttempts, routingHints, Some(routeParams))
-        payFSM ! payment
+        payloadOut.paymentSecret match {
```
pm47, line 163: Why do we have this equivalence between "non-trampoline next node" and "existence of payment secret"?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -259,15 +264,11 @@ object NodeRelayer {
...
-      routeNotFoundCount > failures.length / 2
-    }
-
+    val routeNotFound = failures.collectFirst { case f@LocalFailure(_, RouteNotFound) => f }.nonEmpty
     failures match {
       case Nil => None
-      case LocalFailure(_, PaymentError.BalanceTooLow) :: Nil => Some(TemporaryNodeFailure) // we don't have enough outgoing liquidity at the moment
-      case _ if tooManyRouteNotFound(failures) => Some(TrampolineFeeInsufficient) // if we couldn't find routes, it's likely that the fee/cltv was insufficient
+      case LocalFailure(_, BalanceTooLow) :: Nil => Some(TemporaryNodeFailure) // we don't have enough outgoing liquidity at the moment
+      case _ if routeNotFound => Some(TrampolineFeeInsufficient) // if we couldn't find routes, it's likely that the fee/cltv was insufficient
```
pm47, line 271: What's the rationale behind the change in implementation between `tooManyRouteNotFound` and `routeNotFound`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -72,108 +67,93 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
 
   when(WAIT_FOR_PAYMENT_REQUEST) {
     case Event(r: SendMultiPartPayment, _) =>
-      router ! GetNetworkStats
-      goto(WAIT_FOR_NETWORK_STATS) using WaitingForNetworkStats(sender, r)
+      val routeParams = r.getRouteParams(nodeParams, randomize = false) // we don't randomize the first attempt, regardless of configuration choices
```
pm47, line 70: Why not remove the `GetNetworkStats` altogether? Is that really useful?

Why don't we randomize the first attempt?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -60,6 +54,7 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
   val id = cfg.id
   val paymentHash = cfg.paymentHash
   val start = System.currentTimeMillis
+  var retriedFailedChannels = false
```
pm47, line 57:
```suggestion
  private var retriedFailedChannels = false
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -72,108 +67,93 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
+      log.warning("router error: {}", t.getMessage)
+      if (d.ignore.channels.nonEmpty) {
+        // Channels are mostly ignored for temporary reasons, likely because they didn't have enough balance to forward
+        // the payment. When we're retrying an MPP split, it may make sense to retry those ignored channels because with
+        // a different split, they may have enough balance to forward the payment.
+        val (toSend, maxFee) = remainingToSend(nodeParams, d.request, d.pending.values)
+        retriedFailedChannels = true
+        log.debug("retry sending {} with maximum fee {} without ignoring channels ({})", toSend, maxFee, d.ignore.channels.map(_.shortChannelId).mkString(","))
+        val routeParams = d.request.getRouteParams(nodeParams, randomize = true) // we randomize route selection when we retry
+        router ! createRouteRequest(nodeParams, toSend, maxFee, routeParams, d).copy(ignore = d.ignore.emptyChannels())
```
pm47, line 110: In this order it makes more sense imo:

```suggestion
        log.debug("retry sending {} with maximum fee {} without ignoring channels ({})", toSend, maxFee, d.ignore.channels.map(_.shortChannelId).mkString(","))
        val routeParams = d.request.getRouteParams(nodeParams, randomize = true) // we randomize route selection when we retry
        router ! createRouteRequest(nodeParams, toSend, maxFee, routeParams, d).copy(ignore = d.ignore.emptyChannels())
        retriedFailedChannels = true
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -72,108 +67,93 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
-          d.remainingAttempts + 1
-        } else {
-          d.remainingAttempts
-        }
-        goto(RETRY_WITH_UPDATED_BALANCES) using d.copy(toSend = d.toSend + failedPayment.finalPayload.amount, pending = d.pending - pf.id, failures = d.failures ++ pf.failures, ignoreChannels = ignoreChannels, remainingAttempts = remainingAttempts)
-    }
+    case Event(Status.Failure(t), d: PaymentProgress) =>
+      log.warning("router error: {}", t.getMessage)
+      if (d.ignore.channels.nonEmpty) {
+        // Channels are mostly ignored for temporary reasons, likely because they didn't have enough balance to forward
```
pm47, line 103:
```suggestion
        // If no route can be found, we will retry once with the channels that we previously ignored.
        // Channels are mostly ignored for temporary reasons, likely because they didn't have enough balance to forward
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -193,6 +173,11 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
       require(ps.parts.length == 1, "child payment must contain only one part")
       log.warning(s"payment recipient fulfilled incomplete multi-part payment (id=${ps.parts.head.id})")
       goto(PAYMENT_SUCCEEDED) using PaymentSucceeded(d.sender, d.request, ps.paymentPreimage, ps.parts, d.pending - ps.parts.head.id)
+
+    case Event(Stop, d: PaymentAborted) => myStop(d.sender, Left(PaymentFailed(id, paymentHash, d.failures)))
+
+    case Event(_: RouteResponse, _) => stay
+    case Event(_: Status.Failure, _) => stay
```
pm47, line 180: This should never happen right? Feels like the default behavior of displaying unhandled messages as `WARN` would be better?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
pm47: Maybe using transition for this isn't a good pattern. How about:

```scala
def gotoAbortedOrStop(d: PaymentAborted): State = {
    if (d.pending.isEmpty) {
      myStop(d.sender, Left(PaymentFailed(id, paymentHash, d.failures)))
    } else
      goto(PAYMENT_ABORTED)
  }
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -75,44 +75,39 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
-      } else {
-        router ! RouteRequest(c.getRouteRequestStart(nodeParams), c.targetNodeId, c.finalPayload.amount, c.getMaxFee(nodeParams), c.assistedRoutes, routeParams = c.routeParams, ignoreNodes = ignoredNodes)
-      }
+      log.debug("sending {} to {}", c.finalPayload.amount, c.targetNodeId)
+      router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, c.getMaxFee(nodeParams), c.assistedRoutes, routeParams = c.routeParams)
       if (cfg.storeInDb) {
         paymentsDb.addOutgoingPayment(OutgoingPayment(id, cfg.parentId, cfg.externalId, paymentHash, PaymentType.Standard, c.finalPayload.amount, cfg.recipientAmount, cfg.recipientNodeId, System.currentTimeMillis, cfg.paymentRequest, OutgoingPaymentStatus.Pending))
       }
-      goto(WAITING_FOR_ROUTE) using WaitingForRoute(sender, c, Nil, ignoredNodes, Set.empty)
+      goto(WAITING_FOR_ROUTE) using WaitingForRoute(sender, c, Nil, Ignore.empty)
```
pm47, line 99: Why did you change the behavior here (reset the ignored nodes)?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
pm47: It's fine with me that way.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -312,12 +307,15 @@ object PaymentLifecycle {
...
    *
-   * @param hops         payment route to use.
+   * @param route        payment route to use.
    * @param finalPayload onion payload for the target node.
    */
-  case class SendPaymentToRoute(hops: Seq[PublicKey], finalPayload: FinalPayload, assistedRoutes: Seq[Seq[ExtraHop]] = Nil) {
-    require(hops.nonEmpty, s"payment route must not be empty")
-    val targetNodeId = hops.last
+  case class SendPaymentToRoute(route: Either[Seq[PublicKey], Route], finalPayload: FinalPayload, assistedRoutes: Seq[Seq[ExtraHop]] = Nil) {
+    require(route.fold(_.nonEmpty, _.hops.nonEmpty), s"payment route must not be empty")
```
pm47, line 314: Nit:
```suggestion
    require(route.fold(_.nonEmpty, _.hops.nonEmpty), "payment route must not be empty")
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -75,44 +75,39 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
-      val hops = c.routePrefix ++ routes.head.hops
-      log.info(s"route found: attempt=${failures.size + 1}/${c.maxAttempts} route=${hops.map(_.nextNodeId).mkString("->")} channels=${hops.map(_.lastUpdate.shortChannelId).mkString("->")}")
-      val firstHop = hops.head
-      val (cmd, sharedSecrets) = OutgoingPacket.buildCommand(cfg.upstream, paymentHash, hops, c.finalPayload)
-      register ! Register.ForwardShortId(firstHop.lastUpdate.shortChannelId, cmd)
-      goto(WAITING_FOR_PAYMENT_COMPLETE) using WaitingForComplete(s, c, cmd, failures, sharedSecrets, ignoreNodes, ignoreChannels, Route(c.finalPayload.amount, hops))
-
-    case Event(Status.Failure(t), WaitingForRoute(s, _, failures, _, _)) =>
+    case Event(RouteResponse(route +: _), WaitingForRoute(s, c, failures, ignore)) =>
+      log.info(s"route found: attempt=${failures.size + 1}/${c.maxAttempts} route=${route.hops.map(_.nextNodeId).mkString("->")} channels=${route.hops.map(_.lastUpdate.shortChannelId).mkString("->")}")
```
pm47, line 104: Nit²: there could probably be some factorization on printing routes.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: It's so much easier to read!

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
t-bast: We don't :)
The code here starts by looking at whether we have `invoiceFeatures` available in the onion. This field will only be there when a phoenix node asks us to relay to a non-trampoline node.
Then we check the `paymentSecret` to see whether we can use MPP or not to reach that non-trampoline destination node.
Does that make more sense?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
t-bast: This was a legacy hack :)

The behavior should have always been: if we have a `RouteNotFound` error, that means we couldn't find a route to our destination, so we should ask the sender to retry with a higher fee budget.

But because of the previous MPP split algorithm that looked at channels "blindly", it could generate some `RouteNotFound`s that weren't really `RouteNotFound`s, so we had to relax the condition to "if there are many route not found errors".

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
t-bast: > Why not remove the GetNetworkStats altogether?

Some users told me they're looking at the `NetworkStats` (through the API).
I chose to keep it for now for that reason and because we may want to use it in the future to dynamically choose the values for `min-amount-satoshis` instead of using one provided by users in `eclair.conf`.

> Why don't we randomize the first attempt?

Because it will generate more HTLCs. The non-randomized algorithm is better from an algorithmic point-of-view: it will provide a better set of routes, and a lower total fee so it's preferable. But if these routes fail (because we're blind to remote balances), then it makes sense to introduce randomization in there to try different amounts and routes. It's a heuristic, my guts say it makes sense but we'll have to see it in action!

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
t-bast: This can (and will) happen in normal scenario, this is why I didn't log it.
If you receive an error from an intermediate node and ask the router for a new set of routes, and then receive an error from the final recipient, we will go to `PAYMENT_ABORTED` (because it's an error from the final recipient) while we're still waiting for a route response.
Once we're in that state these routes don't matter anymore, so we just handle the message and ignore it.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
t-bast: Because there can never be any `ignoreNodes` now that we removed the hack-ish imposed "route prefix"

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
pm47: Yes but definitely probably deserves a comment

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/da4f8694edbfc83cfcb3e369079930aac028c748

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/da4f8694edbfc83cfcb3e369079930aac028c748

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/da4f8694edbfc83cfcb3e369079930aac028c748

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/da4f8694edbfc83cfcb3e369079930aac028c748

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/da4f8694edbfc83cfcb3e369079930aac028c748

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/da4f8694edbfc83cfcb3e369079930aac028c748

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
t-bast: Do you really dislike the pattern of sending ourselves a message in the transition?
I quite like this centralized transition, it makes sure we cannot shoot ourselves in the foot by doing a transition without using the `gotoAbortedOrStop` method.

pm47: The more I think about it, the less I like the pattern of intercepting and changing the next state from within a transition. Sending a message to ourself is fine, it's just the overall pattern that I don't like.

t-bast: Really? I think it makes sense that we can only stop from terminal states (`PAYMENT_ABORTED` and `PAYMENT_SUCCEEDED`), what I find lacking is a hook when arriving into a new state, which is somewhat replaced by the `onTransition` + sending events to ourselves trick. Let me know if you find a better solution.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1439/commits/34204e1ca5e8e65b5683672708aadbddb04f7b60

PR #1440, opened 2020-05-28
Accept multiple channels for some API

### LINKED ISSUE #1432: Feature request: API call to update relay fees for bulk or all channels
Would it be possible to change `UpdateRelayFee` such that the fee can be changed for more than one channel at once via providing a list of short and/or long channel IDs?

Due to the constantly changing on-chain fees it would also be very helpful to have another API call to simply update relay fees for all channels in one go.

t-bast:
It's handy to update relay fees for multiple channels at once.
Closing and force-closing channels may also make sense to do in batch.

I've made a note to update the API documentation when we release these changes.

Closes #1432


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -69,15 +70,15 @@ trait Eclair {
 
   def open(nodeId: PublicKey, fundingAmount: Satoshi, pushAmount_opt: Option[MilliSatoshi], fundingFeerateSatByte_opt: Option[Long], flags_opt: Option[Int], openTimeout_opt: Option[Timeout])(implicit timeout: Timeout): Future[ChannelCommandResponse]
 
-  def close(channelIdentifier: Either[ByteVector32, ShortChannelId], scriptPubKey_opt: Option[ByteVector])(implicit timeout: Timeout): Future[ChannelCommandResponse]
+  def close(channels: List[Either[ByteVector32, ShortChannelId]], scriptPubKey_opt: Option[ByteVector])(implicit timeout: Timeout): Future[Map[Either[ByteVector32, ShortChannelId], ChannelCommandResponse]]
```
araspitzu, line 73: nit: we could reduce the verbosity of the code here and below by introducing a type alias like:
```
  type ChannelIdentifier = Either[ByteVector32, ShortChannelId]
```
and then replacing all instances of `Either[ByteVector32, ShortChannelId]` with `ChannelIdentifier`

eclair-node/src/main/scala/fr/acinq/eclair/api/ExtraDirectives.scala
```diff
@@ -50,10 +53,19 @@ trait ExtraDirectives extends Directives {
...
+  }
+
+  def withChannelsIdentifier: Directive1[List[Either[ByteVector32, ShortChannelId]]] = formFields(channelIdFormParam.?, channelIdsFormParam.?, shortChannelIdFormParam.?, shortChannelIdsFormParam.?).tflatMap {
+    case (None, None, None, None) => reject(MalformedFormFieldRejection("channelId(s)/shortChannelId(s)", "Must specify channelId, channelIds, shortChannelId or shortChannelIds"))
+    case (channelId_opt, channelIds_opt, shortChannelId_opt, shortChannelIds_opt) =>
+      val channelId: List[Either[ByteVector32, ShortChannelId]] = channelId_opt.map(cid => Left(cid)).toList
+      val channelIds: List[Either[ByteVector32, ShortChannelId]] = channelIds_opt.map(_.map(cid => Left(cid))).toList.flatten
+      val shortChannelId: List[Either[ByteVector32, ShortChannelId]] = shortChannelId_opt.map(scid => Right(scid)).toList
+      val shortChannelIds: List[Either[ByteVector32, ShortChannelId]] = shortChannelIds_opt.map(_.map(scid => Right(scid))).toList.flatten
+      provide(channelId ++ channelIds ++ shortChannelId ++ shortChannelIds)
```
araspitzu, line 68: Is it worth to call `.distinct` on the resulting collection of ids? We can prevent some user errors in case of repeated ids but not all errors because they can mix `channelId` and `shortChannelId` for the same channel.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: I went back and forth on that one because I didn't know where to put it...
I wasn't a big fan of putting this directly in the `fr.acinq.eclair` namespace, and couldn't find the right object to put it in...let me try again.

eclair-node/src/main/scala/fr/acinq/eclair/api/ExtraDirectives.scala
t-bast: Good idea, I'll add a `distinct` it doesn't hurt.

Users can indeed mess up and send the same command twice to the same channel if they specify it both via `channelId` and `shortChannelId`, but I figured it was mostly harmless (the channels should handle the duplicate command without any issue) and something the user should fix on their side of the script.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1440/commits/75fd29f9229c6c3666c3fac11df59f5225dc1907

eclair-node/src/main/scala/fr/acinq/eclair/api/ExtraDirectives.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1440/commits/75fd29f9229c6c3666c3fac11df59f5225dc1907

t-bast:
Actually I didn't test the failing cases thoroughly enough, in case of failures we don't return a map that allows tracking which channel command succeeded and which failed, we only return the failure.

I should handle that more cleanly.

PR #1441, opened 2020-06-02
More aggressive channel exclusion

t-bast:
We were missing some cases where we should exclude temporarily unavailable channels.

The first missing case was when this was the last attempt (which impacts a lot the MPP scenario, since in MPP child payments do a single attempt).

The second missing case was when a channel from the routing hint is unavailable. Since we send the channel update to the router, the router was correctly removing it from the graph, but then we were re-adding it via the routing hints.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
@@ -278,8 +278,8 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
       // channel flags to indicate that it's disabled
       data.c.assistedRoutes.flatMap(r => RouteCalculation.toChannelDescs(r, data.c.targetNodeId))
         .find(_.shortChannelId == failure.update.shortChannelId)
-        .foreach(desc => router ! ExcludeChannel(desc))
-      data.c.assistedRoutes.map(_.filter(extraHop => extraHop.shortChannelId != failure.update.shortChannelId))
+        .foreach(desc => router ! ExcludeChannel(desc)) // we want the exclusion to be router-wide so that sister payments in the case of MPP are aware the channel is faulty
+      data.c.assistedRoutes
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
@@ -149,6 +149,10 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
           val failure = res match {
             case Success(e@Sphinx.DecryptedFailurePacket(nodeId, failureMessage)) =>
               log.info(s"received an error message from nodeId=$nodeId (failure=$failureMessage)")
+              failureMessage match {
+                case failureMessage: Update => handleUpdate(nodeId, failureMessage, data)
+                case _ =>
+              }
               RemoteFailure(cfg.fullRoute(route), e)
             case Failure(t) =>
               log.warning(s"cannot parse returned error: ${t.getMessage}")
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -266,6 +240,49 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
+        )
+        case extraHop => extraHop
+      })
+    } else {
+      // if the channel is disabled, we temporarily exclude it: this is necessary because the routing hint doesn't contain
+      // channel flags to indicate that it's disabled
+      data.c.assistedRoutes.flatMap(r => RouteCalculation.toChannelDescs(r, data.c.targetNodeId))
+        .find(_.shortChannelId == failure.update.shortChannelId)
+        .foreach(desc => router ! ExcludeChannel(desc))
+      data.c.assistedRoutes.map(_.filter(extraHop => extraHop.shortChannelId != failure.update.shortChannelId))
```
pm47, line 282: I would remove this extra precaution because it makes the processing tree more complex and shouldn't be necessary.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -266,6 +240,49 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
+          feeProportionalMillionths = failure.update.feeProportionalMillionths
+        )
+        case extraHop => extraHop
+      })
+    } else {
+      // if the channel is disabled, we temporarily exclude it: this is necessary because the routing hint doesn't contain
+      // channel flags to indicate that it's disabled
+      data.c.assistedRoutes.flatMap(r => RouteCalculation.toChannelDescs(r, data.c.targetNodeId))
+        .find(_.shortChannelId == failure.update.shortChannelId)
+        .foreach(desc => router ! ExcludeChannel(desc))
```
pm47, line 281:
```suggestion
        .foreach(desc => router ! ExcludeChannel(desc)) // we want the exclusion to be router-wide so that sister payments in the case of MPP are aware the channel is faulty
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1441/commits/f3bb7e33cab0c5fe15343cc8e2d71aeb30122fd9

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1441/commits/f3bb7e33cab0c5fe15343cc8e2d71aeb30122fd9

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -266,6 +240,49 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
     goto(WAITING_FOR_ROUTE) using WaitingForRoute(data.sender, data.c, data.failures :+ failure, ignore1)
   }
 
+  private def handleUpdate(nodeId: PublicKey, failure: Update, data: WaitingForComplete): Seq[Seq[ExtraHop]] = {
```
pm47, line 243:
```suggestion
  /**
   * Apply the channel_update to our routing table
   * 
   * @return Updated routing hints if applicable
   */
  private def handleUpdate(nodeId: PublicKey, failure: Update, data: WaitingForComplete): Seq[Seq[ExtraHop]] = {
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1441/commits/5fa4ae5f146e3c3827ee474ddf555c9b30785aaa

PR #1239, opened 2019-12-07
Prefix all stored LN messages by their length

pm47:
Now that we have optional fields at the end of messages, we have to
store their length in database, otherwise disasters will happen.

PR #1442, opened 2020-06-02
Use length-delimited, byte-aligned codecs

pm47:
All LN protocol message must be stored as length-delimited, because they may have arbitrary trailing data.

Took the opportunity to restore byte alignment. It was mainly due to the use of single-bit `bool` codec.

The goal is to play safe and have simple codecs, so the existing codecs have been kept and moved to a `Legacy` inner object.

This is an alternative to #1239.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
@@ -16,11 +16,9 @@
 
 package fr.acinq.eclair.wire
 
-import java.util.UUID
-
 import akka.actor.ActorRef
 import fr.acinq.bitcoin.DeterministicWallet.{ExtendedPrivateKey, KeyPath}
-import fr.acinq.bitcoin.{ByteVector32, ByteVector64, Crypto, OutPoint, Transaction, TxOut}
+import fr.acinq.bitcoin.{ByteVector32, OutPoint, Transaction, TxOut}
 import fr.acinq.eclair.channel._
 import fr.acinq.eclair.crypto.ShaChain
 import fr.acinq.eclair.payment.relay.Origin
@@ -29,18 +27,18 @@ import fr.acinq.eclair.transactions._
 import fr.acinq.eclair.wire.CommonCodecs._
 import fr.acinq.eclair.wire.LightningMessageCodecs._
 import grizzled.slf4j.Logging
-import scodec.bits.BitVector
 import scodec.codecs._
 import scodec.{Attempt, Codec}
-import shapeless.{HNil, ::}
-
-import scala.compat.Platform
-import scala.concurrent.duration._
 
 /**
  * Created by PM on 02/06/2017.
  */
 object ChannelCodecs extends Logging {
+
+  /**
+   * All LN protocol message must be stored as length-delimited, because they may have arbitrary trailing data
+   */
+  def lengthDelimited[T](codec: Codec[T]): Codec[T] = variableSizeBytesLong(varintoverflow, codec)
 
   val keyPathCodec: Codec[KeyPath] = ("path" | listOfN(uint16, uint32)).xmap[KeyPath](l => new KeyPath(l), keyPath => keyPath.path.toList).as[KeyPath]
 
@@ -51,14 +49,12 @@ object ChannelCodecs extends Logging {
       ("path" | keyPathCodec) ::
       ("parent" | int64)).as[ExtendedPrivateKey]
 
-  val channelVersionCodec: Codec[ChannelVersion] = discriminatorWithDefault[ChannelVersion](
-    discriminator = discriminated[ChannelVersion].by(byte)
-      .typecase(0x01, bits(ChannelVersion.LENGTH_BITS).as[ChannelVersion])
-    // NB: 0x02 and 0x03 are *reserved* for backward compatibility reasons
-    ,
-    fallback = provide(ChannelVersion.ZEROES) // README: DO NOT CHANGE THIS !! old channels don't have a channel version
-    // field and don't support additional features which is why all bits are set to 0.
-  )
+  val channelVersionCodec: Codec[ChannelVersion] = bits(ChannelVersion.LENGTH_BITS).as[ChannelVersion]
+
+  /**
+   * byte-aligned boolean codec
+   */
+  val bool8: Codec[Boolean] = bool(8)
 
   def localParamsCodec(channelVersion: ChannelVersion): Codec[LocalParams] = (
     ("nodeId" | publicKey) ::
@@ -69,8 +65,8 @@ object ChannelCodecs extends Logging {
       ("htlcMinimum" | millisatoshi) ::
       ("toSelfDelay" | cltvExpiryDelta) ::
       ("maxAcceptedHtlcs" | uint16) ::
-      ("isFunder" | bool) ::
-      ("defaultFinalScriptPubKey" | varsizebinarydata) ::
+      ("isFunder" | bool8) ::
+      ("defaultFinalScriptPubKey" | lengthDelimited(bytes)) ::
       ("localPaymentBasepoint" | optional(provide(channelVersion.hasStaticRemotekey), publicKey)) ::
       ("features" | combinedFeaturesCodec)).as[LocalParams]
 
@@ -89,14 +85,9 @@ object ChannelCodecs extends Logging {
       ("htlcBasepoint" | publicKey) ::
       ("features" | combinedFeaturesCodec)).as[RemoteParams]
 
-  val htlcCodec: Codec[DirectedHtlc] = discriminated[DirectedHtlc].by(bool)
-    .typecase(true, updateAddHtlcCodec.as[IncomingHtlc])
-    .typecase(false, updateAddHtlcCodec.as[OutgoingHtlc])
-
-  def setCodec[T](codec: Codec[T]): Codec[Set[T]] = Codec[Set[T]](
-    (elems: Set[T]) => listOfN(uint16, codec).encode(elems.toList),
-    (wire: BitVector) => listOfN(uint16, codec).decode(wire).map(_.map(_.toSet))
-  )
+  val htlcCodec: Codec[DirectedHtlc] = discriminated[DirectedHtlc].by(bool8)
+    .typecase(true, lengthDelimited(updateAddHtlcCodec).as[IncomingHtlc])
+    .typecase(false, lengthDelimited(updateAddHtlcCodec).as[OutgoingHtlc])
 
   val commitmentSpecCodec: Codec[CommitmentSpec] = (
     ("htlcs" | setCodec(htlcCodec)) ::
@@ -104,16 +95,16 @@ object ChannelCodecs extends Logging {
       ("toLocal" | millisatoshi) ::
       ("toRemote" | millisatoshi)).as[CommitmentSpec]
 
-  val outPointCodec: Codec[OutPoint] = variableSizeBytes(uint16, bytes.xmap(d => OutPoint.read(d.toArray), d => OutPoint.write(d)))
+  val outPointCodec: Codec[OutPoint] = lengthDelimited(bytes.xmap(d => OutPoint.read(d.toArray), d => OutPoint.write(d)))
 
-  val txOutCodec: Codec[TxOut] = variableSizeBytes(uint16, bytes.xmap(d => TxOut.read(d.toArray), d => TxOut.write(d)))
+  val txOutCodec: Codec[TxOut] = lengthDelimited(bytes.xmap(d => TxOut.read(d.toArray), d => TxOut.write(d)))
 
-  val txCodec: Codec[Transaction] = variableSizeBytes(uint16, bytes.xmap(d => Transaction.read(d.toArray), d => Transaction.write(d)))
+  val txCodec: Codec[Transaction] = lengthDelimited(bytes.xmap(d => Transaction.read(d.toArray), d => Transaction.write(d)))
 
   val inputInfoCodec: Codec[InputInfo] = (
     ("outPoint" | outPointCodec) ::
       ("txOut" | txOutCodec) ::
-      ("redeemScript" | varsizebinarydata)).as[InputInfo]
+      ("redeemScript" | lengthDelimited(bytes))).as[InputInfo]
 
   val txWithInputInfoCodec: Codec[TransactionWithInputInfo] = discriminated[TransactionWithInputInfo].by(uint16)
     .typecase(0x01, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec)).as[CommitTx])
@@ -127,19 +118,10 @@ object ChannelCodecs extends Logging {
     .typecase(0x09, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec)).as[HtlcPenaltyTx])
     .typecase(0x10, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec)).as[ClosingTx])
 
-  // this is a backward compatible codec (we used to store the sig as DER encoded), now we store it as 64-bytes
-  val sig64OrDERCodec: Codec[ByteVector64] = Codec[ByteVector64](
-    (value: ByteVector64) => bytes(64).encode(value),
-    (wire: BitVector) => bytes.decode(wire).map(_.map {
-      case bin64 if bin64.size == 64 => ByteVector64(bin64)
-      case der => Crypto.der2compact(der)
-    })
-  )
-
   val htlcTxAndSigsCodec: Codec[HtlcTxAndSigs] = (
     ("txinfo" | txWithInputInfoCodec) ::
-      ("localSig" | variableSizeBytes(uint16, sig64OrDERCodec)) :: // we store as variable length for historical purposes (we used to store as DER encoded)
-      ("remoteSig" | variableSizeBytes(uint16, sig64OrDERCodec))).as[HtlcTxAndSigs]
+      ("localSig" | lengthDelimited(bytes64)) :: // we store as variable length for historical purposes (we used to store as DER encoded)
+      ("remoteSig" | lengthDelimited(bytes64))).as[HtlcTxAndSigs]
 
   val publishableTxsCodec: Codec[PublishableTxs] = (
     ("commitTx" | (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec)).as[CommitTx]) ::
@@ -156,7 +138,7 @@ object ChannelCodecs extends Logging {
       ("txid" | bytes32) ::
       ("remotePerCommitmentPoint" | publicKey)).as[RemoteCommit]
 
-  val updateMessageCodec: Codec[UpdateMessage] = lightningMessageCodec.narrow(f => Attempt.successful(f.asInstanceOf[UpdateMessage]), g => g)
+  val updateMessageCodec: Codec[UpdateMessage] = lengthDelimited(lightningMessageCodec.narrow[UpdateMessage](f => Attempt.successful(f.asInstanceOf[UpdateMessage]), g => g))
 
   val localChangesCodec: Codec[LocalChanges] = (
     ("proposed" | listOfN(uint16, updateMessageCodec)) ::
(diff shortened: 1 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -60,6 +62,11 @@ object ChannelCodecs extends Logging {
     // field and don't support additional features which is why all bits are set to 0.
   )
 
+  /**
+   * byte-aligned boolean codec
+   */
+  val boolean: Codec[Boolean] = bool(8)
```
pm47, line 68: Not sure about the name, maybe I should go with `bool8` or even `byteAligned(bool)`

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -363,16 +323,25 @@ object ChannelCodecs extends Logging {
...
-    .typecase(0x08, DATA_WAIT_FOR_FUNDING_CONFIRMED_Codec)
-    .typecase(0x01, DATA_WAIT_FOR_FUNDING_CONFIRMED_COMPAT_01_Codec)
-    .typecase(0x02, DATA_WAIT_FOR_FUNDING_LOCKED_Codec)
-    .typecase(0x03, DATA_NORMAL_COMPAT_03_Codec)
-    .typecase(0x04, DATA_SHUTDOWN_Codec)
-    .typecase(0x05, DATA_NEGOTIATING_Codec)
-    .typecase(0x06, DATA_CLOSING_COMPAT_06_Codec)
-    .typecase(0x07, DATA_WAIT_FOR_REMOTE_PUBLISH_FUTURE_COMMITMENT_Codec)
+  val stateDataCodec: Codec[HasCommitments] = discriminated[HasCommitments].by(byte)
+    .typecase(1, discriminated[HasCommitments].by(uint16)
```
pm47, line 327: I've set a new version, wasn't really mandatory. We're relying on the same ordering trick (see comment) to take care of the migration.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -224,127 +230,81 @@ object ChannelCodecs extends Logging {
       ("localNextHtlcId" | uint64overflow) ::
       ("remoteNextHtlcId" | uint64overflow) ::
       ("originChannels" | originsMapCodec) ::
-      ("remoteNextCommitInfo" | either(bool, waitingForRevocationCodec, publicKey)) ::
+      ("remoteNextCommitInfo" | either(boolean, waitingForRevocationCodec, publicKey)) ::
       ("commitInput" | inputInfoCodec) ::
-      ("remotePerCommitmentSecrets" | ShaChain.shaChainCodec) ::
+      ("remotePerCommitmentSecrets" | byteAligned(ShaChain.shaChainCodec)) ::
```
pm47, line 235: The `shaChainCodec` isn't byte aligned, and there may be room for optimization there. cc @sstone

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -185,17 +192,16 @@ object ChannelCodecs extends Logging {
...
   val trampolineRelayedCodec: Codec[Origin.TrampolineRelayed] = (
     listOfN(uint16, bytes32 ~ int64) ::
       ("sender" | provide(Option.empty[ActorRef]))
     ).as[Origin.TrampolineRelayed]
 
   val originCodec: Codec[Origin] = discriminated[Origin].by(uint16)
-    .typecase(0x03, localCodec) // backward compatible
-    .typecase(0x01, provide(Origin.Local(UNKNOWN_UUID, None)))
     .typecase(0x02, relayedCodec)
+    .typecase(0x03, localCodec)
```
pm47, line 204: I removed the compatibility codec since the new codec will never use it.

pm47:
My plan for tests would be to generate a bunch of channel data before this patch (maybe with scalacheck generators @t-bast?) and their java object checksum.

I would then apply this change, read-then-write-then-read-again the data, and make sure the three checksums are equal.

I think that would bring solid assurance that we don't have a regression.

pm47:
So I validated the change by following the procedure decribed in my previous comment. I'm using an sqlite table with two columns: scodec-serialized and java-serialized, generated on a *pre-migration* branch. Then on the *post-migration* branch, I make sure that `read(scodec) == read(java)` and `read(write(read(scodec))) == read(java)`.

I ended up using three data sets:
- randomly generated data, using scalacheck here (code [here](https://gist.github.com/pm47/13f83a53e52d70d92994b8eae6f1599a)). Caveats: a few fields were empty or hardcoded (transactions, pending commitment changes).
- testnet and mainnet node data, using this code:
```scala
  test("nonreg length-delimited reference data generation") {
    val sqlite = DriverManager.getConnection(s"jdbc:sqlite:${new File("endurance_before.sqlite")}")
    sqlite.createStatement().executeUpdate("CREATE TABLE data (scodec BLOB NOT NULL, java BLOB NOT NULL)")

    val read_sqlite = DriverManager.getConnection("jdbc:sqlite::resource:eclair.sqlite.endurance.bak")
    val rs = read_sqlite.createStatement().executeQuery("SELECT data FROM local_channels")
    var count = 0
    while (rs.next()) {
      val scodec = BitVector(rs.getBytes("data"))
      val d = ChannelCodecs.stateDataCodec.decode(scodec).require.value
      val bos = new ByteArrayOutputStream()
      val oos = new ObjectOutputStream(bos)
      oos.writeObject(d)
      oos.close()
      val java = bos.toByteArray
      val statement = sqlite.prepareStatement("INSERT INTO data VALUES (?, ?)")
      statement.setBytes(1, scodec.toByteArray)
      statement.setBytes(2, java)
      statement.execute()
      count = count + 1
    }
    println(s"processed $count channels")
  }
```

This is the actual test:
```scala
test("nonreg length-delimited migration") {
    val sqlite = DriverManager.getConnection(s"jdbc:sqlite:${new File("endurance_before.sqlite")}")
    val rs = sqlite.createStatement().executeQuery("SELECT * FROM data")
    var count = 0
    while (rs.next()) {
      val scodec = ByteVector(rs.getBytes(1)).bits
      val java = rs.getBytes(2)
      val d_scodec = ChannelCodecs.stateDataCodec.decode(scodec).require.value match {
        case d: DATA_WAIT_FOR_FUNDING_CONFIRMED => d.copy(waitingSince = 42000)
        case d: DATA_CLOSING => d.copy(waitingSince = 123456789)
        case d => d
      }
      val bis = new ByteArrayInputStream(java)
      val ois = new ObjectInputStream(bis)
      val d_java = ois.readObject().asInstanceOf[HasCommitments] match {
        case d: DATA_WAIT_FOR_FUNDING_CONFIRMED => d.copy(waitingSince = 42000)
        case d: DATA_CLOSING => d.copy(waitingSince = 123456789)
        case d => d
      }
      assert(d_scodec === d_java)
      val scodec2 = ChannelCodecs.stateDataCodec.encode(d_scodec).require
      assert(scodec != scodec2)
      assert(scodec2.size % 8 == 0)
      val d_scodec2 = ChannelCodecs.stateDataCodec.decode(scodec2).require.value
      assert(d_scodec2 == d_java)
      count = count + 1
    }
    println(s"processed $count channels")
  }
```

It makes little sense committing the migration tests since they occur on two different versions of the code and they only apply at the time of the migration. I prefer documenting them here.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -224,133 +218,314 @@ object ChannelCodecs extends Logging {
...
-
-  // All channel_announcement's written prior to supporting unknown trailing fields had the same fixed size, because
-  // those are the announcements that *we* created and we always used an empty features field, which was the only
-  // variable-length field.
-  val noUnknownFieldsChannelAnnouncementSizeCodec: Codec[Int] = provide(430)
-
-  // We used to ignore unknown trailing fields, and assume that channel_update size was known. This is not true anymore,
-  // so we need to tell the codec where to stop, otherwise all the remaining part of the data will be decoded as unknown
-  // fields. Fortunately, we can easily tell what size the channel_update will be.
-  val noUnknownFieldsChannelUpdateSizeCodec: Codec[Int] = peek( // we need to take a peek at a specific byte to know what size the message will be, and then rollback to read the full message
```
pm47, line 287: Got rid of this compatibility code.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -184,18 +182,14 @@ object ChannelCodecs extends Logging {
       ("amountIn" | millisatoshi) ::
       ("amountOut" | millisatoshi)).as[Origin.Relayed]
 
-  // this is for backward compatibility to handle legacy payments that didn't have identifiers
-  val UNKNOWN_UUID = UUID.fromString("00000000-0000-0000-0000-000000000000")
```
pm47, line 188: Got rid of this compatibility code.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -126,19 +133,10 @@ object ChannelCodecs extends Logging {
     .typecase(0x09, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec)).as[HtlcPenaltyTx])
     .typecase(0x10, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec)).as[ClosingTx])
 
-  // this is a backward compatible codec (we used to store the sig as DER encoded), now we store it as 64-bytes
-  val sig64OrDERCodec: Codec[ByteVector64] = Codec[ByteVector64](
```
pm47, line 130: Got rid of this compatibility code.

pm47:
I also had this dilemma. I pushed the right button, but I'm really not sure: since the legacy codecs are frozen, shouldn't we remove the non-reg tests?

image

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
t-bast: I think `bool8` makes sense since we have `uint16`, `uint32`, etc.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
t-bast: Nice, that means we write with version 1, and legacy will fail the read with version 1 and fallback to version 0, that sounds good.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
t-bast: I think we should make it byte-aligned, but the question is whether to do it "outside" (what you just did), or inside the shachain codec (use `bool8` instead of `bool`)?

t-bast:
> shouldn't we remove the non-reg tests?

At least it doesn't hurt to keep them. The way I see it the goal is to completely remove the code of these legacy codecs (for example, X releases after the one where we release the new codecs). We can remove these tests at that point?

t-bast:
I think the testing strategy is sound, and at a high-level the code looks good. Could you squash/rebase and then we can review thoroughly?

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: It would have fallen back using the 2nd-level typecase anyway, the change of version is strictly non-necessary.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: Apparently we can be more optimal, the codec should naturally be byte-aligned. Since this is sensitive code I'd rather leave that optimisation to another PR.

pm47:
> At least it doesn't hurt to keep them.

There is a price though: we have several accessible codecs for the same object types, which is a bit error-prone. But we don't use those codecs that often and the risk of using a legacy one by mistake is quite low.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: Done with 06207d1.

pm47:
> Could you squash/rebase and then we can review thoroughly?

Done. I need to re-run the tests. It just occured to me that #1141 had touched the codecs, so a bit more testing is probably not a bad thing.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
```diff
@@ -312,7 +316,7 @@ class ChannelCodecsSpec extends AnyFunSuite {
     // and re-encode it with the new codec
     val bin_new = ByteVector(stateDataCodec.encode(data_new).require.toByteVector.toArray)
     // data should now be encoded under the new format, with version=0 and type=8
-    assert(bin_new.startsWith(hex"000008"))
+    assert(bin_new.startsWith(hex"010020"))
```
t-bast, line 319: The comment above should be updated.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
```diff
@@ -355,7 +359,7 @@ class ChannelCodecsSpec extends AnyFunSuite {
       // and we encode with new codec
       val newbin = stateDataCodec.encode(oldnormal).require.bytes
       // make sure that encoding used the new 0x10 codec
-      assert(newbin.startsWith(hex"000010"))
+      assert(newbin.startsWith(hex"010022"))
```
t-bast, line 362: Same, the comment should be updated.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
t-bast: Got it, indeed since the typecase are different it works without the version.
But it clearly separates the two groups, I think it makes it clearer that it's a milestone/major change.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -26,22 +26,26 @@ import fr.acinq.eclair.crypto.ShaChain
...
 
 /**
  * Created by PM on 02/06/2017.
  */
 object ChannelCodecs extends Logging {
 
+  /**
+   * All LN protocol message must be stored as length-delimited, because they may have arbitrary trailing data
+   */
+  def lengthDelimited[T](codec: Codec[T]): Codec[T] = variableSizeBytes(uint16, codec)
```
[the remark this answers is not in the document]
pm47: `uint16` is enough for lightning messages, but you're right that `varint` is both more optimal and more future proof. This will remove some factorization between old/new codecs, but I guess that's not really important.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
pm47: Done with cd36a9f.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
pm47: Done with cd36a9f.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
```diff
@@ -230,13 +232,15 @@ class ChannelCodecsSpec extends AnyFunSuite {
...
-    assert(originCodec.decodeValue(hex"0001 0123456789abcdef0123456789abcdef".bits).require === Local(UNKNOWN_UUID, None))
+    assert(codec.decodeValue(codec.encode(Local(id, Some(ActorSystem("test").deadLetters))).require).require === Local(id, None))
+    val ZERO_UUID = UUID.fromString("00000000-0000-0000-0000-000000000000")
+    assert(codec.decodeValue(hex"0001 0123456789abcdef0123456789abcdef".bits).require === Local(ZERO_UUID, None))
     val relayed = Relayed(randomBytes32, 4324, 12000000 msat, 11000000 msat)
-    assert(originCodec.decodeValue(originCodec.encode(relayed).require).require === relayed)
+    assert(codec.decodeValue(codec.encode(relayed).require).require === relayed)
     val trampolineRelayed = TrampolineRelayed((randomBytes32, 1L) :: (randomBytes32, 1L) :: (randomBytes32, 2L) :: Nil, None)
-    assert(originCodec.decodeValue(originCodec.encode(trampolineRelayed).require).require === trampolineRelayed)
+    assert(codec.decodeValue(codec.encode(trampolineRelayed).require).require === trampolineRelayed)
```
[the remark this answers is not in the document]
pm47: Done with cd36a9f.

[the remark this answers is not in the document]
pm47: Done with cd36a9f.

pm47:
6e3414bfc was less straightforward than I expected. Being in an inner object of the normal codecs, the legacy codecs have transparent access to all codecs which leads to hidden dependencies. I realize now that copy-pasting the old file would have been the safest approach. I started like that but reverted because it prevented me from sharing some codecs... which I now see wasn't a good idea. WDYT?

I've regenerated the random data to support static remote key and successfully re-ran the migration tests.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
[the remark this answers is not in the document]
pm47: Fixed with 0b956a4e2.

pm47:
Tried a different isolation with d052666ab, and re-ran the migrations tests. I think that design is better.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: We will indeed always read with the legacy backward-compatible codec and write with the new codec. There is no need for the new codec to support legacy encoded data.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -32,16 +32,23 @@ import grizzled.slf4j.Logging
 import scodec.bits.BitVector
 import scodec.codecs._
 import scodec.{Attempt, Codec}
-import shapeless.{HNil, ::}
```
t-bast, line 35: nit: there are other unused imports you can clean up

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -32,16 +32,23 @@ import grizzled.slf4j.Logging
...
 object ChannelCodecs extends Logging {
 
+  /**
+   * All LN protocol message must be stored as length-delimited, because they may have arbitrary trailing data
+   */
+  def lengthDelimited[T](codec: Codec[T]): Codec[T] = variableSizeBytesLong(varintoverflow, codec)
+
+  def mapCodec[K, V](keyCodec: Codec[K], valueCodec: Codec[V]): Codec[Map[K, V]] = listOfN(uint16, keyCodec ~ valueCodec).xmap(_.toMap, _.toList)
+
+  def setCodec[T](codec: Codec[T]): Codec[Set[T]] = listOfN(uint16, codec).xmap(_.toSet, _.toList)
```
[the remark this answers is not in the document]
pm47: Done with b90a9e078.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: Done with b90a9e078.

PR #1443, opened 2020-06-03
Routing hints balance

t-bast:
This is a collection of several small fixes/improvements:

- Ignore routing hints for our own channels
- Assume a bigger capacity than just the HTLC amount in routing hints
- Correctly handle balance change events for private channels

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -109,7 +109,7 @@ object RouteCalculation {
     // should be able to route the payment, so we'll compute an htlcMaximumMsat accordingly.
     // We could also get the channel capacity from the blockchain (since we have the shortChannelId) but that's more expensive.
     // We also need to make sure the channel isn't excluded by our heuristics.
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -109,7 +109,7 @@ object RouteCalculation {
     // should be able to route the payment, so we'll compute an htlcMaximumMsat accordingly.
     // We could also get the channel capacity from the blockchain (since we have the shortChannelId) but that's more expensive.
     // We also need to make sure the channel isn't excluded by our heuristics.
-    val lastChannelCapacity = amount.max(RoutingHeuristics.CAPACITY_CHANNEL_LOW)
+    val lastChannelCapacity = (amount * 2).max(RoutingHeuristics.CAPACITY_CHANNEL_LOW)
```
pm47, line 112: I'm not sure about this, except if we have a real reason behind this change.

t-bast: This indeed shouldn't be necessary, I reverted it in https://github.com/ACINQ/eclair/pull/1443/commits/aaf17b6572a9500e5edb2fed1c1917579fd9e98a

PR #1445, opened 2020-06-04
MPP preimage received event

t-bast:
The MultipartPaymentLifecycle FSM sends a message as soon as it receives the preimage.

This allows removing the use of the node-relayer as a passthrough for fulfills, it can now simply listen to this event.

Long term, this could be sent to the event stream to share with more actors.


t-bast:
Done in https://github.com/ACINQ/eclair/pull/1445/commits/81abeb59c72241e3fd037005e3a7ebd1befea4c2

PR #1446, opened 2020-06-05
Fix edges with empty capacity in the routing graph

t-bast:
Light clients don't always validate channels by fetching the blockchain tx.
That means they don't have access to the exact capacity.
When that happens, we should fallback to `htlc_maximum_msat` if available, or to a default capacity (otherwise path-finding will ignore these edges).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -393,14 +393,7 @@ object Graph {
       def addEdge(edge: GraphEdge): DirectedGraph = {
         val vertexIn = edge.desc.a
         val vertexOut = edge.desc.b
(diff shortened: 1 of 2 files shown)


sstone:
LGTM, but it's a bit difficult to test as I don't think it's covered by our unit tests yet (on master capacity is correct as we get it from the blockchain, on android it's not but "A->B->C" type tests are disabled as we don't relay payments which is why this issue was not detected before, and on phoenix it's not relevant).
I'll test on android and add a comment.

t-bast:
Actually this doesn't cover all the way we can create the graph. Wait before testing on android, I need to add another commit to cover everything.

t-bast:
You should be good now, I was missing the helper method we use to create the graph initially from the DB.

sstone:
I checked that it does fix route computation on Android.

PR #1450, opened 2020-06-08
Add a fee provider getting feerates from database

### LINKED ISSUE #1447: Add a `SqliteFeeProvider`
Similarly to the `SmoothFeeProvider`, it would take an underlying fee provider, and would store the last known feerate. That would allow a faster startup on Android.

Would be defined like so:

```scala
class SqliteFeeProvider(sqlite: Connection, provider: FeeProvider)(implicit ec: ExecutionContext) extends FeeProvider {
...
}
```

dpad85:
This PR adds a `DbFeeProvider` wrapper that will first try to retrieve the feerates from a database, and then fallback on the wrapped fee provider.

This is to be used when we want to quickly start the node, for example on mobile devices.

This PR does not modify the `Setup`, since this new provider will probably only be used on the mobile branch.

Fixes #1447

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
@@ -45,6 +45,8 @@ class BitgoFeeProvider(chainHash: ByteVector32, readTimeOut: Duration)(implicit 
         .send()
       feeRanges = parseFeeRanges(res.unsafeBody)
     } yield extractFeerates(feeRanges)
+
+  override def getName: String = "Bitgo"
 }
 
 object BitgoFeeProvider {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
@@ -46,7 +46,6 @@ class BitgoFeeProvider(chainHash: ByteVector32, readTimeOut: Duration)(implicit 
       feeRanges = parseFeeRanges(res.unsafeBody)
     } yield extractFeerates(feeRanges)
 
-  override def getName: String = "Bitgo"
 }
 
 object BitgoFeeProvider {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
@@ -21,7 +21,7 @@ import fr.acinq.eclair.db.FeeratesDb
 import scala.concurrent.{ExecutionContext, Future}
 
 /**
- * This wrapper retrieves the feerates from the a database for a given provider the first time it's used, then fallbacks
+ * This wrapper retrieves the feerates from the database for a given provider the first time it's used, then fallbacks
  * to the wrapped provider's actual `getFeeRates` future.
  */
 class DbFeeProvider(db: FeeratesDb, provider: FeeProvider)(implicit ec: ExecutionContext) extends FeeProvider {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 ACINQ SAS
+ * Copyright 2020 ACINQ SAS
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,24 +18,31 @@ package fr.acinq.eclair.blockchain.fee
 
 import fr.acinq.eclair.db.FeeratesDb
 
-import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
 
 /**
- * This provider first tries to retrieve the feerates from the a database acting as a cache.
- *
- * If no provider already exists for this provider's name, or if the feerates age exceeds the provided max age,
- * then the provider's actual `getFeeRates` future is called ; subsequent result is saved to the database.
+ * This wrapper retrieves the feerates from the a database for a given provider the first time it's used, then fallbacks
+ * to the wrapped provider's actual `getFeeRates` future.
  */
-class DbFeeProvider(db: FeeratesDb, provider: FeeProvider, maxAgeMillis: Long = 1.day.toMillis)(implicit ec: ExecutionContext) extends FeeProvider {
+class DbFeeProvider(db: FeeratesDb, provider: FeeProvider)(implicit ec: ExecutionContext) extends FeeProvider {
 
-  override def getFeerates: Future[FeeratesPerKB] = db.getFeerates(provider.getName) match {
-    case Some((feeratesInDb, timestamp)) if System.currentTimeMillis() - timestamp <= maxAgeMillis =>
-      Future.successful(feeratesInDb)
-    case _ => provider.getFeerates.flatMap { f =>
-      db.addOrUpdateFeerates(provider.getName, f)
-      Future.successful(f)
+  /** This boolean represents the state of this provider */
+  private var isFirstCall = true
+
+  /** This method should use the database once, and then fallback to the wrapped provider. */
+  override def getFeerates: Future[FeeratesPerKB] = if (isFirstCall) {
+    db.getFeerates(provider.getName) match {
+      case Some(feeratesInDb) =>
+        isFirstCall = false
+        Future.successful(feeratesInDb)
+      case _ => provider.getFeerates.flatMap { f =>
+        isFirstCall = false
+        db.addOrUpdateFeerates(provider.getName, f)
+        Future.successful(f)
+      }
     }
+  } else {
+    provider.getFeerates
   }
 
   override def getName: String = provider.getName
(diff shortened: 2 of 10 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
```diff
@@ -0,0 +1,43 @@
...
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+ * This provider first tries to retrieve the feerates from the a database acting as a cache.
+ *
+ * If no provider already exists for this provider's name, or if the feerates age exceeds the provided max age,
+ * then the provider's actual `getFeeRates` future is called ; subsequent result is saved to the database.
+ */
+class DbFeeProvider(db: FeeratesDb, provider: FeeProvider, maxAgeMillis: Long = 1.day.toMillis)(implicit ec: ExecutionContext) extends FeeProvider {
```
t-bast, line 30: Why not use a `Duration` for `maxAge`? More typed is good :)

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/fee/DbFeeProviderSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+      }
+    }, 1)
+
+    // first call: db empty
+    assert(db.getFeerates("foo").isEmpty)
+    assert(Await.result(provider.getFeerates, Timeout(30 seconds).duration) == feerates1)
+    assert(db.getFeerates("foo").get._1 == feerates1)
+
+    // second call: data from db is already obsolete and should be ignored
+    Thread.sleep(5)
```
t-bast, line 74: Do you really need the `Thread.sleep`?
Why not a non-blocking `awaitCond()` instead?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
t-bast, line 30: Shouldn't we do this the other way around?
First try to get feerate from the provider. If it fails, fetch from the DB. If the DB entry is obsolete, fail.
This would ensure we have more fresh entries in the DB, because with the current code we will only try to fetch from the provider once the db entry is obsolete, so if that single call fails we're screwed...

t-bast: If the main goal is the faster start-up on android, it makes sense to first query the DB, but then even if the DB call succeeds I would do the call to the provider to ensure we have a more recent entry in the DB for next time?

dpad85: In fact I changed this PR with 8651b04, this provider is not a cache anymore, the value in database will only be used once, and then it falls back to the wrapped provider. The only point of this PR is to have a fast startup, which happens only once, so a caching system is not useful.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/fee/DbFeeProviderSpec.scala
dpad85: This is now resolved, see https://github.com/ACINQ/eclair/pull/1450#discussion_r436678682

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
dpad85: This is now resolved, see https://github.com/ACINQ/eclair/pull/1450#discussion_r436678682

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package fr.acinq.eclair.blockchain.fee
+
+import fr.acinq.eclair.db.FeeratesDb
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+ * This wrapper retrieves the feerates from the a database for a given provider the first time it's used, then fallbacks
```
t-bast, line 24:
```suggestion
 * This wrapper retrieves the feerates from the database for a given provider the first time it's used, then fallbacks
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
```diff
@@ -0,0 +1,50 @@
...
+        isFirstCall = false
+        Future.successful(feeratesInDb)
+      case _ => provider.getFeerates.flatMap { f =>
+        isFirstCall = false
+        db.addOrUpdateFeerates(provider.getName, f)
+        Future.successful(f)
+      }
+    }
+  } else {
+    provider.getFeerates
```
t-bast, line 45: I think we should also take this opportunity to update the db with the latest feerates each time we retrieve them, shouldn't we?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
```diff
@@ -0,0 +1,50 @@
...
+  private var isFirstCall = true
+
+  /** This method should use the database once, and then fallback to the wrapped provider. */
+  override def getFeerates: Future[FeeratesPerKB] = if (isFirstCall) {
+    db.getFeerates(provider.getName) match {
+      case Some(feeratesInDb) =>
+        isFirstCall = false
+        Future.successful(feeratesInDb)
+      case _ => provider.getFeerates.flatMap { f =>
+        isFirstCall = false
```
t-bast, line 39: This isn't thread-safe! We should either use an `Atomic` or change the design, let me think about it a bit more...

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
```diff
@@ -45,6 +45,8 @@ class BitgoFeeProvider(chainHash: ByteVector32, readTimeOut: Duration)(implicit
         .send()
       feeRanges = parseFeeRanges(res.unsafeBody)
     } yield extractFeerates(feeRanges)
+
+  override def getName: String = "Bitgo"
```
pm47, line 49: I find it weird that the implementation of one particular provider has impacts on other providers. I'd rather remove the `providerName` from the db.

If one day we want to combine multiple providers into one `MultiProvider` (say we want to do averages or something), then this new provider will probably have to use that kind of things internally. But it's a different concern, and even in that case other providers would be left untouched.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/SmoothFeeProvider.scala
```diff
@@ -34,6 +34,8 @@ class SmoothFeeProvider(provider: FeeProvider, windowSize: Int)(implicit ec: Exe
       _ = append(rate)
     } yield SmoothFeeProvider.smooth(queue)
   }
+
+  override def getName: String = provider.getName
```
pm47, line 38: If we decide to keep provider's name, this should be more specific
```suggestion
  override def getName: String = s"Smooth(${provider.getName})"
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
pm47: And it doesn't do what we want I think. How about we reimplement it that way:

```scala
class DbFeeProvider(db: FeeratesDb, provider: FeeProvider)(implicit ec: ExecutionContext) extends FeeProvider {
  
  /** This method retrieves feerates from the provider, and store results in the database */
  override def getFeerates: Future[FeeratesPerKB] =
    provider.getFeerates map { feerates =>
        db.addOrUpdateFeerates(provider.getName, feerates)
      feerates
    }

  def getLastKnownFeerates: Option[FeeratesPerKB] = db.getFeerates(provider.getName)

  override def getName: String = provider.getName

}
```

There is no need to try to shoe in eveything under the `getFeerates` method. It's easy to use from the mobile `Setup`: if `getLastKnownFeerates` returns `Some(feerates)` then initialize global feerates with these, otherwise do a blocking call to `getFeerates` (which we always do currently).

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BitgoFeeProvider.scala
dpad85: The feerates is used throughout the app and is an important piece of data, and with all the inheritance going on with providers we don't know where it's taken from. Having a name fixes that.

However by adding a `getName` method, all providers must then implement it including those that do not actually get a feerates but only add some intelligence to it, like `SmoothFeeProvider` or this `DbFeeProvider`, where the name is irrelevant. It could be improved by reworking the design of the providers but it seems out of the scope of this PR. Or maybe there's a smarter way.

Anyway I think it should be done in another PR, we can remove the provider name from the db and from the providers.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
pm47: True, but it feels like an encapsulation violation. Not a big deal.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/DbFeeProvider.scala
dpad85: Fixed with 02801fb0a4cf39d4b27566143d61f04374d4f29a 👍

PR #1452, opened 2020-06-09
Scale fulfill safety window based on amount at risk

t-bast:
When an upstream peer is not responsive to htlc fulfills, there is a risk that they will try to cheat us by force-closing the channel and redeeming the htlc-timeout on-chain.

To avoid that, we previously used a constant security window of 6 blocks to force-close the channel ourselves (which gives us time to redeem the htlc-success on-chain).

Instead of using a constant window, it may make sense to scale it based on the amount at risk. Coupled with the `max-htlc-value-in-flight-msat` restriction, it provides a good adaptive defence against those attacks.

Let me know if you think this is worth the added complexity: an easier solution since we use a high `cltv_expiry_delta` is to simply always use a much higher `fulfill-safety-before-timeout-blocks`. With a `cltv_expiry_delta` of 144, we can safely set `fulfill-safety-before-timeout-blocks` to 24 or even 48.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -44,6 +44,37 @@ import scala.util.{Failure, Success, Try}
 object Helpers {
 
   /**
+   * When we receive the preimage for an HTLC and want to fulfill it but the upstream peer stops responding, we want to
+   * avoid letting its HTLC-timeout transaction become enforceable on-chain (otherwise there is a race condition between
+   * our HTLC-success and their HTLC-timeout).
+   * We will close the channel when the HTLC-timeout will happen in less than N blocks, where we scale N linearly depending
+   * on the total amount at risk (there may be multiple HTLCs in that state).
+   *
+   * @param min           if incoming HTLCs expire in less than `min`, we close the channel regardless of the amount at risk.
+   * @param max           incoming HTLCs that expire in more than `max` are ignored.
+   * @param deltaPerBlock linear scaling factor for the safety window.
+   */
+  case class FulfillSafetyBeforeTimeout(min: CltvExpiryDelta, max: CltvExpiryDelta, deltaPerBlock: MilliSatoshi) {
+    require(min >= CltvExpiryDelta(3), "at least 3 blocks are necessary to enforce a fulfill safety window")
+    require(min <= max, "fulfill safety minimum must be lower than maximum")
+
+    /**
+     * @param htlcsAtRisk all htlcs in a given channel that expire soon, for which we know the preimage but couldn't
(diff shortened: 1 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -43,6 +43,37 @@ import scala.util.{Failure, Success, Try}
...
+   * avoid letting its HTLC-timeout transaction become enforceable on-chain (otherwise there is a race condition between
+   * our HTLC-success and their HTLC-timeout).
+   * We will close the channel when the HTLC-timeout will happen in less than N blocks, where we scale N linearly depending
+   * on the total amount at risk (there may be multiple HTLCs in that state).
+   *
+   * @param min           if incoming HTLCs expire in less than `min`, we close the channel regardless of the amount at risk.
+   * @param max           incoming HTLCs that expire in more than `max` are ignored.
+   * @param deltaPerBlock linear scaling factor for the safety window.
+   */
+  case class FulfillSafetyBeforeTimeout(min: CltvExpiryDelta, max: CltvExpiryDelta, deltaPerBlock: MilliSatoshi) {
```
pm47, line 57: I would have put this somewhere else, maybe right below the `aboveReserve` method.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -43,6 +43,37 @@ import scala.util.{Failure, Success, Try}
...
+     * @param htlcsAtRisk all htlcs in a given channel that expire soon, for which we know the preimage but couldn't
+     *                    remove from the commitment (most likely because our remote is malicious or unresponsive).
+     * @return true if the funds at risk are too great and we should close the channel to safely redeem HTLCs on-chain.
+     */
+    def shouldClose(blockHeight: Long, htlcsAtRisk: Set[UpdateAddHtlc]): Boolean = {
+      htlcsAtRisk.toSeq.filter(_.cltvExpiry.toLong <= blockHeight + max.toInt).sortBy(_.cltvExpiry).foldLeft((0 msat, false)) {
+        case ((amountAtRisk, true), _) => (amountAtRisk, true)
+        case ((amountAtRisk, false), htlc) =>
+          val amount2 = amountAtRisk + htlc.amountMsat
+          val shouldClose = deltaPerBlock * (htlc.cltvExpiry.toLong - blockHeight - min.toInt).max(0) < amount2
```
pm47, line 71: I don't understand this formula.

t-bast:
After discussing that offline, we realized that the added complexity may not be worth it.
It's very likely better to instead always use the higher bound and increase `cltv_expiry_delta` accordingly if needed.
I will open a different PR that increases these limits and makes the correlation between them more explicit.

PR #1453, opened 2020-06-10
Monitor on-chain feerate mismatch

t-bast:
This is a first step before actually deciding how we can improve our tolerance to remote feerate mismatch.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -21,6 +21,7 @@ import fr.acinq.eclair.channel.ChannelVersion._
 import fr.acinq.bitcoin.Crypto.{PrivateKey, PublicKey, sha256}
 import fr.acinq.bitcoin.{ByteVector32, ByteVector64, Crypto}
 import fr.acinq.eclair.blockchain.fee.{FeeEstimator, FeeTargets}
+import fr.acinq.eclair.channel.Monitoring.Metrics
 import fr.acinq.eclair.crypto.{Generators, KeyManager, ShaChain, Sphinx}
 import fr.acinq.eclair.payment.relay.{Origin, Relayer}
 import fr.acinq.eclair.transactions.DirectedHtlc._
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -389,13 +390,15 @@ object Commitments {
...
   }
 
-  def receiveFee(commitments: Commitments, feeEstimator: FeeEstimator, feeTargets: FeeTargets, fee: UpdateFee, maxFeerateMismatch: Double): Try[Commitments] = {
+  def receiveFee(commitments: Commitments, feeEstimator: FeeEstimator, feeTargets: FeeTargets, fee: UpdateFee, maxFeerateMismatch: Double)(implicit log: LoggingAdapter): Try[Commitments] = {
     if (commitments.localParams.isFunder) {
       Failure(FundeeCannotSendUpdateFee(commitments.channelId))
     } else if (fee.feeratePerKw < fr.acinq.eclair.MinimumFeeratePerKw) {
       Failure(FeerateTooSmall(commitments.channelId, remoteFeeratePerKw = fee.feeratePerKw))
     } else {
+      Metrics.RemoteFeeratePerKw.withoutTags().record(fee.feeratePerKw)
```
pm47, line 399: How will this make sense if we don't index by channel? (which we can't because of cardinality)

t-bast: Exactly, we can't index by channel so we just get a histogram of the overall mismatch.
The goal is to check how many channels we'd lose if we changed our tolerance to mismatch by N.
Then there's the added log line to check in details the mismatch of each channel.

t-bast: How would you do it differently?

pm47: I missed that you were using a `histogram` vs a `gauge` for the local feerate. We'll be able to display the distribution of values, which alleviates the issue I was describing. Maybe we'll be able to group between various fee providers.

PR #1454, opened 2020-06-10
Fix transaction fee test

sstone:
- Test was not executed (because the "tests" variable was an iterator that was emptied by the call to .size())
- HTLC regex had to be updated to skip over the HTLC number that was added to the reference test vectors


t-bast:
:facepalm: 

I'm not sure fetching the test vectors from the spec repo is a good idea.
To be honest these test vectors will change very rarely, and when a spec PR proposes changes we will integrate those changes into our test suite.
Why not extract from Bolt 3 the test vectors into a test? That removes the need for error-prone regex-based parsing (at the cost of a one-time human parsing).

PR #1455, opened 2020-06-11
Clarify trampoline route not found error

t-bast:
The payment initiator retries trampoline payments by raising the fee/ctlv when the trampoline node couldn't find a route.
If we exhaust all attempts and the trampoline node keep returning a fee insufficient error, it actually means `RouteNotFound`, so we should translate the error for the end user (otherwise it's misleading, they think they should raise the fee but it's a red herring).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
@@ -29,6 +29,7 @@ import fr.acinq.eclair.payment._
 import fr.acinq.eclair.payment.send.MultiPartPaymentLifecycle.SendMultiPartPayment
 import fr.acinq.eclair.payment.send.PaymentError._
 import fr.acinq.eclair.payment.send.PaymentLifecycle.{SendPayment, SendPaymentToRoute}
+import fr.acinq.eclair.router.RouteNotFound
 import fr.acinq.eclair.router.Router.{Hop, NodeHop, Route, RouteParams}
 import fr.acinq.eclair.wire.Onion.FinalLegacyPayload
 import fr.acinq.eclair.wire._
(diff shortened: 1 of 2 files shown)


dpad85:
Rewriting trampoline errors in the core is ok because it avoids tightly coupling the "front" and the "back", while trampoline is still a work in progress and subject to changes.

I think that once things are more stable, the core should just pipe errors to the front, and it's the duty of the front to provide pertinent message to the end user.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -81,16 +82,28 @@ class PaymentInitiator(nodeParams: NodeParams, router: ActorRef, register: Actor
...
-          pp.sender ! pf
-          context.system.eventStream.publish(pf)
-          context become main(pending - pf.id)
+      val shouldRetry = decryptedFailures.contains(TrampolineFeeInsufficient) || decryptedFailures.contains(TrampolineExpiryTooSoon)
+      if (shouldRetry) {
+        pp.remainingAttempts match {
+          case (trampolineFees, trampolineExpiryDelta) :: remainingAttempts =>
+            log.info(s"retrying trampoline payment with trampoline fees=$trampolineFees and expiry delta=$trampolineExpiryDelta")
+            sendTrampolinePayment(pf.id, pp.r, trampolineFees, trampolineExpiryDelta)
+            context become main(pending + (pf.id -> pp.copy(remainingAttempts = remainingAttempts)))
```
dpad85, line 91: nit: `remainingAttempts = remainingAttempts` confused me a bit for a while. `remainingAttempts` could be renamed to `remainingTail` or something like that.

t-bast: Right, the same naming is a bit confusing. Updated in https://github.com/ACINQ/eclair/pull/1455/commits/3effac53b8ef5b33aefa4aa02d077035add75285

PR #1456, opened 2020-06-12
Add friendly network name to GetInfo response

t-bast:
This is friendlier to interpret than mapping this from the chain's genesis block hash.


eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -131,6 +131,15 @@ object NodeParams {
     }
   }
 
+  def chainFromHash(chainHash: ByteVector32): String = {
```
pm47, line 134: Could be refactored to:

```scala
  private val chain2Hash: Map[String, ByteVector32] = Map(
  "regtest" -> Block.RegtestGenesisBlock.hash,
  "testnet" -> Block.TestnetGenesisBlock.hash,
  "mainnet" -> Block.LivenetGenesisBlock.hash
  )

  def makeChainHash(chain: String): ByteVector32 = chain2Hash.getOrElse(chain, throw new RuntimeException(s"invalid chain '$chain'"))

  def chainFromHash(chainHash: ByteVector32): String = chain2Hash.map(_.swap).getOrElse(chainHash, throw new RuntimeException(s"invalid chainHash '$chainHash'"))
```

pm47: Also maybe rename `makeChainHash`->`hashFromChain` for symmetry.

t-bast: Done!

Issue #1459, opened 2020-06-15
On-chain list transactions API


ShahanaFarooqui:
In order to offer better on-chain insights and features to the user, we request to add on-chain list transactions API. This is good to have as an API. [LND](https://api.lightning.community/#v1-transactions) already provides it but [C-Lightning-REST](https://github.com/Ride-The-Lightning/c-lightning-REST/issues/39) has an open issue and will be introducing it soon.

Request Params:
start height
end height 

Response:
block hash on which the transaction is included
transaction hash
date/time of the transaction
block height
number of confirmations
total fees
amount
destination addresses

Sample UI designs:
![image](https://user-images.githubusercontent.com/23240721/84718473-a8a24600-af46-11ea-8da6-356b207b68b9.png)

![image](https://user-images.githubusercontent.com/23240721/84718612-05056580-af47-11ea-8b18-2f154abcaf31.png)

t-bast:
`bitcoind` doesn't offer filtering transactions by `start_height` / `end_height` (see [listtransactions](https://bitcoincore.org/en/doc/0.20.0/rpc/wallet/listtransactions/)).
It only offers pagination using `count` and `skip`, going downwards from the latest transactions.

We have two possibilities:

- we expose filtering only by `count` and `skip` like `bitcoind` does: simplest for us, is this something that could work for the UI side?
- we try to match given `start` / `end` height by dichotomy making calls to `bitcoind`: this is likely going to be somewhat inefficient

What do you prefer?

ShahanaFarooqui:
Hi @t-bast no preference, whatever works or is simpler for you.
Thanks

Issue #1458, opened 2020-06-15
On-chain send funds API


ShahanaFarooqui:
In order to offer better on-chain insights and features to the user, we request to add on-chain send funds API.

Request Params:
bitcoin address - mandatory
amount - mandatory
confirmation blocks - mandatory if fee rate is missing
fee rate - mandatory if confirmation blocks is missing

Response:
transaction id/error details

Sample UI design:
![image](https://user-images.githubusercontent.com/23240721/84654433-2cbfe380-aedd-11ea-9bb4-4f9a2f5b1700.png)

t-bast:
`bitcoind`'s [sendtoaddress](https://bitcoincore.org/en/doc/0.20.0/rpc/wallet/sendtoaddress/) supports setting the confirmation blocks, but not the mode where you set the feerate.

I really don't want eclair to implement the wallet logic of creating a raw transaction based on user's feerate, is it ok if we only implement the "confirmation blocks" option? Otherwise I can do a reverse lookup and transform an incoming "feerate" into a "confirmation blocks" I would give to `bitcoind`, but that means I wouldn't be honoring exactly the user's requested feerate, so I don't know if that would be ok.

saubyk:
Hi @t-bast , the above suggestion for fee-rate/target blocks option is based on the features available on both LND and C-Lightning implementations.
You can decide to offer just one, if you feel that keeps Eclair implementation simple. I am wondering if similar logic is applied in Eclair for channel opening with the `fundingFeerateSateByte` option.

I can tell you from experience and especially for opening channels the feature of specifying sats/byte is pretty popular. As people like the flexibility of specifying low rates and 1 sat/byte is an especially popular rate. :-)

t-bast:
> I am wondering if similar logic is applied in Eclair for channel opening with the fundingFeerateSateByte option

This is because we're building the raw funding transaction ourselves that we're able to choose the exact feerate, but I don't want to support that for arbitrary send transactions (implementing a good enough Bitcoin wallet is a full project in itself, I want eclair to do one thing and do it right and focus on the lightning side of things :slightly_smiling_face:).

We'll start with support for `confirmation_blocks` only for now, and we'll see if we revisit that later!

saubyk:
Hi @t-bast understood. Thanks.

PR #1461, opened 2020-06-16
Add on-chain APIs

### LINKED ISSUE #1457: On-chain get balance API
In order to offer better on-chain insights and features to the user, we request to add on-chain balance API.

Request Params:
None

Response:
Confirmed Balance
Unconfirmed Balance
Total Balance*

*good to have, can calculate otherwise :). Just added to maintain standard response from all LNPs.

Sample UI Design:
image

t-bast:
This is a collection of independent commits.
The goal is to add some basic on-chain APIs (that simply forward to Bitcoin Core) to provide a good integration into [RTL](https://github.com/Ride-The-Lightning/RTL).

We clearly don't want eclair to become a wallet for bitcoin core, but adding basic pass-through APIs can be useful for end-users.

Fixes #1457, #1458 and #1459.

### DIFF
--- a/eclair-core/eclair-cli
+++ b/eclair-core/eclair-cli
@@ -26,19 +26,58 @@ where OPTIONS can be:
   -h                    Show this help
   -s                    Some commands can print a trimmed JSON
 
-and COMMAND is one of:
-  getinfo, connect, disconnect, open, close, forceclose, updaterelayfee,
-  peers, channels, channel, allnodes, allchannels, allupdates
-  findroute, findroutetonode, parseinvoice, payinvoice, sendtonode,
-  sendtoroute, getsentinfo, createinvoice, getinvoice, listinvoices,
-  listpendinginvoices, getreceivedinfo, audit, networkfees,
-  channelstats, usablebalances
+and COMMAND is one of the available commands:
+
+  === Node ===
+    - getinfo
+    - connect
+    - disconnect
+    - peers
+    - allnodes
+    - audit
+
+  === Channel ===
+    - open
+    - close
+    - forceclose
+    - channel
+    - channels
+    - allchannels
+    - allupdates
+    - channelstats
+    - networkfees
+    - updaterelayfee
+
+  === Path-finding ===
+    - findroute
+    - findroutetonode
+    - networkstats
+
+  === Invoice ===
+    - createinvoice
+    - getinvoice
+    - listinvoices
+    - listpendinginvoices
+    - parseinvoice
+
+  === Payment ===
+    - getnewaddress
+    - usablebalances
+    - onchainbalances
+    - payinvoice
+    - sendtonode
+    - sendtoroute
+    - sendonchain
+    - getsentinfo
+    - getreceivedinfo
+    - listtransactions
 
 Examples
 --------
-  eclair-cli -a localhost:1234 peers        list the peers of a node hosted on localhost:1234
-  eclair-cli close --channelId=006fb...     closes the channel with id 006fb...
-
+  eclair-cli -a localhost:1234 peers                  list the peers of a node hosted on localhost:1234
+  eclair-cli connect --nodeId=03864e...               connect to node with id 03864e...
+  eclair-cli open --nodeId=... --fundingSatoshis=...  open a channel to a given node
+  eclair-cli close --channelId=006fb...               closes the channel with id 006fb...
 
 Full documentation here: <https://acinq.github.io/eclair>" 1>&2;
 exit 1;
--- a/eclair-core/eclair-cli
+++ b/eclair-core/eclair-cli
@@ -63,14 +63,14 @@ and COMMAND is one of the available commands:
   === Payment ===
     - getnewaddress
     - usablebalances
-    - onchainbalances
+    - onchainbalance
     - payinvoice
     - sendtonode
     - sendtoroute
     - sendonchain
     - getsentinfo
     - getreceivedinfo
-    - listtransactions
+    - onchaintransactions
 
 Examples
 --------
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -26,6 +26,7 @@ import fr.acinq.bitcoin.{ByteVector32, Satoshi}
 import fr.acinq.eclair.TimestampQueryFilters._
 import fr.acinq.eclair.blockchain.OnChainBalance
 import fr.acinq.eclair.blockchain.bitcoind.BitcoinCoreWallet
+import fr.acinq.eclair.blockchain.bitcoind.BitcoinCoreWallet.WalletTransaction
 import fr.acinq.eclair.channel.Register.{Forward, ForwardShortId}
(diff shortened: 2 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -88,7 +89,7 @@ trait Eclair {
 
   def receive(description: String, amount_opt: Option[MilliSatoshi], expire_opt: Option[Long], fallbackAddress_opt: Option[String], paymentPreimage_opt: Option[ByteVector32])(implicit timeout: Timeout): Future[PaymentRequest]
 
-  def newAddress(): Future[String]
+  def newAddress()(implicit timeout: Timeout): Future[String]
```
pm47, line 93: This is currently useless because the method is implemented by a call to `Eclair.getReceiveAddress` which doesn't take support timeouts. Other methods are implemented with `asks` which will understand the implicit parameter.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -90,7 +90,13 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
     future.map(_.forall(b => b))
   }
 
-  override def getBalance: Future[Satoshi] = rpcClient.invoke("getbalance") collect { case JDecimal(balance) => Satoshi(balance.bigDecimal.scaleByPowerOfTen(8).longValue) }
+  override def getBalance: Future[OnChainBalance] = rpcClient.invoke("getbalances").map(json => {
+    def toSatoshi(v: BigDecimal): Satoshi = Satoshi(v.bigDecimal.scaleByPowerOfTen(8).longValue)
```
pm47: Those decimal/integer conversions are tricky. I think the formula is correct, but would a few unit tests be over the top?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -127,6 +128,8 @@ trait Eclair {
 
   def onChainBalance()(implicit timeout: Timeout): Future[OnChainBalance]
 
+  def listTransactions(count: Int, skip: Int)(implicit timeout: Timeout): Future[Iterable[WalletTransaction]]
```
pm47, line 133: For consistency with `onChainBalance`:

```suggestion
  def listOnChainTransactions(count: Int, skip: Int)(implicit timeout: Timeout): Future[Iterable[WalletTransaction]]
```

or just `onChainTransactions`

eclair-core/eclair-cli
```diff
@@ -26,19 +26,58 @@ where OPTIONS can be:
...
   -s                    Some commands can print a trimmed JSON
 
-and COMMAND is one of:
-  getinfo, connect, disconnect, open, close, forceclose, updaterelayfee,
-  peers, channels, channel, allnodes, allchannels, allupdates
-  findroute, findroutetonode, parseinvoice, payinvoice, sendtonode,
-  sendtoroute, getsentinfo, createinvoice, getinvoice, listinvoices,
-  listpendinginvoices, getreceivedinfo, audit, networkfees,
-  channelstats, usablebalances
+and COMMAND is one of the available commands:
```
pm47, line 29: Nice, but this is so easily out-of-date though.

eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -283,6 +283,9 @@ trait Service extends ExtraDirectives with Logging {
                         path("usablebalances") {
                           complete(eclairApi.usableBalances())
                         } ~
+                        path("onchainbalances") {
```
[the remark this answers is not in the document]
t-bast: Yes, that bothered me too :smile: 
I think that since for off-chain we have `usablebalances` we need to align this one and the on-chain one. I just don't know if plural or singular is better, what do you prefer?
It's probably a good occasion to rename `usablebalance(s)` to `offchainbalance(s)`? Or something else?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: Gotcha, I'll remove it then.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
t-bast: I agree, it's probably worth adding a unit test, it never hurts.

eclair-core/eclair-cli
t-bast: I agree, but the previous one was also out-of-date, so while updating it I figured it was easier to keep it up-to-date with that formatting :)
There's probably a lot of improvement we could make to the CLI to make it keep up-to-date more easily, but I don't think that's very important atm.

eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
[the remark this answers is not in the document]
pm47: The plural for LN balances is for `send`/`receive`, so it doesn't bother me to have `usablebalances` and `onchainbalance`.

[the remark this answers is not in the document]
t-bast: Great in that case I'll go with the simplest change of renaming the JSON API `onchainbalance`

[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1461/commits/80b23340a99b886b6df1df79551b4c2b33ab093a

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1461/commits/80b23340a99b886b6df1df79551b4c2b33ab093a

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1461/commits/80b23340a99b886b6df1df79551b4c2b33ab093a

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
t-bast: I improved the test in https://github.com/ACINQ/eclair/pull/1461/commits/dcb67246eaf0e53f7cdde1346e74f92e4d13204f, if the conversion was incorrect this test wouldn't pass.

PR #1463, opened 2020-06-19
Static key refactoring

pm47:
See commits description. They are independent and can be reviewed separately.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -604,10 +604,7 @@ object Commitments {
     val channelKeyPath = keyManager.channelKeyPath(localParams, channelVersion)
     val localDelayedPaymentPubkey = Generators.derivePubKey(keyManager.delayedPaymentPoint(channelKeyPath).publicKey, localPerCommitmentPoint)
     val localHtlcPubkey = Generators.derivePubKey(keyManager.htlcPoint(channelKeyPath).publicKey, localPerCommitmentPoint)
-    val remotePaymentPubkey = channelVersion match {
-      case v if v.hasStaticRemotekey => remoteParams.paymentBasepoint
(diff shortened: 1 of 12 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -604,16 +604,10 @@ object Commitments {
...
-      case _ => Generators.derivePubKey(remoteParams.paymentBasepoint, localPerCommitmentPoint)
-    }
+    val remotePaymentPubkey = if (channelVersion.hasStaticRemotekey) remoteParams.paymentBasepoint else Generators.derivePubKey(remoteParams.paymentBasepoint, localPerCommitmentPoint)
     val remoteHtlcPubkey = Generators.derivePubKey(remoteParams.htlcBasepoint, localPerCommitmentPoint)
     val localRevocationPubkey = Generators.revocationPubKey(remoteParams.revocationBasepoint, localPerCommitmentPoint)
-    val localPaymentBasepoint = channelVersion match {
-      case v if v.isSet(USE_STATIC_REMOTEKEY_BIT) => localParams.localPaymentBasepoint.get
-      case _ => keyManager.paymentPoint(channelKeyPath).publicKey
-    }
+    val localPaymentBasepoint = localParams.staticPaymentBasepoint.getOrElse(keyManager.paymentPoint(channelKeyPath).publicKey)
```
sstone, line 610: it looks better, but now it does not check that  `localParams.staticPaymentBasepoint` is defined iff `channelVersion` activates `option_static_remote_key`

pm47: It kind of does, here: https://github.com/ACINQ/eclair/blob/59ea4cd606bfbcaae80e2d59b8a6e9f6ac68bc1b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala#L67

I think that's consistent with how remote works: we just use the value they give us and assume it follows the static feature bit.

PR #1466, opened 2020-06-22
Increase fulfill safety window

t-bast:
Improve comments and explanations in various places.

This is a simpler alternative to #1452


eclair-core/src/main/scala/fr/acinq/eclair/CltvExpiry.scala
```diff
@@ -55,6 +55,9 @@ case class CltvExpiryDelta(private val underlying: Int) extends Ordered[CltvExpi
   def +(other: Int): CltvExpiryDelta = CltvExpiryDelta(underlying + other)
   def +(other: CltvExpiryDelta): CltvExpiryDelta = CltvExpiryDelta(underlying + other.underlying)
   def -(other: CltvExpiryDelta): CltvExpiryDelta = CltvExpiryDelta(underlying - other.underlying)
+  def *(m: Int): CltvExpiryDelta = CltvExpiryDelta(underlying * m)
+  def *(m: Double): CltvExpiryDelta = CltvExpiryDelta((underlying * m).toInt)
+  def /(d: Double): CltvExpiryDelta = CltvExpiryDelta((underlying / d).toInt)
```
pm47, line 60: I'd rather remove the `Double` operations. There are not used and create precision issues.

```suggestion
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1466/commits/bbf54f4472354ea5a77210f4fdc3fb09228f9b21

PR #1469, opened 2020-06-24
Add context logging to route request

pm47:


eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -58,43 +59,48 @@ object RouteCalculation {
...
-        case Success(routes) =>
-          Metrics.RouteResults.withTags(tags).record(routes.length)
-          routes.foreach(route => Metrics.RouteLength.withTags(tags).record(route.length))
-          ctx.sender ! RouteResponse(routes)
-        case Failure(t) =>
-          val failure = if (isNeighborBalanceTooLow(d.graph, r)) BalanceTooLow else t
-          Metrics.FindRouteErrors.withTags(tags.withTag(Tags.Error, failure.getClass.getSimpleName)).increment()
-          ctx.sender ! Status.Failure(failure)
+  def handleRouteRequest(d: Data, routerConf: RouterConf, currentBlockHeight: Long, r: RouteRequest)(implicit ctx: ActorContext, log: DiagnosticLoggingAdapter): Data = {
+    Logs.withMdc(log)(Logs.mdc(
```
t-bast, line 63: It may also make sense to add `r.target` in `remoteNodeId_opt`. Or is it confusing because we usually use it to indicate the node we're communicating with?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -396,7 +397,8 @@ object Router {
                           ignore: Ignore = Ignore.empty,
                           routeParams: Option[RouteParams] = None,
                           allowMultiPart: Boolean = false,
-                          pendingPayments: Seq[Route] = Nil)
+                          pendingPayments: Seq[Route] = Nil,
+                          paymentConfig: Option[SendPaymentConfig] = None)
```
t-bast, line 401: It's a bit overkill to send the whole `SendPaymentConfig` whereas what we need is just the ids and the `payment_hash`...
We may in the future wrongly start using other fields of this class, I'd prefer sending only what we need but maybe I'm nit-picking a bit too much :)

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
pm47: I think we should keep this field for the direct peer, otherwise it's confusing.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: How about with a comment like in b1c32b7a6?

t-bast: I'd prefer something like `paymentContext: Option[PaymentContext]` where `case class PaymentContext(id: UUID, parentId: UUID, paymentHash: ByteVector32)` that's easily built from the `SendPaymentConfig`

PR #1473, opened 2020-06-26
Rework feerate mismatch

### LINKED ISSUE #1438: Unusual formula for measuring the fee rate mismatch
The formula eclair uses for calculating the difference between two fee rates is:

```scala
  /**
   * @param referenceFeePerKw reference fee rate per kiloweight
   * @param currentFeePerKw   current fee rate per kiloweight
   * @return the "normalized" difference between i.e local and remote fee rate: |reference - current| / avg(current, reference)
   */
  def feeRateMismatch(referenceFeePerKw: Long, currentFeePerKw: Long): Double =
    Math.abs((2.0 * (referenceFeePerKw - currentFeePerKw)) / (currentFeePerKw + referenceFeePerKw))
```

This seems a bit odd to me. If the reference fee rate is 1btc and the current fee rate is 0.0001btc then the average of the two will be (approximately) 0.5btc - which is half the reference fee rate but 5000 times the current fee rate. It would make more sense to measure this on a log scale so that an infinite difference (in terms of orders-of-magnitude) becomes an infinite `feeRateMismatch`, rather than a `feeRateMismatch` of 2.

Is it worth changing the formula to `|log(reference) - log(current)|`? (And also changing the name of the config option to avoid silently changing its meaning?)

t-bast:
This PR improves the way we handle feerate mismatch, giving more configuration flexibility to users and achieving better security while minimizing channel closures.

In https://github.com/ACINQ/eclair/commit/8e7f4916ce2be3bf2273f381b4aed3d1aae6dd98, we split the `max-feerate-mismatch` into two distinct configuration values. A low feerate is more dangerous than a high feerate (especially since it's the remote node that pays that fee).

In https://github.com/ACINQ/eclair/commit/e222318a108a1f9ae03c3afb367265830301c3d0 we relax the conditions for closing a channel when feerates don't match: if a channel has no HTLC, we can accept the remote feerate. It's only when an HTLC will be added to this channel that we need to evaluate whether we agree on feerate or not (in which case we close).

Fixes #1438

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -101,7 +101,7 @@ eclair {
 
     feerate-tolerance {
       ratio-low = 0.5 // will allow remote fee rates as low as half our local feerate
-      ratio-high = 8.0 // will allow remote fee rates as high as 8 times our local feerate
+      ratio-high = 10.0 // will allow remote fee rates as high as 10 times our local feerate
     }
     close-on-offline-feerate-mismatch = true // do not change this unless you know what you are doing
 
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -99,9 +99,10 @@ eclair {
       claim-main = 12    // target for the claim main transaction (tx that spends main channel output back to wallet)
     }
 
-    // maximum local vs remote feerate mismatch; 1.0 means 100%
-    // actual check is abs((local feerate - remote fee rate) / (local fee rate + remote fee rate)/2) > fee rate mismatch
-    max-feerate-mismatch = 1.56 // will allow remote fee rates up to 8x bigger or smaller than our local fee rate
+    feerate-mismatch {
+      ratio-low = 0.5 // will allow remote fee rates as low as half our local feerate
+      ratio-high = 8.0 // will allow remote fee rates as high as 8 times our local feerate
+    }
     close-on-offline-feerate-mismatch = true // do not change this unless you know what you are doing
 
     // funder will send an UpdateFee message if the difference between current commitment fee and actual current network fee is greater
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -99,7 +99,7 @@ eclair {
       claim-main = 12    // target for the claim main transaction (tx that spends main channel output back to wallet)
     }
 
-    feerate-mismatch {
+    feerate-tolerance {
       ratio-low = 0.5 // will allow remote fee rates as low as half our local feerate
       ratio-high = 8.0 // will allow remote fee rates as high as 8 times our local feerate
     }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -26,7 +26,7 @@ import com.typesafe.config.{Config, ConfigFactory, ConfigValueType}
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{Block, ByteVector32, Satoshi}
 import fr.acinq.eclair.NodeParams.WatcherType
-import fr.acinq.eclair.blockchain.fee.{FeeEstimator, FeeTargets, FeerateMismatch, OnChainFeeConf}
+import fr.acinq.eclair.blockchain.fee.{FeeEstimator, FeeTargets, FeerateTolerance, OnChainFeeConf}
 import fr.acinq.eclair.channel.Channel
 import fr.acinq.eclair.crypto.KeyManager
 import fr.acinq.eclair.db._
@@ -143,7 +143,7 @@ object NodeParams {
       "global-features" -> "features",
       "local-features" -> "features",
       // v0.4.1
-      "on-chain-fees.max-feerate-mismatch" -> "on-chain-fees.feerate-mismatch.ratio-low / on-chain-fees.feerate-mismatch.ratio-high"
+      "on-chain-fees.max-feerate-mismatch" -> "on-chain-fees.feerate-tolerance.ratio-low / on-chain-fees.feerate-tolerance.ratio-high"
     )
     deprecatedKeyPaths.foreach {
       case (old, new_) => require(!config.hasPath(old), s"configuration key '$old' has been replaced by '$new_'")
@@ -246,7 +246,7 @@ object NodeParams {
       onChainFeeConf = OnChainFeeConf(
         feeTargets = feeTargets,
         feeEstimator = feeEstimator,
-        maxFeerateMismatch = FeerateMismatch(config.getDouble("on-chain-fees.feerate-mismatch.ratio-low"), config.getDouble("on-chain-fees.feerate-mismatch.ratio-high")),
+        maxFeerateMismatch = FeerateTolerance(config.getDouble("on-chain-fees.feerate-tolerance.ratio-low"), config.getDouble("on-chain-fees.feerate-tolerance.ratio-high")),
         closeOnOfflineMismatch = config.getBoolean("on-chain-fees.close-on-offline-feerate-mismatch"),
         updateFeeMinDiffRatio = config.getDouble("on-chain-fees.update-fee-min-diff-ratio")
       ),
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -26,7 +26,7 @@ import com.typesafe.config.{Config, ConfigFactory, ConfigValueType}
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{Block, ByteVector32, Satoshi}
 import fr.acinq.eclair.NodeParams.WatcherType
-import fr.acinq.eclair.blockchain.fee.{FeeEstimator, FeeTargets, OnChainFeeConf}
+import fr.acinq.eclair.blockchain.fee.{FeeEstimator, FeeTargets, FeerateMismatch, OnChainFeeConf}
 import fr.acinq.eclair.channel.Channel
 import fr.acinq.eclair.crypto.KeyManager
 import fr.acinq.eclair.db._
@@ -141,7 +141,9 @@ object NodeParams {
       "update-fee_min-diff-ratio" -> "on-chain-fees.update-fee-min-diff-ratio",
       // v0.3.3
       "global-features" -> "features",
-      "local-features" -> "features"
+      "local-features" -> "features",
+      // v0.4.1
+      "on-chain-fees.max-feerate-mismatch" -> "on-chain-fees.feerate-mismatch.ratio-low / on-chain-fees.feerate-mismatch.ratio-high"
     )
     deprecatedKeyPaths.foreach {
       case (old, new_) => require(!config.hasPath(old), s"configuration key '$old' has been replaced by '$new_'")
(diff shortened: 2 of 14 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -99,9 +99,10 @@ eclair {
       claim-main = 12    // target for the claim main transaction (tx that spends main channel output back to wallet)
     }
 
-    // maximum local vs remote feerate mismatch; 1.0 means 100%
-    // actual check is abs((local feerate - remote fee rate) / (local fee rate + remote fee rate)/2) > fee rate mismatch
-    max-feerate-mismatch = 1.56 // will allow remote fee rates up to 8x bigger or smaller than our local fee rate
+    feerate-mismatch {
+      ratio-low = 0.5 // will allow remote fee rates as low as half our local feerate
+      ratio-high = 8.0 // will allow remote fee rates as high as 8 times our local feerate
+    }
```
pm47, line 105:
```suggestion
    feerate-tolerance {
      ratio-low = 0.5 // will allow remote fee rates as low as half our local feerate
      ratio-high = 8.0 // will allow remote fee rates as high as 8 times our local feerate
    }
```

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -244,7 +246,7 @@ object NodeParams {
       onChainFeeConf = OnChainFeeConf(
         feeTargets = feeTargets,
         feeEstimator = feeEstimator,
-        maxFeerateMismatch = config.getDouble("on-chain-fees.max-feerate-mismatch"),
+        maxFeerateMismatch = FeerateMismatch(config.getDouble("on-chain-fees.feerate-mismatch.ratio-low"), config.getDouble("on-chain-fees.feerate-mismatch.ratio-high")),
```
pm47, line 249:
```suggestion
        maxFeerateMismatch = FeerateTolerance(config.getDouble("on-chain-fees.feerate-mismatch.ratio-low"), config.getDouble("on-chain-fees.feerate-mismatch.ratio-high")),
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -815,7 +813,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       else if (Commitments.localHasUnsignedOutgoingHtlcs(d.commitments))
       // TODO: simplistic behavior, we could also sign-then-close
       handleCommandError(CannotCloseWithUnsignedOutgoingHtlcs(d.channelId), c)
-      else if (!Closing.isValidFinalScriptPubkey(localScriptPubKey))
+        else if (!Closing.isValidFinalScriptPubkey(localScriptPubKey))
```
pm47, line 816:
```suggestion
      else if (!Closing.isValidFinalScriptPubkey(localScriptPubKey))
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1807,8 +1810,11 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
     val currentFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw
-    // if the fees are too high we risk to not be able to confirm our current commitment
-    if (networkFeeratePerKw > currentFeeratePerKw && Helpers.isFeeDiffTooHigh(currentFeeratePerKw, networkFeeratePerKw, nodeParams.onChainFeeConf.maxFeerateMismatch)) {
+    // if the network fees are too high we risk to not be able to confirm our current commitment
+    val shouldClose = networkFeeratePerKw > currentFeeratePerKw &&
+      Helpers.isFeeDiffTooHigh(networkFeeratePerKw, currentFeeratePerKw, nodeParams.onChainFeeConf.maxFeerateMismatch) &&
+      d.commitments.hasPendingOrProposedHtlcs // we close only if we have HTLCs potentially at risk
+    if (shouldClose) {
       if (nodeParams.onChainFeeConf.closeOnOfflineMismatch) {
         log.warning(s"closing OFFLINE ${d.channelId} due to fee mismatch: currentFeeratePerKw=$currentFeeratePerKw networkFeeratePerKw=$networkFeeratePerKw")
```
pm47, line 1819: (same for the comment a few lines below)

```suggestion
        log.warning(s"closing OFFLINE channel due to fee mismatch: currentFeeratePerKw=$currentFeeratePerKw networkFeeratePerKw=$networkFeeratePerKw")
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1787,13 +1785,18 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-        self ! CMD_UPDATE_FEE(networkFeeratePerKw, commit = true)
-        stay
-      case false if Helpers.isFeeDiffTooHigh(currentFeeratePerKw, networkFeeratePerKw, nodeParams.onChainFeeConf.maxFeerateMismatch) =>
-        handleLocalError(FeerateTooDifferent(d.channelId, localFeeratePerKw = networkFeeratePerKw, remoteFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw), d, Some(c))
-      case _ => stay
+    val shouldUpdateFee = d.commitments.localParams.isFunder &&
+      Helpers.shouldUpdateFee(currentFeeratePerKw, networkFeeratePerKw, nodeParams.onChainFeeConf.updateFeeMinDiffRatio)
+    val shouldClose = !d.commitments.localParams.isFunder &&
+      Helpers.isFeeDiffTooHigh(networkFeeratePerKw, currentFeeratePerKw, nodeParams.onChainFeeConf.maxFeerateMismatch) &&
+      d.commitments.hasPendingOrProposedHtlcs // we close only if we have HTLCs potentially at risk
```
pm47, line 1792: Proposed htlcs are not at risk (they haven't been signed)

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -189,6 +194,13 @@ object Commitments {
       return Failure(HtlcValueTooSmall(commitments.channelId, minimum = htlcMinimum, actual = cmd.amount))
     }
 
+    // we allowed mismatches between our feerates and our remote's as long as commitments didn't contain any HTLC at risk
+    // we need to verify that we're not disagreeing on feerates anymore before offering new HTLCs
+    val localFeeratePerKw = feeConf.feeEstimator.getFeeratePerKw(target = feeConf.feeTargets.commitmentBlockTarget)
```
pm47, line 199: This error should be translated to a `TemporaryChannelFailure` at relay.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -129,7 +128,7 @@ object Helpers {
     }
 
     val localFeeratePerKw = nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(target = nodeParams.onChainFeeConf.feeTargets.commitmentBlockTarget)
-    if (isFeeDiffTooHigh(open.feeratePerKw, localFeeratePerKw, nodeParams.onChainFeeConf.maxFeerateMismatch)) throw FeerateTooDifferent(open.temporaryChannelId, localFeeratePerKw, open.feeratePerKw)
+    if (isFeeDiffTooHigh(localFeeratePerKw, open.feeratePerKw, nodeParams.onChainFeeConf.maxFeerateMismatch)) throw FeerateTooDifferent(open.temporaryChannelId, localFeeratePerKw, open.feeratePerKw)
```
pm47, line 131: Why have you inverted `localFeeratePerKw` and `open.feeratePerKw`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -189,25 +188,20 @@ object Helpers {
...
+   * To avoid spamming our peers with fee updates every time there's a small variation, we only update the fee when the
+   * difference exceeds a given ratio (updateFeeMinDiffRatio).
    */
-  def feeRateMismatch(referenceFeePerKw: Long, currentFeePerKw: Long): Double =
-    Math.abs((2.0 * (referenceFeePerKw - currentFeePerKw)) / (currentFeePerKw + referenceFeePerKw))
-
-  def shouldUpdateFee(commitmentFeeratePerKw: Long, networkFeeratePerKw: Long, updateFeeMinDiffRatio: Double): Boolean =
-    feeRateMismatch(networkFeeratePerKw, commitmentFeeratePerKw) > updateFeeMinDiffRatio
+  def shouldUpdateFee(currentFeeratePerKw: Long, nextFeeratePerKw: Long, updateFeeMinDiffRatio: Double): Boolean =
+    Math.abs((currentFeeratePerKw - nextFeeratePerKw).toDouble / currentFeeratePerKw) > updateFeeMinDiffRatio
```
pm47, line 195: We should check the division by zero, it isn't much effort.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Because since the calculation was previously symmetric, it was a mess and `isFeeDiffTooHigh` was called with the wrong order in some places. It didn't matter because the result was the same, but now the result is different, so I had to choose an order.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: Good point, I'll write some tests and make sure this is the case.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: With the current implementation, they might be. Because once they're proposed, they'll eventually get signed and added to the commitment without going through the feerate mismatch verification (which we only do in `sendAdd` and `receiveAdd`).

I think for simplicity sake that it's better to close right now, allowing us to check the feerate mismatch only in these two functions while being safe. Otherwise we'd need to check it in a lot more places which is more error-prone.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1473/commits/9037da2978266fb0de40bf13c272d0024dfedcea

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1473/commits/9037da2978266fb0de40bf13c272d0024dfedcea

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1473/commits/9037da2978266fb0de40bf13c272d0024dfedcea

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1473/commits/9037da2978266fb0de40bf13c272d0024dfedcea

eclair-core/src/main/resources/reference.conf
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1473/commits/9037da2978266fb0de40bf13c272d0024dfedcea

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1473/commits/e03d05793ba4f402b1f1c2ab4c61ae92408bacd5

PR #1474, opened 2020-06-29
Add lower bound on to_self_delay

t-bast:
We shouldn't let to_self_delay be arbitrary low.
We apply the lower bound both to our delays and our remote's, for fairness.


t-bast:
I don't think this is an issue. I chose to apply it symmetrically, to ensure we can neither accept nor send a `to_self_delay` below 144 (we don't want to encourage remote nodes to be at risk by allowing us to have a value too low). It also makes the code much more understandable than if I applied it only for the other side (and we could misunderstand whether it's applied to local or remote).

t-bast:
Riiight, I indeed got confused there!
So we always ask the remote to have a 720 blocks `to_self_delay`, that sounds good, no need for my PR!

PR #1475, opened 2020-06-30
Update backup handler and add DB metrics

t-bast:
Akka 2.4 introduced a non-blocking bounded mailbox.
It makes sense to upgrade the backup handler to that mailbox type.

We also add DB metrics and wrap interesting calls.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/FileBackupHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/FileBackupHandler.scala
@@ -21,6 +21,7 @@ import java.nio.file.{Files, StandardCopyOption}
 
 import akka.actor.{Actor, ActorLogging, Props}
 import akka.dispatch.{BoundedMessageQueueSemantics, RequiresMessageQueue}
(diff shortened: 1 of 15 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/FileBackupHandler.scala
```diff
@@ -65,6 +66,8 @@ class FileBackupHandler private(databases: FileBackup, backupFile: File, backupS
 
       // publish a notification that we have updated our backup
       context.system.eventStream.publish(BackupCompleted)
+      Metrics.FileBackupCompleted.withoutTags().increment()
+      Metrics.FileBackupDuration.withoutTags().record(end - start)
```
pm47, line 70: Why don't you use `KamonExt.time`?

t-bast: Done in https://github.com/ACINQ/eclair/pull/1475/commits/84aee65057bf3788b58b1ddae8dd4d0078f3a4ab

PR #1477, opened 2020-07-01
Replace Travis by Github Action

pm47:


.github/workflows/main.yml
```diff
@@ -0,0 +1,43 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
```
t-bast, line 3:
```suggestion
name: Build & Test
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,43 @@
+# This is a basic workflow to help you get started with Actions
```
t-bast, line 1: Remove? To hide the fact that we're noobs and got away with good old copy-pasting? :)

.github/workflows/main.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Cache Maven dependencies
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2
+        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+        restore-keys: ${{ runner.os }}-m2
+
+    # Runs a set of commands using the runners shell
+    - name: Build with Maven
+      run: |
```
t-bast, line 41: Why the `|`?

.github/workflows/main.yml
pm47: Done with 65b691f.

.github/workflows/main.yml
pm47: This to allow multi-line commands, but I went differently with 5470640.

pm47:
We got a test failure in [this build](https://github.com/ACINQ/eclair/runs/827108044), maybe a regression due to #1474?

> FuzzTest(true,2,0,Satoshi(372),220019009 msat,984244262 msat) -> could not setup initial htlcs: fr.acinq.eclair.channel.FeerateTooDifferent: local/remote feerates are too different: remoteFeeratePerKw=0 localFeeratePerKw=10000 (CommitmentsSpec.scala:453)

t-bast:
> We got a test failure in this build, maybe a regression due to #1474?

Right, this is a fuzz test, I see where that's coming from, I'll fix it.

pm47:
> just need to change the travis badge on our README

Right! good call

pm47:
>> just need to change the travis badge on our README

> Right! good call

Done with 3bd3d01. I'd rather have used the file path method as documented [here](https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#example-using-a-workflow-file-path) but couldn't get it to work.

PR #1480, opened 2020-07-02
Handle invoices with 0 amount

### LINKED ISSUE #1478: parseinvoice API throws error for empty invoice generated by CLightning
parseinvoice works fine with empty invoice generated by LND but throws error `The form field 'invoice' was malformed:\\nrequirement failed: amount is not valid` for Clightining's empty invoice.

Successful response from LND's empty Invoice:
{
	"prefix": "lntb",
	"timestamp": 1593632106,
	"nodeId": "02fad37ef58f76c1a5f802d198a07da5a8e55620b9ffbcbdc316fc8c9e9eac02cb",
	"serialized": "lntb1p00eet2pp5hd8s2u9nh8a0vu23rmmvqwp9eewfu54ks22rhjf2fqjxtx03zuqqdzg23jhxarfdenjq3trd3skju3q2pshjmt9de68xgrhd96xsgz9d4c8g7fqf9h8vmmfvdjjqveqcqzpgxqy9gcqsp565fye2jp5tdny2le9azgkylcuj8ghc282g5we6ympj5kula2zn0q9qy9qsqyzmupljnqyzlexk6wfh5aqs587c288005ksszuxxz6dyvyl952hqwuvvm0wquy3zjjzcrv95w24jaqh7cpewnmgjnggxwd0azd4qqespj8saqr",
	"description": "Testing Eclair Payments with Empty Invoice 3 ",
	"paymentHash": "bb4f0570b3b9faf671511ef6c03825ce5c9e52b682943bc92a48246599f11700",
	"expiry": 172800,
	"minFinalCltvExpiry": 40
}

Error while parsing Clightning's empty Invoice:
{
	"name": "StatusCodeError",
	"statusCode": 400,
	"message": "400 - {\"error\":\"The form field 'invoice' was malformed:\\nrequirement failed: amount is not valid\"}",
	"error": {
		"error": "The form field 'invoice' was malformed:\nrequirement failed: amount is not valid"
	},
	"options": {
		"url": "http://localhost:9090/parseinvoice",
		"rejectUnauthorized": false,
		"json": true,
		"form": {
			"invoice": "lntb01p00eezgpp5zprrgpmr8wugjhrpnnwqfc0u3suuvju6z2f7wgk9xm2sv95k90eqdpc23jhxarfdenjqetdwp68jgrfdemx76trv5sxvun0d5sy2cmvv95hygp3xqyjw5qcqp2sp56q5kq5gvftylckzlwppzj994qxvtj8kcwysdjucjltfpycxekcusrzjqg9ncduwavcnd98jqtd98mecucuxf4rrc4fh24u9ke766ftrlmn3jx48jcqqpqsqqqqqqqqpqqqqqzsqqc9qy9qsqvkz059mh3q5ghtcz2343vd4440mpr6dlmfeuv8rlgn3vhssxzk2jwap33lawq8p9s46r9s5pcszvvn7g0a025rj4jq0p5wcdnsljw6sqg6rruu"
		},
		"headers": {},
		"method": "POST",
		"qs": {},
		"simple": true,
		"resolveWithFullResponse": false,
		"transform2xxOnly": false
	},
	"response": {
		"statusCode": 400,
		"body": {
			"error": "The form field 'invoice' was malformed:\nrequirement failed: amount is not valid"
		},
		"headers": {
			"access-control-allow-headers": "Content-Type, Authorization",
			"access-control-allow-methods": "POST",
			"cache-control": "public, no-store, max-age=0",
			"server": "akka-http/10.0.11",
			"date": "Wed, 01 Jul 2020 19:46:25 GMT",
			"connection": "close",
			"content-type": "application/json",
			"content-length": "100"
		},
		"request": {
			"uri": {
				"protocol": "http:",
				"slashes": true,
				"auth": null,
				"host": "localhost:9090",
				"port": "9090",
				"hostname": "localhost",
				"hash": null,
				"search": null,
				"query": null,
				"pathname": "/parseinvoice",
				"path": "/parseinvoice",
				"href": "http://localhost:9090/parseinvoice"
			},
			"method": "POST",
			"headers": {
				"content-type": "application/x-www-form-urlencoded",
				"accept": "application/json",
				"content-length": 403
			}
		}
	}
}

t-bast:
We handled empty amounts (field not specified at all) but not the case
where the amount was specified and equal to 0.

Fixes #1478

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
@@ -451,13 +451,16 @@ object PaymentRequest {
     }
 
     def decode(input: String): Option[MilliSatoshi] =
-      input match {
+      (input match {
         case "" => None
         case a if a.last == 'p' => Some(MilliSatoshi(a.dropRight(1).toLong / 10L)) // 1 pico-bitcoin == 10 milli-satoshis
         case a if a.last == 'n' => Some(MilliSatoshi(a.dropRight(1).toLong * 100L))
         case a if a.last == 'u' => Some(MilliSatoshi(a.dropRight(1).toLong * 100000L))
         case a if a.last == 'm' => Some(MilliSatoshi(a.dropRight(1).toLong * 100000000L))
         case a => Some(MilliSatoshi(a.toLong * 100000000000L))
+      }).flatMap {
+        case MilliSatoshi(0) => None
+        case amount => Some(amount)
       }
 
     def encode(amount: Option[MilliSatoshi]): String = {
--- a/eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentRequestSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentRequestSpec.scala
@@ -53,6 +53,15 @@ class PaymentRequestSpec extends AnyFunSuite {
     assert('m' === Amount.unit((1 btc).toMilliSatoshi))
   }
 
+  test("decode empty amount") {
+    assert(Amount.decode("") === None)
+    assert(Amount.decode("0") === None)
+    assert(Amount.decode("0p") === None)
+    assert(Amount.decode("0n") === None)
+    assert(Amount.decode("0u") === None)
+    assert(Amount.decode("0m") === None)
+  }
+
   test("check that we can still decode non-minimal amount encoding") {
     assert(Amount.decode("1000u") === Some(100000000 msat))
     assert(Amount.decode("1000000n") === Some(100000000 msat))
@@ -75,7 +84,6 @@ class PaymentRequestSpec extends AnyFunSuite {
 
   test("verify that padding is zero") {
     val codec = PaymentRequest.Codecs.alignedBytesCodec(bits)
-
     assert(codec.decode(bin"1010101000").require == DecodeResult(bin"10101010", BitVector.empty))
     assert(codec.decode(bin"1010101001").isFailure) // non-zero padding
   }


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -450,15 +450,17 @@ object PaymentRequest {
...
-      input match {
-        case "" => None
-        case a if a.last == 'p' => Some(MilliSatoshi(a.dropRight(1).toLong / 10L)) // 1 pico-bitcoin == 10 milli-satoshis
-        case a if a.last == 'n' => Some(MilliSatoshi(a.dropRight(1).toLong * 100L))
-        case a if a.last == 'u' => Some(MilliSatoshi(a.dropRight(1).toLong * 100000L))
-        case a if a.last == 'm' => Some(MilliSatoshi(a.dropRight(1).toLong * 100000000L))
-        case a => Some(MilliSatoshi(a.toLong * 100000000000L))
+    def decode(input: String): Option[MilliSatoshi] = {
+      val amount = input match {
+        case "" => 0.msat
```
pm47, line 455: I'm not sure I like the fact that you first convert to `0.msat` before converting to `None`.

Surely there is a more elegant way of doing this?

t-bast: I can instead change all the `Some(MilliSatoshi(a.toLong * multiplier))` to call an internal method that returns `None` if 0 msat and `Some` otherwise, would that be better?

t-bast: What about https://github.com/ACINQ/eclair/pull/1480/commits/19d3ae4790b2354a2a00faa6a07a07e8eda8a391 ?

pm47: How about:

```scala
    def decode(input: String): Option[MilliSatoshi] = {
      val pattern = "([0-9]*)([pnum]?)".r
      input match {
        case pattern(a, _) if a.isEmpty || a.toLong == 0 => None
        case pattern(a, "p") => Some(MilliSatoshi(a.toLong / 10L)) // 1 pico-bitcoin == 10 milli-satoshis
        case pattern(a, "n") => Some(MilliSatoshi(a.toLong * 100L))
        case pattern(a, "u") => Some(MilliSatoshi(a.toLong * 100000L))
        case pattern(a, "m") => Some(MilliSatoshi(a.toLong * 100000000L))
        case pattern(a, "") => Some(MilliSatoshi(a.toLong * 100000000000L))
      }
    }
```

pm47: Regex is evil, I know...

t-bast: I'd prefer avoiding a regex :)
What's wrong with the `flatMap`?

pm47: Just aesthetics, but it's fine.

PR #1483, opened 2020-07-07
Update MinFinalCltvExpiryDelta default value and activate wumbo

t-bast:
## Activate wumbo by default

This is safe as `max-funding-satoshis` is still set to 16777215 sats, which is the non-wumbo limit.
If users want to increase the maximum channel size, they will now only need to update this configuration value.

## Update default `minFinalCltvExpiryDelta`

See https://github.com/lightningnetwork/lightning-rfc/pull/785
A value of `9` is a bit too optimistic and could be exploited.

## Set `minFinalCltvExpiryDelta` in invoices

It's important to set `minFinalCltvExpiryDelta` in the invoices we generate to something bigger than our `fulfill-safety-before-timeout-blocks`, otherwise we'll close channels whenever a block is produced while we were fulfilling a payment for which we're the final recipient.

NB: @dkrm0 `keysend` payments will be a problem, because we don't control the `minFinalCltvExpiryDelta` senders will set and currently have no way of letting them know what value we want. When spec-ing `keysend`, we should add a tlv to our `node_announcement` to advertize our `minFinalCltvExpiryDelta`. Meanwhile we may want to reject `keysend` payments that are below our `minFinalCltvExpiryDelta`.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -63,7 +63,7 @@ object Channel {
   val MAX_NEGOTIATION_ITERATIONS = 20
 
   // this is defined in BOLT 11
-  val MIN_CLTV_EXPIRY_DELTA = CltvExpiryDelta(9)
+  val MIN_CLTV_EXPIRY_DELTA = CltvExpiryDelta(18)
```
pm47, line 66: the comment above is inaccurate now.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -240,11 +244,14 @@ object MultiPartHandler {
...
+  private def validatePayment(nodeParams: NodeParams, payment: IncomingPacket.FinalPacket, record: IncomingPayment)(implicit log: LoggingAdapter): Option[CMD_FAIL_HTLC] = {
     // We send the same error regardless of the failure to avoid probing attacks.
-    val cmdFail = CMD_FAIL_HTLC(payment.add.id, Right(IncorrectOrUnknownPaymentDetails(payment.payload.totalAmount, currentBlockHeight)), commit = true)
+    val cmdFail = CMD_FAIL_HTLC(payment.add.id, Right(IncorrectOrUnknownPaymentDetails(payment.payload.totalAmount, nodeParams.currentBlockHeight)), commit = true)
     val paymentAmountOk = record.paymentRequest.amount.forall(a => validatePaymentAmount(payment, a))
-    val paymentCltvOk = validatePaymentCltv(payment, record.paymentRequest.minFinalCltvExpiryDelta.getOrElse(Channel.MIN_CLTV_EXPIRY_DELTA).toCltvExpiry(currentBlockHeight))
+    // The default `min_final_cltv_expiry` we accept is our `fulfill_safety_before_timeout` to which we add one block for
+    // safety (to handle races between our peer acknowledging the HTLC fulfill and a new block being produced).
+    val defaultMinFinalCltvExpiryDelta = nodeParams.fulfillSafetyBeforeTimeoutBlocks + 1
```
pm47, line 253: Shouldn't this logic take place in `validatePaymentCltv`? Why don't we rely on the value in `record.paymentRequest.minFinalCltvExpiryDelta`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -192,7 +192,7 @@ object PaymentInitiator {
    *                           the payment will automatically be retried in case of TrampolineFeeInsufficient errors.
    *                           For example, [(10 msat, 144), (15 msat, 288)] will first send a payment with a fee of 10
    *                           msat and cltv of 144, and retry with 15 msat and 288 in case an error occurs.
-   * @param finalExpiryDelta   expiry delta for the final recipient.
+   * @param finalExpiryDelta   expiry delta for the final recipient when the [[paymentRequest]] doesn't specify it.
```
pm47, line 195: The name is confusing, it doesn't show that this is a fallback. Maybe `defaultfinalExpiryDelta`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: Because unfortunately `record.paymentRequest.minFinalCltvExpiryDelta` is an `Option`...

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
t-bast: Yeah I agree, why not even `fallbackFinalExpiryDelta`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Depends, I'm hoping we merge https://github.com/lightningnetwork/lightning-rfc/pull/785 soon(ish) so this value will be in Bolt 11 :D

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: I can move that inside `validatePaymentCltv` though, that makes sense.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -56,6 +56,10 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
       Try {
         val paymentPreimage = paymentPreimage_opt.getOrElse(randomBytes32)
         val paymentHash = Crypto.sha256(paymentPreimage)
+        // We must set a `min_final_cltv_expiry` bigger than our `fulfill_safety_before_timeout`. Otherwise when we receive
+        // a payment, if a new block is produced while we're waiting for our peer to acknowledge our fulfill we'll end up
+        // closing the channel because it violates `fulfill_safety_before_timeout`.
+        val minFinalExpiryDelta = nodeParams.fulfillSafetyBeforeTimeoutBlocks + 3
```
pm47, line 62: How did you come up with the value `3`? Is it that pre d5ec6a56e842b8daabdc2dd91d43b238db956811, we had `fulfill-safety-before-timeout-blocks = 6` with `minFinalCltvExpiryDelta= 9` and `9-6=3`?

t-bast: I chose it completely arbitrarily :grinning: 
We need to add at least 1 to handle a new block being produced, and I figured maybe it's worth giving a bit more leeway than 1 block...
Do you think I should make this configurable? I could make `min-final-expiry-delta` a config value and verify that it's `> fulfill-safety-before-timeout`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1483/commits/d28fbba9ef0c69ec911610b2f381297fd62e32a3

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1483/commits/14c023b67b629338570d11815cc9cf9dd16c3216

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: Or maybe `+ 1` is good enough...

pm47: > I could make min-final-expiry-delta a config value and verify that it's > fulfill-safety-before-timeout.

That seems sensible. We will need to introduce some hierarchy in the `nodeParams` soon.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1483/commits/01678eeadf94f0a9f9d5a9aee2dc42bd38d730aa, it feels cleaner.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -185,53 +185,53 @@ object PaymentInitiator {
...
+   * @param assistedRoutes           (optional) routing hints (usually from a Bolt 11 invoice).
+   * @param routeParams              (optional) parameters to fine-tune the routing algorithm.
+   * @param userCustomTlvs           (optional) user-defined custom tlvs that will be added to the onion sent to the target node.
    */
   case class SendPaymentRequest(recipientAmount: MilliSatoshi,
                                 paymentHash: ByteVector32,
                                 recipientNodeId: PublicKey,
                                 maxAttempts: Int,
-                                finalExpiryDelta: CltvExpiryDelta = Channel.MIN_CLTV_EXPIRY_DELTA,
+                                fallbackFinalExpiryDelta: CltvExpiryDelta = Channel.MIN_CLTV_EXPIRY_DELTA,
```
pm47, line 227: On second thought, I'm not too sure about this. The `Channel.MIN_CLTV_EXPIRY_DELTA` is spec-defined, that's why it is a constant (and not a configurable value). Why don't we change the definition of `PaymentRequest.minFinalCltvExpiryDelta` to :

```scala
lazy val minFinalCltvExpiryDelta: CltvExpiryDelta = tags.collectFirst {
    case cltvExpiry: PaymentRequest.MinFinalCltvExpiry => cltvExpiry.toCltvExpiryDelta
  } getOrElse(Channel.MIN_CLTV_EXPIRY_DELTA)
```

We could also keep the `PaymentRequest.minFinalCltvExpiryDelta` optional, and define a new :

```scala
lazy val minFinalCltvExpiryDelta: Option[CltvExpiryDelta] = ...
lazy val minFinalCltvExpiryDeltaWithFallback: CltvExpiryDelta = minFinalCltvExpiryDelta.getOrElse(Channel.MIN_CLTV_EXPIRY_DELTA)
```

But that seems overkill, since we can always iterate on the `tags` if we want to know whether the `minFinalCltvExpiryDelta` was explicitely defined in the payment request or not.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -57,6 +57,7 @@ case class NodeParams(keyManager: KeyManager,
                       maxAcceptedHtlcs: Int,
                       expiryDeltaBlocks: CltvExpiryDelta,
                       fulfillSafetyBeforeTimeoutBlocks: CltvExpiryDelta,
+                      minFinalExpiryDeltaBlocks: CltvExpiryDelta,
```
pm47, line 60: The field `minFinalExpiryDelta` is typed, no need to specify that the unit is in blocks.

```suggestion
                      minFinalExpiryDelta: CltvExpiryDelta,
```

eclair-core/src/main/resources/reference.conf
```diff
@@ -76,6 +77,7 @@ eclair {
   // NB: this number effectively reduces the expiry-delta-blocks, so you may want to take that into account and increase
   // expiry-delta-blocks.
   fulfill-safety-before-timeout-blocks = 24
+  min-final-expiry-delta-blocks = 30 // Bolt 11 invoice's min_final_cltv_expiry; must be strictly greater than fulfill-safety-before-timeout-blocks
```
pm47, line 80: The difference between `expiry-delta-blocks`, `fulfill-safety-before-timeout-blocks`, `min-final-expiry-delta-blocks` is pretty subtle, maybe we should prefix them with `relay` or `receive` when applicable? Can be revamped in a later PR though.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
t-bast: > The Channel.MIN_CLTV_EXPIRY_DELTA is spec-defined, that's why it is a constant (and not a configurable value).

This is changing in the spec as we speak :)
We all agreed yesterday that hard-coding this in the spec was a bad idea from the start, and we want to remove it entirely.
The last commit of the spec PR now says that writers MUST specify a value in their invoice, and the default value of 18 is only there for backwards-compatibility for readers that receive invoices from legacy wallets/nodes that didn't specify this value.

The goal in the long(er) term is that this will allow us to change this value from an `Option[CltvExpiryDelta]` to a `CltvExpiryDelta` once enough nodes have upgraded.

eclair-core/src/main/resources/reference.conf
t-bast: Agreed, I think it's worth a separate PR to put them in a dedicated subsection of the config (which requires some tests in `StartupSpec` to guide users on how to change their config).

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
pm47: That's fine, let's redefine `Channel.MIN_CLTV_EXPIRY_DELTA` to `18` and take my first proposal, it is consistent with the decision to make writer `MUST` define the value?

t-bast: And we do need to keep this `fallbackFinalExpiryDelta` in these request classes to allow senders to tweak this.
Ideally when you receive an `expiryTooSoon` error you could guess that you need to provide a higher value here.

t-bast: Or at least the node operator can manually act on this error when the payment fails, whereas if we hide that inside `PaymentRequest.scala` we lose this ability.

pm47: > And we do need to keep this fallbackFinalExpiryDelta in these request classes to allow senders to tweak this.

Why would senders need to tweak this?

> Ideally when you receive an expiryTooSoon error you could guess that you need to provide a higher value here.
> Or at least the node operator can manually act on this error when the payment fails, whereas if we hide that inside PaymentRequest.scala we lose this ability.

If the `minFinalExpiryTooSoon` is enforced in the payment request, then the only reason this would fail is a disagreement in the current block height, in that case the payment should be retried as-is? Anyway for the use-case you are describing, this isn't a fallback but an override.

t-bast: Senders may need to tweak this when paying invoices that don't specify this value.

I think older nodes will want to increase the values they accept (given the recent noise about flood and loot) which they can likely do in other implementations by just changing a configuration value, without specifying it in invoices (because the code isn't yet released in implementation to add this to invoices).

I agree that we should remove this fallback value at some point, but I think we should wait a bit before doing this because it removes flexibility...

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1483/commits/135f40e29b8c1ef0cc7fe2ebfd4f13fdb45cc472

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
t-bast: There are some unknowns that may play into this as well, like hodl invoices, so I'm a bit wary when making the API less flexible than before

pm47: Looking deeper into this, I don't think we are handling `FinalIncorrectCltvExpiry` at all in the payment lifecycle?

We have four possible values for the `minFinalCltvExpiry`:
1) returned in a `FinalIncorrectCltvExpiry`
2) manual override (what we are discussing here)
3) in the payment request
4) default suggested value in the spec (9 blocks then, 18 blocks now)

Do we agree on the order of precedence? If we handle the `FinalIncorrectCltvExpiry` error correctly, I'm not sure we need the manual override.

t-bast: Thanks :), once lnd and c-lightning release support for setting it by default in invoices, I think we'll be able to safely remove those and simplify the code

PR #1484, opened 2020-07-09
Anchor output transaction format

t-bast:
This PR contains a first step towards the anchor outputs commitment format.
It implements all the transaction format changes and passes official test vectors.

It's currently only raw transaction/script code and will at first be unreachable (until support is added in higher layers to negotiate the feature at channel opening time).
We're implementing this feature from the bottom up to be able to integrate small chunks to master safely.

Some clean-up/refactoring was necessary, so I advise to review commit-by-commit.

~NB: there is one area where we currently diverge from lnd, in the case where only one of the anchor outputs is added (pruned `to_remote`). We are still clarifying what the behavior should be on the spec PR, I'll update here once there's consensus between implementations on the expected behavior).~

EDIT: we confirmed during the spec meeting that both anchors should always be deduced for simplicity's sake. This is done in https://github.com/ACINQ/eclair/pull/1484/commits/fefd07642e9d0ba292f0526d3f767917730857da, which makes this PR fully spec compliant.


t-bast:
> In receiveCommit() and sendCommit() shouldn't we change the SIGHASH flag used when signing HTLC txs if anchor_output is active ?

Yes we will :)

But this is done in a later PR (currently being worked on) that focuses on the changes to `Channel.scala`, `Helpers.scala` and `Commitments.scala` to avoid a huge feature PR.

The current PR adds all the script/transaction format capabilities, but doesn't exploit them yet.

PR #1485, opened 2020-07-10
Implement the KeySend feature, spontaneous payments

dkrm0:
## What is KeySend?
For an explanation of the feature see the PR from [c-lightning](https://github.com/ElementsProject/lightning/pull/3611).

## Testing
Support for receiving and sending is added. This was tested against each Lightning implementations (`Sender -> Receiver`):
- `Eclair -> Eclair`
- `Eclair -> lnd`
- `Eclair -> c-lightning`
- `lnd -> Eclair`

Sending support [is not currently merged](https://github.com/ElementsProject/lightning/pull/3792) in c-lightning, so testing in this way `c-lightning -> Eclair` cannot be performed yet.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -48,6 +48,7 @@ eclair {
     payment_secret = optional
     basic_mpp = optional
     option_support_large_channel = optional
+	keysend = optional
   }
   override-features = [ // optional per-node features
     #  {
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -48,7 +48,6 @@ eclair {
     payment_secret = optional
     basic_mpp = optional
     option_support_large_channel = optional
-	keysend = optional
   }
   override-features = [ // optional per-node features
     #  {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -22,7 +22,7 @@ import akka.actor.ActorRef
 import akka.pattern._
 import akka.util.Timeout
 import fr.acinq.bitcoin.Crypto.PublicKey
-import fr.acinq.bitcoin.{ByteVector32, Satoshi}
+import fr.acinq.bitcoin.{ByteVector32, Crypto, Satoshi}
 import fr.acinq.eclair.TimestampQueryFilters._
 import fr.acinq.eclair.blockchain.OnChainBalance
 import fr.acinq.eclair.blockchain.bitcoind.BitcoinCoreWallet
@@ -38,7 +38,7 @@ import fr.acinq.eclair.payment.relay.Relayer.{GetOutgoingChannels, OutgoingChann
 import fr.acinq.eclair.payment.send.PaymentInitiator.{SendPaymentRequest, SendPaymentToRouteRequest, SendPaymentToRouteResponse}
 import fr.acinq.eclair.router.Router._
 import fr.acinq.eclair.router.{NetworkStats, RouteCalculation}
-import fr.acinq.eclair.wire.{ChannelAnnouncement, ChannelUpdate, NodeAddress, NodeAnnouncement}
+import fr.acinq.eclair.wire.{ChannelAnnouncement, ChannelUpdate, NodeAddress, NodeAnnouncement, GenericTlv}
 import scodec.bits.ByteVector
 
 import scala.concurrent.duration._
@@ -131,6 +131,8 @@ trait Eclair {
   def onChainBalance(): Future[OnChainBalance]
 
   def onChainTransactions(count: Int, skip: Int): Future[Iterable[WalletTransaction]]
+
+  def sendWithPreimage(externalId_opt: Option[String], recipientNodeId: PublicKey, amount: MilliSatoshi, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
 
 }
 
@@ -378,4 +380,18 @@ class EclairImpl(appKit: Kit) extends Eclair {
 
   override def usableBalances()(implicit timeout: Timeout): Future[Iterable[UsableBalance]] =
     (appKit.relayer ? GetOutgoingChannels()).mapTo[OutgoingChannels].map(_.channels.map(_.toUsableBalance))
+
+  override def sendWithPreimage(externalId_opt: Option[String], recipientNodeId: PublicKey, amount: MilliSatoshi, maxAttempts_opt: Option[Int], feeThreshold_opt: Option[Satoshi], maxFeePct_opt: Option[Double])(implicit timeout: Timeout): Future[UUID] = {
+    val maxAttempts = maxAttempts_opt.getOrElse(appKit.nodeParams.maxPaymentAttempts)
+    val defaultRouteParams = RouteCalculation.getDefaultRouteParams(appKit.nodeParams.routerConf)
+    val routeParams = defaultRouteParams.copy(
+      maxFeePct = maxFeePct_opt.getOrElse(defaultRouteParams.maxFeePct),
+      maxFeeBase = feeThreshold_opt.map(_.toMilliSatoshi).getOrElse(defaultRouteParams.maxFeeBase)
+    )
+    val paymentPreimage = randomBytes32
+    val paymentHash: ByteVector32 = Crypto.sha256(paymentPreimage)
+    val keySendTlvRecords = Seq(GenericTlv(UInt64(5482373484L), paymentPreimage))
+    val sendPayment = SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts = maxAttempts, externalId = externalId_opt, routeParams = Some(routeParams), userCustomTlvs = keySendTlvRecords)
+    (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
+  }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -392,7 +392,9 @@ class EclairImpl(appKit: Kit) extends Eclair {
     val paymentPreimage = randomBytes32
     val paymentHash: ByteVector32 = Crypto.sha256(paymentPreimage)
     val keySendTlvRecords = Seq(GenericTlv(keySendTagValue, paymentPreimage))
-    val sendPayment = SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts = maxAttempts, externalId = externalId_opt, routeParams = Some(routeParams), userCustomTlvs = keySendTlvRecords)
+    // TODO For now MIN_CLTV_EXPIRY_DELTA is set to 9 (spec requirement). But the lnd implementation require 13 as a minimum (see https://github.com/lightningnetwork/lnd/blob/master/invoices/invoiceregistry.go#L675)
+    // The merge request #1483 will bump the MIN_CLTV_EXPIRY_DELTA to 18. Until then we use this "magic value" of 13 to be interoperable with all implementations
+    val sendPayment = SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts = maxAttempts, finalExpiryDelta = CltvExpiryDelta(13), externalId = externalId_opt, routeParams = Some(routeParams), userCustomTlvs = keySendTlvRecords)
     (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
   }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -98,7 +98,7 @@ trait Eclair {
 
   def send(externalId_opt: Option[String], recipientNodeId: PublicKey, amount: MilliSatoshi, paymentHash: ByteVector32, invoice_opt: Option[PaymentRequest] = None, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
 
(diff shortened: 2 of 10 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -47,6 +47,7 @@ eclair {
     var_onion_optin = optional
     payment_secret = optional
     basic_mpp = optional
+	keysend = optional
```
t-bast, line 50: We shouldn't activate it by default until it is in the spec.
It should be opt-in since it may break once spec-ed (change in the feature bit used).

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -132,6 +133,8 @@ trait Eclair {
 
   def onChainTransactions(count: Int, skip: Int): Future[Iterable[WalletTransaction]]
 
+  def sendWithPreimage(externalId_opt: Option[String], recipientNodeId: PublicKey, amount: MilliSatoshi, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
```
t-bast, line 136: nit: can you put it right below the `send` method? This trait is a bit of a mess, let's try to add new methods close to related ones (same for the implementation below).

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -192,6 +192,11 @@ object Features {
     val mandatory = 50
   }
 
+  case object KeySend extends Feature {
```
t-bast, line 195: You should also add this in the `featuresDependency` map.
KeySend depends on `VariableLengthOnion` (you couldn't have known since it's not spec-ed yet!).

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -228,13 +230,15 @@ object Onion {
     val expiry: CltvExpiry
     val paymentSecret: Option[ByteVector32]
     val totalAmount: MilliSatoshi
+    val keySend: Option[KeySend]
```
t-bast, line 233: I would directly put the preimage here to remove one unnecessary level of nesting:

```suggestion
    val paymentPreimage: Option[ByteVector32]
```

And in the `FinalTlvPayload` you would do:

```
override val paymentPreimage = records.get[KeySend].map(_.paymentPreimage) 
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
```diff
@@ -290,6 +295,7 @@ object Onion {
     FinalTlvPayload(TlvStream(AmountToForward(amount), OutgoingCltv(expiry), PaymentData(paymentSecret, totalAmount), TrampolineOnion(trampolinePacket)))
   }
 
+  val keySendTagValue = UInt64(5482373484L)
```
t-bast, line 298: Remove, just directly put this value inline in the `typecase` in the `onionTlvCodec` like we do for other records

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -93,10 +93,22 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
...
+          case None => p.payload.keySend match {
+            case Some(keySend) =>
+              val paymentPreimage = keySend.paymentPreimage
+              val amount = Some(p.payload.totalAmount)
+              val paymentHash = Crypto.sha256(paymentPreimage)
+              val desc = "Donation"
+              // Insert a fake invoice and then restart the incoming payment handler
+              val paymentRequest = PaymentRequest(nodeParams.chainHash, amount, paymentHash, nodeParams.privateKey, desc)
+              log.debug("generated fake payment request={} from amount={} (KeySend)", PaymentRequest.write(paymentRequest), amount)
+              db.addIncomingPayment(paymentRequest, paymentPreimage)
```
t-bast, line 105: `addIncomingPayment` has an optional `paymentType` argument.
I think it would be useful to add a new `PaymentType.KeySend`, that would make the filtering very easy for the wallet.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -93,10 +93,22 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
...
-            val cmdFail = CMD_FAIL_HTLC(p.add.id, Right(IncorrectOrUnknownPaymentDetails(p.payload.totalAmount, nodeParams.currentBlockHeight)), commit = true)
-            PendingRelayDb.safeSend(register, nodeParams.db.pendingRelay, p.add.channelId, cmdFail)
+          case None => p.payload.keySend match {
+            case Some(keySend) =>
+              val paymentPreimage = keySend.paymentPreimage
+              val amount = Some(p.payload.totalAmount)
+              val paymentHash = Crypto.sha256(paymentPreimage)
+              val desc = "Donation"
+              // Insert a fake invoice and then restart the incoming payment handler
+              val paymentRequest = PaymentRequest(nodeParams.chainHash, amount, paymentHash, nodeParams.privateKey, desc)
```
t-bast, line 103: I think you need to set the invoice features like we do in `ReceivePayment`.
Right now you're not setting support for multipart payments even if it's activated in `nodeParams.features`, which means you will reject a payment if it's sent in multiple HTLCs (which I think lnd may do if they're unable to send the whole amount in a single HTLC).

You'll want something like:

```
val features =  if (nodeParams.features.hasFeature(Features.BasicMultiPartPayment)) {
    Seq(Features.BasicMultiPartPayment.optional, Features.PaymentSecret.optional, Features.VariableLengthOnion.optional)
} else {
    Seq(Features.PaymentSecret.optional, Features.VariableLengthOnion.optional)
}
val paymentRequest = PaymentRequest(nodeParams.chainHash, amount, paymentHash, nodeParams.privateKey, desc, features = features)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -93,10 +93,22 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
                   pendingPayments = pendingPayments + (p.add.paymentHash -> (record.paymentPreimage, handler))
               }
           }
-          case None =>
-            Metrics.PaymentFailed.withTag(Tags.Direction, Tags.Directions.Received).withTag(Tags.Failure, "InvoiceNotFound").increment()
-            val cmdFail = CMD_FAIL_HTLC(p.add.id, Right(IncorrectOrUnknownPaymentDetails(p.payload.totalAmount, nodeParams.currentBlockHeight)), commit = true)
-            PendingRelayDb.safeSend(register, nodeParams.db.pendingRelay, p.add.channelId, cmdFail)
+          case None => p.payload.keySend match {
+            case Some(keySend) =>
```
t-bast, line 97: You also need to check that key-send is active. If the user doesn't activate keySend in his features, he wants to reject those payments:

```suggestion
            case Some(keySend) if nodeParams.features.hasFeature(Features.KeySend) =>
```

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -388,4 +388,18 @@ class EclairImplSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with I
...
+
+    val eclair = new EclairImpl(kit)
+    val nodeId = randomKey.publicKey
+
+    eclair.sendWithPreimage(None, nodeId, 12345 msat)
+    val send = paymentInitiator.expectMsgType[SendPaymentRequest]
+    assert(send.externalId === None)
+    assert(send.recipientNodeId === nodeId)
+    assert(send.recipientAmount === 12345.msat)
+    assert(send.paymentRequest === None)
```
t-bast, line 402: You should verify that the tlv records contain the keySend with a preimage that matches `send.paymentHash`

eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -224,9 +224,13 @@ trait Service extends ExtraDirectives with Logging {
                           }
                         } ~
                         path("sendtonode") {
-                          formFields(amountMsatFormParam, paymentHashFormParam, nodeIdFormParam, "maxAttempts".as[Int].?, "feeThresholdSat".as[Satoshi].?, "maxFeePct".as[Double].?, "externalId".?) {
-                            (amountMsat, paymentHash, nodeId, maxAttempts_opt, feeThresholdSat_opt, maxFeePct_opt, externalId_opt) =>
+                          formFields(amountMsatFormParam, paymentHashFormParam.?, nodeIdFormParam, "maxAttempts".as[Int].?, "feeThresholdSat".as[Satoshi].?, "maxFeePct".as[Double].?, "externalId".?, "keysend".?) {
```
t-bast, line 227: nit: let's group optional parameters together (and keysend should be typed):

```suggestion
                          formFields(amountMsatFormParam, nodeIdFormParam, paymentHashFormParam.?, "keysend".as[Boolean].?, "maxAttempts".as[Int].?, "feeThresholdSat".as[Satoshi].?, "maxFeePct".as[Double].?, "externalId".?) {
```

I'm wondering if it makes sense to allow users to provide the preimage themselves (in which case "keysend" would be a `ByteVector32` and we wouldn't generate it ourselves in `sendWithPreimage`). It's probably fine to start with just a `Boolean` and that can evolve if users ask for the ability to choose the preimage themselves.

dkrm0: Yes, I don't think it's needed right now. Nevertheless, I've modified `sendWithPreimage` to take an optional `paymentPreimage`. So if we want to give the user the opportunity to use their own preimage, we would just need to change `Boolean` to `ByteVector32`.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala
dkrm0: Done

eclair-core/src/main/resources/reference.conf
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
dkrm0: Done

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
dkrm0: Done

eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
dkrm0: Done

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
t-bast: I don't see where it's done...?
I meant you should check:
```
assert(send.userCustomTlvs.length === 1)
val keySendTlv = send.userCustomTlvs.head
assert(keySendTlv.tag === UInt64(5482373484L))
val preimage = ByteVector32(keySendTlv.value)
assert(Crypto.sha256(preimage) === send.paymentHash)
```

eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -224,9 +224,17 @@ trait Service extends ExtraDirectives with Logging {
...
+                              keySend match  {
+                                case Some(true) => reject(MalformedFormFieldRejection("paymentHash", "You cannot request a KeySend payment and specify a paymentHash"))
+                                case _ => complete(eclairApi.send(externalId_opt, nodeId, amountMsat, paymentHash, maxAttempts_opt = maxAttempts_opt, feeThresholdSat_opt = feeThresholdSat_opt, maxFeePct_opt = maxFeePct_opt))
+                              }
+                            case (amountMsat, nodeId, None, maxAttempts_opt, feeThresholdSat_opt, maxFeePct_opt, externalId_opt, keySend) =>
+                              keySend match {
+                                case Some(true) => complete(eclairApi.sendWithPreimage(externalId_opt, nodeId, amountMsat, maxAttempts_opt = maxAttempts_opt, feeThresholdSat_opt = feeThresholdSat_opt, maxFeePct_opt = maxFeePct_opt))
+                                case _ => reject(MalformedFormFieldRejection("paymentHash", "No payment type specified. Either give a paymentHash or use --keysend=true"))
```
t-bast, line 236: nit: "give" sounds a bit harsh, what about:

```suggestion
                                case _ => reject(MalformedFormFieldRejection("paymentHash", "No payment type specified. Either provide a paymentHash or use --keysend=true"))
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartHandlerSpec.scala
```diff
@@ -460,4 +467,41 @@ class MultiPartHandlerSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike
     })
   }
 
+  test("KeySend payment in a single HTLC") { f =>
```
t-bast, line 470: nit: it would be nice for consistency to harmonize the test names with the rest of the file, for example:

- PaymentHandler should handle single-part keysend payment
- PaymentHandler should reject keysend payment when feature is disabled

dkrm0: Done

PR #1489, opened 2020-07-17
Add delay before setting watches at startup

pm47:
Watching the blockchain is an asynchronous task, so it is "always late"
and it doesn't matter if we don't synchronously set the watch back when
the node is restarted.

It allows us to smoothen the load if needed at startup, because setting
tens of thousands of `watch-spent` all at once at startup is pretty
expensive.


pm47:
Maybe we could start by only delaying watches on the external channels in the router.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -222,8 +222,13 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
             case None =>
               // in all other cases we need to be ready for any type of closing
               // TODO: should we wait for an acknowledgment from the watcher?
-              blockchain ! WatchSpent(self, data.commitments.commitInput.outPoint.txid, data.commitments.commitInput.outPoint.index.toInt, data.commitments.commitInput.txOut.publicKeyScript, BITCOIN_FUNDING_SPENT)
-              blockchain ! WatchLost(self, data.commitments.commitInput.outPoint.txid, nodeParams.minDepthBlocks, BITCOIN_FUNDING_LOST)
+              // avoid herd effect at startup because watch-spent are intensive in terms of rpc calls to bitcoind
+              val watchSpentDelay = Random.nextLong(1 + nodeParams.watchSpentWindow.toSeconds).seconds
```
t-bast, line 226: Why do you add 1?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -222,8 +222,13 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
             case None =>
               // in all other cases we need to be ready for any type of closing
               // TODO: should we wait for an acknowledgment from the watcher?
-              blockchain ! WatchSpent(self, data.commitments.commitInput.outPoint.txid, data.commitments.commitInput.outPoint.index.toInt, data.commitments.commitInput.txOut.publicKeyScript, BITCOIN_FUNDING_SPENT)
-              blockchain ! WatchLost(self, data.commitments.commitInput.outPoint.txid, nodeParams.minDepthBlocks, BITCOIN_FUNDING_LOST)
+              // avoid herd effect at startup because watch-spent are intensive in terms of rpc calls to bitcoind
```
t-bast, line 225: Maybe we should start with only delaying channels that are going to `NORMAL`, not the `CLOSING` ones?

t-bast:
> Maybe we could start by only delaying watches on the external channels in the router.

Yeah that's probably the biggest gain already...then if it's not enough, we can do it on transitions offline -> normal.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Can't have a zero duration unfortunately. Pretty ugly...

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Good idea

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: But you *can* have a zero duration with your current code!
You need to do `1 + Random.nextLong(nodeParams.watchSpentWindow.toSeconds).seconds` if you want to ensure a non-zero duration

pm47: My bad, zero duration are ok but `Random.nextLong(0)` is *verboten*.

pm47: With 5dcdf5e, only the router watches are delayed. Since it's less sensitive than channel watches, I've chosen a non-zero default and replaced `+ 1` guard by a warning in the documentation.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: I've even restricted to the router watches. We'll see how it performs, given that the expensive path is to recursively look for spending transactions block-by-block, which we only do for channels.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Actually, better to have a conf check.... Done in 5d3c1506a.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -36,7 +36,7 @@ import fr.acinq.eclair.tor.Socks5ProxyParams
 import fr.acinq.eclair.wire.{Color, EncodingType, NodeAddress}
 import scodec.bits.ByteVector
 
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.{FiniteDuration, _}
```
t-bast, line 39: nit:
```suggestion
import scala.concurrent.duration._
```

PR #1490, opened 2020-07-20
Use github action for scoverage

pm47:


pm47:
> Why do you remove the .codecov.yml file?
> It was useful to remove the gui and node parts from code coverage.

Because it didn't work when I was running the tests locally. Instead I use a `directory` filter when uploading the reports in the `workflow.yml`.

pm47:
> Instead I use a directory filter when uploading the reports in the workflow.yml.

Except it's not working 🙄

PR #1494, opened 2020-07-23
Refactor integration tests

t-bast:
The IntegrationSpec's force-close tests were very hard to work with; I found out (with great pain) that they relied on many unwritten assumptions about event ordering which made them very brittle and hard to extend (in preparation for anchor outputs).

So I decided to re-work these tests to make them more robust and easier to extend, with the following changes:

- we don't need one node per force-close scenario, we can use different channels to the same node which makes the spec simpler.
- force-close tests now have better isolation: they create the channel at the beginning of the test, and the test ends with that channel closed.
- common parts have been refactored to reduce code duplication.
- force-close tests use `push_msat` to ensure both peers have an output in the commit tx, which wasn't previously tested

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/ForwardHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/ForwardHandler.scala
@@ -24,5 +24,7 @@ import akka.event.DiagnosticLoggingAdapter
  * Simple handler that forwards all messages to an actor
  */
 class ForwardHandler(actor: ActorRef) extends ReceiveHandler {
-  override def handle(implicit ctx: ActorContext, log: DiagnosticLoggingAdapter): Receive = { case msg => actor forward msg}
+  override def handle(implicit ctx: ActorContext, log: DiagnosticLoggingAdapter): Receive = {
+    case msg if !msg.isInstanceOf[ReceiveHandler] => actor forward msg
+  }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/ForwardHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/ForwardHandler.scala
@@ -25,6 +25,6 @@ import akka.event.DiagnosticLoggingAdapter
  */
 class ForwardHandler(actor: ActorRef) extends ReceiveHandler {
   override def handle(implicit ctx: ActorContext, log: DiagnosticLoggingAdapter): Receive = {
-    case msg if !msg.isInstanceOf[ReceiveHandler] => actor forward msg
+    case msg => actor forward msg
   }
 }
--- a/eclair-core/src/test/resources/logback-test.xml
+++ b/eclair-core/src/test/resources/logback-test.xml
@@ -50,9 +50,9 @@
     <logger name="fr.acinq.eclair.db.FileBackupHandler" level="OFF"/>
 
     <root level="INFO">
-        <!--appender-ref ref="FILE"/>
-        <appender-ref ref="CONSOLEWARN"/>
-        <appender-ref ref="CONSOLE"/-->
+        <!--<appender-ref ref="FILE"/>-->
+        <!--<appender-ref ref="CONSOLEWARN"/>-->
+        <!--<appender-ref ref="CONSOLE"/>-->
     </root>
 
 </configuration>
--- a/eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
@@ -20,6 +20,7 @@ import java.io.File
 import java.util.{Properties, UUID}
 
 import akka.actor.{ActorRef, ActorSystem}
+import akka.pattern.pipe
 import akka.testkit.{TestKit, TestProbe}
 import com.google.common.net.HostAndPort
 import com.typesafe.config.{Config, ConfigFactory}
@@ -31,13 +32,12 @@ import fr.acinq.eclair.blockchain.bitcoind.BitcoindService.BitcoinReq
 import fr.acinq.eclair.blockchain.bitcoind.rpc.ExtendedBitcoinClient
 import fr.acinq.eclair.blockchain.{Watch, WatchConfirmed}
 import fr.acinq.eclair.channel.Channel.{BroadcastChannelUpdate, PeriodicRefresh}
-import fr.acinq.eclair.channel.ChannelCommandResponse.ChannelOpened
 import fr.acinq.eclair.channel.Register.{Forward, ForwardShortId}
 import fr.acinq.eclair.channel._
 import fr.acinq.eclair.crypto.Sphinx.DecryptedFailurePacket
 import fr.acinq.eclair.crypto.TransportHandler
 import fr.acinq.eclair.db._
-import fr.acinq.eclair.io.Peer.{Disconnect, PeerRoutingMessage}
+import fr.acinq.eclair.io.Peer.PeerRoutingMessage
 import fr.acinq.eclair.io.{Peer, PeerConnection}
 import fr.acinq.eclair.payment.PaymentRequest.ExtraHop
 import fr.acinq.eclair.payment._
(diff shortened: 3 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/ForwardHandler.scala
```diff
@@ -24,5 +24,7 @@ import akka.event.DiagnosticLoggingAdapter
  * Simple handler that forwards all messages to an actor
  */
 class ForwardHandler(actor: ActorRef) extends ReceiveHandler {
-  override def handle(implicit ctx: ActorContext, log: DiagnosticLoggingAdapter): Receive = { case msg => actor forward msg}
+  override def handle(implicit ctx: ActorContext, log: DiagnosticLoggingAdapter): Receive = {
+    case msg if !msg.isInstanceOf[ReceiveHandler] => actor forward msg
```
pm47, line 28: Interesting, my first reaction was that this could surely be done with regular pattern matching on type, but it doesn't seem possible.

eclair-core/src/test/resources/logback-test.xml
```diff
@@ -50,9 +50,9 @@
     <logger name="fr.acinq.eclair.db.FileBackupHandler" level="OFF"/>
 
     <root level="INFO">
-        <!--appender-ref ref="FILE"/>
-        <appender-ref ref="CONSOLEWARN"/>
-        <appender-ref ref="CONSOLE"/-->
+        <!--<appender-ref ref="FILE"/>-->
+        <!--<appender-ref ref="CONSOLEWARN"/>-->
+        <!--<appender-ref ref="CONSOLE"/>-->
```
pm47, line 55: Is this to make intellij auto-uncomment work?

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -863,38 +840,79 @@ class IntegrationSpec extends TestKitBaseClass with BitcoindService with AnyFunS
...
+
+    val bitcoinClient = new ExtendedBitcoinClient(bitcoinrpcclient)
+    bitcoinClient.lookForSpendingTx(None, fundingOutpoint.txid, fundingOutpoint.index.toInt).pipeTo(sender.ref)
+    val closingTx = sender.expectMsgType[Transaction]
+    assert(closingTx.txOut.map(_.publicKeyScript).toSet === Set(finalPubKeyScriptC, finalPubKeyScriptF))
+
+    awaitAnnouncements(nodes.filterKeys(_ == "A").toMap, 5, 7, 16)
+  }
+
+  test("open channel C <-> F2, send payments and close (option_anchor_outputs, option_static_remotekey)") {
```
pm47, line 880: Do I understand correctly that you are introducing a new naming convention that F1 = standard commitment and F2 = anchor output? If so, I'm not sure that's a good idea because `C` doesn't have the same naming convention.

eclair-core/src/test/resources/logback-test.xml
t-bast: Exactly, I was annoyed at having to "manually" fix those every time I wanted to add logs to tests :)
Now it works with IntelliJ (and other editors such as VS code) shortcuts.

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
t-bast: It's not necessarily a naming convention, just two test nodes that support different commitment formats (just like we previously had one of the F nodes support trampoline and not the others).
I agree this file is a a bit complex to follow, especially since we want to batch some things to avoid making the tests unnecessarily slow :/

t-bast:
> Have you considered splitting the file in several parts?

Can you detail? Maybe that could be a further improvement, I'm not sure we should do it in the same PR though, it's already a bit hard to review (because these force-close tests are very subtle) but at least right now it follows quite the same flow as before.

pm47:
>>Have you considered splitting the file in several parts?

>Can you detail?

There just doesn't seem to be the need to put everything in the same test. Maybe we could break it three parts:
- routing table broadcast tests
- payment test
- channel failure scenarii.

A big plus would be that we would be able to parallelize and reduce the overall build time.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/ForwardHandler.scala
pm47: but... why do we need this in the first place? ❓ 😕 🤷‍♂️

t-bast:
> Maybe we could break it three parts:

Indeed that's a great idea and would considerably speed up the test suite.
Since I based all the anchor outputs PRs on this one, is it ok if I do that after we've merged the current anchor outputs PRs?

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/ForwardHandler.scala
t-bast: We're intercepting HTLCs in some integration tests in order to fetch the channel state right before the HTLC is relayed/fulfilled/failed to get some commitment-specific data...for the force-close tests (especially with revoked commitments) it's quite handy!

pm47: > We're intercepting HTLCs in some integration tests

I got that but why do we need to catch this particular case? Didn't we catch htlcs in the previous version of the tests, without needing this hack?

pm47:
> Since I based all the anchor outputs PRs on this one, is it ok if I do that after we've merged the current anchor outputs PRs?

Sure! I'm keen on merging this one asap so I can rebase #1520 on top of it

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/ForwardHandler.scala
t-bast: It worked before because we used one eclair node per close test, so we only set a `ForwardHandler` once on each node.

Now that I removed some of the test eclair nodes (F4, F5, etc) and we re-use the same nodes across multiple tests, we need to be able to set a new `ForwardHandler`, which was simply ignored without this line. Does that make sense? An alternative would be to share the `ForwardHandler` between tests, but it feels cleaner to give the possibility to change the `ReceiveHandler` at runtime.

pm47: Gotcha, how about this then:
```diff
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/PaymentHandler.scala b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/PaymentHandler.scala
index a8784d26a..03cf21756 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/PaymentHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/PaymentHandler.scala
@@ -36,13 +36,15 @@ class PaymentHandler(nodeParams: NodeParams, register: ActorRef) extends Actor w

   override def receive: Receive = normal(defaultHandler.handle(context, log))

-  def normal(handle: Receive): Receive = handle orElse {
+  def addReceiveHandler(handle: Receive): Receive = {
     case handler: ReceiveHandler =>
       log.info(s"registering handler of type=${handler.getClass.getSimpleName}")
       // NB: the last handler that was added will be the first called
       context become normal(handler.handle(context, log) orElse handle)
   }

+  def normal(handle: Receive): Receive = addReceiveHandler(handle) orElse handle
+
   override def mdc(currentMessage: Any): MDC = Logs.mdc(category_opt = Some(Logs.LogCategory.PAYMENT))
 }
```

Makes your change to the `ForwardHandler`  unnecessary.

t-bast: Good idea, I'll apply that!

pm47: A comment seems appropriate:

```scala
  /**
   * This is a bit subtle because we want handlers to be as generic as possible, but we also want to catch a particular
   * type of message (the [[ReceiveHandler]]s themselves) to update the list of handlers.
   *
   * That why we *prepend* new handlers, but after the first special handler [[addReceiveHandler]]:
   * {{{
   *    paymentHandler ! handler1
   *    paymentHandler ! handler2
   *    paymentHandler ! handler3
   *    // the current handler is now addReceiveHandler :: handler3 :: handler2 :: handler1
   * }}}
   *
   * @param handle
   * @return
   */
  def normal(handle: Receive): Receive = addReceiveHandler(handle) orElse handle
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1494/commits/e65f5a5380f409543f5c98dad1e6bed745bb51f9

pm47 (approved):

PR #1495, opened 2020-07-27
Verify feature graph in Init

t-bast:
If our remote peer doesn't correctly set transitive feature dependencies, we close the connection until they fix it.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -118,9 +118,15 @@ class PeerConnection(nodeParams: NodeParams, switchboard: ActorRef, router: Acto
 
         log.info(s"peer is using features=${remoteInit.features}, networks=${remoteInit.networks.mkString(",")}")
 
+        val featureGraphErr = Features.validateFeatureGraph(remoteInit.features)
         if (remoteInit.networks.nonEmpty && !remoteInit.networks.contains(d.nodeParams.chainHash)) {
           log.warning(s"incompatible networks (${remoteInit.networks}), disconnecting")
           d.pendingAuth.origin_opt.foreach(_ ! ConnectionResult.InitializationFailed("incompatible networks"))
+          d.transport ! PoisonPill
+          stay
+        } else if (featureGraphErr.nonEmpty) {
+          log.warning(featureGraphErr.get.message)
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -118,11 +118,17 @@ class PeerConnection(nodeParams: NodeParams, switchboard: ActorRef, router: Acto
...
         if (remoteInit.networks.nonEmpty && !remoteInit.networks.contains(d.nodeParams.chainHash)) {
           log.warning(s"incompatible networks (${remoteInit.networks}), disconnecting")
           d.pendingAuth.origin_opt.foreach(_ ! ConnectionResult.InitializationFailed("incompatible networks"))
           d.transport ! PoisonPill
           stay
+        } else if (featureGraphErr.nonEmpty) {
+          log.warning(featureGraphErr.get.message)
+          d.pendingAuth.origin_opt.foreach(_ ! ConnectionResult.InitializationFailed(featureGraphErr.get.message))
+          d.transport ! PoisonPill
+          stay
```
pm47, line 131: It troubles me a little bit that we call `Features.validateFeatureGraph` *before* comparing the `chain_hash`es, because results won't be accurate if the latter differ.

Also nit: I understand why you go imperative here, but in that case I'd rather only call `.get` once, and as close as possible to the `nonEmpty` (we got bitten by those before)

```suggestion
        val featureGraphErr_opt = Features.validateFeatureGraph(remoteInit.features)
        if (remoteInit.networks.nonEmpty && !remoteInit.networks.contains(d.nodeParams.chainHash)) {
          log.warning(s"incompatible networks (${remoteInit.networks}), disconnecting")
          d.pendingAuth.origin_opt.foreach(_ ! ConnectionResult.InitializationFailed("incompatible networks"))
          d.transport ! PoisonPill
          stay
        } else if (featureGraphErr_opt.nonEmpty) {
          val featureGraphErr = featureGraphErr_opt.get
          log.warning(featureGraphErr.message)
          d.pendingAuth.origin_opt.foreach(_ ! ConnectionResult.InitializationFailed(featureGraphErr.message))
          d.transport ! PoisonPill
          stay
```

t-bast: Agreed, done in https://github.com/ACINQ/eclair/pull/1495/commits/ad4f0375e21fb2707c6eb35c61c844a55c4a99e6
I'm always a bit unsatisfied when mixing imperative validation and `Option`s, there may be a more idiomatic way of doing it but at least it's easy to read.

PR #1498, opened 2020-07-29
Remove the `override-features` conf parameter

pm47:
The per-node override was only used in the `init` message, and not in
e.g. channel operations, which means it was completely broken.

The use-case for this is Eclair Mobile, because we only sync the routing
table from the ACINQ node (because there is no public channel
validation as it is too expensive on mobile devices).

I'm not sure what to do with this. Should we only have this feature on
the `android` branch?

Related to #1361.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -49,7 +49,6 @@ case class NodeParams(keyManager: KeyManager,
                       color: Color,
                       publicAddresses: List[NodeAddress],
                       features: Features,
-                      overrideFeatures: Map[PublicKey, Features],
                       syncWhitelist: Set[PublicKey],
                       dustLimit: Satoshi,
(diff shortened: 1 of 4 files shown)


pm47:
An alternative to this would be to fix the implementation (that is: to use the overriding features in `Peer.makeChannelParams`). But even then, some ambiguities may remain. E.g. in `MultiPartHandler` we're looking at `nodeParams.features`.

t-bast:
I think the feature in itself can be useful: when trying out experimental features (wumbo, anchor outputs) you may want to first try it with a friend's node or a node you trust.

If it doesn't become overly complex, I think it would be more useful to fix the feature than remove it.
We may have to make `nodeParams.features` a method that takes an optional `nodeId` as parameter, or something like this.

pm47:
> We may have to make nodeParams.features a method that takes an optional nodeId as parameter, or something like this.

That's a pretty good idea! Probably better than what I did in #1548 where I put it in the `switchboard` handler.

pm47:
Superseded by #1549.

PR #1499, opened 2020-07-30
Add API commands to sign & verify arbitrary messages

### LINKED ISSUE #1496: Add `signmessage` and `verifymessage` APIs
It can be useful to sign arbitrary messages with the key associated with our `node_id` (to prove ownership of a node).
It can be done implicitly by generating an invoice, but it shouldn't be much work to have an explicit way of doing it.

We should provide an interface similar to lnd, see:

- https://api.lightning.community/#signmessage
- https://api.lightning.community/#verifymessage

We will probably differ on the encoding to keep it simple and avoid adding new dependencies.
The message should probably be a base64 encoding of raw bytes, and the signature a raw 64-byte signature in hex format (or base64?), so the response would look something like:

```json
{
    "nodeId": "03af0ed6052cf28d670665549bc86f4b721c9fdb309d40c58f5811f63966e005d0",
    "message": "base64 encoding of a raw message",
    "signature": "64 bytes signature, hex-encoded"
}
```

dkrm0:
Fixes #1496

## End to End tests
`Sign` by -> `Verified` by
- [x] Eclair -> Eclair
- [x] Eclair -> Lnd
- [x] Eclair -> C-lightning
- [x] Lnd -> Eclair
- [x] C-lightning -> Eclair

## How to use the `signmessage` command
```
$> eclair-cli signmessage --msg=$(echo -n 'hello, world' | base64)
{
  "nodeId": "02ca7361934233f6d4defd4d792fb2ce2cd0b50c7605e6c5cd16006bcd5be2bf70",
  "message": "aGVsbG8gd29ybGQK",
  "signature": "1f730dce842c31b692dc041c2d0f00423d2a2a67b0c63c1a905d500f09652a5b1a036763a1603333fa589ae92d1f7963428ff170e976d0966a113f4b9f9d0efc7f"
}
```
## How to use the `verifymessage` command
```
$> eclair-clir verifymessage --msg=$(echo -n 'hello, world' | base64) --sig=1f730dce842c31b692dc041c2d0f00423d2a2a67b0c63c1a905d500f09652a5b1a036763a1603333fa589ae92d1f7963428ff170e976d0966a113f4b9f9d0efc7f
{
  "valid": true,
  "publicKey": "02ca7361934233f6d4defd4d792fb2ce2cd0b50c7605e6c5cd16006bcd5be2bf70"
}
```

## Signature format conversion
The signature we return are in hexadecimal format. Lnd & C-lightning use a zbase32 format. To convert from and to zbase32 you can use [zbase32](https://github.com/tv42/zbase32) and some basics Unix tools.

### Hexadecimal to zbase32
```
echo -n "${eclair_signature}" | xxd -revert -plain | zbase32-encode
```

### zbase32 to Hexadecimal
```
zbase32-decode "${zbase32_signature}" | xxd -cols 256 -plain
```

## Message format
We expect the `msg` argument, for both commands, to be a valid base64 string, to avoid issues with charset and encoding:
```
$> eclair-cli signmessage --msg=0123456789-invalid
The form field 'msg' was malformed:
Invalid base 64 character '-' at index 10
```

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -74,6 +74,9 @@ object ApiTypes {
 
 trait Eclair {
 
+  // String prefixed when signing arbitrary data to ensure we don't sign sensitive material
+  private val messagePrefix = ByteVector("Lightning Signed Message:".getBytes)
+
   def connect(target: Either[NodeURI, PublicKey])(implicit timeout: Timeout): Future[String]
 
   def disconnect(nodeId: PublicKey)(implicit timeout: Timeout): Future[String]
@@ -138,9 +141,9 @@ trait Eclair {
 
   def onChainTransactions(count: Int, skip: Int): Future[Iterable[WalletTransaction]]
 
-  def signMessage(base64Message: ByteVector): SignedMessage
+  def signMessage(base64Message: ByteVector, prefix: ByteVector = messagePrefix): SignedMessage
 
-  def verifyMessage(base64Message: ByteVector, signature: ByteVector64): VerifiedMessage
+  def verifyMessage(base64Message: ByteVector, signature: ByteVector64, prefix: ByteVector = messagePrefix): VerifiedMessage
 }
 
 class EclairImpl(appKit: Kit) extends Eclair {
@@ -401,11 +404,11 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
   }
 
-  override def signMessage(base64Message: ByteVector): SignedMessage = {
+  override def signMessage(base64Message: ByteVector, prefix: ByteVector): SignedMessage = {
     SignedMessage(appKit.nodeParams.privateKey.publicKey, ByteVector.empty, ByteVector64.Zeroes)
   }
 
-  override def verifyMessage(base64Message: ByteVector, signature: ByteVector64): VerifiedMessage = {
+  override def verifyMessage(base64Message: ByteVector, signature: ByteVector64, prefix: ByteVector): VerifiedMessage = {
     VerifiedMessage(false, None)
   }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -22,7 +22,7 @@ import akka.actor.ActorRef
 import akka.pattern._
 import akka.util.Timeout
 import fr.acinq.bitcoin.Crypto.PublicKey
-import fr.acinq.bitcoin.{ByteVector32, Crypto, Satoshi}
+import fr.acinq.bitcoin.{ByteVector32, ByteVector64, Crypto, Satoshi}
 import fr.acinq.eclair.TimestampQueryFilters._
 import fr.acinq.eclair.blockchain.OnChainBalance
 import fr.acinq.eclair.blockchain.bitcoind.BitcoinCoreWallet
@@ -38,7 +38,7 @@ import fr.acinq.eclair.payment.relay.Relayer.{GetOutgoingChannels, OutgoingChann
 import fr.acinq.eclair.payment.send.PaymentInitiator.{SendPaymentRequest, SendPaymentToRouteRequest, SendPaymentToRouteResponse}
 import fr.acinq.eclair.router.Router._
 import fr.acinq.eclair.router.{NetworkStats, RouteCalculation}
-import fr.acinq.eclair.wire.{ChannelAnnouncement, ChannelUpdate, NodeAddress, NodeAnnouncement, GenericTlv}
+import fr.acinq.eclair.wire._
 import scodec.bits.ByteVector
 
 import scala.concurrent.duration._
@@ -134,6 +134,9 @@ trait Eclair {
 
   def onChainTransactions(count: Int, skip: Int): Future[Iterable[WalletTransaction]]
 
+  def signMessage(base64Message: ByteVector): (PublicKey, ByteVector, ByteVector64)
+
+  def verifyMessage(base64Message: ByteVector, signature: ByteVector64): (Boolean, Option[PublicKey])
 }
 
 class EclairImpl(appKit: Kit) extends Eclair {
@@ -392,5 +395,13 @@ class EclairImpl(appKit: Kit) extends Eclair {
     val keySendTlvRecords = Seq(GenericTlv(UInt64(5482373484L), paymentPreimage))
     val sendPayment = SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts, externalId = externalId_opt, routeParams = Some(routeParams), userCustomTlvs = keySendTlvRecords)
     (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
+  }
+
+  override def signMessage(base64Message: ByteVector): (PublicKey, ByteVector, ByteVector64) = {
+    (appKit.nodeParams.privateKey.publicKey, ByteVector.empty, ByteVector64.Zeroes)
+  }
+
+  override def verifyMessage(base64Message: ByteVector, signature: ByteVector64): (Boolean, Option[PublicKey]) = {
+    (false, None)
   }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -69,7 +69,7 @@ object TimestampQueryFilters {
 }
 
 object SignedMessage {
-  def signedBytes(message: String): ByteVector32 = Crypto.hash256(ByteVector(("Lightning Signed Message:" ++ message).getBytes))
+  def signedBytes(message: ByteVector): ByteVector32 = Crypto.hash256(ByteVector("Lightning Signed Message:".getBytes) ++ message)
 }
 
 object ApiTypes {
@@ -142,9 +142,9 @@ trait Eclair {
 
   def onChainTransactions(count: Int, skip: Int): Future[Iterable[WalletTransaction]]
 
-  def signMessage(message: String): SignedMessage
+  def signMessage(message: ByteVector): SignedMessage
 
-  def verifyMessage(message: String, recoverableSignature: ByteVector): VerifiedMessage
+  def verifyMessage(message: ByteVector, recoverableSignature: ByteVector): VerifiedMessage
 }
 
 class EclairImpl(appKit: Kit) extends Eclair {
@@ -405,13 +405,13 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
   }
 
-  override def signMessage(message: String): SignedMessage = {
+  override def signMessage(message: ByteVector): SignedMessage = {
     val bytesToSign = SignedMessage.signedBytes(message)
     val (signature, recoveryId) = appKit.nodeParams.keyManager.signDigest(bytesToSign)
-    SignedMessage(appKit.nodeParams.nodeId, message, (recoveryId + 31).toByte +: signature)
+    SignedMessage(appKit.nodeParams.nodeId, message.toBase64, (recoveryId + 31).toByte +: signature)
   }
 
-  override def verifyMessage(message: String, recoverableSignature: ByteVector): VerifiedMessage = {
+  override def verifyMessage(message: ByteVector, recoverableSignature: ByteVector): VerifiedMessage = {
     val signedBytes = SignedMessage.signedBytes(message)
     val signature = ByteVector64(recoverableSignature.tail)
     val recoveryId = recoverableSignature.head.toInt - 31
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -68,14 +68,15 @@ object TimestampQueryFilters {
   }
 }
 
+object SignedMessage {
+  def signedBytes(message: String): ByteVector32 = Crypto.hash256(ByteVector(("Lightning Signed Message:" ++ message).getBytes))
+}
+
 object ApiTypes {
   type ChannelIdentifier = Either[ByteVector32, ShortChannelId]
 }
 
 trait Eclair {
-
-  // String prefixed when signing arbitrary data to ensure we don't sign sensitive material
-  private val messagePrefix = ByteVector("Lightning Signed Message:".getBytes)
 
   def connect(target: Either[NodeURI, PublicKey])(implicit timeout: Timeout): Future[String]
 
@@ -141,9 +142,9 @@ trait Eclair {
 
   def onChainTransactions(count: Int, skip: Int): Future[Iterable[WalletTransaction]]
 
-  def signMessage(message: String, prefix: ByteVector = messagePrefix): SignedMessage
+  def signMessage(message: String): SignedMessage
 
-  def verifyMessage(message: String, recoverableSignature: ByteVector, prefix: ByteVector = messagePrefix): VerifiedMessage
+  def verifyMessage(message: String, recoverableSignature: ByteVector): VerifiedMessage
 }
 
 class EclairImpl(appKit: Kit) extends Eclair {
@@ -404,18 +405,18 @@ class EclairImpl(appKit: Kit) extends Eclair {
     (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
   }
 
-  override def signMessage(message: String, prefix: ByteVector): SignedMessage = {
-    val hash256Message = Crypto.hash256(prefix ++ ByteVector(message.getBytes))
-    val (signature, recoveryId) = appKit.nodeParams.keyManager.signDigest(hash256Message)
+  override def signMessage(message: String): SignedMessage = {
+    val bytesToSign = SignedMessage.signedBytes(message)
+    val (signature, recoveryId) = appKit.nodeParams.keyManager.signDigest(bytesToSign)
     SignedMessage(appKit.nodeParams.nodeId, message, (recoveryId + 31).toByte +: signature)
   }
 
-  override def verifyMessage(message: String, recoverableSignature: ByteVector, prefix: ByteVector): VerifiedMessage = {
-    val hash256Message = Crypto.hash256(prefix ++ ByteVector(message.getBytes))
+  override def verifyMessage(message: String, recoverableSignature: ByteVector): VerifiedMessage = {
+    val signedBytes = SignedMessage.signedBytes(message)
     val signature = ByteVector64(recoverableSignature.tail)
     val recoveryId = recoverableSignature.head.toInt - 31
-    val pubKeyFromSignature = Crypto.recoverPublicKey(signature, hash256Message, recoveryId)
-    if (Crypto.verifySignature(hash256Message, signature, pubKeyFromSignature))
+    val pubKeyFromSignature = Crypto.recoverPublicKey(signature, signedBytes, recoveryId)
+    if (Crypto.verifySignature(signedBytes, signature, pubKeyFromSignature))
       VerifiedMessage(true, Some(pubKeyFromSignature))
     else
       VerifiedMessage(false, None)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -51,7 +51,7 @@ case class AuditResponse(sent: Seq[PaymentSent], received: Seq[PaymentReceived],
 
 case class TimestampQueryFilters(from: Long, to: Long)
 
-case class SignedMessage(nodeId: PublicKey, message: ByteVector, signature: ByteVector64)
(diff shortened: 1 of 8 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -134,6 +141,9 @@ trait Eclair {
 
   def onChainTransactions(count: Int, skip: Int): Future[Iterable[WalletTransaction]]
 
+  def signMessage(base64Message: ByteVector, prefix: ByteVector = messagePrefix): SignedMessage
```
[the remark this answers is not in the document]
t-bast: Also, it doesn't make sense to mention a string encoding (base64) when the field is a `ByteVector`.
Base64 is a way of representing bytes into a string, so either you accept a `base64Message: String` or just a `message: ByteVector` (same below for `verifyMessage`).

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -393,4 +403,21 @@ class EclairImpl(appKit: Kit) extends Eclair {
...
+  override def signMessage(base64Message: ByteVector, prefix: ByteVector): SignedMessage = {
+    val hash256Message = Crypto.hash256(prefix ++ base64Message)
+    val signature = appKit.nodeParams.keyManager.signDigest(hash256Message)
+    SignedMessage(appKit.nodeParams.keyManager.nodeId, base64Message.toBase64, signature)
+  }
+
+  override def verifyMessage(base64Message: ByteVector, signature: ByteVector64, prefix: ByteVector): VerifiedMessage = {
+    val hash256Message = Crypto.hash256(prefix ++ base64Message)
+    val (pubKey1, pubKey2) = Crypto.recoverPublicKey(signature, hash256Message)
+    if (appKit.nodeParams.db.network.getNode(pubKey1).isDefined)
```
t-bast, line 416: I don't think that's how we should do it, it feels hackish.
Thinking about this more, I think we should also include the `recoveryId` byte in our signature and use it (which allows us to use the `Crypto.recoverPublicKey` that takes a `recoveryId` and returns a single `PublicKey` (and internally uses `libsecp256k1` which is better than bouncycastle).
What do you think? I don't remember the exact details of how that `recoveryId` works but there's probably some documentation somewhere.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -64,6 +68,10 @@ object TimestampQueryFilters {
   }
 }
 
+object SignedMessage {
+  def signedBytes(message: String): ByteVector32 = Crypto.hash256(ByteVector(("Lightning Signed Message:" ++ message).getBytes))
```
t-bast, line 72: I think it's misleading to have `message` be a `String` here, what encoding should it be using?
If it's base64 we can't simply do `message.getBytes`.
This is why my suggestion was to use `def signedBytes(message: ByteVector)` and deal with the encoding (base64) outside of this method.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -393,4 +404,24 @@ class EclairImpl(appKit: Kit) extends Eclair {
     val sendPayment = SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts, externalId = externalId_opt, routeParams = Some(routeParams), userCustomTlvs = keySendTlvRecords)
     (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
   }
+
+  override def signMessage(message: String): SignedMessage = {
+    val bytesToSign = SignedMessage.signedBytes(message)
+    val (signature, recoveryId) = appKit.nodeParams.keyManager.signDigest(bytesToSign)
+    SignedMessage(appKit.nodeParams.nodeId, message, (recoveryId + 31).toByte +: signature)
```
t-bast, line 411: Shouldn't this be:

```suggestion
    SignedMessage(appKit.nodeParams.nodeId, message, signature :+ (recoveryId + 31).toByte)
```

t-bast: In invoices we put the recoveryId at the end of the signature, what does lnd and c-lightning do for this API?

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -393,4 +404,24 @@ class EclairImpl(appKit: Kit) extends Eclair {
     val sendPayment = SendPaymentRequest(amount, paymentHash, recipientNodeId, maxAttempts, externalId = externalId_opt, routeParams = Some(routeParams), userCustomTlvs = keySendTlvRecords)
     (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
   }
+
+  override def signMessage(message: String): SignedMessage = {
```
t-bast, line 408: You're missing a base64.decode step on the message, aren't you?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/KeyManager.scala
```diff
@@ -110,6 +110,16 @@ trait KeyManager {
...
    */
   def signChannelAnnouncement(fundingKeyPath: DeterministicWallet.KeyPath, chainHash: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, remoteFundingKey: PublicKey, features: Features): (ByteVector64, ByteVector64)
+
+  /**
+   * Sign a digest, primarily used to prove ownership of the current node
+   *
+   * @param digest     SHA256 digest
+   * @param privateKey private key to sign with, default the one from the current node
+   * @return a (signature, recoveryId) pair. signature is a signature of the digest parameter generated with the
+             private key given in parameter. recoveryId is the corresponding recoveryId of the signature
```
t-bast, line 120: I would take this opportunity to explain the recoveryId in more details:

```
When recovering a public key from an ECDSA signature for secp256k1, there are 4 possible matching curve points that can be found. The recoveryId identifies which of these points is the correct one to speed up recovery (otherwise we would need to check the signature's validity for each of these points to find the correct one).
```

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -393,4 +404,24 @@ class EclairImpl(appKit: Kit) extends Eclair {
...
+    SignedMessage(appKit.nodeParams.nodeId, message, (recoveryId + 31).toByte +: signature)
+  }
+
+  override def verifyMessage(message: String, recoverableSignature: ByteVector): VerifiedMessage = {
+    val signedBytes = SignedMessage.signedBytes(message)
+    val signature = ByteVector64(recoverableSignature.tail)
+    val recoveryId = recoverableSignature.head.toInt - 31
+    val pubKeyFromSignature = Crypto.recoverPublicKey(signature, signedBytes, recoveryId)
+    if (!pubKeyFromSignature.isValid)
+      VerifiedMessage(false, None)
```
t-bast, line 420: I don't think this check is necessary, public key recovery should always return a valid public key.
However it may return a public key for which the signature doesn't match, in case the recoveryId was incorrect (so the two checks below are useful).

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -393,4 +404,24 @@ class EclairImpl(appKit: Kit) extends Eclair {
...
+
+    if (Crypto.verifySignature(signedBytes, signature, pubKeyFromSignature))
+      VerifiedMessage(true, Some(pubKeyFromSignature))
+    else
+      VerifiedMessage(false, None)
```
t-bast, line 425: We try to always include the brackets, it doesn't use more lines and is less footgun-ish:

```suggestion
    if (Crypto.verifySignature(signedBytes, signature, pubKeyFromSignature)) {
      VerifiedMessage(true, Some(pubKeyFromSignature))
    } else {
      VerifiedMessage(false, None)
    }
```

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -429,4 +430,66 @@ class EclairImplSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with I
...
+    val msg = "super secret message"
+    // sha256(sha256("Lightning Signed Message:super secret message"))
+    val signature = 32.toByte +: ByteVector64(hex"97918faa65e1a49e1b0661299a4755a328d412dffe8852bc95420fb323639c840fc1c85a9ba7b909908e5654a5c6b19074b569ff3afbe284a7c208b2cdd0197e")
+
+    val verifiedMessage = eclair.verifyMessage(msg, signature)
+    assert(verifiedMessage.valid)
+    assert(verifiedMessage.signerNodeId === Some(expectedNodeId))
+  }
+
+  test("verify an invalid signature for the given message") { f =>
```
t-bast, line 483: This doesn't work, but I think you can still test the invalid case with the following flow:

- Sign a message with the `signMessage` API
- Pass a different message to `verifyMessage`
- Verify that you receive `valid == true` but that `signerNodeId` is different from what you expect.

Another test case that may be useful to add is the following:

- Sign a message with the `signMessage` API
- Change the recoveryId (if it's 0, use 1, or the other way around)
- Pass this modified signature to `verifyMessage`
- It should return `valid == false` (I think)

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -429,4 +430,66 @@ class EclairImplSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with I
     assert(expectedPaymentPreimage === ByteVector32(keySendTlv.value))
   }
 
+  test("sign an arbitrary message with the node's private key") { f =>
```
t-bast, line 433: I think this test and the next one are redundant.
They don't even need to inject a custom `KeyManager` with a hard-coded seed to test what you want.
You can group them in a single test that does the following:

- Use the default `KeyManager` (no need to inject a custom seed)
- Sign a message with `signMessage`
- Verify it with `verifyMessage`
- Ensure that the recovered `signerNodeId` matches `kit.nodeParams.nodeId` and verify that the signature is correct (with "normal" signature validation with `Crypto.verifySignature`)

eclair-core/src/test/scala/fr/acinq/eclair/crypto/LocalKeyManagerSpec.scala
```diff
@@ -141,4 +141,21 @@ class LocalKeyManagerSpec extends AnyFunSuite {
     assert(keyManager.htlcPoint(channelKeyPath).publicKey == PrivateKey(hex"b1be27b5232e3bc5d6a261949b4ee68d96fa61f481998d36342e2ad99444cf8a").publicKey)
     assert(keyManager.commitmentSecret(channelKeyPath, 0).value == ShaChain.shaChainFromSeed(ByteVector32.fromValidHex("eeb3bad6808e8bb5f1774581ccf64aa265fef38eca80a1463d6310bb801b3ba7"), 0xFFFFFFFFFFFFL))
   }
+
+  test("generate a signature from a digest") {
+    val seed = ByteVector.fromValidHex("deadbeef")
```
t-bast, line 146: nit:
```suggestion
    val seed = hex"deadbeef"
```

eclair-core/src/test/scala/fr/acinq/eclair/crypto/LocalKeyManagerSpec.scala
```diff
@@ -141,4 +141,21 @@ class LocalKeyManagerSpec extends AnyFunSuite {
...
+    val expectedRecoveryId = 1
+    val expectedPublicKey = PublicKey(hex"02ec8f6801d53fba134d032af7bc33ce1fd3d6373743f1317272d6a7d05ee9b30d")
+    val digest = ByteVector32(hex"d7914fe546b684688bb95f4f888a92dfc680603a75f23eb823658031fff766d9") // sha256(sha256("hello"))
+
+    assert(privkey.publicKey === expectedPublicKey)
+
+    val (signature, recid) = testKeyManager.signDigest(digest, privkey)
+    assert(signature === expectedSignature)
+    assert(recid === expectedRecoveryId)
+  }
```
t-bast, line 160: Same as with the eclair tests, I think this test doesn't gain anything by checking things against hard-coded values.
A better and simpler test (in my opinion) would be:

```suggestion
  test("generate a signature from a digest") {
    val seed = ByteVector.fromValidHex("deadbeef")
    val testKeyManager = new LocalKeyManager(seed, Block.RegtestGenesisBlock.hash)
    val digest = ByteVector32(hex"d7914fe546b684688bb95f4f888a92dfc680603a75f23eb823658031fff766d9") // sha256(sha256("hello"))

    val (signature, recid) = testKeyManager.signDigest(digest)
    val recoveredPubkey = Crypto.recoverPublicKey(signature, signedBytes, recoveryId)
    assert(recoveredPubkey === testKeyManager.nodeId)
    assert(Crypto.verifySignature(digest, signature, testKeyManager.nodeId))
  }
```

eclair-node/src/main/scala/fr/acinq/eclair/api/ExtraDirectives.scala
```diff
@@ -20,13 +20,14 @@ import akka.http.scaladsl.marshalling.ToResponseMarshaller
 import akka.http.scaladsl.model.StatusCodes.NotFound
 import akka.http.scaladsl.model.{ContentTypes, HttpResponse}
 import akka.http.scaladsl.server.{Directive1, Directives, MalformedFormFieldRejection, Route}
-import fr.acinq.bitcoin.ByteVector32
+import fr.acinq.bitcoin.{ByteVector32, ByteVector64}
```
t-bast, line 23: nit: remove changes from this file? ctrl + alt + o

eclair-node/src/main/scala/fr/acinq/eclair/api/FormParamExtractors.scala
```diff
@@ -59,6 +59,8 @@ object FormParamExtractors {
 
   implicit val millisatoshiUnmarshaller: Unmarshaller[String, MilliSatoshi] = Unmarshaller.strict { str => MilliSatoshi(str.toLong) }
 
+  implicit val base64DataUnmarshaller: Unmarshaller[String, ByteVector] = Unmarshaller.strict { str => ByteVector.fromValidBase64(str) }
```
t-bast, line 62: How is that being used? It looks like you're not base64-decoding (yet) string inputs from the API

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
dkrm0: I don't think the format of the message is that much important. If we receive base64 or hexadecimal data we just encode the ASCII representation of that data.

I don't see why this command:
```
eclair-cli signmessage --msg=$(echo -n "I was here" | base64)
```

should be preferable to this:
```
eclair-cli signmessage --msg="I was here"
```

If someone wants to encode binaries data, they're free to use base64, hexadecimal or some exotic encoding like zbase32 for the ASCII representation.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
dkrm0: They put it at the front. See https://twitter.com/rusty_twit/status/1182102005914800128 #specinatweet

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/crypto/KeyManager.scala
dkrm0: Done

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: It is very important, otherwise we won't sign the same thing as lnd or c-lightning and won't be compatible.
*What you sign are bytes, not strings,* so you absolutely need to specify the encoding that converts between the two, otherwise the user doesn't know what is really signed.
Note that ASCII (or UTF8) _is_ an encoding, just like base64.

With your current code, the two commands you mention will **not** result in the same signature.
So how can the user know which one will match lnd and c-lightning?
We want the user to provide us the message as *bytes*, but since these are messy to handle in CLI/API we simply ask that these bytes be provided as a base64 string.

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
dkrm0: Done

eclair-core/src/test/scala/fr/acinq/eclair/crypto/LocalKeyManagerSpec.scala
dkrm0: Done

eclair-core/src/test/scala/fr/acinq/eclair/crypto/LocalKeyManagerSpec.scala
dkrm0: Done

eclair-node/src/main/scala/fr/acinq/eclair/api/FormParamExtractors.scala
dkrm0: Yes, I forgot to remove it.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
dkrm0: Right now we sign the same thing as lnd and c-lightning.

This command:
```
eclair-cli signmessage --msg="I was here"
```
will produce a signature that lnd and c-lightning can verify like this:
```
lightning-cli checkmessage 'I was here' ${signature}
lncli verifymessage --msg 'I was here' --sig ${signature}
```

If I do what you asks, the user will need to use this command instead:
```
eclair-cli signmessage --msg=$(echo -n 'I was here' | base64)
```
It puts the burden on the users to remember to convert in base64 the message they want to sign.

eclair-node/src/main/scala/fr/acinq/eclair/api/ExtraDirectives.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -64,6 +68,10 @@ object TimestampQueryFilters {
   }
 }
 
+object SignedMessage {
+  def signedBytes(message: ByteVector): ByteVector32 = Crypto.hash256(ByteVector("Lightning Signed Message:".getBytes) ++ message)
```
t-bast, line 72: lnd answered me on Twitter and told me that Go guarantees that the conversion from string to bytes will always use UTF8 regardless of the machine's locale, so we should force UTF8 for the prefix:

```suggestion
  def signedBytes(message: ByteVector): ByteVector32 = Crypto.hash256(ByteVector("Lightning Signed Message:".getBytes(StandardCharsets.UTF_8)) ++ message)
```

eclair-core/src/main/scala/fr/acinq/eclair/crypto/KeyManager.scala
```diff
@@ -110,6 +110,20 @@ trait KeyManager {
    *         private key, bitcoinSig is the signature of the channel announcement with our funding private key
    */
   def signChannelAnnouncement(fundingKeyPath: DeterministicWallet.KeyPath, chainHash: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, remoteFundingKey: PublicKey, features: Features): (ByteVector64, ByteVector64)
+
+  /**
+   * Sign a digest, primarily used to prove ownership of the current node
+   *
+   * When recovering a public key from an ECDSA signature for secp256k1, there are 4 possible matching curve points
+   * that can be found. The recoveryId identifies which of these points is the correct one to speed up recovery
+   * (otherwise we would need to check the signature's validity for each of these points to find the correct one).
```
t-bast, line 119: We realized that this is incorrect, all points will yield a valid signature...

```suggestion
   * that can be found. The recoveryId identifies which of these points is the correct one.
```

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/crypto/KeyManager.scala
dkrm0: Done

PR #1500, opened 2020-07-30
Improve blockchain watchers

t-bast:
- Correctly handle the case where a tx has both a sequence and locktime set (anchor output htlcs)
- Add tx publish tests to ZmqWatcher and ElectrumWatcher
- Add documentation on watcher types

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
@@ -26,8 +26,8 @@ import scodec.bits.ByteVector
 import scala.util.{Failure, Success, Try}
 
 /**
-  * Created by PM on 19/01/2016.
-  */
+ * Created by PM on 19/01/2016.
+ */
 
 // @formatter:off
 
@@ -35,11 +35,19 @@ sealed trait Watch {
   def channel: ActorRef
   def event: BitcoinEvent
 }
-// we need a public key script to use electrum apis
+
+/**
+ * Watch for confirmation of a given transaction.
+ *
+ * @param channel         channel to notify once the transaction is confirmed.
+ * @param txId            txid of the transaction to watch.
+ * @param publicKeyScript when using electrum, we need to specify a public key script; any of the output scripts should work.
+ * @param minDepth        number of confirmations.
+ * @param event           channel event related to the transaction.
+ */
 final case class WatchConfirmed(channel: ActorRef, txId: ByteVector32, publicKeyScript: ByteVector, minDepth: Long, event: BitcoinEvent) extends Watch
 object WatchConfirmed {
-  // if we have the entire transaction, we can get the redeemScript from the witness, and re-compute the publicKeyScript
-  // we support both p2pkh and p2wpkh scripts
+  // if we have the entire transaction, we can get the publicKeyScript from any of the outputs
   def apply(channel: ActorRef, tx: Transaction, minDepth: Long, event: BitcoinEvent): WatchConfirmed = WatchConfirmed(channel, tx.txid, tx.txOut.map(_.publicKeyScript).headOption.getOrElse(ByteVector.empty), minDepth, event)
 
   def extractPublicKeyScript(witness: ScriptWitness): ByteVector = Try(PublicKey(witness.stack.last)) match {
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
```diff
@@ -26,20 +26,28 @@ import scodec.bits.ByteVector
...
 /**
-  * Created by PM on 19/01/2016.
-  */
+ * Created by PM on 19/01/2016.
+ */
 
 // @formatter:off
 
 sealed trait Watch {
   def channel: ActorRef
```
pm47, line 35: `replyTo` is all the rage these days and it's actually a better name since for `WatchSpentBasic` it's the router.
```suggestion
  def replyTo: ActorRef
```

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcherSpec.scala
```diff
@@ -146,8 +155,102 @@ class ZmqWatcherSpec extends TestKitBaseClass with AnyFunSuiteLike with Bitcoind
...
+
+    // tx1 has an absolute delay but no relative delay
+    val tx1 = {
+      wallet.fundTransaction(Transaction(2, Nil, TxOut(150000 sat, Script.pay2wpkh(priv.publicKey)) :: Nil, blockCount.get + 5), lockUnspents = true, FeeratePerKw(250 sat)).pipeTo(probe.ref)
+      val funded = probe.expectMsgType[FundTransactionResponse].tx
+      wallet.signTransaction(funded).pipeTo(probe.ref)
+      probe.expectMsgType[SignTransactionResponse].tx
+    }
+    probe.send(watcher, PublishAsap(tx1))
+    generateBlocks(bitcoincli, 5)
```
pm47, line 219: Regarding my comment about potential off-by-one errors, how about:
```suggestion
    probe.send(watcher, PublishAsap(tx1))
    generateBlocks(bitcoincli, 4)
    probe.expectNoMsg()
    generateBlocks(bitcoincli, 1)
```

Maybe too paranoid?

pm47 (commented):
I wonder if there is any chance we could factor tests for Bitcoind/Electrum.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcherSpec.scala
t-bast: Nope, that makes sense!
Even though if we end up broadcasting too soon, `bitcoind` will reject the tx and many tests will fail.
If we end up broadcasting too late, then we're wasting a 10 minutes delay but at least we're safe :)

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
```diff
@@ -220,11 +219,15 @@ class ElectrumWatcher(blockCount: AtomicLong, client: ActorRef) extends Actor wi
       context become disconnected(watches, sent.map(PublishAsap), block2tx, Queue.empty)
   }
 
+  def publish(tx: Transaction): Unit = {
+    log.info(s"publishing tx=$tx")
```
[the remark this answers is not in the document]
t-bast: Sounds good, I'll revert some of them.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -165,24 +165,25 @@ class ZmqWatcher(blockCount: AtomicLong, client: ExtendedBitcoinClient)(implicit
...
         val block2tx1 = block2tx.updated(cltvTimeout, block2tx.getOrElse(cltvTimeout, Seq.empty[Transaction]) :+ tx)
-        context become watching(watches, watchedUtxos, block2tx1, None)
+        context become watching(watches, watchedUtxos, block2tx1, nextTick)
       } else publish(tx)
 
     case WatchEventConfirmed(BITCOIN_PARENT_TX_CONFIRMED(tx), blockHeight, _, _) =>
-      log.info(s"parent tx of txid=${tx.txid} has been confirmed")
+      log.info("parent tx of txid={} has been confirmed", tx.txid)
       val blockCount = this.blockCount.get()
+      val cltvTimeout = Scripts.cltvTimeout(tx)
```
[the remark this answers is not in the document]
t-bast: Agreed, at some point we can migrate quite a few things to bitcoin-lib, there are also some helpers on bitcoin amounts that could be migrated

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcherSpec.scala
pm47: There are probably several places where this pattern could be used.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
```diff
@@ -74,8 +74,8 @@ class ElectrumWatcher(blockCount: AtomicLong, client: ActorRef) extends Actor wi
           client ! ElectrumClient.GetScriptHashHistory(scriptHash)
       }
       val toPublish = block2tx.filterKeys(_ <= newheight)
-      toPublish.values.flatten.foreach(tx => self ! PublishAsap(tx))
-      context become running(newheight, newtip, watches, scriptHashStatus, block2tx -- toPublish.keys, sent)
+      toPublish.values.flatten.foreach(publish)
+      context become running(newheight, newtip, watches, scriptHashStatus, block2tx -- toPublish.keys, sent ++ toPublish.values.flatten)
```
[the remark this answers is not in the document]
t-bast: We were previously relying on the fact that the handler for `PublishAsap` added the tx to this `sent` queue.
Now we are directly publishing these txs (line 77) so we need to add them to the `sent` queue.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcherSpec.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1500/commits/fc4e6b0c72d3d743124f8615141e8f858eb9d03c, this is really the only test that benefited explicitly from this.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1500/commits/699b6d9ef8f725340a61cf8048e494848a358bc9

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1500/commits/7f319935c023e03fa83e2dc5c0e702d26f63c212

t-bast:
> I wonder if there is any chance we could factor tests for Bitcoind/Electrum.

That was my first intuition as well, but in fact they don't behave the same, especially with our testing helpers so it's quite hard...

pm47 (approved):

PR #1508, opened 2020-08-10
Maven Wrapper

SalomonBrys:
The [Maven Wrapper](https://github.com/takari/maven-wrapper) is a project that has been officially accepted as part of the future Maven 3.7.0 release.

It allows the project to:

- contain its own build system
- define precisely the Maven version used to build
- be easier to build (no dependency needed outside of the JDK)

This PR adds a Maven wrapper to the project and defines the Maven version to download and use as the latest stable at the time (3.6.3).


t-bast:
It's hard to find expected release dates about maven, but since their last release was more than a year ago maybe we can expect maven 3.7 to release soon-ish, and integrate this once we move to 3.7?

PR #1513, opened 2020-08-12
Add DB entry for payment router error

### LINKED ISSUE #1512: Early payment failures are not reflected in `getsentinfo`
When payments fail before a `PaymentLifecycle` FSM is instantiated, nothing is logged in the DB.
This is confusing for users that send payments through the API/CLI.

In particular, for MPP payments for which we can't find a route, nothing will be written to the DB, whereas non-MPP payments that don't find a route will have an entry in the DB.

t-bast:
When using MPP, if we can't find a route, we need to add an entry to the
DB. Otherwise when users query their payment status, nothing will be
returned which is a bad UX.

Fixes #1512

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
@@ -23,6 +23,7 @@ import akka.actor.{ActorRef, FSM, Props, Status}
 import akka.event.Logging.MDC
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.eclair.db.{OutgoingPayment, OutgoingPaymentStatus, PaymentType}
 import fr.acinq.eclair.payment.Monitoring.{Metrics, Tags}
 import fr.acinq.eclair.payment.OutgoingPacket.Upstream
 import fr.acinq.eclair.payment.PaymentRequest.ExtraHop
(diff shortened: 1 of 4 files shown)


pm47:
At first sight, I'm not convinced by the rationale as laid out in the description of the PR and would agree with [@fiatjaf's comment](https://github.com/ACINQ/eclair/issues/1512#issuecomment-673691382), *but* the inconsistency between non-mpp and mpp definitely needs solving. Have you considered removing the db write in the non-mpp case?

t-bast:
I agree that the way it's implemented feels a bit hackish...

> Have you considered removing the db write in the non-mpp case?

It would require moving the initial DB write in `PaymentLifecycle` at a later stage, which I think is going in the wrong direction...the reason why we started by storing something in the DB was to ensure an entry was there for bookkeeping and let users know what happened with the payment!

I agree that in theory we could instead directly answer the `payinvoice` call with that error, but that's really hard because we're using actor message translated to HTTP response, and we can only have one response, so only the first message (which contains the paymentId) is sent back to the caller...

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -216,7 +217,19 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
 
   private def gotoAbortedOrStop(d: PaymentAborted): State = {
     if (d.pending.isEmpty) {
-      myStop(d.request.replyTo, Left(PaymentFailed(id, paymentHash, d.failures)))
+      val pf = PaymentFailed(id, paymentHash, d.failures)
+      if (cfg.storeInDb) {
+        // In cases where we fail early (router error during the first attempt), the DB won't have an entry for that
+        // payment, which may be confusing for users.
+        nodeParams.db.payments.listOutgoingPayments(cfg.parentId) match {
```
pm47, line 224: Writing to db here forces you to call `db.payments.listOutgoingPayments` in order to tell whether there were child payments. Why don't you do it earlier, when you have that information at hand?

```scala
when(WAIT_FOR_ROUTES) {
  ...
  case Event(Status.Failure(t), d: PaymentProgress) =>
    if (d.pending.isEmpty && d.failures.isEmpty) {
    // no child payment was ever attempted, write to db here
    }
  ..
}
```

t-bast: Good idea, done!

pm47 (approved):

PR #1514, opened 2020-08-13
Use the `replyTo` pattern for Register messages

pm47:
In preparation of the migration to typed actors, we need to remove the
use of `sender`, which doesn't exist in typed actors (and even returns
dead letters when mixing typed/untyped actors).

This also means that, in the tests, we should stop using the
`TestProbe.send()` method, which also relies on the recipient replying
to `sender`. Instead, we should use `targetActor ! msg` which guarantees
that the sender is void.

Using the `replyTo` pattern doesn't mix well with the `ask` pattern,
because we don't know the reference to the temporary actor. To deal with
that, we set `replyTo = ActorRef.noSender` which is a bit hackish.

NB: the [akka doc](https://doc.akka.io/docs/akka/current/typed/interaction-patterns.html#request-response-with-ask-from-outside-an-actor) shows the proper way of dealing with `ask` with typed actors:

```scala
val result: Future[CookieFabric.Reply] = cookieFabric.ask(ref => CookieFabric.GiveMeCookies(3, ref))
```

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
@@ -62,16 +62,20 @@ class Register extends Actor with ActorLogging {
 
     case Symbol("channelsTo") => sender ! channelsTo
 
-    case fwd@Forward(channelId, msg) =>
+    case fwd@Forward(replyTo, channelId, msg) =>
+      // for backward compatibility with legacy ask, we use the replyTo as sender
+      val compatReplyTo = if (replyTo == ActorRef.noSender) sender else replyTo
       channels.get(channelId) match {
-        case Some(channel) => channel forward msg
-        case None => sender ! Failure(ForwardFailure(fwd))
+        case Some(channel) => channel.tell(msg, compatReplyTo)
+        case None => compatReplyTo ! Failure(ForwardFailure(fwd))
       }
 
-    case fwd@ForwardShortId(shortChannelId, msg) =>
+    case fwd@ForwardShortId(replyTo, shortChannelId, msg) =>
+      // for backward compatibility with legacy ask, we use the replyTo as sender
+      val compatReplyTo = if (replyTo == ActorRef.noSender) sender else replyTo
       shortIds.get(shortChannelId).flatMap(channels.get) match {
-        case Some(channel) => channel forward msg
-        case None => sender ! Failure(ForwardShortIdFailure(fwd))
+        case Some(channel) => channel.tell(msg, compatReplyTo) // for backward compatibility, we use the replyTo as sender
+        case None => compatReplyTo ! Failure(ForwardShortIdFailure(fwd))
       }
   }
 }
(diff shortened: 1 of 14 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
```diff
@@ -62,25 +62,29 @@ class Register extends Actor with ActorLogging {
...
       }
 
-    case fwd@ForwardShortId(shortChannelId, msg) =>
+    case fwd@ForwardShortId(replyTo, shortChannelId, msg) =>
+      // for backward compatibility with legacy ask, we use the replyTo as sender
+      val compatReplyTo = if (replyTo == ActorRef.noSender) sender else replyTo
       shortIds.get(shortChannelId).flatMap(channels.get) match {
-        case Some(channel) => channel forward msg
-        case None => sender ! Failure(ForwardShortIdFailure(fwd))
+        case Some(channel) => channel.tell(msg, compatReplyTo) // for backward compatibility, we use the replyTo as sender
```
t-bast, line 77: nit:
```suggestion
        case Some(channel) => channel.tell(msg, compatReplyTo)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -234,6 +236,10 @@ object Relayer extends Logging {
       isPublic = commitments.announceChannel)
   }
   case class OutgoingChannels(channels: Seq[OutgoingChannel])
+
+  // internal classes, used for testing
+  private[payment] case class GetChildActors(replyTo: ActorRef)
+  private[payment] case class ChildActors(postRestartCleaner: ActorRef, channelRelayer: ActorRef, nodeRelayer: ActorRef)
```
t-bast, line 242: This is nice! I had to do a very ugly workaround [here](https://github.com/ACINQ/eclair/blob/2c437421d7d03435a761a76796ff438a93295617/eclair-core/src/test/scala/fr/acinq/eclair/payment/RelayerSpec.scala#L558), can you fix this too?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
```diff
@@ -62,25 +62,29 @@ class Register extends Actor with ActorLogging {
 
     case Symbol("channelsTo") => sender ! channelsTo
 
-    case fwd@Forward(channelId, msg) =>
+    case fwd@Forward(replyTo, channelId, msg) =>
+      // for backward compatibility with legacy ask, we use the replyTo as sender
+      val compatReplyTo = if (replyTo == ActorRef.noSender) sender else replyTo
```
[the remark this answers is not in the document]
pm47: Correct

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
pm47: Done with 3b1a4d5fa. I went one step further in another PR and added those child actors to the fixture.

pm47:
Rebased

t-bast (approved):

PR #1515, opened 2020-08-14
Route search fix

btcontract:
I guess provided test describes this best so I'll walk through what it does:

- First, routes `((1, 2) and (1, 4))` are successfully found, but when `split` called it turns out their total capacity is `30 sat` while we try to send a `60 sat` payment.
- Next, what this code change does is it removes a number of lowest capacity channels and runs a search once again in hope that more appropriate routes will be found this time.
- Second attempt turns out to be successful with routes `((1, 5) and (1, 4))`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -289,6 +289,7 @@ object RouteCalculation {
     }
   }
 
+  @tailrec
   private def findMultiPartRouteInternal(g: DirectedGraph,
                                          localNodeId: PublicKey,
                                          targetNodeId: PublicKey,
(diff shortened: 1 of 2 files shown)


t-bast:
Congrats on getting through this particularly tricky part of the code ;)

However I think there is no issue there and your proposal kind of violates layers.
You are right that the behaviour of the path-finding algorithm is that it will miss the potential `(1, 5) + (1, 4)` split because it orders routes based on fee/length/cltv according to the `WeightRatio` configured.

But that's because you're using `randomize = false` and a small `maxParts`.
IMHO retrying should be done in upper layers, should set `randomize` to `true` (see what's done in the `MultiPartPaymentLifecycle.scala`) and potentially set a higher `maxParts`, which would find the `(1, 5) + (1, 4)` route.

What do you think?

t-bast:
Another important point is that you're not setting `WeightRatio` at all in your test. Setting those with a non-0 `capacityFactor` would ensure the (1, 5) channel would appear before the (1, 2) channel, completely avoiding this issue in the first place.

btcontract:
I don't have a particular opinion on a fix, this just seemed like incorrect behavior to me because, although maybe it's a rather synthetic case, a payment does get failed right away while a retry could be used. Fine with me if this is OK in practice.

t-bast:
This is a low-level component that isn't directly accessible to users, this is why I think it's ok that it behaves that way.

We could retry at that layer like your proposal does, but there is an efficiency trade-off to consider. MPP is already more costly than single-part payment because dijkstra is run several times, so we'd need to accept not always getting the optimal results. My state of mind right now is that randomization is the best way of dealing with LN path-finding because of the uncertainty of route balances, which is why the payment state machines retry with randomization activated if the first attempt fails (and the scenario described in your test should result in a payment success because of this E2E behavior).

PR #1518, opened 2020-08-17
Set the channel version in the init msg for the fundee

pm47:
Currently, the fundee computes it after receiving the `open_channel`
message, by calling `ChannelVersion.pickChannelVersion`.

Instead, we call this method in the `Peer` and add the resulting channel
version to the initialization parameters.

In the end, the behavior is exactly the same, but:
- it is more consistent with how the funder works;
- it may make sense to compute the channel version a bit earlier in the
process, because we may initiate a different kind of channel actor in
the future for some particular versions (?)


t-bast (approved):
LGTM

PR #1520, opened 2020-08-19
Use the `replyTo` pattern for Channel messages

pm47:
Builds on #1516.

This turned out to be a significant change impacting many files and is **best reviewed commit by commit**.

I pretty happy with the new model that cleanly encapsulates what was before the `HtlcAddFailed` exception and the `ForwardMessage`, and the new handling of disconnections which is not an awkward corner case anymore.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1352,7 +1352,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       // for our outgoing payments, let's send events if we know that they will settle on chain
       Closing
         .onchainOutgoingHtlcs(d.commitments.localCommit, d.commitments.remoteCommit, d.commitments.remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit), tx)
-        .map(add => (add, d.commitments.originChannels.get(add.id).collect { case o: Origin.Local => o.upstream.id })) // we resolve the payment id if this was a local payment
+        .map(add => (add, d.commitments.originChannels.get(add.id).collect { case o: Origin.Local => o.id })) // we resolve the payment id if this was a local payment
         .collect { case (add, Some(id)) => context.system.eventStream.publish(PaymentSettlingOnChain(id, amount = add.amountMsat, add.paymentHash)) }
       // and we also send events related to fee
       Closing.networkFeePaid(tx, d1) foreach { case (fee, desc) => feePaid(fee, tx, desc, d.channelId) }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -632,7 +632,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       // note: spec would allow us to keep sending new htlcs after having received their shutdown (and not sent ours)
       // but we want to converge as fast as possible and they would probably not route them anyway
       val error = NoMoreHtlcsClosingInProgress(d.channelId)
-      handleAddHtlcCommandError(c, d.channelId, error, origin(c), Some(d.channelUpdate))
+      handleAddHtlcCommandError(c, error, Some(d.channelUpdate))
 
     case Event(c: CMD_ADD_HTLC, d: DATA_NORMAL) =>
       Commitments.sendAdd(d.commitments, c, origin(c), nodeParams.currentBlockHeight, nodeParams.onChainFeeConf) match {
@@ -640,7 +640,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           if (c.commit) self ! CMD_SIGN
           context.system.eventStream.publish(AvailableBalanceChanged(self, d.channelId, d.shortChannelId, commitments1))
           handleCommandSuccess(c, d.copy(commitments = commitments1)) sending add
-        case Failure(cause) => handleAddHtlcCommandError(c, d.channelId, cause, origin(c), Some(d.channelUpdate))
+        case Failure(cause) => handleAddHtlcCommandError(c, cause, Some(d.channelUpdate))
       }
 
     case Event(add: UpdateAddHtlc, d: DATA_NORMAL) =>
@@ -976,18 +976,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         log.info(s"updating channel_update announcement (reason=disabled)")
         val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.localCommit.spec.totalFunds, enable = false)
         d.commitments.localChanges.proposed.collect {
-//          case add: UpdateAddHtlc =>
-//            relayer ! RES_ADD_FAILED(d.channelId, add.paymentHash, ChannelUnavailable(d.channelId), d.commitments.originChannels(add.id), Some(channelUpdate), None)
-          case add: UpdateAddHtlc =>
-            val addFailed = RES_ADD_FAILED(d.channelId, add.paymentHash, ChannelUnavailable(d.channelId), d.commitments.originChannels(add.id), Some(channelUpdate), None)
-            d.commitments.originChannels(add.id) match {
-              case Origin.Local(_, None) => () // TODO: we used to log an error here
-              case Origin.Local(_, Some(replyTo)) => replyTo ! addFailed
-              case Origin.Relayed(_, _, _, _, None) => () // TODO: this case didn't exist before
-              case Origin.Relayed(_, _, _, _, Some(replyTo)) => replyTo ! addFailed
-              case Origin.TrampolineRelayed(_, None) => ()// TODO: we used to log an error here
-              case Origin.TrampolineRelayed(_, Some(replyTo)) => replyTo ! addFailed
-            }
+          case add: UpdateAddHtlc => relayer ! RES_ADD_COMPLETED(Relayer.RelayBackward.RelayFailDisconnected(channelUpdate, d.commitments.originChannels(add.id), add))
         }
         d.copy(channelUpdate = channelUpdate)
       } else {
@@ -1665,7 +1654,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     case Event(c: CMD_ADD_HTLC, d: HasCommitments) =>
       log.info(s"rejecting htlc request in state=$stateName")
       val error = ChannelUnavailable(d.channelId)
-      handleAddHtlcCommandError(c, d.channelId, error, origin(c), None) // we don't provide a channel_update: this will be a permanent channel failure
+      handleAddHtlcCommandError(c, error, None) // we don't provide a channel_update: this will be a permanent channel failure
 
     case Event(c: CMD_CLOSE, d) => handleCommandError(CommandUnavailableInThisState(Helpers.getChannelId(d), "close", stateName), c)
 
@@ -1872,13 +1861,13 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     stay using newData
   }
 
-  def handleAddHtlcCommandError(c: CMD_ADD_HTLC, channelId: ByteVector32, cause: Throwable, origin: Origin, channelUpdate: Option[ChannelUpdate]) = {
+  def handleAddHtlcCommandError(c: CMD_ADD_HTLC, cause: Throwable, channelUpdate: Option[ChannelUpdate]) = {
     log.warning(s"${cause.getMessage} while processing cmd=${c.getClass.getSimpleName} in state=$stateName")
     val replyTo = c match {
       case hasReplyTo: HasReplyTo if hasReplyTo.replyTo != ActorRef.noSender => hasReplyTo.replyTo
       case _ => sender
     }
-    replyTo ! RES_ADD_FAILED(channelId, c.paymentHash, cause, origin, channelUpdate, Some(c))
+    replyTo ! RES_ADD_FAILED(c, cause, channelUpdate)
     context.system.eventStream.publish(ChannelErrorOccurred(self, Helpers.getChannelId(stateData), remoteNodeId, stateData, LocalError(cause), isFatal = false))
     stay
   }
@@ -1984,7 +1973,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     } else {
       // channel is already disabled, we reply to the request
       val error = ChannelUnavailable(d.channelId)
-      handleAddHtlcCommandError(c, d.channelId, error, origin(c), Some(d.channelUpdate)) // can happen if we are in OFFLINE or SYNCING state (channelUpdate will have enable=false)
+      handleAddHtlcCommandError(c, error, Some(d.channelUpdate)) // can happen if we are in OFFLINE or SYNCING state (channelUpdate will have enable=false)
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -665,7 +665,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       Commitments.receiveFulfill(d.commitments, fulfill) match {
         case Success((commitments1, origin, htlc)) =>
           // we forward preimages as soon as possible to the upstream channel because it allows us to pull funds
-          relayer ! Relayer.ForwardRemoteFulfill(fulfill, origin, htlc)
+          relayer ! RES_ADD_COMPLETED(Relayer.RelayBackward.RelayRemoteFulfill(fulfill, origin, htlc))
           stay using d.copy(commitments = commitments1)
         case Failure(cause) => handleLocalError(cause, d, Some(fulfill))
       }
@@ -782,9 +782,13 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         case Success((commitments1, forwards)) =>
           cancelTimer(RevocationTimeout.toString)
           log.debug("received a new rev, spec:\n{}", Commitments.specs2String(commitments1))
-          forwards.foreach { forward =>
-            log.debug("forwarding {} to relayer", forward)
-            relayer ! forward
+          forwards.foreach {
+            case forwardAdd: Relayer.RelayForward =>
+              log.debug("forwarding {} to relayer", forwardAdd)
+              relayer ! forwardAdd
+            case forward: Relayer.RelayBackward =>
+              log.debug("forwarding {} to relayer", forward)
+              relayer ! RES_ADD_COMPLETED(forward)
           }
           if (Commitments.localHasChanges(commitments1) && d.commitments.remoteNextCommitInfo.left.map(_.reSignAsap) == Left(true)) {
             self ! CMD_SIGN
@@ -1024,7 +1028,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       Commitments.receiveFulfill(d.commitments, fulfill) match {
         case Success((commitments1, origin, htlc)) =>
           // we forward preimages as soon as possible to the upstream channel because it allows us to pull funds
-          relayer ! Relayer.ForwardRemoteFulfill(fulfill, origin, htlc)
+          relayer ! RES_ADD_COMPLETED(Relayer.RelayBackward.RelayRemoteFulfill(fulfill, origin, htlc))
           stay using d.copy(commitments = commitments1)
         case Failure(cause) => handleLocalError(cause, d, Some(fulfill))
       }
@@ -1132,13 +1136,13 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           cancelTimer(RevocationTimeout.toString)
           log.debug("received a new rev, spec:\n{}", Commitments.specs2String(commitments1))
           forwards.foreach {
-            case forwardAdd: Relayer.ForwardAdd =>
+            case forwardAdd: Relayer.RelayForward =>
               // BOLT 2: A sending node SHOULD fail to route any HTLC added after it sent shutdown.
               log.debug("closing in progress: failing {}", forwardAdd.add)
               self ! CMD_FAIL_HTLC(forwardAdd.add.id, Right(PermanentChannelFailure), commit = true)
-            case forward =>
+            case forward: Relayer.RelayBackward =>
               log.debug("forwarding {} to relayer", forward)
-              relayer ! forward
+              relayer ! RES_ADD_COMPLETED(forward)
           }
           if (commitments1.hasNoPendingHtlcs) {
             log.debug("switching to NEGOTIATING spec:\n{}", Commitments.specs2String(commitments1))
@@ -1298,7 +1302,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         d.commitments.originChannels.get(htlc.id) match {
           case Some(origin) =>
             log.info(s"fulfilling htlc #${htlc.id} paymentHash=${htlc.paymentHash} origin=$origin")
-            relayer ! Relayer.ForwardOnChainFulfill(preimage, origin, htlc)
+            relayer ! RES_ADD_COMPLETED(Relayer.RelayBackward.RelayOnChainFulfill(preimage, origin, htlc))
           case None =>
             // if we don't have the origin, it means that we already have forwarded the fulfill so that's not a big deal.
             // this can happen if they send a signature containing the fulfill, then fail the channel before we have time to sign it
@@ -1337,7 +1341,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         d.commitments.originChannels.get(add.id) match {
           case Some(origin) =>
             log.info(s"failing htlc #${add.id} paymentHash=${add.paymentHash} origin=$origin: htlc timed out")
-            relayer ! Relayer.ForwardOnChainFail(HtlcsTimedoutDownstream(d.channelId, Set(add)), origin, add)
+            relayer ! RES_ADD_COMPLETED(Relayer.RelayBackward.RelayOnChainFail(HtlcsTimedoutDownstream(d.channelId, Set(add)), origin, add))
           case None =>
             // same as for fulfilling the htlc (no big deal)
             log.info(s"cannot fail timedout htlc #${add.id} paymentHash=${add.paymentHash} (origin not found)")
@@ -1348,7 +1352,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         d.commitments.originChannels.get(add.id) match {
           case Some(origin) =>
             log.info(s"failing htlc #${add.id} paymentHash=${add.paymentHash} origin=$origin: overridden by local commit")
-            relayer ! Relayer.ForwardOnChainFail(HtlcOverriddenByLocalCommit(d.channelId, add), origin, add)
+            relayer ! RES_ADD_COMPLETED(Relayer.RelayBackward.RelayOnChainFail(HtlcOverriddenByLocalCommit(d.channelId, add), origin, add))
           case None =>
             // same as for fulfilling the htlc (no big deal)
             log.info(s"cannot fail overridden htlc #${add.id} paymentHash=${add.paymentHash} (origin not found)")
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -632,7 +632,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       // note: spec would allow us to keep sending new htlcs after having received their shutdown (and not sent ours)
       // but we want to converge as fast as possible and they would probably not route them anyway
       val error = NoMoreHtlcsClosingInProgress(d.channelId)
-      handleCommandError(AddHtlcFailed(d.channelId, c.paymentHash, error, origin(c), Some(d.channelUpdate), Some(c)), c)
+      handleAddHtlcCommandError(c, d.channelId, error, origin(c), Some(d.channelUpdate))
 
     case Event(c: CMD_ADD_HTLC, d: DATA_NORMAL) =>
       Commitments.sendAdd(d.commitments, c, origin(c), nodeParams.currentBlockHeight, nodeParams.onChainFeeConf) match {
@@ -640,7 +640,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           if (c.commit) self ! CMD_SIGN
           context.system.eventStream.publish(AvailableBalanceChanged(self, d.channelId, d.shortChannelId, commitments1))
           handleCommandSuccess(c, d.copy(commitments = commitments1)) sending add
-        case Failure(cause) => handleCommandError(AddHtlcFailed(d.channelId, c.paymentHash, cause, origin(c), Some(d.channelUpdate), Some(c)), c)
+        case Failure(cause) => handleAddHtlcCommandError(c, d.channelId, cause, origin(c), Some(d.channelUpdate))
       }
 
     case Event(add: UpdateAddHtlc, d: DATA_NORMAL) =>
@@ -972,7 +972,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         log.info(s"updating channel_update announcement (reason=disabled)")
         val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.localCommit.spec.totalFunds, enable = false)
         d.commitments.localChanges.proposed.collect {
-          case add: UpdateAddHtlc => relayer ! Status.Failure(AddHtlcFailed(d.channelId, add.paymentHash, ChannelUnavailable(d.channelId), d.commitments.originChannels(add.id), Some(channelUpdate), None))
+          case add: UpdateAddHtlc => relayer ! RES_ADD_FAILED(d.channelId, add.paymentHash, ChannelUnavailable(d.channelId), d.commitments.originChannels(add.id), Some(channelUpdate), None)
         }
         d.copy(channelUpdate = channelUpdate)
       } else {
@@ -1650,7 +1650,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     case Event(c: CMD_ADD_HTLC, d: HasCommitments) =>
       log.info(s"rejecting htlc request in state=$stateName")
       val error = ChannelUnavailable(d.channelId)
-      handleCommandError(AddHtlcFailed(d.channelId, c.paymentHash, error, origin(c), None, Some(c)), c) // we don't provide a channel_update: this will be a permanent channel failure
+      handleAddHtlcCommandError(c, d.channelId, error, origin(c), None) // we don't provide a channel_update: this will be a permanent channel failure
 
     case Event(c: CMD_CLOSE, d) => handleCommandError(CommandUnavailableInThisState(Helpers.getChannelId(d), "close", stateName), c)
 
@@ -1681,7 +1681,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     case Event("disconnecting", _) => stay
 
     // when we realize we need to update our network fees, we send a CMD_UPDATE_FEE to ourselves which may result in this error being sent back to ourselves, this can be ignored
-    case Event(Status.Failure(_: CannotAffordFees), _) => stay
+    case Event(RES_FAILURE(_: CMD_UPDATE_FEE, _: CannotAffordFees), _) => stay
 
     // funding tx was confirmed in time, let's just ignore this
     case Event(BITCOIN_FUNDING_TIMEOUT, _: HasCommitments) => stay
@@ -1857,14 +1857,30 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     stay using newData
   }
 
-  def handleCommandError(cause: Throwable, cmd: Command) = {
-    log.warning(s"${cause.getMessage} while processing cmd=${cmd.getClass.getSimpleName} in state=$stateName")
+  def handleAddHtlcCommandError(c: CMD_ADD_HTLC, channelId: ByteVector32, cause: Throwable, origin: Origin, channelUpdate: Option[ChannelUpdate]) = {
+    log.warning(s"${cause.getMessage} while processing cmd=${c.getClass.getSimpleName} in state=$stateName")
+    val replyTo = c match {
+      case hasReplyTo: HasReplyTo if hasReplyTo.replyTo != ActorRef.noSender => hasReplyTo.replyTo
+      case _ => sender
+    }
+    replyTo ! RES_ADD_FAILED(channelId, c.paymentHash, cause, origin, channelUpdate, Some(c))
+    context.system.eventStream.publish(ChannelErrorOccurred(self, Helpers.getChannelId(stateData), remoteNodeId, stateData, LocalError(cause), isFatal = false))
+    stay
+  }
+
+  def handleCommandError(cause: Throwable, c: Command) = {
+    log.warning(s"${cause.getMessage} while processing cmd=${c.getClass.getSimpleName} in state=$stateName")
     cause match {
       case _: ChannelException => ()
-      case _ => log.error(cause, s"msg=$cmd stateData=$stateData ")
+      case _ => log.error(cause, s"msg=$c stateData=$stateData ")
     }
+    val replyTo = c match {
+      case hasReplyTo: HasReplyTo if hasReplyTo.replyTo != ActorRef.noSender => hasReplyTo.replyTo
+      case _ => sender
+    }
+    replyTo ! RES_FAILURE(c, cause)
     context.system.eventStream.publish(ChannelErrorOccurred(self, Helpers.getChannelId(stateData), remoteNodeId, stateData, LocalError(cause), isFatal = false))
-    stay replying Status.Failure(cause)
+    stay
   }
 
   /**
@@ -1952,7 +1968,8 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       goto(stateName) using d.copy(channelUpdate = channelUpdate)
     } else {
       // channel is already disabled, we reply to the request
-      handleCommandError(AddHtlcFailed(d.channelId, c.paymentHash, ChannelUnavailable(d.channelId), origin(c), Some(d.channelUpdate), Some(c)), c) // can happen if we are in OFFLINE or SYNCING state (channelUpdate will have enable=false)
+      val error = ChannelUnavailable(d.channelId)
+      handleAddHtlcCommandError(c, d.channelId, error, origin(c), Some(d.channelUpdate)) // can happen if we are in OFFLINE or SYNCING state (channelUpdate will have enable=false)
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -972,7 +972,18 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         log.info(s"updating channel_update announcement (reason=disabled)")
         val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.localCommit.spec.totalFunds, enable = false)
         d.commitments.localChanges.proposed.collect {
-          case add: UpdateAddHtlc => relayer ! RES_ADD_FAILED(d.channelId, add.paymentHash, ChannelUnavailable(d.channelId), d.commitments.originChannels(add.id), Some(channelUpdate), None)
+//          case add: UpdateAddHtlc =>
+//            relayer ! RES_ADD_FAILED(d.channelId, add.paymentHash, ChannelUnavailable(d.channelId), d.commitments.originChannels(add.id), Some(channelUpdate), None)
+          case add: UpdateAddHtlc =>
+            val addFailed = RES_ADD_FAILED(d.channelId, add.paymentHash, ChannelUnavailable(d.channelId), d.commitments.originChannels(add.id), Some(channelUpdate), None)
+            d.commitments.originChannels(add.id) match {
+              case Origin.Local(_, None) => () // TODO: we used to log an error here
+              case Origin.Local(_, Some(replyTo)) => replyTo ! addFailed
+              case Origin.Relayed(_, _, _, _, None) => () // TODO: this case didn't exist before
+              case Origin.Relayed(_, _, _, _, Some(replyTo)) => replyTo ! addFailed
+              case Origin.TrampolineRelayed(_, None) => ()// TODO: we used to log an error here
+              case Origin.TrampolineRelayed(_, Some(replyTo)) => replyTo ! addFailed
+            }
         }
         d.copy(channelUpdate = channelUpdate)
       } else {
@@ -2338,7 +2349,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
   def origin(c: CMD_ADD_HTLC): Origin = c.upstream match {
     case Upstream.Local(id) => Origin.Local(id, Some(c.replyTo)) // we were the origin of the payment
-    case Upstream.Relayed(u) => Origin.Relayed(u.channelId, u.id, u.amountMsat, c.amount) // this is a relayed payment to an outgoing channel
+    case Upstream.Relayed(u) => Origin.Relayed(u.channelId, u.id, u.amountMsat, c.amount, Some(c.replyTo)) // this is a relayed payment to an outgoing channel
     case Upstream.TrampolineRelayed(us) => Origin.TrampolineRelayed(us.map(u => (u.channelId, u.id)).toList, Some(c.replyTo)) // this is a relayed payment to an outgoing node
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -2337,9 +2337,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   }
 
   def origin(c: CMD_ADD_HTLC): Origin = c.upstream match {
-    case Upstream.Local(id) => Origin.Local(id, Some(sender)) // we were the origin of the payment
+    case Upstream.Local(id) => Origin.Local(id, Some(c.replyTo)) // we were the origin of the payment
     case Upstream.Relayed(u) => Origin.Relayed(u.channelId, u.id, u.amountMsat, c.amount) // this is a relayed payment to an outgoing channel
-    case Upstream.TrampolineRelayed(us) => Origin.TrampolineRelayed(us.map(u => (u.channelId, u.id)).toList, Some(sender)) // this is a relayed payment to an outgoing node
+    case Upstream.TrampolineRelayed(us) => Origin.TrampolineRelayed(us.map(u => (u.channelId, u.id)).toList, Some(c.replyTo)) // this is a relayed payment to an outgoing node
   }
 
   def feePaid(fee: Satoshi, tx: Transaction, desc: String, channelId: ByteVector32): Unit = Try { // this may fail with an NPE in tests because context has been cleaned up, but it's not a big deal
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -30,7 +30,6 @@ import fr.acinq.eclair.crypto.ShaChain
 import fr.acinq.eclair.db.PendingRelayDb
 import fr.acinq.eclair.io.Peer
 import fr.acinq.eclair.payment._
-import fr.acinq.eclair.payment.relay.{Origin, Relayer}
 import fr.acinq.eclair.router.Announcements
 import fr.acinq.eclair.transactions.Transactions.TxOwner
 import fr.acinq.eclair.transactions._
@@ -635,7 +634,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       handleAddHtlcCommandError(c, error, Some(d.channelUpdate))
 
     case Event(c: CMD_ADD_HTLC, d: DATA_NORMAL) =>
-      Commitments.sendAdd(d.commitments, c, origin(c), nodeParams.currentBlockHeight, nodeParams.onChainFeeConf) match {
+      Commitments.sendAdd(d.commitments, c, nodeParams.currentBlockHeight, nodeParams.onChainFeeConf) match {
         case Success((commitments1, add)) =>
           if (c.commit) self ! CMD_SIGN
           context.system.eventStream.publish(AvailableBalanceChanged(self, d.channelId, d.shortChannelId, commitments1))
@@ -665,7 +664,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       Commitments.receiveFulfill(d.commitments, fulfill) match {
         case Success((commitments1, origin, htlc)) =>
           // we forward preimages as soon as possible to the upstream channel because it allows us to pull funds
-          relayer ! RES_ADD_COMPLETED(Relayer.RelayBackward.RelayRemoteFulfill(fulfill, origin, htlc))
+          relayer ! RES_ADD_COMPLETED(origin, htlc, HtlcResult.RemoteFulfill(fulfill))
           stay using d.copy(commitments = commitments1)
         case Failure(cause) => handleLocalError(cause, d, Some(fulfill))
       }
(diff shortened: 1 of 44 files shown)


pm47:
Sidenote: should probably rename `RES_ADD_COMPLETED` to `RES_ADD_SETTLED` for consistency with existing terminology.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -145,11 +196,51 @@ final case class RES_GETINFO(nodeId: PublicKey, channelId: ByteVector32, state:
...
+  sealed trait Fulfill extends HtlcResult { def paymentPreimage: ByteVector32 }
+  case class RemoteFulfill(fulfill: UpdateFulfillHtlc) extends Fulfill { override val paymentPreimage = fulfill.paymentPreimage }
+  case class OnChainFulfill(paymentPreimage: ByteVector32) extends Fulfill
+  sealed trait Fail extends HtlcResult
+  case class RemoteFail(fail: UpdateFailHtlc) extends Fail
+  case class RemoteFailMalformed(fail: UpdateFailMalformedHtlc) extends Fail
+  case class OnChainFail(cause: ChannelException) extends Fail
+  case class Disconnected(channelUpdate: ChannelUpdate) extends Fail { assert(!Announcements.isEnabled(channelUpdate.channelFlags), "channel update must have disabled flag set") }
+}
+final case class RES_ADD_COMPLETED[+O <: Origin[_ <: Upstream], +R <: HtlcResult](to: O, htlc: UpdateAddHtlc, result: R) extends CommandSuccess[CMD_ADD_HTLC]
```
pm47, line 226: Sidenote: should probably rename for consistency with existing terminology.

```suggestion
final case class RES_ADD_SETTLED[+O <: Origin[_ <: Upstream], +R <: HtlcResult](to: O, htlc: UpdateAddHtlc, result: R) extends CommandSuccess[CMD_ADD_HTLC]
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -145,11 +196,51 @@ final case class RES_GETINFO(nodeId: PublicKey, channelId: ByteVector32, state:
...
+final case class RES_SUCCESS[+C <: Command](cmd: C, channelId: ByteVector32) extends CommandSuccess[C]
+final case class RES_FAILURE[+C <: Command, +T <: Throwable](cmd: C, t: T) extends CommandFailure[C, T]
+
+/**
+ * special case for [[CMD_ADD_HTLC]]
+ * note that for this command there is gonna be two response patterns:
+ * - either [[RES_ADD_FAILED]]
+ * - or [[RES_SUCCESS[CMD_ADD_HTLC]]] followed by [[RES_ADD_COMPLETED]] (possibly a while later)
+ */
+final case class RES_ADD_FAILED[+T <: Throwable](c: CMD_ADD_HTLC, t: T, channelUpdate: Option[ChannelUpdate]) extends CommandFailure[CMD_ADD_HTLC, T] { override def toString = s"cannot add htlc with origin=${c.origin} reason=${t.getMessage}" }
```
t-bast, line 214: It would be nice to be able to restrict to `+T <: ChannelException`, isn't is possible already?
Can't you simplify the signature like that:

```
final case class RES_ADD_FAILED(c: CMD_ADD_HTLC, t: ChannelException, channelUpdate: Option[ChannelUpdate]) extends CommandFailure[CMD_ADD_HTLC, ChannelException]
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
t-bast, line 226: I think the `to` field is misleading, shouldn't this be either a `from` or an `origin`?
It made sense when that type was called `ForwardXXX` but not anymore.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -145,11 +196,51 @@ final case class RES_GETINFO(nodeId: PublicKey, channelId: ByteVector32, state:
...
+final case class RES_GETSTATE[+S <: State](state: S) extends CommandSuccess[CMD_GETSTATE.type]
+final case class RES_GETSTATEDATA[+D <: Data](data: D) extends CommandSuccess[CMD_GETSTATEDATA.type]
+final case class RES_GETINFO(nodeId: PublicKey, channelId: ByteVector32, state: State, data: Data) extends CommandSuccess[CMD_GETINFO.type]
+final case class RES_CLOSE(channelId: ByteVector32) extends CommandSuccess[CMD_CLOSE]
+
+/**
+ * Those are not response to [[Command]], but to [[fr.acinq.eclair.io.Peer.OpenChannel]]
+ *
+ * If actor A sends a [[fr.acinq.eclair.io.Peer.OpenChannel]] and actor B sends a [[CMD_CLOSE]], then A will receive a
+ * [[ChannelOpenResponse.ChannelClosed]] whereas B will receive a [[]]
```
t-bast, line 238: What B will receive seems to be missing...it's a `[[RES_CLOSE]]` IIUC

eclair-core/src/main/scala/fr/acinq/eclair/payment/Monitoring.scala
```diff
@@ -87,7 +87,8 @@ object Monitoring {
     val Success = "success"
     val Failure = "failure"
 
-    object FailureType {
+    object
+    FailureType {
```
t-bast, line 91: nit: revert?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -144,7 +144,7 @@ object PaymentFailure {
    */
   def transformForUser(failures: Seq[PaymentFailure]): Seq[PaymentFailure] = {
     failures.map {
-      case LocalFailure(hops, AddHtlcFailed(_, _, t, _, _, _)) => LocalFailure(hops, t) // we're interested in the error which caused the add-htlc to fail
+      case LocalFailure(hops, t) => LocalFailure(hops, t) // we're interested in the error which caused the add-htlc to fail
```
t-bast, line 147: Now that this is just a passthrough, the whole `.map` is unnecessary, you can remove it:
```
failures match {
      case previousFailures :+ LocalFailure(_, RouteNotFound) if previousFailures.nonEmpty => previousFailures
      case other => other
}
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -56,35 +57,41 @@ class ChannelRelayer(nodeParams: NodeParams, relayer: ActorRef, register: ActorR
...
-    case ChannelCommandResponse.Ok => // ignoring responses from channels
+    case _: RES_SUCCESS[_] => // ignoring responses from channels
   }
 
   override def mdc(currentMessage: Any): MDC = {
     val paymentHash_opt = currentMessage match {
       case relay: RelayHtlc => Some(relay.r.add.paymentHash)
       case Register.ForwardShortIdFailure(Register.ForwardShortId(_, _, c: CMD_ADD_HTLC)) => Some(c.paymentHash)
-      case Status.Failure(addFailed: AddHtlcFailed) => Some(addFailed.paymentHash)
+      case addFailed: RES_ADD_FAILED[_] @unchecked => Some(addFailed.c.paymentHash)
```
t-bast, line 94: I don't think you need the `@unchecked`, do you?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -148,51 +130,25 @@ class Relayer(nodeParams: NodeParams, router: ActorRef, register: ActorRef, paym
...
 
   override def mdc(currentMessage: Any): MDC = {
     val paymentHash_opt = currentMessage match {
-      case ForwardAdd(add, _) => Some(add.paymentHash)
-      case Status.Failure(addFailed: AddHtlcFailed) => Some(addFailed.paymentHash)
-      case ff: ForwardFulfill => Some(ff.htlc.paymentHash)
-      case ff: ForwardFail => Some(ff.htlc.paymentHash)
+      case RelayForward(add, _) => Some(add.paymentHash)
+      case addFailed: RES_ADD_FAILED[_] @unchecked => Some(addFailed.c.paymentHash)
+      case addCompleted: RES_ADD_COMPLETED[_, _] @unchecked => Some(addCompleted.htlc.paymentHash)
```
t-bast, line 151: I think the `@unchecked` aren't necessary.
nit2: there are unused imports in this file.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -129,12 +130,12 @@ class PaymentInitiator(nodeParams: NodeParams, router: ActorRef, register: Actor
           val trampolineSecret = r.trampolineSecret.getOrElse(randomBytes32)
           sender ! SendPaymentToRouteResponse(paymentId, parentPaymentId, Some(trampolineSecret))
           val (trampolineAmount, trampolineExpiry, trampolineOnion) = buildTrampolinePayment(SendTrampolinePaymentRequest(r.recipientAmount, r.paymentRequest, trampoline, Seq((r.trampolineFees, r.trampolineExpiryDelta)), r.fallbackFinalExpiryDelta), r.trampolineFees, r.trampolineExpiryDelta)
-          payFsm forward SendPaymentToRoute(Left(r.route), Onion.createMultiPartPayload(r.amount, trampolineAmount, trampolineExpiry, trampolineSecret, Seq(OnionTlv.TrampolineOnion(trampolineOnion))), r.paymentRequest.routingInfo)
+          payFsm forward SendPaymentToRoute(sender, Left(r.route), Onion.createMultiPartPayload(r.amount, trampolineAmount, trampolineExpiry, trampolineSecret, Seq(OnionTlv.TrampolineOnion(trampolineOnion))), r.paymentRequest.routingInfo)
```
t-bast, line 133:
```suggestion
          payFsm ! SendPaymentToRoute(sender, Left(r.route), Onion.createMultiPartPayload(r.amount, trampolineAmount, trampolineExpiry, trampolineSecret, Seq(OnionTlv.TrampolineOnion(trampolineOnion))), r.paymentRequest.routingInfo)
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -156,28 +156,30 @@ object ChannelCodecs extends Logging {
       ("sentAfterLocalCommitIndex" | uint64overflow) ::
       ("reSignAsap" | bool8)).as[WaitingForRevocation]
 
-  val localCodec: Codec[Origin.Local] = (
-    ("id" | uuid) ::
-      ("sender" | provide(Option.empty[ActorRef]))
-    ).as[Origin.Local]
+  val localColdCodec: Codec[Origin.LocalCold] = (
+    ("id" | uuid)).as[Upstream.Local].as[Origin.LocalCold]
```
t-bast, line 160: nit: unnecessary parenthesis:
```suggestion
  val localColdCodec: Codec[Origin.LocalCold] = ("id" | uuid).as[Upstream.Local].as[Origin.LocalCold]
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -156,28 +156,30 @@ object ChannelCodecs extends Logging {
...
     ("originChannelId" | bytes32) ::
       ("originHtlcId" | int64) ::
       ("amountIn" | millisatoshi) ::
-      ("amountOut" | millisatoshi)).as[Origin.Relayed]
+      ("amountOut" | millisatoshi)).as[Upstream.ChannelRelayedCold].as[Origin.ChannelRelayedCold]
+
+  val relayedCodec: Codec[Origin.ChannelRelayed] = relayedColdCodec.xmap[Origin.ChannelRelayed](o => o: Origin.ChannelRelayed, o => Origin.ChannelRelayedCold(Upstream.ChannelRelayedCold(o.upstream.originChannelId, o.upstream.originHtlcId, o.upstream.amountIn, o.upstream.amountOut)))
+
+  val trampolineRelayedColdCodec: Codec[Origin.TrampolineRelayedCold] = (
+    listOfN(uint16, bytes32 ~ int64)).as[Upstream.TrampolineRelayedCold].as[Origin.TrampolineRelayedCold]
```
t-bast, line 173: nit: unnecessary parenthesis:
```suggestion
  val trampolineRelayedColdCodec: Codec[Origin.TrampolineRelayedCold] = 
    listOfN(uint16, bytes32 ~ int64).as[Upstream.TrampolineRelayedCold].as[Origin.TrampolineRelayedCold]
```

t-bast (commented):
First pass done commit-by-commit, I now have a good mental model of all the changes.
My biggest feedback before starting a more thorough review is on the type hierarchy for `Origin`/`Upstream`, I think we can go further and simplify them (see comment).

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -105,34 +107,83 @@ case class BITCOIN_PARENT_TX_CONFIRMED(childTx: Transaction) extends BitcoinEven
...
  */
 
-sealed trait Upstream
-object Upstream {
+/**
+ * Origin of a payment, answering both questions:
+ * - what actor in the app sent that htlc? (Origin.replyTo)
+ * - what are the parent(s) of this payment in the htlc chain? (Upstream)
+ */
+sealed trait Origin[+U <: Origin.Upstream] { def upstream: U }
```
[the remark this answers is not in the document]
t-bast: Here is my proposal:

```scala
  /**
   * Origin of a payment, answering both questions:
   * - what actor in the app sent that htlc? (Origin.replyTo)
   * - what are the upstream parent(s) of this payment in the htlc chain?
   */
  sealed trait Origin
  /** We haven't restarted since we sent the payment downstream: the origin actor is known. */
  sealed trait Hot extends Origin { def replyTo: ActorRef }
  /** We have restarted after the payment was sent, we have limited info and the origin actor doesn't exist anymore. */
  sealed trait Cold extends Origin

  /** The downstream HTLCs have been sent by a payment FSM and may be retried in case of failures. */
  sealed trait HasPaymentLifecycle extends Origin

  /** Our node is the origin of the payment. */
  sealed trait Local extends HasPaymentLifecycle { def id: UUID }
  case class LocalHot(replyTo: ActorRef, id: UUID) extends Local with Hot
  case class LocalCold(id: UUID) extends Local with Cold

  /** Our node forwarded a single incoming HTLC to an outgoing channel. */
  sealed trait ChannelRelayed extends Origin {
    def originChannelId: ByteVector32
    def originHtlcId: Long
    def amountIn: MilliSatoshi
    def amountOut: MilliSatoshi
  }
  case class ChannelRelayedHot(replyTo: ActorRef, add: UpdateAddHtlc, override val amountOut: MilliSatoshi) extends ChannelRelayed with Hot {
    override def originChannelId: ByteVector32 = add.channelId
    override def originHtlcId: Long = add.id
    override def amountIn: MilliSatoshi = add.amountMsat
  }
  case class ChannelRelayedCold(originChannelId: ByteVector32, originHtlcId: Long, amountIn: MilliSatoshi, amountOut: MilliSatoshi) extends ChannelRelayed with Cold

  /** Our node forwarded an incoming HTLC set to a remote outgoing node (potentially producing multiple downstream HTLCs).*/
  sealed trait TrampolineRelayed extends HasPaymentLifecycle { def htlcs: List[(ByteVector32, Long)] }
  case class TrampolineRelayedHot(replyTo: ActorRef, adds: Seq[UpdateAddHtlc]) extends TrampolineRelayed with Hot {
    override def htlcs: List[(ByteVector32, Long)] = adds.map(u => (u.channelId, u.id)).toList
    val amountIn: MilliSatoshi = adds.map(_.amountMsat).sum
    val expiryIn: CltvExpiry = adds.map(_.cltvExpiry).min
  }
  case class TrampolineRelayedCold(override val htlcs: List[(ByteVector32, Long)]) extends TrampolineRelayed with Cold
```

There are two distinct ways to then pattern match on `Origin`:
* separate them by Hot vs Cold (post-restart-handler handles only cold, relayers handle only hot and then pattern match further)
* separate them by type of relay (Local, ChannelRelayed, TrampolineRelayed) and then pattern match on hot vs cold

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -105,34 +107,83 @@ case class BITCOIN_PARENT_TX_CONFIRMED(childTx: Transaction) extends BitcoinEven
...
+  /** Our node forwarded an incoming HTLC set to a remote outgoing node (potentially producing multiple downstream HTLCs).*/
+  sealed trait TrampolineRelayed extends Origin[Upstream.TrampolineRelayed]
+  case class TrampolineRelayedHot(upstream: Upstream.TrampolineRelayedHot, replyTo: ActorRef) extends TrampolineRelayed with Hot[Upstream.TrampolineRelayed]
+  case class TrampolineRelayedCold(upstream: Upstream.TrampolineRelayedCold) extends TrampolineRelayed with Cold[Upstream.TrampolineRelayed]
+
+  /** Parents of a payment in the htlc chain */
+  sealed trait Upstream
+  object Upstream {
+    /** there is a payment lifecycle involved for local and trampoline payments */
+    sealed trait SentByPaymentLifecycle extends Upstream
```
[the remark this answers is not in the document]
t-bast: In fact I don't think this trait is useful at all. The only place where we really use it in the `SendPaymentConfig` of the `PaymentInitiator`, but using `Hot` there would be enough.
The only pattern matching we do on it is in `MultiPartPaymentLifecycle.spawnChildPaymentFsm` where we just need to do something for `Local` origin, and leave other kinds of origin unchanged.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1840,18 +1850,44 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     }
   }
 
-  def handleCommandSuccess(sender: ActorRef, newData: Data) = {
-    stay using newData replying ChannelCommandResponse.Ok
+  def handleFastClose(c: CloseCommand, channelId: ByteVector32) = {
+    goto(CLOSED) replying RES_SUCCESS(c, channelId)
```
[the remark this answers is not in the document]
pm47: I'm aware of that but chose to not address it right now.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1840,18 +1850,44 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+    val replyTo = c match {
+      case hasReplyTo: HasReplyTo if hasReplyTo.replyTo != ActorRef.noSender => hasReplyTo.replyTo
+      case _ => sender
+    }
+    val channelId = Helpers.getChannelId(newData)
+    replyTo ! RES_SUCCESS(c, channelId)
+    stay using newData
+  }
+
+  def handleAddHtlcCommandError(c: CMD_ADD_HTLC, cause: Throwable, channelUpdate: Option[ChannelUpdate]) = {
```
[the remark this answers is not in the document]
pm47: True, although it also would break the symmetry with other `Commitments` functions. For some operations (e.g. signature) we can't restrict to known `ChannelException`. Hmm... should we encapsulate those is a `BitcoinlibException`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -105,34 +107,83 @@ case class BITCOIN_PARENT_TX_CONFIRMED(childTx: Transaction) extends BitcoinEven
...
-object Upstream {
+/**
+ * Origin of a payment, answering both questions:
+ * - what actor in the app sent that htlc? (Origin.replyTo)
+ * - what are the parent(s) of this payment in the htlc chain? (Upstream)
+ */
+sealed trait Origin[+U <: Origin.Upstream] { def upstream: U }
+object Origin {
+  /** We have full details and the origin actor is known. */
+  sealed trait Hot[+U <: Upstream] extends Origin[U]
```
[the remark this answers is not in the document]
pm47: This sounds like a logical thing to do, but that actor needs to be typed, something like:

```scala
sealed trait Hot[+U <: Upstream] extends Origin[U] { def replyTo: ActorRef[RES_ADD_COMPLETED[Hot[U], HtlcResult]] }
```

I briefly looked into it and type isn't trivial to get right. Feels like we're close to hitting the limit of what can be done with vanilla scala.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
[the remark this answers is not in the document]
t-bast: Agreed, that's probably the last thing to do once we've fully changed the channel actor to use a clean `replyTo`/`command` pattern everywhere

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
pm47: I mean, this change makes complete sense now, but the `ActorRef` would need to be typed *in the future*, so we need to be careful not to put ourselves in a corner.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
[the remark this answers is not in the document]
t-bast: Good question...probably for later though, even if we had such a `BitcoinlibException` I think it's still useful that every method returns the "smallest" type it can, so `sendAdd` could still return only `ChannelExceptions`

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
pm47: Actually, the reason why I introduced this abstraction is to avoid an impossible match/case here:

```scala
def buildCommand(replyTo: ActorRef, upstream: Upstream.SentByPaymentLifecycle, paymentHash: ByteVector32, hops: Seq[ChannelHop], finalPayload: Onion.FinalPayload): (CMD_ADD_HTLC, Seq[(ByteVector32, PublicKey)]) = {
    val (firstAmount, firstExpiry, onion) = buildPacket(Sphinx.PaymentPacket)(paymentHash, hops, finalPayload)
    val origin = upstream match {
      case u: Upstream.Local => Origin.LocalHot(u, replyTo)
      case u: Upstream.TrampolineRelayedHot => Origin.TrampolineRelayedHot(u, replyTo)
    }
    CMD_ADD_HTLC(replyTo, firstAmount, paymentHash, firstExpiry, onion.packet, origin, commit = true) -> onion.sharedSecrets
  }
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
t-bast: I heard Scala 3 has much better handling of generics, variance, and could probably fill the last gaps for us :)

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
pm47: It would be nice, but I can't right now, because we're relying on generic `Try` for commitment methods.  See https://github.com/ACINQ/eclair/pull/1520#discussion_r482796394. However we could move to a `scodec`-style custom error messages but a bit of work is needed.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
t-bast: But is it really harmful to add `case u: Upstream.ChannelRelayedHot => Origin.ChannelRelayedHot(u, replyTo)` here? I don't have a very strong opinion, but I think the `Origin` type hierarchy is a bit complex already...

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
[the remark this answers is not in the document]
pm47: That's sounds reasonable, I'll look into it.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
pm47: It depends how hard we want to rely on the typing system to prevent us from making mistakes. Having a `Upstream.ChannelRelayedHot` is a bug that the compiler could detect.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
t-bast: We can also chose to add it for now since it fits the current code, and remove it if the typing becomes horrible when moving to Akka Typed?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
t-bast: That's true, let's keep it then.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
pm47: Agreed, I'll try that right away.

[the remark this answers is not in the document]
pm47: Wait, I remember now. There are cases where we know the `Upstream` but we don't yet have the sender. When sending trampoline payments if my memory serves right. Will still try though...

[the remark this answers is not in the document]
t-bast: If that's the case, maybe we can add a dedicated type but only for that specific case (instead of having every relay scenario have both a hot and cold upstream type)? I'm looking into it as well :)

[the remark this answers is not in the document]
t-bast: > Not sure where to put the new Upstream trait though.

I think that since it originates from `PaymentInitiator` (and only used for `SendPaymentConfig`) it's fine where you put it for now.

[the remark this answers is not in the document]
pm47: Done with 4a25b39c4.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -105,34 +107,83 @@ case class BITCOIN_PARENT_TX_CONFIRMED(childTx: Transaction) extends BitcoinEven
...
+  case class TrampolineRelayedCold(upstream: Upstream.TrampolineRelayedCold) extends TrampolineRelayed with Cold[Upstream.TrampolineRelayed]
+
+  /** Parents of a payment in the htlc chain */
+  sealed trait Upstream
+  object Upstream {
+    /** there is a payment lifecycle involved for local and trampoline payments */
+    sealed trait SentByPaymentLifecycle extends Upstream
+
+    /** Our node is the origin of the payment. */
+    final case class Local(id: UUID) extends Upstream with SentByPaymentLifecycle // we don't persist reference to local actors
```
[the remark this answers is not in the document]
pm47: Done with 4a25b39.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
pm47: Done with 4a25b39 and 35a676350.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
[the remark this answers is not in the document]
pm47: The issue disappears with 4a25b39.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
pm47: Done with 35a6763.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
pm47: Done with a0ae9f81e.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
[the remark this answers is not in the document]
pm47: Done with a0ae9f81e.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
pm47: Done with c46fff704.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Monitoring.scala
pm47: Done with c46fff704.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
pm47: Hawk eyes 👍 . Done with c46fff704.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
pm47: Done with c46fff704.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
pm47: Done with c46fff704.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
pm47: Done with c46fff704.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: Done with c46fff704.

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: Done with 89ffc887b.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
pm47: Done with 27f11ffc3.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -148,51 +128,21 @@ class Relayer(nodeParams: NodeParams, router: ActorRef, register: ActorRef, paym
...
 
-    case Status.Failure(addFailed: AddHtlcFailed) => addFailed.origin match {
-      case Origin.Local(id, None) => log.error(s"received unexpected add failed with no sender (paymentId=$id)")
-      case Origin.Local(_, Some(sender)) => sender ! Status.Failure(addFailed)
-      case _: Origin.Relayed => channelRelayer forward Status.Failure(addFailed)
-      case Origin.TrampolineRelayed(htlcs, None) => log.error(s"received unexpected add failed with no sender (upstream=${htlcs.mkString(", ")}")
-      case Origin.TrampolineRelayed(_, Some(paymentSender)) => paymentSender ! Status.Failure(addFailed)
+    case r: RES_ADD_SETTLED[_, _] => r.origin match {
+      case _: Origin.Cold => postRestartCleaner ! r
+      case o: Origin.Hot => o.replyTo ! r
```
pm47, line 133: Next logical step is to do that in the `Channel`, since it's completely stateless. Would that go to far?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -151,7 +151,23 @@ class PostRestartHtlcCleaner(nodeParams: NodeParams, register: ActorRef, initial
...
+          if (relayedOut1.isEmpty) {
+            log.info(s"payment with paymentHash=${fulfilledHtlc.paymentHash} successfully relayed")
+            // We could emit a TrampolinePaymentRelayed event but that requires more book-keeping on incoming HTLCs.
+            // It seems low priority so isn't done at the moment but can be added when we feel we need it.
+            Metrics.PendingRelayedOut.decrement()
+            context become main(brokenHtlcs.copy(relayedOut = brokenHtlcs.relayedOut - origin, settledUpstream = brokenHtlcs.settledUpstream - origin))
+          } else {
+            context become main(brokenHtlcs.copy(relayedOut = brokenHtlcs.relayedOut + (origin -> relayedOut1), settledUpstream = brokenHtlcs.settledUpstream + origin))
+          }
+
```
t-bast, line 169: nit: remove empty line?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -151,7 +151,23 @@ class PostRestartHtlcCleaner(nodeParams: NodeParams, register: ActorRef, initial
...
+        case Origin.ChannelRelayedCold(originChannelId, originHtlcId, _, _) =>
+          if (!brokenHtlcs.settledUpstream.contains(origin)) {
+            log.info(s"received preimage for paymentHash=${fulfilledHtlc.paymentHash}: fulfilling 1 HTLC upstream")
+            PendingRelayDb.safeSend(register, nodeParams.db.pendingRelay, originChannelId, CMD_FULFILL_HTLC(originHtlcId, paymentPreimage, commit = true))
+          }
+          val relayedOut1 = relayedOut diff Set((fulfilledHtlc.channelId, fulfilledHtlc.id))
+          if (relayedOut1.isEmpty) {
+            log.info(s"payment with paymentHash=${fulfilledHtlc.paymentHash} successfully relayed")
+            // We could emit a TrampolinePaymentRelayed event but that requires more book-keeping on incoming HTLCs.
+            // It seems low priority so isn't done at the moment but can be added when we feel we need it.
```
t-bast, line 163: In that case I think we can (and should) emit a `ChannelPaymentRelayed`

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -151,7 +151,23 @@ class PostRestartHtlcCleaner(nodeParams: NodeParams, register: ActorRef, initial
           // instead spread across multiple local origins) so we can now forget this origin.
           Metrics.PendingRelayedOut.decrement()
           context become main(brokenHtlcs.copy(relayedOut = brokenHtlcs.relayedOut - origin))
-        case Origin.TrampolineRelayed(origins, _) =>
+        case Origin.ChannelRelayedCold(originChannelId, originHtlcId, _, _) =>
+          if (!brokenHtlcs.settledUpstream.contains(origin)) {
+            log.info(s"received preimage for paymentHash=${fulfilledHtlc.paymentHash}: fulfilling 1 HTLC upstream")
+            PendingRelayDb.safeSend(register, nodeParams.db.pendingRelay, originChannelId, CMD_FULFILL_HTLC(originHtlcId, paymentPreimage, commit = true))
+          }
+          val relayedOut1 = relayedOut diff Set((fulfilledHtlc.channelId, fulfilledHtlc.id))
```
t-bast, line 159: In the channel-relay case, `relayedOut` should always have a single element (one HTLC in, one HTLC out).
The code we need should be a bit simpler:

```scala
case Origin.ChannelRelayedCold(originChannelId, originHtlcId, amountIn, amountOut) =>
          log.info(s"received preimage for paymentHash=${fulfilledHtlc.paymentHash}: fulfilling 1 HTLC upstream")
          if (relayedOut != Set((fulfilledHtlc.channelId, fulfilledHtlc.id))) {
            log.error(s"unexpected channel relay downstream HTLCs: expected (${fulfilledHtlc.channelId},${fulfilledHtlc.id}), found $relayedOut")
          }
          PendingRelayDb.safeSend(register, nodeParams.db.pendingRelay, originChannelId, CMD_FULFILL_HTLC(originHtlcId, paymentPreimage, commit = true))
          context.system.eventStream.publish(ChannelPaymentRelayed(amountIn, amountOut, fulfilledHtlc.paymentHash, originChannelId, fulfilledHtlc.channelId))
          Metrics.PendingRelayedOut.decrement()
          context become main(brokenHtlcs.copy(relayedOut = brokenHtlcs.relayedOut - origin))

```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -193,24 +206,27 @@ class PostRestartHtlcCleaner(nodeParams: NodeParams, register: ActorRef, initial
...
                   context.system.eventStream.publish(PaymentFailed(p.parentId, failedHtlc.paymentHash, Nil))
                 }
               })
-              case Origin.TrampolineRelayed(origins, _) =>
-                log.warning(s"payment failed for paymentHash=${failedHtlc.paymentHash}: failing ${origins.length} upstream HTLCs")
+              case Origin.ChannelRelayedCold(originChannelId, originHtlcId, _, _) =>
+                log.warning(s"payment failed for paymentHash=${failedHtlc.paymentHash}: failing 1 HTLC upstream")
+                Metrics.Resolved.withTag(Tags.Success, value = false).withTag(Metrics.Relayed, value = true).increment()
+                // We don't bother decrypting the downstream failure to forward a more meaningful error upstream, it's
+                // very likely that it won't be actionable anyway because of our node restart.
```
t-bast, line 220: Actually that's not true for channel relay.
I think we should pipe the `HtlcResult.Fail` and pattern match on it to forward the error onion upstream when available.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
t-bast: If we want to move to a full `request-reply` model and move that into the channel, we'll need to promote the `PostRestartHtlcCleaner`: it shouldn't be a child of the `Relayer` anymore but rather a top-level actor that the channel knows about. Maybe worth a try in a later PR?

pm47: > we'll need to promote the PostRestartHtlcCleaner: it shouldn't be a child of the Relayer anymore

Indeed, see my early unfinished prototyped on branch `channel-postrestart`.

> Maybe worth a try in a later PR?

Okay

eclair-core/src/test/scala/fr/acinq/eclair/payment/RelayerSpec.scala
```diff
@@ -467,32 +466,25 @@ class RelayerSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike {
...
-    assert(register.expectMsgType[Register.Forward[CMD_FAIL_HTLC]].message.reason === Right(TemporaryChannelFailure(channelUpdate_bc)))
-
+    val upstreamAdd = UpdateAddHtlc(channelId_ab, 42, 1_100_000 msat, paymentHash, CltvExpiry(288), TestConstants.emptyOnionPacket)
+    val origin = Origin.ChannelRelayedHot(channelRelayer, upstreamAdd, 1_000_000.msat)
+    val cmdAdd = CMD_ADD_HTLC(channelRelayer, 1_000_000.msat, paymentHash, CltvExpiry(144), TestConstants.emptyOnionPacket, origin)
+
+    assert(ChannelRelayer.translateError(ExpiryTooSmall(channelId_bc, CltvExpiry(100), CltvExpiry(0), 0), Some(channelUpdate_bc)) === ExpiryTooSoon(channelUpdate_bc))
+    assert(ChannelRelayer.translateError(ExpiryTooBig(channelId_bc, CltvExpiry(100), CltvExpiry(200), 0), Some(channelUpdate_bc)) === ExpiryTooFar)
+    assert(ChannelRelayer.translateError(InsufficientFunds(channelId_bc, cmdAdd.amount, 100 sat, 0 sat, 0 sat), Some(channelUpdate_bc)) === TemporaryChannelFailure(channelUpdate_bc))
+    assert(ChannelRelayer.translateError(FeerateTooDifferent(channelId_bc, FeeratePerKw(1000 sat), FeeratePerKw(300 sat)), Some(channelUpdate_bc)) === TemporaryChannelFailure(channelUpdate_bc))
```
t-bast, line 476: Can't we test like we used to by sending a `RES_ADD_FAILED` to the `ChannelRelayer` and checking the messages sent to the `Register`? I slightly preferred that, it tested a bit more of the flow.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
```diff
@@ -265,41 +271,87 @@ class PaymentLifecycleSpec extends BaseRouterSpec {
...
+    routerForwarder.expectMsg(defaultRouteRequest(nodeParams.nodeId, d, cfg))
+    routerForwarder.forward(routerFixture.router)
+    awaitCond(paymentFSM.stateName == WAITING_FOR_PAYMENT_COMPLETE)
+    val WaitingForComplete(_, cmd1, Nil, _, _, _) = paymentFSM.stateData
+
+    register.expectMsg(ForwardShortId(paymentFSM, channelId_ab, cmd1))
+    sender.send(paymentFSM, addCompleted(HtlcResult.OnChainFail(HtlcsTimedoutDownstream(randomBytes32, Set.empty))))
+
+    // this error is fatal
+    routerForwarder.expectNoMessage()
```
t-bast, line 311: nit: we need to use `expectNoMsg(100 millis)` if we want this to build on the android branch...

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
```diff
@@ -265,41 +271,87 @@ class PaymentLifecycleSpec extends BaseRouterSpec {
...
+    sender.expectMsgType[PaymentFailed]
+  }
+
+  test("payment failed (disconnected before signing the first htlc)") { routerFixture =>
```
t-bast, line 315: Cool stuff!

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -419,6 +421,45 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+    sender.send(relayer, Relayer.GetChildActors(sender.ref))
+    val postRestartHtlcCleaner = sender.expectMsgType[Relayer.ChildActors].postRestartCleaner
+
+    // This downstream HTLC has two upstream HTLCs.
+    sender.send(relayer, buildForwardFulfill(testCase.downstream, testCase.origin, preimage1))
+    register.expectMsg(Register.Forward(postRestartHtlcCleaner, testCase.origin.originChannelId, CMD_FULFILL_HTLC(testCase.origin.originHtlcId, preimage1, commit = true)))
+
+    sender.send(relayer, buildForwardFulfill(testCase.downstream, testCase.origin, preimage1))
+    register.expectNoMsg(100 millis) // the payment has already been fulfilled upstream
+    eventListener.expectNoMsg(100 millis)
```
t-bast, line 459: We should expect a `ChannelPaymentRelayed` (once implemented of course)

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -419,6 +421,45 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+
```
t-bast, line 460: nit: remove empty line

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -419,6 +421,45 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+    val testCase = setupChannelRelayedPayments(nodeParams)
+    val relayer = f.createRelayer()
+    register.expectNoMsg(100 millis)
+
+    // we need a reference to the post-htlc-restart child actor
+    sender.send(relayer, Relayer.GetChildActors(sender.ref))
+    val postRestartHtlcCleaner = sender.expectMsgType[Relayer.ChildActors].postRestartCleaner
+
+    // This downstream HTLC has two upstream HTLCs.
```
t-bast, line 453: This doesn't really make sense for channel-relays, it should always be 1-to-1, right? I think you can simply remove the comment as the code has a single upstream htlc IIUC

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -419,6 +421,45 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+    register.expectNoMsg(100 millis)
+
+    // we need a reference to the post-htlc-restart child actor
+    sender.send(relayer, Relayer.GetChildActors(sender.ref))
+    val postRestartHtlcCleaner = sender.expectMsgType[Relayer.ChildActors].postRestartCleaner
+
+    sender.send(relayer, buildForwardFail(testCase.downstream, testCase.origin))
+    register.expectMsgType[Register.Forward[CMD_FAIL_HTLC]]
+
+    register.expectNoMsg(100 millis) // the payment has already been fulfilled upstream
```
t-bast, line 438: nit: comment doesn't seem to be accurate? But maybe it would be interesting to re-send `buildForwardFail(testCase.downstream, testCase.origin)` a second time to the `relayer` and assert it's a noop?

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -296,32 +297,34 @@ class IntegrationSpec extends TestKitBaseClass with BitcoindService with AnyFunS
...
+      sender.expectMsgType[RES_GETSTATE[_]].state == WAIT_FOR_FUNDING_LOCKED
     })
 
     generateBlocks(bitcoincli, 6)
 
     // after 8 blocks the fundee is still waiting for more confirmations
-    fundee.register ! Register.Forward(sender.ref, channelId, CMD_GETSTATE)
-    assert(sender.expectMsgType[State] == WAIT_FOR_FUNDING_CONFIRMED)
+    //fundee.register ! Register.Forward(sender.ref, channelId, CMD_GETSTATE)
+    new EclairImpl(fundee).channelInfo(Left(channelId))(20 seconds).pipeTo(sender.ref)
```
t-bast, line 312: nit: not sure this change is really useful to do in this PR, maybe later? Even though I'm more in favor of improving the tests of `EclairImpl` independently of the `IntegrationSpec`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -237,9 +239,13 @@ object OutgoingPacket {
    *
    * @return the command and the onion shared secrets (used to decrypt the error in case of payment failure)
    */
-  def buildCommand(upstream: Upstream, paymentHash: ByteVector32, hops: Seq[ChannelHop], finalPayload: Onion.FinalPayload): (CMD_ADD_HTLC, Seq[(ByteVector32, PublicKey)]) = {
+  def buildCommand(replyTo: ActorRef, upstream: Upstream, paymentHash: ByteVector32, hops: Seq[ChannelHop], finalPayload: Onion.FinalPayload): (CMD_ADD_HTLC, Seq[(ByteVector32, PublicKey)]) = {
     val (firstAmount, firstExpiry, onion) = buildPacket(Sphinx.PaymentPacket)(paymentHash, hops, finalPayload)
-    CMD_ADD_HTLC(firstAmount, paymentHash, firstExpiry, onion.packet, upstream, commit = true) -> onion.sharedSecrets
+    val origin = upstream match {
```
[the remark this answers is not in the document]
pm47: I went with something a bit different with b2864ab01.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -110,11 +114,11 @@ object ChannelRelayer {
    *         - a CMD_FAIL_HTLC to be sent back upstream
    *         - a CMD_ADD_HTLC to propagate downstream
    */
-  def handleRelay(relayPacket: IncomingPacket.ChannelRelayPacket, channelUpdates: ChannelUpdates, node2channels: NodeChannels, previousFailures: Seq[AddHtlcFailed], chainHash: ByteVector32)(implicit log: LoggingAdapter): RelayResult = {
+  def handleRelay(replyTo: ActorRef, relayPacket: IncomingPacket.ChannelRelayPacket, channelUpdates: ChannelUpdates, node2channels: NodeChannels, previousFailures: Seq[RES_ADD_FAILED[ChannelException]], chainHash: ByteVector32)(implicit log: LoggingAdapter): RelayResult = {
```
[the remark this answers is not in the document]
pm47: Done with 67193f7af.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -137,6 +116,7 @@ class Relayer(nodeParams: NodeParams, router: ActorRef, register: ActorRef, paym
           } else {
             nodeRelayer forward r
           }
+
```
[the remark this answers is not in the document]
pm47: Done with 67193f7af.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
pm47: Done with 6be5ba134.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
pm47: Thanks for the snippet! Done with 6be5ba134.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
pm47: Done with 6be5ba134.

eclair-core/src/test/scala/fr/acinq/eclair/payment/ChannelSelectionSpec.scala
```diff
@@ -41,34 +42,36 @@ class ChannelSelectionSpec extends AnyFunSuite {
     Announcements.makeChannelUpdate(Block.RegtestGenesisBlock.hash, randomKey, randomKey.publicKey, shortChannelId, cltvExpiryDelta, htlcMinimumMsat, feeBaseMsat, feeProportionalMillionths, htlcMaximumMsat, enable)
 
   test("convert to CMD_FAIL_HTLC/CMD_ADD_HTLC") {
-    val onionPayload = RelayLegacyPayload(ShortChannelId(12345), 998900 msat, CltvExpiry(60))
+    val onionPayload = RelayLegacyPayload(ShortChannelId(12345), 998_900 msat, CltvExpiry(60))
```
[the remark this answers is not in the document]
pm47: Damn! forgot about that.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
pm47: Done with 6be5ba134.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
pm47: Done with 6be5ba134.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
pm47: Done with 6be5ba134.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
pm47: Done with 6be5ba134.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -579,9 +620,39 @@ object PostRestartHtlcCleanerSpec {
...
+
+  def setupChannelRelayedPayments(nodeParams: NodeParams): ChannelRelayedPaymentTest = {
+    // Upstream HTLCs.
+    val htlc_ab_1 = Seq(
+      buildHtlcIn(0, channelId_ab_1, paymentHash1)
+    )
+
+    val origin_1 = Origin.ChannelRelayedCold(htlc_ab_1.head.add.channelId, htlc_ab_1.head.add.id, htlc_ab_1.head.add.amountMsat, htlc_ab_1.head.add.amountMsat - 100.msat)
+
+    //val notRelayed = Set((1L, channelId_bc_1), (0L, channelId_bc_2), (3L, channelId_bc_3), (5L, channelId_bc_3), (7L, channelId_bc_4))
```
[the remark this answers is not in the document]
pm47: Done with 6be5ba134.

eclair-core/src/test/scala/fr/acinq/eclair/payment/RelayerSpec.scala
pm47: It's not that easy, because I removed this weird piece of code:
https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala#L70-L75

It allowed us to skip the retry through the relayer, which requires a valid onion. I could fix it, but with the switch to typed I'll have to rewrite the test anyway...

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala
```diff
@@ -69,7 +69,7 @@ class OfflineStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
     import f._
     val sender = TestProbe()
 
-    sender.send(alice, CMD_ADD_HTLC(1000000 msat, ByteVector32.Zeroes, CltvExpiryDelta(144).toCltvExpiry(currentBlockHeight), TestConstants.emptyOnionPacket, Upstream.Local(UUID.randomUUID())))
+    sender.send(alice, CMD_ADD_HTLC(ActorRef.noSender, 1000000 msat, ByteVector32.Zeroes, CltvExpiryDelta(144).toCltvExpiry(currentBlockHeight), TestConstants.emptyOnionPacket, localOrigin(sender.ref)))
```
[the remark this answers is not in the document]
pm47: Done with f24cc1002.

eclair-core/src/test/scala/fr/acinq/eclair/payment/ChannelSelectionSpec.scala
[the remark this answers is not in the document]
pm47: Done with f24cc1002.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
```diff
@@ -243,25 +242,41 @@ class ChannelCodecsSpec extends AnyFunSuite {
...
-    val trampolineRelayed = TrampolineRelayed((randomBytes32, 1L) :: (randomBytes32, 1L) :: (randomBytes32, 2L) :: Nil, None)
-    assert(originCodec.decodeValue(originCodec.encode(trampolineRelayed).require).require === trampolineRelayed)
+    val replyTo = TestProbe("replyTo")(ActorSystem("system")).ref
+
+    val localHot = Origin.LocalHot(replyTo, UUID.randomUUID())
+    val localCold = Origin.LocalCold(localHot.id)
+    assert(originCodec.decodeValue(originCodec.encode(localHot).require).require === localCold)
+    assert(originCodec.decodeValue(originCodec.encode(localCold).require).require === localCold)
+
+    val add = UpdateAddHtlc(randomBytes32, 4324, 11_000_000 msat, randomBytes32, CltvExpiry(400_000), TestConstants.emptyOnionPacket)
```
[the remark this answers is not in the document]
pm47: Done with f24cc1002.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
```diff
@@ -18,24 +18,23 @@ package fr.acinq.eclair.payment
...
+import fr.acinq.eclair.channel._
 import fr.acinq.eclair.crypto.Sphinx
 import fr.acinq.eclair.db.{OutgoingPayment, OutgoingPaymentStatus, PaymentType}
 import fr.acinq.eclair.io.Peer.PeerRoutingMessage
 import fr.acinq.eclair.payment.PaymentRequest.ExtraHop
 import fr.acinq.eclair.payment.PaymentSent.PartialPayment
-import fr.acinq.eclair.payment.relay.Origin.Local
-import fr.acinq.eclair.payment.relay.{Origin, Relayer}
-import fr.acinq.eclair.payment.send.PaymentInitiator.{SendPaymentConfig, SendPaymentRequest}
+import fr.acinq.eclair.payment.relay.Relayer
```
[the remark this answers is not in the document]
pm47: Done with f24cc1002.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
pm47: Done with f24cc1002.

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
pm47: Done with f24cc1002.

eclair-core/src/test/scala/fr/acinq/eclair/payment/RelayerSpec.scala
t-bast: Gotcha, no worries

t-bast (approved):
LGTM :rocket:

PR #1523, opened 2020-09-01
Trampoline relay exhausted liquidity

t-bast:
We were previously returning TemporaryNodeFailure for trampoline payments to neighbour nodes with depleted liquidity. This prevented us from finding alternative, indirect routes.

In such cases, we now roughly estimate whether the fee is "big enough" to allow finding alternative routes; if not we ask the sender to raise the fee before telling them we're lacking liquidity.

A more costly alternative that we may implement in the future would be to run the path-finding to find a route without bounding the fee, and send that information back to the sender.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
@@ -21,7 +21,6 @@ import java.util.UUID
 import akka.actor.{Actor, ActorRef, DiagnosticActorLogging, PoisonPill, Props}
 import akka.event.Logging.MDC
 import fr.acinq.bitcoin.ByteVector32
-import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.channel.{CMD_FAIL_HTLC, CMD_FULFILL_HTLC, Upstream}
 import fr.acinq.eclair.db.PendingRelayDb
 import fr.acinq.eclair.payment.Monitoring.{Metrics, Tags}
@@ -133,7 +132,7 @@ class NodeRelayer(nodeParams: NodeParams, router: ActorRef, register: ActorRef) 
     case PaymentFailed(id, paymentHash, failures, _) =>
       log.debug("trampoline payment failed downstream (id={})", id)
       pendingOutgoing.get(paymentHash).foreach(p => if (!p.fulfilledUpstream) {
-        rejectPayment(p.upstream, translateError(failures, p.nextPayload.outgoingNodeId))
+        rejectPayment(p.upstream, translateError(nodeParams, failures, p))
       })
       context become main(pendingIncoming, pendingOutgoing - paymentHash)
   }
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -257,15 +256,25 @@ object NodeRelayer {
    * This helper method translates relaying errors (returned by the downstream nodes) to a BOLT 4 standard error that we
    * should return upstream.
    */
-  private def translateError(failures: Seq[PaymentFailure], outgoingNodeId: PublicKey): Option[FailureMessage] = {
+  private def translateError(nodeParams: NodeParams, failures: Seq[PaymentFailure], p: PendingResult): Option[FailureMessage] = {
```
pm47, line 259: One more occurence (see https://github.com/ACINQ/eclair/pull/1361#issue-396455215) in favor of traits that would be implemented by `NodeParams`:

```scala

sealed trait RelayFeeConf {
  def feeBase: MilliSatoshi
  def feeProportionalMillionth: Int
}

case class NodeParams extends RelayFeeConf {
...
}
```

```suggestion
  private def translateError(relayFeeConf: RelayFeeConf, failures: Seq[PaymentFailure], p: PendingResult): Option[FailureMessage] = {
```

Probably not worth it here since it's kind of a hack, but still. It's annoying to have to carry the whole `NodeParams` just for two parameters.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -257,15 +256,25 @@ object NodeRelayer {
...
     failures match {
       case Nil => None
-      case LocalFailure(_, BalanceTooLow) :: Nil => Some(TemporaryNodeFailure) // we don't have enough outgoing liquidity at the moment
+      case LocalFailure(_, BalanceTooLow) :: Nil =>
+        // We have direct channels to the target node, but not enough outgoing liquidity to use those channels.
+        // We check the total routing fee proposed by the sender:
+        //  - if it's low, we tell them to retry with higher fees to find alternative, indirect routes
+        //  - if it's high, we tell them that we don't have enough outgoing liquidity at the moment
+        val routingFeeOk = p.upstream.amountIn - p.nextPayload.amountToForward >= nodeFee(nodeParams.feeBase, nodeParams.feeProportionalMillionth, p.nextPayload.amountToForward) * 5
+        if (routingFeeOk) {
```
pm47, line 269:
```suggestion
        val routingFeeHigh = p.upstream.amountIn - p.nextPayload.amountToForward >= nodeFee(nodeParams.feeBase, nodeParams.feeProportionalMillionth, p.nextPayload.amountToForward) * 5
        if (routingFeeHigh) {
```

pm47: Nit²: you have used a different code style for the `case _ if routeNotFound => Some(TrampolineFeeInsufficient) // if we couldn't find routes, it's likely that the fee/cltv was insufficient`.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1523/commits/d0de62e8c1e153beaf06f2415b8bcbfef8b43685

pm47 (approved):

PR #1525, opened 2020-09-04
Migrate relayers to akka typed

pm47:


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -0,0 +1,347 @@
...
+                         paymentHash: ByteVector32,
+                         context: ActorContext[NodeRelay.Command],
+                         adapters: typed.ActorRef[Any],
+                         fsmFactory: FsmFactory) {
+
+  import NodeRelay._
+
+
+  def apply(): Behavior[Command] =
+    Behaviors.receiveMessagePartial {
```
t-bast, line 164: Why do you need the `Partial`? Can't we be stricter and explicitly state that in that initial state, we only receive `WrappedNodeRelayPacket` messages?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -0,0 +1,347 @@
...
+      // We make sure we receive all payment parts before forwarding to the next trampoline node.
+      case WrappedNodeRelayPacket(IncomingPacket.NodeRelayPacket(add, outer, inner, next)) => outer.paymentSecret match {
+        case None =>
+          // TODO: @pm: maybe those checks should be done later in the flow (by the mpp FSM?)
```
t-bast, line 168: Maybe, it's worth exploring (in a separate PR, it's already quite hard to diff because of the file change!).
What's a bit tricky is that you want to reject HTLCs that don't have the right payment secret *without* impacting the state of the MPP FSM; you don't want to reject the other HTLCs (that have the right secret) because an attacker inserted a malicious HTLC that doesn't have it.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -0,0 +1,347 @@
...
+        val payFSM = fsmFactory.spawnOutgoingPayFSM(context, nodeParams, router, register, paymentCfg, multiPart = true)
+        val paymentSecret = randomBytes32 // we generate a new secret to protect against probing attacks
+        val payment = SendMultiPartPayment(adapters.toClassic, paymentSecret, payloadOut.outgoingNodeId, payloadOut.amountToForward, payloadOut.outgoingCltv, nodeParams.maxPaymentAttempts, routeParams = Some(routeParams), additionalTlvs = Seq(OnionTlv.TrampolineOnion(packetOut)))
+        payFSM ! payment
+        payFSM
+    }
+    payFSM
+  }
+
+  private def rejectExtraHtlcPartialFunction: PartialFunction[Command, Behavior[Command]] = {
```
t-bast, line 304: This is used in only one place, is it really worth putting in a dedicated function (I think it makes it a bit harder to read than if all the `case`s were in the same place)?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -0,0 +1,347 @@
...
+          rejectHtlc(add.id, add.channelId, add.amountMsat)
+          Behaviors.stopped
+        case Some(secret) =>
+          import akka.actor.typed.scaladsl.adapter._
+          context.log.debug("creating a payment FSM")
+          val mppFsm = context.actorOf(MultiPartPaymentFSM.props(nodeParams, add.paymentHash, outer.totalAmount, adapters.toClassic))
+          context.log.debug("forwarding incoming htlc to the payment FSM")
+          mppFsm ! MultiPartPaymentFSM.HtlcPart(outer.totalAmount, add)
+          receiving(Queue(add), secret, inner, next, mppFsm)
+        case _ => Behaviors.unhandled
```
t-bast, line 179: This case is unnecessary, everything is already handled

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -0,0 +1,347 @@
...
+          Behaviors.same
+        case Some(incomingSecret) if incomingSecret == secret =>
+          context.log.debug("forwarding incoming htlc to the payment FSM")
+          handler ! MultiPartPaymentFSM.HtlcPart(outer.totalAmount, add)
+          receiving(htlcs :+ add, secret, nextPayload, nextPacket, handler)
+      }
+      case WrappedMultiPartPaymentFailed(MultiPartPaymentFSM.MultiPartPaymentFailed(_, failure, parts)) =>
+        context.log.warn("could not relay payment (paidAmount={} failure={})", parts.map(_.amount).sum, failure)
+        Metrics.recordPaymentRelayFailed(failure.getClass.getSimpleName, Tags.RelayType.Trampoline)
+        handler ! PoisonPill
```
t-bast, line 212: We need to do some experimentation, but now that you spawn a `NodeRelay` per-trampoline payment, we shouldn't need to explicitly `PoisonPill` the `MultiPartPaymentFSM`, it should be stopped automatically when the `NodeRelay` stops (which may be a bit later than before when we actually relay downstream, but shouldn't be an issue).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+  /**
+   * Once the payment is forwarded, we're waiting for fail/fulfill responses from downstream nodes.
+   *
+   * @param upstream          complete HTLC set received.
+   * @param nextPayload       relay instructions.
+   * @param fulfilledUpstream true if we already fulfilled the payment upstream.
+   */
+  private def sending(payFSM: ActorRef, upstream: Upstream.Trampoline, nextPayload: Onion.NodeRelayPayload, fulfilledUpstream: Boolean): Behavior[Command] =
```
t-bast, line 238: I don't think you need the `payFSM` here, do you?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -0,0 +1,347 @@
...
+            // We want to fulfill upstream as soon as we receive the preimage (even if not all HTLCs have fulfilled downstream).
+            context.log.debug("trampoline payment successfully relayed")
+            fulfillPayment(upstream, paymentPreimage)
+            sending(payFSM, upstream, nextPayload, fulfilledUpstream = true)
+          } else {
+            // we don't want to fulfill multiple times
+            Behaviors.same
+          }
+        case WrappedPaymentSent(paymentSent) =>
+          // We may have already fulfilled upstream, but we can now emit an accurate relayed event and clean-up resources.
```
t-bast, line 253: nit: this comment should go in the `success` handler

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -0,0 +1,347 @@
...
+        case WrappedPaymentFailed(PaymentFailed(id, _, failures, _)) =>
+          context.log.debug(s"trampoline payment failed downstream (id={})", id)
+          if (!fulfilledUpstream) {
+            rejectPayment(upstream, translateError(nodeParams, failures, upstream, nextPayload))
+            Behaviors.stopped
+          } else {
+            // Hmm, this is weird
+            // TODO @t-bast : is this ok?
+            Behaviors.unhandled
+          }
```
t-bast, line 266: I think this should simply be:
```suggestion
          if (!fulfilledUpstream) {
            rejectPayment(upstream, translateError(nodeParams, failures, upstream, nextPayload))
          }
          Behaviors.stopped
```

t-bast (commented):
Sorry for the lengthy review, Akka Typed is quite new to me so I'm trying to match bits of the Akka documentation to what you did :)
I haven't reviewed the tests yet, but already have quite a lot of questions and feedback!

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -161,7 +161,7 @@ sealed trait HasHtlcId { this: Command => def id: Long }
 final case class CMD_FULFILL_HTLC(id: Long, r: ByteVector32, commit: Boolean = false) extends Command with HasHtlcId
 final case class CMD_FAIL_HTLC(id: Long, reason: Either[ByteVector, FailureMessage], commit: Boolean = false) extends Command with HasHtlcId
 final case class CMD_FAIL_MALFORMED_HTLC(id: Long, onionHash: ByteVector32, failureCode: Int, commit: Boolean = false) extends Command with HasHtlcId
-final case class CMD_ADD_HTLC(replyTo: ActorRef, amount: MilliSatoshi, paymentHash: ByteVector32, cltvExpiry: CltvExpiry, onion: OnionRoutingPacket, origin: Origin.Hot, commit: Boolean = false, previousFailures: Seq[RES_ADD_FAILED[ChannelException]] = Seq.empty) extends Command with HasReplyTo
+final case class CMD_ADD_HTLC(replyTo: ActorRef, amount: MilliSatoshi, paymentHash: ByteVector32, cltvExpiry: CltvExpiry, onion: OnionRoutingPacket, origin: Origin.Hot, commit: Boolean = false) extends Command with HasReplyTo
```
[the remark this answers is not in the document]
pm47: Channel relay was always a stateful operation but was maskeraded as stateless, now it is explicit.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -52,69 +52,25 @@ class Relayer(nodeParams: NodeParams, router: ActorRef, register: ActorRef, paym
   // we pass these to helpers classes so that they have the logging context
   implicit def implicitLog: LoggingAdapter = log
 
-  context.system.eventStream.subscribe(self, classOf[LocalChannelUpdate])
-  context.system.eventStream.subscribe(self, classOf[LocalChannelDown])
-  context.system.eventStream.subscribe(self, classOf[AvailableBalanceChanged])
-  context.system.eventStream.subscribe(self, classOf[ShortChannelIdAssigned])
```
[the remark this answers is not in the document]
pm47: I like the symmetry between `NodeRelayer` and `ChannelRelayer`. Why would we remove the latter and keep the former? If we do, then we break assumptions on the type of child actors.

eclair-core/src/test/resources/logback-test.xml
```diff
@@ -52,7 +52,7 @@
     <root level="INFO">
         <!--<appender-ref ref="FILE"/>-->
         <!--<appender-ref ref="CONSOLEWARN"/>-->
-        <!--<appender-ref ref="CONSOLE"/>-->
+        <!--appender-ref ref="CONSOLE"/-->
```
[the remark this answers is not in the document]
pm47: Don't know why I did that, doesn't work in my intellij either... Fixed with 5467945e8.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -16,220 +16,107 @@
...
-            Some(requestedShortChannelId)
-          case None =>
-            // no channel seem to work for this payment and we have already tried the requested channel id: we give up
-            None
+  def apply(
+             nodeParams: NodeParams,
+             register: ActorRef,
+             channelUpdates: ChannelUpdates = Map.empty,
+             node2channels: NodeChannels = mutable.MultiDict.empty[PublicKey, ShortChannelId]
+           ): Behavior[Command] =
```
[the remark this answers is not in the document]
pm47: Fixed with 5467945e8.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -158,7 +116,7 @@ object Relayer extends Logging {
   type NodeChannels = mutable.MultiDict[PublicKey, ShortChannelId]
```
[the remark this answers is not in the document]
pm47: Fixed with 5467945e8.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
```diff
@@ -0,0 +1,267 @@
...
+ * see https://doc.akka.io/docs/akka/current/typed/style-guide.html#passing-around-too-many-parameters
+ */
+class ChannelRelay private(
+                            nodeParams: NodeParams,
+                            register: ActorRef,
+                            channelUpdates: ChannelUpdates,
+                            node2channels: NodeChannels,
+                            r: IncomingPacket.ChannelRelayPacket,
+                            context: ActorContext[ChannelRelay.Command],
+                            adapters: typed.ActorRef[Any]) {
```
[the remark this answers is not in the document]
pm47: I remember feeling confused about this indentation myself. Fixed with 5467945e8.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
```diff
@@ -0,0 +1,267 @@
...
+        Behaviors.stopped
+
+      case WrappedAddResponse(RES_ADD_SETTLED(o: Origin.ChannelRelayedHot, _, fail: HtlcResult.Fail)) =>
+        Metrics.recordPaymentRelayFailed(Tags.FailureType.Remote, Tags.RelayType.Channel)
+        val cmd = fail match {
+          case f: HtlcResult.RemoteFail => CMD_FAIL_HTLC(o.originHtlcId, Left(f.fail.reason), commit = true)
+          case f: HtlcResult.RemoteFailMalformed => CMD_FAIL_MALFORMED_HTLC(o.originHtlcId, f.fail.onionHash, f.fail.failureCode, commit = true)
+          case _: HtlcResult.OnChainFail => CMD_FAIL_HTLC(o.originHtlcId, Right(PermanentChannelFailure), commit = true)
+          case f: HtlcResult.Disconnected => CMD_FAIL_HTLC(o.originHtlcId, Right(TemporaryChannelFailure(f.channelUpdate)), commit = true)
+        }
```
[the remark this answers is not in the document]
pm47: Nice catch. Fixed with 5467945e8.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -18,264 +18,55 @@ package fr.acinq.eclair.payment.relay
...
-import fr.acinq.eclair.router.Router.RouteParams
-import fr.acinq.eclair.router.{BalanceTooLow, RouteCalculation, RouteNotFound}
-import fr.acinq.eclair.wire._
-import fr.acinq.eclair.{CltvExpiry, Features, Logs, MilliSatoshi, NodeParams, nodeFee, randomBytes32}
-
-import scala.collection.immutable.Queue
+import akka.actor.ActorRef
+import akka.actor.typed.Behavior
+import akka.actor.typed.scaladsl.Behaviors
+import fr.acinq.eclair.payment.{relay, _}
```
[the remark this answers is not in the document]
pm47: Fixed with 5467945e8.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: It's actually used in feature branches.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: Fixed with 5467945e8.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: We do?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: Fixed with 5467945e8.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: Is there any benefit in having a warning log when that happens?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -16,220 +16,107 @@
...
+    Behaviors.setup { context =>
+      context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[LocalChannelUpdate](WrappedLocalChannelUpdate))
+      context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[LocalChannelDown](WrappedLocalChannelDown))
+      context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[AvailableBalanceChanged](WrappedAvailableBalanceChanged))
+      context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[ShortChannelIdAssigned](WrappedShortChannelIdAssigned))
+      context.messageAdapter[IncomingPacket.ChannelRelayPacket](WrappedChannelRelayPacket)
+      Behaviors.withMdc(Logs.mdc(category_opt = Some(Logs.LogCategory.PAYMENT)), mdc) {
+        Behaviors.receiveMessagePartial {
```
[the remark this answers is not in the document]
pm47: To be honest, I used `receiveMessage`/`receiveMessagePartial` interchangeably without noticing the difference. `receiveMessage` seems stricter but now I'm getting those warnings:
```
Exhaustivity analysis reached max recursion depth, not all missing cases are reported.
(Please try with scalac -Ypatmat-exhaust-depth 40 or -Ypatmat-exhaust-depth off.)
    Behaviors.receiveMessage {
```

But I can't get this configuration to be accepted by `scala-maven-plugin` right away. Needs more digging.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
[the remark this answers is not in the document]
t-bast: Allright, the symmetry makes it indeed easy to navigate

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
t-bast: It's not necessary here, there is already a warning log line in the `MultiPartPaymentLifecycle` in that case

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -18,264 +18,55 @@ package fr.acinq.eclair.payment.relay
...
-        // Otherwise, we try to find a downstream error that we could decrypt.
-        val outgoingNodeFailure = failures.collectFirst { case RemoteFailure(_, e) if e.originNode == p.nextPayload.outgoingNodeId => e.failureMessage }
-        val otherNodeFailure = failures.collectFirst { case RemoteFailure(_, e) => e.failureMessage }
-        val failure = outgoingNodeFailure.getOrElse(otherNodeFailure.getOrElse(TemporaryNodeFailure))
-        Some(failure)
+  def apply(nodeParams: NodeParams, router: ActorRef, register: ActorRef): Behavior[Command] =
+    Behaviors.setup { context =>
+      context.messageAdapter[IncomingPacket.NodeRelayPacket](WrappedNodeRelayPacket)
+      Behaviors.withMdc(Logs.mdc(category_opt = Some(Logs.LogCategory.PAYMENT)), mdc) {
+        Behaviors.receiveMessagePartial {
```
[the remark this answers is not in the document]
pm47: See https://github.com/ACINQ/eclair/pull/1525#discussion_r484837338.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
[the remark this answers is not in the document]
pm47: Ok, found out what the difference is.

With `receiveMessage`, any unhandled message will cause the actor to fail (that is, to stop), whereas with `receiveMessagePartial` it will just print a deadletter log. That's why the former generates a compilation warning. Obviously we should use `receiveMessage` as much as possible.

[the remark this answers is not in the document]
pm47: Actually, if we model our typed actors as FSM with behaviors as states, we have no other choise than using `receiveMessagePartial` almost everywhere, since each state typically handles a subset of all the channel commands. We could use `receiveMessage` and manually ignore all unmatched message, but that's exactly what `receiveMessagePartial` does.

Side note : there are limits to what the compiler can detect apparently. E.g. in the `sending` method below, not all messages are handled but there are no warning (probably due to the `orElse` construct).

https://github.com/ACINQ/eclair/blob/typed-relayers/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala#L236-L265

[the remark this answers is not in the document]
t-bast: In that case, do you think we could have a top-level `sealed trait Command` and below that, other `sealed trait SomeStateCommand extends Command` for each of the FSM states, and in each state declare ourselves as `ActorRef[SomeStateCommand]`? Or is it overkill?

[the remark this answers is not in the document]
pm47: We could, but then we'll have to use `Behaviors.[SubCommand].unsafeUpcast[Command]` somewhere. That may be better because more explicit, I'm not sure.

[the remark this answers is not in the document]
t-bast: Would it be better with the less functional way of implementing the actor?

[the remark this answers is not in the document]
pm47: It could even get messier if we have three states: 

```scala

sealed trait Command
sealed trait State1Command extends Command
sealed trait State2Command extends Command

def apply(): Behavior[Command] = state1.unsafeUpcast[Command]

def state1(): Behavior[State1Command] = Behaviors.receiveMessage {
  case s1: State1Command => state2.unsafeUpcast[State1Command]
}

def state2(): Behavior[State2Command] = Behaviors.receiveMessage {
  case s2: State2Command => Behaviors.stopped
}
```

The `state2.unsafeUpcast[State1Command].unsafeUpcast[Command]` is... let's say not very elegant.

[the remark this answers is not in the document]
pm47: > Would it be better with the less functional way of implementing the actor?

I don't think so, this is a logical problem, not a technical one. We could also do:


```scala

sealed trait Command
sealed trait State1Command extends Command
sealed trait State2Command extends Command

def apply(): Behavior[Command] = state1.unsafeUpcast[Command]

def state1(): Behavior[Command] = Behaviors.receiveMessage[State1Command] {
  ...
}.unsafeUpcast[State1Command]

def state2(): Behavior[Command] = Behaviors.receiveMessage[State2Command] {
  ...
}.unsafeUpcast[State2Command]
```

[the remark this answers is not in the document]
t-bast: It's weird, what we're doing really sounds like a very common usage of akka typed, it's surprising that there isn't a more elegant solution :confused:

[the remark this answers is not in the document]
pm47: In the [akka doc](https://doc.akka.io/docs/akka/current/typed/fsm.html#behaviors-as-finite-state-machines) they use a combination of `receiveMessage` + `case _ => Behaviors.unhandled` for one state, and `receivePartial` for another one. Looks like there isn't a real good way of doing what we want to do.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: Fixed with 3b729a6.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
```diff
@@ -0,0 +1,267 @@
...
+import fr.acinq.eclair.router.Announcements
+import fr.acinq.eclair.wire._
+import fr.acinq.eclair.{Logs, MilliSatoshi, NodeParams, ShortChannelId, channel, nodeFee}
+
+object ChannelRelay {
+
+  // @formatter:off
+  sealed trait Command
+  case class WrappedForwardShortIdFailure(failure: Register.ForwardShortIdFailure[CMD_ADD_HTLC]) extends Command
+  case class WrappedForwardFailure(failure: Register.ForwardFailure[CMD_FAIL_HTLC]) extends Command
```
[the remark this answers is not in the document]
pm47: Fixed with 0d98217ae.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
```diff
@@ -0,0 +1,267 @@
...
+  case class RelayFailure(cmdFail: CMD_FAIL_HTLC) extends RelayResult
+  case class RelaySuccess(shortChannelId: ShortChannelId, cmdAdd: CMD_ADD_HTLC) extends RelayResult
+  case class RelayPayToOpen(amount: MilliSatoshi, add: UpdateAddHtlc) extends RelayResult
+  // @formatter:on
+
+  def apply(nodeParams: NodeParams, register: ActorRef, channelUpdates: ChannelUpdates, node2channels: NodeChannels, relayId: UUID, r: IncomingPacket.ChannelRelayPacket): Behavior[Command] =
+    Behaviors.setup { context =>
+      Behaviors.withMdc(Logs.mdc(
+        category_opt = Some(Logs.LogCategory.PAYMENT),
+        parentPaymentId_opt = Some(relayId),
```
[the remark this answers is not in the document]
pm47: Done with ecb75d167.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -18,264 +18,55 @@ package fr.acinq.eclair.payment.relay
...
+          case WrappedNodeRelayPacket(nodeRelayPacket) =>
+            import nodeRelayPacket.add.paymentHash
+            val handler = context.child(paymentHash.toString) match {
+              case Some(handler) =>
+                //@TODO: pm: we could also maintain a list of children
```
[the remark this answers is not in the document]
pm47: Fixed with bd1f11939.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
```diff
@@ -0,0 +1,267 @@
...
+        paymentHash_opt = Some(r.add.paymentHash),
+        channelId_opt = Some(r.add.channelId))) {
+        val adapters = {
+          context.messageAdapter[Register.ForwardShortIdFailure[CMD_ADD_HTLC]](WrappedForwardShortIdFailure)
+          context.messageAdapter[CommandResponse[CMD_ADD_HTLC]](WrappedAddResponse)
+        }.unsafeUpcast[Any]
```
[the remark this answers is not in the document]
pm47: This is just a way to get one single `ActorRef[Any]` in charge of all conversions (it is a different actor from `context.self`). Not particularly elegant, I've tried a different pattern with bdc167592, let me know what you think.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: Alright, done with 2935b6837.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -0,0 +1,347 @@
...
+  case class WrappedPreimageReceived(preimageReceived: PreimageReceived) extends Command
+  case class WrappedPaymentSent(paymentSent: PaymentSent) extends Command
+  case class WrappedPaymentFailed(paymentFailed: PaymentFailed) extends Command
+  // @formatter:on
+
+  def mdc: Command => Map[String, String] = {
+    case c: WrappedNodeRelayPacket => Logs.mdc(
+      channelId_opt = Some(c.nodeRelayPacket.add.channelId))
+    case _ => Map.empty
+  }
```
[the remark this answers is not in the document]
pm47: Indeed have the upstream `channelId` isn't useful, removed it in f21d51852. Added the `parentPaymentId` as part of ecb75d167.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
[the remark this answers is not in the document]
t-bast: :+1: that looks better!

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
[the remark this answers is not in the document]
t-bast: :+1: that looks better!

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
t-bast: I don't see where :thinking:? Here it seems that you just pass it through, but you never actually use it to send it messages, so you don't need to keep the `ActorRef` around, do you?

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/RelayerSpec.scala
```diff
@@ -0,0 +1,196 @@
...
+    import f._
+
+    // we use this to build a valid onion
+    val (cmd, _) = buildCommand(ActorRef.noSender, Upstream.Local(UUID.randomUUID()), paymentHash, hops, FinalLegacyPayload(finalAmount, finalExpiry))
+    // and then manually build an htlc
+    val add_ab = UpdateAddHtlc(channelId = randomBytes32, id = 123456, cmd.amount, cmd.paymentHash, cmd.cltvExpiry, cmd.onion)
+
+    relayer ! RelayForward(add_ab)
+
+    register.expectMessageType[Register.Forward[CMD_FAIL_HTLC]]
```
t-bast, line 82: Can't we emit a `LocalChannelUpdate` like we did before to have this be a `CMD_ADD_HTLC`? Or maybe since this is already tested in `ChannelRelayerSpec`, just assert that the `replyTo` of the `Register.Forward` is a `ChannelRelay` actor?

eclair-core/src/test/scala/fr/acinq/eclair/payment/ChannelSelectionSpec.scala
```diff
@@ -41,82 +31,82 @@ class ChannelSelectionSpec extends AnyFunSuite {
...
+//    // insufficient fee
+//    val relayPayload_insufficientfee = relayPayload.copy(payload = onionPayload.copy(amountToForward = 998910 msat))
+//    assert(relayOrFail(ActorRef.noSender, relayPayload_insufficientfee, Some(channelUpdate)) === RelayFailure(CMD_FAIL_HTLC(relayPayload.add.id, Right(FeeInsufficient(relayPayload_insufficientfee.add.amountMsat, channelUpdate)), commit = true)))
+//    // note that a generous fee is ok!
+//    val relayPayload_highfee = relayPayload.copy(payload = onionPayload.copy(amountToForward = 900000 msat))
+//    val originHighFee = origin.copy(amountOut = relayPayload_highfee.payload.amountToForward)
+//    assert(relayOrFail(ActorRef.noSender, relayPayload_highfee, Some(channelUpdate)) === RelaySuccess(ShortChannelId(12345), CMD_ADD_HTLC(ActorRef.noSender, relayPayload_highfee.payload.amountToForward, relayPayload_highfee.add.paymentHash, relayPayload_highfee.payload.outgoingCltv, relayPayload_highfee.nextPacket, originHighFee, commit = true)))
+//  }
+//
+//  test("channel selection") {
```
t-bast, line 67: I think we can remove the previous test (it's already covered in `ChannelRelayerSpec`), but don't we want to keep this one? Or migrate it inside `ChannelRelayerSpec`?

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/RelayerSpec.scala
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  test("fail to relay an htlc-add when the onion is malformed") { f =>
+    import f._
+
+    // we use this to build a valid onion
+    val (cmd, _) = buildCommand(ActorRef.noSender, Upstream.Local(UUID.randomUUID()), paymentHash, hops, FinalLegacyPayload(finalAmount, finalExpiry))
+    // and then manually build an htlc with an invalid onion (hmac)
+    val add_ab = UpdateAddHtlc(channelId = channelId_ab, id = 123456, cmd.amount, cmd.paymentHash, cmd.cltvExpiry, cmd.onion.copy(hmac = cmd.onion.hmac.reverse))
+    relayer ! LocalChannelUpdate(null, channelId_bc, channelUpdate_bc.shortChannelId, c, None, channelUpdate_bc, makeCommitments(channelId_bc))
```
t-bast, line 135: This line doesn't make sense anymore, remove?

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/RelayerSpec.scala
```diff
@@ -0,0 +1,196 @@
...
+    val a = PaymentPacketSpec.a
+
+    // we use this to build a valid trampoline onion inside a normal onion
+    val trampolineHops = NodeHop(a, b, channelUpdate_ab.cltvExpiryDelta, 0 msat) :: NodeHop(b, c, channelUpdate_bc.cltvExpiryDelta, fee_b) :: Nil
+    val (trampolineAmount, trampolineExpiry, trampolineOnion) = OutgoingPacket.buildPacket(Sphinx.TrampolinePacket)(paymentHash, trampolineHops, Onion.createSinglePartPayload(finalAmount, finalExpiry))
+    val (cmd, _) = buildCommand(ActorRef.noSender, Upstream.Local(UUID.randomUUID()), paymentHash, ChannelHop(a, b, channelUpdate_ab) :: Nil, Onion.createTrampolinePayload(trampolineAmount, trampolineAmount, trampolineExpiry, randomBytes32, trampolineOnion.packet))
+
+    // and then manually build an htlc
+    val add_ab = UpdateAddHtlc(channelId = channelId_ab, id = 123456, cmd.amount, cmd.paymentHash, cmd.cltvExpiry, cmd.onion)
+    relayer ! LocalChannelUpdate(null, channelId_bc, channelUpdate_bc.shortChannelId, c, None, channelUpdate_bc, makeCommitments(channelId_bc))
```
t-bast, line 159: Same, remove?

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
```diff
@@ -0,0 +1,423 @@
...
+    val channelId1 = channelIds(shortId1)
+    val payload = RelayLegacyPayload(shortId1, outgoingAmount, outgoingExpiry)
+    val r = createValidIncomingPacket(1100000 msat, CltvExpiry(400100), payload)
+    val u = createChannelUpdate(shortId1)
+    val downstream_htlc = UpdateAddHtlc(channelId1, 7, outgoingAmount, paymentHash, outgoingExpiry, emptyOnionPacket)
+
+    case class TestCase(result: HtlcResult)
+
+    val testCases = Seq(
+      TestCase(HtlcResult.RemoteFulfill(UpdateFulfillHtlc(channelId1, downstream_htlc.id, paymentPreimage)))
```
t-bast, line 309: Can't we also test the `OnChainFulfill`?

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -0,0 +1,561 @@
...
+ */
+
+class NodeRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("application")) with FixtureAnyFunSuiteLike {
+
+  import NodeRelayerSpec._
+
+  case class FixtureParam(nodeParams: NodeParams, nodeRelayer: ActorRef[NodeRelay.Command], router: TestProbe[Any], register: TestProbe[Any], mockPayFSM: TestProbe[Any], eventListener: TestProbe[PaymentEvent])
+
+  override def withFixture(test: OneArgTest): Outcome = {
+    eventually {
```
t-bast, line 60: Are you sure we should use `eventually` (same in `RelayerSpec` and `ChannelRelayerSpec`)?
IIUC it retries tests if they fail, that would hide random failures that probably need to be fixed, doesn't it?

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -0,0 +1,561 @@
...
+            fsm
+          }
+        }
+      }
+      val nodeRelay = testKit.spawn(NodeRelay(nodeParams, router.ref.toClassic, register.ref.toClassic, relayId, paymentHash, fsmFactory))
+      withFixture(test.toNoArgTest(FixtureParam(nodeParams, nodeRelay, router, register, mockPayFSM, eventListener)))
+    }
+  }
+
+  test("fail to relay when incoming multi-part payment times out") { f =>
```
t-bast, line 88: I think this should use a `mockFsm`, otherwise we're waiting for a timeout and this test takes 5s whereas it could take 100ms :)

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -0,0 +1,561 @@
...
+    import f._
+
+    // Receive a partial upstream multi-part payment.
+    incomingMultiPart.dropRight(1).foreach(incoming => nodeRelayer ! NodeRelay.WrappedNodeRelayPacket(incoming))
+    // after a while the payment times out
+    incomingMultiPart.dropRight(1).foreach(p => register.expectMessageType[Register.Forward[CMD_FAIL_HTLC]](30 seconds))
+    //    incomingMultiPart.dropRight(1).foreach(p => register.expectMessage(30 seconds, Register.Forward(nodeRelayer.toClassic, p.add.channelId, CMD_FAIL_HTLC(p.add.id, Right(PaymentTimeout), commit = true))))
```
t-bast, line 95: nit: remove?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -16,220 +16,107 @@
...
-  sealed trait RelayResult
-  case class RelayFailure(cmdFail: CMD_FAIL_HTLC) extends RelayResult
-  case class RelaySuccess(shortChannelId: ShortChannelId, cmdAdd: CMD_ADD_HTLC) extends RelayResult
+  sealed trait Command
+  case class WrappedGetOutgoingChannels(replyTo: ActorRef, getOutgoingChannels: GetOutgoingChannels) extends Command
+  case class WrappedChannelRelayPacket(channelRelayPacket: IncomingPacket.ChannelRelayPacket) extends Command
+  case class WrappedLocalChannelUpdate(localChannelUpdate: LocalChannelUpdate) extends Command
+  case class WrappedLocalChannelDown(localChannelDown: LocalChannelDown) extends Command
+  case class WrappedAvailableBalanceChanged(availableBalanceChanged: AvailableBalanceChanged) extends Command
+  case class WrappedShortChannelIdAssigned(shortChannelIdAssigned: ShortChannelIdAssigned) extends Command
```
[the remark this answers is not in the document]
t-bast: EDIT: for the commands that are requests of this actor, the naming should not be `WrappedXXX` for clarity (the first two).
For commands that are wrapped responses (the last four), we should probably make them private (or only expose them to tests) and let message adapters do the automatic wrapping.

Same for `NodeRelayer`/`NodeRelay`.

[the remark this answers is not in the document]
pm47: Done with 90ba84f61.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: Right! Fixed as part of 43c6ca216.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
pm47: Done with 9adec95c9.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
pm47: I disagree: this allows us to test the actual timeout logic, and tests are parallelized anyway.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
pm47: Fixed with 9adec95c9.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
t-bast: Ok, it's already tested in the multi-part FSM spec but it never hurts to test multiple times!

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/RelayerSpec.scala
pm47: Fixed in 626b4027f.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/RelayerSpec.scala
pm47: Fixed in 626b4027f.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/RelayerSpec.scala
pm47: Fixed in 626b4027f.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
pm47: Fixed in 626b4027f.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/RelayerSpec.scala
t-bast: The `expectMessageType` is inside a `println`, was it for your local debugging or is it expected?

pm47: Oops, fixed in 0b2e5bd7b.

eclair-core/src/test/scala/fr/acinq/eclair/payment/ChannelSelectionSpec.scala
t-bast: What shall we do about this last test (`channel selection`)?
Otherwise the PR looks good to me.

pm47: Done with 3453ba627 (and fixed a bug). I also simplified `ChannelRelay` by using the appropriate projection of `ChannelUpdates`/`NodeChannels`. I believe we could simplify further by only call the channel selection function once at the beginning to sort the channels in order of preference and try them in order instead of calling it at every round.

t-bast (approved):
:rocket: :full_moon:

t-bast (approved):

PR #1526, opened 2020-09-08
Update to Bitcoin Core 0.20.1

t-bast:
The only impactful change is that by default on regtest and testnet the `fallbackfee` (used when there is not enough historical data to correctly estimate the feerate) is now set to 0, whereas it was set to 0.0002 btc in previous versions. This change only impacts regtest and testnet, on mainnet this parameter has always been set to 0.

We set it manually in tests `bitcoin.conf` to preserve the past behavior (as recommended by bitcoind's release notes).

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ You will find detailed guides and frequently asked questions there.
 
 ### Configuring Bitcoin Core
 
-:warning: Eclair requires Bitcoin Core 0.19.1 or 0.20.1. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
+:warning: Eclair requires Bitcoin Core 0.18.1, 0.19.1 or 0.20.1. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
 
 Eclair needs a _synchronized_, _segwit-ready_, **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ You will find detailed guides and frequently asked questions there.
 
 ### Configuring Bitcoin Core
 
-:warning: Eclair requires Bitcoin Core 0.18.1 or 0.19.1. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
+:warning: Eclair requires Bitcoin Core 0.19.1 or 0.20.1. If you are upgrading an existing wallet, you need to create a new address and send all your funds to that address.
```
pm47, line 44: Any reason why we drop the 0.18.1 compatibility?

pm47 (approved):
I verified all the checksums ✔️

README.md
t-bast: Not really, we could still support it as we haven't done anything breaking. If you prefer I can change that to 0.18.1, 0.19.1 or 0.20.1.
I'm mostly waiting for release 0.21 that will add a `psbtbumpfee` RPC, which is required for automatic fee bumping of anchor output channels, but for now we don't need to deprecate support.

pm47: I would say if we don't need to, let's not break compatibility.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1526/commits/b8170cd7c5bdf04e49b10350deb9ff52e19b06a7

pm47 (approved):
I see you're not into oxford comma.

t-bast:
I'm too biased by french grammar :laughing:

PR #1527, opened 2020-09-16
Do not reply to htlc settlement commands

pm47:
The replies are always ignored currently anyway. A new trait `NoReplyTo`
has been introduced. Those commands have a particular workflow because
they are persisted in the pending relay db.


t-bast (approved):
Nice clean-up, the model feels better!

PR #1528, opened 2020-09-16
Plugin messaging

btcontract:
This is a draft of how plugin messaging between two peers could work.

It makes sense to enable plugins to communicate with each other via an already existing and authenticated socket connection rather than for example getting them to open another HTTP endpoint just for that.

The idea is as follows:

1. Once connected, a peer who wants to know about other peer's messaging-enabled plugins sends `QuerySupportedPlugins` message.
2. The other peer replies with `ReplySupportedPlugins` so interested peer knows what to expect.
3. Remote peer can then send `PluginMessage` with `pluginId` attached, a `Peer.scala` would issue an event on seeing such a message which in turn would be intercepted by plugin and further processed according to plugin's specific rules.


eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -326,6 +364,11 @@ object LightningMessageCodecs {
     .typecase(263, queryChannelRangeCodec)
     .typecase(264, replyChannelRangeCodec)
     .typecase(265, gossipTimestampFilterCodec)
+  //
+    .typecase(1487, provide(QuerySupportedPlugins))
```
btcontract, line 368: these are magic odd numbers, should they be picked on some specific principle?

Christewart:
Hi @btcontract 

We've begun hacking away at eclair to introduce some of the PTLC work (and future DLC work) we've been writing about at Suredbits for awhile now. Would these two use cases apply here?

https://github.com/bitcoin-s/eclair/pulls?q=is%3Apr+label%3Aptlc+

btcontract:
@Christewart I don't yet understand how PTLC/DLC works but if this involves communication between two peers then yes, a related plugin could define a completely independant system of messages which is convenient for your use case, no need for tlv data even.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
rorp: Yes, as per https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#lightning-message-format :

> Custom (types 32768-65535): experimental and application-specific messages

t-bast:
> 1. Once connected, a peer who wants to know about other peer's messaging-enabled plugins sends QuerySupportedPlugins message.
> 2. The other peer replies with ReplySupportedPlugins so interested peer knows what to expect.

I think this part is unnecessary, that is exactly why we have feature bits in place so I wouldn't want to add this to the base eclair layer.
I think it would be much cleaner to reserve some feature bits far in the future to each feature that you're adding.
If you use feature bits that are higher than 128, you're pretty much guaranteed that the spec will not override your choice before at least 10 years given the rate at which we add spec features :laughing: 

And from an abstraction's point of view, it's also better. A node connecting to you doesn't care that a feature works via a plugin or via the base code; it shouldn't even care what implementation you're running, as long as you support the feature they want. And ideally in the long run, custom features that users like will end up being a standard implemented by many, which is better for the ecosystem.

There is a risk that another app uses the same feature bits for its experiments, but I don't think it's an issue in practice because if you think they support your feature because they've activate the same feature bit, when you'll exchange custom messages they will fail to decode and either side will close the connection.

btcontract:
Here's what's done so far:

- Each plugin may declare an optional feature bit which must be higher than 128 and must not overlap with core used features (checks on this are done, I guess it makes sense to have plugin features always optional since plugins themselves are optional).
- Plugin features are injected into `Init` message and added to `Peer.scala`.
- `PluginMessage` now looks like `PluginMessage(featureBit: Int, data: ByteVector)` instead of `PluginMessage(pluginId: Int, data: ByteVector)`, `Peer` can still quickly check if it's supported.
- While doing this `sealed trait Feature` became just `trait Feature` to reference it in `Plugin.scala`, is this acceptable?

> At start-up, plugins inject the list of message types (the 2-byte number at the beginning of every lightning message) they're interested in

Do we really need to do it this way? Having actual plugin-specific message as `data: ByteVector` inside of `PluginMessage` gives plugin devs more freedom, as an extreme example some plugin may decide to exchange data in, say, Json format (converted to bytes).

t-bast (commented):
Great, it looks like this won't be too much added code and will add a lot of usefulness to plugins!

> While doing this sealed trait Feature became just trait Feature to reference it in Plugin.scala, is this acceptable?

Yes I think it's totally acceptable, the `Feature` trait is not something we want to exhaustively pattern-match on so I'd go with it.

> > At start-up, plugins inject the list of message types (the 2-byte number at the beginning of every lightning message) they're interested in

> Do we really need to do it this way? Having actual plugin-specific message as data: ByteVector inside of PluginMessage gives plugin devs more freedom, as an extreme example some plugin may decide to exchange data in, say, Json format (converted to bytes).

Yes I think my suggestion is more in-line with how lightning messages are used and will evolve (and fits what we're doing on the Acinq node for Phoenix features).

When the peer receives an unknown message, it checks in its `nodeParams` if something subscribed to that message type; if yes, it simply sends it to the eventStrean. WDYT?

btcontract:
> When the peer receives an unknown message

To make sure I understand this right: there will be something like `UnknownMessage(tag: Int, data: ByteVector)`,

t-bast:
Yes that's exactly what I had in mind (if there's not blocking point when implementing that).

btcontract:
I've started on this, but right at the outset this feels a bit futile since:

- `discriminatorWithDefault` returns `Codec[A]` so `discriminatorWithDefault(lightningMessageCodec ...)` will have to return a `Codec[LightningMessage]` so we will need to have `UnknownMessage(tag: Int, data: ByteVector) extends LightningMessage` anyway (identical to `PluginMessage`).
- since there are no checks on `data: ByteVector` at core level (and should not be) a plugin may sort of fool core and build its entire messaging system through one `UnknownMessage` with the same `tag` which is the same as `PluginMessage`.

Am I missing something?

t-bast:
> UnknownMessage(tag: Int, data: ByteVector) extends LightningMessage anyway (identical to PluginMessage).

Yes that's totally ok, it is a `LightningMessage` so it makes sense.

> since there are no checks on data: ByteVector at core level (and should not be) a plugin may sort of fool core and build its entire messaging system through one UnknownMessage with the same tag which is the same as PluginMessage.

That's ok too, it's up to the plugin to decide how it wants to use it.
But what's important is that at the core layer, we make all usages possible (to either use multiple messages or use a single one).

btcontract:
Okay, this might be overly complex, maybe there are easier ways but anyway:

- there are two new LN message types: 
  - `KnownUnknownMessage` (has a standard `tag == STANDARD_UNKNOWN_MESSAGE_TAG` so can be handled by `lightningMessageCodec`)
  - `UnknownUnknownMessage` (has `tag != STANDARD_UNKNOWN_MESSAGE_TAG`) so `lightningMessageCodec` fails on it and then `lightningMessageCodecWithFallback` handles it.

- `Peer` only handles `UnknownUnknownMessage`s as the ones which are clearly "unknown unknown".

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -405,3 +407,7 @@ case object EmptyAPIPasswordException extends RuntimeException("must set a passw
 case object IncompatibleDBException extends RuntimeException("database is not compatible with this version of eclair")
 
 case object IncompatibleNetworkDBException extends RuntimeException("network database is not compatible with this version of eclair")
+
+case class PluginTagsAndFeature(tags: Set[Int], feature: Feature) {
```
t-bast, line 411: What about `PluginParams`?
It would be more in-line with `nodeParams` and we could enrich this case class with other plugin parameters that we may want to add in the future.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -28,6 +28,8 @@ import scodec.{Attempt, Codec}
  */
 object LightningMessageCodecs {
 
+  final val STANDARD_UNKNOWN_MESSAGE_TAG = 42263
```
t-bast, line 31: We should not have to hardcode something plugin-specific like that inside eclair-core.
See other comments below.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -297,6 +299,34 @@ object LightningMessageCodecs {
...
+  val knownUnknownMessageCodec: Codec[KnownUnknownMessage] = (
+    ("tag" | uint16) ::
+      ("data" | varsizebinarydata)
+    ).as[KnownUnknownMessage]
+
+  val unknownUnknownMessageCodec: Codec[UnknownUnknownMessage] = (
+    ("tag" | uint16) ::
+      ("message" | knownUnknownMessageCodec)
+    ).as[UnknownUnknownMessage]
+
```
t-bast: I don't understand why you want to distinguish between these two.
Why can't you just have a single `UnknownMessage` that you use in the fallback of the discriminator?

btcontract: indeed

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -405,3 +407,7 @@ case object EmptyAPIPasswordException extends RuntimeException("must set a passw
 case object IncompatibleDBException extends RuntimeException("database is not compatible with this version of eclair")
 
 case object IncompatibleNetworkDBException extends RuntimeException("network database is not compatible with this version of eclair")
+
+case class PluginParams(tags: Set[Int], feature: Feature) {
+  def activatedFeature: ActivatedFeature = ActivatedFeature(feature, FeatureSupport.Optional)
+}
```
t-bast, line 413: I think this doesn't belong in `Setup.scala`. Maybe in `NodeParams.scala`?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -49,6 +49,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
 
   import Peer._
 
+  val pluginMessageTags: Seq[Int] = nodeParams.pluginParams.flatMap(_.tags)
```
t-bast, line 52: You could make that a `Set`

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -197,7 +199,16 @@ object NodeParams {
...
+
     require(featuresErr.isEmpty, featuresErr.map(_.message))
+    require(pluginParams.forall(_.feature.mandatory > 128), "Plugin mandatory feature bit is too low, must be > 128")
+    require(pluginParams.forall(_.feature.mandatory % 2 == 0), "Plugin mandatory feature bit is odd, must be even")
+    require(pluginParams.flatMap(_.tags).forall(_ > 32768), "Plugin messages tags must be > 32768")
+    val pluginFeatureSet = pluginParams.map(_.feature.mandatory).toSet
+    require(Features.knownFeatures.map(_.mandatory).intersect(pluginFeatureSet).isEmpty, "Plugin feature bit overlaps with known feature bit")
+    require(pluginFeatureSet.size == pluginParams.size, "Duplicate plugin feature bits found")
+
+    val coreAndPluginFeatures = features.copy(activated = features.activated ++ pluginParams.map(_.activatedFeature))
```
t-bast, line 211: Is there a reason you can't use `unknown` set? It looks like it would be a good fit:

```suggestion
    val coreAndPluginFeatures = features.copy(unknown = pluginParams.map(p => UnknownFeature(p.feature.optional)).toSet)
```

eclair-node/src/main/scala/fr/acinq/eclair/Plugin.scala
```diff
@@ -18,12 +18,13 @@ package fr.acinq.eclair
 
 import java.io.File
 import java.net.{JarURLConnection, URL, URLClassLoader}
-
 import grizzled.slf4j.Logging
-
 import scala.util.{Failure, Success, Try}
 
 trait Plugin {
+  // Set of LightningMessage tags this plugin is interested in and a Feature bit it advertizes through Init message
```
t-bast, line 25: nit: this comment will become obsolete as soon as we add more data to the `PluginParams` structure.
Why not move these descriptions inside `PluginParams`? The name of the field here is self-explanatory and shouldn't need any comment.

t-bast (commented):
This is shaping up nicely, let us know when you've tested this E2E with a sample plugin!

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -203,10 +203,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
         d.channels.values.toSet[ActorRef].foreach(_ ! INPUT_DISCONNECTED) // we deduplicate with toSet because there might be two entries per channel (tmp id and final id)
         gotoConnected(connectionReady, d.channels)
 
-      case Event(unknownMsg: UnknownMessage, _: ConnectedData) =>
-        if (pluginMessageTags.contains(unknownMsg.tag)) {
-          context.system.eventStream.publish(UnknownMessageReceived(self, remoteNodeId, unknownMsg))
-        }
+      case Event(unknownMsg: UnknownMessage, _: ConnectedData) if pluginMessageTags.contains(unknownMsg.tag) =>
```
btcontract, line 206: Besides mentioned fixes this one makes an `UnknownMessage` without subscribers to be logged in next case instead of failing silently.

btcontract:
I've made a sample here: https://github.com/btcontract/eclair-plugin-messaging-sample. 

It pushes a json-encoded local feerate to remote peer on connect if remote peer supports this plugin (makes use of `remoteInit` and I've chosen json just to show anything is possible).

It compiles but I could not figure out how to build my snapshot jar such that it could run. Looks like something changed in build process and jars do not contain all dependencies inside now? Some instructions on building it would be appreciated.

t-bast:
> It compiles but I could not figure out how to build my Eclair snapshot jar such that it could run. Looks like something changed in build process and jars do not contain all dependencies inside now? Some instructions on building it would be appreciated.

We now use standard maven builds (to get reproducible builds) instead of capsule.
This means there is no single-jar deployment anymore, but `mvn package` will instead create a folder with all the necessary jar.
Isn't that working for you? I agree instructions for building plugins could be better than the small section we have on the `README.md`, if you have suggestions feel free to edit it!

btcontract:
My bad, a zip file with everything needed was being created all the time, but I did not pay attention to it.

OK, I've just tested this with two Eclair instances and it works, I can see `Remote peer sent feerates=50602 via plugin` in both logs, as it should be according to https://github.com/btcontract/eclair-plugin-messaging-sample/blob/master/src/main/scala/fr/acinq/messagingsample/app/Notifier.scala#L21

t-bast:
Great, the overall architecture sounds good, let me know when this is ready for review.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerEvents.scala
```diff
@@ -18,9 +18,12 @@ package fr.acinq.eclair.io
 
 import akka.actor.ActorRef
 import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.eclair.wire.UnknownMessage
 
 sealed trait PeerEvent
 
-case class PeerConnected(peer: ActorRef, nodeId: PublicKey) extends PeerEvent
+case class PeerConnected(peer: ActorRef, nodeId: PublicKey, data: Peer.Data) extends PeerEvent
```
t-bast, line 25: Can you explain why you need to expose the `Peer.Data`? It's internal data, I'd prefer to keep it hidden if possible (and maybe expose only what you found useful for plugins as a new `case class`).

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -26,17 +26,16 @@ import com.google.common.net.HostAndPort
...
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.FeatureSupport.Optional
-import fr.acinq.eclair.Features.{Wumbo, StaticRemoteKey}
+import fr.acinq.eclair.Features.{StaticRemoteKey, Wumbo}
 import fr.acinq.eclair.TestConstants._
 import fr.acinq.eclair._
 import fr.acinq.eclair.blockchain.{EclairWallet, TestWallet}
 import fr.acinq.eclair.channel.states.StateTestsHelperMethods
-import fr.acinq.eclair.channel.{CMD_GETINFO, Channel, ChannelCreated, ChannelVersion, DATA_WAIT_FOR_ACCEPT_CHANNEL, HasCommitments, RES_GETINFO, WAIT_FOR_ACCEPT_CHANNEL}
+import fr.acinq.eclair.channel.{CMD_GETINFO, Channel, ChannelCreated, DATA_WAIT_FOR_ACCEPT_CHANNEL, HasCommitments, RES_GETINFO, WAIT_FOR_ACCEPT_CHANNEL}
```
t-bast, line 34: I think it's worth adding tests here:
1) Test that when receiving an `UnknownMessage` that a plugin subscribed to, it's forwarded to the event stream
2) Test that when receiving an `UnknownMessage` for which no plugin subscribed, nothing is sent to the event stream

eclair-core/src/test/scala/fr/acinq/eclair/wire/LightningMessageCodecsSpec.scala
```diff
@@ -195,20 +195,33 @@ class LightningMessageCodecsSpec extends AnyFunSuite {
...
+        val decoded = lightningMessageCodecWithFallback.decode(encoded).require
         assert(msg === decoded.value)
       }
     }
   }
 
+  test("Unknown messages") {
+    // Non-standard tag number so this message can only be handled by a codec with a fallback
+    val unknown = UnknownMessage(tag = 47282, data = ByteVector32.Zeroes.bytes)
+    assertThrows[java.lang.IllegalArgumentException](lightningMessageCodec.encode(unknown).require)
```
t-bast, line 218:
```suggestion
    assert(lightningMessageCodec.encode(unknown).isFailure)
```

eclair-core/src/test/scala/fr/acinq/eclair/wire/LightningMessageCodecsSpec.scala
```diff
@@ -195,20 +195,33 @@ class LightningMessageCodecsSpec extends AnyFunSuite {
...
+    val encoded1 = lightningMessageCodecWithFallback.encode(unknown).require
+    val decoded1 = lightningMessageCodecWithFallback.decode(encoded1).require.value
+    assertThrows[java.lang.IllegalArgumentException](lightningMessageCodec.decode(encoded1).require.value)
```
t-bast, line 221:
```suggestion
    assert(lightningMessageCodec.decode(encoded1).isFailure)
```

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerEvents.scala
btcontract: To be able to send an `OutgoingMessage` back to `Peer`, a `peerConnection` is needed for that as far as I understand and `Peer.Data` always has the latest one, I use it like this: https://github.com/btcontract/eclair-plugin-messaging-sample/blob/master/src/main/scala/fr/acinq/messagingsample/app/Notifier.scala#L20

btcontract: Also `Peer.Data` has a `remoteInit` which allows a plugin to check if connected remote peer supports this plugin and push some data to remote if it does.

t-bast: Got it, in that case I'd rather only expose what's needed:
```
case class ConnectionInfo(peerConnection: ActorRef, remoteInit: wire.Init)
case class PeerConnected(peer: ActorRef, nodeId: PublicKey, connectionInfo: ConnectionInfo) extends PeerEvent
case class UnknownMessageReceived(peer: ActorRef, nodeId: PublicKey, message: UnknownMessage, , connectionInfo: ConnectionInfo) extends PeerEvent
```

btcontract:
There is a slight complication when using `ConnectionInfo` class: in `onTransition` only `nextStateData` is visible which is `Peer.Data`, not `ConnectedData`. 

To solve it I've defined a method `connectionInfo: Option[ConnectionInfo] = None` on `Peer.Data` and overrided it in `ConnectedData`. As a result `PeerConnected` and `UnknownMessageReceived` now have `Option[ConnectionInfo]` instead of `ConnectionInfo` which is still easily matchable in plugin.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -162,6 +162,19 @@ class PeerSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with StateTe
...
   }
 
+  test("handle unknown messages") { f =>
+    import f._
+
+    val listener = TestProbe()
+    system.eventStream.subscribe(listener.ref, classOf[UnknownMessageReceived])
+    connect(remoteNodeId, peer, peerConnection, channels = Set(ChannelCodecsSpec.normal))
+
+    peerConnection.send(peer, UnknownMessage(tag = 50001, data = ByteVector.empty))
```
t-bast, line 172: When someone reads this test out of context, they will have no way of understand why 50001 triggers an event whereas 60001 doesn't...

I think you should add a `val TestPluginParams = PluginParams(Set(50001), TestFeature)` inside `TestConstants`, and in this test you should do:

```
peerConnection.send(peer, UnknownMessage(tag = TestPluginParams.tags.head, data = ByteVector.empty))
```

You should also use different values for the `TestFeature` and the message tag, otherwise it's misleading. They have no reason to be the same (usually a feature will be associated with multiple messages). Set `TestFeature` to any random value you like.

t-bast (commented):
While I agree that `nextStateData` isn't properly typed in transitions, we know that for transitions that go to the `CONNECTED` state (which is where the `PeerConnected` event is sent) it will always be a `ConnectedData`, so that shouldn't be an `Option`.

You'll need to case `nextStateData.asInstanceOf[ConnectedData]` in the transition.
Once we migrate this actor to Akka typed we probably won't need this explicit cast.

t-bast (approved):
LGTM, thanks!

PR #1529, opened 2020-09-17
Improve relayer logs

pm47:
We can be less repetitive given that we now have a `relayId` that we can
use to factor all info.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
@@ -74,11 +74,15 @@ object ChannelRelayer {
           Behaviors.receiveMessage {
             case Relay(channelRelayPacket) =>
               val relayId = UUID.randomUUID()
-              val channels: Map[ShortChannelId, Relayer.OutgoingChannel] = channelUpdates.get(channelRelayPacket.payload.outgoingChannelId) match {
-                case Some(channel) => node2channels.get(channel.nextNodeId).map(channelUpdates).map(c => c.channelUpdate.shortChannelId -> c).toMap
+              val nextNodeId_opt: Option[PublicKey] = channelUpdates.get(channelRelayPacket.payload.outgoingChannelId) match {
+                case Some(channel) => Some(channel.nextNodeId)
+                case None => None
+              }
+              val channels: Map[ShortChannelId, Relayer.OutgoingChannel] = nextNodeId_opt match {
+                case Some(nextNodeId) => node2channels.get(nextNodeId).map(channelUpdates).map(c => c.channelUpdate.shortChannelId -> c).toMap
                 case None => Map.empty
               }
-              context.log.debug(s"spawning a new handler with relayId=$relayId channels={}", channels.keys.mkString(","))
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -74,11 +74,15 @@ object ChannelRelayer {
           Behaviors.receiveMessage {
             case Relay(channelRelayPacket) =>
               val relayId = UUID.randomUUID()
-              val channels: Map[ShortChannelId, Relayer.OutgoingChannel] = channelUpdates.get(channelRelayPacket.payload.outgoingChannelId) match {
-                case Some(channel) => node2channels.get(channel.nextNodeId).map(channelUpdates).map(c => c.channelUpdate.shortChannelId -> c).toMap
+              val nextNodeId_opt: Option[PublicKey] = channelUpdates.get(channelRelayPacket.payload.outgoingChannelId) match {
+                case Some(channel) => Some(channel.nextNodeId)
+                case None => None
+              }
```
t-bast, line 80: nit:
```suggestion
              val nextNodeId_opt: Option[PublicKey] = channelUpdates.get(channelRelayPacket.payload.outgoingChannelId).map(_.nextNodeId)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -94,11 +98,13 @@ object ChannelRelayer {
             case WrappedLocalChannelUpdate(LocalChannelUpdate(_, channelId, shortChannelId, remoteNodeId, _, channelUpdate, commitments)) =>
               context.log.debug(s"updating local channel info for channelId=$channelId shortChannelId=$shortChannelId remoteNodeId=$remoteNodeId channelUpdate={} commitments={}", channelUpdate, commitments)
               val channelUpdates1 = channelUpdates + (channelUpdate.shortChannelId -> Relayer.OutgoingChannel(remoteNodeId, channelUpdate, commitments))
-              apply(nodeParams, register, channelUpdates1, node2channels.addOne(remoteNodeId, channelUpdate.shortChannelId))
+              val node2channels1 = node2channels.addOne(remoteNodeId, channelUpdate.shortChannelId)
+              apply(nodeParams, register, channelUpdates1, node2channels1)
```
t-bast, line 102: Why did you make it an explicit variable?
Is it used differently in feature branches?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
pm47: The format I went with allows for easier merge in feature branches.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
pm47: It is more consistent than doing the operation in the function call.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
t-bast: Ok, got it!

t-bast (approved):

PR #1531, opened 2020-09-18
Minor improvements to the watcher

pm47:
This is mostly about cleaning up useless logs.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
@@ -53,7 +53,13 @@ class ZmqWatcher(blockCount: AtomicLong, client: ExtendedBitcoinClient)(implicit
   // this is to initialize block count
   self ! TickNewBlock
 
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -146,17 +158,22 @@ class ZmqWatcher(blockCount: AtomicLong, client: ExtendedBitcoinClient)(implicit
                   }
               }
           }
+          Keep
 
-        case w: WatchConfirmed => checkConfirmed(w) // maybe the tx is already tx, in that case the watch will be triggered and removed immediately
-
-        case _: WatchLost => () // TODO: not implemented
+        case w: WatchConfirmed =>
+          checkConfirmed(w) // maybe the tx is already tx, in that case the watch will be triggered and removed immediately
```
t-bast, line 164:
```suggestion
          checkConfirmed(w) // maybe the tx is already confirmed, in that case the watch will be triggered and removed immediately
```

t-bast (approved):

PR #1532, opened 2020-09-18
Update to scala 2.13.3 and akka 2.6.9

pm47:


t-bast (approved):
Progress never stops!
Anything in particular you're interested in for these releases, or just keeping up to date?

pm47:
Just keeping up to date, there are a few bugfixes that are interesting to get. If you remember, we had to downgrade to scala 2.13.1 because of a bug in scala 2.13.2 ( #1394).

PR #1533, opened 2020-09-18
Handle the `NoReplyTo` also in the error case

pm47:


t-bast (approved):

pm47:
I'm not sure about this change anymore. I had to update tests so that they don't expect `RES_FAILURE[CMD_FAIL/FULFILL_HTLC]` anymore but... now those tests don't mean much and the channel is kind of a black box.

I wonder if we should replace `NoReplyTo` by `OptionalReplyTo` or something like that.

t-bast:
True, it's quite helpful in fact to have a response here just to acknowledge correct execution, if only for the tests...

pm47:
Superseded by #1543.

PR #1534, opened 2020-09-21
[OpenChannel] Use TLV for channel version

t-bast:
We were previously using an encoding that was incompatible with TLV.
We keep the old format, but also understand the channel version when provided in a TLV field.

For backwards-compatibility, if the old format is provided, it will be preferred.
Since Phoenix only reads `open_channel` messages, it only needs support to *decode* this new TLV.
It will be on the eclair-node side that we will to switch to write the new TLV instead of the old format once we're confident enough Phoenix users have upgraded.

NB: this PR doesn't merge to master, but to `android-phoenix`

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/wire/OpenTlvSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/wire/OpenTlvSpec.scala
@@ -16,16 +16,11 @@
 
 package fr.acinq.eclair.wire
 
-import fr.acinq.eclair.UInt64
 import fr.acinq.eclair.channel.ChannelVersion
-import fr.acinq.eclair.wire.ChannelTlv.UpfrontShutdownScript
-import fr.acinq.eclair.wire.CommonCodecs.{varint, varintoverflow}
 import fr.acinq.eclair.wire.OpenChannelTlv.ChannelVersionTlv
-import fr.acinq.eclair.wire.TlvCodecs.tlvStream
 import org.scalatest.funsuite.AnyFunSuite
 import scodec.bits._
-import scodec.codecs.{bits, bytes, discriminated, variableSizeBytesLong}
-import scodec.{Attempt, Codec, DecodeResult}
+import scodec.{Attempt, DecodeResult}
 
 class OpenTlvSpec extends AnyFunSuite {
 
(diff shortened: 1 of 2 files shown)


sstone (commented):
LGTM, just a nit: would it be easy to add a test that reads the new channel version as a generic TLV (to check thtat it does comply to the TLV encoding requirements) ?

t-bast:
Done in https://github.com/ACINQ/eclair/pull/1534/commits/4914a636a89f28d3a36e3b3c7c3bcf6b5b13c7bc, I had tested it once before changing the codec, but it's worth having as a temporary separate test indeed.

dpad85 (approved):

eclair-core/src/test/scala/fr/acinq/eclair/wire/OpenTlvSpec.scala
```diff
@@ -16,22 +16,43 @@
...
+    // This is the codec that was used previously.
+    val previousCodec: Codec[TlvStream[OpenChannelTlv]] = tlvStream(discriminated[OpenChannelTlv].by(varint)
+      .typecase(UInt64(0), variableSizeBytesLong(varintoverflow, bytes).as[UpfrontShutdownScript])
+      .typecase(UInt64(0x47000000), bits(ChannelVersion.LENGTH_BITS).as[ChannelVersion].as[ChannelVersionTlv])
     )
 
-    refs.foreach { case (cv, bits) => assert(OpenChannelTlv.openTlvCodec.encode(TlvStream(ChannelVersionTlv(cv))) === Attempt.Successful(bits)) }
-    refs.foreach { case (cv, bits) => assert(OpenChannelTlv.openTlvCodec.decode(bits) === Attempt.successful(DecodeResult(TlvStream(ChannelVersionTlv(cv)), BitVector.empty))) }
+    assert(previousCodec.decode(hex"fe47000001 04 00000001".bits).require.value === TlvStream(Nil, Seq(GenericTlv(UInt64(0x47000001), hex"00000001"))))
+    assert(previousCodec.decode(hex"fe47000001 04 00000009".bits).require.value === TlvStream(Nil, Seq(GenericTlv(UInt64(0x47000001), hex"00000009"))))
```
pm47, line 55: The test is okay, I would have sent both tlvs (new and old) to this old codec and make sure it was ignoring the new tlv and interpreting the old one). Seems closer to our use case?

t-bast: The encoder will never add the two in the same tlv stream since it's the same Scala type.
On the contrary, what whill happen is that eclair-node will continue sending the old encoding, but once Phoenix is upgraded to support the new one, we can switch eclair-node to use the new encoding when writing.
Does that make sense?

pm47: Gotcha. But the test is misleading: even if old Phoenix can parse the new format, they will still break because they will use an invalid channel version, different from the funder. If we want a smooth transition, we need to modify the encoding to have it write both tags.

t-bast: I removed that test. We'll guarantee backwards-compatibility by simply waiting for users to upgrade, as writing both tags will require hacks to the tlv stream codecs...

pm47 (approved):

PR #1535, opened 2020-09-21
Split IntegrationSpec

t-bast:
Split the IntegrationSpec into two distinct test suites:

- Payment tests (HTLC, Trampoline, Routing, etc)
- Channel management tests (wumbo, Closing, etc)

The main benefit is that it speeds up full runs of the test suite (because specs are run in parallel), both locally and during the CI.
It uses a bit more CPU and memory because it's one more `bitcoind` to run, but the result still looks like a net improvement.

Note that I split the review in several commits: the first one just moves code around with minimal changes. Only the other commits apply meaningful changes to the tests.


pm47:
14 min -> 12 min build time on ci, that is a 15% reduction. Nice!

pm47:
The speed improvement is spectacular on my machine : 
- eclair-core tests : 7m45 -> 4m32 (-42 %)
- `mvn clean test` : 9m18 -> 6m06 (-34 %)

t-bast:
That means you have a good machine with enough cores and RAM :smiley:

pm47:
There is probably room for further improvements. On my machine these 6 tests run alone at the end and are very slow:
```
- propagate a failure upstream when a downstream htlc times out (local commit)
- propagate a failure upstream when a downstream htlc times out (local commit, anchor outputs)
- propagate a failure upstream when a downstream htlc times out (remote commit)
- propagate a failure upstream when a downstream htlc times out (remote commit, anchor outputs)
- punish a node that has published a revoked commit tx
- punish a node that has published a revoked commit tx (anchor outputs)
```

t-bast:
But those are basically all the tests in `ChannelIntegrationSpec` :laughing: 
I'll try splitting it in two, but then it will be sacrificing some consistency.

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -131,11 +109,19 @@ class IntegrationSpec extends TestKitBaseClass with BitcoindService with AnyFunS
...
+    awaitCond({
+      sender.send(bitcoincli, BitcoinReq("getnetworkinfo"))
+      sender.receiveOne(5 second).isInstanceOf[JValue]
+    }, max = 30 seconds, interval = 500 millis)
+    logger.info(s"generating initial blocks...")
+    generateBlocks(bitcoincli, 150, timeout = 30 seconds)
   }
 
-  override def afterAll: Unit = {
+  override def afterAll(): Unit = {
```
pm47, line 124: Ha, we are not calling `super.afterAll()`, which would be `TestKitBaseClass.afterAll()`, which calls `TestKit.shutdownActorSystem(system)`. Maybe the explanation behind intermittent freezing after tests pass?

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -17,59 +17,37 @@
...
-import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.jdk.CollectionConverters._
 
 /**
  * Created by PM on 15/03/2017.
  */
 
-class IntegrationSpec extends TestKitBaseClass with BitcoindService with AnyFunSuiteLike with BeforeAndAfterAll with Logging {
+trait IntegrationSpec extends TestKitBaseClass with BitcoindService with AnyFunSuiteLike with BeforeAndAfterAll with Logging {
```
pm47, line 50: It feels like this really is an `abstract class`

```suggestion
abstract class IntegrationSpec extends TestKitBaseClass with BitcoindService with AnyFunSuiteLike with BeforeAndAfterAll with Logging {
```

pm47:
> But those are basically all the tests in ChannelIntegrationSpec 😆
> I'll try splitting it in two, but then it will be sacrificing some consistency for performance...

I know, maybe we can be a bit smarter with test execution instead? Longer tests should be started first.

eclair-core/src/test/scala/fr/acinq/eclair/integration/ChannelIntegrationSpec.scala
```diff
@@ -0,0 +1,724 @@
...
+import scala.jdk.CollectionConverters._
+
+/**
+ * Created by t-bast on 21/09/2020.
+ */
+
+class ChannelIntegrationSpec extends IntegrationSpec {
+
+  test("start eclair nodes") {
+    instantiateEclairNode("A", ConfigFactory.parseMap(Map("eclair.node-alias" -> "A", "eclair.expiry-delta-blocks" -> 40, "eclair.fulfill-safety-before-timeout-blocks" -> 12, "eclair.server.port" -> 29740, "eclair.api.port" -> 28090).asJava).withFallback(commonFeatures).withFallback(commonConfig))
```
pm47, line 54: Looks like you had fun with port allocation 😜

pm47: Nice move to reduce the `expiry-delta`.

pm47 (commented):
Looks already pretty good to me as-is. Maybe worth looking into the test execution, if we can get an additional 10-20%.

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
t-bast: Interesting, good catch!

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1535/commits/9a85d4b7dd44c39d854fac9e0f72991a53891719

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1535/commits/9a85d4b7dd44c39d854fac9e0f72991a53891719

t-bast:
> There is probably room for further improvements. 

Can you try on your machine with the latest commit (https://github.com/ACINQ/eclair/pull/1535/commits/f8c6f599d5e35d518bc9cb773d0d4fb4d42689a5). On mine it still makes the perf better, and the last test to finish is now one of the `PeerSpec`!

pm47:
>> There is probably room for further improvements.

> Can you try on your machine with the latest commit (f8c6f59). On mine it still makes the perf better, and the last test to finish is now one of the PeerSpec!

Yep! now:
- eclair-core tests : 7m45 -> 4m32 (-42 %) -> 3m1 (-30 %)
- mvn clean test : 9m18 -> 6m06 (-34 %) -> 4m43 (-23 %)

I'm surprised the split wasn't more straightforward tough.

t-bast:
> I'm surprised the split wasn't more straightforward tough.

We could factor a bit more inside `ChannelIntegrationSpec`, but I think it may be limiting for future tests we'll want to add, so I went with the simple route with some duplication in each child class, but it was quite straightforward, just moving stuff around and changing a few ports / cleaning up the references to F1 and F2 to always be F.

t-bast:
It doesn't speed up the CI though, it's exactly the same time (11 min 30) as before, so I think we're hitting the dual-core limits :)

pm47:
How about we put `StandardChannelIntegrationSpec`/`AnchorOutputChannelIntegrationSpec` in separate files? More readable and easier to compare.

t-bast:
> How about we put StandardChannelIntegrationSpec/AnchorOutputChannelIntegrationSpec in separate files?

TBH I find it harder to navigate between multiple files...since it's not too big I like it in the same file

pm47 (approved):

PR #1540, opened 2020-09-24
Configure bitcoin wallet

### LINKED ISSUE #1538: BasicBitcoinJsonRPCClient uses hard-coded wallet name
Tracked it down to this [line][0]:
https://github.com/ACINQ/eclair/blob/662e0c4bcc1dec24f9aad2bd526866a8a056153f/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala#L40

### Background story

Today I wanted to fix broken tests in [simverse](https://github.com/darwin/simverse).

Here are my findings (note that I don't follow bitcoin core development that closely, so I may be missing something): 

1. bitcoind used to create a default wallet with empty name. It seems to be no longer the case. Newly I have to first create some wallet via `bitcoin-cli createwallet <name>`. See my commit [here][1].

2. eclair started failing with `"bitcoind must have wallet support enabled"`, I double checked that my bitcoind really had wallet support enabled and started googling the issue. I found the same problem reported [here][2].

3. I looked at the place where the [error is raised][3] and decided to investigate further.

4. I found out that the issue is that eclair has [hardcoded empty wallet name][0].

5. bitcoind logic for selecting wallet for RPC request is implemented [here][4].

### Possible suggestions

1. BasicBitcoinJsonRPCClient should be configurable to allow wallet name selection (not sure how to propagate this configurability further)

2. By default BasicBitcoinJsonRPCClient should have unspecified name (which is different from empty string), If I understand the rpcwallet.cpp correctly, it would be more flexible in this case not to add "/wallet/" into endpoint url. Bitcoin RPC would then pick the only wallet it has loaded regardless of its name (or return an error about multi-wallet scenario via `RPC_WALLET_NOT_SPECIFIED` or an error about no-wallet scenario via `RPC_METHOD_NOT_FOUND` or any other error which could be implemented on bitcoind side in the future). This would be more flexible than assuming default wallet name to be empty string.

3. Secondary problem is that Setup.scala#L150 swallows original RPC error. Even if you didn't resolve points 1 or 2, the code should be updated to include original RPC error. It should not blindly return `"bitcoind must have wallet support enabled"` which may be misleading, because it assumes particular behaviour of bitcoind which may be subject of future changes.

[0]: https://github.com/ACINQ/eclair/blob/662e0c4bcc1dec24f9aad2bd526866a8a056153f/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala#L40
[1]: https://github.com/darwin/simverse/commit/296805226fd8d249cd5cfabe1a6d20f712cdbd7c
[2]: https://gitter.im/ACINQ/eclair?at=5e48f793c8da1343d4540a44
[3]: https://github.com/ACINQ/eclair/blob/daddfc007fe569c89bb854ef5d672614d397e91e/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala#L150
[4]: https://github.com/bitcoin/bitcoin/blob/1b313cacc99a1b372238f9036abed5491f9d28f7/src/wallet/rpcwallet.cpp#L102

t-bast:
It makes sense to allow users to use a different wallet from the default one.

There's one important caveat: once set, users shouldn't change it while they have open channels.
We mention this clearly in the documentation, I don't think we need an explicit mechanism in the code to prevent starting with a different wallet (for example, if no channels are open, it's safe to change your wallet). It's use-at-your-own-risk if users choose to ignore the documented caveats. @darwin is that trade-off ok?

Fixes #1538

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,15 +45,8 @@ You will find detailed guides and frequently asked questions there.
 
 Eclair needs a _synchronized_, _segwit-ready_, **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node.
 
-Eclair will use the default loaded Bitcoin Core wallet to fund any channels you choose to open.
-If you want to use a different wallet from the default one, you must set `eclair.bitcoind.wallet` accordingly in your `eclair.conf`.
-
-:warning: once a wallet is configured, you must be very careful if you want to change it: changing the wallet when you have channels open may result in a loss of funds (or a complex recovery procedure).
-
-Eclair will return BTC from closed channels to the wallet configured.
-Any BTC found in the wallet can be used to fund the channels you choose to open.
-You can configure your Bitcoin Node to use either `p2sh-segwit` addresses or `bech32` addresses, Eclair is compatible with both modes.
-If your wallet has "non-segwit UTXOs" (outputs that are neither `p2sh-segwit` or `bech32`), you must send them to a `p2sh-segwit` or `bech32` address.
+You can configure your Bitcoin node to use either `p2sh-segwit` addresses or `bech32` addresses, Eclair is compatible with both modes.
+If your wallet has "non-segwit UTXOs" (outputs that are neither `p2sh-segwit` or `bech32`), you must send them to a `p2sh-segwit` or `bech32` address before running eclair.
 
 Run bitcoind with the following minimal `bitcoin.conf`:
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -41,13 +41,19 @@ You will find detailed guides and frequently asked questions there.
...
-Eclair will use any BTC it finds in the default Bitcoin Core wallet to fund any channels you choose to open. Eclair will return BTC from closed channels to this wallet. You can have multiple Bitcoin Core wallets but make sure that the default one is always available.
-Any BTC found in the wallet can be used to fund the channels you choose to open and the BTC from closed channels will return to this wallet. 
+
+Eclair will use the default loaded Bitcoin Core wallet to fund any channels you choose to open.
+If you want to use a different wallet from the default one, you must set `eclair.bitcoind.wallet` accordingly in your `eclair.conf`.
+
+:warning: once a wallet is configured, you must be very careful if you want to change it: changing the wallet when you have channels open may result in a loss of funds (or a complex recovery procedure).
+
+Eclair will return BTC from closed channels to the wallet configured.
+Any BTC found in the wallet can be used to fund the channels you choose to open.
```
pm47, line 54: This addition makes the paragraph quite big for a very particular use case. Maybe we should in subsections:
- Configuration file (like in Configuring Eclair)
- Custom wallet

darwin:
LGTM

README.md
t-bast: I agree, it's worth its separate section.

t-bast: Let me know if https://github.com/ACINQ/eclair/pull/1540/commits/ba9da7fe87223cad6f458d114145726ca2323d0c is better

README.md
```diff
@@ -100,11 +99,22 @@ name                         | description
...
 Quotes are not required unless the value contains special characters. Full syntax guide [here](https://github.com/lightbend/config/blob/master/HOCON.md).
 
 &rarr; see [`reference.conf`](eclair-core/src/main/resources/reference.conf) for full reference. There are many more options!
 
+#### Configure Bitcoin Core wallet
+
+Eclair will use the default loaded Bitcoin Core wallet to fund any channels you choose to open.
+If you want to use a different wallet from the default one, you must set `eclair.bitcoind.wallet` accordingly in your `eclair.conf`.
+
+:warning: once a wallet is configured, you must be very careful if you want to change it: changing the wallet when you have channels open may result in a loss of funds (or a complex recovery procedure).
```
pm47, line 113:
```suggestion
:warning: Once a wallet is configured, you must be very careful if you want to change it: changing the wallet when you have channels open may result in a loss of funds (or a complex recovery procedure).
```

t-bast: Done!

pm47 (approved):

PR #1542, opened 2020-09-25
Make a ChannelCommitments trait, extend Commitments from it

btcontract:
Why?

I want to enable hosted channels (HC) entirely as a plugin as has been discussed before, three things are needed to enable this:

- `Peer` should be able to handle HC-specific messages (enabled already with plugin messaging)
- HC should be able to inject its commitments data into `Router` and `Relayer` to enable routing from HC to normal channels and vice versa, enabled by this PR.
- `PostRestartHtlcCleaner` should be aware of HC-routed broken and pending HTLCs (next PR, perhaps another plugin extension).

Before going further I'd like to know if this change is acceptable.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -246,7 +246,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
           // we rebuild a new channel_update with values from the configuration because they may have changed while eclair was down
           val candidateChannelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, normal.channelUpdate.shortChannelId, nodeParams.expiryDelta,
-            normal.commitments.remoteParams.htlcMinimum, normal.channelUpdate.feeBaseMsat, normal.channelUpdate.feeProportionalMillionths, normal.commitments.localCommit.spec.totalFunds, enable = Announcements.isEnabled(normal.channelUpdate.channelFlags))
+            normal.commitments.remoteParams.htlcMinimum, normal.channelUpdate.feeBaseMsat, normal.channelUpdate.feeProportionalMillionths, normal.commitments.capacity.toMilliSatoshi, enable = Announcements.isEnabled(normal.channelUpdate.channelFlags))
           val channelUpdate1 = if (Announcements.areSame(candidateChannelUpdate, normal.channelUpdate)) {
             // if there was no configuration change we keep the existing channel update
             normal.channelUpdate
@@ -598,7 +598,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       blockchain ! WatchConfirmed(self, commitments.commitInput.outPoint.txid, commitments.commitInput.txOut.publicKeyScript, ANNOUNCEMENTS_MINCONF, BITCOIN_FUNDING_DEEPLYBURIED)
       context.system.eventStream.publish(ShortChannelIdAssigned(self, commitments.channelId, shortChannelId, None))
       // we create a channel_update early so that we can use it to send payments through this channel, but it won't be propagated to other nodes since the channel is not yet announced
-      val initialChannelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, shortChannelId, nodeParams.expiryDelta, d.commitments.remoteParams.htlcMinimum, nodeParams.feeBase, nodeParams.feeProportionalMillionth, commitments.localCommit.spec.totalFunds, enable = Helpers.aboveReserve(d.commitments))
+      val initialChannelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, shortChannelId, nodeParams.expiryDelta, d.commitments.remoteParams.htlcMinimum, nodeParams.feeBase, nodeParams.feeProportionalMillionth, commitments.capacity.toMilliSatoshi, enable = Helpers.aboveReserve(d.commitments))
       // we need to periodically re-send channel updates, otherwise channel will be considered stale and get pruned by network
       context.system.scheduler.schedule(initialDelay = REFRESH_CHANNEL_UPDATE_INTERVAL, interval = REFRESH_CHANNEL_UPDATE_INTERVAL, receiver = self, message = BroadcastChannelUpdate(PeriodicRefresh))
       goto(NORMAL) using DATA_NORMAL(commitments.copy(remoteNextCommitInfo = Right(nextPerCommitmentPoint)), shortChannelId, buried = false, None, initialChannelUpdate, None, None) storing()
@@ -898,7 +898,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         // we need to re-announce this shortChannelId
         context.system.eventStream.publish(ShortChannelIdAssigned(self, d.channelId, shortChannelId, Some(d.shortChannelId)))
         // we re-announce the channelUpdate for the same reason
-        Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.localCommit.spec.totalFunds, enable = Helpers.aboveReserve(d.commitments))
+        Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.capacity.toMilliSatoshi, enable = Helpers.aboveReserve(d.commitments))
       } else d.channelUpdate
       val localAnnSigs_opt = if (d.commitments.announceChannel) {
         // if channel is public we need to send our announcement_signatures in order to generate the channel_announcement
@@ -945,14 +945,14 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
     case Event(c: CMD_UPDATE_RELAY_FEE, d: DATA_NORMAL) =>
       log.info("updating relay fees: prevFeeBaseMsat={} nextFeeBaseMsat={} prevFeeProportionalMillionths={} nextFeeProportionalMillionths={}", d.channelUpdate.feeBaseMsat, c.feeBase, d.channelUpdate.feeProportionalMillionths, c.feeProportionalMillionths)
-      val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, c.feeBase, c.feeProportionalMillionths, d.commitments.localCommit.spec.totalFunds, enable = Helpers.aboveReserve(d.commitments))
+      val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, c.feeBase, c.feeProportionalMillionths, d.commitments.capacity.toMilliSatoshi, enable = Helpers.aboveReserve(d.commitments))
       // we use GOTO instead of stay because we want to fire transitions
       c.replyTo ! RES_SUCCESS(c, d.channelId)
       goto(NORMAL) using d.copy(channelUpdate = channelUpdate) storing()
 
     case Event(BroadcastChannelUpdate(reason), d: DATA_NORMAL) =>
       val age = System.currentTimeMillis.milliseconds - d.channelUpdate.timestamp.seconds
-      val channelUpdate1 = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.localCommit.spec.totalFunds, enable = Helpers.aboveReserve(d.commitments))
+      val channelUpdate1 = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.capacity.toMilliSatoshi, enable = Helpers.aboveReserve(d.commitments))
       reason match {
         case Reconnected if Announcements.areSame(channelUpdate1, d.channelUpdate) && age < REFRESH_CHANNEL_UPDATE_INTERVAL =>
           // we already sent an identical channel_update not long ago (flapping protection in case we keep being disconnected/reconnected)
@@ -976,7 +976,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       // if we have pending unsigned htlcs, then we cancel them and advertise the fact that the channel is now disabled
       val d1 = if (d.commitments.localChanges.proposed.collectFirst { case add: UpdateAddHtlc => add }.isDefined) {
         log.info(s"updating channel_update announcement (reason=disabled)")
-        val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.localCommit.spec.totalFunds, enable = false)
+        val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.capacity.toMilliSatoshi, enable = false)
         d.commitments.localChanges.proposed.collect {
           case add: UpdateAddHtlc => relayer ! RES_ADD_SETTLED(d.commitments.originChannels(add.id), add, HtlcResult.Disconnected(channelUpdate))
         }
@@ -1445,7 +1445,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
     case Event(c: CMD_UPDATE_RELAY_FEE, d: DATA_NORMAL) =>
       log.info(s"updating relay fees: prevFeeBaseMsat={} nextFeeBaseMsat={} prevFeeProportionalMillionths={} nextFeeProportionalMillionths={}", d.channelUpdate.feeBaseMsat, c.feeBase, d.channelUpdate.feeProportionalMillionths, c.feeProportionalMillionths)
-      val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, c.feeBase, c.feeProportionalMillionths, d.commitments.localCommit.spec.totalFunds, enable = false)
+      val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, c.feeBase, c.feeProportionalMillionths, d.commitments.capacity.toMilliSatoshi, enable = false)
       // we're in OFFLINE state, we don't broadcast the new update right away, we will do that when next time we go to NORMAL state
       c.replyTo ! RES_SUCCESS(c, d.channelId)
       stay using d.copy(channelUpdate = channelUpdate) storing()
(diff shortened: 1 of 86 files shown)


engenegr:
I would use eclair nodes with such plugins at least just for connecting them "virtually". It would motivate me to launch at least two new nodes.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -46,6 +46,40 @@ case class RemoteCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, r
...
+    Sphinx.PaymentPacket.peel(nodeSecret, add.paymentHash, add.onionRoutingPacket) match {
+      case Right(Sphinx.DecryptedPacket(_, _, sharedSecret)) =>
+        val reason = cmd.reason match {
+          case Left(forwarded) => Sphinx.FailurePacket.wrap(forwarded, sharedSecret)
+          case Right(failure) => Sphinx.FailurePacket.create(sharedSecret, failure)
+        }
+        Success(UpdateFailHtlc(channelId, cmd.id, reason))
+      case Left(_) => Failure(CannotExtractSharedSecret(channelId, add))
+    }
+  }
```
t-bast, line 80: I don't think this makes sense as a method inside the trait.
It's probably better to make this a static helper function inside the `Commitments` object, WDYT?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -87,16 +88,16 @@ object ChannelRelayer {
...
-              val channels = if (enabledOnly) {
-                channelUpdates.values.filter(o => Announcements.isEnabled(o.channelUpdate.channelFlags))
-              } else {
-                channelUpdates.values
-              }
+              val channels = for {
+                out @ OutgoingChannel(_, channelUpdate, _: Commitments) <- channelUpdates.values // We are only interested in normal channel Commitments here
+                if !enabledOnly || (enabledOnly && Announcements.isEnabled(channelUpdate.channelFlags))
+              } yield out
+
```
t-bast, line 95: Why do you filter it out here? It feels very weird to filter here because of something external from eclair-core.
Is that filter really necessary? If so, can't it be done outside of here (in the plugins that enable HC)?

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/GUIUpdater.scala
```diff
@@ -79,16 +79,16 @@ class GUIUpdater(mainController: MainController) extends Actor with ActorLogging
       runInGuiThread(() => mainController.channelBox.getChildren.addAll(root))
       context.become(main(m + (channel -> channelPaneController)))
 
-    case ChannelRestored(channel, peer, remoteNodeId, isFunder, channelId, currentData) =>
+    case ChannelRestored(channel, peer, remoteNodeId, isFunder, currentData: Commitments) => // We are only interested in normal channel Commitments here
```
t-bast, line 82: Same, is it really necessary? Do HC need to emit a `ChannelRestored` event in the first place?

t-bast (commented):
It's quite a small code-change, which simply adds a layer of abstraction, so I think it's acceptable to go in that direction.
We'll need @pm47 to also agree on that though.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
btcontract: I did this because this chunk can be directly reused in HC while `Commitments` object is related to specifically normal channel commitments (?). Maybe `ChannelCommitments` object?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
btcontract: It's done this way to preserve existing behavior, if HC is present in system then its commitments data will be injected into `ChannelRelayer` maps and will be returned alongside with normal channel commitments unless filtered. 

This may or may not be desired so I've decided to take a safe route and filter out anything except what core Eclair already understands.

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/GUIUpdater.scala
btcontract: Like with `ChannelRelayer`, HC also needs to inject itself into `Register` which listens to `ChannelRestored` event. This is again an approach where anything except stuff known at core level is ignored.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: Actually a good place to put that helper function could be in `PaymentPacket.scala`, in the `OutgoingPacket` object.
I would also rename it `buildHtlcFailure` to be more consistent with that file's functions.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
t-bast: Would a user running a node with hosted channels be interested in getting them in the `channels` API?
It seems to me that yes, it would be interesting for them, wouldn't it?

I think it's weird to filter here for things eclair-core doesn't know about that will be built on top: if a user adds a plugin to eclair-core that adds hosted channels, that user knows about them and can decide whether to filter them or not, so the decision should be taken by the plugin?

The only reason I see for filtering them out explicitly is if adding hosted channels and having them returned here would break something that eclair-core does; you should check what actors depend on this channels list and evaluate whether that's the case.

btcontract: > Would a user running a node with hosted channels be interested in getting them in the channels API?
It seems to me that yes, it would be interesting for them, wouldn't it?

the plan for this in particular is to have a separate plugin-specific API (similar to keysend plugin which establishes its own API). 

But in general, yes, there is no reason to filter out HC in particular because they are designed to be indistinguishable from normal channels to the rest of the system. 

For instance, if user wants to send (or receive) a payment then not filtering HC out would make it a candidate for a first channel in a route when sending or another route hint when receiving, and it's perfectly fine. 

I guess this is a precautionary measure if someone decides to add some other type of injectable commitments via plugin without taking care of full compatibility, but maybe this is unjustified and this part should be well documented instead?

t-bast: Yes, I think that for now we shouldn't do any filtering.
It will be up to future plugin developers to figure out if they need their new channel type to be filtered out or not; if it needs to be filtered out, at that point a PR should be opened to eclair-core to introduce a filtering mechanism, or a mechanism in the plugin itself should be found.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -193,7 +193,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
     case Event(INPUT_RESTORED(data), _) =>
       log.info(s"restoring channel channelId=${data.channelId}")
-      context.system.eventStream.publish(ChannelRestored(self, peer, remoteNodeId, data.commitments.localParams.isFunder, data.channelId, data))
+      context.system.eventStream.publish(ChannelRestored(self, data.commitments.channelId, peer, remoteNodeId, data.commitments.localParams.isFunder, data.commitments))
```
t-bast, line 196:
```suggestion
      context.system.eventStream.publish(ChannelRestored(self, data.channelId, peer, remoteNodeId, data.commitments.localParams.isFunder, data.commitments))
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -221,8 +221,8 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
             case None =>
               // in all other cases we need to be ready for any type of closing
               // TODO: should we wait for an acknowledgment from the watcher?
-              blockchain ! WatchSpent(self, data.commitments.commitInput.outPoint.txid, data.commitments.commitInput.outPoint.index.toInt, data.commitments.commitInput.txOut.publicKeyScript, BITCOIN_FUNDING_SPENT)
-              blockchain ! WatchLost(self, data.commitments.commitInput.outPoint.txid, nodeParams.minDepthBlocks, BITCOIN_FUNDING_LOST)
+              blockchain ! WatchSpent(self, data.commitments.fundingTxId, data.commitments.commitInput.outPoint.index.toInt, data.commitments.commitInput.txOut.publicKeyScript, BITCOIN_FUNDING_SPENT)
+              blockchain ! WatchLost(self, data.commitments.fundingTxId, nodeParams.minDepthBlocks, BITCOIN_FUNDING_LOST)
```
t-bast, line 225: Why did you change that? I think the previous code was less confusing since we still need to refer to `data.commitments.commitInput` in the other parameters...and same for the other places in this file where that code was changed.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -25,6 +25,7 @@ import akka.actor.typed.scaladsl.Behaviors
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.channel._
 import fr.acinq.eclair.payment.IncomingPacket
+import fr.acinq.eclair.payment.relay.Relayer.OutgoingChannel
```
t-bast, line 28: nit: remove?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Auditor.scala
```diff
@@ -98,7 +98,7 @@ class Auditor(nodeParams: NodeParams) extends Actor with ActorLogging {
         case _: RecoveryClose => "recovery"
         case _: RevokedClose => "revoked"
       }
-      db.add(ChannelLifecycleEvent(e.channelId, e.commitments.remoteParams.nodeId, e.commitments.commitInput.txOut.amount, e.commitments.localParams.isFunder, !e.commitments.announceChannel, event))
+      db.add(ChannelLifecycleEvent(e.commitments.channelId, e.commitments.remoteParams.nodeId, e.commitments.commitInput.txOut.amount, e.commitments.localParams.isFunder, !e.commitments.announceChannel, event))
```
t-bast, line 101: nit: why that change?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -99,7 +140,17 @@ case class Commitments(channelVersion: ChannelVersion,
...
+
+  lazy val localNodeId: PublicKey = localParams.nodeId
+
+  lazy val remoteNodeId: PublicKey = remoteParams.nodeId
+
+  lazy val capacity: MilliSatoshi = localCommit.spec.totalFunds
+
+  lazy val fundingTxId: ByteVector32 = commitInput.outPoint.txid
+
+  lazy val announceChannel: Boolean = (channelFlags & 0x01) != 0
```
t-bast, line 153: nit: I don't see a reason to make these `lazy val`, they will always be computed so let's just make them `val`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -355,30 +390,21 @@ object Commitments {
...
-              case Left(forwarded) => Sphinx.FailurePacket.wrap(forwarded, sharedSecret)
-              case Right(failure) => Sphinx.FailurePacket.create(sharedSecret, failure)
-            }
-            val fail = UpdateFailHtlc(commitments.channelId, cmd.id, reason)
-            val commitments1 = addLocalProposal(commitments, fail)
-            Success((commitments1, fail))
-          case Left(_) => Failure(CannotExtractSharedSecret(commitments.channelId, htlc))
-        }
+        val failTry: Try[UpdateFailHtlc] = OutgoingPacket.buildHtlcFailure(nodeSecret, cmd, htlc)
+        failTry.map(updateFail => (addLocalProposal(commitments, updateFail), updateFail))
```
t-bast, line 407: The name is very misleading, what about:

```suggestion
        OutgoingPacket.buildHtlcFailure(nodeSecret, cmd, htlc).map(fail => (addLocalProposal(commitments, fail), fail))
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -46,6 +47,32 @@ case class RemoteCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, r
...
+  def availableBalanceForReceive: MilliSatoshi
+
+  def availableBalanceForSend: MilliSatoshi
+
+  def originChannels: Map[Long, Origin]
+
+  def channelId: ByteVector32
+
+  def announceChannel: Boolean
+}
```
t-bast, line 74: Is this really all you need for hosted channels?
To help my understanding, are hosted channels using a different channel FSM than the one in eclair-core (I'd guess yes)?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -769,7 +769,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           }
           if (d.commitments.availableBalanceForSend != commitments1.availableBalanceForSend) {
             // we send this event only when our balance changes
-            context.system.eventStream.publish(AvailableBalanceChanged(self, d.channelId, d.shortChannelId, commitments1))
+            context.system.eventStream.publish(AvailableBalanceChanged(self, commitments1.channelId, d.shortChannelId, commitments1))
```
t-bast, line 772:
```suggestion
            context.system.eventStream.publish(AvailableBalanceChanged(self, d.channelId, d.shortChannelId, commitments1))
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1362,7 +1362,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       closingType_opt match {
         case Some(closingType) =>
           log.info(s"channel closed (type=$closingType)")
-          context.system.eventStream.publish(ChannelClosed(self, d.channelId, closingType, d.commitments))
+          context.system.eventStream.publish(ChannelClosed(self, d.commitments.channelId, closingType, d.commitments))
```
t-bast, line 1365:
```suggestion
          context.system.eventStream.publish(ChannelClosed(self, d.channelId, closingType, d.commitments))
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
```diff
@@ -74,7 +74,7 @@ class ChannelRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("a
 
   /** we parse logs to make sure the listener has received the event and prevent race conditions */
   def publishUpdate(e: LocalChannelUpdate) = {
-    LoggingTestKit.debug(s"updating local channel info for channelId=${e.channelId} shortChannelId=${e.shortChannelId}").expect {
+    LoggingTestKit.debug(s"updating local channel info for channelId=${e.commitments.channelId} shortChannelId=${e.shortChannelId}").expect {
```
t-bast, line 77: Please revert (and same elsewhere, why did you add these unrelated and unnecessary changes?)

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
btcontract: Originally it's because of this place: https://github.com/ACINQ/eclair/pull/1542/files#diff-86ff21f0015ed1b8252f65273dd4d06aR108 It accepts `ChannelCommitments` trait but then tries to access a funding txid which `ChannelCommitments` trait does not have (because HC has no on-chain funding). The plan to set it to all zeros in HC data. Then I've decided to apply it elsewhere since it's already defined.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
btcontract: My dad, just did not pay attention to what exactly was there before, will revert.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
btcontract: Maybe another way to do it is to add something more general, like "fundingDescription" field which would resolve to txid in normal channel and, say, human readable "No on-chain funding" in HC?

t-bast: Honestly the GUI is just a demo tool, if that's the only place where this is used, don't bother, it should not make eclair-core more complex

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
```diff
@@ -397,8 +397,9 @@ object PaymentPacketSpec {
   }
 
   def makeCommitments(channelId: ByteVector32, testAvailableBalanceForSend: MilliSatoshi = 50000000 msat, testAvailableBalanceForReceive: MilliSatoshi = 50000000 msat): Commitments = {
-    val params = LocalParams(null, null, null, null, null, null, null, 0, isFunder = true, null, None, null)
-    new Commitments(ChannelVersion.STANDARD, params, null, 0.toByte, null, null, null, null, 0, 0, Map.empty, null, null, null, channelId) {
+    val localParams = LocalParams(null, null, null, null, null, null, null, 0, isFunder = true, null, None, null)
+    val remoteParams = RemoteParams(randomKey.publicKey, null, null, null, null, null, maxAcceptedHtlcs = 0, null, null, null, null, null, null)
```
btcontract, line 401: Since `ChannelCommitments` has a `val remoteNodeId: PublicKey = remoteParams.nodeId` now it would throw if `remoteParams` is `null`.

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/GUIUpdater.scala
```diff
@@ -79,16 +79,16 @@ class GUIUpdater(mainController: MainController) extends Actor with ActorLogging
       runInGuiThread(() => mainController.channelBox.getChildren.addAll(root))
       context.become(main(m + (channel -> channelPaneController)))
 
-    case ChannelRestored(channel, peer, remoteNodeId, isFunder, channelId, currentData) =>
+    case ChannelRestored(channel, channelId, peer, remoteNodeId, isFunder, currentData: Commitments) => // We are specifically interested in normal Commitments with funding txid here
```
btcontract, line 82: I've put this filtering back as an easiest fix.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
t-bast: Good catch

t-bast (commented):
ACK https://github.com/ACINQ/eclair/pull/1542/commits/cd786406fb96b7f1248ea6e12a6b5ca572189240
Let's see what @pm47 thinks about this change.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -46,6 +47,28 @@ case class RemoteCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, r
 case class WaitingForRevocation(nextRemoteCommit: RemoteCommit, sent: CommitSig, sentAfterLocalCommitIndex: Long, reSignAsap: Boolean = false)
 // @formatter:on
 
+trait ChannelCommitments {
```
pm47, line 50: What is the rationale for the naming? The following may sound a bit java-ish, but at least it conveys the information that `Commitments <: AbstractCommitments`.

```suggestion
trait AbstractCommitments {
```

pm47 (commented):
While I'm not sure about the final goal, this change is rather small and doesn't introduce much complexity, so I guess why not.

Have you considered introducing an abstraction in `LocalChannelUpdate` instead? That's what the router and the relayer are receiving.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
btcontract: I tried to convey it's something, well, abstract so this also sounds good.

btcontract:
> Have you considered introducing an abstraction in LocalChannelUpdate instead?

I don't understand how this could work, can you elaborate here?

pm47:
>> Have you considered introducing an abstraction in LocalChannelUpdate instead?

> I don't understand how this could work, can you elaborate here?

Your piece of code (that manages hosted channels) need to emit `LocalChannelUpdate` right? But you don't have a `Commitment`, which is why you had to do this PR that allows creating a custom `Commitment`. Instead, you could change `LocalChannelUpdate` to contain the information that the relayer and the router need, without touching the existing `Commitment`. My guess is that you would end up with something close to the `Relayer.OutgoingChannel` class.

Basically, replace `Commitments` by `Relayer.OutgoingChannel` in `LocalChannelUpdate` and see where it breaks.

btcontract:
I'd say this approach is less elegant and more laborious because:

- HC has a commitments object which is similar to normal channel commitments in many ways, the main difference is in HC peers commit to new state by cross-signing it with their node id instead of producing a tx, so it makes sense to have a trait. 

- There will need to be more changes made in `Register` and `Relayer`, I guess another abstraction to treat HC and normal channel data in a same way inside of them while with common trait it comes for free.

- Besides `Register` and `Relayer` there are some places where `AbstractCommitments`'s methods are used (such as `getIncomingHtlcCrossSigned` in PostRestartCleaner or `localNodeId`/`remoteNodeId` in router). Having them in common trait is again simpler than custom workarounds for each such case.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -75,8 +75,6 @@ final case class CommitmentSpec(htlcs: Set[DirectedHtlc], feeratePerKw: FeerateP
   def findIncomingHtlcById(id: Long): Option[IncomingHtlc] = htlcs.collectFirst { case htlc: IncomingHtlc if htlc.add.id == id => htlc }
 
   def findOutgoingHtlcById(id: Long): Option[OutgoingHtlc] = htlcs.collectFirst { case htlc: OutgoingHtlc if htlc.add.id == id => htlc }
-
-  val totalFunds: MilliSatoshi = toLocal + toRemote + htlcs.toSeq.map(_.add.amountMsat).sum
```
btcontract, line 79: This was duplicating what `val capacity: Satoshi = commitInput.txOut.amount` now does so I've decided to remove it.

t-bast:
Looking good, we'd like to release a new version of eclair soon, we'll include this PR right after the release, is that ok?

btcontract:
Yes, it's fine since I'm interested in this and https://github.com/ACINQ/eclair/pull/1547 anyway.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -249,4 +250,15 @@ object OutgoingPacket {
     CMD_ADD_HTLC(replyTo, firstAmount, paymentHash, firstExpiry, onion.packet, Origin.Hot(replyTo, upstream), commit = true) -> onion.sharedSecrets
   }
 
+  def buildHtlcFailure(nodeSecret: PrivateKey, cmd: CMD_FAIL_HTLC, add: UpdateAddHtlc): Try[UpdateFailHtlc] = {
```
btcontract, line 253: turns out this method gets inlined and is not visible to plugin, `@noinline` does not help. Are there any other ways to prevent inlining? Otherwise it makes no sense to have it and code can be copy-pasted in plugin.

t-bast: That sounds very surprising...why would this one be inlined and not all others? Are you sure you're referencing the right jar?

btcontract: yes, I'm sure since `AbstractCommitments` are visible. I've also clicked "decompile to java" in `Commitments.scala` and the method body, not a function call can be seen in `sendFail`.

t-bast: Damn this is weird...and are you able to see the other functions in the `OutgoingPacket` object?

t-bast: The only explanation I see is that it's used in only one place, maybe if you add a unit test for it it won't be inlined?

btcontract: Adding a method to tests solved an issue.

t-bast (approved):
LGTM, if @pm47 is happy with that change we can merge it.

btcontract:
@pm47 is there an issue here?

pm47 (approved):

PR #1543, opened 2020-09-28
Make all commands extend `Has*ReplyToCommand`

pm47:
Commands have to be of type `HasReplyToCommand` or `HasOptionalReplyToCommand`. The latter is mostly for commands which don't need replies except in tests.

With this change, there is no more `replying` in `Channel`, and a lot of `sender.send()` have been removed from tests, in preparation to full akka typed.

Supersedes #1533.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/CommandCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/CommandCodecs.scala
@@ -16,6 +16,7 @@
 
 package fr.acinq.eclair.wire
 
+import akka.actor.ActorRef
 import fr.acinq.eclair.channel._
 import fr.acinq.eclair.wire.CommonCodecs._
 import fr.acinq.eclair.wire.FailureMessageCodecs.failureMessageCodec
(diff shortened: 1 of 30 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/CommandCodecs.scala
```diff
@@ -27,18 +28,21 @@ object CommandCodecs {
   val cmdFulfillCodec: Codec[CMD_FULFILL_HTLC] =
     (("id" | int64) ::
       ("r" | bytes32) ::
-      ("commit" | provide(false))).as[CMD_FULFILL_HTLC]
+      ("commit" | provide(false)) ::
+      ("replytTo_opt" | provide(Option.empty[ActorRef]))).as[CMD_FULFILL_HTLC]
```
t-bast, line 32: nit (and same for the 2 others below):
```suggestion
      ("replyTo_opt" | provide(Option.empty[ActorRef]))).as[CMD_FULFILL_HTLC]
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/b/WaitForFundingCreatedInternalStateSpec.scala
```diff
@@ -68,7 +68,10 @@ class WaitForFundingCreatedInternalStateSpec extends TestKitBaseClass with Fixtu
 
   test("recv CMD_CLOSE") { f =>
     import f._
-    alice ! CMD_CLOSE(None)
+    val sender = TestProbe()
+    val c = CMD_CLOSE(sender.ref, None)
```
t-bast, line 72: Should we have one of these tests test the fallback to `sender` when `ActorRef.noSender` is provided?

```scala
val sender = TestProbe()
val c = CMD_CLOSE(ActorRef.noSender, None)
sender.send(alice, c)
sender.expectMsg(RES_SUCCESS(c, ByteVector32.Zeroes))
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/f/ShutdownStateSpec.scala
```diff
@@ -328,54 +326,52 @@ class ShutdownStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike wit
...
     bob2alice.forward(alice)
-    sender.send(bob, CMD_SIGN)
+    bob ! CMD_SIGN(replyTo_opt = Some(sender.ref))
+    sender.expectMsgType[RES_SUCCESS[CMD_SIGN]]
     bob2alice.expectMsgType[CommitSig]
     bob2alice.forward(alice)
     alice2bob.expectMsgType[RevokeAndAck]
     alice2bob.forward(bob)
-    sender.send(alice, CMD_SIGN)
-    sender.expectMsgType[RES_SUCCESS[CMD_SIGN.type]]
```
t-bast, line 340: Good catch, this wasn't necessary!

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/b/WaitForFundingCreatedInternalStateSpec.scala
pm47: Done with 2bad181ab... and found a bug!

eclair-core/src/main/scala/fr/acinq/eclair/wire/CommandCodecs.scala
pm47: Done with b80e0b50e.

t-bast (dismissed):

t-bast (approved):

PR #1545, opened 2020-09-30
Add blockchain watchdog

t-bast:
Introduce a blockchain watchdog, that fetches headers from multiple sources in order to detect whether we're being eclipsed.

The current blockchain sources we use are:

- blockchainheaders.net (headers over DNS)
- blockstream.info
- mempool.space
- blockcypher.com

We can add more sources later if needed.


t-bast:
I'm not sure how I should instantiate my top-level actor (`BlockchainWatchdog`).
It looks like the simplest way is to do it as child of an existing actor (does it make sense as child of `ZmqWatcher`?) otherwise I think I need to create a full akka typed ActorSystem in `Setup.scala`.

pm47:
> I'm not sure how I should instantiate my top-level actor (BlockchainWatchdog).
It looks like the simplest way is to do it as child of an existing actor (does it make sense as child of ZmqWatcher?) otherwise I think I need to create a full akka typed ActorSystem in Setup.scala.

I believe a child of `ZmqWatcher` would make sense. Or maybe could create a `MultiSourceWatcher` that would itself be a watcher, and would instantiate a `BlockchainWatchdog` and `ZmqWatcher`?

pm47:
You should probably add https://mempool.space/api.

pm47 (commented):
I think that instead of looking for a `blockCountDelta` in the future, we should simply compare tips. In the case of bitcoinheaders.net which doesn't have this feature, we can look for `n + 10` blocks. Maybe we are even more late than that but it's enough to raise the alarm.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
```diff
@@ -0,0 +1,139 @@
...
+      }.toClassic
+      Behaviors.receiveMessage {
+        case CheckLatestHeaders(replyTo) => chainHash match {
+          case Block.LivenetGenesisBlock.hash =>
+            // We try to get the next 10 blocks; if we're late by more than 10 blocks, this is bad, no need to even look further.
+            (currentBlockCount until currentBlockCount + 10).foreach(blockCount => {
+              val hostname = s"$blockCount.${blockCount / 10000}.bitcoinheaders.net"
+              IO(Dns)(context.system.classicSystem).tell(DnsProtocol.resolve(hostname, DnsProtocol.Ip(ipv4 = false, ipv6 = true)), dnsAdapters)
+            })
+            collect(replyTo, currentBlockCount, Set.empty, 10)
```
pm47, line 61: Why don't we simply try to get the n+10th block?

t-bast: If we do that and it doesn't exist, we can't properly update the metric to know how many blocks we're missing. I think it's useful to have that metric be more precise than we're more than 10 blocks late vs we're less than 10 blocks late, even only for consistency with the other watchdogs

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
```diff
@@ -0,0 +1,139 @@
...
+            // Headers over DNS is only supported for mainnet.
+            Behaviors.stopped
+        }
+        case _ => Behaviors.unhandled
+      }
+    }
+  }
+
+  private def collect(replyTo: ActorRef[BlockchainWatchdog.LatestHeaders], currentBlockCount: Long, received: Set[BlockHeaderAt], remaining: Int): Behavior[Command] = {
+    Behaviors.setup { context =>
```
pm47, line 72: I thought `Behaviors.setup` was only to be used when we are, you know, _setting up_ the actor, but it turns out we can use it in later `Behaviors` too. Ha!

> setup is typically used as the outer most behavior when spawning an actor, but it can also be returned as the next behavior when processing a message or signal. In that case it will be started immediately after it is returned, i.e. next message will be processed by the started behavior.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
```diff
@@ -0,0 +1,139 @@
...
+        context.messageAdapter[DnsProtocol.Resolved](WrappedDnsResolved)
+        context.messageAdapter[Status.Failure](f => WrappedDnsFailed(f.cause))
+      }.toClassic
+      Behaviors.receiveMessage {
+        case CheckLatestHeaders(replyTo) => chainHash match {
+          case Block.LivenetGenesisBlock.hash =>
+            // We try to get the next 10 blocks; if we're late by more than 10 blocks, this is bad, no need to even look further.
+            (currentBlockCount until currentBlockCount + 10).foreach(blockCount => {
+              val hostname = s"$blockCount.${blockCount / 10000}.bitcoinheaders.net"
+              IO(Dns)(context.system.classicSystem).tell(DnsProtocol.resolve(hostname, DnsProtocol.Ip(ipv4 = false, ipv6 = true)), dnsAdapters)
```
pm47, line 59: There doesn't seem to be a _typed_ version of DNS resolution 😢

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
```diff
@@ -0,0 +1,139 @@
...
+  val Source = "bitcoinheaders.net"
+
+  // @formatter:off
+  sealed trait Command
+  case class CheckLatestHeaders(replyTo: ActorRef[BlockchainWatchdog.LatestHeaders]) extends Command
+  private case class WrappedDnsResolved(response: DnsProtocol.Resolved) extends Command
+  private case class WrappedDnsFailed(cause: Throwable) extends Command
+  // @formatter:on
+
+  def apply(chainHash: ByteVector32, currentBlockCount: Long): Behavior[Command] = {
```
pm47, line 47: I think you are _almost_ at the point where it makes sense to use the [private class](https://doc.akka.io/docs/akka/current/typed/style-guide.html#passing-around-too-many-parameters) pattern.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
```diff
@@ -0,0 +1,139 @@
...
+        case CheckLatestHeaders(replyTo) => chainHash match {
+          case Block.LivenetGenesisBlock.hash =>
+            // We try to get the next 10 blocks; if we're late by more than 10 blocks, this is bad, no need to even look further.
+            (currentBlockCount until currentBlockCount + 10).foreach(blockCount => {
+              val hostname = s"$blockCount.${blockCount / 10000}.bitcoinheaders.net"
+              IO(Dns)(context.system.classicSystem).tell(DnsProtocol.resolve(hostname, DnsProtocol.Ip(ipv4 = false, ipv6 = true)), dnsAdapters)
+            })
+            collect(replyTo, currentBlockCount, Set.empty, 10)
+          case _ =>
+            // Headers over DNS is only supported for mainnet.
```
pm47, line 63: Would have made that a debug comment.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/ExplorerApi.scala
```diff
@@ -0,0 +1,201 @@
...
+  case class BlockstreamExplorer() extends Esplora {
+    override val name = "blockstream.info"
+    override val baseUris = Map(
+      Block.TestnetGenesisBlock.hash -> uri"https://blockstream.info/testnet/api",
+      Block.LivenetGenesisBlock.hash -> uri"https://blockstream.info/api"
+    )
+  }
+
+  /** Query https://mempool.space/ to fetch block headers. */
+  case class MempoolSpaceExplorer() extends Esplora {
```
pm47, line 193: Oh, Mempool.space is based on Esplora, didn't know.

pm47 (approved):
I have just suggestions, but we can go with this version if you want.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
t-bast: Yeah I couldn't find it either, it's sad...hopefully they release that someday!

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
t-bast: I'll look into it

pm47: See `ChannelRelay`/`NodeRelay` for inspiration 😄

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1545/commits/50b74e84b38396d085ca29cc74d60b91ab179d9d

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
t-bast: TBH I think I don't have that many arguments, so let's keep it like that :)

pm47: Fair enough.

pm47 (approved):

PR #1547, opened 2020-10-01
Plugin HTLCs on restart

btcontract:
This enriches plugins with ability to provide their own broken and pending HTLCs on restart which are then used in `PostRestartHtlcCleaner` methods and events.

This is needed in particular as a final piece to enable HC entirely as a plugin, it builds upon plugin messaging and https://github.com/ACINQ/eclair/pull/1542.

Without this functionality it's not possible to correctly handle broken and pending HTLCs which were somehow routed through plugins (in case of HC that would be routings like `peer -> normal/HC -> normal/HC -> peer`).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
@@ -79,7 +79,7 @@ class PostRestartHtlcCleaner(nodeParams: NodeParams, register: ActorRef, initial
       val acked = brokenHtlcs.notRelayed
         .filter(_.add.channelId == data.channelId) // only consider htlcs coming from this channel
         .filter {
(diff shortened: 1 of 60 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -71,8 +71,8 @@ class PostRestartHtlcCleaner(nodeParams: NodeParams, register: ActorRef, initial
       Metrics.PendingNotRelayed.update(brokenHtlcs.notRelayed.size)
       Metrics.PendingRelayedOut.update(brokenHtlcs.relayedOut.keySet.size)
       // Once we've loaded the channels and identified broken HTLCs, we let other components know they can proceed.
-      Try(initialized.map(_.success(Done)))
       context become main(brokenHtlcs)
+      sender ! Done
```
btcontract, line 75: Does this make sense? Since now PostRestartHtlcCleaner is initialized by a message it seems reasonable use ask/tell.

btcontract:
I'll provide a more clearly structured PR.

PR #1548, opened 2020-10-01
Separate configuration for peer-connection

pm47:
Features are now provided by the `switchboard`, in response to the
`PeerConnection.Authenticated` message.

The `switchboard` will also decide whether or not we sync with that
peer, depending on the `syncWhiteList` configuration.

This is an alternative to #1361. The `features`/`override-features`/`sync-whitelist` remain top-level parameters in `eclair.conf`, which was one of the main issues with #1361.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -123,11 +123,14 @@ eclair {
 
   revocation-timeout = 20 seconds // after sending a commit_sig, we will wait for at most that duration before disconnecting
 
-  auth-timeout = 10 seconds // will disconnect if connection authentication doesn't happen within that timeframe
-  init-timeout = 10 seconds // will disconnect if initialization doesn't happen within that timeframe
-  ping-interval = 30 seconds
-  ping-timeout = 10 seconds // will disconnect if peer takes longer than that to respond
-  ping-disconnect = true // disconnect if no answer to our pings
+  peer-connection {
+    auth-timeout = 10 seconds // will disconnect if connection authentication doesn't happen within that timeframe
+    init-timeout = 10 seconds // will disconnect if initialization doesn't happen within that timeframe
+    ping-interval = 30 seconds
+    ping-timeout = 10 seconds // will disconnect if peer takes longer than that to respond
+    ping-disconnect = true // disconnect if no answer to our pings
+  }
+
   auto-reconnect = true
   initial-random-reconnect-delay = 5 seconds // we add a random delay before the first reconnection attempt, capped by this value
   max-reconnect-interval = 1 hour // max interval between two reconnection attempts, after the exponential backoff period
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -20,6 +20,7 @@ import java.net.InetSocketAddress
 
 import akka.actor.{ActorRef, FSM, OneForOneStrategy, PoisonPill, Props, SupervisorStrategy, Terminated}
 import akka.event.Logging.MDC
+import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.Logs.LogCategory
 import fr.acinq.eclair.crypto.Noise.KeyPair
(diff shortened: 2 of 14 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -123,11 +123,14 @@ eclair {
...
+    auth-timeout = 10 seconds // will disconnect if connection authentication doesn't happen within that timeframe
+    init-timeout = 10 seconds // will disconnect if initialization doesn't happen within that timeframe
+    ping-interval = 30 seconds
+    ping-timeout = 10 seconds // will disconnect if peer takes longer than that to respond
+    ping-disconnect = true // disconnect if no answer to our pings
+  }
+
   auto-reconnect = true
   initial-random-reconnect-delay = 5 seconds // we add a random delay before the first reconnection attempt, capped by this value
   max-reconnect-interval = 1 hour // max interval between two reconnection attempts, after the exponential backoff period
```
t-bast, line 136: Shouldn't these three also be in the peer connection section?

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -530,10 +535,9 @@ object PeerConnection {
    *         - true if there is a filter and msg has no timestamp, or has one that matches the filter
    *         - false otherwise
    */
-  def timestampInRange(nodeParams: NodeParams, msg: RoutingMessage, origins: Set[GossipOrigin], gossipTimestampFilter_opt: Option[GossipTimestampFilter]): Boolean = {
+  def timestampInRange(msg: RoutingMessage, origins: Set[GossipOrigin], gossipTimestampFilter_opt: Option[GossipTimestampFilter]): Boolean = {
     // For our own gossip, we should ignore the peer's timestamp filter.
     val isOurGossip = msg match {
-      case n: NodeAnnouncement if n.nodeId == nodeParams.nodeId => true
```
t-bast, line 536: We're sure this is now redundant with the line below and can safely be removed?

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
```diff
@@ -63,27 +63,27 @@ class PeerConnectionSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike wi
     val remoteNodeId = Bob.nodeParams.nodeId
 
     import com.softwaremill.quicklens._
-    val aliceParams = TestConstants.Alice.nodeParams
-      .modify(_.syncWhitelist).setToIf(test.tags.contains("sync-whitelist-bob"))(Set(remoteNodeId))
-      .modify(_.syncWhitelist).setToIf(test.tags.contains("sync-whitelist-random"))(Set(randomKey.publicKey))
+    val aliceParams = TestConstants.Alice.nodeParams//.peerConnectionConf
+//      .modify(_.syncWhitelist).setToIf(test.tags.contains("sync-whitelist-bob"))(Set(remoteNodeId))
+//      .modify(_.syncWhitelist).setToIf(test.tags.contains("sync-whitelist-random"))(Set(randomKey.publicKey))
```
t-bast, line 68: nit: needs a bit of clean-up?

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
```diff
@@ -151,7 +151,7 @@ class PeerConnectionSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike wi
     probe.watch(transport.ref)
     probe.send(peerConnection, PeerConnection.PendingAuth(connection.ref, Some(remoteNodeId), address, origin_opt = Some(origin.ref), transport_opt = Some(transport.ref)))
     transport.send(peerConnection, TransportHandler.HandshakeCompleted(remoteNodeId))
-    probe.send(peerConnection, PeerConnection.InitializeConnection(peer.ref))
+    probe.send(peerConnection, PeerConnection.InitializeConnection(peer.ref, nodeParams.chainHash, nodeParams.features, doSync = true))
```
t-bast, line 154: We're not testing `doSync = false` anywhere, maybe one of these tests should set it to `false`?

eclair-core/src/main/resources/reference.conf
pm47: No, those are related to the reconnection task. The scope of peer-connection is the handling of an already established connection (incoming or outgoing).

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
pm47: Yes, it was already removed in feature branches.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
pm47: Indeed!

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
pm47: Yes we are in the `connect` method, which has `doSync: Boolean = false`.

t-bast: Great, I missed that

t-bast (approved):

PR #1549, opened 2020-10-07
Fix override-features implementation

pm47:
This is an alternative to #1498.

We were calling `nodeParams.features` from inside the channel, which is
problematic because we may have overriden those features for specific
peers. This is now fixed.

`IncomingPacket.decrypt` needs features to tell whether we are using
variable onion or not. What features should we use? Do we have an
unwritten rule that node-level features shouldn't be overriden? This
happens in the relayer and in the post-restart-htlc-cleaner.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -51,7 +51,7 @@ case class NodeParams(keyManager: KeyManager,
                       color: Color,
                       publicAddresses: List[NodeAddress],
                       features: Features,
-                      overrideFeatures: Map[PublicKey, Features],
+                      private val overrideFeatures: Map[PublicKey, Features],
                       syncWhitelist: Set[PublicKey],
                       pluginParams: Seq[PluginParams],
                       dustLimit: Satoshi,
@@ -92,6 +92,8 @@ case class NodeParams(keyManager: KeyManager,
   val keyPair = KeyPair(nodeId.value, privateKey.value)
 
   def currentBlockHeight: Long = blockCount.get
+
+  def featuresFor(nodeId: PublicKey) = overrideFeatures.getOrElse(nodeId, features)
 }
 
 object NodeParams {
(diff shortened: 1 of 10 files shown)


pm47:
Note that if we change the `var_onion_optin` after a restart, then the post-restart-htlc-cleaner task may not work. We *could* make `PostRestartHtlcCleaner.checkBrokenHtlcs` use `channels.commitments.localParams.features` to decrypt htlcs, wdyt?

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -92,6 +92,8 @@ case class NodeParams(keyManager: KeyManager,
   val keyPair = KeyPair(nodeId.value, privateKey.value)
 
   def currentBlockHeight: Long = blockCount.get
+
+  def featuresFor(nodeId: PublicKey) = overrideFeatures.getOrElse(nodeId, features)
```
pm47, line 96: See https://github.com/ACINQ/eclair/pull/1498#issuecomment-702827782.

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -72,10 +72,8 @@ class Switchboard(nodeParams: NodeParams, watcher: ActorRef, relayer: ActorRef,
     case authenticated: PeerConnection.Authenticated =>
       // if this is an incoming connection, we might not yet have created the peer
       val peer = createOrGetPeer(authenticated.remoteNodeId, offlineChannels = Set.empty)
-      val features = nodeParams.overrideFeatures.get(authenticated.remoteNodeId) match {
-        case Some(f) => f
-        case None => nodeParams.features.maskFeaturesForEclairMobile()
-      }
+      // TODO: @tbast: we now apply maskFeaturesForEclairMobile to custom features as well, is that ok?
```
t-bast, line 75: I think this is ok. The two features we mask off are really invoice features; the only place where it really matters to set them is inside invoices.
We can omit them in `Init`, I'm not seeing any issue it could cause.

t-bast (commented):
> Do we have an unwritten rule that node-level features shouldn't be overriden?

I think so, yes. It feels to me that turning `var_onion_optin` off is asking for trouble...
There are some features that we may want to make mandatory at some point, and disallow users turning them off, don't you think? It will also allow us to simplify the code by removing some unnecessary `if` statements. I'm thinking that `var_onion_optin` is exactly a good candidate for that, there's too much risk of footgun for users if we let them turn it off (it's been battle-tested with 1.5 years in the wild and 0 drawbacks)

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
pm47:
```suggestion
```

t-bast (approved):

PR #1551, opened 2020-10-08
Use BitVector instead of ByteVector in UnknownMessage

btcontract:
This is a minor change which I find to be more convenient while working on my first plugin. 

Spec following plugin is expected to define its own set of codecs and these codecs take `BitVector`, not `ByteVector` so with `UnknownMessage` containing a `BitVector` data we can pass it to codec directly and save on `.toBitVector` conversion.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
@@ -298,7 +298,7 @@ object LightningMessageCodecs {
 
   val unknownMessageCodec: Codec[UnknownMessage] = (
     ("tag" | uint16) ::
-      ("message" | varsizebinarydata)
+      ("message" | variableSizeBits(uint16, bits))
     ).as[UnknownMessage]
 
   // NB: blank lines to minimize merge conflicts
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -298,7 +298,7 @@ object LightningMessageCodecs {
 
   val unknownMessageCodec: Codec[UnknownMessage] = (
     ("tag" | uint16) ::
-      ("message" | varsizebinarydata)
+      ("message" | variableSizeBits(uint16, bits))
```
btcontract, line 301: Except this probably reduces max size of message by 8? Should `uint24` be used here?

t-bast:
I think using `ByteVector` is better because everything is byte-aligned in lightning messages.
Note that if you have a `ByteVector`, you can simply call `.bits` on it to get a `BitVector`, do your thing, and then call `.bytes` on that `BitVector` to get a `ByteVector` back.
Would that work for you?

btcontract:
`bits` seems to be identical to `toBitVector` in computation cost, but this is really minor anyway and if convention is to have `ByteVector` everywhere then it's fine.

PR #1552, opened 2020-10-08
Prevent disabling `var_onion_optin`

t-bast:
There is no good reason to completely disable `var_onion_optin` and it can even be harmful: pending HTLCs after a restart may be lost if they used a variable-length onion, causing a loss of funds and potential channel closes.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -198,10 +198,14 @@ object NodeParams {
     val nodeAlias = config.getString("node-alias")
     require(nodeAlias.getBytes("UTF-8").length <= 32, "invalid alias, too long (max allowed 32 bytes)")
 
+    def validateFeatures(features: Features): Unit = {
+      val featuresErr = Features.validateFeatureGraph(features)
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -214,7 +215,10 @@ object NodeParams {
     val overrideFeatures: Map[PublicKey, Features] = config.getConfigList("override-features").asScala.map { e =>
       val p = PublicKey(ByteVector.fromValidHex(e.getString("nodeid")))
       val f = Features.fromConfiguration(e)
-      p -> f
+      val err = Features.validateFeatureGraph(f)
+      require(err.isEmpty, err.map(_.message))
+      require(f.hasFeature(Features.VariableLengthOnion), s"${Features.VariableLengthOnion.rfcName} must be enabled")
```
pm47, line 220: That should probably be factored in a `def validateFeatures`, defined inside `def makeNodeParams`.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1552/commits/778b857a752a6cedf6953b4052e7f55581b47570

pm47 (approved):

PR #1556, opened 2020-10-09
Fix channel CMD_SIGN

t-bast:
We're sending the type instead of sending an instantiated class...

NB: we're doing the same thing with `CMD_GETSTATE`, `CMD_GETSTATEDATA` and `CMD_GETINFO`: everywhere in the codebase we use it as if it were an object without any paramter, omitting the `replyTo`...I think the compiler should warn for these things, not sure if we need to fix it now for these commands, WDYT?

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -203,16 +203,16 @@ class EclairImpl(appKit: Kit) extends Eclair {
   override def channelsInfo(toRemoteNode_opt: Option[PublicKey])(implicit timeout: Timeout): Future[Iterable[RES_GETINFO]] = toRemoteNode_opt match {
     case Some(pk) => for {
       channelIds <- (appKit.register ? Symbol("channelsTo")).mapTo[Map[ByteVector32, PublicKey]].map(_.filter(_._2 == pk).keys)
-      channels <- Future.sequence(channelIds.map(channelId => sendToChannel[RES_GETINFO](Left(channelId), CMD_GETINFO)))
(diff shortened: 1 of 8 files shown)


pm47:
Definitely need to fix imo.

pm47 (commented):
I think adding a `bob2alice.expectMsgType[CommitSig]` right after this line: https://github.com/ACINQ/eclair/blob/50363569c771afd9e617cb6f7d98ccac34a72911/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/OfflineStateSpec.scala#L472

would make the test catch this bug.

t-bast:
> Definitely need to fix imo.

Done in the last commit.

pm47 (changes requested):
You are missing one:

```diff
diff --git a/eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/controllers/ChannelPaneController.scala b/eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/controllers/ChannelPaneController.scala
index 34b358027..7722f34af 100644
--- a/eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/controllers/ChannelPaneController.scala
+++ b/eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/controllers/ChannelPaneController.scala
@@ -107,7 +107,7 @@ class ChannelPaneController(val channelRef: ActorRef, val peerNodeId: String) ex
             """.stripMargin, ButtonType.YES, ButtonType.NO)
         alert.showAndWait
         if (alert.getResult eq ButtonType.YES) {
-          channelRef ! CMD_FORCECLOSE
+          channelRef ! CMD_FORCECLOSE(ActorRef.noSender)
         }
       }
     })
```

regex: `CMD_[a-zA-Z_]+(?=\s)`

pm47 (approved):

PR #1557, opened 2020-10-10
Use maven plugin to include tests

btcontract:
This will result in `eclair-core_2.13-<version>-tests.jar` file and allow to easily reuse Eclair testing classes in plugins.

### DIFF
--- a/eclair-core/pom.xml
+++ b/eclair-core/pom.xml
@@ -69,6 +69,13 @@
                         </manifestEntries>
                     </archive>
                 </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
             </plugin>
         </plugins>
     </build>


pm47:
That's awesome, I didn't know about that feature.

t-bast (approved):
Gave it a try, nice find, LGTM

pm47:
Actually this is less awesome than I thought:
> Note: The downside of this solution is that you don't get the transitive test-scoped dependencies automatically. Maven only resolves the compile-time dependencies, so you'll have to add all the other required test-scoped dependencies by hand.

> The preferred way
In order to let Maven resolve all test-scoped transitive dependencies you should create a separate project.

https://maven.apache.org/plugins/maven-jar-plugin/examples/create-test-jar.html

btcontract:
I've stopped at what was working (as far as I understand this way requires both `core` and `core-tests` jars included, but plugins need both of these jars anyway).

pm47:
Indeed, it's still great for our use-case 👍

PR #1564, opened 2020-10-13
Fix race condition in integration test

t-bast:
In the revoked commit tx case, both nodes are competing to claim the HTLC outputs from the commit tx.

The test incorrectly assumed that node F would always win that race.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/integration/ChannelIntegrationSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/integration/ChannelIntegrationSpec.scala
@@ -738,10 +738,9 @@ class AnchorOutputChannelIntegrationSpec extends ChannelIntegrationSpec {
     sender.expectMsg(revokedCommitTx.txid)
     // get the revoked commitment confirmed: now HTLC txs can be published
     generateBlocks(bitcoincli, 2)
-    bitcoinClient.publishTransaction(htlcSuccess.head).pipeTo(sender.ref)
-    sender.expectMsg(htlcSuccess.head.txid)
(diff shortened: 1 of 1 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/integration/ChannelIntegrationSpec.scala
```diff
@@ -738,10 +738,9 @@ class AnchorOutputChannelIntegrationSpec extends ChannelIntegrationSpec {
     sender.expectMsg(revokedCommitTx.txid)
     // get the revoked commitment confirmed: now HTLC txs can be published
     generateBlocks(bitcoincli, 2)
-    bitcoinClient.publishTransaction(htlcSuccess.head).pipeTo(sender.ref)
-    sender.expectMsg(htlcSuccess.head.txid)
-    bitcoinClient.publishTransaction(htlcTimeout.head).pipeTo(sender.ref)
-    sender.expectMsg(htlcTimeout.head.txid)
+    // NB: there is a race between C and F here; C may publish more quickly and claim the HTLC outputs directly from the commit tx
```
pm47, line 741:
```suggestion
    // NB: The test cannot be deterministic because there is a race between C and F here; C may publish more quickly and claim the HTLC outputs directly from the commit tx. As a result we may have different combinations of transactions if the test is run several times. It's okay, we just need to make sure that the test never fails.
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1564/commits/80c43cfbfe6bfcedb53a4e3f6c3bbcd08434c559

pm47 (approved):

PR #1566, opened 2020-10-14
Distribute connection-handling on multiple machines using akka-cluster

pm47:
### Architecture overview

Here is what a deployment looks like:
```
                 +-----------+
                 |           |
                 | +-------+ |
            -----|-|       | |
            -----|-|       | |
            -----|-| FRONT |-|---,
            -----|-|       | |    \
     P      -----|-|       | |     \
     U           | +-------+ |      \
     B           |           |       \
     L           | +-------+ |        \  +------+
     I      -----|-|       | |         \ |      |
     C      -----|-|       | |          `|      |    router
            -----|-| FRONT |-|-----------| BACK |<-- switchboard
     N      -----|-|       | |          ,|      |    peers
     E      -----|-|       | |         / |      |    channels
     T           | +-------+ |        /  +------+    reconnection-tasks
     W           |           |       /
     O           | +-------+ |      /
     R      -----|-|       | |     /
     K      -----|-|       | |    /
            -----|-| FRONT |-|---'
            -----|-|       | |       front-router
            -----|-|       |<------- server
                 | +-------+ |       clients
                 |           |       peer-connections
                 +-----------+
```

### Features

The goal is to offload the back from everything connection-related. Following is handled by the front:
- incoming connections
- outgoing connections
- gossip queries + pings
- incoming gossip aggregation
- outgoing gossip dispatch (rebroadcast)

### Security

See:
- https://doc.akka.io/docs/akka/2.6.10/remoting-artery.html?language=scala#remote-security

According to the documentation, the base assumption with `akka-remote` is that it operates within a datacenter, isolated from the outside with firewall rules.

If you have access to the front, out-of-the-box with the default java-based serialization you could pretty much do whatever you want and send any commands to the backend (e.g. send a `Peer.OpenChannel` with `push_msat`>0 to the switchboard). In other words hacking one of the front servers would be the same as hacking the backend node.

We do mitigate that, using:

#### Custom serialization

We control what messages can be sent to the back using a custom scodec-based serialization, which is effectively a whitelist.

#### Untrusted mode

See:
- https://doc.akka.io/docs/akka/2.6.10/remoting-artery.html?language=scala#untrusted-mode

Caveat, this is no silver bullet:
> Untrusted mode does not give full protection against attacks by itself. It makes it slightly harder to perform malicious or unintended actions but it should be noted that Java serialization should still not be enabled. Additional protection can be achieved when running in an untrusted network by network security (e.g. firewalls) and/or enabling TLS with mutual authentication.

#### Remote deployment whitelist

We use this feature to explicitly disallow any actor creation on the back from the front.

#### Encryption

Communication between the front and the back over `akka-remote` can (and should in production) be encrypted, using TLS 1.2 (`TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`).

Note that it is also possible to enable TLS mutual authentication.

### Backpressure

Internally `akka-remote` uses message queues, and if they overflow (slow network, busy target), then messages are dropped. This could be problematic, especially during network sync.

I have implemented back-pressure for gossip messages: they are aggregated/deduplicated by the front, forwarded to the router, and only when the router replies are they acked.

Channel-level messages (forwarded to the peer) don't have backpressure : they are acknowledged by the front upon reception, and forwarded to the back.

I have also increased the default queue size with `akka.actor.remote.artery.advanced.outbound-message-queue-size = 30720` (x10 compared to default).

### Tor

There is some interference with the support of Tor for incoming/outgoing connections:
- outgoing to hidden service
We should be able to run a tor proxy that would be accessible by front servers. Not implemented, for now we instantiate those connections on the back.
- incoming to hidden service
That's tricky: it's not enough to run a tor node on each front server, they would also need to share the same service address. I haven't looked deep into that. Otherwise we could default to only have one single front handling incoming tor connections.
- incoming from tor to clearnet address
This is the Phoenix use case when Tor is enabled. That's ok, fortunately because on our side it is a regular clearnet connection.
 
### Deployment

An AWS Beanstalk bundle is automatically generated for the module `eclair-front`, which makes the deployment pretty straightforward on that platform. It is recommended to use AWS Secrets Manager to store the node private key.


eclair-front/src/main/resources/application.conf
```diff
@@ -0,0 +1,86 @@
...
+      # more data is read from the network the connection actor will try to
+      # do other work.
+      # The purpose of this setting is to impose a smaller limit than the
+      # configured receive buffer size. When using value 'unlimited' it will
+      # try to read all from the receive buffer.
+      # As per BOLT#8 lightning messages are at most 2 + 16 + 65535 + 16 = 65569bytes
+      # Currently the largest message is update_add_htlc (~1500b).
+      # As a tradeoff to reduce the RAM consumption, in conjunction with tcp pull mode,
+      # the default value is chosen to allow for a decent number of messages to be prefetched.
+      max-received-message-size = 16384b
```
t-bast, line 83: Actually this isn't accurate anymore, gossip messages can fill up to the 64k limit, can't they?

eclair-front/src/main/resources/logback.xml
```diff
@@ -0,0 +1,32 @@
...
+    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+        <target>System.out</target>
+        <withJansi>false</withJansi>
+        <encoder>
+            <pattern>${HOSTNAME} %d %-5level %logger{24}%X{category}%X{nodeId}%X{channelId}%X{paymentHash}%.-11X{parentPaymentId}%.-11X{paymentId} - %msg%ex{12}%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="INFO">
+        <appender-ref ref="CONSOLE"/>
```
t-bast, line 29: Why don't we propose a log file appender?
Because we only optimize for AWS deployment for now and they pipe the console to some AWS logging output?

eclair-front/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -0,0 +1,63 @@
...
+import akka.actor.ActorSystem
+import com.typesafe.config.{ConfigFactory, ConfigParseOptions, ConfigSyntax}
+import grizzled.slf4j.Logging
+import kamon.Kamon
+
+import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success}
+
+/**
+ * Created by PM on 25/01/2016.
```
t-bast, line 30: I know you started working early on this feature, but I don't believe you there :smile:

eclair-front/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -0,0 +1,63 @@
...
+ */
+object Boot extends App with Logging {
+  try {
+    val datadir = new File(System.getProperty("eclair.datadir", System.getProperty("user.home") + "/.eclair"))
```
t-bast, line 34: nit: isn't that redundant with the default you added to `reference.conf`?

eclair-front/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -0,0 +1,63 @@
...
+    val config = ConfigFactory.parseString(
+      Option(System.getenv("AKKA_CONF")).getOrElse("").replace(";", "\n"),
+      ConfigParseOptions.defaults().setSyntax(ConfigSyntax.PROPERTIES))
+      .withFallback(ConfigFactory.load())
+
+    implicit val system: ActorSystem = ActorSystem("eclair-node", config)
```
t-bast, line 40:
```suggestion
    implicit val system: ActorSystem = ActorSystem("eclair-front", config)
```

eclair-front/src/main/scala/fr/acinq/eclair/ClusterListener.scala
```diff
@@ -0,0 +1,49 @@
...
+      }
+    case MemberLeft(member) => maybeShutdown(member)
+    case MemberDowned(member) => maybeShutdown(member)
+    case UnreachableMember(member) => maybeShutdown(member)
+  }
+
+  private def maybeShutdown(member: Member): Unit = {
+    if (member.roles.contains(BackendRole)) {
+      log.info(s"backend is down, stopping...")
+      System.exit(0)
```
t-bast, line 45: Why don't we use `context.system.terminate()`? Isn't that the recommended way of shutting down the actor system?

eclair-front/src/main/scala/fr/acinq/eclair/FrontSetup.scala
```diff
@@ -0,0 +1,108 @@
...
+
+      _ = system.actorOf(Props(new ClusterListener(backendAddressFound)), name = "cluster-listener")
+      backendAddress <- backendAddressFound.future
+
+      switchBoardSelection = system.actorSelection(RootActorPath(backendAddress) / "user" / "*" / "switchboard")
+      remoteSwitchboard <- switchBoardSelection.resolveOne()
+      routerSelection = system.actorSelection(RootActorPath(backendAddress) / "user" / "*" / "router")
+      remoteRouter <- routerSelection.resolveOne()
+
+      RouterPeerConf(routerConf, peerConnectionConf) <- (switchBoardSelection ? GetRouterPeerConf).mapTo[RouterPeerConf]
```
t-bast, line 96: Can't we do:

```suggestion
      RouterPeerConf(routerConf, peerConnectionConf) <- (remoteSwitchboard ? GetRouterPeerConf).mapTo[RouterPeerConf]
```

t-bast (commented):
Cool stuff! One thing that I don't understand though: for now both the `Router` and `FrontRouter` will regularly publish `Rebroadcast` messages. Does this means we're duplicating the rebroadcast? Is that intended?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -84,7 +84,7 @@ case object INPUT_DISCONNECTED
 case class INPUT_RECONNECTED(remote: ActorRef, localInit: Init, remoteInit: Init)
 case class INPUT_RESTORED(data: HasCommitments)
```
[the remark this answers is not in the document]
pm47: Yes that's the reason. Note that our codec provide a greater protection than this flag, which can be bypassed by simply disabling untrusted mode. I almost feel like we should not use the `PossiblyHarmful` trait for this reason.

eclair-core/src/main/scala/fr/acinq/eclair/remote/ScodecSerializer.scala
```diff
@@ -0,0 +1,40 @@
...
+
+package fr.acinq.eclair.remote
+
+import java.nio.ByteBuffer
+
+import akka.serialization.{ByteBufferSerializer, SerializerWithStringManifest}
+import scodec.Codec
+import scodec.bits.BitVector
+
+class ScodecSerializer[T <: AnyRef](override val identifier: Int, codec: Codec[T]) extends SerializerWithStringManifest  with ByteBufferSerializer {
```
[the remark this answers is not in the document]
pm47: Done with 645183a55.

eclair-core/src/main/scala/fr/acinq/eclair/io/ClientSpawner.scala
```diff
@@ -18,19 +18,42 @@ package fr.acinq.eclair.io
...
+      context.system.eventStream.subscribe(self, classOf[DeadLetter])
+    }
+  }
+
   override def receive: Receive = {
     case req: ClientSpawner.ConnectionRequest =>
       log.info("initiating new connection to nodeId={} origin={}", req.remoteNodeId, sender)
       context.actorOf(Client.props(keyPair, socks5ProxyParams_opt, peerConnectionConf, switchboard, router, req.address, req.remoteNodeId, origin_opt = Some(req.origin)))
+    case DeadLetter(req: ClientSpawner.ConnectionRequest, _, _) =>
+      log.warning("handling outgoing connection request locally")
```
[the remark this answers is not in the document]
pm47: No, we only subscribe to the deadletters event stream when in cluster mode.

Added a comment in 7fb4433c5.

eclair-core/src/main/scala/fr/acinq/eclair/io/ClientSpawner.scala
```diff
@@ -18,19 +18,42 @@ package fr.acinq.eclair.io
...
 import fr.acinq.eclair.tor.Socks5ProxyParams
 
 class ClientSpawner(keyPair: KeyPair, socks5ProxyParams_opt: Option[Socks5ProxyParams], peerConnectionConf: PeerConnection.Conf, switchboard: ActorRef, router: ActorRef) extends Actor with ActorLogging {
 
   context.system.eventStream.subscribe(self, classOf[ClientSpawner.ConnectionRequest])
 
+  if (context.system.hasExtension(Cluster)) {
+    val roles = context.system.extension(Cluster).selfRoles
+    if (roles.contains("frontend")) {
+      import akka.cluster.pubsub.DistributedPubSubMediator.Put
```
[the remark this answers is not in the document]
pm47: Fixed with 8071b7dba.

eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
```diff
@@ -0,0 +1,176 @@
...
+
+  val singleChannelDiscoveredCodec: Codec[SingleChannelDiscovered] = (lengthPrefixedChannelAnnouncementCodec :: satoshi :: optional(bool(8), lengthPrefixedChannelUpdateCodec) :: optional(bool(8), lengthPrefixedChannelUpdateCodec)).as[SingleChannelDiscovered]
+
+  val readAckCodec: Codec[TransportHandler.ReadAck] = lightningMessageCodec.upcast[Any].as[TransportHandler.ReadAck]
+
+  def codec(system: ExtendedActorSystem): Codec[RemoteTypes] = discriminated[RemoteTypes].by(uint16)
+    .typecase(0, provide(Switchboard.GetRouterPeerConf))
+    .typecase(1, (routerConfCodec :: peerConnectionConfCodec).as[RouterPeerConf])
+    .typecase(5, readAckCodec)
+    .typecase(7, connectionRequestCodec(system))
```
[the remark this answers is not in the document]
pm47: They are more or less grouped by target actor. I don't think it matters a lot since this is for internal messages.

eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
```diff
@@ -0,0 +1,176 @@
...
+
+class EclairInternalsSerializer(val system: ExtendedActorSystem) extends ScodecSerializer(43, EclairInternalsSerializer.codec(system))
+
+object EclairInternalsSerializer {
+
+  trait RemoteTypes extends Serializable
+
+  def finiteDurationCodec: Codec[FiniteDuration] = int64.xmap(_.milliseconds, _.toMillis)
+
+  def set[A](codec: Codec[A]): Codec[Set[A]] = listOfN(uint16, codec).xmap(_.toSet, _.toList)
```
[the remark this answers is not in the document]
pm47: You are right, it's unused. Removed in 825a3136a.

eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
```diff
@@ -0,0 +1,176 @@
...
+import fr.acinq.eclair.router._
+import fr.acinq.eclair.wire.CommonCodecs._
+import fr.acinq.eclair.wire.LightningMessageCodecs._
+import fr.acinq.eclair.wire.QueryChannelRangeTlv.queryFlagsCodec
+import fr.acinq.eclair.wire.{AnnouncementMessage, EncodingType, QueryChannelRangeTlv, RoutingMessage}
+import fr.acinq.eclair.{CltvExpiryDelta, Features}
+import scodec._
+import scodec.codecs._
+
+import scala.concurrent.duration.{FiniteDuration, _}
```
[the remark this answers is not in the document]
pm47: For some reason, this stays after optimizing imports on intellij. Oh well.

Fixed with b2a17e3c2.

eclair-core/src/main/scala/fr/acinq/eclair/remote/ScodecSerializer.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  override def toBinary(o: AnyRef, buf: ByteBuffer): Unit = buf.put(toBinary(o))
+
+  override def fromBinary(buf: ByteBuffer, manifest: String): AnyRef =  {
+    val bytes = Array.ofDim[Byte](buf.remaining())
+    buf.get(bytes)
+    fromBinary(bytes, manifest)
+  }
+
+  override def manifest(o: AnyRef): String = fr.acinq.eclair.getSimpleClassName(o)
```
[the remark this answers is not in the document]
pm47: Correct, done with d2028a735.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -127,6 +128,33 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
...
+        .values
+        .map(pc => SingleChannelDiscovered(pc.ann, pc.capacity, pc.update_1_opt, pc.update_2_opt))
+        .sliding(100, 100)
+        .map(ChannelsDiscovered)
+        .foreach(listener ! _)
+      listener ! RoutingStateStreamingUpToDate
+      context.actorOf(Props(new Actor with ActorLogging {
+        log.info(s"subscribing listener=$listener to network events")
+        context.system.eventStream.subscribe(listener, classOf[NetworkEvent])
+        context.watch(listener)
```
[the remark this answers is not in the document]
pm47: You are right, although it is a tiny memory leak (front nodes should be long lived). Will look into it...

eclair-front/modules/assembly.xml
```diff
@@ -0,0 +1,59 @@
...
+    </fileSet>
+    <!--fileSet>
+      <directory>${project.basedir}/modules/awseb</directory>
+      <outputDirectory>.</outputDirectory>
+      <includes>
+        <include>*.jks</include>
+      </includes>
+      <fileMode>0400</fileMode>
+      <lineEnding>keep</lineEnding>
+    </fileSet-->
```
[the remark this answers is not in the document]
pm47: Yeah it's just an example for people who wish to use aws eb and enable tls. 

Added a comment in 7db91903f.

eclair-front/src/main/resources/logback.xml
pm47: Yes.

eclair-front/src/main/scala/fr/acinq/eclair/Boot.scala
pm47: I *may* have copy pasted that one. Removed in aed39363b.

eclair-front/src/main/scala/fr/acinq/eclair/Boot.scala
pm47: True, but there is a chicken and egg problem. We need the data dir to look up the `eclair.conf` file. What's in the `reference.conf` is actually a reminder to where the datadir actually is, but it doesn't actively set it.

eclair-front/src/main/scala/fr/acinq/eclair/Boot.scala
pm47: Okay this is embarrassing. I may have done this on purpose to enable grouping of actors in the monitoring...

eclair-front/src/main/scala/fr/acinq/eclair/FrontSetup.scala
pm47: Yes, and it's probably easier to understand.

eclair-front/src/main/scala/fr/acinq/eclair/Boot.scala
pm47: Yeah, I think keeping `eclair-node` is good. WDYT?

t-bast: Sounds good, a comment would be useful in that case to explain that choice

pm47: Okay, can't have a different actorsystem name across the cluster apparently. Tried, didn't work.

https://github.com/akka/akka/blob/1bf012837cf7f0120acb61e6a83eb66a93b23ab0/akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala#L696-L700

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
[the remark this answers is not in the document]
pm47: I verified that, in untrusted mode, death watches are still triggered when the whole actor system goes down (as opposed to: when the specific actor that is watched dies). It's good, because here the lifespan of the watched actor is the same as the actor system, so there is no memory leak.

We do have a nasty race condition here though. When we restart the front, the death watch may trigger immediately because of a race with the cluster membership lifecycle (see https://github.com/akka/akka/blob/93a69c42ff465603504fa32f4cb91070618676ce/akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala#L174). The subscription to `NetworkEvent`s will be cancelled right after it is set. This has nasty side effects, because we rely on those `NetworkEvent`s to acknowledge messages from our peers on the front. This resulted in stuck tcp connections.

The fix is in two steps:
- d4191cd53 or 29a1c38bb (competing commits, the latter partially reverts the former) make the message acknowledgement independent of the `NetworkEvent`s. This is a bit complicated because there are several potential race conditions to take into account. See the comment in the `GossipDecision.Accepted` handler of the `FrontRouter` for more details.
- 1d800e8b4 and 8971d9725 solve the death watch issue by adding some time for the cluster to properly detect the new incarnation of the same front server when we restart.

eclair-front/src/main/scala/fr/acinq/eclair/router/FrontRouter.scala
```diff
@@ -0,0 +1,320 @@
...
+      }
+      // NB: we will also shortly receive a NetworkEvent from the router for this announcement, where we put it in the
+      // rebroadcast map. We keep the origin peers in the accepted map, so we don't send back the same announcement to
+      // the peers that sent us in the first place.
+      // Why do we not just leave the announcement in the processing map? Because we would have a race between other
+      // peers that send us that same announcement, and the NetworkEvent. If the other peers win the race, we will defer
+      // acknowledging their message (because the announcement is still in the processing map) and we will
+      // wait forever for the very gossip decision that we are processing now, resulting in a stuck connection
+      val origins1 = d.processing.getOrElse(accepted.ann, Set.empty[RemoteGossip])
+      stay using d.copy(processing = d.processing - accepted.ann, accepted = d.accepted + (accepted.ann -> origins1))
```
pm47, line 158: The `accepted` map will be cleaned up when receiving the `NetworkEvent` for that announcement. If that event never arrives, because of a bug, we will never free the memory.

We could choose to arbitrarily clean up this map when processing `TickBroadcast` events, but then we have a new race condition, where we regularly forget the origin information and send back announcements to peers that already know it.

eclair-front/src/main/resources/application.conf
pm47: Here is the doc for that configuration parameter:
```
      # The maximum number of bytes delivered by a `Received` message. Before
      # more data is read from the network the connection actor will try to
      # do other work.
      # The purpose of this setting is to impose a smaller limit than the
      # configured receive buffer size. When using value 'unlimited' it will
      # try to read all from the receive buffer.
      max-received-message-size = unlimited
```

I think we should use a multiple of the max message size, maybe 262144 bytes, WDYT?

t-bast: I agree, that's a good starting parameter :+1: 
And we'll adjust if we see issues related to that.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -127,6 +128,33 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
       sender ! GetNetworkStatsResponse(d.stats)
       stay
 
+    case Event(GetRoutingStateStreaming, d) =>
```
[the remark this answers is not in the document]
pm47: Done with f04f6b2.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -505,16 +512,16 @@ object PeerConnection {
...
+  case class ConnectionReady(peerConnection: ActorRef, remoteNodeId: PublicKey, address: InetSocketAddress, outgoing: Boolean, localInit: wire.Init, remoteInit: wire.Init) extends RemoteTypes
 
-  sealed trait ConnectionResult
+  sealed trait ConnectionResult extends RemoteTypes
   object ConnectionResult {
     sealed trait Success extends ConnectionResult
     sealed trait Failure extends ConnectionResult
 
-    case class NoAddressFound(remoteNodeId: PublicKey) extends ConnectionResult.Failure { override def toString: String = "no address found" }
+    case object NoAddressFound extends ConnectionResult.Failure { override def toString: String = "no address found" }
```
pm47, line 524: @btcontract I reverted that change made in #1594. Please don't just randomely add a `remoteNodeId` there because you need it, and break consistency among the different `ConnectionResult` types. I'm not opposed to adding it at the `ConnectionResult` level.

eclair-front/src/main/resources/application.conf
pm47: Done in 06d4d12.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
pm47: Actually just tried it and this creates issues because in some cases (e.g. for `ConnectionResult.AuthenticationFailed`) we don't have it. Sigh...

eclair-front/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -0,0 +1,60 @@
...
+
+object Boot extends App with Logging {
+  try {
+    val datadir = new File(System.getProperty("eclair.datadir", System.getProperty("user.home") + "/.eclair"))
+    val config = ConfigFactory.parseString(
+      Option(System.getenv("AKKA_CONF")).getOrElse("").replace(";", "\n"),
+      ConfigParseOptions.defaults().setSyntax(ConfigSyntax.PROPERTIES))
+      .withFallback(ConfigFactory.load())
+
+    implicit val system: ActorSystem = ActorSystem("eclair-node", config)
```
t-bast, line 37: I still think this is worth a comment explaining that it really must be `eclair-node` and not `eclair-front`

eclair-front/src/main/scala/fr/acinq/eclair/router/FrontRouter.scala
t-bast: The current solution feels ok. We can always monitor memory consumption, that should indicate potential bugs.
And front machines should be expendable, so it's not a big issue if we have to restart one at some point because of a memory leak (and will help figure out the bug and fix it afterwards).

t-bast (commented):
The `FrontRouter` logic is quite subtle, but I didn't notice anything obviously wrong.
We'll see how that behaves in the wild!

It would be nice to add a `Cluster.md` documentation file in the `docs` folder, to walk people through the process of deploying and monitoring this beast, don't you think? It can be done in a separate PR if you want to get this merged to avoid too many rebase.

eclair-front/src/main/scala/fr/acinq/eclair/Boot.scala
pm47: Fixed with c0e330a11.

pm47:
> It would be nice to add a Cluster.md documentation file in the docs folder, to walk people through the process of deploying and monitoring this beast, don't you think? It can be done in a separate PR if you want to get this merged to avoid too many rebase.

Definitely! Let's do this in a separate PR so I can finally squash this 😌

t-bast (approved):
:rocket: :rocket: :rocket:

PR #1569, opened 2020-10-16
Add an option to avoid bitcoin sig checking for ChannelAnnouncement

btcontract:
This change will make it easier to check public hosted channels' announcements which don't have a chain funding part.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
@@ -149,12 +149,10 @@ object Announcements {
     )
   }
 
-  def checkSigs(ann: ChannelAnnouncement): Boolean = {
+  def checkSigs(ann: ChannelAnnouncement, checkBitcoinSigs: Boolean = true): Boolean = {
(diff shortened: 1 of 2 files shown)


pm47:
This is starting to feel very hackish...

btcontract:
I guess this post can explain best what I'm up to: https://www.rene-pickhardt.de/index.html%3Fp=2131.html

tl/dr it's not as bad as it may seem if used right.

btcontract:
Another point is this is needed for a plugin only, won't be used in core anywhere. I could just copy related code without bitcoin sig part but doing it this way seems nicer.

t-bast:
I don't like this change either, and I really want to avoid adding this footgun in eclair-core.
It's very easy to avoid needing it: hosted channels can simply use the same public key in `nodeId` and `bitcoinKey`, and provide a valid signature of the message.

btcontract:
Thanks for the suggestion, my initial plan was that one of distinguishing parts of PHC announcements is some predetermined static bitcoin key values and zeroed out bitcoin signatures, but bitcoin keys equal to node keys seem an equally good idea, I'll go with it.

PR #1574, opened 2020-10-20
Post restart plugin htlcs

btcontract:
This enriches plugins with ability to provide their own broken and pending HTLCs on restart which are then used in PostRestartHtlcCleaner methods and events.

This is needed in particular as a final piece to enable HC entirely as a plugin, it builds upon plugin messaging and #1542.

Without this functionality it's not possible to correctly handle broken and pending HTLCs which were somehow routed through plugins (in case of HC that would be routings like `peer1 -> HC1 -> HC2 -> peer2` or  `peer1 -> normal -> HC -> peer2` and a like).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -46,7 +46,8 @@ import fr.acinq.eclair.db.{Databases, FileBackupHandler}
 import fr.acinq.eclair.io.{ClientSpawner, Server, Switchboard}
 import fr.acinq.eclair.payment.Auditor
 import fr.acinq.eclair.payment.receive.PaymentHandler
-import fr.acinq.eclair.payment.relay.Relayer
+import fr.acinq.eclair.payment.relay.PostRestartHtlcCleaner.PluginHtlcs
+import fr.acinq.eclair.payment.relay.{PostRestartHtlcCleaner, Relayer}
 import fr.acinq.eclair.payment.send.{Autoprobe, PaymentInitiator}
 import fr.acinq.eclair.router._
 import fr.acinq.eclair.tor.TorProtocolHandler.OnionServiceVersion
@@ -56,6 +57,7 @@ import grizzled.slf4j.Logging
 import org.json4s.JsonAST.JArray
 import scodec.bits.ByteVector
 
+import akka.pattern.ask
 import scala.concurrent._
 import scala.concurrent.duration._
 import scala.util.{Failure, Success}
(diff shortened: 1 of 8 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -65,12 +67,15 @@ import scala.util.{Failure, Success}
  *
  * Created by PM on 25/01/2016.
  *
- * @param datadir  directory where eclair-core will write/read its data.
- * @param seed_opt optional seed, if set eclair will use it instead of generating one and won't create a seed.dat file.
- * @param db       optional databases to use, if not set eclair will create the necessary databases
+ * @param datadir       directory where eclair-core will write/read its data.
+ * @param pluginParams  a list of (node feature bit, LN message tags) each network-enabled plugin is interested in.
```
t-bast, line 71: You shouldn't describe what's inside `PluginParams`, otherwise the comment will be obsolete and probably not fixed in we change the contents of `PluginParams`.

```suggestion
 * @param pluginParams  parameters for all configured plugins.
```

t-bast (changes requested):
NACK. This is way too invasive, we don't want implementation details of a specific plugin to be bleeding inside the core implementation like that.

I think that either:

- htlcs routed via hosted channels should setup the necessary data to be automatically picked up by the `PostRestartHtlcCleaner` with minimal changes
- or the hosted channel plugin should create its own, separate post-restart handler for hosted channels HTLCs

It's hard to judge without knowing how HTLCs are relayed through hosted channels, can you explain how that works and the exact data that is updated at each step?

btcontract:
> It's hard to judge without knowing how HTLCs are relayed through hosted channels, can you explain how that works and the exact data that is updated at each step?

The process mimics normal channel almost exactly: same `UpdateAddHtlc` message followed by signed `StateUpdate` message with sender's signature (which signs a next proposed state with updated balance, added in-flight HTLC taken into account and so on), followed by receiver's `StateUpdate` reply message (so updated cross-signed state is reached and receiver can start routing a payment further, similar to how it's done on obtaining a cross-signed commit tx in normal channel).

The problem to be solved here is when doing initial broken HTLC resolution, payments from normal channels and from HC must meet somewhere so to say. 

For example: in `peer1 -> normal -> HC -> peer2` an HTLC may be cross-signed in `normal` and successfully relayed through `HC` and pending. Then on restart, if `PostRestartHtlcCleaner` does not see that HTLC is still pending in `HC` it would think it's a `notRelayed` one when doing checks in `checkBrokenHtlcs` method.

Similarly in `cleanupRelayDb` method, if `PostRestartHtlcCleaner` does not see that fulfilled `... -> HC -> peer` HTLC is still pending in `HC` (say, because it was OFFLINE when fulfill happened) it would think that related db record should be cleared.

btcontract:
> This is way too invasive, we don't want implementation details of a specific plugin to be bleeding inside the core implementation like that.

I just want to note that I use `HTLCEnabled...` trait name to emphasize it's not just HC but a plugin which somehow gets involved with HTLC processing. In theory I can imagine a plugin which deals with a different cryptocurrency or maybe not cryptocurrency at all.

t-bast:
What I don't understand is that if the `commitments` are correctly updated for hosted-channel htlcs (regardless of how you update them, in your case thanks to a `StateUpdate` message IIUC), `checkBrokenHtlcs` should find them without any change (or with minimal changes). The fact that you need to manually inject HTLC-like objects is a big red herring and a no-go.

It feels to me that what's missing is that your HC commitments are missing parts compared to normal commitments.
If `checkBrokenHtlcs` could work with `AbstractCommitments`, it would work for hosted-channel htlcs transparently which would be much better. Is that possible?

btcontract:
> What I don't understand is that if the commitments are correctly updated for hosted-channel htlcs (regardless of how you update them, in your case thanks to a StateUpdate message IIUC), checkBrokenHtlcs should find them without any change (or with minimal changes).

Yes, that's true, all relevant HTLCs can be extracted from HC commitment objects. 

My initial line of thinking was like this: `PostRestartHtlcCleaner` calls `listLocalChannels` method, gets list of normal channel's commitments and then extracts HTLCs from those. Since `listLocalChannels` does not see HC I should supply additional HTLCs out-of-band.

But sure, somehow abstracting on top of `listLocalChannels` to return a list of `AbstractCommitments` and then working with those with minimal changes is another way to do that. I'll provide another PR which does that.

PR #1575, opened 2020-10-20
Add per-node feerate tolerance

t-bast:
When we have a trusted relationship with some of our peers (business relations, family members, our own mobile wallet, etc) it makes sense to relax the feerate mismatch constraint.

This must be done per-node, to avoid leaving the gates open for attackers.

### DIFF
--- a/docs/Configure.md
+++ b/docs/Configure.md
@@ -1,0 +1,176 @@
+# Configuring Eclair
+
+Eclair reads its configuration file, and writes its logs, to `~/.eclair` by default.
+You can change this behavior with the `eclair.datadir` parameter:
+
+```sh
+eclair-node.sh -Declair.datadir="/path/to/custom/eclair/data/folder"
+```
+
+## Change your node's configuration
+
+The first step is to **actually create the configuration file**.
+Go to `eclair.datadir` and create a file named `eclair.conf`.
+The dncoding should be UTF-8.
+
+Options are set as key-value pairs and follow the [HOCON syntax](https://github.com/lightbend/config/blob/master/HOCON.md).
+Values do not need to be surrounded by quotes, except if they contain special characters.
+
+## Options reference
+
+Here are some of the most common options:
+
+name                         | description                                                | default value
+-----------------------------|------------------------------------------------------------|--------------
+ eclair.chain                | Which blockchain to use: *regtest*, *testnet* or *mainnet* | mainnet
+ eclair.server.port          | Lightning TCP port                                         | 9735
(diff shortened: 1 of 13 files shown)


docs/Configure.md
```diff
@@ -0,0 +1,176 @@
...
+
+```sh
+eclair-node.sh -Declair.datadir="/path/to/custom/eclair/data/folder"
+```
+
+## Change your node's configuration
+
+The first step is to **actually create the configuration file**.
+Go to `eclair.datadir` and create a file named `eclair.conf`.
+The dncoding should be UTF-8.
```
pm47, line 14:
```suggestion
The encoding should be UTF-8.
```

eclair-core/src/main/resources/reference.conf
```diff
@@ -114,6 +114,16 @@ eclair {
...
       ratio-high = 10.0 // will allow remote fee rates as high as 10 times our local feerate
     }
+    override-feerate-tolerance = [ // optional per-node feerate tolerance
+      //      {
+      //        nodeid = "02aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+      //        feerate-tolerance {
+      //          ratio-low = 0.1
+      //          ratio-high = 20.0
+      //        }
+      //      }
```
pm47, line 124: Nit, for consistency with `override-features`:

```suggestion
      #      {
      #        nodeid = "02aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
      #        feerate-tolerance {
      #          ratio-low = 0.1
      #          ratio-high = 20.0
      #        }
      #      }
```

pm47 (commented):
Looking good, just nits.

docs/Configure.md
t-bast: I clearly couldn't decide between encoding and decoding.

eclair-core/src/main/resources/reference.conf
t-bast: Done in https://github.com/ACINQ/eclair/pull/1575/commits/463f39a919ef93f525095441f6449ba1090612a7

docs/Configure.md
t-bast: Done in https://github.com/ACINQ/eclair/pull/1575/commits/463f39a919ef93f525095441f6449ba1090612a7

pm47 (approved):

PR #1576, opened 2020-10-20
Rework features compatibility

t-bast:
The `supportedMandatoryFeatures` set made less and less sense as we start supporting more features and allowing users to turn features on and off depending on the peers they connect to.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -62,9 +62,10 @@ case class Features(activated: Set[ActivatedFeature], unknown: Set[UnknownFeatur
 
   def hasPluginFeature(feature: UnknownFeature): Boolean = unknown.contains(feature)
 
+  /** NB: this method is not reflexive, see [[Features.areCompatible]] if you want symmetric validation. */
   def areSupported(remoteFeatures: Features): Boolean = {
     // we allow unknown odd features (it's ok to be odd)
-    val unknownFeaturesOk = !remoteFeatures.unknown.exists(_.bitIndex % 2 == 0)
+    val unknownFeaturesOk = remoteFeatures.unknown.forall(_.bitIndex % 2 == 1)
     // we verify that we activated every mandatory feature they require
     val knownFeaturesOk = remoteFeatures.activated.forall {
       case ActivatedFeature(_, Optional) => true
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -62,6 +62,17 @@ case class Features(activated: Set[ActivatedFeature], unknown: Set[UnknownFeatur
 
   def hasPluginFeature(feature: UnknownFeature): Boolean = unknown.contains(feature)
 
+  def areSupported(remoteFeatures: Features): Boolean = {
+    // we allow unknown odd features (it's ok to be odd)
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -62,6 +62,17 @@ case class Features(activated: Set[ActivatedFeature], unknown: Set[UnknownFeatur
 
   def hasPluginFeature(feature: UnknownFeature): Boolean = unknown.contains(feature)
 
+  def areSupported(remoteFeatures: Features): Boolean = {
+    // we allow unknown odd features (it's ok to be odd)
+    val unknownFeaturesOk = !remoteFeatures.unknown.exists(_.bitIndex % 2 == 0)
```
pm47, line 67: rewritten in a positive way:

```suggestion
    val unknownFeaturesOk = remoteFeatures.unknown.forall(_.bitIndex % 2 == 1)
```

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -62,6 +62,17 @@ case class Features(activated: Set[ActivatedFeature], unknown: Set[UnknownFeatur
 
   def hasPluginFeature(feature: UnknownFeature): Boolean = unknown.contains(feature)
 
+  def areSupported(remoteFeatures: Features): Boolean = {
```
pm47, line 65:
```suggestion
  /**
  * NB: this method is not reflexive, see [[Features.areCompatible]].
  */
  def areSupported(remoteFeatures: Features): Boolean = {
```

eclair-core/src/test/scala/fr/acinq/eclair/FeaturesSpec.scala
```diff
@@ -105,47 +102,37 @@ class FeaturesSpec extends AnyFunSuite {
...
-      bin"        001000000000000000000000" -> true,
-      bin"        100000000000000000000000" -> true,
-      // those are useful for nonreg testing of the areSupported method (which needs to be updated with every new supported mandatory bit)
-      bin"        000100000000000000000000" -> false,
-      bin"        010000000000000000000000" -> false,
-      bin"    0001000000000000000000000000" -> false,
-      bin"    0100000000000000000000000000" -> false,
-      bin"00010000000000000000000000000000" -> false,
-      bin"01000000000000000000000000000000" -> false
+    case class TestCase(ours: Features, theirs: Features, compatible: Boolean)
```
pm47, line 105: Would it be worth it to also check the `areSupported`?
```scala
case class TestCase(ours: Features, theirs: Features, oursSupportTheirs: Boolean, theirSupportOurs: Boolean, compatible: Boolean)
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1576/commits/c026fa00a7c9e2223760850eae220f810829067b

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1576/commits/c026fa00a7c9e2223760850eae220f810829067b

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1576/commits/c026fa00a7c9e2223760850eae220f810829067b

pm47 (approved):

PR #1578, opened 2020-10-21
Clarify crossSign helper functions

t-bast:
It's important in particular to highlight the case where an HTLC is half-fulfilled or half-failed.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -106,6 +106,10 @@ case class Commitments(channelVersion: ChannelVersion,
       remoteNextCommitInfo.left.toSeq.flatMap(_.nextRemoteCommit.spec.htlcs.collect(incoming).filter(expired).toSet)
   }
 
+  /**
+   * Find an outgoing HTLC that is signed by both parties in at least one of the current commitment transactions (local/remote).
+   * NB: if we're in the middle of fulfilling or failing that HTLC, it will not be returned by this function.
+   */
   def getOutgoingHtlcCrossSigned(htlcId: Long): Option[UpdateAddHtlc] = for {
     localSigned <- remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit).spec.findIncomingHtlcById(htlcId)
     remoteSigned <- localCommit.spec.findOutgoingHtlcById(htlcId)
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -106,6 +106,10 @@ case class Commitments(channelVersion: ChannelVersion,
       remoteNextCommitInfo.left.toSeq.flatMap(_.nextRemoteCommit.spec.htlcs.collect(incoming).filter(expired).toSet)
   }
 
+  /**
+   * Find an outgoing HTLC that is signed by both parties in at least one of the current commitment transactions (local/remote).
```
pm47, line 110: I find this confusing, we actually look for an HTLC that is signed by both parties in both (local and remote transactions). How about:

```suggestion
   * Find an outgoing HTLC that is present in both (local and remote) unrevoked commitment transactions. This means the HTLC is ready to be forwarded (it may already have been forwarded).
```

t-bast: > signed by both parties in both

But that's *not* what the function does, since it can find an HTLC that is signed only by us in the remote commitment (nextRemoteCommitInfo)...

pm47: It wouldn't, otherwise `remoteSigned` would equal `None` and we would yield `None`.

pm47:
```scala
for {
 a <- Some(1)
 b <- Option.empty[Int]
} yield a
> val res0: Option[Int] = None
```

t-bast: Yes of course, but that's more subtle than that...
What you're implying is that if the HTLC is found in the `localCommit`, it will have been cross-signed in the `remoteCommit` before, right? And the reason we look into `remoteNextCommitInfo` is to explicitly NOT return it if the next commit fulfills/fails that HTLC?

pm47: I'm not making assumptions on the order. Actually I think the function could be more precise because some cases (like the one you describe) are impossible. We just went with the simplest implementation.

I'm just saying that we check that the HTLC is present in both last valid local/remote transactions. If we are mid-failing, it's gonna only be in one transaction and the function won't return it.

t-bast: Clarified in https://github.com/ACINQ/eclair/pull/1578/commits/c4db58c5fcc6659669e122cc6afa9b8cebad6e6c

pm47 (approved):

PR #1580, opened 2020-10-23
Post restart plugin htlcs 3

btcontract:
This yet another attempt to to enable plugin HTLCs, currently it's a hybrid of original https://github.com/ACINQ/eclair/pull/1577 and subsequent review suggestions:

- A suggested `PluginParams` trait is added with its sub-traits specifying custom plugin functions. This allows `Setup` to remain unchanged and then relevant `PluginParams` sub-traits can be filtered out and used  wherever needed (an example is `PostRestartHtlcCleaner`).

- `PluginWithHTLCStateChanged` event is removed and `ChannelStateChanged` event can now be used for all types of channels including non-standard ones.

- `HasAbstractCommitments` trait is retained and now extends `Data`. This is what enables a unified `ChannelStateChanged` event as well as minimal changes in `PostRestartHtlcCleaner` methods.

- `AbstractCommitments` has a new method called `getIncomingCrossSignedAndRelayedHtlcs`, also added extensive comments to clarify how it differs from similar sounding `getIncomingHtlcCrossSigned`.

An originally suggested

```
/** Parameters for a plugin that defines custom commitment transactions (or non-standard HTLCs). */
sealed trait CustomCommitmentsPlugin extends PluginParams {
  /**
   * If we do nothing after a restart, incoming HTLCs that were committed upstream but not relayed will eventually
   * expire. If your plugin defines non-standard HTLCs, and they need to be automatically failed, they should be
   * returned by this method.
   *
   * Outgoing HTLC sets that are still pending may either succeed or fail: we need to watch them to properly forward the
   * result upstream to preserve channels. If you have non-standard HTLCs that may be in this situation, they should be
   * returned by this method.
   */
  def getBrokenHtlcs: BrokenHtlcs
}
```

was not implemented because plugin can't know beforehand if its HTLCs are broken, this is something to be decided in `PostRestartHtlcCleaner` when checking all HTLCs from all channels. This is another reason for having `HasAbstractCommitments` and `getIncomingCrossSignedAndRelayedHtlcs` method in `AbstractCommitments`.


btcontract:
Another point to add is suggested https://github.com/ACINQ/eclair/pull/1577#discussion_r510183790 indeed seems to be a dead end. The only way I could think of is redefining `HasCommitments` to

```
sealed trait HasCommitments[T <: AbstractCommitments] extends HasAbstractCommitments {
  def commitments: T
}
```

but this has triggered the need to peremetrize it as `HasCommitments[Commitments]` in too many places.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
```diff
@@ -32,7 +32,7 @@ trait ChannelEvent
 
 case class ChannelCreated(channel: ActorRef, peer: ActorRef, remoteNodeId: PublicKey, isFunder: Boolean, temporaryChannelId: ByteVector32, initialFeeratePerKw: FeeratePerKw, fundingTxFeeratePerKw: Option[FeeratePerKw]) extends ChannelEvent
 
-case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, isFunder: Boolean, currentData: AbstractCommitments) extends ChannelEvent
+case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, isFunder: Boolean, currentData: HasAbstractCommitments) extends ChannelEvent
```
t-bast, line 35: Ins't it `ChannelStateChanged` that you should update?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -298,6 +306,30 @@ final case class DATA_CLOSING(commitments: Commitments,
                               revokedCommitPublished: List[RevokedCommitPublished] = Nil) extends Data with HasCommitments {
   val spendingTxes = mutualClosePublished ::: localCommitPublished.map(_.commitTx).toList ::: remoteCommitPublished.map(_.commitTx).toList ::: nextRemoteCommitPublished.map(_.commitTx).toList ::: futureRemoteCommitPublished.map(_.commitTx).toList ::: revokedCommitPublished.map(_.commitTx)
   require(spendingTxes.nonEmpty, "there must be at least one tx published in this state")
+
+  override def htlcsToFailInOrigin(log: LoggingAdapter): Set[Long] = {
```
t-bast, line 310: This has nothing to do inside `DATA_CLOSING` :thinking:

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -248,9 +250,15 @@ sealed trait Data
...
 case object Nothing extends Data
 
-sealed trait HasCommitments extends Data {
-  def commitments: Commitments
+trait HasAbstractCommitments extends Data {
+  def commitments: AbstractCommitments
   def channelId: ByteVector32 = commitments.channelId
+  /** HTLCs which can't be relayed to remote peer so need to be failed in origin channel */
+  def htlcsToFailInOrigin(log: LoggingAdapter): Set[Long] = Set.empty
+}
```
t-bast, line 258: I think we can (should) avoid introducing this new trait, it's weird that it's used almost nowhere.

Why can't you change `ChannelStateChanged` to `case class ChannelStateChanged(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, previousState: State, currentState: State, commitments: AbstractCommitments) extends ChannelEvent`? And the potential new function you need to add should be added to the `AbstractCommitments` trait, unless I'm missing something.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -48,6 +48,9 @@ case class WaitingForRevocation(nextRemoteCommit: RemoteCommit, sent: CommitSig,
 // @formatter:on
 
 trait AbstractCommitments {
+  /** Specifically return incoming HTLCs which were cross-signed AND relayed into another local channel. */
+  def getIncomingCrossSignedAndRelayedHtlcs: Set[UpdateAddHtlc]
```
t-bast, line 52: I think it's confusing that it's named that way.
I think it would be better to expose in this trait:

```
/** Returns all HTLCs in the current remote commitment. */
def htlcsRemoteCommit: Set[DirectedHtlc]
/** Returns all HTLCs in the current local commitment. */
def htlcsLocalCommit: Set[DirectedHtlc]
```

And then you compose the results with the usual `outgoing` / `incoming` partial functions.

t-bast (changes requested):

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
btcontract: I was looking where this trait would be appropriate, applying it here saves a bit of code here: https://github.com/ACINQ/eclair/pull/1580/files#diff-9242cfc225576251a04aa532dfe001172c5b6614dd9411c7c965d59457130f23L196

but indeed if `HasAbstractCommitments` is to be thrown away this event will be just fine without it.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
btcontract: It's related to `DATA_CLOSING` in a sense that here: https://github.com/ACINQ/eclair/blob/b0716aea8839bb0593293e19fb6397a578365989/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala#L325 `checkBrokenHtlcs` matches against data and applies this code if `DATA_CLOSING` is found.

This change basically removes a need for a match and opens a possibility for every `Data` type to provide its own `htlcsToFailInOrigin`. Currently it's only overridden in `DATA_CLOSING` but it would be convenient to have it in plugin `Data` types, too. This is where `HasAbstractCommitments` will come handy, plugin data types are supposed to extend it.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
btcontract: Maybe I'm missing something here, I'll once again look if this trait can be avoided a provide a list of reasons which I think prevents that (if found).

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
btcontract: As to `ChannelStateChanged`, it needs not be updated if `HasAbstractCommitments` is to stay since it uses `currentData: Data` and `ChannelStateChanged` extends from `Data` so can be published from base channels as well as plugins.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
t-bast: Oh right, I reviewed this one too quickly. Will revisit.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
btcontract: `checkBrokenHtlcs` method is the main reason `HasAbstractCommitments` exists because in order to determine which HTLCs are actually broken it needs both an instance of `AbstractCommitments` and a `Data` instance which wraps it for each channel.

Originally it takes a `Seq[HasCommitments]` and applies additional rules if wrapping `HasCommitments` instance happens to be `CLOSING_DATA`. With HTLC-enabled plugins this method needs a way to:

- access various commitment object methods (such as `htlcsRemoteCommit`, `originChannels`. Currently solved by `AbstractCommitments` abstract methods).

- know if wrapping `Data` type provides any additional rules for determining which HTLCs are broken (Currently solved by `HasAbstractCommitments` trait and its `htlcsToFailInOrigin` method which makes each data type (including plugin types) responsible for providing such rules.

Other ways of solving this that I can think of:

1. Unsealing `Data` trait (for plugin types to extend it), adding `htlcsToFailInOrigin` method to it, supplying a tuple of `(AbstractCommitments, Data)` to `checkBrokenHtlcs` method (because `Data` does not necessarily has commitments yet, so `Data`s without commitments should be filtered out). At least to me it seems less elegant than introducing `HasAbstractCommitments`.

2. Changing `HasCommitments` trait to have `AbstractCommitments` instead of `Commitments`, but https://github.com/ACINQ/eclair/pull/1580#issuecomment-715229358.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -51,9 +51,6 @@ trait AbstractCommitments {
   /** Returns all HTLCs in the current remote commitment. */
   def htlcsRemoteCommit: Set[DirectedHtlc]
 
-  /** Returns all HTLCs in the current local commitment. */
-  def htlcsLocalCommit: Set[DirectedHtlc]
```
btcontract: I'm not sure it makes sense to introduce this yet because so far it's not used anywhere except Commitments object itself.

t-bast:
I think this is still too invasive, and can be avoided with a small refactoring of the `PostRestartHtlcCleaner`.

I separated the steps to compute the broken HTLCs, and I think this makes it much simpler for plugins to inject their own: see what I've done in https://github.com/ACINQ/eclair/commit/9d63b94c3bd028e5b6fec4fbbb87ad509eea5bd0. No need for `HasAbstractCommitments`, less code changes, and we keep `PostRestartHtlcCleaner` almost unchanged (so less risks of breaking something).

Let me know if that would work for your use-case, and if so you can base your PR on this change. You will probably need to change the method parameters for the two functions I introduced in `CustomCommitmentsPlugin` to suit your needs, and add tests, but I think this should unblock you.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -91,14 +92,20 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
...
   val nodeId: PublicKey = nodeKeyManager.nodeId
-  val keyPair = KeyPair(nodeId.value, privateKey.value)
+
+  val keyPair: KeyPair = KeyPair(nodeId.value, privateKey.value)
 
   val pluginMessageTags: Set[Int] = pluginParams.collect { case p: CustomFeaturePlugin => p.messageTags }.toSet.flatten
 
+  def nonStandardIncomingHtlcs: Seq[PostRestartHtlcCleaner.IncomingHtlc] = pluginParams.collect { case p: CustomCommitmentsPlugin => p.getIncomingHtlcs }.flatten
+
+  def nonStandardRelayedOutHtlcs(htlcsIn: Seq[PostRestartHtlcCleaner.IncomingHtlc]): Map[Origin, Set[(ByteVector32, Long)]] = pluginParams.collect { case p: CustomCommitmentsPlugin => p.getHtlcsRelayedOut(htlcsIn) }.flatten.toMap
```
btcontract, line 104: This taken from this discussion: https://gitter.im/ACINQ/developers?at=5fa1289c2a3544071500e1b3

t-bast:
The diff is weird, github shows me changes for things that are already on master, am I the only one?
Can you clean the history by rebasing on top of master and squashing everything in a single new commit?

PR #1582, opened 2020-10-27
Minor refactoring

btcontract:
This turns a piece of channel code into a separate method, the reason is it's going to be reused exactly in HC.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -97,15 +97,11 @@ object Channel {
    */
   case class OutgoingMessage(msg: LightningMessage, peerConnection: ActorRef)
 
-  def failTimedout(adds: Set[UpdateAddHtlc], relayer: ActorRef, commits: AbstractCommitments, log: akka.event.LoggingAdapter): Unit =
+  def failHtlcs(adds: Set[UpdateAddHtlc], commits: AbstractCommitments, log: akka.event.LoggingAdapter)(onFound: (Origin, UpdateAddHtlc) => Unit): Unit =
     adds.foreach { add =>
       commits.originChannels.get(add.id) match {
-        case Some(origin) =>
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -97,15 +97,11 @@ object Channel {
    */
   case class OutgoingMessage(msg: LightningMessage, peerConnection: ActorRef)
 
-  def failTimedout(adds: Set[UpdateAddHtlc], relayer: ActorRef, commits: AbstractCommitments, log: akka.event.LoggingAdapter): Unit =
+  def failHtlcs(adds: Set[UpdateAddHtlc], commits: AbstractCommitments, log: akka.event.LoggingAdapter)(onFound: (Origin, UpdateAddHtlc) => Unit): Unit =
```
btcontract, line 100: This goes a bit further since failing timed out and overridden HTLCs code is similar. If this is too much I'll undo this change.

t-bast (commented):
Can you explain why it's useful to have this factorized for HC?
The resulting `failHtlcs` is just a wapper around `commitments.originChannels.get(add.it)`, it doesn't feel like a big gain.

While it's true that there is some duplication, we want to preserve the distinct logs and since it's a rather critical piece of code, I'd rather keep all the code in the same place to help readability.

btcontract:
It's just to avoid duplication, nothing beyond that. What about more straightforward https://github.com/ACINQ/eclair/commit/7bbcf66ee746396aa791bf5f4cb36b537a1b0d7f?

t-bast:
TBH I'm fine with very small duplication (such as here) when the minor differences are painful to refactor (ie in this case the log messages).
I don't see the need for this refactoring for the following reasons:

- I like keeping all messages sent to actor directly visible in the handler (so no moving `relayer ! ...` elsewhere), this makes it easy to see side effects in only one read (without jumping in and out of function calls)
- I want to reasonably avoid having to jump inside too many different functions when reading the code for a critical handler such as this one

btcontract:
OK, will keep this in mind in future PRs.

PR #1584, opened 2020-10-30
Segregate the node seed from the channel seed

dkrm0:
# Segregate the node seed from the channel seed

We want to have 2 separated seeds:
- one for the node (to sign protocol messages, etc.)
- one for the channels (to sign commit txs, etc.)

This PR creates 2 new files, `nodeSeed.dat` & `channelSeed.dat`, both stored in `$datadir`.

## Migration from existing setup
This flow is executed twice, for `nodeSeed.dat` and `channelSeed.dat`:
```
+----------------------------+   Yes   +--------------------+
| Does `newSeed.dat` exists? +-------->+ Use that seed file |
+------------+---------------+         +---+----------------+
             |                             ^
             | No                          |
             v                             |
+------------+---------------+   Yes   +---+----------------------+
| Does `seed.dat` exists?    +-------->+ Copy it to `newSeed.dat` |
+------------+---------------+         +---+----------------------+
             |                             ^
             | No                          |
             v                             |
+------------+-------------+               |
| Generate a random seed   |               |
| into a file              +---------------+
+--------------------------+
```

The old `seed.dat` is not removed after the migration.


eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -118,18 +120,49 @@ object NodeParams {
...
+    logger.info(s"use seed file: ${seedPath.getCanonicalPath}")
+    ByteVector(Files.readAllBytes(seedPath.toPath))
+  }
+
+  private def writeSeedToFile(path: File, seed: ByteVector): Unit = {
+    Files.write(path.toPath, seed.toArray)
+    logger.info(s"create new seed file: ${path.getCanonicalPath}")
+  }
+
+  private def migrateSeedFile(old: File, `new`: File): Unit = {
```
t-bast, line 133: I think the migration is misleading. We cannot really migrate an existing node, so maybe we should instead be explicit about that?

Using the same seed for node and channel means we didn't segregate anything and if the user moves his `channelSeed.dat` to a hardware wallet, he will get a false sense of security.

It also won't make sense to generate the channel seed on a normal machine, it should be generated by a hardware wallet, shouldn't it?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -102,7 +102,7 @@ object Channel {
 class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId: PublicKey, blockchain: ActorRef, relayer: ActorRef, origin_opt: Option[ActorRef] = None)(implicit ec: ExecutionContext = ExecutionContext.Implicits.global) extends FSM[State, Data] with FSMDiagnosticActorLogging[State, Data] {
 
   import Channel._
-  import nodeParams.keyManager
+  import nodeParams.channelKeyManager
```
t-bast, line 105: What about:

```
private val keyManager = nodeParams.channelKeyManager
```

So that all the other changes to this file are unnecessary :)

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -519,20 +519,20 @@ object Commitments {
 
   def revocationHash(seed: ByteVector32, index: Long): ByteVector32 = Crypto.sha256(revocationPreimage(seed, index))
 
-  def sendCommit(commitments: Commitments, keyManager: KeyManager)(implicit log: LoggingAdapter): Try[(Commitments, CommitSig)] = {
+  def sendCommit(commitments: Commitments, channelKeyManager: ChannelKeyManager)(implicit log: LoggingAdapter): Try[(Commitments, CommitSig)] = {
```
t-bast, line 522: Same here (and everywhere in the PR). I think we should keep the name of these variables `keyManager`, as long as the type is correct we don't need the variable name change. It also removes noise from the PR.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChannelKeyManager.scala
```diff
@@ -109,23 +106,10 @@ trait KeyManager {
...
-   * When recovering a public key from an ECDSA signature for secp256k1, there are 4 possible matching curve points
-   * that can be found. The recoveryId identifies which of these points is the correct.
-   *
-   * @param digest     SHA256 digest
-   * @param privateKey private key to sign with, default the one from the current node
-   * @return a (signature, recoveryId) pair. signature is a signature of the digest parameter generated with the
-   *         private key given in parameter. recoveryId is the corresponding recoveryId of the signature
-   */
-  def signDigest(digest: ByteVector32, privateKey: PrivateKey = nodeKey.privateKey): (ByteVector64, Int)
+  def signChannelAnnouncement(localNodeSecret: PrivateKey, fundingKeyPath: DeterministicWallet.KeyPath, chainHash: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, remoteFundingKey: PublicKey, features: Features): (ByteVector64, ByteVector64)
```
t-bast, line 109: This function becomes a bit weird now, doesn't it?
Maybe we should clearly separate signing the channel announcement with the node's key and the channel key; this will result in two distinct `signChannelAnnouncement` methods, each returning only one signature, one defined in `ChannelKeyManager` and the other in `NodeKeyManager`.

sstone (dismissed):

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
sstone: Yes but reusing an existing seed that was used on a regular computer (moving i into to a hardware wallet for example) is not safe anyway ?

t-bast: I agree, maybe it's the term "migrating" that I find misleading there, but it may make sense here...
I need to think about this more.

t-bast: Ok so what we're doing for now is migrating nodes from a single seed to two seeds, so the wording makes sense.
I guess later we'll provide tutorials and additional changes to initialize a node with a hardware device that handles all the `channelSeed` part.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -65,13 +65,13 @@ import scala.util.{Failure, Success}
...
- * @param seed_opt optional seed, if set eclair will use it instead of generating one and won't create a seed.dat file.
- * @param db       optional databases to use, if not set eclair will create the necessary databases
+ * @param datadir   directory where eclair-core will write/read its data.
+ * @param seeds_opt optional seeds, if set eclair will use them instead of generating them and won't create a nodeSeed.dat and channelSeed.dat files.
+ * @param db        optional databases to use, if not set eclair will create the necessary databases
  */
 class Setup(datadir: File,
             pluginParams: Seq[PluginParams],
-            seed_opt: Option[ByteVector] = None,
+            seeds_opt: Option[(ByteVector, ByteVector)] = None,
```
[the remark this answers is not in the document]
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -118,18 +120,49 @@ object NodeParams {
...
+    if (old.exists() && !`new`.exists()) {
+      Files.copy(old.toPath, `new`.toPath)
+      logger.info(s"migrate seed file: ${old.getCanonicalPath} → ${`new`.getCanonicalPath}")
     }
   }
 
+  def getSeeds(datadir: File, nodeSeedFilename: String = "nodeSeed.dat", channelSeedFilename: String = "channelSeed.dat"): (ByteVector, ByteVector) = {
```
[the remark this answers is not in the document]
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -118,18 +120,49 @@ object NodeParams {
...
+    // Previously we used one seed file ("seed.dat") to generate the node and the channel private keys
+    // Now we use two separate files and thus we need to migrate the old seed file if necessary
+    val oldSeedPath = new File(datadir, "seed.dat")
+
+    def getSeed(filename: String): ByteVector = {
+      val seedPath = new File(datadir, filename)
+      if (seedPath.exists()) {
+        readSeedFromFile(seedPath)
+      }
+      else if (oldSeedPath.exists()) {
```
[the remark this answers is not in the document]
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChannelKeyManager.scala
dkrm0: See d7418a97
Does it seems right to you?

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -118,18 +121,49 @@ object NodeParams {
...
+      logger.info(s"migrate seed file: ${old.getCanonicalPath} → ${`new`.getCanonicalPath}")
     }
   }
 
+  def getSeeds(datadir: File): Seeds = {
+    // Previously we used one seed file ("seed.dat") to generate the node and the channel private keys
+    // Now we use two separate files and thus we need to migrate the old seed file if necessary
+    val oldSeedPath = new File(datadir, "seed.dat")
+    val nodeSeedFilename: String = "nodeSeed.dat"
+    val channelSeedFilename: String = "channelSeed.dat"
```
pm47, line 146: I don't know why, but camel case doesn't work well for files.

```suggestion
    val nodeSeedFilename: String = "node_seed.dat"
    val channelSeedFilename: String = "channel_seed.dat"
```

pm47 (commented):
I would move all the `*KeyManager` files to a sub-package of `crypto`.

Otherwise seems ok to me.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -118,18 +121,49 @@ object NodeParams {
...
+    logger.info(s"use seed file: ${seedPath.getCanonicalPath}")
+    ByteVector(Files.readAllBytes(seedPath.toPath))
+  }
+
+  private def writeSeedToFile(path: File, seed: ByteVector): Unit = {
+    Files.write(path.toPath, seed.toArray)
+    logger.info(s"create new seed file: ${path.getCanonicalPath}")
+  }
+
+  private def migrateSeedFile(old: File, `new`: File): Unit = {
```
t-bast, line 134: nit: I'm not a fan of naming a variable `new` and having to use the backtick workaround...
Why not (to match `Files.copy`):

```suggestion
  private def migrateSeedFile(source: File, destination: File): Unit = {
```

t-bast (commented):
LGTM, tiny nit

eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChannelKeyManager.scala
t-bast: Yes that's perfect.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
dkrm0: Done

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
dkrm0: Done

t-bast (approved):
LGTM

pm47 (approved):

PR #1585, opened 2020-11-04
Improve channel exceptions

t-bast:
We were heavily using `Try`, which unfortunately loses the type of exception (in our case `ChannelException`) and prevents us from doing a complete pattern matching on all error cases.

I'm replacing that with an `Either[ChannelException, X]` where possible. The goal is to slowly move away from untyped exceptions.

This may be a dangerous change, so please review carefully!

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -290,9 +290,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   when(WAIT_FOR_OPEN_CHANNEL)(handleExceptions {
     case Event(open: OpenChannel, d@DATA_WAIT_FOR_OPEN_CHANNEL(INPUT_INIT_FUNDEE(_, localParams, _, remoteInit, channelVersion))) =>
       log.info("received OpenChannel={}", open)
-      Try(Helpers.validateParamsFundee(nodeParams, localParams.features, open, remoteNodeId)) match {
-        case Failure(t) => handleLocalError(t, d, Some(open))
-        case Success(_) =>
+      Helpers.validateParamsFundee(nodeParams, localParams.features, open, remoteNodeId) match {
+        case Some(t) => handleLocalError(t, d, Some(open))
+        case None =>
           context.system.eventStream.publish(ChannelCreated(self, peer, remoteNodeId, isFunder = false, open.temporaryChannelId, open.feeratePerKw, None))
           val fundingPubkey = keyManager.fundingPublicKey(localParams.fundingKeyPath).publicKey
           val channelKeyPath = keyManager.keyPath(localParams, channelVersion)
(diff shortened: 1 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -97,80 +97,84 @@ object Helpers {
   /**
    * Called by the fundee
    */
-  def validateParamsFundee(nodeParams: NodeParams, features: Features, open: OpenChannel, remoteNodeId: PublicKey): Unit = {
+  def validateParamsFundee(nodeParams: NodeParams, features: Features, open: OpenChannel, remoteNodeId: PublicKey): Option[ChannelException] = {
```
pm47, line 100: For consistency, and to better convey the idea that we expect a `Right`?

```suggestion
  def validateParamsFundee(nodeParams: NodeParams, features: Features, open: OpenChannel, remoteNodeId: PublicKey): Left[ChannelException, Done] = {
```

NB: this is the builtin `akka.Done`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -389,23 +389,26 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-        fundingTxid = fundingTx.hash,
-        fundingOutputIndex = fundingTxOutputIndex,
-        signature = localSigOfRemoteTx
-      )
-      val channelId = toLongId(fundingTx.hash, fundingTxOutputIndex)
-      peer ! ChannelIdAssigned(self, remoteNodeId, temporaryChannelId, channelId) // we notify the peer asap so it knows how to route messages
-      context.system.eventStream.publish(ChannelIdAssigned(self, remoteNodeId, temporaryChannelId, channelId))
-      // NB: we don't send a ChannelSignatureSent for the first commit
-      goto(WAIT_FOR_FUNDING_SIGNED) using DATA_WAIT_FOR_FUNDING_SIGNED(channelId, localParams, remoteParams, fundingTx, fundingTxFee, localSpec, localCommitTx, RemoteCommit(0, remoteSpec, remoteCommitTx.tx.txid, remoteFirstPerCommitmentPoint), open.channelFlags, channelVersion, fundingCreated) sending fundingCreated
+      Funding.makeFirstCommitTxs(keyManager, channelVersion, temporaryChannelId, localParams, remoteParams, fundingAmount, pushMsat, initialFeeratePerKw, fundingTx.hash, fundingTxOutputIndex, remoteFirstPerCommitmentPoint) match {
```
pm47, line 394: Uh, didn't realize we were relying on the top-level `handleExceptions` for that one!

pm47 (commented):
The risk I see is for unexpected exceptions, that were previously caught by the enclosing `Try` and will now escape the `Either`. I'm particularly looking at the `Commitments.sendCommit` method.

However we still have a higher level `handleExceptions`, so I think we are fine.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Yes, that was on my todo list when getting back to this PR, it should be an Either[ChannelException, Done]` or `Either[ChannelException, Unit]`. Let me fix that!

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Yep, but I think that since we have that top-level exception handler, we should try to remove all exceptional code below it.
It will help us figure out where some lower level functions should be wrapped into an `Either` or a `Try`.

pm47: Wholly agree.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1585/commits/737b51da59fa27b795ba2edeb27ce4fc47efc1c8

pm47 (approved):

PR #1586, opened 2020-11-05
htlc post restart 4

btcontract:
This is continuation of https://github.com/ACINQ/eclair/pull/1580 (rather than wrestling with git I've decided to send a new PR). It follows suggestions outlined in https://github.com/ACINQ/eclair/commit/9d63b94c3bd028e5b6fec4fbbb87ad509eea5bd0.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -29,12 +29,11 @@ import fr.acinq.bitcoin.{Block, ByteVector32, Crypto, Satoshi}
 import fr.acinq.eclair.NodeParams.WatcherType
 import fr.acinq.eclair.Setup.Seeds
 import fr.acinq.eclair.blockchain.fee.{FeeEstimator, FeeTargets, FeerateTolerance, OnChainFeeConf}
-import fr.acinq.eclair.channel.{Channel, Origin}
+import fr.acinq.eclair.channel.Channel
 import fr.acinq.eclair.crypto.Noise.KeyPair
 import fr.acinq.eclair.crypto.keymanager.{ChannelKeyManager, NodeKeyManager}
 import fr.acinq.eclair.db._
 import fr.acinq.eclair.io.PeerConnection
-import fr.acinq.eclair.payment.relay.PostRestartHtlcCleaner
 import fr.acinq.eclair.router.Router.RouterConf
 import fr.acinq.eclair.tor.Socks5ProxyParams
 import fr.acinq.eclair.wire.{Color, EncodingType, NodeAddress}
@@ -98,10 +97,6 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
   val keyPair: KeyPair = KeyPair(nodeId.value, privateKey.value)
 
   val pluginMessageTags: Set[Int] = pluginParams.collect { case p: CustomFeaturePlugin => p.messageTags }.toSet.flatten
-
-  def nonStandardIncomingHtlcs: Seq[PostRestartHtlcCleaner.IncomingHtlc] = pluginParams.collect { case p: CustomCommitmentsPlugin => p.getIncomingHtlcs }.flatten
-
-  def nonStandardRelayedOutHtlcs(htlcsIn: Seq[PostRestartHtlcCleaner.IncomingHtlc]): Map[Origin, Set[(ByteVector32, Long)]] = pluginParams.collect { case p: CustomCommitmentsPlugin => p.getHtlcsRelayedOut(htlcsIn) }.flatten.toMap
 
   def currentBlockHeight: Long = blockCount.get
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -94,7 +94,7 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
   val nodeId: PublicKey = nodeKeyManager.nodeId
   val keyPair = KeyPair(nodeId.value, privateKey.value)
 
-  val pluginMessageTags: Set[Int] = pluginParams.collect { case p: MessageFeaturePluginParams => p.messageTags }.toSet.flatten
+  val pluginMessageTags: Set[Int] = pluginParams.collect { case p: CustomFeaturePlugin => p.messageTags }.toSet.flatten
 
   def currentBlockHeight: Long = blockCount.get
 
@@ -241,7 +241,7 @@ object NodeParams extends Logging {
       require(features.hasFeature(Features.VariableLengthOnion), s"${Features.VariableLengthOnion.rfcName} must be enabled")
     }
 
-    val pluginMessageParams = pluginParams.collect { case p: MessageFeaturePluginParams => p }
+    val pluginMessageParams = pluginParams.collect { case p: CustomFeaturePlugin => p }
     val features = Features.fromConfiguration(config)
     validateFeatures(features)
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -94,7 +94,7 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
   val nodeId: PublicKey = nodeKeyManager.nodeId
   val keyPair = KeyPair(nodeId.value, privateKey.value)
 
-  val pluginMessageTags: Set[Int] = pluginParams.flatMap(_.tags).toSet
+  val pluginMessageTags: Set[Int] = pluginParams.collect { case p: MessageFeaturePluginParams => p.messageTags }.toSet.flatten
 
   def currentBlockHeight: Long = blockCount.get
 
@@ -241,23 +241,24 @@ object NodeParams extends Logging {
       require(features.hasFeature(Features.VariableLengthOnion), s"${Features.VariableLengthOnion.rfcName} must be enabled")
     }
 
+    val pluginMessageParams = pluginParams.collect { case p: MessageFeaturePluginParams => p }
     val features = Features.fromConfiguration(config)
     validateFeatures(features)
 
-    require(pluginParams.forall(_.feature.mandatory > 128), "Plugin mandatory feature bit is too low, must be > 128")
-    require(pluginParams.forall(_.feature.mandatory % 2 == 0), "Plugin mandatory feature bit is odd, must be even")
-    require(pluginParams.flatMap(_.tags).forall(_ > 32768), "Plugin messages tags must be > 32768")
-    val pluginFeatureSet = pluginParams.map(_.feature.mandatory).toSet
+    require(pluginMessageParams.forall(_.feature.mandatory > 128), "Plugin mandatory feature bit is too low, must be > 128")
+    require(pluginMessageParams.forall(_.feature.mandatory % 2 == 0), "Plugin mandatory feature bit is odd, must be even")
+    require(pluginMessageParams.flatMap(_.messageTags).forall(_ > 32768), "Plugin messages tags must be > 32768")
+    val pluginFeatureSet = pluginMessageParams.map(_.feature.mandatory).toSet
     require(Features.knownFeatures.map(_.mandatory).intersect(pluginFeatureSet).isEmpty, "Plugin feature bit overlaps with known feature bit")
-    require(pluginFeatureSet.size == pluginParams.size, "Duplicate plugin feature bits found")
+    require(pluginFeatureSet.size == pluginMessageParams.size, "Duplicate plugin feature bits found")
 
-    val coreAndPluginFeatures = features.copy(unknown = features.unknown ++ pluginParams.map(_.pluginFeature))
+    val coreAndPluginFeatures = features.copy(unknown = features.unknown ++ pluginMessageParams.map(_.pluginFeature))
 
     val overrideFeatures: Map[PublicKey, Features] = config.getConfigList("override-features").asScala.map { e =>
       val p = PublicKey(ByteVector.fromValidHex(e.getString("nodeid")))
       val f = Features.fromConfiguration(e)
       validateFeatures(f)
-      p -> f.copy(unknown = f.unknown ++ pluginParams.map(_.pluginFeature))
+      p -> f.copy(unknown = f.unknown ++ pluginMessageParams.map(_.pluginFeature))
     }.toMap
 
     val syncWhitelist: Set[PublicKey] = config.getStringList("sync-whitelist").asScala.map(s => PublicKey(ByteVector.fromValidHex(s))).toSet
@@ -381,14 +382,4 @@ object NodeParams extends Logging {
       enableTrampolinePayment = config.getBoolean("trampoline-payments-enable")
     )
   }
-}
-
-/**
- * @param tags    a set of LightningMessage tags that plugin is interested in
- * @param feature a Feature bit that plugin advertises through Init message
- */
-case class PluginParams(tags: Set[Int], feature: Feature) {
-  def pluginFeature: UnknownFeature = UnknownFeature(feature.optional)
-
-  override def toString: String = s"Messaging enabled plugin=${feature.rfcName} with feature bit=${feature.optional} and LN message tags=${tags.mkString(",")}"
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -29,11 +29,12 @@ import fr.acinq.bitcoin.{Block, ByteVector32, Crypto, Satoshi}
 import fr.acinq.eclair.NodeParams.WatcherType
 import fr.acinq.eclair.Setup.Seeds
 import fr.acinq.eclair.blockchain.fee.{FeeEstimator, FeeTargets, FeerateTolerance, OnChainFeeConf}
-import fr.acinq.eclair.channel.Channel
+import fr.acinq.eclair.channel.{Channel, Origin}
 import fr.acinq.eclair.crypto.Noise.KeyPair
 import fr.acinq.eclair.crypto.keymanager.{ChannelKeyManager, NodeKeyManager}
 import fr.acinq.eclair.db._
 import fr.acinq.eclair.io.PeerConnection
+import fr.acinq.eclair.payment.relay.PostRestartHtlcCleaner
 import fr.acinq.eclair.router.Router.RouterConf
 import fr.acinq.eclair.tor.Socks5ProxyParams
 import fr.acinq.eclair.wire.{Color, EncodingType, NodeAddress}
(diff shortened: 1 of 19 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -91,14 +92,20 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
...
-  val keyPair = KeyPair(nodeId.value, privateKey.value)
 
-  val pluginMessageTags: Set[Int] = pluginParams.flatMap(_.tags).toSet
+  val keyPair: KeyPair = KeyPair(nodeId.value, privateKey.value)
+
+  val pluginMessageTags: Set[Int] = pluginParams.collect { case p: CustomFeaturePlugin => p.messageTags }.toSet.flatten
+
+  def nonStandardIncomingHtlcs: Seq[PostRestartHtlcCleaner.IncomingHtlc] = pluginParams.collect { case p: CustomCommitmentsPlugin => p.getIncomingHtlcs }.flatten
+
+  def nonStandardRelayedOutHtlcs(htlcsIn: Seq[PostRestartHtlcCleaner.IncomingHtlc]): Map[Origin, Set[(ByteVector32, Long)]] = pluginParams.collect { case p: CustomCommitmentsPlugin => p.getHtlcsRelayedOut(htlcsIn) }.flatten.toMap
```
btcontract, line 104: This taken from this discussion: https://gitter.im/ACINQ/developers?at=5fa1289c2a3544071500e1b3

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -58,10 +56,23 @@ class PostRestartHtlcCleaner(nodeParams: NodeParams, register: ActorRef, initial
...
+  // this and fast-fail those HTLCs and thus preserve channels.
+  //
+  // Outgoing HTLC sets that are still pending may either succeed or fail: we need to watch them to properly forward the
+  // result upstream to preserve channels.
+  val brokenHtlcs: BrokenHtlcs = {
     val channels = listLocalChannels(nodeParams.db.channels)
-    cleanupRelayDb(channels, nodeParams.db.pendingRelay)
-    checkBrokenHtlcs(channels, nodeParams.db.payments, nodeParams.privateKey)
+    val htlcsIn: Seq[IncomingHtlc] = getIncomingHtlcs(channels, nodeParams.db.payments, nodeParams.privateKey) ++ nodeParams.nonStandardIncomingHtlcs
+    // TODO: this part needs unit tests to verify maps are correctly merged
```
btcontract, line 68: Meaning tests in plugin? On base layer I can't think of any interesting test (I'd have to supply an HTLC which either matches or not?)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -290,35 +301,33 @@ object PostRestartHtlcCleaner {
...
-  /**
-   * If we do nothing after a restart, incoming HTLCs that were committed upstream but not relayed will eventually
-   * expire and we won't lose money, but the channel will get closed, which is a major inconvenience. We want to detect
-   * this and fast-fail those HTLCs and thus preserve channels.
-   *
-   * Outgoing HTLC sets that are still pending may either succeed or fail: we need to watch them to properly forward the
-   * result upstream to preserve channels.
-   */
-  private def checkBrokenHtlcs(channels: Seq[HasCommitments], paymentsDb: IncomingPaymentsDb, privateKey: PrivateKey)(implicit log: LoggingAdapter): BrokenHtlcs = {
+  def remoteOutgoingToLocalIncoming(paymentsDb: IncomingPaymentsDb): PartialFunction[Either[FailureMessage, IncomingPacket], IncomingHtlc] = {
```
t-bast, line 304: Rename to `decryptedIncomingHtlcs`? I'm not a big fan of exposing partial functions like that, but I guess you want to re-use it in your plugin?

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -540,7 +540,7 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
         system.eventStream.publish(chcr)
         wsClient.expectMessage(expectedSerializedChcr)
 
-        val chsc = ChannelStateChanged(system.deadLetters, system.deadLetters, bobNodeId, OFFLINE, NORMAL, null)
+        val chsc = ChannelStateChanged(system.deadLetters, null, system.deadLetters, bobNodeId, OFFLINE, NORMAL, null)
```
t-bast, line 543: What's nice is that now we'll be able to easily add the `channelId` here! It's not your job to do it in this PR though, I'll do it in a follow-up one.

t-bast (commented):
We're getting close, this time is the right one ;)
Have you been able to test that this correctly gives you the hooks you need for HC?

btcontract:
> Have you been able to test that this correctly gives you the hooks you need for HC?

Yes, `getIncomingHtlcs` and `getHtlcsRelayedOut` methods are very close to how it works now (where HC is part of core).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
btcontract: Yes, it also uses private `shouldFulfill` method so both of them will be reused.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -44,22 +44,6 @@ import scala.util.{Failure, Success, Try}
  */
 
 object Helpers {
-
-  /**
-   * Depending on the state, returns the current temporaryChannelId or channelId
-   *
-   * @return the long identifier of the channel
-   */
-  def getChannelId(stateData: Data): ByteVector32 = stateData match {
```
btcontract, line 53: what about this transfer of `channelId` from special method to `Data` perameter overridden where appropriate? Seems to make a code a bit clearer in `Channel`.

t-bast: I like it, it's better than what we had before :+1: IMO

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1729,6 +1722,14 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
       }
 
+      if (state != nextState) {
+        val commitments_opt = nextStateData match {
+          case hasCommitments: HasCommitments => Some(hasCommitments.commitments)
+          case _ => None
+        }
+        context.system.eventStream.publish(ChannelStateChanged(self, nextStateData.channelId, peer, remoteNodeId, state, nextState, commitments_opt))
+      }
+
```
t-bast, line 1732: There's no need to change where this is done, can you put it back at the beginning of the handler?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -244,13 +244,15 @@ object ChannelOpenResponse {
       8888888P" d88P     888     888  d88P     888
  */
 
-sealed trait Data
+sealed trait Data {
+  def channelId: ByteVector32 = ByteVector32.Zeroes
```
t-bast, line 248: This shouldn't have a default value, that's too dangerous.
We must force all children to implement it explicitly (`Nothing` will be the only one returning `ByteVector32.Zeroes`).

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/GUIUpdater.scala
```diff
@@ -101,11 +101,11 @@ class GUIUpdater(mainController: MainController) extends Actor with ActorLogging
       val channelPaneController = m(channel)
       runInGuiThread(() => channelPaneController.channelId.setText(channelId.toHex))
 
-    case ChannelStateChanged(channel, _, _, _, currentState, currentData) if m.contains(channel) =>
+    case ChannelStateChanged(channel, _, _, _, _, currentState, Some(commitments)) if m.contains(channel) =>
```
t-bast, line 104: There's a small regression here:
```suggestion
    case ChannelStateChanged(channel, _, _, _, _, currentState, commitments_opt) if m.contains(channel) =>
```

eclair-node-gui/src/main/scala/fr/acinq/eclair/gui/GUIUpdater.scala
```diff
@@ -101,11 +101,11 @@ class GUIUpdater(mainController: MainController) extends Actor with ActorLogging
...
       val channelPaneController = m(channel)
       runInGuiThread { () =>
-        (currentState, currentData) match {
-          case (WAIT_FOR_FUNDING_CONFIRMED, d: HasCommitments) => channelPaneController.txId.setText(d.commitments.commitInput.outPoint.txid.toHex)
+        (currentState, commitments) match {
+          case (WAIT_FOR_FUNDING_CONFIRMED, c: Commitments) => channelPaneController.txId.setText(c.commitInput.outPoint.txid.toHex)
```
t-bast, line 108: And here:

```suggestion
        (currentState, commitments_opt) match {
          case (WAIT_FOR_FUNDING_CONFIRMED, Some(c: Commitments)) => channelPaneController.txId.setText(c.commitInput.outPoint.txid.toHex)
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -543,6 +544,99 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+    val c = ChannelCodecsSpec.makeChannelDataNormal(List(relayedhtlc1Out), Map(0L -> trampolineRelayed))
+    nodeParams1.db.channels.addOrUpdateChannel(c)
+
+    val channel = TestProbe()
+    f.createRelayer(nodeParams1)
+    register.expectNoMsg(100 millis) // nothing should happen while channels are still offline.
+
+    val cs = new AbstractCommitments {
+      def getOutgoingHtlcCrossSigned(htlcId: Long): Option[UpdateAddHtlc] = None
+      def getIncomingHtlcCrossSigned(htlcId: Long): Option[UpdateAddHtlc] = Some(relayedHtlc1In.add)
```
t-bast, line 573: This should return both `relayedHtlc1In` and `nonRelayedHtlc2In` for the test to be accurate. Or always return `None`, since it's in fact not called for your non-standard channel.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -543,6 +544,99 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+    val relayedPaymentHash = randomBytes32
+    val trampolineRelayedPaymentHash = randomBytes32
+    val trampolineRelayed = Origin.TrampolineRelayedCold((channelId_ab_2, 0L) :: Nil)
+    val relayedHtlcIn = buildHtlcIn(0L, channelId_ab_2, trampolineRelayedPaymentHash)
+    val nonRelayedHtlcIn = buildHtlcIn(1L, channelId_ab_2, relayedPaymentHash)
+
+    val pluginParams = new CustomCommitmentsPlugin {
+      def name = "test with outgoing HTLC to remote"
+      def getIncomingHtlcs: Seq[PostRestartHtlcCleaner.IncomingHtlc] = List.empty
+      def getHtlcsRelayedOut(htlcsIn: Seq[PostRestartHtlcCleaner.IncomingHtlc]): Map[Origin, Set[(ByteVector32, Long)]] = Map(trampolineRelayed -> Set((channelId_ab_2, 0L)))
```
t-bast, line 603: It doesn't make sense that htlc #0 in `channel_ab_2` has been relayed as htlc #0 in `channel_ab_2`...

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -543,6 +544,99 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+    f.createRelayer(nodeParams1)
+    register.expectNoMsg(100 millis) // nothing should happen while channels are still offline.
+
+    // Standard channel goes to NORMAL state:
+    system.eventStream.publish(ChannelStateChanged(channel.ref, c.commitments.channelId, system.deadLetters, a, OFFLINE, NORMAL, Some(c.commitments)))
+    channel.expectMsg(CMD_FAIL_HTLC(1L, Right(TemporaryNodeFailure), commit = true))
+    channel.expectNoMsg(100 millis)
+  }
+
+  test("Non-standard HTLC CMD_FAIL in relayDb is retained") { f =>
```
t-bast, line 620: This test would be more interesting if it also had a command that was cleaned up.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -543,6 +544,99 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+      def getIncomingHtlcs: Seq[PostRestartHtlcCleaner.IncomingHtlc] = List(PostRestartHtlcCleaner.IncomingHtlc(relayedHtlc1In.add, None))
+      def getHtlcsRelayedOut(htlcsIn: Seq[PostRestartHtlcCleaner.IncomingHtlc]): Map[Origin, Set[(ByteVector32, Long)]] = Map.empty
+    }
+
+    val cmd = CMD_FAIL_HTLC(id = 0L, reason = Left(ByteVector.empty), replyTo_opt = None)
+    val nodeParams1 = nodeParams.copy(pluginParams = List(pluginParams))
+    nodeParams1.db.pendingRelay.addPendingRelay(channelId_ab_1, cmd)
+    f.createRelayer(nodeParams1)
+    register.expectNoMsg(100 millis)
+    assert(nodeParams1.db.pendingRelay.listPendingRelay(channelId_ab_1).head == cmd)
```
t-bast, line 637: You need an `awaitCond` otherwise there's a race condition (you might hit your assert before the PostRestartHtlcCleaner has the opportunity to call `cleanUpRelayDb`).

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
btcontract: Indeed, but the method does get called here: https://github.com/ACINQ/eclair/blob/b0716aea8839bb0593293e19fb6397a578365989/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala#L82

t-bast: Right, so it really needs to return `nonRelayedHtlc2In`

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -543,6 +544,105 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+    val nonRelayedHtlc2In = buildHtlcIn(1L, channelId_ab_1, relayedPaymentHash)
+
+    val pluginParams = new CustomCommitmentsPlugin {
+      def name = "test with incoming HTLC from remote"
+      def getIncomingHtlcs: Seq[PostRestartHtlcCleaner.IncomingHtlc] = List(PostRestartHtlcCleaner.IncomingHtlc(relayedHtlc1In.add, None), PostRestartHtlcCleaner.IncomingHtlc(nonRelayedHtlc2In.add, None))
+      def getHtlcsRelayedOut(htlcsIn: Seq[PostRestartHtlcCleaner.IncomingHtlc]): Map[Origin, Set[(ByteVector32, Long)]] = Map.empty
+    }
+
+    val nodeParams1 = nodeParams.copy(pluginParams = List(pluginParams))
+    val c = ChannelCodecsSpec.makeChannelDataNormal(List(relayedhtlc1Out), Map(50L -> trampolineRelayed))
```
t-bast, line 564: Why did you change the htlcId to 50? It doesn't make any sense now...this one is supposed to map `relayedHtlc1Out` to `relayedHtlc1In` so it must be `0L -> trampolineRelayed`.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
```diff
@@ -543,6 +544,105 @@ class PostRestartHtlcCleanerSpec extends TestKitBaseClass with FixtureAnyFunSuit
...
+
+    val channel = TestProbe()
+    f.createRelayer(nodeParams1)
+    register.expectNoMsg(100 millis) // nothing should happen while channels are still offline.
+
+    val cs = new AbstractCommitments {
+      def getOutgoingHtlcCrossSigned(htlcId: Long): Option[UpdateAddHtlc] = None
+      def getIncomingHtlcCrossSigned(htlcId: Long): Option[UpdateAddHtlc] = {
+        if (htlcId == 0L) Some(relayedHtlc1In.add)
+        else if (htlcId == 1L) Some(relayedHtlc1In.add)
```
t-bast: Come on...

```suggestion
        else if (htlcId == 1L) Some(relayedHtlc2In.add)
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/PostRestartHtlcCleanerSpec.scala
btcontract: Strange, I thought that number is local HTLC id in a channel so i can differ from HTLC id in origin. Also wonder why the test still works.

btcontract: OK, that number does not matter here, it's used at https://github.com/ACINQ/eclair/blob/83b50347447266537426f905affb656cfb1e17bc/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala#L362 to match `htlcsToIgnore` which are empty in this test.

t-bast (approved):

PR #1587, opened 2020-11-05
Only sync with channel peers

t-bast:
Our previous behaviour was to do a full routing table sync whenever we connected to a node (or a node connected to us).
This was wasting a lot of bandwidth for no reason.

Instead, we only sync with nodes that are either:

- in our sync whitelist
- have channels with us

On top of that, we trigger a sync whenever our first channel is opened with a new peer: this ensures we following flow works:

- connect to a new node
- we don't have channels together yet, so we're not syncing
- open a channel
- once the channel is open, sync the routing table
- we can now make payments (yay!)

The syncing state is a bit hard to read: we've started discussions in the spec to make changes to channel queries (see https://github.com/lightningnetwork/lightning-rfc/issues/804 and https://github.com/lightningnetwork/lightning-rfc/issues/811), it will be a good opportunity to refactor the code later.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -166,7 +166,7 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
         log.info(s"channel id switch: previousId=$temporaryChannelId nextId=$channelId")
         // we have our first channel with that peer: let's sync our routing table
         if (!d.channels.keys.exists(_.isInstanceOf[FinalChannelId])) {
-          d.peerConnection ! PeerConnection.DoSync
+          d.peerConnection ! PeerConnection.DoSync(replacePrevious = false)
         }
         // NB: we keep the temporary channel id because the switch is not always acknowledged at this point (see https://github.com/lightningnetwork/lightning-rfc/pull/151)
         // we won't clean it up, but we won't remember the temporary id on channel termination
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -196,6 +196,7 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
         log.info(s"channel closed: channelId=${channelIds.mkString("/")}")
         if (d.channels.values.toSet - actor == Set.empty) {
           log.info(s"that was the last open channel, closing the connection")
+          context.system.eventStream.publish(PeerLastChannelClosed(self, remoteNodeId))
           d.peerConnection ! PeerConnection.Kill(KillReason.NoRemainingChannel)
         }
         stay using d.copy(channels = d.channels -- channelIds)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -25,6 +25,7 @@ import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{ByteVector32, DeterministicWallet, Satoshi, SatoshiLong, Script}
 import fr.acinq.eclair.Features.Wumbo
 import fr.acinq.eclair.Logs.LogCategory
+import fr.acinq.eclair._
 import fr.acinq.eclair.blockchain.EclairWallet
 import fr.acinq.eclair.blockchain.fee.FeeratePerKw
 import fr.acinq.eclair.channel._
@@ -32,7 +33,6 @@ import fr.acinq.eclair.io.Monitoring.Metrics
 import fr.acinq.eclair.io.PeerConnection.KillReason
 import fr.acinq.eclair.remote.EclairInternalsSerializer.RemoteTypes
 import fr.acinq.eclair.wire._
-import fr.acinq.eclair.{wire, _}
 import scodec.bits.ByteVector
 
 import java.net.InetSocketAddress
@@ -164,6 +164,10 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
 
       case Event(ChannelIdAssigned(channel, _, temporaryChannelId, channelId), d: ConnectedData) if d.channels.contains(TemporaryChannelId(temporaryChannelId)) =>
         log.info(s"channel id switch: previousId=$temporaryChannelId nextId=$channelId")
+        // we have our first channel with that peer: let's sync our routing table
+        if (!d.channels.keys.exists(_.isInstanceOf[FinalChannelId])) {
+          d.peerConnection ! PeerConnection.DoSync
+        }
         // NB: we keep the temporary channel id because the switch is not always acknowledged at this point (see https://github.com/lightningnetwork/lightning-rfc/pull/151)
         // we won't clean it up, but we won't remember the temporary id on channel termination
         stay using d.copy(channels = d.channels + (FinalChannelId(channelId) -> channel))
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -196,7 +196,7 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
         log.info(s"channel closed: channelId=${channelIds.mkString("/")}")
         if (d.channels.values.toSet - actor == Set.empty) {
           log.info(s"that was the last open channel, closing the connection")
-          context.system.eventStream.publish(PeerLastChannelClosed(self, remoteNodeId))
+          context.system.eventStream.publish(LastChannelClosed(self, remoteNodeId))
           d.peerConnection ! PeerConnection.Kill(KillReason.NoRemainingChannel)
         }
         stay using d.copy(channels = d.channels -- channelIds)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -151,7 +151,7 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
             val flags_opt = if (canUseChannelRangeQueriesEx) Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL)) else None
             if (d.doSync) {
               log.info(s"sending sync channel range query with flags_opt=$flags_opt")
-              router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, flags_opt = flags_opt)
+              router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious = true, flags_opt = flags_opt)
             } else {
               log.info("not syncing with this peer")
             }
@@ -358,6 +358,15 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
         }
         stay using d.copy(behavior = behavior1)
 
+      case Event(DoSync, d: ConnectedData) =>
+        val flags_opt = if (d.localInit.features.hasFeature(Features.ChannelRangeQueriesExtended) && d.remoteInit.features.hasFeature(Features.ChannelRangeQueriesExtended)) {
+          Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL))
+        } else {
+          None
+        }
+        router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious = false, flags_opt)
+        stay
+
       case Event(ResumeAnnouncements, d: ConnectedData) =>
         log.info(s"resuming processing of network announcements for peer")
         stay using d.copy(behavior = d.behavior.copy(fundingTxAlreadySpentCount = 0, ignoreNetworkAnnouncement = false))
@@ -472,6 +481,7 @@ object PeerConnection {
   case object InitTimeout
   case object SendPing
   case object ResumeAnnouncements
+  case object DoSync
   // @formatter:on
 
   val IGNORE_NETWORK_ANNOUNCEMENTS_PERIOD: FiniteDuration = 5 minutes
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -345,7 +345,7 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
         val canUseChannelRangeQueriesEx = Features.canUseFeature(d.localInit.features, d.remoteInit.features, Features.ChannelRangeQueriesExtended)
         if (canUseChannelRangeQueries || canUseChannelRangeQueriesEx) {
           val flags_opt = if (canUseChannelRangeQueriesEx) Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL)) else None
-          log.info(s"sending sync channel range query with flags_opt=$flags_opt")
+          log.info(s"sending sync channel range query with flags_opt=$flags_opt replacePrevious=$replacePrevious")
           router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious, flags_opt)
         } else if (d.remoteInit.features.hasFeature(Features.InitialRoutingSync) && replacePrevious) {
           // For "old" nodes that don't support channel queries, we send them the full routing table
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -137,22 +137,10 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
           d.peer ! ConnectionReady(self, d.remoteNodeId, d.pendingAuth.address, d.pendingAuth.outgoing, d.localInit, remoteInit)
           d.pendingAuth.origin_opt.foreach(_ ! ConnectionResult.Connected)
 
-          val canUseChannelRangeQueries = Features.canUseFeature(d.localInit.features, remoteInit.features, Features.ChannelRangeQueries)
-          val canUseChannelRangeQueriesEx = Features.canUseFeature(d.localInit.features, remoteInit.features, Features.ChannelRangeQueriesExtended)
-          if (canUseChannelRangeQueries || canUseChannelRangeQueriesEx) {
-            // if they support channel queries we don't send routing info yet, if they want it they will query us
-            // we will query them, using extended queries if supported
-            val flags_opt = if (canUseChannelRangeQueriesEx) Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL)) else None
-            if (d.doSync) {
-              log.info(s"sending sync channel range query with flags_opt=$flags_opt")
-              router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious = true, flags_opt = flags_opt)
-            } else {
-              log.info("not syncing with this peer")
-            }
-          } else if (remoteInit.features.hasFeature(Features.InitialRoutingSync)) {
-            // "old" nodes, do as before
-            log.info("peer requested a full routing table dump")
-            router ! GetRoutingState
+          if (d.doSync) {
+            self ! DoSync(replacePrevious = true)
+          } else {
+            log.info("not syncing with this peer")
           }
 
           // we will delay all rebroadcasts with this value in order to prevent herd effects (each peer has a different delay)
@@ -352,16 +340,17 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
         }
         stay using d.copy(behavior = behavior1)
 
-      case Event(DoSync, d: ConnectedData) =>
+      case Event(DoSync(replacePrevious), d: ConnectedData) =>
         val canUseChannelRangeQueries = Features.canUseFeature(d.localInit.features, d.remoteInit.features, Features.ChannelRangeQueries)
         val canUseChannelRangeQueriesEx = Features.canUseFeature(d.localInit.features, d.remoteInit.features, Features.ChannelRangeQueriesExtended)
         if (canUseChannelRangeQueries || canUseChannelRangeQueriesEx) {
-          val flags_opt = if (canUseChannelRangeQueriesEx) {
-            Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL))
-          } else {
-            None
-          }
-          router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious = false, flags_opt)
+          val flags_opt = if (canUseChannelRangeQueriesEx) Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL)) else None
+          log.info(s"sending sync channel range query with flags_opt=$flags_opt")
+          router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious, flags_opt)
+        } else if (d.remoteInit.features.hasFeature(Features.InitialRoutingSync) && replacePrevious) {
+          // For "old" nodes that don't support channel queries, we send them the full routing table
+          log.info("peer requested a full routing table dump")
+          router ! GetRoutingState
         }
         stay
 
@@ -479,7 +468,7 @@ object PeerConnection {
   case object InitTimeout
   case object SendPing
   case object ResumeAnnouncements
-  case object DoSync
+  case class DoSync(replacePrevious: Boolean)
   // @formatter:on
 
   val IGNORE_NETWORK_ANNOUNCEMENTS_PERIOD: FiniteDuration = 5 minutes
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -16,8 +16,6 @@
 
 package fr.acinq.eclair.io
 
-import java.net.InetSocketAddress
-
 import akka.actor.{ActorRef, FSM, OneForOneStrategy, PoisonPill, Props, SupervisorStrategy, Terminated}
 import akka.event.Logging.MDC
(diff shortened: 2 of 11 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -162,6 +162,8 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, watcher: ActorRe
 
       case Event(ChannelIdAssigned(channel, _, temporaryChannelId, channelId), d: ConnectedData) if d.channels.contains(TemporaryChannelId(temporaryChannelId)) =>
         log.info(s"channel id switch: previousId=$temporaryChannelId nextId=$channelId")
+        // we have a new channel with that peer: let's sync our routing table
+        d.peerConnection ! PeerConnection.DoSync
```
t-bast, line 166: I'm currently doing this at every channel creation. If we think it's too much, I can restrict that to only the first `FinalChannelId` (ie the first channel we open to that peer). Let me know what you prefer.

eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
```diff
@@ -128,6 +128,7 @@ object EclairInternalsSerializer {
     ("chainsHash" | bytes32) ::
       ("remoteNodeId" | publicKey) ::
       ("to" | actorRefCodec(system)) ::
+      ("replacePrevious" | bool(8)) ::
```
t-bast, line 131: Is that ok? Or do I need to do some magic for backwards-compatibility?

eclair-core/src/main/scala/fr/acinq/eclair/router/Sync.scala
```diff
@@ -502,18 +515,17 @@ object Sync {
       checksums = checksums)
   }
 
-  def addToSync(syncMap: Map[PublicKey, Syncing], remoteNodeId: PublicKey, pending: List[RoutingMessage]): (Map[PublicKey, Syncing], Option[RoutingMessage]) = {
+  def addToSync(syncMap: Map[PublicKey, Syncing], current: Syncing, remoteNodeId: PublicKey, pending: List[RoutingMessage]): (Map[PublicKey, Syncing], Option[RoutingMessage]) = {
```
pm47, line 518: Really not sure why we didn't type this further?

```suggestion
  def addToSync(syncMap: Map[PublicKey, Syncing], current: Syncing, remoteNodeId: PublicKey, pending: List[QueryShortChannelIds]): (Map[PublicKey, Syncing], Option[QueryShortChannelIds]) = {
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Sync.scala
```diff
@@ -502,18 +515,17 @@ object Sync {
...
         // they may send back several reply_channel_range messages for a single query_channel_range query, and we must not
         // send another query_short_channel_ids query if they're still processing one
-        syncMap.get(remoteNodeId) match {
-          case None =>
-            // we don't have a pending query with this peer, let's send it
-            (syncMap + (remoteNodeId -> Syncing(rest, pending.size)), Some(head))
-          case Some(sync) =>
-            // we already have a pending query with this peer, add missing ids to our "sync" state
-            (syncMap + (remoteNodeId -> Syncing(sync.pending ++ pending, sync.total + pending.size)), None)
+        if (current.started) {
```
pm47, line 523: Why can't we just check the presence in the `syncMap` like before?

pm47 (commented):
Just a first pass.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -151,7 +151,7 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
             val flags_opt = if (canUseChannelRangeQueriesEx) Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL)) else None
             if (d.doSync) {
               log.info(s"sending sync channel range query with flags_opt=$flags_opt")
-              router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, flags_opt = flags_opt)
+              router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious = true, flags_opt = flags_opt)
```
pm47, line 154: I wonder if we should factor this inside one single handler in state `CONNECTED` for a `MaybeSync` event?

eclair-core/src/main/scala/fr/acinq/eclair/router/Sync.scala
```diff
@@ -47,23 +46,30 @@ object Sync {
...
-    s.to ! filter
-
-    // clean our sync state for this peer: we receive a SendChannelQuery just when we connect/reconnect to a peer and
-    // will start a new complete sync process
-    d.copy(sync = d.sync - s.remoteNodeId)
+    // we currently send query_channel_range when:
+    //  * we just (re)connected to a peer with whom we have channels
+    //  * we validate a new channel with that peer
+    // we must ensure we don't send a new query_channel_range while another query is still in progress
+    if (s.replacePrevious || !d.sync.contains(s.remoteNodeId)) {
```
pm47, line 53: It seems we replace the previous sync session when the `peerConnection` dies. Another strategy could be to listen to `Peer.ConnectionDown` events?

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -34,7 +34,7 @@ class Switchboard(nodeParams: NodeParams, watcher: ActorRef, relayer: ActorRef,
   import Switchboard._
 
   // we load channels from database
-  {
+  val peersWithChannels: Set[PublicKey] = {
```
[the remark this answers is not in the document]
t-bast: True, but I don't see any other way (because we do need this information available until we reconnect all peers), and is it really an issue? With 1.000 distinct peers this would only be around 50KB of data.

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -75,7 +75,8 @@ class Switchboard(nodeParams: NodeParams, watcher: ActorRef, relayer: ActorRef,
       // if this is an incoming connection, we might not yet have created the peer
       val peer = createOrGetPeer(authenticated.remoteNodeId, offlineChannels = Set.empty)
       val features = nodeParams.featuresFor(authenticated.remoteNodeId)
-      val doSync = nodeParams.syncWhitelist.isEmpty || nodeParams.syncWhitelist.contains(authenticated.remoteNodeId)
+      // if the peer is whitelisted, we sync with them, otherwise we only sync with peers with whom we have at least one channel
+      val doSync = nodeParams.syncWhitelist.contains(authenticated.remoteNodeId) || (nodeParams.syncWhitelist.isEmpty && peersWithChannels.contains(authenticated.remoteNodeId))
```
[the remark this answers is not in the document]
t-bast: No, that's why we need what's done in the second commit, to trigger sync when a new channel is created.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
t-bast: I don't understand what you mean and what that would bring :sweat_smile:

eclair-core/src/main/scala/fr/acinq/eclair/router/Sync.scala
t-bast: Can you explain why that would be better?

eclair-core/src/main/scala/fr/acinq/eclair/router/Sync.scala
t-bast: Because there will *always* be an entry in the `syncMap` (check where that is called, in `handleReplyChannelRange`).

t-bast: We could get it from the map here and call `.get`, but that would be a bit ugly, forcing the caller to pass it ensures we don't have an impossible case (the case where it would be `None`).

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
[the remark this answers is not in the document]
t-bast: Actually I think I can avoid needing this map, let me dig into it a bit.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: I think we should only do it for the first channel, otherwise we'd be mostly wasting bandwidth, I'll change that.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -358,6 +358,15 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
...
 
+      case Event(DoSync, d: ConnectedData) =>
+        val flags_opt = if (d.localInit.features.hasFeature(Features.ChannelRangeQueriesExtended) && d.remoteInit.features.hasFeature(Features.ChannelRangeQueriesExtended)) {
+          Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL))
+        } else {
+          None
+        }
+        router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious = false, flags_opt)
+        stay
+
```
sstone, line 369: What if our peer does not support channel queries at all (i.e they don't even set `Features.ChannelRangeQueries`) ?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -566,7 +567,13 @@ object Router {
 
   case class ShortChannelIdAndFlag(shortChannelId: ShortChannelId, flag: Long)
 
-  case class Syncing(pending: List[RoutingMessage], total: Int)
+  /**
+   * @param remainingQueries remaining queries to send, the next one will be popped after we receive a [[ReplyShortChannelIdsEnd]]
+   * @param totalQueries     total number of *queries* (not channels) that will be sent during this syncing session
+   */
+  case class Syncing(remainingQueries: List[RoutingMessage], totalQueries: Int) {
+    def started: Boolean = totalQueries > 0
```
sstone, line 575: nit: I read the comments about `val` vs `def` here but even when we call `.copy()` on a case class its constructors will be called and a `val` would be safe

t-bast: Interesting, let's keep it like this though where it's obvious that it will work?

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
t-bast: Good question, I'll dig into it.

t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1587/commits/e5ba331c3cdc0984c999126ecf4eacd5ea53a701

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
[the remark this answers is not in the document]
sstone: It seems that nodes we don't have channels with anymore will be kept in the  `peersWithChannels` map until the next restart and we will still sync with them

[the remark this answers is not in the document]
t-bast: I think it's not an issue, we won't because once we don't have channels anymore we disconnect and forget the peer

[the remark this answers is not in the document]
sstone: Yes but if that peer connects again we will sync with them even if they're not white-listed and don't have a channel with us ?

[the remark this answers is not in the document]
t-bast: Good point, I'll see if we can easily fix that scenario.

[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1587/commits/2b5c07959a563076512994cf7c614d3a45456110, let me know what you think.

sstone (dismissed):

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
pm47: This is what I have in mind, we factor the sync handling in one place.

I think we should consider dropping the support for `Features.InitialRoutingSync` alltogether, which would make this refactoring even more logical.

```diff
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
index deedcab67..be6e88075 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -137,22 +137,10 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
           d.peer ! ConnectionReady(self, d.remoteNodeId, d.pendingAuth.address, d.pendingAuth.outgoing, d.localInit, remoteInit)
           d.pendingAuth.origin_opt.foreach(_ ! ConnectionResult.Connected)

-          val canUseChannelRangeQueries = Features.canUseFeature(d.localInit.features, remoteInit.features, Features.ChannelRangeQueries)
-          val canUseChannelRangeQueriesEx = Features.canUseFeature(d.localInit.features, remoteInit.features, Features.ChannelRangeQueriesExtended)
-          if (canUseChannelRangeQueries || canUseChannelRangeQueriesEx) {
-            // if they support channel queries we don't send routing info yet, if they want it they will query us
-            // we will query them, using extended queries if supported
-            val flags_opt = if (canUseChannelRangeQueriesEx) Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL)) else None
-            if (d.doSync) {
-              log.info(s"sending sync channel range query with flags_opt=$flags_opt")
-              router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious = true, flags_opt = flags_opt)
-            } else {
-              log.info("not syncing with this peer")
-            }
-          } else if (remoteInit.features.hasFeature(Features.InitialRoutingSync)) {
-            // "old" nodes, do as before
-            log.info("peer requested a full routing table dump")
-            router ! GetRoutingState
+          if (d.doSync) {
+            self ! DoSync
+          } else {
+            log.info("not syncing with this peer")
           }

           // we will delay all rebroadcasts with this value in order to prevent herd effects (each peer has a different delay)
@@ -356,12 +344,19 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
         val canUseChannelRangeQueries = Features.canUseFeature(d.localInit.features, d.remoteInit.features, Features.ChannelRangeQueries)
         val canUseChannelRangeQueriesEx = Features.canUseFeature(d.localInit.features, d.remoteInit.features, Features.ChannelRangeQueriesExtended)
         if (canUseChannelRangeQueries || canUseChannelRangeQueriesEx) {
+          // This is us syncing from them
           val flags_opt = if (canUseChannelRangeQueriesEx) {
             Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL))
           } else {
             None
           }
           router ! SendChannelQuery(d.chainHash, d.remoteNodeId, self, replacePrevious = false, flags_opt)
+        } else if (d.remoteInit.features.hasFeature(Features.InitialRoutingSync)) {
+          // This is them syncing from us
+          // For "old" nodes that don't support channel queries, we send them the full routing table, otherwise we will
+          // just wait for their queries
+          log.info("peer requested a full routing table dump")
+          router ! GetRoutingState
         }
         stay

```

eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
pm47: We just need to make sure that we upgrade the front and back at the same time.

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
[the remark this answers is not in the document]
pm47: A more immutable/functional version:

```diff
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala b/eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
index 2d46e4a84..b9d048b56 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
@@ -25,8 +25,6 @@ import fr.acinq.eclair.channel._
 import fr.acinq.eclair.remote.EclairInternalsSerializer.RemoteTypes
 import fr.acinq.eclair.router.Router.RouterConf

-import scala.collection.mutable
-
 /**
  * Ties network connections to peers.
  * Created by PM on 14/02/2017.
@@ -39,7 +37,7 @@ class Switchboard(nodeParams: NodeParams, watcher: ActorRef, relayer: ActorRef,
   context.system.eventStream.subscribe(self, classOf[PeerLastChannelClosed])

   // we load channels from database
-  val peersWithChannels: mutable.Set[PublicKey] = {
+  private def initialPeersWithChannels: Set[PublicKey] = {
     // Check if channels that are still in CLOSING state have actually been closed. This can happen when the app is stopped
     // just after a channel state has transitioned to CLOSED and before it has effectively been removed.
     // Closed channels will be removed, other channels will be restored.
@@ -51,10 +49,12 @@ class Switchboard(nodeParams: NodeParams, watcher: ActorRef, relayer: ActorRef,

     val peerChannels = channels.groupBy(_.commitments.remoteParams.nodeId)
     peerChannels.foreach { case (remoteNodeId, states) => createOrGetPeer(remoteNodeId, offlineChannels = states.toSet) }
-    mutable.Set.empty.addAll(peerChannels.keys)
+    peerChannels.keySet
   }

-  def receive: Receive = {
+  def receive: Receive = normal(initialPeersWithChannels)
+
+  def normal(peersWithChannels: Set[PublicKey]): Receive = {

     case Peer.Connect(publicKey, _) if publicKey == nodeParams.nodeId =>
       sender ! Status.Failure(new RuntimeException("cannot open connection with oneself"))
@@ -84,9 +84,9 @@ class Switchboard(nodeParams: NodeParams, watcher: ActorRef, relayer: ActorRef,
       val doSync = nodeParams.syncWhitelist.contains(authenticated.remoteNodeId) || (nodeParams.syncWhitelist.isEmpty && peersWithChannels.contains(authenticated.remoteNodeId))
       authenticated.peerConnection ! PeerConnection.InitializeConnection(peer, nodeParams.chainHash, features, doSync)

-    case ChannelIdAssigned(_, remoteNodeId, _, _) => peersWithChannels.add(remoteNodeId)
+    case ChannelIdAssigned(_, remoteNodeId, _, _) => context become normal(peersWithChannels + remoteNodeId)

-    case PeerLastChannelClosed(_, remoteNodeId) => peersWithChannels.remove(remoteNodeId)
+    case PeerLastChannelClosed(_, remoteNodeId) => context become normal(peersWithChannels - remoteNodeId)

     case Symbol("peers") => sender ! context.children

```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -519,7 +520,7 @@ object Router {
   // @formatter:on
 
   // @formatter:off
-  case class SendChannelQuery(chainHash: ByteVector32, remoteNodeId: PublicKey, to: ActorRef, flags_opt: Option[QueryChannelRangeTlv]) extends RemoteTypes
+  case class SendChannelQuery(chainHash: ByteVector32, remoteNodeId: PublicKey, to: ActorRef, replacePrevious: Boolean, flags_opt: Option[QueryChannelRangeTlv]) extends RemoteTypes
```
pm47, line 523: We don't seem to be using `replacePrevious=true` anywhere except in tests?

eclair-core/src/main/scala/fr/acinq/eclair/router/Sync.scala
```diff
@@ -86,59 +92,67 @@ object Sync {
...
 
-    Metrics.ReplyChannelRange.Blocks.withTag(Tags.Direction, Tags.Directions.Incoming).record(r.numberOfBlocks)
-    Metrics.ReplyChannelRange.ShortChannelIds.withTag(Tags.Direction, Tags.Directions.Incoming).record(r.shortChannelIds.array.size)
+    d.sync.get(origin.nodeId) match {
+      case None =>
+        log.info("received unsolicited reply_channel_range with {} channels", r.shortChannelIds.array.size)
+        d // we didn't request a sync from this node, ignore
+      case Some(currentSync) if currentSync.remainingQueries.isEmpty && r.shortChannelIds.array.isEmpty =>
+        log.info("received empty reply_channel_range, sync is complete")
+        d.copy(sync = d.sync - origin.nodeId)
```
pm47, line 101: This isn't how the spec indicates that we should detect the last `reply_channel_range` message:

> the final reply_channel_range message:
>    MUST have first_blocknum plus number_of_blocks equal or greater than the query_channel_range first_blocknum plus number_of_blocks.

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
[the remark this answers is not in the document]
pm47: This patch solves this: https://github.com/ACINQ/eclair/pull/1587#discussion_r569494216.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: We do, in `PeerConnection` when receiving their `init` message (that's the only case where we replace the previous sync)

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerEvents.scala
```diff
@@ -31,4 +31,6 @@ case class PeerConnected(peer: ActorRef, nodeId: PublicKey, connectionInfo: Conn
 
 case class PeerDisconnected(peer: ActorRef, nodeId: PublicKey) extends PeerEvent
 
+case class PeerLastChannelClosed(peer: ActorRef, nodeId: PublicKey) extends PeerEvent
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1587/commits/a632b719b03aa69a32bfe08efda49ed3dc43ab20

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1587/commits/c87bbe4c2f284ed6cecbb5bdcf289bde5d4a9e7a

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1587/commits/ea4f2a3de7e7a35e2b7d924699cfe32027323c53

eclair-core/src/main/scala/fr/acinq/eclair/router/Sync.scala
t-bast: Right, this is probably a leftover made to handle nodes that didn't correctly signal termination?
I'll look into it.
Note that we'll be bringing back the `complete` field which will greatly simplify our lives.

t-bast: This case doesn't handle the normal termination of `reply_channel_range`, it deals with a special case where the *first* `reply_channel_range` is empty (meaning the remote node doesn't want to send us anything basically). Otherwise we'll have something in `remainingQueries` and we'll wait for these to complete before we consider the sync done. I had to add this because we have a test case for it. 

An interesting note is that we don't detect the end of a `reply_channel_range` at all: we rely on `reply_short_channel_ids_end` and the fact that our `remainingQueries` is empty. _That's different from the spec and should be fixed_. In practice it will very likely work (we've always had this behavior and we've been fine) and it's orthogonal to this PR since it was our previous behavior. I think that can be fixed in a later PR (and we can wait to rely on the `complete` field for a cleaner signal that the sync is complete once we correctly set it again).

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -358,6 +340,20 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
         }
         stay using d.copy(behavior = behavior1)
 
+      case Event(DoSync(replacePrevious), d: ConnectedData) =>
+        val canUseChannelRangeQueries = Features.canUseFeature(d.localInit.features, d.remoteInit.features, Features.ChannelRangeQueries)
+        val canUseChannelRangeQueriesEx = Features.canUseFeature(d.localInit.features, d.remoteInit.features, Features.ChannelRangeQueriesExtended)
+        if (canUseChannelRangeQueries || canUseChannelRangeQueriesEx) {
+          val flags_opt = if (canUseChannelRangeQueriesEx) Some(QueryChannelRangeTlv.QueryFlags(QueryChannelRangeTlv.QueryFlags.WANT_ALL)) else None
+          log.info(s"sending sync channel range query with flags_opt=$flags_opt")
```
pm47, line 348:
```suggestion
          log.info(s"sending sync channel range query with flags_opt=$flags_opt replacePrevious=$replacePrevious")
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Sync.scala
pm47: Maybe you should include your reply as a `TODO` note here?

pm47 (dismissed):

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1587/commits/8f292b9f7dc2bb40e65cc576f05c04a6722ce5ab

eclair-core/src/main/scala/fr/acinq/eclair/router/Sync.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1587/commits/8f292b9f7dc2bb40e65cc576f05c04a6722ce5ab and added to my todo list.

pm47 (approved):

PR #1588, opened 2020-11-05
Fix NodeRelayer flaky test

t-bast:
This test was randomly failing because of a race condition: we sometimes sent the payment failure before the payment FSM asked for routes, so it was ignored and the test could not progress.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
@@ -299,7 +299,7 @@ class NodeRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("appl
     payFSM ! Status.Failure(BalanceTooLow)
 
     incoming.foreach { p =>
-      val fwd = register.expectMessageType[Register.Forward[CMD_FAIL_HTLC]] // TODO: timing out?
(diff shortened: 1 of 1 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -295,10 +295,11 @@ class NodeRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("appl
     incoming.foreach(p => nodeRelayer ! NodeRelay.Relay(p))
 
     val payFSM = mockPayFSM.expectMessageType[akka.actor.ActorRef]
+    router.expectMessageType[RouteRequest]
     payFSM ! Status.Failure(BalanceTooLow)
 
     incoming.foreach { p =>
-      val fwd = register.expectMessageType[Register.Forward[CMD_FAIL_HTLC]]
+      val fwd = register.expectMessageType[Register.Forward[CMD_FAIL_HTLC]] // TODO: timing out?
```
pm47, line 302: What is this `TODO` for?

t-bast: Woops, that was my comment on where the test was failing :smile:

pm47 (approved):

PR #1590, opened 2020-11-06
Add router API for invoice routing hints

t-bast:
The router can now return local channels, with all necessary information to build routing hints from these channels.

It's not eclair-core's responsibility to choose what channels to use in routing hints: this is really a wallet strategy, so each wallet should have its own heuristics to choose what channels to include (if any).

I think @btcontract was interested in these.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -162,25 +162,25 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
       log.info("reinstating shortChannelId={} from nodeId={}", shortChannelId, nodeId)
       stay using d.copy(excludedChannels = d.excludedChannels - desc)
 
-    case Event(Symbol("nodes"), d) =>
+    case Event(GetNodes, d) =>
       sender ! d.nodes.values
       stay
 
-    case Event(Symbol("localChannels"), d) =>
+    case Event(GetLocalChannels, d) =>
       val scids = d.graph.getIncomingEdgesOf(nodeParams.nodeId).map(_.desc.shortChannelId)
       val localChannels = scids.flatMap(scid => d.channels.get(scid).map(c => LocalChannel(Right(c))).orElse(d.privateChannels.get(scid).map(c => LocalChannel(Left(c)))))
       sender ! localChannels
       stay
 
-    case Event(Symbol("channels"), d) =>
+    case Event(GetChannels, d) =>
       sender ! d.channels.values.map(_.ann)
       stay
 
-    case Event(Symbol("channelsMap"), d) =>
+    case Event(GetChannelsMap, d) =>
       sender ! d.channels
       stay
 
-    case Event(Symbol("updates"), d) =>
+    case Event(GetChannelUpdates, d) =>
       val updates: Iterable[ChannelUpdate] = d.channels.values.flatMap(d => d.update_1_opt ++ d.update_2_opt) ++ d.privateChannels.values.flatMap(d => d.update_1_opt ++ d.update_2_opt)
       sender ! updates
       stay
@@ -493,6 +493,11 @@ object Router {
   case object GetRoutingStateStreaming
   case object RoutingStateStreamingUpToDate
   case object GetRouterData
+  case object GetNodes
+  case object GetLocalChannels
+  case object GetChannels
+  case object GetChannelsMap
+  case object GetChannelUpdates
   // @formatter:on
 
   // @formatter:off
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -168,7 +168,7 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
 
     case Event(GetLocalChannels, d) =>
       val scids = d.graph.getIncomingEdgesOf(nodeParams.nodeId).map(_.desc.shortChannelId)
-      val localChannels = scids.flatMap(scid => d.channels.get(scid).map(c => LocalChannel(Right(c))).orElse(d.privateChannels.get(scid).map(c => LocalChannel(Left(c)))))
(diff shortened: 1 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -466,6 +493,11 @@ object Router {
   case object GetRoutingStateStreaming
   case object RoutingStateStreamingUpToDate
   case object GetRouterData
+  case object GetNodes
+  case object GetLocalChannels
+  case object GetChannels
+  case object GetChannelsMap
+  case object GetChannelUpdates
```
pm47, line 500: You could have used the `replyTo` pattern for those new messages, but we may as well handle them all at once when the day comes.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -162,19 +162,25 @@ class Router(val nodeParams: NodeParams, watcher: ActorRef, initialized: Option[
...
 
-    case Event(Symbol("nodes"), d) =>
+    case Event(GetNodes, d) =>
       sender ! d.nodes.values
       stay
 
-    case Event(Symbol("channels"), d) =>
+    case Event(GetLocalChannels, d) =>
+      val scids = d.graph.getIncomingEdgesOf(nodeParams.nodeId).map(_.desc.shortChannelId)
+      val localChannels = scids.flatMap(scid => d.channels.get(scid).map(c => LocalChannel(Right(c))).orElse(d.privateChannels.get(scid).map(c => LocalChannel(Left(c)))))
```
pm47, line 171: Slightly shorter version. I tried linebreaks, but it wasn't much more readable. We could save one iteration by removing the `.map(_.desc.shortChannelId)`.

```suggestion
      val localChannels = scids.flatMap(scid => d.channels.get(scid).map(Right(_)).orElse(d.privateChannels.get(scid).map(Left(_)).map(LocalChannel)))
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -322,6 +328,27 @@ object Router {
       case Right(rcu) => updateChannelUpdateSameSideAs(rcu.channelUpdate)
     }
   }
+  case class LocalChannel(channel: Either[PrivateChannel, PublicChannel]) {
```
pm47, line 331: Should we have `PrivateChannel` and `PublicChannel` extend the same `trait`?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -322,6 +328,27 @@ object Router {
...
+    def getRemoteUpdate(localNodeId: PublicKey): Option[ChannelUpdate] = channel.fold(
+      c => if (localNodeId == c.nodeId1) c.update_2_opt else c.update_1_opt,
+      c => if (localNodeId == c.ann.nodeId1) c.update_2_opt else c.update_1_opt
+    )
+
+    /** Create an invoice routing hint from that channel. Note that if the channel is private, the invoice will leak its existence. */
+    def toExtraHop(localNodeId: PublicKey): Option[ExtraHop] = getRemoteUpdate(localNodeId).map(remoteUpdate =>
+      ExtraHop(getRemoteNodeId(localNodeId), remoteUpdate.shortChannelId, remoteUpdate.feeBaseMsat, remoteUpdate.feeProportionalMillionths, remoteUpdate.cltvExpiryDelta)
+    )
+  }
```
pm47, line 351: It seems to me that we could make `remoteNodeId` part of the constructor. It is a bit strange that we need to provide more info for a class named `LocalChannel` to tell which side is the local one.

pm47 (commented):
This doesn't "add a router API for invoice routing hints"?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Agreed, I'd rather change the whole router at once (or at least everything we can update in the router at once).

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Right, probably, I don't know why that wasn't done initially.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Agreed, I'll revisit that.

t-bast:
> This doesn't "add a router API for invoice routing hints"?

Yes it does, but only to the router. It doesn't add it to the HTTP API as I think it's unnecessary, it's meant to be used by plugins for nodes that offer custodial wallet capabilities (for public nodes you usually don't need any routing hint, you have public channels).

btcontract:
Yes, my rationale is that some nodes may start pruning smaller public channels at some point to keep graph manageable, so adding hints for smaller public channels can be useful.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1590/commits/273dd81593614f3323350c7ce0e3c52933841ca7, let me know what you think.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1590/commits/273dd81593614f3323350c7ce0e3c52933841ca7
The only downside is the methods that return a new instance of `PublicChannel/PrivateChannel` now return that shared abstraction, not sure whether it's an issue or not.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: I think it's better.

pm47 (approved):

PR #1594, opened 2020-11-09
Allow plugins to force reconnect

btcontract:
Discussed at https://gitter.im/ACINQ/developers?at=5fa1289c2a3544071500e1b3, this will allow message-enabled plugins to keep a connection open to a given node even if there are no channels left.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
@@ -17,6 +17,7 @@
 package fr.acinq.eclair
 
 import fr.acinq.bitcoin.ByteVector32
+import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.channel.Origin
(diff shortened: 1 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
```diff
@@ -34,6 +35,9 @@ trait CustomFeaturePlugin extends PluginParams {
   /** Feature bit that the plugin wants to advertise through Init message. */
   def feature: Feature
 
+  /** Once disconnect happens, should Eclair attempt periodic reconnects to a given remote node even if there is no normal channels left */
+  def forceReconnect(nodeId: PublicKey): Boolean
```
t-bast, line 39: That doesn't make sense in the `CustomFeaturePlugin`, it's not feature-related.
You should create a new trait for that.

btcontract: OK, my rationale was that `CustomFeaturePlugin` defines network messages it's interested it so it's inherently something network-related, might as well care about connection.

t-bast (approved):

PR #1595, opened 2020-11-09
High fee tolerance for small channels

pm47:
ℹ️ this is a PR to `android-phoenix`

For small channels we blindly accept feerate from funder.
This is a temporary fix to allow making payment in high fees
environment and will be removed when "anchor outputs" is deployed


pm47:
I considered defining a `def capacity: Satoshi = commitments.commitInput.txOut.amount` but went with the smallest change.

t-bast (approved):

PR #1596, opened 2020-11-11
Remove unused receiveFail/receiveFailMalformed return values

btcontract:
`receiveFail` and `receiveFailMalformed` return values which are not used anywhere, this removes those values.


t-bast:
It is used in feature branches, so we should keep these return values.

PR #1599, opened 2020-11-12
Improve MPP-send for direct channels

t-bast:
When the recipient is a direct peer, we can use the accurate knowledge of our local channels instead of hard-coded MPP parameters to choose the number of routes and the minimum route amount.

This change makes it possible to easily send all local funds in one payment when we're directly connected to the recipient.

@dpad85 I'll cherry-pick this commit on the `android-phoenix` branch once this is merged. It made me think that we probably want to change the default (15000 sats, 6 parts) for eclair-mobile MPP: no-one has complained yet, but maybe (1 000 sat, 10) would be better suited for a wallet.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -327,11 +327,10 @@ object RouteCalculation {
     // When the recipient is a direct peer, we have complete visibility on our local channels so we can use more accurate MPP parameters.
     val routeParams1 = {
       case class DirectChannel(balance: MilliSatoshi, isEmpty: Boolean)
-      val directChannels = g.getEdgesBetween(localNodeId, targetNodeId).map(e => {
-        val balance = e.balance_opt.getOrElse(e.capacity.toMilliSatoshi)
-        val isEmpty = balance < e.update.htlcMinimumMsat
-        DirectChannel(balance, isEmpty)
-      })
+      val directChannels = g.getEdgesBetween(localNodeId, targetNodeId).collect {
+        // We should always have balance information available for local channels.
+        case GraphEdge(_, update, _, Some(balance)) => DirectChannel(balance, balance < update.htlcMinimumMsat)
+      }
       // If we have direct channels to the target, we can use them all.
       val numRoutes = routeParams.mpp.maxParts.max(directChannels.length)
       // If we have direct channels to the target, we can use them all, even if they have only a small balance left.
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -324,15 +324,24 @@ object RouteCalculation {
                                          routeParams: RouteParams,
                                          currentBlockHeight: Long): Either[RouterException, Seq[Route]] = {
     // We use Yen's k-shortest paths to find many paths for chunks of the total amount.
-    val numRoutes = {
-      val directChannelsCount = g.getEdgesBetween(localNodeId, targetNodeId).length
-      routeParams.mpp.maxParts.max(directChannelsCount) // if we have direct channels to the target, we can use them all
+    // When the recipient is a direct peer, we have complete visibility on our local channels so we can use more accurate MPP parameters.
+    val routeParams1 = {
+      case class DirectChannel(balance: MilliSatoshi, isEmpty: Boolean)
+      val directChannels = g.getEdgesBetween(localNodeId, targetNodeId).map(e => {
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -324,15 +324,24 @@ object RouteCalculation {
...
                                          currentBlockHeight: Long): Either[RouterException, Seq[Route]] = {
     // We use Yen's k-shortest paths to find many paths for chunks of the total amount.
-    val numRoutes = {
-      val directChannelsCount = g.getEdgesBetween(localNodeId, targetNodeId).length
-      routeParams.mpp.maxParts.max(directChannelsCount) // if we have direct channels to the target, we can use them all
+    // When the recipient is a direct peer, we have complete visibility on our local channels so we can use more accurate MPP parameters.
+    val routeParams1 = {
+      case class DirectChannel(balance: MilliSatoshi, isEmpty: Boolean)
+      val directChannels = g.getEdgesBetween(localNodeId, targetNodeId).map(e => {
+        val balance = e.balance_opt.getOrElse(e.capacity.toMilliSatoshi)
```
pm47, line 331: Why `getOrElse`? Shouldn't we always know the balance for local channels.

If so, we may as well set `0.msat` so that unexpected behavior is more evident?

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -324,15 +324,24 @@ object RouteCalculation {
...
+        val isEmpty = balance < e.update.htlcMinimumMsat
+        DirectChannel(balance, isEmpty)
+      })
+      // If we have direct channels to the target, we can use them all.
+      val numRoutes = routeParams.mpp.maxParts.max(directChannels.length)
+      // If we have direct channels to the target, we can use them all, even if they have only a small balance left.
+      val routeAmount = (amount +: routeParams.mpp.minPartAmount +: directChannels.filter(!_.isEmpty).map(_.balance)).min
```
pm47, line 338: I don't get that line:
```scala
val routeAmount = (amount +: routeParams.mpp.minPartAmount +: directChannels.filter(!_.isEmpty).map(_.balance)).min
```

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
t-bast: We always know the balance for local channels but only once they've come up online and sent their `LocalChannelUpdate`, there's a potential race here...but you're right that if we don't have the balance available yet, the path-finding algorithm will very likely fail because it will try to use too much liquidity from the channel, so your proposal of disabling those channels by setting balance to 0 msat makes sense.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
t-bast: We want to use the lowest common denominator between:

- the total amount we want to send
- the configured minimum part amount
- the lowest balance of our our usable channels when the recipient is a direct peer

That's why we put these all in a `Seq` and take the min.

t-bast: The path-finding algorithm will find a bunch of non-overlapping routes that can carry that amount, then the `split` method will "fill" each of these routes with a bigger amount to send depending on the route's capacity.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
t-bast: Actually that was incorrect: for local channels, we should always have balance information available (otherwise the channel will not yet be in the graph).

Since it's an `Option` maybe what we should do instead is `collect` only those for which it's defined?

t-bast: Done in https://github.com/ACINQ/eclair/pull/1599/commits/337ba17b5a57b987738f30f866338839843892ed

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
pm47: I think the naming is what got me confused:

```suggestion
      val minPartAmount = (amount +: routeParams.mpp.minPartAmount +: directChannels.filter(!_.isEmpty).map(_.balance)).min
```

t-bast: Right, that makes sense, fixed in https://github.com/ACINQ/eclair/pull/1599/commits/3ee956d451af173399db4d92878aa3acdf960342

pm47 (approved):

PR #1601, opened 2020-11-13
Improve MPP-send for direct channels (#1599)

t-bast:
When the recipient is a direct peer, we can use the accurate knowledge
of our local channels instead of hard-coded MPP parameters to choose
the number of routes and the minimum route amount.

This change makes it possible to easily send all local funds in one
payment when we're directly connected to the recipient.

NB: this PR imports in `android` a fix that has been committed to `master`.


dpad85 (approved):
👍

PR #1602, opened 2020-11-13
Add `replyToCommand` method, use it in Channel

btcontract:
Main reason is this logic is exactly reused in HC, it seems nice to define it as precise method.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -99,9 +99,8 @@ object Channel {
   case class OutgoingMessage(msg: LightningMessage, peerConnection: ActorRef)
 
   def replyToCommand(sender: ActorRef, reply: CommandResponse[Command], cmd: Command): Unit = cmd match {
+    case cmd1: HasReplyToCommand => if (cmd1.replyTo == ActorRef.noSender) sender ! reply else cmd1.replyTo ! reply
     case cmd1: HasOptionalReplyToCommand => cmd1.replyTo_opt.foreach(_ ! reply)
-    case cmd1: HasReplyToCommand if cmd1.replyTo == ActorRef.noSender => sender ! reply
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -98,6 +98,10 @@ object Channel {
    */
   case class OutgoingMessage(msg: LightningMessage, peerConnection: ActorRef)
 
+  def replyToCommand(sender: ActorRef, reply: CommandResponse[Command], cmd: Command): Unit = cmd match {
+    case cmd1: HasReplyToCommand => if (cmd1.replyTo == ActorRef.noSender) sender ! reply else cmd1.replyTo ! reply
+    case cmd1: HasOptionalReplyToCommand => cmd1.replyTo_opt.foreach(_ ! reply)
+  }
```
t-bast, line 104: I'm not sure this makes sense in the `object`, we've put everything in the `class` up until now.
Let's see what @pm47 thinks and why he didn't add such a method initially.

pm47: > Let's see what @pm47 thinks and why he didn't add such a method initially.

My initial goal was to make all commands a `HasReplyTo`

> I'm not sure this makes sense in the object, we've put everything in the class up until now.

There is a major difference : the `!` method has an implicit argument that would be lost if the method is defined in the `object`. We may want to do it on purpose though, since with _akka typed_ there is no more `sender`.
```scala
def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit
```

btcontract: so... can we proceed with this then?

t-bast: I don't think so, this is a change in behavior since the implicit `sender` is lost when putting the function in the `object`.
I would not change this for now, and wait for our full migration to akka typed which will completely remove that code.

PR #1603, opened 2020-11-14
Optimize replaying on restart

btcontract:
This change is two-fold:

1. It puts FAIL/FULFILL replay logic into a separate method since it's mostly duplicated in `Channel` (and is reused in HC).
2. method is only called if there are pending HTLCs present.

As is it's a bit coarse-grained since method is called if incoming or outgoing HTLCs are present while checking for incoming only is sufficient. But it nicely reuses an existing `hasPendingOrProposedHtlcs` and it's still better than unconditional DB call.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -18,6 +18,7 @@ package fr.acinq.eclair.channel
 
 import akka.actor.{ActorRef, FSM, OneForOneStrategy, Props, Status, SupervisorStrategy}
 import akka.event.Logging.MDC
+import akka.event.LoggingAdapter
 import akka.pattern.pipe
 import fr.acinq.bitcoin.Crypto.{PrivateKey, PublicKey}
 import fr.acinq.bitcoin.{ByteVector32, OutPoint, Satoshi, Script, ScriptFlags, Transaction}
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1778,22 +1789,13 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-      PendingRelayDb.getPendingFailsAndFulfills(nodeParams.db.pendingRelay, nextStateData.asInstanceOf[HasCommitments].channelId) match {
-        case Nil =>
-          log.debug("nothing to replay")
-        case cmds =>
-          log.info("replaying {} unacked fulfills/fails", cmds.size)
-          cmds.foreach(self ! _) // they all have commit = false
-          self ! CMD_SIGN() // so we can sign all of them at once
+    case state -> nextState =>
+      (state, nextState, nextStateData) match {
+        case (_, CLOSING, data: HasCommitments) if data.commitments.hasPendingOrProposedHtlcs =>
```
t-bast, line 1794: On the contrary, we want a DB call here in all cases for safety.
It allows us to detect if there's a bug that creates incoherences between channel state and pendingRelayDb.
And the previous code was clearer by only focusing on the states we're transitioning to/from.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -98,6 +99,16 @@ object Channel {
    */
   case class OutgoingMessage(msg: LightningMessage, peerConnection: ActorRef)
 
+  def replayPendingFailsAndFulfills(sendTo: ActorRef, pendingRelay: PendingRelayDb, channelId: ByteVector32, sendSignCmd: Boolean)(log: LoggingAdapter): Unit = {
```
t-bast, line 102: I don't know if we're ready to put this kind of helper function in the object instead of the class.
It feels a premature, that's probably something we'll do when we move this actor to akka typed.

btcontract:
I'll just close it, then. Rewriting from scratch in HC is fine, I thought optimization has more value here (which turned to not be the case).

PR #1604, opened 2020-11-16
Remove `timedOutOutgoingHtlcs` from AbstractCommitments

btcontract:
It turns out that:

- `def timedOutOutgoingHtlcs(blockHeight: Long): Set[wire.UpdateAddHtlc]` defined in `AbstractCommitments` is currently only used in `Commitments` object and nowhere else.
- Logic in HC commitments object can be simplified if `timedOutOutgoingHtlcs` takes additional arguments besides `blockHeight`.

So it makes sense to just remove `timedOutOutgoingHtlcs` from `AbstractCommitments` and let HC commitments object to do things differently internally without implementing of (useless) `timedOutOutgoingHtlcs`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -53,8 +53,6 @@ trait AbstractCommitments {
 
   def getIncomingHtlcCrossSigned(htlcId: Long): Option[UpdateAddHtlc]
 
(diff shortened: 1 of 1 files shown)


t-bast:
Please spend more time testing those thoroughly.
Changing these abstractions creates issues for our merges to feature branches, so we need some stability on these abstractions.

btcontract:
It's safe to say this is it, the work on HC is done and I don't see a need for any more PRs (besides maybe https://gitter.im/ACINQ/developers?at=5fae6c8bc6fe0131d4f1ea88).

t-bast (approved):

PR #1608, opened 2020-11-19
Retrieve bitcoin's mempool-min-fee

t-bast:
* Retrieve the minimum fee required to enter our mempool
* Add a metric to track this value
* Remove fallback fee providers when Bitcoin Core is used
* Add a metric to track feerate retrieval errors

For fallback fee providers that don't have any mempool information, we default to the 1008 block target. It doesn't matter as these fee providers will not be used anyway.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -224,6 +224,7 @@ class Setup(datadir: File,
 
       defaultFeerates = {
         val confDefaultFeerates = FeeratesPerKB(
+          mempoolMinFee = FeeratePerKB(Satoshi(config.getLong("on-chain-fees.default-feerates.1008"))),
           block_1 = FeeratePerKB(Satoshi(config.getLong("on-chain-fees.default-feerates.1"))),
           blocks_2 = FeeratePerKB(Satoshi(config.getLong("on-chain-fees.default-feerates.2"))),
           blocks_6 = FeeratePerKB(Satoshi(config.getLong("on-chain-fees.default-feerates.6"))),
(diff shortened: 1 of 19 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -244,7 +245,7 @@ class Setup(datadir: File,
         case (Block.RegtestGenesisBlock.hash, _) =>
           new FallbackFeeProvider(new ConstantFeeProvider(defaultFeerates) :: Nil, minFeeratePerByte)
         case (_, Bitcoind(bitcoinClient)) =>
-          new FallbackFeeProvider(new SmoothFeeProvider(new BitcoinCoreFeeProvider(bitcoinClient, defaultFeerates), smoothFeerateWindow) :: new SmoothFeeProvider(new BitgoFeeProvider(nodeParams.chainHash, readTimeout), smoothFeerateWindow) :: new SmoothFeeProvider(new EarnDotComFeeProvider(readTimeout), smoothFeerateWindow) :: Nil, minFeeratePerByte) // order matters!
+          new FallbackFeeProvider(new SmoothFeeProvider(new BitcoinCoreFeeProvider(bitcoinClient, defaultFeerates), smoothFeerateWindow) :: Nil, minFeeratePerByte)
```
pm47, line 248: I think we can remove the FallbackFeeProvider in that case.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
```diff
@@ -102,8 +102,8 @@ object FeeratePerKw {
 }
 
 /** Fee rates in satoshi-per-kilo-bytes (1 kb = 1000 bytes). */
-case class FeeratesPerKB(block_1: FeeratePerKB, blocks_2: FeeratePerKB, blocks_6: FeeratePerKB, blocks_12: FeeratePerKB, blocks_36: FeeratePerKB, blocks_72: FeeratePerKB, blocks_144: FeeratePerKB, blocks_1008: FeeratePerKB) {
-  require(block_1.feerate > 0.sat && blocks_2.feerate > 0.sat && blocks_6.feerate > 0.sat && blocks_12.feerate > 0.sat && blocks_36.feerate > 0.sat && blocks_72.feerate > 0.sat && blocks_144.feerate > 0.sat && blocks_1008.feerate > 0.sat, "all feerates must be strictly greater than 0")
+case class FeeratesPerKB(mempoolMinFee: FeeratePerKB, block_1: FeeratePerKB, blocks_2: FeeratePerKB, blocks_6: FeeratePerKB, blocks_12: FeeratePerKB, blocks_36: FeeratePerKB, blocks_72: FeeratePerKB, blocks_144: FeeratePerKB, blocks_1008: FeeratePerKB) {
```
pm47, line 105: Maybe you can squeeze a comment somewhere explaining what `mempoolMinFee` is and why it's ok to set it to `blocks_1008` if the provider doesn't offer finer-grained data?

pm47 (commented):
Nits, otherwise LGTM.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
t-bast: In fact we can't, it's that `FallbackFeeProvider` that enforces the minimum feerate...
Once we've removed bitgo / earndotcom, we'll be able to clean this up a bit, but let's keep it for now?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1608/commits/ce14f3742f74fd134762673a40b99225b198cd2f

pm47 (approved):

PR #1609, opened 2020-11-19
Add metrics to track utxos

t-bast:
We track the following metrics:

* Total utxos amount
* Number of utxos
* Split by status: unconfirmed, unsafe, safe and confirmed

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
@@ -99,7 +99,7 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
           blockCount.set(count)
           context.system.eventStream.publish(CurrentBlockCount(count))
       }
-      checkUtxos()
+      checkUtxos().recoverWith { case ex => log.warning(s"could not check utxos: $ex") }
       // TODO: beware of the herd effect
       KamonExt.timeFuture(Metrics.NewBlockCheckConfirmedDuration.withoutTags()) {
         Future.sequence(watches.collect { case w: WatchConfirmed => checkConfirmed(w) })
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
@@ -32,7 +32,7 @@ import fr.acinq.eclair.blockchain.bitcoind.rpc.ExtendedBitcoinClient
 import fr.acinq.eclair.blockchain.watchdogs.BlockchainWatchdog
 import fr.acinq.eclair.channel.BITCOIN_PARENT_TX_CONFIRMED
 import fr.acinq.eclair.transactions.Scripts
-import org.json4s.JsonAST.JDecimal
+import org.json4s.JsonAST.{JArray, JBool, JDecimal, JInt}
 import scodec.bits.ByteVector
 
 import scala.collection.immutable.SortedMap
@@ -100,9 +100,28 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
           blockCount.set(count)
           context.system.eventStream.publish(CurrentBlockCount(count))
       }
-      client.rpcClient.invoke("getbalance").collect {
-        // We track our balance in mBTC because a rounding issue in BTC would be too impactful.
-        case JDecimal(balance) => Metrics.BitcoinBalance.withoutTags().update(balance.doubleValue * 1000)
+      // TODO: move to a separate actor that listens to CurrentBlockCount
+      client.rpcClient.invoke("listunspent").collect {
+        case JArray(values) =>
+          val utxos = values.map(utxo => {
+            val JInt(confirmations) = utxo \ "confirmations"
+            val JBool(safe) = utxo \ "safe"
+            val JDecimal(amount) = utxo \ "amount"
+            (amount.doubleValue * 1000, confirmations.toLong, safe)
+          })
+          val filteredByStatus = Seq(
+            (Monitoring.Tags.UtxoStatuses.Confirmed, utxos.filter(_._2 > 0)),
+            (Monitoring.Tags.UtxoStatuses.Unconfirmed, utxos.filter(_._2 == 0)),
+            (Monitoring.Tags.UtxoStatuses.Safe, utxos.filter(_._3)),
+            (Monitoring.Tags.UtxoStatuses.Unsafe, utxos.filter(!_._3)),
+          )
+          filteredByStatus.foreach {
+            case (status, filteredUtxos) =>
+              val amount = filteredUtxos.map(_._1).sum
+              log.info(s"we have ${filteredUtxos.length} $status utxos ($amount mBTC)")
+              Monitoring.Metrics.UtxoCount.withTag(Monitoring.Tags.UtxoStatus, status).update(filteredUtxos.length)
+              Monitoring.Metrics.BitcoinBalance.withTag(Monitoring.Tags.UtxoStatus, status).update(amount)
+          }
       }
       // TODO: beware of the herd effect
       KamonExt.timeFuture(Metrics.NewBlockCheckConfirmedDuration.withoutTags()) {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
@@ -18,7 +18,6 @@ package fr.acinq.eclair.blockchain.bitcoind
 
 import java.util.concurrent.Executors
 import java.util.concurrent.atomic.AtomicLong
-
 import akka.actor.typed.SupervisorStrategy
 import akka.actor.typed.scaladsl.Behaviors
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -100,9 +100,28 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
           blockCount.set(count)
           context.system.eventStream.publish(CurrentBlockCount(count))
       }
-      client.rpcClient.invoke("getbalance").collect {
-        // We track our balance in mBTC because a rounding issue in BTC would be too impactful.
-        case JDecimal(balance) => Metrics.BitcoinBalance.withoutTags().update(balance.doubleValue * 1000)
+      // TODO: move to a separate actor that listens to CurrentBlockCount
```
stanislavkozlovski, line 103: nit: would it be better to move this to a Github Issue such that newer contributors could take a stab at this?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -100,9 +100,28 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
...
+          val filteredByStatus = Seq(
+            (Monitoring.Tags.UtxoStatuses.Confirmed, utxos.filter(_._2 > 0)),
+            (Monitoring.Tags.UtxoStatuses.Unconfirmed, utxos.filter(_._2 == 0)),
+            (Monitoring.Tags.UtxoStatuses.Safe, utxos.filter(_._3)),
+            (Monitoring.Tags.UtxoStatuses.Unsafe, utxos.filter(!_._3)),
+          )
+          filteredByStatus.foreach {
+            case (status, filteredUtxos) =>
+              val amount = filteredUtxos.map(_._1).sum
+              log.info(s"we have ${filteredUtxos.length} $status utxos ($amount mBTC)")
```
stanislavkozlovski, line 121: Would this read better if we printed one log line that has a mapping of status -> amount? A bit of log duplication (easy to misread) since we would have 4 such lines per block and potentially more if we ever make the statuses more granular

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: I have plans for this one, it will not be a good first issue: this will be the basis of our fee bumping / utxo manager component, I'll handle this one ;)

t-bast: But we do have a "good first issue" tag and some issues waiting to be picked up by new contributors! Don't hesitate to ask questions directly on these issues if one looks interesting to you.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: IIUC, you'd like to replace the current multiple lines with one single log line that would look like:

```
unconfirmed -> 0.3 mBTC (7 utxos), confirmed -> 0.2 mBTC (3 utxo), safe -> 0.5 mBTC (5 utxo)
```

I prefer separating the log lines, it makes it easier to grep only the status you're interested in and automate analysis.

stanislavkozlovski: sounds good

pm47: This only counts the tips right?

Here is a script (courtesy @sstone) which also counts the number of unconfirmed txes that we can create:

```shell
#!/bin/bash

echo "utxos that can be created from confirmed utxos:"
echo `bitcoin-cli listunspent | jq '. | length * 25'`
echo "utxos that can be created from unconfirmed utxos:"
for i in `bitcoin-cli listunspent 0 | jq --raw-output '.[] | select(.confirmations == 0) | .txid'`
do
        bitcoin-cli getmempoolentry $i | jq '. | select(.ancestorcount == 1) | 25 - .descendantcount'
done | paste -sd+ | bc
```

t-bast: Is that really helpful to track as a metric?
We'd need to refresh that every time a new mempool entry consumes one of our unspent utxos, don't we?

The question is how do you want to use this metric.
The way I see it, it's very simple: if it's becoming low we create new utxos from our cold wallet (or something similar).
I don't think the added complexity is necessary in that case, is it?

pm47: It's pretty useful, because this actually tracks our ability to publish unconfirmed txes, which the other values are a proxy for.

t-bast: But you agree that it's a lot harder to implement correctly, and more expensive as well (we'll need to recompute it basically every time the mempool changes if we want it to be correct).

Or maybe we don't care about such variations since we won't be able to react immediately anyway, and we're ok with computing this only once every block, regardless of the txs we'll submit to the mempool?

pm47: Oh yes definitely, only once every block, not in realtime.

t-bast: Allright, I'll experiment with that.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1609/commits/e3d8f222254f704653682598399303a6a4d86b3a
I chose a slightly different route, because as discussed offline estimating the number of transactions we may be able to create can be wrong by an order of magnitude.
We instead simply discard unconfirmed utxos that have too many ancestors.
And when the metric is low, we should send more funds from a cold wallet.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -251,6 +247,50 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
     }
   }
 
+  // TODO: move to a separate actor that listens to CurrentBlockCount and manages utxos for RBF
+  def checkUtxos(): Future[Unit] = {
+    case class Utxo(txId: ByteVector32, amount: MilliBtc, confirmations: Long, safe: Boolean)
+
+    def listUnspent(): Future[Seq[Utxo]] = client.rpcClient.invoke("listunspent", /* minconf */ 0).collect {
```
pm47, line 254: Is the commented `minconf` normal?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -251,6 +247,50 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
     }
   }
 
+  // TODO: move to a separate actor that listens to CurrentBlockCount and manages utxos for RBF
+  def checkUtxos(): Future[Unit] = {
```
pm47, line 251: Making this method returns a `Future` is good because it will save the parent watcher actor from failing in case of problems.

However I think an `onFailure` or `recover` with a `log.warn` would be nice to detect if such problems happen.

pm47 (commented):
Minor comment, otherwise LGTM

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: Yes, it's to indicate the name of the RPC parameter, otherwise there's no way to document that inside the code (since bitcoin requests are not typed).

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: Definitely, good idea.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1609/commits/43d726be7184e941f80b10b94f541fc8f239e4e3

pm47: This the build is failing, I'd rather do the `recover` in the method itself, it's more consistent with the rest of the file.

t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1609/commits/59b9e32e6eec41d6afae9426a6d514a710ba458b :sweat_smile:

pm47 (approved):

PR #1610, opened 2020-11-19
Set initial relay fees during channel open

### LINKED ISSUE #1507: Feature request: Open channel with specific relay and base fee setting
for some peers (like LOOP) it is necessary to open channels with a vastly higher fee than the regular one due to the one way flow characteristics of the liquidity and in order to not have them fully exhausted right away. 

Therefor it would be very helpful if channels could be opened with a specific fee setting overwriting the global setting in the config.

t-bast:
It can be useful to override the default relay fees when opening channels to specific nodes.

Note that these initial relay fees are not persisted in the DB. That means that if your node reboots before the funding transaction confirms, the channel will be opened with the default relay fees, not the overridden values.

Here is the command to open a channel with overridden relay fees:

```
eclair-cli open --nodeId=<nodeId> --fundingSatoshis=500000 --feeBaseMsat=1000 --feeProportionalMillionths=10000
```

Fixes #1507

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
@@ -77,7 +77,17 @@ case object ERR_INFORMATION_LEAK extends State
       8888888888     Y8P     8888888888 888    Y888     888     "Y8888P"
  */
 
-case class INPUT_INIT_FUNDER(temporaryChannelId: ByteVector32, fundingAmount: Satoshi, pushAmount: MilliSatoshi, initialFeeratePerKw: FeeratePerKw, fundingTxFeeratePerKw: FeeratePerKw, localParams: LocalParams, remote: ActorRef, remoteInit: Init, channelFlags: Byte, channelVersion: ChannelVersion)
+case class INPUT_INIT_FUNDER(temporaryChannelId: ByteVector32,
+                             fundingAmount: Satoshi,
+                             pushAmount: MilliSatoshi,
+                             initialFeeratePerKw: FeeratePerKw,
+                             fundingTxFeeratePerKw: FeeratePerKw,
+                             initialRelayFees_opt: Option[(MilliSatoshi, Int)],
+                             localParams: LocalParams,
+                             remote: ActorRef,
+                             remoteInit: Init,
+                             channelFlags: Byte,
+                             channelVersion: ChannelVersion)
 case class INPUT_INIT_FUNDEE(temporaryChannelId: ByteVector32, localParams: LocalParams, remote: ActorRef, remoteInit: Init, channelVersion: ChannelVersion)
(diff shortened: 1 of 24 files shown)


pm47 (commented):
Have you considered going with a simpler best-effort non-persisted implementation ?

If we go with a change that involves a db migration, we may as well squeeze in #1616?

t-bast:
> Have you considered going with a simpler best-effort non-persisted implementation ?

But this is the best effort non-persisted implementation... :thinking: 
Can you clarify why you're saying otherwise?

pm47:
>> Have you considered going with a simpler best-effort non-persisted implementation ?

> But this is the best effort non-persisted implementation... 🤔
> Can you clarify why you're saying otherwise?

🤦‍♂️  I saw that you modified channel types and associated codecs, and jumped to the conclusion.

That said even if we are not changing the database, we are still changing the codecs and a regression here could be fatal. Let me try to convince myself that there is absolutely no risk with the `provide` method.

t-bast:
I get it, yes I had to change the codec but only to provide a `None` value, so I *think* it should be ok, but it's worth an independent review ;)

I did test this E2E on regtest, with the following scenarios:

- Upgrading from the latest release
- Create channel with initial relay fees set
- Create channel with initial relay fees set but restart before it confirms

Everything looked good.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -77,7 +77,17 @@ case object ERR_INFORMATION_LEAK extends State
       8888888888     Y8P     8888888888 888    Y888     888     "Y8888P"
  */
 
-case class INPUT_INIT_FUNDER(temporaryChannelId: ByteVector32, fundingAmount: Satoshi, pushAmount: MilliSatoshi, initialFeeratePerKw: FeeratePerKw, fundingTxFeeratePerKw: FeeratePerKw, localParams: LocalParams, remote: ActorRef, remoteInit: Init, channelFlags: Byte, channelVersion: ChannelVersion)
+case class INPUT_INIT_FUNDER(temporaryChannelId: ByteVector32,
+                             fundingAmount: Satoshi,
+                             pushAmount: MilliSatoshi,
+                             initialFeeratePerKw: FeeratePerKw,
+                             fundingTxFeeratePerKw: FeeratePerKw,
+                             initialRelayFees_opt: Option[(MilliSatoshi, Int)],
```
pm47, line 85: I suppose you have considered defining a like so?
```scala
case class RelayFees(base: MilliSatoshi, proportional: Int)
```

The only issue I would have is that I'm not sure how we should then define `ChannelUpdate`. Except if we define this `RelayFees` in the scope of the channel? In the future we will probably play a lot more with those feerates.

t-bast: Yes I did experiment with this, but I felt that then it made sense to use it (almost) everywhere in the codebase if we added this class, and it touched many files, so I figured we would do it in a separate PR to avoid hiding the real change.

pm47 (approved):

PR #1612, opened 2020-11-20
Router handling of non-standard channel events

btcontract:
This tries to address what's described at https://gitter.im/ACINQ/developers?at=5fae6c8bc6fe0131d4f1ea88.

First, it seems like events from certain types of non-standard channels should be just blocked from `Router` altogether because validation is inherently connected with base chain data.

Second, I've decided to go with another `AbstractCommitments` parameter (instead of pattern matching) because some future non-standard channels may still have a base chain tx and so could be eligible for processing in `Router`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
@@ -412,7 +412,7 @@ object Validation {
     } else {
       log.debug("ignoring announcement {} (unknown channel)", u)
       sendDecision(origins, GossipDecision.NoRelatedChannel(u))
-      d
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -412,7 +412,7 @@ object Validation {
     } else {
       log.debug("ignoring announcement {} (unknown channel)", u)
       sendDecision(origins, GossipDecision.NoRelatedChannel(u))
-      d
+      d.copy(privateChannels = d.privateChannels - u.shortChannelId)
```
btcontract, line 415: I think this change should be applied irregardless of other changes because without removal HCs will be staying in this map until `LocalChannelDown` is received which is somewhat undesired.

PR #1613, opened 2020-11-22
Improve CustomCommitmentsPlugin methods

btcontract:
This is a bunch of small changes to make `CustomCommitmentsPlugin` methods easier to works with.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
@@ -51,7 +51,7 @@ trait CustomCommitmentsPlugin extends PluginParams {
    * expire. If your plugin defines non-standard HTLCs, and they need to be automatically failed, they should be
    * returned by this method.
    */
-  def getIncomingHtlcs: Seq[IncomingHtlc]
+  def getIncomingHtlcs(nodeParams: NodeParams): Seq[IncomingHtlc]
 
   /**
    * Outgoing HTLC sets that are still pending may either succeed or fail: we need to watch them to properly forward the
--- a/eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
@@ -16,6 +16,7 @@
 
 package fr.acinq.eclair
 
+import akka.event.LoggingAdapter
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.channel.Origin
@@ -51,7 +52,7 @@ trait CustomCommitmentsPlugin extends PluginParams {
    * expire. If your plugin defines non-standard HTLCs, and they need to be automatically failed, they should be
    * returned by this method.
    */
-  def getIncomingHtlcs(nodeParams: NodeParams): Seq[IncomingHtlc]
+  def getIncomingHtlcs(nodeParams: NodeParams)(implicit log: LoggingAdapter): Seq[IncomingHtlc]
 
   /**
    * Outgoing HTLC sets that are still pending may either succeed or fail: we need to watch them to properly forward the
--- a/eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
@@ -52,12 +52,12 @@ trait CustomCommitmentsPlugin extends PluginParams {
    * expire. If your plugin defines non-standard HTLCs, and they need to be automatically failed, they should be
    * returned by this method.
    */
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
```diff
@@ -51,7 +52,7 @@ trait CustomCommitmentsPlugin extends PluginParams {
    * expire. If your plugin defines non-standard HTLCs, and they need to be automatically failed, they should be
    * returned by this method.
    */
-  def getIncomingHtlcs: Seq[IncomingHtlc]
+  def getIncomingHtlcs(nodeParams: NodeParams)(implicit log: LoggingAdapter): Seq[IncomingHtlc]
```
btcontract, line 55: `getIncomingHtlcs` now takes `NodeParams` (because plugin sees `Kit` class too late in `onKit` method while `getIncomingHtlcs` is called at `Setup` phase) and `LoggingAdapter` (because `IncomingPacket.decrypt` is reused in plugin and it takes a `LoggingAdapter` implicitly).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -251,6 +251,8 @@ class PostRestartHtlcCleaner(nodeParams: NodeParams, register: ActorRef, initial
 
 object PostRestartHtlcCleaner {
 
+  type ChannelIdAndHtlcId = (ByteVector32, Long)
```
btcontract, line 254: I've decided to add this because I've had troubles immediately understanding what exactly `(ByteVector32, Long)` represents.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -329,9 +331,25 @@ object PostRestartHtlcCleaner {
   private def isPendingUpstream(channelId: ByteVector32, htlcId: Long, htlcsIn: Seq[IncomingHtlc]): Boolean =
     htlcsIn.exists(htlc => htlc.add.channelId == channelId && htlc.add.id == htlcId)
 
+  def groupByOrigin(htlcsOut: Seq[(Origin, ByteVector32, Long)], htlcsIn: Seq[IncomingHtlc]): Map[Origin, Set[ChannelIdAndHtlcId]] =
```
btcontract, line 334: This piece of code is non-trivial and reused in plugin so I've made it a separate method similar to `PostRestartHtlcCleaner.decryptedIncomingHtlcs`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -329,9 +331,25 @@ object PostRestartHtlcCleaner {
...
+    htlcsOut
+      .groupBy { case (origin, _, _) => origin }
+      .view
```
btcontract, line 337: Added a view here since `mapValues` without it was marked as deprecated.

eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
t-bast: Then it makes sense to do the same for `getHtlcsRelayedOut` (add `nodeParams` and the `LoggingAdapter`), it will always be useful in some cases isn't it?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
t-bast: But you're only doing half of the job, because the function declaration in `PluginParams` returns a `Set[(ByteVector32, Long)]`.
This needs to be consistent, either you introduce a `case class HtlcAndChannelIds` and use it everywhere, or you keep `(ByteVector32, Long)` (which IMHO is good enough).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
t-bast: Please revert, this is done in order to keep compatibility with eclair-mobile / phoenix who are on older akka versions.
We will fix these deprecation warnings once phoenix / eclair-mobile don't depend on eclair-core anymore.

eclair-core/src/main/scala/fr/acinq/eclair/PluginParams.scala
btcontract: Probably won't hurt, also I've decided to remove `implicit` part from log since that seems too scala-specific while plugins could be done in other jvm languages.

t-bast (approved):

PR #1614, opened 2020-11-23
Move reusable API parts to a trait

btcontract:
This takes all reusable parts from `Service` trait and puts them into a more generally applicable `AbstractService` trait. The goal is to provide reusable parts for plugins which would establish API of their own (HC is such a plugin).

### DIFF
--- a/eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
+++ b/eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
@@ -77,7 +77,7 @@ object Boot extends App with Logging {
         override val actorSystem = system
         override val password = apiPassword
         override val eclairApi: Eclair = new EclairImpl(kit)
-      }.route
+      }.finalRoute
       Http().newServerAt(config.getString("api.binding-ip"), config.getInt("api.port")).bindFlow(apiRoute).recover {
         case _: BindFailedException => onError(TCPBindException(config.getInt("api.port")))
       }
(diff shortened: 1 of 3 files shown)


btcontract:
`Materializer` is not needed as of akka 2.6 (https://stackoverflow.com/questions/58768399/akka-stream-2-6-how-to-create-actormaterializer)

btcontract:
What about this one, especially in light of https://github.com/ACINQ/eclair/issues/1640?

FWIW my opinion is current approach is good for these reasons:
1. API is a basic function which users expect to be packaged, not optional.
2. Once boilerplate code is factored out defining API becomes pretty straightforward both in core and plugins.
3. Having a common base interface feels better than letting each API-enabled plugin to build this completely from scratch.

t-bast:
We have plans to heavily re-work the JSON API in the not-too-distant future, so I don't think we'll take in changes like the ones you propose until we've figured out where we want to go in more details.

Extending the API for custom nodes can very easily be done in a fork for now, it's a part that doesn't move much so it's easy to keep up-to-date with `master`.

btcontract:
This is somewhat unfortunate as I hope to be not the only plugin user. Maybe this can be used as temporary solution for the next release for example, and on next releases with API changes I'll update accordingly. If not I guess I'll just copy-paste current API code into plugin which is a bit ugly but will work.

btcontract:
Will keep it open to give it a chance to be considered still.

btcontract:
Just forgot about this one, not relevant any more indeed.

PR #1615, opened 2020-11-23
Ack htlc settlement commands after writing state

pm47:
There are two issues:
1. because we forward commands *before* writing to disk in
   `PendingRelayDb.safeSend` in order to reduce latency, there is a race
   where the channel can process and acknowledge the command before the
   db write. As a result, the command stays in the pending relay db and
   will be cleaned up by the post-restart-htlc-cleaner at next restart.
2. in the general case, the channel acknowledges commands *before* it
   writes its state to disk, which opens a window for losing the command
   if we stop eclair at that exact time.

In order to fix 2., we introduce a new `acking` transition method, which
will be called after `storing`. This method adds a delay before actually
acknowledging the commands, which should be more than enough to solve 1.

I'm not sure we need that additional delay, because now that we
acknowledge the commands *after* storing the state, the channel should
lose the race most of the time.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -657,8 +657,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           handleCommandSuccess(c, d.copy(commitments = commitments1)) sending fulfill
         case Failure(cause) =>
           // we acknowledge the command right away in case of failure
-          PendingRelayDb.ackCommand(nodeParams.db.pendingRelay, d.channelId, c)
-          handleCommandError(cause, c)
+          handleCommandError(cause, c) acking(d.channelId, c)
       }
 
     case Event(fulfill: UpdateFulfillHtlc, d: DATA_NORMAL) =>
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2390,6 +2382,23 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+     * @param cmd fail/fulfill command that has been processed
+     */
+    def acking(channelId: ByteVector32, cmd: HtlcSettlementCommand): FSM.State[fr.acinq.eclair.channel.State, Data] = {
+      log.debug("scheduling acknowledgement of cmd id={}", cmd.id)
+      context.system.scheduler.scheduleOnce(10 seconds)(PendingRelayDb.ackCommand(nodeParams.db.pendingRelay, channelId, cmd))(context.system.dispatcher)
+      state
+    }
+
+    def acking(updates: List[UpdateMessage]): FSM.State[fr.acinq.eclair.channel.State, Data] = {
+      log.debug("scheduling acknowledgement of cmds ids={}", updates.collect { case s: HtlcSettlementMessage => s.id }.mkString(","))
```
t-bast, line 2397: nit:

```suggestion
      log.debug("scheduling acknowledgement of htlc ids={}", updates.collect { case s: HtlcSettlementMessage => s.id }.mkString(","))
```

t-bast (commented):
LGTM, I think the additional delay doesn't hurt, let's keep it.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: We really acknowledge commands, don't we?

t-bast: That's true, it makes sense.

t-bast (approved):

PR #1617, opened 2020-11-27
Allow overriding of Tor target

pm47:
This is useful to redirect incoming tor traffic to eclair-front.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -206,7 +206,7 @@ eclair {
     host = "127.0.0.1"
     port = 9051
     private-key-file = "tor.dat"
-    targets = [] // a list of [address:]port, for advanced use (e.g. to send traffic to front servers)
+    targets = [] // a list of address:port, for advanced use (e.g. to send traffic to front servers). See the tor man page for syntax details.
   }
 
   db {
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -206,6 +206,7 @@ eclair {
     host = "127.0.0.1"
     port = 9051
     private-key-file = "tor.dat"
+    targets = [] // a list of [address:]port, for advanced use (e.g. to send traffic to front servers)
   }
 
   db {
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -206,6 +206,7 @@ eclair {
     host = "127.0.0.1"
     port = 9051
     private-key-file = "tor.dat"
+    targets = [] // a list of [address:]port, for advanced use (e.g. to send traffic to front servers)
```
t-bast, line 209: Do you really mean [address:]port and not [address:port]?
That means the [address:] part is optional?
If that's so, it's worth a clearer comment, and maybe changes to `docs/Tor.md` to describe this feature.

pm47: Yes, actually both the address and port are optional:

> HiddenServicePort VIRTPORT [TARGET]
Configure a virtual port VIRTPORT for a hidden service. You may use this option multiple times; each time applies to the service using the most recent HiddenServiceDir. By default, this option maps the virtual port to the same port on 127.0.0.1 over TCP. You may override the target port, address, or both by specifying a target of addr, port, addr:port, or unix:path. (You can specify an IPv6 target as [addr]:port. Unix paths may be quoted, and may use standard C escapes.) You may also have multiple lines with the same VIRTPORT: when a user connects to that VIRTPORT, one of the TARGETs from those lines will be chosen at random. Note that address-port pairs have to be comma-separated.

To be frank I even hesitated documenting the feature at all. I'm afraid getting into to much details will be more confusing. I figure users who need this advanced feature will know what they are doing.

```suggestion
    targets = [] // a list of address, port, or address:port, for advanced use (e.g. to send traffic to front servers). Examplel: ["a1:p1", "a2", "a3", "p4"]. Address defaults to 127.0.0.1 and port to 9735.
```

t-bast: Makes sense, maybe just provide a link to the corresponding Tor documentation, and only users that understand it and know what they are doing will modify this?

pm47: I'm a bit wary of linking to https://2019.www.torproject.org/docs/tor-manual.html.en, with the risk of broken links.

How about a polite RTFM?

```suggestion
    targets = [] // a list of address:port, for advanced use (e.g. to send traffic to front servers). See the tor man page for syntax details.
```

t-bast: Sounds good to me :grinning:

t-bast (approved):

Issue #1619, opened 2020-11-28
Overwriting from 4.0 with 4.1 appears to fail.


dscotese:
1. I downloaded the version 4.1 zip file and overwrote the contents of my eclair folder (from which I have been running my node successfully) with the contents of the zip file.
2. I ran the eclair.bat file.
3. I used a bash window to run `ecliar-cli getinfo`

I expected it to report that it was running `  "version": "0.4.1-e5fb281",`

It reported that it was still running `  "version": "0.4-69c538e",`

I tracked this down to an assumption in eclair.bat that `dir /b` in Windows command prompt will list files putting xyz.4.jar before xyz.4.1.jar, and that's not true.  The result is that `APP_ENTRYPOINT` is set to the version 4.0 jar instead of the 4.1 jar.  If we continue using the convention of leaving the .0 off of the initial major version, this will happen for every upgrade done similar to mine.  I am about to make a pull request for an update to the batch file which orders the `eclair-node*` files by their timestamp, under the assumption that the latest `eclair-node*.jar` file will be the one with the highest version number.

t-bast:
Fixed with eclair 0.5.0.

PR #1620, opened 2020-11-28
Find the latest version of eclair-node jar

dscotese:
Until the build process is updated so that changing the major version number x to x+1 appends the `.0` to the new version number in the filename, Windows needs a different strategy to find the highest version number.  I chose to use the eclair-node*.jar file with the latest write time.  This is a fix for #1619 .

### DIFF
--- a/eclair-node/src/main/resources/eclair-node.bat
+++ b/eclair-node/src/main/resources/eclair-node.bat
@@ -48,7 +48,7 @@ rem "-J" is stripped, "-D" is left as is, and everything is appended to JAVA_OPT
 set _JAVA_PARAMS=
 set _APP_ARGS=
 
-for /f %%i in ('dir /b %APP_LIB_DIR%\eclair-node*') do set APP_ENTRYPOINT=%%i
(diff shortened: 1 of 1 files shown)


t-bast:
I don't have a windows machine so I can't test this, but I'm wary of modifying these standard scripts.

IIUC, this issue only appears when you have both versions of eclair on your machine, right?
I don't think it's really the responsibility of this script to choose between multiple eclair executables on your machine.
It's your responsibility as a node operator to ensure there's no ambiguity there, otherwise there's an operational risk for you.

dscotese:
Yes, you must have the version that has no .0 on the end of it.  I agree that the node operator should ensure there's no ambiguity, so the installation instructions should always indicate that older versions of the jar file should be deleted manually, given that #1619 is still open.

t-bast:
We've released 0.5.0, and will make sure we don't forget the trailing `.0` in future releases, so I think this PR is not necessary and can be closed.

t-bast:
Closing this PR as it's not an issue anymore with eclair 0.5.0 (and future versions).

PR #1622, opened 2020-11-30
[CI] Exclude tests that call external APIs

t-bast:
These tests have been flaky because external APIs tend to be down or throttle our calls. and we don't want our test suite to fail because of that.

In practice it's enough if developers run these tests locally once in a while.


pm47 (commented):
Nice trick

pm47 (approved):

t-bast:
:sunglasses:

PR #1624, opened 2020-12-07
Sync fix

btcontract:
Make `ReplyShortChannelIdsEnd` low priority in `TransportHandler`, otherwise it is often sent before the related bunch of gossip messages is done transmitting which breaks sync logic for remote peers.


btcontract:
Does it make sense to make the whole `RoutingMessage` trait as low priority instead of just those 4 subtypes?

pm47:
Hi Anton, great find! Does this solve your sync issue?

> Does it make sense to make the whole RoutingMessage trait as low priority instead of just those 4 subtypes?

Maybe, but need to double check for `AnnouncementSignatures`.

btcontract:
> Does this solve your sync issue?

Yes, it does.

pm47 (approved):
Acking as-is, because this change is small, testing this is hard and the risk of a regression is small.

PR #1625, opened 2020-12-08
Remove old compatibility workaround for eclair mobile

pm47:


t-bast (approved):
Glad to get rid of this :sweat_smile:

PR #1630, opened 2020-12-14
Settle HTLCs revoked commit

t-bast:
When a revoked commitment is published, we didn't correctly settle pending HTLCs, which could lead to upstream channel closure.

This would not cause a loss of funds (especially since we would gain funds from the revoked channel) but it's a temporary liquidity loss that we'd like to avoid.

I took this opportunity to rework the revoked close tests to cover more scenarios.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -976,7 +976,10 @@ object Helpers {
         // their last commitment got confirmed, so no htlcs will be overridden, they will timeout or be fulfilled on chain
         Set.empty
       } else if (d.revokedCommitPublished.map(_.commitTx.txid).contains(tx.txid)) {
-        // a revoked commitment got confirmed: we will claim its outputs, but we also need to fail htlcs that are pending in the latest commitment
+        // a revoked commitment got confirmed: we will claim its outputs, but we also need to fail htlcs that are pending in the latest commitment:
+        //  - outgoing htlcs that are in the local commitment but not in remote/nextRemote have already been fulfilled/failed so we don't care about them
+        //  - outgoing htlcs that are in the remote/nextRemote commitment may not really be overridden, but since we are going to claim their output as a
+        //    punishment we will never get the preimage and may as well consider them failed in the context of relaying htlcs
         nextRemoteCommit_opt.getOrElse(remoteCommit).spec.htlcs.collect(incoming)
       } else {
         Set.empty
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -20,6 +20,7 @@ import akka.event.{DiagnosticLoggingAdapter, LoggingAdapter}
 import fr.acinq.bitcoin.Crypto.{PrivateKey, PublicKey, ripemd160, sha256}
 import fr.acinq.bitcoin.Script._
 import fr.acinq.bitcoin._
+import fr.acinq.eclair._
 import fr.acinq.eclair.blockchain.EclairWallet
 import fr.acinq.eclair.blockchain.fee.{FeeEstimator, FeeTargets, FeeratePerKw, FeerateTolerance}
 import fr.acinq.eclair.channel.Channel.REFRESH_CHANNEL_UPDATE_INTERVAL
@@ -32,7 +33,6 @@ import fr.acinq.eclair.transactions.Scripts._
 import fr.acinq.eclair.transactions.Transactions._
 import fr.acinq.eclair.transactions._
 import fr.acinq.eclair.wire._
-import fr.acinq.eclair._
 import scodec.bits.ByteVector
 
 import scala.concurrent.Await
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -1057,17 +1045,19 @@ object Helpers {
...
         // is this the commit tx itself ? (we could do this outside of the loop...)
         val isCommitTx = revokedCommitPublished.commitTx.txid == tx.txid
         // does the tx spend an output of the remote commitment tx?
         val spendsTheCommitTx = revokedCommitPublished.commitTx.txid == outPoint.txid
-        // is the tx one of our 3rd stage delayed txes? (a 3rd stage tx is a tx spending the output of an htlc tx, which
+        // is the tx one of our 3rd stage delayed txs? (a 3rd stage tx is a tx spending the output of an htlc tx, which
         // is itself spending the output of the commitment tx)
         val is3rdStageDelayedTx = revokedCommitPublished.claimHtlcDelayedPenaltyTxs.map(_.txid).contains(tx.txid)
-        isCommitTx || spendsTheCommitTx || is3rdStageDelayedTx
+        // does the tx spend an output of an htlc tx? (in which case it may invalidate one of our claim-htlc-delayed-penalty)
```
pm47, line 1058: I don't understand this. In what scenario would that happen?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -1115,11 +1105,13 @@ object Helpers {
...
-        .filter(tx => (tx.txIn.map(_.outPoint.txid).toSet -- revokedCommitPublished.irrevocablySpent.values).isEmpty) // only the txes which parents are already confirmed may get confirmed (note that this also eliminates outputs that have been double-spent by a competing tx)
-        .filterNot(tx => revokedCommitPublished.irrevocablySpent.values.toSet.contains(tx.txid)) // has the tx already been confirmed?
-      isCommitTxConfirmed && commitOutputsSpendableByUs.isEmpty && unconfirmedHtlcDelayedTxes.isEmpty
+      // which htlc delayed txs can we expect to be confirmed?
+      val unconfirmedHtlcDelayedTxs = revokedCommitPublished.claimHtlcDelayedPenaltyTxs
+        // only the txs which parents are already confirmed may get confirmed (note that this also eliminates outputs that have been double-spent by a competing tx)
+        .filter(tx => (tx.txIn.map(_.outPoint.txid).toSet -- revokedCommitPublished.irrevocablySpent.values).isEmpty)
+        // if one of the tx inputs has been spent, the tx has already been confirmed or a competing tx has been confirmed
+        .filterNot(tx => tx.txIn.exists(txIn => revokedCommitPublished.irrevocablySpent.contains(txIn.outPoint)))
```
pm47, line 1113: This is giving me a headache. Aren't we changing the meaning of `irrevocablySpent`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -990,6 +975,9 @@ object Helpers {
       } else if (nextRemoteCommit_opt.map(_.txid).contains(tx.txid)) {
         // their last commitment got confirmed, so no htlcs will be overridden, they will timeout or be fulfilled on chain
         Set.empty
+      } else if (d.revokedCommitPublished.map(_.commitTx.txid).contains(tx.txid)) {
+        // a revoked commitment got confirmed: we will claim its outputs, but we also need to fail htlcs that are pending in the latest commitment
+        nextRemoteCommit_opt.getOrElse(remoteCommit).spec.htlcs.collect(incoming)
```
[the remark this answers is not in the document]
t-bast: >  there should be some comparisons with existing local

Why? It feels to me that it's exactly the same case as elsewhere: we want to fail every pending signed HTLC (they will never be settled since the remote published a revoked commitment, which means that all the content of this channel can be ignored). All our pending HTLCs have been overridden by this revoked commitment, so the naming feels correct to me.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: It happens when the remote spends one of his htlc txs (ie we failed to claim our penalty tx).
In that case our claim-htlc-penalty tx will never confirm, but we still want to move to the closed state (we acknowledge that this output is not spendable anymore, we won't be able to claim it).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: I don't see why...one of our inputs has been irrevocably spent, so our tx is now obsolete and will never confirm. Can you detail?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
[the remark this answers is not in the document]
t-bast: > since we are going to claim their output as a punishment we will never get the preimage and may as well consider them failed in the context of relaying htlcs

That's part of the intuition indeed. Once the remote has published a revoked tx, every HTLC in the current commit tx can be safely failed upstream. There should be no way the downstream peer can claim these HTLCs, we should claim them all via penalty txs (but without learning the preimage).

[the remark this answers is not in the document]
pm47: My point is that the original meaning of this method is to identify `htlcs that will never reach the blockchain`. In the case of a revoked close, those htlcs will hit the chain, but we decide to fail them for other reason. That's why I'm proposing to add the second bullet point above as comment. The first bullet point probably can be ignored.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: I thought that in all other cases the `irrevocablySpent` map only contained transactions that are in the `xxxPublished` object, but that's wrong: there are other instances where we have competing transactions, and we use that code:
```scala
  // does the tx spend an output of the local commitment tx?
  val spendsTheCommitTx = localCommitPublished.commitTx.txid == outPoint.txid
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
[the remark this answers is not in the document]
t-bast: Got it, it's indeed a weird case because these HTLCs will hit the chain, but not in a meaningful way - this doesn't count as normal htlc settlement. So it could be interpreted in different ways, I choose to interpret it as a punishment to the remote, but not as HTLCs being settled.

I'll update the comment.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Yes, we have to, we record in `irrevocablySpent` anything that confirms and consumes an outpoint we're interested in.
This is clearly a headache-heavy part of our codebase :)

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1630/commits/1ad2c80a7078dd7f321cecc5e82c4f86ead7c91f

pm47 (approved):

PR #1631, opened 2020-12-14
Move fee provider conf

pm47:
Took the opportunity to improve the naming as well.

I also considered putting them under a  `providers` section, WDYT?
```
on-chain-fees {
  providers {
    min-feerate = 1 // minimum feerate in satoshis per byte
    smoothing-window = 6 // 1 = no smoothing
    timeout = 5 seconds // max time we'll wait for answers from a fee provider before we fallback to the next one
  }
...
}
```

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -36,10 +36,6 @@ eclair {
     zmqtx = "tcp://127.0.0.1:29000"
   }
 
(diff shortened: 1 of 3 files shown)


t-bast (approved):
LGTM.

I don't think we need to add a `providers` section, since we'll eventually get rid of providers like bitgo and earndotcom and instead only use bitcoin core.

PR #1633, opened 2020-12-14
Accept remote's closing fee when we have nothing at stake

pm47:
If we have nothing at stake there is no need to negotiate and we accept
their fee right away.

⚠️ This is a small change but a bug here would be costly.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1175,16 +1175,14 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         case Success(signedClosingTx) =>
           // if we are fundee and we were waiting for them to send their first closing_signed, we don't have a lastLocalClosingFee, so we compute a firstClosingFee
           val lastLocalClosingFee = d.closingTxProposed.last.lastOption.map(_.localClosingSigned.feeSatoshis)
-          val nextClosingFee = Closing.nextClosingFee(
-            localClosingFee = lastLocalClosingFee.getOrElse {
-              if (d.commitments.localCommit.spec.toLocal == 0.msat) {
-                // if we have nothing at stake there is no need to negotiate and we accept their fee right away
-                remoteClosingFee
-              } else {
-                Closing.firstClosingFee(d.commitments, d.localShutdown.scriptPubKey, d.remoteShutdown.scriptPubKey, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)
-              }
-            },
-            remoteClosingFee = remoteClosingFee)
+          val nextClosingFee = if (d.commitments.localCommit.spec.toLocal == 0.msat) {
+            // if we have nothing at stake there is no need to negotiate and we accept their fee right away
+            remoteClosingFee
+          } else {
+            Closing.nextClosingFee(
+              localClosingFee = lastLocalClosingFee.getOrElse(Closing.firstClosingFee(d.commitments, d.localShutdown.scriptPubKey, d.remoteShutdown.scriptPubKey, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)),
+              remoteClosingFee = remoteClosingFee)
+          }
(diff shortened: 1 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/channel/states/g/NegotiatingStateSpec.scala
```diff
@@ -137,6 +137,18 @@ class NegotiatingStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike
...
+  test("recv ClosingSigned (nothing at stake)", Tag("no_push_msat")) { f =>
+    import f._
+    val aliceCloseFee = alice2bob.expectMsgType[ClosingSigned].feeSatoshis
+    alice2bob.forward(bob)
+    val bobCloseFee = bob2alice.expectMsgType[ClosingSigned].feeSatoshis
+    assert(aliceCloseFee === bobCloseFee)
+    val mutualCloseTx = bob2blockchain.expectMsgType[PublishAsap].tx
+    assert(bob.stateData.asInstanceOf[DATA_CLOSING].mutualClosePublished == List(mutualCloseTx))
+    assert(bob2blockchain.expectMsgType[WatchConfirmed].event === BITCOIN_TX_CONFIRMED(mutualCloseTx))
+    alice ! WatchEventSpent(BITCOIN_FUNDING_SPENT, mutualCloseTx)
```
t-bast, line 149: It feels like this test isn't finished...don't you want to test that both parties correctly go to the `CLOSED` state once the tx confirms?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1176,7 +1176,14 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           // if we are fundee and we were waiting for them to send their first closing_signed, we don't have a lastLocalClosingFee, so we compute a firstClosingFee
           val lastLocalClosingFee = d.closingTxProposed.last.lastOption.map(_.localClosingSigned.feeSatoshis)
           val nextClosingFee = Closing.nextClosingFee(
-            localClosingFee = lastLocalClosingFee.getOrElse(Closing.firstClosingFee(d.commitments, d.localShutdown.scriptPubKey, d.remoteShutdown.scriptPubKey, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)),
+            localClosingFee = lastLocalClosingFee.getOrElse {
```
t-bast, line 1179: I think it would be clearer to completely skipe the `Closing.nextClosingFee` call in that case:

```scala
val nextClosingFee = if (d.commitments.localCommit.spec.toLocal == 0.msat) {
    // if we have nothing at stake there is no need to negotiate and we accept their fee right away
    remoteClosingFee
} else {
    Closing.nextClosingFee(
        localClosingFee = lastLocalClosingFee.getOrElse(Closing.firstClosingFee(d.commitments, d.localShutdown.scriptPubKey, d.remoteShutdown.scriptPubKey, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)),
        remoteClosingFee = remoteClosingFee)
}
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1176,7 +1176,14 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+              if (d.commitments.localCommit.spec.toLocal == 0.msat) {
```
t-bast, line 1180: NB: this is only possible for 0-reserve channels, right?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/g/NegotiatingStateSpec.scala
pm47: Fixed in 772ffd9.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Fixed in 772ffd9.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: It also applies to unused channels with no `push_msat`.

t-bast (approved):

PR #1634, opened 2020-12-14
fixup! Distribute connection-handling on multiple machines (#1566)

pm47:
This fixes the Docker build which was failing due to the new child module `eclair-front`.

We should probably add a way to start an `eclair-front` from docker too.


t-bast (approved):

t-bast:
> We should probably add a way to start an eclair-front from docker too.

I don't know, since we don't recommend running it via docker for any kind of meaningful deployment, I'm not sure it's useful, is it?

pm47:
>> We should probably add a way to start an eclair-front from docker too.

> I don't know, since we don't recommend running it via docker for any kind of meaningful deployment, I'm not sure it's useful, is it?

True!

PR #1637, opened 2020-12-15
Fix duplicate implicit btc amount conversions

### LINKED ISSUE #1632: Remove duplicate implicit methods for bitcoin amounts conversion
They conflict with the ones in `bitcoin-lib` since https://github.com/ACINQ/bitcoin-lib/commit/a89851c04b197eb8663d38db58f0f9fc1f6ee05b.

```
Note that implicit conversions are not applicable because they are ambiguous:
[ERROR] both method SatoshiLong in package bitcoin of type (n: Long): fr.acinq.bitcoin.package.SatoshiLong
[ERROR] and method LongToBtcAmount in package eclair of type (l: Long): fr.acinq.eclair.package.LongToBtcAmount
[ERROR] are possible conversion functions from Long(150000L) to ?{def sat: ?}
```

t-bast:
Bitcoin-lib introduces conversions from `Long` to bitcoin amounts.
It clashes with the ones previously defined in eclair and was causing issues in IntelliJ IDEA.

The meaningful change is in `package.scala`, the rest is fixing imports.

Fixes #1632

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/package.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/package.scala
@@ -114,7 +114,7 @@ package object eclair {
     override def toFloat(x: MilliSatoshi): Float = x.toLong
     override def toDouble(x: MilliSatoshi): Double = x.toLong
     override def compare(x: MilliSatoshi, y: MilliSatoshi): Int = x.compare(y)
-    override def parseString(str: String): Option[MilliSatoshi] = None
+    override def parseString(str: String): Option[MilliSatoshi] = ???
     // @formatter:on
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/package.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/package.scala
@@ -16,13 +16,12 @@
 
 package fr.acinq
 
-import java.security.SecureRandom
-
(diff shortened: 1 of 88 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -120,7 +114,7 @@ package object eclair {
     override def toFloat(x: MilliSatoshi): Float = x.toLong
     override def toDouble(x: MilliSatoshi): Double = x.toLong
     override def compare(x: MilliSatoshi, y: MilliSatoshi): Int = x.compare(y)
-    override def parseString(str: String): Option[MilliSatoshi] = ???
+    override def parseString(str: String): Option[MilliSatoshi] = None
```
pm47, line 117: Wasn't the `???` (which throws an `NotImplementedException`) actually better here?

t-bast: I did that to align with bitcoin-lib, which implements this as `None`, but I can revert.
It's true that since we don't want to use it, we probably want to raise an exception.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1637/commits/9ebf61f61aba31ef08b4f295877f4f7b77c25272

pm47 (approved):

PR #1642, opened 2020-12-15
Disable the distibuted pubsub extension by default

pm47:
It only needs to be enabled in cluster mode.

### DIFF
--- a/eclair-node/src/main/resources/application.conf
+++ b/eclair-node/src/main/resources/application.conf
@@ -3,7 +3,7 @@ eclair {
 }
 
 akka {
-  actor.provider = local // set to cluster when using eclair-front
+  actor.provider = local // set to cluster when using eclair-front (must also enable the distributed pubsub extension below)
   remote {
     artery {
       canonical.hostname = "127.0.0.1"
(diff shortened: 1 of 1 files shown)


eclair-node/src/main/resources/application.conf
```diff
@@ -27,9 +27,10 @@ akka {
   cluster.roles = [backend]
   coordinated-shutdown.terminate-actor-system = on
   coordinated-shutdown.exit-jvm = on
-  //It is recommended to load the extension when the actor system is started by defining it in akka.extensions
-  //configuration property. Otherwise it will be activated when first used and then it takes a while for it to be populated.
-  extensions = ["akka.cluster.pubsub.DistributedPubSub"]
+  // It is recommended to load the extension when the actor system is started by defining it in akka.extensions
+  // configuration property. Otherwise it will be activated when first used and then it takes a while for it to be populated.
+  // Uncomment the line below in cluster mode
+  #extensions = ["akka.cluster.pubsub.DistributedPubSub"]
```
t-bast, line 33: nit: why a comment with `#` whereas we use `//` everywhere else?

pm47: It was to separate between comment and disable feature, we seem to be using this convention in `reference.conf` (cf `backup-notify-script`, `override-features`...)

t-bast: Ok I thought so, makes sense.

t-bast (approved):

PR #1646, opened 2020-12-16
Documentation for cluster mode

pm47:


eclair-front/src/main/resources/application.conf
```diff
@@ -3,12 +3,15 @@ eclair {
...
-    priv-key-provider = "aws-sm" // aws-sm (AWS Secrets Manager) or env (environment variable)
+    priv-key-provider = "seed" // aws-sm (AWS Secrets Manager) or env (environment variable) or seed (seed.dat)
     priv-key = ${?NODE_PRIV_KEY} // used if priv-key-provider = env
     aws-sm.priv-key-name = "node-priv-key" // used if priv-key-provider = aws-sm
     // As a security measure, we also require the pub key, which will be matched against the priv key to make sure the
     // front is really using the expected key
     pub = ${NODE_PUB_KEY}
+
+    backend.ip = 127.0.0.1
+    backend.ip = ${?BACKEND_IP}
```
t-bast, line 14: That looks like a duplicate, the second one is probably the correct one? Or is it a way to set it from the environment variable if available, otherwise default to `127.0.0.1`?

docs/Cluster.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+NB: we override the ports, otherwise they would conflict since in this example everything runs on the same server. You can run multiple `frontend`s on the same server, just make sure to change the ports.
+
+### Production setup
+
+In production you should:
+- run multiple `frontend`s
+- run one app per server
+- enable `tcp-tls` to encrypted between members of the cluster with your own generated certificate (see below)
```
t-bast, line 77:
```suggestion
- enable `tcp-tls` to encrypt communications between members of the cluster with your own generated certificate (see below)
```

docs/Cluster.md
```diff
@@ -0,0 +1,177 @@
...
+> - Adding nodes to the cluster is simple as the key material can be deployed / distributed to the new node.
+
+Generate a self-signed certificate (set a strong password):
+```shell
+$ keytool -genkeypair -v \
+          -keystore akka-cluster-tls.jks \
+          -dname "O=ACME, C=FR" \
+          -keypass:env <password> \
+          -storepass:env <password> \
+          -keyalg RSA \
```
t-bast, line 98: Why do you use RSA and not elliptic curves (which are safer, faster and create smaller certificates)?

docs/Cluster.md
```diff
@@ -0,0 +1,177 @@
...
+For convenience, we provide a prebuilt AWS Beanstalk bundle for the `frontend` (choose a WebServer environment type, and Java platform).
+
+You can run it as-is for testing.
+
+#### TLS encryption
+
+If you intend to use it in production, you need to enable encryption with your own certificate:
+1. Follow the procedure above to generate your `akka-tls.jks`
+
+2. We recommend forking the project and build your own bundle:
```
t-bast, line 125: nit:

```suggestion
2. We recommend forking the project and building your own bundle:
```

docs/Cluster.md
```diff
@@ -0,0 +1,177 @@
...
+$ vi eclair-core/src/main/reference.conf # set akka.remote.artery.transport = "tls-tcp"
+$ cp akka-cluster-tls.jks eclair-front/modules/awseb/ # copy the file you generated
+$ vi eclair-front/modules/awseb.xml # uncomment the relevant parts
+$ mvn package -DskipTests
+```
+Alternatively, you can also edit the existing bundle and manually add the `akka-cluster-tls.jks` file to the root of the zip archive. You will also need to set `akka.remote.artery.transport=tls-tcp` at runtime. 
+
+#### Private key
+
+In production, we highly recommend using AWS Secrets manager to provide the node private key. This is done by setting `eclair.front.priv-key-provider=aws-sm`. Default secret name is "node-priv-key", but it is configurable with `eclair.front.aws-sm.aws-sm.priv-key-name`
```
t-bast, line 137: Is the duplicate `aws-sm.aws-sm` expected?

t-bast (commented):
First pass, I'm testing the single-server demo right now.

eclair-front/src/main/resources/application.conf
pm47: Yes that's a way to have a default (see https://github.com/lightbend/config/blob/master/HOCON.md#substitutions).

docs/Cluster.md
pm47: I used one of the recommended algorithms in [akka doc](https://doc.akka.io/docs/akka/current/remoting-artery.html#configuring-ssl-tls-for-akka-remoting), but advanced users can change if they prefer.

docs/Cluster.md
pm47: Good catch!

```suggestion
In production, we highly recommend using AWS Secrets manager to provide the node private key. This is done by setting `eclair.front.priv-key-provider=aws-sm`. Default secret name is "node-priv-key", but it is configurable with `eclair.front.aws-sm.priv-key-name`
```

t-bast (approved):

PR #1653, opened 2020-12-18
Fix rare race conditions in integration tests

t-bast:
When a channel goes to the CLOSED state, the actor will stop itself.

We were previously sending messages to the actor asking for its state, which returns a failure when the actor is stopped. We can simply listen to state events to safely get the same result.


pm47 (approved):
Nice!

PR #1659, opened 2020-12-30
Add tx signing metrics

t-bast:
Monitor the rate at which we sign channel txs and the duration of the signing operations.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/crypto/keymanager/LocalChannelKeyManager.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/crypto/keymanager/LocalChannelKeyManager.scala
@@ -102,6 +102,7 @@ class LocalChannelKeyManager(seed: ByteVector, chainHash: ByteVector32) extends 
    * @return a signature generated with the private key that matches the input extended public key
    */
   override def sign(tx: TransactionWithInputInfo, publicKey: ExtendedPublicKey, txOwner: TxOwner, commitmentFormat: CommitmentFormat): ByteVector64 = {
+    // NB: not all those transactions are actually commit txs (especially during closing), but this is good enough for monitoring purposes
     val tags = TagSet.Empty.withTag(Tags.TxOwner, txOwner.toString).withTag(Tags.TxType, Tags.TxTypes.CommitTx)
     Metrics.SignTxCount.withTags(tags).increment()
     KamonExt.time(Metrics.SignTxDuration.withTags(tags)) {
@@ -121,6 +122,7 @@ class LocalChannelKeyManager(seed: ByteVector, chainHash: ByteVector32) extends 
    * @return a signature generated with a private key generated from the input key's matching private key and the remote point.
    */
   override def sign(tx: TransactionWithInputInfo, publicKey: ExtendedPublicKey, remotePoint: PublicKey, txOwner: TxOwner, commitmentFormat: CommitmentFormat): ByteVector64 = {
+    // NB: not all those transactions are actually htlc txs (especially during closing), but this is good enough for monitoring purposes
     val tags = TagSet.Empty.withTag(Tags.TxOwner, txOwner.toString).withTag(Tags.TxType, Tags.TxTypes.HtlcTx)
     Metrics.SignTxCount.withTags(tags).increment()
     KamonExt.time(Metrics.SignTxDuration.withTags(tags)) {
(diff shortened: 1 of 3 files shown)


pm47 (commented):
The PR is fine, but should we rename the `sign` methods to `signCommitment`, `signHtlc`, `signRevoked`?

t-bast:
It's a bit more complicated than that: the metric tags will be slightly inaccurate because some 2nd-level txs (when closing) will be incorrectly assigned to htlc / commit txs.

But the rate of closing txs will be negligible compared to the rate of signing updated commitment txs, so I think the metric can ignore this subtlety, but renaming the methods could be confusing for readers when studying the closing flows.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/keymanager/LocalChannelKeyManager.scala
```diff
@@ -97,12 +99,15 @@ class LocalChannelKeyManager(seed: ByteVector, chainHash: ByteVector32) extends
...
    * @param txOwner          owner of the transaction (local/remote)
    * @param commitmentFormat format of the commitment tx
-   * @return a signature generated with the private key that matches the input
-   *         extended public key
+   * @return a signature generated with the private key that matches the input extended public key
    */
   override def sign(tx: TransactionWithInputInfo, publicKey: ExtendedPublicKey, txOwner: TxOwner, commitmentFormat: CommitmentFormat): ByteVector64 = {
-    val privateKey = privateKeys.get(publicKey.path)
-    Transactions.sign(tx, privateKey.privateKey, txOwner, commitmentFormat)
+    val tags = TagSet.Empty.withTag(Tags.TxOwner, txOwner.toString).withTag(Tags.TxType, Tags.TxTypes.CommitTx)
```
pm47, line 105:
```suggestion
    // NB: not all those transactions are actually commit txes (especially during closing), but this is good enough for monitoring purposes
    val tags = TagSet.Empty.withTag(Tags.TxOwner, txOwner.toString).withTag(Tags.TxType, Tags.TxTypes.CommitTx)
```

eclair-core/src/main/scala/fr/acinq/eclair/crypto/keymanager/LocalChannelKeyManager.scala
```diff
@@ -113,13 +118,16 @@ class LocalChannelKeyManager(seed: ByteVector, chainHash: ByteVector32) extends
...
    * @param commitmentFormat format of the commitment tx
-   * @return a signature generated with a private key generated from the input keys's matching
-   *         private key and the remote point.
+   * @return a signature generated with a private key generated from the input key's matching private key and the remote point.
    */
   override def sign(tx: TransactionWithInputInfo, publicKey: ExtendedPublicKey, remotePoint: PublicKey, txOwner: TxOwner, commitmentFormat: CommitmentFormat): ByteVector64 = {
-    val privateKey = privateKeys.get(publicKey.path)
-    val currentKey = Generators.derivePrivKey(privateKey.privateKey, remotePoint)
-    Transactions.sign(tx, currentKey, txOwner, commitmentFormat)
+    val tags = TagSet.Empty.withTag(Tags.TxOwner, txOwner.toString).withTag(Tags.TxType, Tags.TxTypes.HtlcTx)
```
pm47, line 124:
```suggestion
    // NB: not all those transactions are actually htlc txes (especially during closing), but this is good enough for monitoring purposes
    val tags = TagSet.Empty.withTag(Tags.TxOwner, txOwner.toString).withTag(Tags.TxType, Tags.TxTypes.HtlcTx)
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1659/commits/11e54785be3e67dc6c2598dbe60c9b45191c3216

eclair-core/src/main/scala/fr/acinq/eclair/crypto/keymanager/LocalChannelKeyManager.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1659/commits/11e54785be3e67dc6c2598dbe60c9b45191c3216

pm47 (approved):

PR #1661, opened 2020-12-30
Shutdown and UpdateFee should not be intertwined

t-bast:
If a plugin (or a very reactive user) tries to mutual-close a channel right when our peer reconnects, we can end up with the following message flow:

```
UpdateFee ----->
Shutdown  ----->
<-----  Shutdown
CommigSig ----->
```

Which was not properly handled and ended up in a locked state between shutdown and negotiating.

Just like we don't initiate a mutual close while we have unsigned outgoing htlcs, we should not initiate a mutual close if we have a pending unsigned fee update. We'll be signing it almost immediately, and the `CMD_CLOSE` can be replayed after a small delay.

We also now correctly handle the transition to the negotiating state in case our peer sends us messages in this order.


eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -1803,6 +1803,23 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
     awaitCond(alice.stateName == NORMAL)
   }
 
+  test("recv CMD_CLOSE (with unsigned fee update)") { f =>
```
pm47, line 1806: Have you considered adding a test for the first commit?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -814,8 +814,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       if (d.localShutdown.isDefined) {
         handleCommandError(ClosingAlreadyInProgress(d.channelId), c)
       } else if (Commitments.localHasUnsignedOutgoingHtlcs(d.commitments)) {
-        // TODO: simplistic behavior, we could also sign-then-close
+        // NB: simplistic behavior, we could also sign-then-close
         handleCommandError(CannotCloseWithUnsignedOutgoingHtlcs(d.channelId), c)
+      } else if (Commitments.localHasUnsignedUpdateFee(d.commitments)) {
+        handleCommandError(CannotCloseWithUnsignedUpdateFee(d.channelId), c)
```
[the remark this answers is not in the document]
t-bast: I think it can be useful, and hopefully in X years we'll be able to completely get rid of it once `update_fee` is gone!

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1661/commits/d49e2f7cc295bb8bd991900c8d2203fe499c173b

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1570,7 +1570,8 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           if (d.commitments.localParams.isFunder) {
             val currentFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw
             val networkFeeratePerKw = nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(nodeParams.onChainFeeConf.feeTargets.commitmentBlockTarget)
-            if (Helpers.shouldUpdateFee(currentFeeratePerKw, networkFeeratePerKw, nodeParams.onChainFeeConf.updateFeeMinDiffRatio)) {
+            val shutdownInProgress = d.localShutdown.nonEmpty || d.remoteShutdown.nonEmpty
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1661/commits/d49e2f7cc295bb8bd991900c8d2203fe499c173b

pm47 (commented):
I think there is also this scenario, where the fundee initiates the close:

```
UpdateFee ----->
<-----  Shutdown
Shutdown  ----->
CommigSig ----->
```

t-bast:
I'll add a test case for that, I believe it will be covered by the code change already.

t-bast:
> I think there is also this scenario, where the fundee initiates the close

Done in https://github.com/ACINQ/eclair/pull/1661/commits/838d2d4a7c52069d4f08e5c49155ed335b15ced6
It showed that the channel code was not handling this correctly.
It's a subtle change, I'm not 100% sure I got it right.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -836,12 +838,11 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       //      we did not send a shutdown message
       //        there are pending signed htlcs    => go to SHUTDOWN
       //        there are no htlcs                => go to NEGOTIATING
-
       if (!Closing.isValidFinalScriptPubkey(remoteScriptPubKey)) {
         handleLocalError(InvalidFinalScript(d.channelId), d, Some(remoteShutdown))
       } else if (Commitments.remoteHasUnsignedOutgoingHtlcs(d.commitments)) {
         handleLocalError(CannotCloseWithUnsignedOutgoingHtlcs(d.channelId), d, Some(remoteShutdown))
-      } else if (Commitments.localHasUnsignedOutgoingHtlcs(d.commitments)) { // do we have unsigned outgoing htlcs?
+      } else if (Commitments.localHasUnsignedOutgoingHtlcs(d.commitments) || Commitments.localHasUnsignedUpdateFee(d.commitments)) { // do we have unsigned outgoing htlcs?
```
pm47, line 845: Nit: comments below should be updated to replace "unsigned htlcs" by "unsigned changes"

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -84,6 +84,8 @@ case class Commitments(channelVersion: ChannelVersion,
 
   def hasNoPendingHtlcs: Boolean = localCommit.spec.htlcs.isEmpty && remoteCommit.spec.htlcs.isEmpty && remoteNextCommitInfo.isRight
 
+  def hasNoPendingFeeUpdate: Boolean = (localChanges.all ++ remoteChanges.all).collectFirst { case _: UpdateFee => true }.isEmpty && remoteNextCommitInfo.isRight
```
pm47, line 87: I believe if want this method to be consisten with `hasNoPendingHtlcs` (and with the comments in Channel.scala), it should be:

```suggestion
  def hasNoPendingFeeUpdate: Boolean = (localChanges.signed ++ localChanges.acked ++ remoteChanges.acked ++ remoteChanges.signed).collectFirst { case _: UpdateFee => true }.isEmpty && remoteNextCommitInfo.isRight
```

pm47 (commented):
The `UpdateFee` handler should be removed in state `SHUTDOWN` I think, because IIUC the assumption is now that new changes cannot be added in that state (we treat `UpdateFee` exactly like `UpdateAddHtlc`) . It looks like an issue at the spec level.

t-bast:
> The UpdateFee handler should be removed in state SHUTDOWN I think, because IIUC the assumption is now that new changes cannot be added in that state

I think you're right. To avoid being too strict with senders, I changed our behavior to stop emitting `update_fee` but allow receiving them in https://github.com/ACINQ/eclair/pull/1661/commits/200c2c03f98348bb8387359bdad91641f0d0998d, let me know what you think.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: You're right! Done in https://github.com/ACINQ/eclair/pull/1661/commits/200c2c03f98348bb8387359bdad91641f0d0998d

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1661/commits/200c2c03f98348bb8387359bdad91641f0d0998d

pm47:
>>The UpdateFee handler should be removed in state SHUTDOWN I think, because IIUC the assumption is now that new changes cannot be added in that state

>I think you're right. To avoid being too strict with senders, I changed our behavior to stop emitting update_fee but allow receiving them in 200c2c0, let me know what you think.

I *think* it works but I'm a bit confused as to why. How can it work where the whole purpose of this PR was to precisely prevent that from happening.

t-bast:
I simplified the PR, it's much closer now to #1691 with additional tests.
I believe it makes sense to skip the `CMD_UPDATE_FEE` at reconnect when we're shutting down (in the `SYNCING` handler), we can always wait for the handler of `currentBlockCount` to update the fee without fear of a race condition.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -515,6 +521,10 @@ object Commitments {
 
   def remoteHasUnsignedOutgoingHtlcs(commitments: Commitments): Boolean = commitments.remoteChanges.proposed.collectFirst { case u: UpdateAddHtlc => u }.isDefined
 
+  def localHasUnsignedUpdateFee(commitments: Commitments): Boolean = commitments.localChanges.proposed.collectFirst { case u: UpdateFee => u }.isDefined
```
pm47, line 524: For consistency, and it's also clearer I think?

```suggestion
  def localHasUnsignedOutgoingUpdateFee(commitments: Commitments): Boolean = commitments.localChanges.proposed.collectFirst { case u: UpdateFee => u }.isDefined
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -515,6 +521,10 @@ object Commitments {
...
+
+  def remoteHasUnsignedUpdateFee(commitments: Commitments): Boolean = commitments.remoteChanges.proposed.collectFirst { case u: UpdateFee => u }.isDefined
```
pm47, line 526:
```suggestion
  def remoteHasUnsignedOutgoingUpdateFee(commitments: Commitments): Boolean = commitments.remoteChanges.proposed.collectFirst { case u: UpdateFee => u }.isDefined
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelExceptions.scala
```diff
@@ -49,8 +49,10 @@ case class ChannelReserveBelowOurDustLimit     (override val channelId: ByteVect
...
 case class ChannelFundingError                 (override val channelId: ByteVector32) extends ChannelException(channelId, "channel funding error")
 case class NoMoreHtlcsClosingInProgress        (override val channelId: ByteVector32) extends ChannelException(channelId, "cannot send new htlcs, closing in progress")
+case class NoMoreFeeUpdateClosingInProgress    (override val channelId: ByteVector32) extends ChannelException(channelId, "cannot send new update_fee, closing in progress")
 case class ClosingAlreadyInProgress            (override val channelId: ByteVector32) extends ChannelException(channelId, "closing already in progress")
 case class CannotCloseWithUnsignedOutgoingHtlcs(override val channelId: ByteVector32) extends ChannelException(channelId, "cannot close when there are unsigned outgoing htlcs")
+case class CannotCloseWithUnsignedUpdateFee    (override val channelId: ByteVector32) extends ChannelException(channelId, "cannot close when there is an unsigned fee update")
```
pm47, line 55:
```suggestion
case class CannotCloseWithUnsignedOutgoingUpdateFee    (override val channelId: ByteVector32) extends ChannelException(channelId, "cannot close when there is an unsigned fee update")
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -831,33 +833,34 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-      //        there are pending signed htlcs    => go to SHUTDOWN
+      //        there are pending signed changes  => go to SHUTDOWN
       //        there are no htlcs                => go to NEGOTIATING
-
       if (!Closing.isValidFinalScriptPubkey(remoteScriptPubKey)) {
         handleLocalError(InvalidFinalScript(d.channelId), d, Some(remoteShutdown))
       } else if (Commitments.remoteHasUnsignedOutgoingHtlcs(d.commitments)) {
         handleLocalError(CannotCloseWithUnsignedOutgoingHtlcs(d.channelId), d, Some(remoteShutdown))
+      } else if (Commitments.remoteHasUnsignedUpdateFee(d.commitments)) {
+        handleLocalError(CannotCloseWithUnsignedUpdateFee(d.channelId), d, Some(remoteShutdown))
```
pm47, line 846:
```suggestion
        handleLocalError(CannotCloseWithUnsignedOutgoingUpdateFee(d.channelId), d, Some(remoteShutdown))
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -812,8 +812,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       if (d.localShutdown.isDefined) {
         handleCommandError(ClosingAlreadyInProgress(d.channelId), c)
       } else if (Commitments.localHasUnsignedOutgoingHtlcs(d.commitments)) {
-        // TODO: simplistic behavior, we could also sign-then-close
+        // NB: simplistic behavior, we could also sign-then-close
         handleCommandError(CannotCloseWithUnsignedOutgoingHtlcs(d.channelId), c)
+      } else if (Commitments.localHasUnsignedUpdateFee(d.commitments)) {
+        handleCommandError(CannotCloseWithUnsignedUpdateFee(d.channelId), c)
```
pm47, line 818:
```suggestion
        handleCommandError(CannotCloseWithUnsignedOutgoingUpdateFee(d.channelId), c)
```

pm47 (commented):
Looking good! Just an OCD nit

t-bast:
Done, I added `Outgoing` in https://github.com/ACINQ/eclair/pull/1661/commits/44db5a54afa227e552bbcfb01355c71c560e1a8d

pm47 (approved):

PR #1662, opened 2021-01-05
Remove kamon tracing

t-bast:
It's costly, we're not using it, and it's too invasive.
There's no reason to keep it at the moment.


pm47:
Is the issue more the cost, or the invasiveness? If it's the former, then why not simply disable it?

t-bast:
It's mostly the invasiveness (the cost is an added benefit, but it's quite negligible).

pm47 (approved):

PR #1666, opened 2021-01-07
Fix HTLC fulfill race condition in integration spec

t-bast:
We were extracting F's commit tx from its internal state right after receiving the `PaymentSent` event. The issue is that this could happen before the fulfill was completely signed on both sides, so the commit tx we obtained would still contain the HTLC and would be different from the one F would publish when closing.

This is the race condition that was sometimes happening:

```
C            F
 add ------->
 <--- fulfill

 payment-sent
 F's commit tx 1

 <------- sig
 rev ------->
 sig ------->
 <------- rev

 force-close
 F's commit tx 2
```

And F's commit tx1 and commit tx2 would be different, so we were trying to get the wrong tx from bitcoind.


pm47:
Wasn't it a nice corner case to test though? (force close with only the unsigned fulfill).

t-bast:
True, but we can't test it reliably in an integration test because we can't control the timing to ensure we're exactly at that step.
But it can be tested in the channel tests, I'll add it there (if it's missing).

t-bast:
> Wasn't it a nice corner case to test though? (force close with only the unsigned fulfill).

Done in https://github.com/ACINQ/eclair/pull/1666/commits/079920bd5f62df4ace8305277882a4ac658824ac

pm47 (approved):

PR #1667, opened 2021-01-07
Blockchain watchdogs use unique actor name

### LINKED ISSUE #1665: Blockchain Watchdog : duplicate actor name
There seem to be an issue related to supervision:

```
2020-12-22 22:15:30,457 ERROR akka.actor.ActorCell - Supervisor ResumeSupervisor saw failure: actor name [mempool.space] is not unique!akka.actor.InvalidActorNameException: actor name [mempool.space] is not unique!
2020-12-23 04:47:24,627 ERROR akka.actor.ActorCell - Supervisor ResumeSupervisor saw failure: actor name [blockcypher] is not unique!akka.actor.InvalidActorNameException: actor name [blockcypher] is not unique!
```

t-bast:
Actor names cannot conflict.

Even though blockchain watchdog actors stop themselves after fetching block data, when blocks are found in a short interval, we may end up with multiple actors of the same type simultaneously alive, so we need them to have unique names.

The blockcount isn't sufficient to make their names unique because forks can happen.

Fixes #1665

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/BlockchainWatchdog.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/BlockchainWatchdog.scala
@@ -23,6 +23,7 @@ import fr.acinq.bitcoin.{BlockHeader, ByteVector32}
 import fr.acinq.eclair.blockchain.CurrentBlockCount
 import fr.acinq.eclair.blockchain.watchdogs.Monitoring.{Metrics, Tags}
 
+import java.util.UUID
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import scala.util.Random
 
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/BlockchainWatchdog.scala
```diff
@@ -66,10 +67,11 @@ object BlockchainWatchdog {
             timers.startSingleTimer(CheckLatestHeaders(blockCount), delay)
             Behaviors.same
           case CheckLatestHeaders(blockCount) =>
-            context.spawn(HeadersOverDns(chainHash, blockCount), HeadersOverDns.Source) ! HeadersOverDns.CheckLatestHeaders(context.self)
-            context.spawn(ExplorerApi(chainHash, blockCount, ExplorerApi.BlockstreamExplorer()), "blockstream") ! ExplorerApi.CheckLatestHeaders(context.self)
-            context.spawn(ExplorerApi(chainHash, blockCount, ExplorerApi.BlockcypherExplorer()), "blockcypher") ! ExplorerApi.CheckLatestHeaders(context.self)
-            context.spawn(ExplorerApi(chainHash, blockCount, ExplorerApi.MempoolSpaceExplorer()), "mempool.space") ! ExplorerApi.CheckLatestHeaders(context.self)
+            val id = UUID.randomUUID()
```
pm47, line 70: A thread-safe counter (`AtomicInteger`) could also have been used.

pm47 (approved):

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/BlockchainWatchdog.scala
t-bast: True, but it's stateful and needs locks :)

PR #1672, opened 2021-01-19
Add a maximum fee threshold for anchor outputs

t-bast:
With anchor outputs, the actual feerate for the commit tx can be decided when broadcasting the tx by using CPFP on the anchor.

That means we don't need to constantly keep the channel feerate close to what's happening on-chain. We just need a feerate that's good enough to get the tx to propagate through the bitcoin network.

We set the upper threshold to 10 sat/byte, which is what lnd does as well. We let the feerate be lower than that when possible, but do note that depending on your configured `feerate-tolerance`, that means you can still experience some force-close events because of feerate mismatch.

The last two commits fix small related issues and can be reviewed independently.

NB: anchor outputs is still not safe to use after this commit: we don't bump the feerate yet by doing a CPFP on the anchor. This will come in a later PR.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
@@ -17,6 +17,9 @@
 package fr.acinq.eclair.blockchain.fee
 
 import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.bitcoin.{Satoshi, SatoshiLong}
+import fr.acinq.eclair.blockchain.CurrentFeerates
+import fr.acinq.eclair.channel.ChannelVersion
 
 trait FeeEstimator {
   // @formatter:off
@@ -25,10 +28,51 @@ trait FeeEstimator {
   // @formatter:on
 }
 
+object FeeEstimator {
+  /** When using anchor outputs, we only need to set a feerate that allows the tx to propagate: we will use CPFP to speed up confirmation if needed. */
+  val AnchorOutputMaxCommitFeerate = FeeratePerKw(FeeratePerByte(10 sat))
+}
+
 case class FeeTargets(fundingBlockTarget: Int, commitmentBlockTarget: Int, mutualCloseBlockTarget: Int, claimMainBlockTarget: Int)
 
-case class FeerateTolerance(ratioLow: Double, ratioHigh: Double)
+case class FeerateTolerance(ratioLow: Double, ratioHigh: Double) {
+  /**
+   * @param channelVersion  channel version
+   * @param networkFeerate  reference fee rate (value we estimate from our view of the network)
+   * @param proposedFeerate fee rate proposed (new proposal through update_fee or previous proposal used in our current commit tx)
+   * @return true if the difference between proposed and reference fee rates is too high.
+   */
+  def isFeeDiffTooHigh(channelVersion: ChannelVersion, networkFeerate: FeeratePerKw, proposedFeerate: FeeratePerKw): Boolean = {
+    if (channelVersion.hasAnchorOutputs) {
+      proposedFeerate < networkFeerate * ratioLow || FeeEstimator.AnchorOutputMaxCommitFeerate * ratioHigh < proposedFeerate
+    } else {
+      proposedFeerate < networkFeerate * ratioLow || networkFeerate * ratioHigh < proposedFeerate
+    }
+  }
+}
 
 case class OnChainFeeConf(feeTargets: FeeTargets, feeEstimator: FeeEstimator, closeOnOfflineMismatch: Boolean, updateFeeMinDiffRatio: Double, private val defaultFeerateTolerance: FeerateTolerance, private val perNodeFeerateTolerance: Map[PublicKey, FeerateTolerance]) {
+
   def maxFeerateMismatchFor(nodeId: PublicKey): FeerateTolerance = perNodeFeerateTolerance.getOrElse(nodeId, defaultFeerateTolerance)
+
+  /** To avoid spamming our peers with fee updates every time there's a small variation, we only update the fee when the difference exceeds a given ratio. */
+  def shouldUpdateFee(currentFeeratePerKw: FeeratePerKw, nextFeeratePerKw: FeeratePerKw): Boolean =
+    currentFeeratePerKw.toLong == 0 || Math.abs((currentFeeratePerKw.toLong - nextFeeratePerKw.toLong).toDouble / currentFeeratePerKw.toLong) > updateFeeMinDiffRatio
+
+  /**
+   * Get the feerate that should apply to a channel commitment transaction:
+   *  - if we're using anchor outputs, we use a feerate that allows network propagation of the commit tx: we will use CPFP to speed up confirmation if needed
+   *  - otherwise we use a feerate that should get the commit tx confirmed in the configured number of blocks
+   */
+  def getCommitmentFeerate(channelVersion: ChannelVersion, channelCapacity: Satoshi, currentFeerates_opt: Option[CurrentFeerates]): FeeratePerKw = {
+    val networkFeerate = currentFeerates_opt match {
+      case Some(currentFeerates) => currentFeerates.feeratesPerKw.feePerBlock(feeTargets.commitmentBlockTarget)
+      case None => feeEstimator.getFeeratePerKw(feeTargets.commitmentBlockTarget)
+    }
+    if (channelVersion.hasAnchorOutputs) {
+      networkFeerate.min(FeeEstimator.AnchorOutputMaxCommitFeerate)
+    } else {
+      networkFeerate
+    }
+  }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
@@ -62,7 +62,10 @@ case class OnChainFeeConf(feeTargets: FeeTargets, feeEstimator: FeeEstimator, cl
   /**
    * Get the feerate that should apply to a channel commitment transaction:
    *  - if we're using anchor outputs, we use a feerate that allows network propagation of the commit tx: we will use CPFP to speed up confirmation if needed
-   *  - otherwise we use a feerate that should get the commit tx confirmed in the configured number of blocks
+   *  - otherwise we use a feerate that should get the commit tx confirmed within the configured block target
+   *
+   * @param channelVersion      channel version
+   * @param currentFeerates_opt if provided, will be used to compute the most up-to-date network fee, otherwise we rely on the fee estimator
    */
   def getCommitmentFeerate(channelVersion: ChannelVersion, channelCapacity: Satoshi, currentFeerates_opt: Option[CurrentFeerates]): FeeratePerKw = {
     val networkFeerate = currentFeerates_opt match {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -276,11 +276,12 @@ object Commitments {
 
     // we allowed mismatches between our feerates and our remote's as long as commitments didn't contain any HTLC at risk
     // we need to verify that we're not disagreeing on feerates anymore before offering new HTLCs
-    // NB: there may be a pending update_fee that hasn't been signed yet that needs to be taken into account
-    val currentFeeratePerKw = commitments.remoteChanges.proposed.collect { case f: UpdateFee => f.feeratePerKw }.lastOption.getOrElse(commitments.localCommit.spec.feeratePerKw)
+    // NB: there may be a pending update_fee that hasn't been applied yet that needs to be taken into account
     val localFeeratePerKw = feeConf.getCommitmentFeerate(commitments.channelVersion, commitments.capacity, None)
-    if (feeConf.maxFeerateMismatchFor(commitments.remoteNodeId).isFeeDiffTooHigh(commitments.channelVersion, localFeeratePerKw, currentFeeratePerKw)) {
-      return Left(FeerateTooDifferent(commitments.channelId, localFeeratePerKw = localFeeratePerKw, remoteFeeratePerKw = commitments.localCommit.spec.feeratePerKw))
+    val remoteFeeratePerKw = commitments.localCommit.spec.feeratePerKw +: commitments.remoteChanges.all.collect { case f: UpdateFee => f.feeratePerKw }
+    remoteFeeratePerKw.find(feerate => feeConf.maxFeerateMismatchFor(commitments.remoteNodeId).isFeeDiffTooHigh(commitments.channelVersion, localFeeratePerKw, feerate)) match {
+      case Some(feerate) => return Left(FeerateTooDifferent(commitments.channelId, localFeeratePerKw = localFeeratePerKw, remoteFeeratePerKw = feerate))
+      case None =>
     }
 
     // let's compute the current commitment *as seen by them* with this change taken into account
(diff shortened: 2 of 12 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
```diff
@@ -25,10 +28,51 @@ trait FeeEstimator {
...
+case class FeerateTolerance(ratioLow: Double, ratioHigh: Double) {
+  /**
+   * @param channelVersion  channel version
+   * @param networkFeerate  reference fee rate (value we estimate from our view of the network)
+   * @param proposedFeerate fee rate proposed (new proposal through update_fee or previous proposal used in our current commit tx)
+   * @return true if the difference between proposed and reference fee rates is too high.
+   */
+  def isFeeDiffTooHigh(channelVersion: ChannelVersion, networkFeerate: FeeratePerKw, proposedFeerate: FeeratePerKw): Boolean = {
+    if (channelVersion.hasAnchorOutputs) {
+      proposedFeerate < networkFeerate * ratioLow || FeeEstimator.AnchorOutputMaxCommitFeerate * ratioHigh < proposedFeerate
```
pm47, line 47: Shouldn't the `AnchorOutputMaxCommitFeerate` be an hard limit? Or do you consider that `UpdateFee` contains the actual network feerate, and we will cap it with `AnchorOutputMaxCommitFeerate`? The problem I have with the latter is that there is no such thing as a reasonable network feerate, since it all depends on CPFP.

```suggestion
      proposedFeerate < networkFeerate * ratioLow || FeeEstimator.AnchorOutputMaxCommitFeerate < proposedFeerate
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
```diff
@@ -25,10 +28,51 @@ trait FeeEstimator {
...
+  }
+}
 
 case class OnChainFeeConf(feeTargets: FeeTargets, feeEstimator: FeeEstimator, closeOnOfflineMismatch: Boolean, updateFeeMinDiffRatio: Double, private val defaultFeerateTolerance: FeerateTolerance, private val perNodeFeerateTolerance: Map[PublicKey, FeerateTolerance]) {
+
   def maxFeerateMismatchFor(nodeId: PublicKey): FeerateTolerance = perNodeFeerateTolerance.getOrElse(nodeId, defaultFeerateTolerance)
+
+  /** To avoid spamming our peers with fee updates every time there's a small variation, we only update the fee when the difference exceeds a given ratio. */
+  def shouldUpdateFee(currentFeeratePerKw: FeeratePerKw, nextFeeratePerKw: FeeratePerKw): Boolean =
+    currentFeeratePerKw.toLong == 0 || Math.abs((currentFeeratePerKw.toLong - nextFeeratePerKw.toLong).toDouble / currentFeeratePerKw.toLong) > updateFeeMinDiffRatio
```
pm47, line 60: I wonder what that `currentFeeratePerKw.toLong == 0` was for (I know it hasn't been introduced by this PR).

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
```diff
@@ -25,10 +28,51 @@ trait FeeEstimator {
...
+
+  /**
+   * Get the feerate that should apply to a channel commitment transaction:
+   *  - if we're using anchor outputs, we use a feerate that allows network propagation of the commit tx: we will use CPFP to speed up confirmation if needed
+   *  - otherwise we use a feerate that should get the commit tx confirmed in the configured number of blocks
```
pm47, line 65:
```suggestion
   *  - otherwise we use a feerate that should get the commit tx confirmed within the configured block target
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
```diff
@@ -25,10 +28,51 @@ trait FeeEstimator {
...
+   */
```
pm47, line 66:
```suggestion
   *
   *  @param channelVersion channel version
   *  @param currentFeerates_opt if provided, will be used to compute the most up-to-date network fee, otherwise we rely
   *                             on the fee estimator
   */
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
t-bast: The problem is that we don't control what value our remote will use there (when they're funder).
LND uses 10 sat/byte by default, but they may change it to 5 sat/byte or 20 sat/byte.
If we want to use 10 sat/byte for the channels where we're funder and allow some leeway for the channels where we're fundee, we need to use a ratio here.

The only other option I see is to blindly accept the remote feerate when we're fundee, do you think that's better (it's a reasonable alternative in my opinion)? I think we can only blindly accept high feerates, we still want to ensure our remote doesn't use a feerate that's so low that the tx will not even relay...

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -443,8 +427,12 @@ object Helpers {
...
     def firstClosingFee(commitments: Commitments, localScriptPubkey: ByteVector, remoteScriptPubkey: ByteVector, feeEstimator: FeeEstimator, feeTargets: FeeTargets)(implicit log: LoggingAdapter): Satoshi = {
       val requestedFeerate = feeEstimator.getFeeratePerKw(feeTargets.mutualCloseBlockTarget)
-      // we "MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction"
-      val feeratePerKw = requestedFeerate.min(commitments.localCommit.spec.feeratePerKw)
+      val feeratePerKw = if (commitments.channelVersion.hasAnchorOutputs) {
+        requestedFeerate
+      } else {
+        // we "MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction"
+        requestedFeerate.min(commitments.localCommit.spec.feeratePerKw)
+      }
```
pm47, line 435: I can't see where in the spec there is a special case for anchor outputs? It only says:

> MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction, as calculated in BOLT #3.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -276,8 +276,10 @@ object Commitments {
 
     // we allowed mismatches between our feerates and our remote's as long as commitments didn't contain any HTLC at risk
     // we need to verify that we're not disagreeing on feerates anymore before offering new HTLCs
-    val localFeeratePerKw = feeConf.feeEstimator.getFeeratePerKw(target = feeConf.feeTargets.commitmentBlockTarget)
-    if (Helpers.isFeeDiffTooHigh(localFeeratePerKw, commitments.localCommit.spec.feeratePerKw, feeConf.maxFeerateMismatchFor(commitments.remoteNodeId))) {
+    // NB: there may be a pending update_fee that hasn't been signed yet that needs to be taken into account
+    val currentFeeratePerKw = commitments.remoteChanges.proposed.collect { case f: UpdateFee => f.feeratePerKw }.lastOption.getOrElse(commitments.localCommit.spec.feeratePerKw)
```
pm47, line 280: We should use `remoteChanges.acked ++ remoteChanges.signed` instead, because we haven't yet acknowledged what is in `remoteChanges.proposed`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -337,8 +339,10 @@ object Commitments {
 
     // we allowed mismatches between our feerates and our remote's as long as commitments didn't contain any HTLC at risk
     // we need to verify that we're not disagreeing on feerates anymore before accepting new HTLCs
-    val localFeeratePerKw = feeConf.feeEstimator.getFeeratePerKw(target = feeConf.feeTargets.commitmentBlockTarget)
-    if (Helpers.isFeeDiffTooHigh(localFeeratePerKw, commitments.localCommit.spec.feeratePerKw, feeConf.maxFeerateMismatchFor(commitments.remoteNodeId))) {
+    // NB: there may be a pending update_fee that hasn't been signed yet that needs to be taken into account
+    val currentFeeratePerKw = commitments.remoteChanges.proposed.collect { case f: UpdateFee => f.feeratePerKw }.lastOption.getOrElse(commitments.localCommit.spec.feeratePerKw)
```
pm47, line 343: Same comment as above.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: It's really an implementation choice, not a spec thing. This is because we enforce a low feerate on the commit tx (decorrelated from on-chain fees) so applying that `min` doesn't make any sense.

If we force the commit tx feerate to be 10 sat/byte but the on-chain feerate for our `mutualCloseBlockTarget` is higher, we should use that higher value instead of using 10 sat/byte, otherwise the mutual close will never confirm in time. Is that clearer? I'll add a comment in the code.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
t-bast: IIRC it's to ensure we never, ever divide by zero (in the next condition) to avoid destroying the universe?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: I get your point, but maybe the spec needs to be updated then, because technically we are not compliant here?

t-bast: Maybe, the spec could be amended with a `SHOULD` instead of a `MUST`, I'll open a PR.

t-bast: Done in https://github.com/lightningnetwork/lightning-rfc/pull/847

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1672/commits/11eaa284c4299878b6eb2f2437037aad6bc0636a

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1672/commits/11eaa284c4299878b6eb2f2437037aad6bc0636a

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
pm47: > The problem is that we don't control what value our remote will use there (when they're funder).
LND uses 10 sat/byte by default, but they may change it to 5 sat/byte or 20 sat/byte.
If we want to use 10 sat/byte for the channels where we're funder and allow some leeway for the channels where we're fundee, we need to use a ratio here.

Nevermind, I thought the `AnchorOutputMaxCommitFeerate` value was defined in spec. With that in mind, we either set a tolerance, or remove the check alltogether as you propose below.

> The only other option I see is to blindly accept the remote feerate when we're fundee, do you think that's better (it's a reasonable alternative in my opinion)? I think we can only blindly accept high feerates, we still want to ensure our remote doesn't use a feerate that's so low that the tx will not even relay (at least until we get package relay)...but even that may not be correctly enforceable, or we may be too late when we realize that the fee isn't enough :/

True, since the funder pays, I don't see why we wouldn't let him pay as much as he wants either. Not sure why we need this notion of a higher limit on the fees.

t-bast: > Not sure why we need this notion of a higher limit on the fees.

I think if we use a tolerance, it still makes sense to have a higher bound as well, because by raising the fee the remote peer can make all HTLC outputs become dust / non-enforceable on-chain, which can be undesirable.

t-bast: Let's stick with this tolerance ratio for now and re-visit later once we're doing interop tests with other implementations and see how they behave?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: I tried moving that check in `sendCommit`, but it's not working well, because a failure in `sendCommit` only fails the command but doesn't close the channel. If that happens the channel will just be kinda stuck, but not clearly closed.

So I decided in https://github.com/ACINQ/eclair/pull/1672/commits/0602beefd487eda7148a7741d43aeae95cf0bbbd to be quite aggressive and check all pending `update_fee` from our remote. Maybe there are some we could ignore because they would be overridden by later ones, but I don't want to take any risk by adding complexity here.

The advantage of checking this in `sendAdd` / `receiveAdd` is that in `sendAdd`, we simply reject the HTLC (without closing the channel, we wait for the channel to be back at a feerate we agree with before using it) and only in `receiveAdd` we close the channel. That sounds like a good compromise, let me know what you think.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
pm47: Fair enough

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: I think a better idea would be to use `CommitmentSpec.reduce` exactly like in the `receiveFee` method, it would take care of deciding which fee to take into account and be consistent with the rest.

t-bast: I'm afraid that wouldn't be enough, there are still issues when we add an HTLC after letting some `update_fee`s we disagree with slip into the `remoteChanges`.

Imagine there is no pending HTLC but there is an `update_fee` in `remoteChanges.proposed` and one in `remoteChanges.acked`, and we're disagree only with the one in `remoteChanges.proposed`. Then the remote adds an HTLC.

If we only check against the `update_fee` in `remoteChanges.signed`, the `update_fee` in `remoteChanges.proposed` will get signed at the same time as the HTLC and we end up with a signed commitment, with a pending HTLC and a feerate we disagree with.

If on the other hand we agree with the `update_fee` in `remoteChanges.proposed` but disagree with the one in `remoteChanges.signed`, but this time we add an HTLC (not the remote). They can delay the ack/sign process on the `update_fee` in `remoteChanges.proposed` and we can end up with a commitment with our HTLC and the feerate we disagree with.

It's really hard to explore all the potential cases since the protocol is asynchronous, and I'm really not feeling safe leaving a hole there, so I'd rather check all pending `update_fee`s and close if we disagree with any of them, this way we're sure our HTLCs are safe. Does that make sense?

pm47: I agree, my proposal doesn't work.

After reading your comment I immediately went to check if we weren't exposed to the opposite scenario, where we first send an `update_add_htlc`, and then received an `update_fee` that we disagree with. But we _do_ check all (unsigned and signed, incoming and outgoing) htlcs in `hasPendingOrProposedHtlcs` so we are okay. But it doesn't make me feel safe either.

This makes me go back to the original idea of enforcing a check in `sendCommit`, which is really the only way to be sure. Going back to your comment https://github.com/ACINQ/eclair/pull/1672#discussion_r577466335:

> I tried moving that check in sendCommit, but it's not working well, because a failure in sendCommit only fails the command but doesn't close the channel. If that happens the channel will just be kinda stuck, but not clearly closed.

We could define a `ChannelException` that makes us force close the channel.

> The advantage of checking this in sendAdd / receiveAdd is that in sendAdd, we simply reject the HTLC (without closing the channel, we wait for the channel to be back at a feerate we agree with before using it) and only in receiveAdd we close the channel. That sounds like a good compromise, let me know what you think.

That is a really good point. Sigh... But maybe we should do both? Our actual security check would be in `sendCommit`, and we would have a best-effort check in `sendAdd` to prevent unnecessary force closes. What do you think?

t-bast: > We could define a ChannelException that makes us force close the channel.

We could indeed, but that breaks the symmetry between commands and external messages and feels a bit hackish...

> But maybe we should do both? Our actual security check would be in sendCommit, and we would have a best-effort check in sendAdd to prevent unnecessary force closes. What do you think?

I think that if we have this check in `sendAdd`, we should have it in `receiveAdd` for consistency (when we read the code later, we'll wonder why it's not in both, even with a comment explaining it since this is quite subtle). And in that case the check in `sendCommit` is unnecessary! What do you dislike about checking `proposed + acked + signed` here? It's quite concise and we're sure it does the job, doesn't it?

pm47: > It's quite concise and we're sure it does the job, doesn't it?

That's the thing. I _think_ it works, but I'm not 100% sure it's enough. Like you said, it's really hard to explore all potential cases.

Anyway, there are a lot of other things that are even more sensitive than fee rate, that we don't double check in the `sendCommit` (e.g. expiries), so it wouldn't really make sense to do a special case here. Let's forget about it.

pm47 (approved):

PR #1673, opened 2021-01-20
Update relay fees from eclair.conf

t-bast:
If the `eclair.conf` changes while we're offline, we want to apply the fee changes to existing channels once they come back online.


pm47:
This could be a problem for people who fine-tune their relay fees channel per channel, because all the fees are overriden at restart. I believe we should keep the behavior as it is currently (maybe adding a comment to explain why we keep the existing fees), and tell users to use the API if they want to mass-update their relay fees settings.

pm47 (approved):

PR #1674, opened 2021-01-22
Add esplora feerates provider

btcontract:


eclair-core/src/test/scala/fr/acinq/eclair/blockchain/fee/EsploraFeeProviderSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+    val json = parse(sample_response)
+    val feeRanges = parseFeeRanges(json)
+    assert(feeRanges.size === 28)
+  }
+
+  test("extract fee for a particular block delay") {
+    val json = parse(sample_response)
+    val feeRanges = parseFeeRanges(json)
+    val fee = extractFeerate(feeRanges, 6)
+    assert(fee === FeeratePerKB(368000 sat))
```
t-bast, line 53: The tests should more explicitly show that what's returned uses `FeeratePerKw`, here we don't understand where your magical values come from since you've done the conversion to `FeeratePerKB` outside of the tests.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/EsploraFeeProvider.scala
```diff
@@ -0,0 +1,57 @@
+package fr.acinq.eclair.blockchain.fee
```
t-bast, line 1: nit: don't forget the license/copyright header (same in the spec file)

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/EsploraFeeProvider.scala
```diff
@@ -0,0 +1,57 @@
...
+  def extractFeerate(feeRanges: Seq[BlockTarget], maxBlockDelay: Int): FeeratePerKB = {
+    // first we keep only fee ranges with a max block delay below the limit
+    val belowLimit = feeRanges.filter(_.block <= maxBlockDelay)
+    // out of all the remaining fee ranges, we select the one with the minimum higher bound
+    FeeratePerKB(FeeratePerKw(FeeratePerVByte(Satoshi(belowLimit.map(_.satPerVByte).min))))
+  }
+
+  def extractFeerates(feeRanges: Seq[BlockTarget]): FeeratesPerKB =
+    FeeratesPerKB(
+      mempoolMinFee = extractFeerate(feeRanges, 1008),
```
t-bast, line 48: It would be nice if esplora had a way of returning the mempool-min-fee, are you sure this doesn't exist?

t-bast:
Thinking about this more, I'm really not sure this should be added at all.
On the contrary, we're in the process of removing feerate providers that are not a local bitcoin core.
And with anchor outputs, you won't need to rely on feerate providers as much as today, so why add more code around something that's being deprecated?

btcontract:
If the plan is to remove these from node code (and keep it in mobile-specific code only I assume) then indeed there's no reason.

t-bast:
TBH I don't think it makes sense to add it at this point.

Maybe (much) later when (if) we plan on supporting a complete "light" backend without a bitcoin core node, but that's a lot of work in various areas so probably not going to happen anytime soon.

PR #1677, opened 2021-01-27
Relax single tx input requirements

t-bast:
In some places of the codebase we relied on the fact that lightning transactions had a single input. That was correct with the standard commitments format, but will not be the case with anchor outputs: 2nd-stage txs (htlc-txs) and 3rd-stage txs (claim-htlc-txs) can be RBF-ed and have any number of inputs and outputs.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
@@ -40,10 +40,13 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 
 /**
- * A blockchain watcher that:
- * - receives bitcoin events (new blocks and new txes) directly from the bitcoin network
- * - also uses bitcoin-core rpc api, most notably for tx confirmation count and blockcount (because reorgs)
  * Created by PM on 21/02/2016.
+ */
+
+/**
+ * A blockchain watcher that:
+ *  - receives bitcoin events (new blocks and new txs) directly from the bitcoin network
+ *  - also uses bitcoin-core rpc api, most notably for tx confirmation count and block count (because reorgs)
  */
 class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: ExtendedBitcoinClient)(implicit ec: ExecutionContext = ExecutionContext.global) extends Actor with ActorLogging {
 
@@ -180,13 +183,15 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
     case PublishAsap(tx) =>
       val blockCount = this.blockCount.get()
       val cltvTimeout = Scripts.cltvTimeout(tx)
-      val csvTimeout = Scripts.csvTimeout(tx)
-      if (csvTimeout > 0) {
-        require(tx.txIn.size == 1, s"watcher only supports tx with 1 input, this tx has ${tx.txIn.size} inputs")
-        val parentTxid = tx.txIn.head.outPoint.txid
-        log.info(s"txid=${tx.txid} has a relative timeout of $csvTimeout blocks, watching parenttxid=$parentTxid tx={}", tx)
-        val parentPublicKey = fr.acinq.bitcoin.Script.write(fr.acinq.bitcoin.Script.pay2wsh(tx.txIn.head.witness.stack.last))
-        self ! WatchConfirmed(self, parentTxid, parentPublicKey, minDepth = 1, BITCOIN_PARENT_TX_CONFIRMED(tx))
+      val csvTimeouts = Scripts.csvTimeouts(tx)
+      if (csvTimeouts.nonEmpty) {
+        // watcher supports txs with multiple csv-delayed inputs: we watch all delayed parents and try to publish every
+        // time a parent's relative delays are satisfied, so we will eventually succeed.
+        csvTimeouts.foreach { case (parentTxId, csvTimeout) =>
+          log.info(s"txid=${tx.txid} has a relative timeout of $csvTimeout blocks, watching parentTxId=$parentTxId tx={}", tx)
+          val parentPublicKeyScript = Script.write(Script.pay2wsh(tx.txIn.find(_.outPoint.txid == parentTxId).get.witness.stack.last))
+          self ! WatchConfirmed(self, parentTxId, parentPublicKeyScript, minDepth = csvTimeout, BITCOIN_PARENT_TX_CONFIRMED(tx))
+        }
       } else if (cltvTimeout > blockCount) {
         log.info(s"delaying publication of txid=${tx.txid} until block=$cltvTimeout (curblock=$blockCount)")
         val block2tx1 = block2tx.updated(cltvTimeout, block2tx.getOrElse(cltvTimeout, Seq.empty[Transaction]) :+ tx)
@@ -197,11 +202,9 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
       log.info(s"parent tx of txid=${tx.txid} has been confirmed")
       val blockCount = this.blockCount.get()
       val cltvTimeout = Scripts.cltvTimeout(tx)
-      val csvTimeout = Scripts.csvTimeout(tx)
-      val absTimeout = math.max(blockHeight + csvTimeout, cltvTimeout)
-      if (absTimeout > blockCount) {
-        log.info(s"delaying publication of txid=${tx.txid} until block=$absTimeout (curblock=$blockCount)")
-        val block2tx1 = block2tx.updated(absTimeout, block2tx.getOrElse(absTimeout, Seq.empty[Transaction]) :+ tx)
+      if (cltvTimeout > blockCount) {
+        log.info(s"delaying publication of txid=${tx.txid} until block=$cltvTimeout (curblock=$blockCount)")
+        val block2tx1 = block2tx.updated(cltvTimeout, block2tx.getOrElse(cltvTimeout, Seq.empty[Transaction]) :+ tx)
         context become watching(watches, watchedUtxos, block2tx1, nextTick)
       } else publish(tx)
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
@@ -173,13 +173,15 @@ class ElectrumWatcher(blockCount: AtomicLong, client: ActorRef) extends Actor wi
     case PublishAsap(tx) =>
       val blockCount = this.blockCount.get()
       val cltvTimeout = Scripts.cltvTimeout(tx)
-      val csvTimeout = Scripts.csvTimeout(tx)
-      if (csvTimeout > 0) {
-        require(tx.txIn.size == 1, s"watcher only supports tx with 1 input, this tx has ${tx.txIn.size} inputs")
-        val parentTxid = tx.txIn.head.outPoint.txid
-        log.info(s"txid=${tx.txid} has a relative timeout of $csvTimeout blocks, watching parenttxid=$parentTxid tx={}", tx)
-        val parentPublicKeyScript = WatchConfirmed.extractPublicKeyScript(tx.txIn.head.witness)
-        self ! WatchConfirmed(self, parentTxid, parentPublicKeyScript, minDepth = 1, BITCOIN_PARENT_TX_CONFIRMED(tx))
+      val csvTimeouts = Scripts.csvTimeouts(tx)
+      if (csvTimeouts.nonEmpty) {
+        // watcher supports txs with multiple csv-delayed inputs: we watch all delayed parents and try to publish every
+        // time a parent's relative delays are satisfied, so we will eventually succeed.
+        csvTimeouts.foreach { case (parentTxId, csvTimeout) =>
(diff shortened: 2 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -443,8 +427,12 @@ object Helpers {
...
     def firstClosingFee(commitments: Commitments, localScriptPubkey: ByteVector, remoteScriptPubkey: ByteVector, feeEstimator: FeeEstimator, feeTargets: FeeTargets)(implicit log: LoggingAdapter): Satoshi = {
       val requestedFeerate = feeEstimator.getFeeratePerKw(feeTargets.mutualCloseBlockTarget)
-      // we "MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction"
-      val feeratePerKw = requestedFeerate.min(commitments.localCommit.spec.feeratePerKw)
+      val feeratePerKw = if (commitments.channelVersion.hasAnchorOutputs) {
+        requestedFeerate
+      } else {
+        // we "MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction"
+        requestedFeerate.min(commitments.localCommit.spec.feeratePerKw)
+      }
```
pm47, line 435: I can't see where in the spec there is a special case for anchor outputs? It only says:

> MUST set fee_satoshis less than or equal to the base fee of the final commitment transaction, as calculated in BOLT #3.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -276,8 +276,10 @@ object Commitments {
 
     // we allowed mismatches between our feerates and our remote's as long as commitments didn't contain any HTLC at risk
     // we need to verify that we're not disagreeing on feerates anymore before offering new HTLCs
-    val localFeeratePerKw = feeConf.feeEstimator.getFeeratePerKw(target = feeConf.feeTargets.commitmentBlockTarget)
-    if (Helpers.isFeeDiffTooHigh(localFeeratePerKw, commitments.localCommit.spec.feeratePerKw, feeConf.maxFeerateMismatchFor(commitments.remoteNodeId))) {
+    // NB: there may be a pending update_fee that hasn't been signed yet that needs to be taken into account
+    val currentFeeratePerKw = commitments.remoteChanges.proposed.collect { case f: UpdateFee => f.feeratePerKw }.lastOption.getOrElse(commitments.localCommit.spec.feeratePerKw)
```
pm47, line 280: We should use `remoteChanges.acked ++ remoteChanges.signed` instead, because we haven't yet acknowledged what is in `remoteChanges.proposed`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -337,8 +339,10 @@ object Commitments {
 
     // we allowed mismatches between our feerates and our remote's as long as commitments didn't contain any HTLC at risk
     // we need to verify that we're not disagreeing on feerates anymore before accepting new HTLCs
-    val localFeeratePerKw = feeConf.feeEstimator.getFeeratePerKw(target = feeConf.feeTargets.commitmentBlockTarget)
-    if (Helpers.isFeeDiffTooHigh(localFeeratePerKw, commitments.localCommit.spec.feeratePerKw, feeConf.maxFeerateMismatchFor(commitments.remoteNodeId))) {
+    // NB: there may be a pending update_fee that hasn't been signed yet that needs to be taken into account
+    val currentFeeratePerKw = commitments.remoteChanges.proposed.collect { case f: UpdateFee => f.feeratePerKw }.lastOption.getOrElse(commitments.localCommit.spec.feeratePerKw)
```
pm47, line 343: Same comment as above.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -180,24 +183,26 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
...
         context become watching(watches, watchedUtxos, block2tx1, nextTick)
       } else publish(tx)
 
-    case WatchEventConfirmed(BITCOIN_PARENT_TX_CONFIRMED(tx), blockHeight, _, _) =>
+    case WatchEventConfirmed(BITCOIN_PARENT_TX_CONFIRMED(tx), blockHeight, _, parentTx) =>
       log.info(s"parent tx of txid=${tx.txid} has been confirmed")
       val blockCount = this.blockCount.get()
       val cltvTimeout = Scripts.cltvTimeout(tx)
-      val csvTimeout = Scripts.csvTimeout(tx)
+      val csvTimeout = Scripts.csvTimeouts(tx).getOrElse(parentTx.txid, 0L)
```
pm47, line 205: How about putting the `csvTimeout` in the watch event?
```scala
case class BITCOIN_PARENT_TX_CONFIRMED(childTx: Transaction, csvTimeout: Long) extends BitcoinEvent
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
```diff
@@ -173,13 +173,15 @@ class ElectrumWatcher(blockCount: AtomicLong, client: ActorRef) extends Actor wi
...
-        log.info(s"txid=${tx.txid} has a relative timeout of $csvTimeout blocks, watching parenttxid=$parentTxid tx={}", tx)
-        val parentPublicKeyScript = WatchConfirmed.extractPublicKeyScript(tx.txIn.head.witness)
-        self ! WatchConfirmed(self, parentTxid, parentPublicKeyScript, minDepth = 1, BITCOIN_PARENT_TX_CONFIRMED(tx))
+      val csvTimeouts = Scripts.csvTimeouts(tx)
+      if (csvTimeouts.nonEmpty) {
+        // watcher supports txs with multiple csv-delayed inputs: we watch all delayed parents and try to publish every
+        // time a parent's relative delays are satisfied, so we will eventually succeed.
+        csvTimeouts.foreach { case (parentTxId, csvTimeout) =>
+          log.info(s"txid=${tx.txid} has a relative timeout of $csvTimeout blocks, watching parentTxId=$parentTxId tx={}", tx)
+          val parentPublicKeyScript = WatchConfirmed.extractPublicKeyScript(tx.txIn.find(_.outPoint.txid == parentTxId).get.witness)
```
pm47, line 182: Same comment as in `ZmqWatcher`.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/electrum/ElectrumWatcher.scala
```diff
@@ -189,11 +191,11 @@ class ElectrumWatcher(blockCount: AtomicLong, client: ActorRef) extends Actor wi
         context become running(height, tip, watches, scriptHashStatus, block2tx, sent :+ tx)
       }
 
-    case WatchEventConfirmed(BITCOIN_PARENT_TX_CONFIRMED(tx), blockHeight, _, _) =>
+    case WatchEventConfirmed(BITCOIN_PARENT_TX_CONFIRMED(tx), blockHeight, _, parentTx) =>
       log.info(s"parent tx of txid=${tx.txid} has been confirmed")
       val blockCount = this.blockCount.get()
       val cltvTimeout = Scripts.cltvTimeout(tx)
-      val csvTimeout = Scripts.csvTimeout(tx)
+      val csvTimeout = Scripts.csvTimeouts(tx).getOrElse(parentTx.txid, 0L)
```
pm47, line 198: Same comment as in `ZmqWatcher`.

pm47 (commented):
This works buy may generate warnings in logs. Another strategy would have been to handle delayed input one-by-one, and keep track of the remaining inputs in the `BITCOIN_PARENT_TX_CONFIRMED` event.
 
Something like:

```scala
case class BITCOIN_PARENT_TX_CONFIRMED(childTx: Transaction, nextCsvTimeouts: Map[ByteVector32, Long]) extends BitcoinEvent
```

t-bast:
> This works but may generate warnings in logs. Another strategy would have been to handle delayed input one-by-one, and keep track of the remaining inputs in the BITCOIN_PARENT_TX_CONFIRMED event.

TBH this code is very temporary; I'll completely remove and change how we track and publish txs to support regularly bumping tx fees for anchor outputs, so I'd rather go with the simplest approach / the less code changes for now.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: I'll try that, that makes sense.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -180,24 +183,26 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
...
-        val parentPublicKey = fr.acinq.bitcoin.Script.write(fr.acinq.bitcoin.Script.pay2wsh(tx.txIn.head.witness.stack.last))
-        self ! WatchConfirmed(self, parentTxid, parentPublicKey, minDepth = 1, BITCOIN_PARENT_TX_CONFIRMED(tx))
+      val csvTimeouts = Scripts.csvTimeouts(tx)
+      if (csvTimeouts.nonEmpty) {
+        // watcher supports txs with multiple csv-delayed inputs: we watch all delayed parents and try to publish every
+        // time a parent's relative delays are satisfied, so we will eventually succeed.
+        csvTimeouts.foreach { case (parentTxId, csvTimeout) =>
+          log.info(s"txid=${tx.txid} has a relative timeout of $csvTimeout blocks, watching parentTxId=$parentTxId tx={}", tx)
+          val parentPublicKeyScript = Script.write(Script.pay2wsh(tx.txIn.find(_.outPoint.txid == parentTxId).get.witness.stack.last))
+          self ! WatchConfirmed(self, parentTxId, parentPublicKeyScript, minDepth = 1, BITCOIN_PARENT_TX_CONFIRMED(tx))
```
sstone, line 193: Could we set `minDepth` to `csvTimeout` instead ? (in fact I wonder why we didn't do that before)

t-bast: That's exactly one of the improvements I have in the pipe for a later PR, I wondered the same thing myself ;)
But I'd rather do that in a second time to minimize changes for now, since this will likely be 100% changed soon.

t-bast: Actually it's simple enough to do right now, I'll update that.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: I think this isn't needed anymore, since I now use `minDepth = csvTimeout` in https://github.com/ACINQ/eclair/pull/1677/commits/7c7606b62b4d2149299cef777721e421d5cd2025

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1677/commits/7c7606b62b4d2149299cef777721e421d5cd2025

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -180,24 +183,26 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
...
-        log.info(s"txid=${tx.txid} has a relative timeout of $csvTimeout blocks, watching parenttxid=$parentTxid tx={}", tx)
-        val parentPublicKey = fr.acinq.bitcoin.Script.write(fr.acinq.bitcoin.Script.pay2wsh(tx.txIn.head.witness.stack.last))
-        self ! WatchConfirmed(self, parentTxid, parentPublicKey, minDepth = 1, BITCOIN_PARENT_TX_CONFIRMED(tx))
+      val csvTimeouts = Scripts.csvTimeouts(tx)
+      if (csvTimeouts.nonEmpty) {
+        // watcher supports txs with multiple csv-delayed inputs: we watch all delayed parents and try to publish every
+        // time a parent's relative delays are satisfied, so we will eventually succeed.
+        csvTimeouts.foreach { case (parentTxId, csvTimeout) =>
+          log.info(s"txid=${tx.txid} has a relative timeout of $csvTimeout blocks, watching parentTxId=$parentTxId tx={}", tx)
+          val parentPublicKeyScript = Script.write(Script.pay2wsh(tx.txIn.find(_.outPoint.txid == parentTxId).get.witness.stack.last))
```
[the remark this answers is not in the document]
t-bast: I think it would require the `csvTimeouts` method to also do some work around this, which isn't really linked to CSV...
To be honest `txIn` will always have only a few elements, this is quite inexpensive to iterate over it again.

pm47 (approved):

sstone (approved):

PR #1680, opened 2021-02-02
CIB-99 make possible to create invoice without preimage

prushin:


prushin:
Sorry, did this PR by mistake...

PR #1682, opened 2021-02-03
Refactor channel test helpers

t-bast:
It's useful to separate these test methods in a dedicated trait instead of always bundling it with `FixtureTestSuite`.

In particular, it was previously impossible to use both `BitcoindService` and `StateTestsHelperMethods` because `BitcoindService` doesn't work with fixtures (it leverages `beforeAll` and `afterAll` instead because launching one bitcoind instance per-test would be too expensive and useless).

My goal is to then use this new `StateTestsHelperMethods` in the `ZmqWatcherSpec` (and other related actors) to test the CPFP and RBF logic against real commitments.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
@@ -49,6 +49,17 @@ trait StateTestsBase extends StateTestsHelperMethods with FixtureTestSuite with 
 
 }
 
+object StateTestsTags {
+  /** If set, channels will use option_static_remotekey. */
+  val StaticRemoteKey = "static_remotekey"
+  /** If set, channels will use option_anchor_outputs. */
+  val AnchorOutputs = "anchor_outputs"
+  /** If set, channels will be public (otherwise we don't announce them by default). */
+  val ChannelsPublic = "channels_public"
+  /** If set, no amount will be pushed when opening a channel (by default we push a small amount). */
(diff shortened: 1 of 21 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -39,7 +38,29 @@ import scala.concurrent.duration._
...
+object StateTestsTags {
+  /** If set, channels will use option_static_remotekey. */
+  val StaticRemoteKey = "static_remotekey"
+  /** If set, channels will use option_anchor_outputs. */
+  val AnchorOutputs = "anchor_outputs"
+  /** If set, channels will be public (otherwise we don't announce them by default). */
+  val ChannelsPublic = "channels_public"
+  /** If set, no amount will be pushed when opening a channel (by default we push a small amount). */
+  val NoPushMsat = "no_push_msat"
+}
```
pm47, line 61: Is there a rationale behind leaving some tags as string? E.g. "funding_unconfirmed", "wumbo", ...

t-bast: Yes, I set here only the tags that are used in this file, because they can be used in several different files across the codebase. The other tags you mention are usually only used in their own test suite, so they don't need to be shared.

The rationale is that if the tag is used in a helper function that is called from many test suites, it's probably worth extracting the tag into a constant.

pm47 (approved):

PR #1683, opened 2021-02-04
Remove support for initial_routing_sync

t-bast:
We keep the `GetRoutingState` API available in the router as it's useful to query network information locally (or between actors), but we stop sending that data to remote nodes.

I've checked the last week on our node, and there were no queries using initial_routing_sync, so it feels ok to drop support.

### DIFF
--- a/docs/Configure.md
+++ b/docs/Configure.md
@@ -61,7 +61,7 @@ Note that you can also disable some default features:
 
 ```conf
 eclair.features {
-    initial_routing_sync = disabled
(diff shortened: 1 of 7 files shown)


pm47 (approved):

pm47:
cc @roasbeef @rustyrussell

Roasbeef:
Cool, we don't actively use it on the `lnd` end, and have a flag that lets users ignore the message all together.

Issue #1393, opened 2020-04-27
Re-enable strict compiler parameters


pm47:
They were relaxed in https://github.com/ACINQ/eclair/pull/1390 for better compatibility with the android branch.

pm47:
Starting from scala 2.13.2 we can re-enable fatal warnings while ignoring deprecation errors: https://alexn.org/blog/2020/05/26/scala-fatal-warnings.html.

Remember that we are stuck in 2.13.1 for now because of #1394.

t-bast:
Fixed by #1885

PR #1684, opened 2021-02-06
Fix/typos

ariskk:
Hi!
I am dipping my toes in Lightning and Eclair. While I was reading the docs, IntelliJ was suggesting a few typo fixes and I thought I should contribute them.

If that's ok, I may next fix collection deprecation warnings as I am reading and understanding the code.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -22,7 +22,7 @@ Notes:
 
 - This command will build all modules (core, node, gui).
 - If the build fails, you may need to clean previously built artifacts with the `mvn clean` command.
-- Some of the tests use [Docker](https://www.docker.com/), so make sure your docker daemon is running.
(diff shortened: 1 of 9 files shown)


t-bast (commented):
LGTM, thanks.

>  I may next fix collection deprecation warnings

Please don't fix them yet. We're waiting a bit before we do, because we need to keep the master branch merge-able with the `android` branch that is on an older Java version (because it's used for [eclair-mobile](https://github.com/acinq/eclair-mobile)). We will fix all these warnings once we can safely move mobile apps to a different lightning backend (based on Kotlin). There are some additional details in https://github.com/ACINQ/eclair/issues/1393

t-bast (approved):

PR #1687, opened 2021-02-12
Publish txs with min-relay-fee met

### LINKED ISSUE #1686: Check mempool min-relay-fee for each funding tx
We currently update our internal `mempool-min-relay-fee` once per block, but it actually changes whenever the mempool changes, so it may be obsolete when we try to publish a funding tx.

In `BitcoinCoreWallet`, in `fundTransaction`, we should start by querying `getmempoolinfo` and raise the feerate if necessary.

t-bast:
When our mempool is full, its min-relay-fee may be constantly changing.
To ensure our txs can be published, we need to check the min-relay-fee when we fund the transaction, and raise it if necessary.

Fixes #1686

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
@@ -42,12 +42,20 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
   def fundTransaction(tx: Transaction, lockUnspents: Boolean, feeRatePerKw: FeeratePerKw): Future[FundTransactionResponse] = fundTransaction(Transaction.write(tx).toHex, lockUnspents, feeRatePerKw)
 
   private def fundTransaction(hex: String, lockUnspents: Boolean, feeRatePerKw: FeeratePerKw): Future[FundTransactionResponse] = {
-    val feeRatePerKB = BigDecimal(FeeratePerKB(feeRatePerKw).toLong)
-    rpcClient.invoke("fundrawtransaction", hex, Options(lockUnspents, feeRatePerKB.bigDecimal.scaleByPowerOfTen(-8))).map(json => {
-      val JString(hex) = json \ "hex"
-      val JInt(changepos) = json \ "changepos"
-      val JDecimal(fee) = json \ "fee"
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreWallet.scala
```diff
@@ -42,12 +42,20 @@ class BitcoinCoreWallet(rpcClient: BitcoinJsonRPCClient)(implicit ec: ExecutionC
...
-      FundTransactionResponse(Transaction.read(hex), changepos.intValue, toSatoshi(fee))
+    val requestedFeeRatePerKB = FeeratePerKB(feeRatePerKw)
+    rpcClient.invoke("getmempoolinfo").map(json => json \ "mempoolminfee" match {
+      case JDecimal(feerate) => FeeratePerKB(Btc(feerate).toSatoshi).max(requestedFeeRatePerKB)
+      case JInt(feerate) => FeeratePerKB(Btc(feerate.toLong).toSatoshi).max(requestedFeeRatePerKB)
+      case other =>
+        logger.warn(s"cannot retrieve mempool minimum fee: $other")
+        requestedFeeRatePerKB
+    }).flatMap(feeRatePerKB => {
+      rpcClient.invoke("fundrawtransaction", hex, Options(lockUnspents, BigDecimal(feeRatePerKB.toLong).bigDecimal.scaleByPowerOfTen(-8))).map(json => {
```
pm47, line 53: Nit (there is also an unused import)
```suggestion
      rpcClient.invoke("fundrawtransaction", hex, Options(lockUnspents, feeRatePerKB.toLong.bigDecimal.scaleByPowerOfTen(-8))).map(json => {
```

t-bast: That doesn't compile... :thinking:

pm47: Requires `import scala.math.BigDecimal.long2bigDecimal`.

t-bast: Good catch, my currently broken IntelliJ IDEA didn't suggest that :)
Done in https://github.com/ACINQ/eclair/pull/1687/commits/ec55eff5058166a65afb4760e7fbe152d76625be

pm47 (approved):

PR #1689, opened 2021-02-12
Refactor latest remote commit

btcontract:
A small change to replace a rather heavy expression with a val whenever possible.


t-bast:
This looks reasonable, I can't remember why but I thought there was a reason we wanted to be explicit about `nextRemoteCommitInfo` but can remember why...I need to spend a bit more time on this.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -93,7 +93,7 @@ case class Commitments(channelVersion: ChannelVersion,
 
     localCommit.spec.htlcs.collect(outgoing).filter(expired) ++
       remoteCommit.spec.htlcs.collect(incoming).filter(expired) ++
-      remoteNextCommitInfo.left.toSeq.flatMap(_.nextRemoteCommit.spec.htlcs.collect(incoming).filter(expired).toSet)
+      latestRemoteCommit.spec.htlcs.collect(incoming).filter(expired)
```
t-bast, line 96: This one shouldn't change, it becomes confusing in the case where `remoteNextCommitInfo` is a `Right`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -963,9 +963,8 @@ object Helpers {
       val nextRemoteCommit_opt = d.commitments.remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit)
       if (localCommit.publishableTxs.commitTx.tx.txid == tx.txid) {
         // our commit got confirmed, so any htlc that we signed but they didn't sign will never reach the chain
-        val mostRecentRemoteCommit = nextRemoteCommit_opt.getOrElse(remoteCommit)
         // NB: from the p.o.v of remote, their incoming htlcs are our outgoing htlcs
-        mostRecentRemoteCommit.spec.htlcs.collect(incoming) -- localCommit.spec.htlcs.collect(outgoing)
+        d.commitments.latestRemoteCommit.spec.htlcs.collect(incoming) -- localCommit.spec.htlcs.collect(outgoing)
```
t-bast, line 967: Same here, it's confusing to change that since we clearly distinguish the `remoteNextCommit` case for business logic below. I think it should be reverted.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
```diff
@@ -389,7 +391,8 @@ object PaymentPacketSpec {
     val params = LocalParams(null, null, null, null, null, null, null, 0, isFunder = true, null, None, null)
     val remoteParams = RemoteParams(randomKey.publicKey, null, null, null, null, null, maxAcceptedHtlcs = 0, null, null, null, null, null, null)
     val commitInput = InputInfo(OutPoint(randomBytes32, 1), TxOut(testCapacity, Nil), Nil)
-    new Commitments(ChannelVersion.STANDARD, params, remoteParams, 0.toByte, null, null, null, null, 0, 0, Map.empty, null, commitInput, null, channelId) {
+    val remoteCommit = RemoteCommit(0, CommitmentSpec(Set.empty, FeeratePerKw.MinimumFeeratePerKw, toRemote = 100L.msat, toLocal = 100L.msat), randomBytes32, randomKey.publicKey)
```
t-bast, line 394:
```suggestion
    val remoteCommit = RemoteCommit(0, CommitmentSpec(Set.empty, FeeratePerKw.MinimumFeeratePerKw, toRemote = testAvailableBalanceForSend, toLocal = testAvailableBalanceForReceive), randomBytes32, randomKey.publicKey)
```

t-bast (commented):
LGTM, will include right after the release.

By the way, https://github.com/ACINQ/eclair/commit/bf2a35f74cf79ce4f76e095916db43d9b4345ec6 is a good example that shows that it's always a subtle choice to use the `nextRemoteCommit` or not, so we need to be wary of not using this `latestRemoteCommit` without thinking long and hard about it.

Maybe adding a comment to it to warn users would be useful, something like:

```
/** Latest commitment we signed. Be careful when using it:
 *   - sometimes we may want to ignore it because the previous commitment is not revoked yet 
 *   - sometimes we may want to take both remote commitments into account
 */
val latestRemoteCommit: RemoteCommit = remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit).getOrElse(remoteCommit)
```

t-bast (commented):
ACK, but I'd like @pm47 to also have a look at this before we merge, as this is a subtle part of the protocol.

pm47 (commented):
This may be a bit error-prone, but the comment is pretty clear about what is going on. That said, we could use a smarter typing to make it harder to shoot ourselves in the feet (pardon the poor naming):
```scala
  sealed trait RemoteCommits {
    def latest: RemoteCommit
    def allSignedHtlcs: Set[DirectedHtlc] // example of a method that may apply to more than one remote commits
  }
  case class Single(current: RemoteCommit) extends RemoteCommits { 
    override def latest = current
    override def allSignedHtlcs: Set[DirectedHtlc] = current.spec.htlcs
  }
  case class Double(current: RemoteCommit, next: RemoteCommit) extends RemoteCommits {
    override def latest = next
    override def allSignedHtlcs: Set[DirectedHtlc] = current.spec.htlcs + next.spec.htlcs
  }
```
Looks a bit over engineered at first glance, but maybe it would make our closing process safer wrt handing of remote commits @t-bast?

Another thing, we should probably acknowledge the fact that `Either` is right-biased (since Scala 2.12), and use the `swap` instead of `left.toOption`. See 40abbe3803d2b697b52c5005134d3f46e6597544.

Also, you could have used the new field here: https://github.com/btcontract/eclair/blob/9fe0975cd2aa30d2addf6d12a61173dab43b4e5e/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1565.

t-bast:
> Looks a bit over engineered at first glance, but maybe it would make our closing process safer wrt handing of remote commits @t-bast?

It's hard to tell yet whether this will be better or worse.
I'd rather introduce that hierarchy in a PR that actually uses it to simplify the closing code.
I'll look into it in my future tx handling types refactoring.

btcontract:
Revisiting this, should not reduced latest remote spec be used in more places (see last commits)? Using it like this also makes `alreadyProposed` method obsolete.

PR #1690, opened 2021-02-14
Feature/simplify api dsl

tompro:
So here is my first try on a bit of API restructuring. Handler are now isolated from each other and grouped first by topic (of my choosing, currently just getting started with Lightning tech) and then for the public route.

Common directives like auth control and default headers are only applied on the top level entry point and free up some setup in unit tests. What I currently don't really like is the customized implicit timeout as it has to be injected into every completer but I could'nt figure out a way to create a completer that provides an implicit implicitly ;)

What do you think?


t-bast:
We're currently finalizing a few features because we want to publish a new release of eclair, I'll have a look at this PR once I'm done with the release work.

Don't worry if we're unable to review it this week, we will pick this up once we have some time available.

t-bast:
Thanks, that's a really nice refactoring, things are much cleaner afterwards.

I'd change a bit the categorization of the existing APIs (it's really hard because some APIs fit in multiple categories, and some don't really fit in any category). There is already a categorization that you can find in `eclair-cli` (https://github.com/ACINQ/eclair/blob/master/eclair-core/eclair-cli). I'd like to modify it to the following to take your proposal into account:

```
  === Node ===
    - getinfo
    - connect
    - disconnect
    - peers
    - audit

  === Channel ===
    - open
    - close
    - forceclose
    - channel
    - channels
    - allchannels
    - allupdates
    - channelstats

  === Fees ===
    - networkfees
    - updaterelayfee

  === Path-finding ===
    - findroute
    - findroutetonode
    - findroutebetweennodes
    - networkstats
    - nodes

  === Invoice ===
    - createinvoice
    - getinvoice
    - listinvoices
    - listpendinginvoices
    - parseinvoice

  === Payment ===
    - usablebalances
    - payinvoice
    - sendtonode
    - sendtoroute
    - getsentinfo
    - getreceivedinfo

  === Message ===
    - signmessage
    - verifymessage

  === OnChain ===
    - getnewaddress
    - sendonchain
    - onchainbalance
    - onchaintransactions
```

Can you update the `eclair-cli` file with that listing and apply it for the API handlers?

Apart from that everything looks good in the PR. Just a couple of comments about styling:

- we like long lines and usually break at around 180 characters (I'm sure everyone has a decent screen nowadays and can affort these long lines)
- we use IntelliJ's default formatting for scala (ctrl+alt+L) and avoid having multiple blank lines

Thanks again, I think this is a really nice refactoring and paves the way for more improvements afterwards, we will include this PR once we're finished with the next eclair release.

tompro:
Yes, noticed the long lines, will get used to it :). Will apply your suggestions as soon as I have some spare time on my hand.

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/ExtraDirectives.scala
```diff
@@ -14,37 +14,53 @@
...
-  val channelIdFormParam = "channelId".as[ByteVector32](sha256HashUnmarshaller)
-  val channelIdsFormParam = "channelIds".as[List[ByteVector32]](sha256HashesUnmarshaller)
-  val nodeIdFormParam = "nodeId".as[PublicKey]
-  val nodeIdsFormParam = "nodeIds".as[List[PublicKey]](pubkeyListUnmarshaller)
-  val paymentHashFormParam = "paymentHash".as[ByteVector32](sha256HashUnmarshaller)
-  val fromFormParam = "from".as[Long]
-  val toFormParam = "to".as[Long]
-  val amountMsatFormParam = "amountMsat".as[MilliSatoshi]
-  val invoiceFormParam = "invoice".as[PaymentRequest]
+  val shortChannelIdFormParam: NameUnmarshallerReceptacle[ShortChannelId] =
```
t-bast, line 40: nit: can you put these all back to a single line?

eclair-node/src/main/scala/fr/acinq/eclair/api/serde/FormParamExtractors.scala
```diff
@@ -14,15 +14,15 @@
...
+package fr.acinq.eclair.api.serde
 
 import java.util.UUID
 
 import akka.http.scaladsl.unmarshalling.Unmarshaller
 import akka.util.Timeout
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{ByteVector32, Satoshi}
-import fr.acinq.eclair.api.JsonSupport._
+import JsonSupport._
```
t-bast, line 25: nit: revert?

eclair-node/src/main/scala/fr/acinq/eclair/api/serde/FormParamExtractors.scala
```diff
@@ -32,6 +32,7 @@ import scodec.bits.ByteVector
 import scala.concurrent.duration._
 import scala.util.Try
 
+
```
t-bast, line 35: nit: revert?

t-bast (commented):
Thanks, a few nits but otherwise looks good.

There is a regression though that needs to be fixed:

```sh
> alice-eclair-cli getinf
The requested resource could not be found.

> bob-eclair-cli getinf
parse error: Invalid numeric literal at line 1, column 4
```

where `alice-eclair-cli` uses `master` and `bob-eclair-cli` uses your branch.

I have the same issue if I provide an invalid password, `master` correctly returns `The supplied authentication is invalid` whereas your change returns `parse error: Invalid numeric literal at line 1, column 4`.

You may be interested in https://github.com/t-bast/lightning-cfg to simplify E2E testing locally (this is what I used with tweaks to use your branch for Bob).

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/ExtraDirectives.scala
tompro: Think I need to change some more Idea settings

t-bast (approved):
LGTM, thanks @tompro!

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Channel.scala
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  val channelStats: Route = postRequest("channelstats") { implicit t =>
+    formFields(fromFormParam.?, toFormParam.?) { (from_opt, to_opt) =>
+      complete(eclairApi.channelStats(from_opt, to_opt))
+    }
+  }
+
+  val channelRoutes: Route = open ~ close ~ forceClose ~ channel ~ channels ~ allChannels ~ allUpdates ~ channelStats
```
pm47, line 96: Defining a variable for each single path seems a bit overkill, and now we must not forget to add each one of them to the `channelRoutes`.

pm47 (commented):
Looks good! Just one nit.

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Channel.scala
tompro: It would be very easy to collect them in a block (you still need to chain them), you loose the ability to use/test them independently though and they will be intended again (but just 1 level).

pm47: Fair enough.

pm47 (approved):

t-bast:
Thanks @tompro!

PR #1691, opened 2021-02-15
Do not send shutdown with unsigned `update_fee`

pm47:
We didn't consider  pending `updating_fee`s when going from `NORMAL` to
`SHUTDOWN` or `NEGOTIATING`. This lead to a deadlock in the following
scenario:

```
                    A                 B
                    |-- update_fee -->|
                    |--- shutdown --->|
 A goes to SHUTDOWN |<-- shutdown ----| B goes to NEGOTIATING
                    |-- commit_sig -->|
                    |                 |
```
because B will ignore A's `commit_sig`, and A will stay in state
`SHUTDOWN` forever, waiting for B's `revoke_and_ack`.

This is an alternative to #1661.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -84,6 +84,8 @@ case class Commitments(channelVersion: ChannelVersion,
 
   def hasNoPendingHtlcs: Boolean = localCommit.spec.htlcs.isEmpty && remoteCommit.spec.htlcs.isEmpty && remoteNextCommitInfo.isRight
 
+  def hasNoPendingFeeUpdate: Boolean = (localChanges.signed ++ localChanges.acked ++ remoteChanges.signed ++ remoteChanges.acked).collectFirst { case _: UpdateFee => true }.isEmpty && remoteNextCommitInfo.isRight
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -84,6 +84,8 @@ case class Commitments(channelVersion: ChannelVersion,
 
   def hasNoPendingHtlcs: Boolean = localCommit.spec.htlcs.isEmpty && remoteCommit.spec.htlcs.isEmpty && remoteNextCommitInfo.isRight
```
pm47, line 85: This method is actually misleading, because the `remoteNextCommitInfo.isRight` statement will be true if the next signature only contains an `update_fee`. We should probably merge `hasNoPendingHtlcs` and `hasNoPendingFeeUpdate` (they are always called together anyway).

t-bast: Good catch

t-bast: Actually not really, it's called independently in `hasPendingOrProposedHtlcs`. But we can factor them in a new method that calls this one.

PR #723, opened 2018-09-25
Handle mutual close in state `DATA_WAIT_FOR_FUNDING_CONFIRMED`

pm47:
This is a clone of #358 which apparently lost track of the remote branch (`unknown repository`).

PR #1692, opened 2021-02-17
Update funding timeout fundee

t-bast:
Our previous timeout was based on timestamps, mostly because `blockCount` could be `0` on mobile using Electrum until a new block was received. Now that we're diverging from the mobile wallet codebase, we can use block heights instead which is more accurate.
    
See lightningnetwork/lightning-rfc#839

I've done some E2E tests on regtest for the "migration" code. There is one edge case where a channel in `WAIT_FOR_FUNDING_CONFIRMED` will not be automatically cleaned up: if the peer never connects back to you, the channel will stay in `OFFLINE` and won't check for funding timeout. I think this edge case is quite harmless and can be ignored (I don't think it's worth adding more code to fix it).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -69,8 +69,8 @@ object Channel {
   // since BOLT 1.1, there is a max value for the refund delay of the main commitment tx
   val MAX_TO_SELF_DELAY = CltvExpiryDelta(2016)
 
-  // as a fundee, we will wait that much time for the funding tx to confirm (funder will rely on the funding tx being double-spent)
-  val FUNDING_TIMEOUT_FUNDEE = 5 days
+  // as a fundee, we will wait that many blocks for the funding tx to confirm (funder will rely on the funding tx being double-spent)
+  val FUNDING_TIMEOUT_FUNDEE = 2016
 
   // pruning occurs if no new update has been received in two weeks (BOLT 7)
   val REFRESH_CHANNEL_UPDATE_INTERVAL = 10 days
@@ -161,7 +161,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   startWith(WAIT_FOR_INIT_INTERNAL, Nothing)
 
   when(WAIT_FOR_INIT_INTERNAL)(handleExceptions {
-    case Event(initFunder@INPUT_INIT_FUNDER(temporaryChannelId, fundingSatoshis, pushMsat, initialFeeratePerKw, fundingTxFeeratePerKw, initialRelayFees_opt, localParams, remote, _, channelFlags, channelVersion), Nothing) =>
+    case Event(initFunder@INPUT_INIT_FUNDER(temporaryChannelId, fundingSatoshis, pushMsat, initialFeeratePerKw, fundingTxFeeratePerKw, _, localParams, remote, _, channelFlags, channelVersion), Nothing) =>
       context.system.eventStream.publish(ChannelCreated(self, peer, remoteNodeId, isFunder = true, temporaryChannelId, initialFeeratePerKw, Some(fundingTxFeeratePerKw)))
       activeConnection = remote
       val fundingPubKey = keyManager.fundingPublicKey(localParams.fundingKeyPath).publicKey
@@ -238,7 +238,11 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
               }
           }
           // no need to go OFFLINE, we can directly switch to CLOSING
-          goto(CLOSING) using closing
+          if (closing.waitingSinceBlock > 1500000) {
+            goto(CLOSING) using closing.copy(waitingSinceBlock = nodeParams.currentBlockHeight)
+          } else {
+            goto(CLOSING) using closing
+          }
 
         case normal: DATA_NORMAL =>
           // TODO: should we wait for an acknowledgment from the watcher?
@@ -280,7 +284,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           blockchain ! WatchLost(self, data.commitments.commitInput.outPoint.txid, nodeParams.minDepthBlocks, BITCOIN_FUNDING_LOST)
           // we make sure that the funding tx has been published
           blockchain ! GetTxWithMeta(funding.commitments.commitInput.outPoint.txid)
-          goto(OFFLINE) using data
+          if (funding.waitingSinceBlock > 1500000) {
+            // we were using timestamps instead of block heights when the channel was created: we reset it to use block heights
+            goto(OFFLINE) using funding.copy(waitingSinceBlock = nodeParams.currentBlockHeight)
+          } else {
+            goto(OFFLINE) using funding
+          }
 
         case _ =>
           // TODO: should we wait for an acknowledgment from the watcher?
@@ -474,8 +483,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
               val fundingMinDepth = Helpers.minDepthForFunding(nodeParams, fundingAmount)
               blockchain ! WatchSpent(self, commitInput.outPoint.txid, commitInput.outPoint.index.toInt, commitInput.txOut.publicKeyScript, BITCOIN_FUNDING_SPENT) // TODO: should we wait for an acknowledgment from the watcher?
               blockchain ! WatchConfirmed(self, commitInput.outPoint.txid, commitInput.txOut.publicKeyScript, fundingMinDepth, BITCOIN_FUNDING_DEPTHOK)
-              context.system.scheduler.scheduleOnce(FUNDING_TIMEOUT_FUNDEE, self, BITCOIN_FUNDING_TIMEOUT)
-              goto(WAIT_FOR_FUNDING_CONFIRMED) using DATA_WAIT_FOR_FUNDING_CONFIRMED(commitments, None, initialRelayFees_opt, now, None, Right(fundingSigned)) storing() sending fundingSigned
+              goto(WAIT_FOR_FUNDING_CONFIRMED) using DATA_WAIT_FOR_FUNDING_CONFIRMED(commitments, None, initialRelayFees_opt, nodeParams.currentBlockHeight, None, Right(fundingSigned)) storing() sending fundingSigned
           }
       }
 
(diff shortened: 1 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -69,8 +69,8 @@ object Channel {
   // since BOLT 1.1, there is a max value for the refund delay of the main commitment tx
   val MAX_TO_SELF_DELAY = CltvExpiryDelta(2016)
 
-  // as a fundee, we will wait that much time for the funding tx to confirm (funder will rely on the funding tx being double-spent)
-  val FUNDING_TIMEOUT_FUNDEE = 5 days
+  // as a fundee, we will wait that many blocks for the funding tx to confirm (funder will rely on the funding tx being double-spent)
+  val FUNDING_TIMEOUT_FUNDEE = 2016
```
pm47, line 73: Okay I have to say it. How about we simply replaced this with:
```suggestion
  val FUNDING_TIMEOUT_FUNDEE = 2 weeks
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -238,7 +238,11 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
               }
           }
           // no need to go OFFLINE, we can directly switch to CLOSING
-          goto(CLOSING) using closing
+          if (closing.waitingSinceBlock > 1500000) {
```
pm47, line 241:
```suggestion
          if (closing.waitingSinceBlock > 1500000) {
            // we were using timestamps instead of block heights when the channel was created: we reset it *and* we use block heights
```

NB: This code will break in 15.7 year 😱

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -280,7 +284,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           blockchain ! WatchLost(self, data.commitments.commitInput.outPoint.txid, nodeParams.minDepthBlocks, BITCOIN_FUNDING_LOST)
           // we make sure that the funding tx has been published
           blockchain ! GetTxWithMeta(funding.commitments.commitInput.outPoint.txid)
-          goto(OFFLINE) using data
+          if (funding.waitingSinceBlock > 1500000) {
+            // we were using timestamps instead of block heights when the channel was created: we reset it to use block heights
```
pm47, line 288:
```suggestion
            // we were using timestamps instead of block heights when the channel was created: we reset it *and* we use block heights
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -238,7 +238,11 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
               }
           }
           // no need to go OFFLINE, we can directly switch to CLOSING
-          goto(CLOSING) using closing
+          if (closing.waitingSinceBlock > 1500000) {
+            goto(CLOSING) using closing.copy(waitingSinceBlock = nodeParams.currentBlockHeight)
```
pm47, line 242: If we don't update the db we'll push back forever:
```suggestion
            goto(CLOSING) using closing.copy(waitingSinceBlock = nodeParams.currentBlockHeight) storing()
```

NB: we don't seem to be using `waitingSinceBlock` for anything in state `CLOSING`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -280,7 +284,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           blockchain ! WatchLost(self, data.commitments.commitInput.outPoint.txid, nodeParams.minDepthBlocks, BITCOIN_FUNDING_LOST)
           // we make sure that the funding tx has been published
           blockchain ! GetTxWithMeta(funding.commitments.commitInput.outPoint.txid)
-          goto(OFFLINE) using data
+          if (funding.waitingSinceBlock > 1500000) {
+            // we were using timestamps instead of block heights when the channel was created: we reset it to use block heights
+            goto(OFFLINE) using funding.copy(waitingSinceBlock = nodeParams.currentBlockHeight)
```
pm47, line 289: If we don't update the db we'll push back forever:
```suggestion
            goto(OFFLINE) using funding.copy(waitingSinceBlock = nodeParams.currentBlockHeight) storing()
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -592,10 +600,19 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
 
+    case Event(c: CurrentBlockCount, d: DATA_WAIT_FOR_FUNDING_CONFIRMED) => d.fundingTx match {
+      case Some(_) => stay // we are funder, we're still waiting for the funding tx to be confirmed
+      case None if c.blockCount - d.waitingSinceBlock > FUNDING_TIMEOUT_FUNDEE =>
+        log.warning(s"funding tx hasn't been published in ${c.blockCount - d.waitingSinceBlock} blocks")
+        self ! BITCOIN_FUNDING_TIMEOUT
+        stay
+      case None => stay // let's wait longer
+    }
+
```
pm47, line 615: This needs to be done in `OFFLINE`/`SYNCING` state too I believe.

Edit: saw your comment
> There is one edge case where a channel in WAIT_FOR_FUNDING_CONFIRMED will not be automatically cleaned up: if the peer never connects back to you, the channel will stay in OFFLINE and won't check for funding timeout. I think this edge case is quite harmless and can be ignored (I don't think it's worth adding more code to fix it).

I can definitely think of a scenario when peers open channels to us, then fail to publish and disappear. What makes it okay is that we will clean up at next startup.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: That was my first attempt, but then I thought it would be much better to do the right thing and start using block heights here instead of timestamps, it ensures we're really spec compliant (and our previous code was constantly pushing back the timeout after each restart, which I think was unintended).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: I hope that in 15.7 years every eclair node will run a more recent version than eclair 0.5.1 and we'll have removed that code :smile:

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: We are, because we do a `GetTxWithMeta` if the funding tx is unconfirmed (and it allows us to move to the `CLOSED` state).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Good idea!

pm47:
Related: #723

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Exactly, I think this is ok and not worth adding more code.

eclair-core/src/test/scala/fr/acinq/eclair/wire/ChannelCodecsSpec.scala
```diff
@@ -335,7 +335,7 @@ class ChannelCodecsSpec extends AnyFunSuite {
     // let's decode the old data (this will use the old codec that provides default values for new fields)
     val data_new = stateDataCodec.decode(bin_old.toBitVector).require.value
     assert(data_new.asInstanceOf[DATA_WAIT_FOR_FUNDING_CONFIRMED].fundingTx === None)
-    assert(System.currentTimeMillis.milliseconds.toSeconds - data_new.asInstanceOf[DATA_WAIT_FOR_FUNDING_CONFIRMED].waitingSince < 3600) // we just set this timestamp to current time
+    assert(System.currentTimeMillis.milliseconds.toSeconds - data_new.asInstanceOf[DATA_WAIT_FOR_FUNDING_CONFIRMED].waitingSinceBlock < 3600) // we just set this timestamp to current time
```
sstone, line 338: Is this comparison still valid ?

t-bast: We will still do that "trick" for upgrades from old eclair nodes, but it will then be changed again in `Channel.scala` to use a block height. I thought about directly using `currentBlockHeight` in the legacy codec, but it requires changing a `val` to a `def` to pipe the `blockCount` which was messy...

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Right. The meaning of `waitingSinceBlock ` is really ambiguous though, because we are naming it and setting it as "waiting in `CLOSING` since" but we are actually using it as "waiting for funding tx to confirm since". Don't you find this odd? I know it's unrelated to this PR.

For example here we could be using `waitingSinceBlock = waitForFundingConfirmed.waitingSinceBlock`
https://github.com/ACINQ/eclair/blob/48ac6b522dd9ee0b1fae8a143925e54f2e55f1a1/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L2190

t-bast: > The meaning of waitingSinceBlock is really ambiguous though, because we are naming it and setting it as "waiting in CLOSING since" but we are actually using it as "waiting for funding tx to confirm since". Don't you find this odd?

Yes I agree, I can dive more into this and clean that up.

pm47:
How about, the following backward compatibility code? I think we can leave it like that (since this happens before the channel restore, where you handle the timestamp->blockheight migration).

https://github.com/ACINQ/eclair/blob/48ac6b522dd9ee0b1fae8a143925e54f2e55f1a1/eclair-core/src/main/scala/fr/acinq/eclair/wire/LegacyChannelCodecs.scala#L265

https://github.com/ACINQ/eclair/blob/48ac6b522dd9ee0b1fae8a143925e54f2e55f1a1/eclair-core/src/main/scala/fr/acinq/eclair/wire/LegacyChannelCodecs.scala#L333

t-bast:
> How about, the following backward compatibility code? I think we can leave it like that (since this happens before the channel restore, where you handle the timestamp->blockheight migration).

I agree, this is related to https://github.com/ACINQ/eclair/pull/1692#discussion_r579091249 and I think we can leave it like that.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: I think it could be worth it.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1692/commits/4019b6693dd52e5537b0c1c7b9db3cde7fa08942

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1692/commits/4019b6693dd52e5537b0c1c7b9db3cde7fa08942

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: I added a comment in https://github.com/ACINQ/eclair/pull/1692/commits/4019b6693dd52e5537b0c1c7b9db3cde7fa08942

I think it's useful to have `Closing.waitingSinceBlock` really be the block when the *closing* was initiated (and not the waiting-for-confirmed-since-block): it can help check how much time we spend in closing (and draw nice graphs for later analysis).

The only drawback is that it "resets" the funding timeout counter (we allow more time for the funder) but I think that's ok.

pm47 (approved):

PR #1693, opened 2021-02-17
Sort addresses in node announcement

t-bast:
Addresses in node announcement should be sorted.
We accept node announcements that don't do this, but we should do it for our own announcements.

See https://github.com/lightningnetwork/lightning-rfc/pull/842

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
@@ -72,7 +72,13 @@ object Announcements {
 
   def makeNodeAnnouncement(nodeSecret: PrivateKey, alias: String, color: Color, nodeAddresses: List[NodeAddress], features: Features, timestamp: Long = System.currentTimeMillis.milliseconds.toSeconds): NodeAnnouncement = {
     require(alias.length <= 32)
-    val witness = nodeAnnouncementWitnessEncode(timestamp, nodeSecret.publicKey, color, alias, features, nodeAddresses, unknownFields = ByteVector.empty)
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
```diff
@@ -72,7 +72,13 @@ object Announcements {
 
   def makeNodeAnnouncement(nodeSecret: PrivateKey, alias: String, color: Color, nodeAddresses: List[NodeAddress], features: Features, timestamp: Long = System.currentTimeMillis.milliseconds.toSeconds): NodeAnnouncement = {
     require(alias.length <= 32)
-    val witness = nodeAnnouncementWitnessEncode(timestamp, nodeSecret.publicKey, color, alias, features, nodeAddresses, unknownFields = ByteVector.empty)
+    val sortedAddresses = nodeAddresses.map {
+      case address@(_: IPv4) => (1, address)
+      case address@(_: IPv6) => (2, address)
+      case address@(_: Tor2) => (3, address)
+      case address@(_: Tor3) => (4, address)
+    }.sortBy(_._1).map(_._2)
```
pm47, line 80: We could have defined an `Ordering[NodeAddress]`, but that would probably be over engineering.

pm47 (approved):

eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
t-bast: I toyed with it :smiley: but decided that it was indeed a bit over-engineered

PR #1694, opened 2021-02-17
Reject unreasonable remote dust limit

t-bast:
It makes no sense to use a dust limit that's much higher than what bitcoin mandates, so we should not allow our peer to use invalid values.


eclair-core/src/main/resources/reference.conf
```diff
@@ -64,6 +64,7 @@ eclair {
   channel-flags = 1 // announce channels
 
   dust-limit-satoshis = 546
+  max-remote-dust-limit-satoshis = 1000
```
t-bast, line 67: Should we even set this to 546? I think no-one ever used something else.

pm47: If we can't see any other value on our logs in the past month or so I guess yes.

t-bast: Actually it looks like we have some people using `573 sat`. I'll change this limit to `600 sat` then.

pm47 (approved):

Issue #1068, opened 2019-07-10
Feature request: FindRouteToNode, but any 2 nodes


STAWKEYE:
Feature request: API call FindRouteToNode will use Eclair's routing algorithm to find a route from your node to any designated node. I request an API call such as FindRouteBetweenNodes that does exactly the same thing, but between any two defined nodes of your choosing. This could be used to find payment loops between 2 nodes that you already have channels with, that you wish to balance. In this instance, FindRouteToNode will simply return your own payment channel since it is meant for finding the best route for a simple payment.

Now that I think about it, my approach will simply return "Node A -> Your Node -> Node B" if you already have channels with A and B. Perhaps return several paths, or return a single path but omit the obvious path that uses your node as the intermediary.

PR #1695, opened 2021-02-17
Find route between nodes. #1068

tompro:
Added additional method to Eclair like findRoute but allowing for 2 nodeIds. Also added a new endpoint to the http Api "findroutebetweennodes" which takes sourceNode and targetNode as params.

This is the impl of feature request #1068. The ticket is rather old so I don't know whether this is still a thing, but straightforward so I gave it a try. Depending on pull request for #1690 I will either impl the API change there or with this pull.

### DIFF
--- a/eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
+++ b/eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
@@ -230,6 +230,12 @@ trait Service extends ExtraDirectives with Logging {
                             complete(eclairApi.findRoute(nodeId, amount))
                           }
                         } ~
+                        path("findroutebetweennodes") {
+                          formFields("sourceNodeId".as[PublicKey], "targetNodeId".as[PublicKey], amountMsatFormParam) {
+                            (sourceNodeId, targetNodeId, amount) =>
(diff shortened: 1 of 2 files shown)


eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
```diff
@@ -230,6 +230,12 @@ trait Service extends ExtraDirectives with Logging {
                             complete(eclairApi.findRoute(nodeId, amount))
                           }
                         } ~
+                        path("findroutebetweennodes") {
+                          formFields("sourceNodeId".as[PublicKey], "targetNodeId".as[PublicKey], amountMsatFormParam) {
+                            (sourceNodeId, targetNodeId, amount) =>
+                            complete(eclairApi.findRouteBetween(sourceNodeId, targetNodeId, amount))
```
t-bast, line 236: nit: indentation is a bit weird here?
```suggestion
                            (sourceNodeId, targetNodeId, amount) => complete(eclairApi.findRouteBetween(sourceNodeId, targetNodeId, amount))
```

t-bast (commented):
This PR is small enough and trivial enough to include now before our upcoming release, thanks for the contribution! Just a small nit and we should be good to go.

t-bast (approved):

PR #1696, opened 2021-02-18
Create invoice with preimageHash only

prushin:
This is the change of Eclair api receive function - to be able to use PreimageHash explicitly without encrypting it from Preimage.
In this case it will store NULL for Preimage and use PreimageHash instead.

This change is requirement for proper functioning of https://github.com/coingaming/eclair-plugin-hodlinvoice plugin.

As an addition I have added nix (+mavenix) support. This is useful for building Eclair as a dependency for Eclair Plugin HodlInvoice and can be used for any other nix built projects. If not needed, will be removed from PR.

Also otj-pg-embedded has been replaced with it's fork embedded-postgres which supports all it's features but more compatible. Also can be removed if not needed.
https://github.com/zonkyio/embedded-postgres


t-bast:
First high-level comment on the PR (not having looked at the code in details yet): don't do many different things in the same PR unless they are trivial. This PR contains 3 very different things:

- introduce nix build scripts
- change the test postgres backend
- add support for external invoice preimages

These 3 things should be done in separate PRs. Please split accordingly so we can review each of these changes independently.

t-bast:
Closing inactive PR. Feel free to re-open when you pick this up.

PR #1697, opened 2021-02-18
Refactor bitcoin clients

t-bast:
And improve test coverage specifically for the calls we'll rely on for CPFP and RBF.
I wanted to separate this PR from the one that actually use these calls for RBF/CPFP to avoid a monster PR, as these refactoring/clean-ups make sense anyway.

I also split a test in the `ElectrumWatcherSpec` into two tests because it was a bit flaky (block count synchronization between `bitcoind` and `electrum` in the test suite is a bit flaky when a test generates multiple blocks in a sequence).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
@@ -19,10 +19,13 @@ package fr.acinq.eclair.blockchain.bitcoind.rpc
 import fr.acinq.bitcoin._
 import fr.acinq.eclair.ShortChannelId.coordinates
 import fr.acinq.eclair.TxCoordinates
+import fr.acinq.eclair.blockchain.fee.{FeeratePerKB, FeeratePerKw}
 import fr.acinq.eclair.blockchain.{GetTxWithMetaResponse, UtxoStatus, ValidateResult}
+import fr.acinq.eclair.transactions.Transactions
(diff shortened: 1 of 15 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -84,6 +89,41 @@ class ExtendedBitcoinClient(val rpcClient: BitcoinJsonRPCClient) {
       index = txs.indexOf(JString(txid.toHex))
     } yield (height.toInt, index)
 
+  def fundTransaction(tx: Transaction, options: FundTransactionOptions)(implicit ec: ExecutionContext): Future[FundTransactionResponse] = {
+    rpcClient.invoke("fundrawtransaction", tx.toString(), options).map(json => {
+      val JString(hex) = json \ "hex"
+      val JInt(changePos) = json \ "changepos"
+      val JDecimal(fee) = json \ "fee"
+      val fundedTx = Transaction.read(hex)
+      val amountIn = toSatoshi(fee) + fundedTx.txOut.map(_.amount).sum
```
pm47, line 98: 👀 

Can you make `amountIn` a method of `FundTransactionResponse`? I'd prefer we'd keep the response object a mirror of what bitcoin core returns.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1697/commits/b89cd280923c39e1f9207c2368e79c3ba3e9d21b

pm47 (approved):

PR #1699, opened 2021-02-19
Optimize watching for spending txs

pm47:
Since we almost always know which transactions will spend the utxos that we are watching, we can optimize the watcher to look for those instead of starting from scratch.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
@@ -19,7 +19,7 @@ package fr.acinq.eclair.blockchain
 import akka.actor.ActorRef
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{ByteVector32, Script, ScriptWitness, Transaction}
-import fr.acinq.eclair.channel.BitcoinEvent
+import fr.acinq.eclair.channel.{BITCOIN_FUNDING_SPENT, BitcoinEvent, Commitments}
 import fr.acinq.eclair.wire.ChannelAnnouncement
 import scodec.bits.ByteVector
 
@@ -72,11 +72,18 @@ object WatchConfirmed {
  * @param outputIndex     index of the outpoint to watch.
  * @param publicKeyScript electrum requires us to specify a public key script; the script of the outpoint must be provided.
  * @param event           channel event related to the outpoint.
+ * @param hints           txids of potential spending transactions; most of the time we know the txs, and it allows for optimizations.
+ *                        This argument can safely be ignored by watcher implementations.
  */
-final case class WatchSpent(replyTo: ActorRef, txId: ByteVector32, outputIndex: Int, publicKeyScript: ByteVector, event: BitcoinEvent) extends Watch
+final case class WatchSpent(replyTo: ActorRef, txId: ByteVector32, outputIndex: Int, publicKeyScript: ByteVector, event: BitcoinEvent, hints: Seq[ByteVector32]) extends Watch
 object WatchSpent {
   // if we have the entire transaction, we can get the publicKeyScript from the relevant output
-  def apply(replyTo: ActorRef, tx: Transaction, outputIndex: Int, event: BitcoinEvent): WatchSpent = WatchSpent(replyTo, tx.txid, outputIndex, tx.txOut(outputIndex).publicKeyScript, event)
+  def apply(replyTo: ActorRef, tx: Transaction, outputIndex: Int, event: BitcoinEvent, hints: Seq[ByteVector32]): WatchSpent = WatchSpent(replyTo, tx.txid, outputIndex, tx.txOut(outputIndex).publicKeyScript, event, hints)
+
+  def watchFundingTx(channel: ActorRef, commitments: Commitments): WatchSpent = {
+    val knownSpendingTxs = commitments.localCommit.publishableTxs.commitTx.tx.txid +: commitments.remoteCommit.txid +: commitments.remoteNextCommitInfo.left.toSeq.map(_.nextRemoteCommit.txid)
+    WatchSpent(channel, commitments.commitInput.outPoint.txid, commitments.commitInput.outPoint.index.toInt, commitments.commitInput.txOut.publicKeyScript, BITCOIN_FUNDING_SPENT, knownSpendingTxs)
+  }
 }
 
 /**
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
@@ -79,11 +79,6 @@ final case class WatchSpent(replyTo: ActorRef, txId: ByteVector32, outputIndex: 
 object WatchSpent {
   // if we have the entire transaction, we can get the publicKeyScript from the relevant output
   def apply(replyTo: ActorRef, tx: Transaction, outputIndex: Int, event: BitcoinEvent, hints: Seq[ByteVector32]): WatchSpent = WatchSpent(replyTo, tx.txid, outputIndex, tx.txOut(outputIndex).publicKeyScript, event, hints)
-
-  def watchFundingTx(channel: ActorRef, commitments: Commitments): WatchSpent = {
-    val knownSpendingTxs = commitments.localCommit.publishableTxs.commitTx.tx.txid +: commitments.remoteCommit.txid +: commitments.remoteNextCommitInfo.left.toSeq.map(_.nextRemoteCommit.txid)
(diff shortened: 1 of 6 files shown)


pm47:
This is much more efficient that what we currently and will significantly reduce resource usage at startup, but this part of the code is obviously sensitive. I think it is worth it, can you review extra carefully @sstone @t-bast? I've done my best to reduce the possibility of regressions.

sstone:
LGTM from a functional point of view but we never test that we actually use the provided hints for find a spending txs. What about extracting the relevant  methods from `ZmqWatcher` and adding them to `ExtendedBitcoinClient` ?

pm47:
> LGTM from a functional point of view but we never test that we actually use the provided hints for find a spending txs.

Indeed, I verified manually by running the tests with logging enabled. The point of those tests (when run automatically) is to prove non-regression (particularly with a wrong hint), they don't prove the optimization.

>What about extracting the relevant methods from ZmqWatcher and adding them to ExtendedBitcoinClient?

I don't know, WDYT @t-bast?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
```diff
@@ -72,11 +72,18 @@ object WatchConfirmed {
...
+ *                        This argument can safely be ignored by watcher implementations.
  */
-final case class WatchSpent(replyTo: ActorRef, txId: ByteVector32, outputIndex: Int, publicKeyScript: ByteVector, event: BitcoinEvent) extends Watch
+final case class WatchSpent(replyTo: ActorRef, txId: ByteVector32, outputIndex: Int, publicKeyScript: ByteVector, event: BitcoinEvent, hints: Seq[ByteVector32]) extends Watch
 object WatchSpent {
   // if we have the entire transaction, we can get the publicKeyScript from the relevant output
-  def apply(replyTo: ActorRef, tx: Transaction, outputIndex: Int, event: BitcoinEvent): WatchSpent = WatchSpent(replyTo, tx.txid, outputIndex, tx.txOut(outputIndex).publicKeyScript, event)
+  def apply(replyTo: ActorRef, tx: Transaction, outputIndex: Int, event: BitcoinEvent, hints: Seq[ByteVector32]): WatchSpent = WatchSpent(replyTo, tx.txid, outputIndex, tx.txOut(outputIndex).publicKeyScript, event, hints)
+
+  def watchFundingTx(channel: ActorRef, commitments: Commitments): WatchSpent = {
```
t-bast, line 83: I'd prefer putting that function in a channel-related file (why not in `Channel.scala` itself?).
Watcher/blockchain components don't really need to know about `commitments` and other channel-related data.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
```diff
@@ -72,11 +72,18 @@ object WatchConfirmed {
...
+    val knownSpendingTxs = commitments.localCommit.publishableTxs.commitTx.tx.txid +: commitments.remoteCommit.txid +: commitments.remoteNextCommitInfo.left.toSeq.map(_.nextRemoteCommit.txid)
```
t-bast, line 84: Would it make sense to also include `mutualClosePublished` when we have them?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -140,28 +140,41 @@ class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client: Extend
...
-                  log.info(s"$txid:$outputIndex has already been spent, looking for the spending tx in the mempool")
-                  client.getMempool().map { mempoolTxs =>
-                    mempoolTxs.filter(tx => tx.txIn.exists(i => i.outPoint.txid == txid && i.outPoint.index == outputIndex)) match {
-                      case Nil =>
-                        log.warning(s"$txid:$outputIndex has already been spent, spending tx not in the mempool, looking in the blockchain...")
-                        client.lookForSpendingTx(None, txid, outputIndex).map { tx =>
-                          log.warning(s"found the spending tx of $txid:$outputIndex in the blockchain: txid=${tx.txid}")
-                          self ! NewTransaction(tx)
+              // if we know some potential spending txs, we try to fetch them directly
+              Future.sequence(hints.map(txid => client.getTransaction(txid).map(Some(_)).recover { case _ => None }))
```
t-bast, line 148: Is there a reason why you start with these instead of doing first `client.isTransactionOutputSpendable` and then try to find the culprit?

t-bast (commented):
This is definitely an improvement, whenever we can avoid a call to `getMempool` let's avoid it!

> What about extracting the relevant methods from ZmqWatcher and adding them to ExtendedBitcoinClient?

I don't have a strong opinion here, it could make sense in `ExtendedBitcoinClient` but if only `ZmqWatcher` is using it, then it also makes sense to keep it in `ZmqWatcher`...maybe leave it like that for now and after the re-architecting that will happen around `ZmqWatcher` we'll see?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
pm47: Absolutely! good catch

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
pm47: Interesting, will try that

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
pm47: Done with 3f69778.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
pm47: Done with b16fc45.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
pm47: Done with 848ac2b.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
```diff
@@ -19,7 +19,7 @@ package fr.acinq.eclair.blockchain
 import akka.actor.ActorRef
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{ByteVector32, Script, ScriptWitness, Transaction}
-import fr.acinq.eclair.channel.BitcoinEvent
+import fr.acinq.eclair.channel.{BITCOIN_FUNDING_SPENT, BitcoinEvent, Commitments}
```
t-bast, line 22: nit: unused imports

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1944,6 +1936,14 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     stay
   }
 
+  def watchFundingTx(commitments: Commitments, additionalKnownSpendingTxs: Seq[ByteVector32] = Nil): Unit = {
+    // TODO: should we wait for an acknowledgment from the watcher?
+    val knownSpendingTxs: Seq[ByteVector32] = List(commitments.localCommit.publishableTxs.commitTx.tx.txid, commitments.remoteCommit.txid) ++ commitments.remoteNextCommitInfo.left.toSeq.map(_.nextRemoteCommit.txid) ++ additionalKnownSpendingTxs
```
t-bast, line 1941: nit: probably worth using a `Set` here, otherwise the commit txs will be duplicated (they are in `DATA_CLOSING.spendingTxes` and added manually here.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
pm47: Done with 1b32f188e.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Indeed duplicates don't make sense for hints. Done with fa8ceac59.

t-bast (approved):
LGTM :rocket:

Issue #1206, opened 2019-11-11
eclair-cli usage example color formatting not working in different shells


fjahr:
I have been using the default version (old) of Bash in the latest macOS version and the color formatting in the usage example was not working as intended. I checked and the \e is not supported in the version that ships with macOS currently. However, I switched to `zsh` (5.7.1) and it was the same case there. Then upgraded Bash via homebrew (5.0.11) and I am still seeing the issue. I have been trying the See example usage line:

```
$ eclair-core/eclair-cli
==============================
Command line client for eclair
==============================

This tool requires the eclair node's API to be enabled and listening
on <http://localhost:8080>.

Usage
-----
\e[93meclair-cli\e[39m [\e[93mOPTIONS\e[39m]... <\e[93mCOMMAND\e[39m> [--command-param=command-value]...

where OPTIONS can be:
  -p <password>         API's password
  -a <address>          Override the API URL with <address>
  -h                    Show this help
  -s                    Some commands can print a trimmed JSON

and COMMAND is one of:
  getinfo, connect, disconnect, open, close, forceclose, updaterelayfee,
  peers, channels, channel, allnodes, allchannels, allupdates
  findroute, findroutetonode, parseinvoice, payinvoice, sendtonode,
  sendtoroute, getsentinfo, createinvoice, getinvoice, listinvoices,
  listpendinginvoices, getreceivedinfo, audit, networkfees,
  channelstats, usablebalances

Examples
--------
  eclair-cli -a localhost:1234 peers        list the peers of a node hosted on localhost:1234
  eclair-cli close --channelId=006fb...     closes the channel with id 006fb...


Full documentation here: <https://acinq.github.io/eclair>
```

Using the color formatting in `printf` it comes out correctly, so there seems to be some kind of escaping issue?
```
$ printf '\e[93meclair-cli\e[39m [\e[93mOPTIONS\e[39m]... <\e[93mCOMMAND\e[39m> [--command-param=command-value]'
eclair-cli [OPTIONS]... <COMMAND> [--command-param=command-value]
```

PR #1700, opened 2021-02-21
#1206 Use multiplatform escape sequence, added cli findroutebetweennodes

tompro:
Using this escape sequence should solve the coloring issue on macOs. I don't have access to Win or Mac but it works fine in several Linux terminals so it won't get worse ;)

I also added the new "findroutebetweennodes" hint to the cli and completion file.


pm47:
Can confirm it works on windows (mingw).

@sstone @dpad85 any of you can check on MacOS?

t-bast (approved):
I just tested this on macOS and linux, it works fine, thanks!

PR #1705, opened 2021-02-24
Address all intellij warnings for Channel.scala

pm47:
Fixing those small warnings one by one in unrelated PRs makes reviews more difficult.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -47,33 +47,33 @@ import scala.util.{Failure, Success, Try}
  */
 
 object Channel {
-  def props(nodeParams: NodeParams, wallet: EclairWallet, remoteNodeId: PublicKey, blockchain: ActorRef, relayer: ActorRef, origin_opt: Option[ActorRef]) = Props(new Channel(nodeParams, wallet, remoteNodeId, blockchain, relayer, origin_opt))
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -47,33 +47,33 @@ import scala.util.{Failure, Success, Try}
...
-  val MIN_DUSTLIMIT = 546 sat
+  val MIN_DUSTLIMIT: Satoshi = 546 sat
 
   // we won't exchange more than this many signatures when negotiating the closing fee
   val MAX_NEGOTIATION_ITERATIONS = 20
 
-  // this is defined in BOLT 11
-  val MIN_CLTV_EXPIRY_DELTA = CltvExpiryDelta(18)
-  val MAX_CLTV_EXPIRY_DELTA = CltvExpiryDelta(7 * 144) // one week
+  // this is private defined in BOLT 11
```
t-bast, line 65: What do you mean?

pm47: It's just the result of a reckless replace to convert all `def` to `private def`...

pm47: Fixed with 8b8b042.

t-bast (approved):

pm47:
I think I can merge this, there shouldn't be many conflicts with #1702 WDYT @t-bast?

t-bast:
Sure, let's merge this, if it does conflict it should be a simple rebase which shouldn't change anything in #1702

PR #1706, opened 2021-02-24
Relay partially failed htlcs when closing

t-bast:
If a channel closes when we've received an `UpdateFailHtlc`, signed it but not yet received our peer's revocation, we need to fail the htlc upstream.

That specific scenario was not correctly handled, resulting in upstream htlcs that were not failed which would force our upstream peer to close the channel.

In that scenario the HTLC is in the `remoteCommit` but not in the `nextRemoteCommit`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -954,8 +954,8 @@ object Helpers {
       val remoteCommit = d.commitments.remoteCommit
       val nextRemoteCommit_opt = d.commitments.remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit)
       if (localCommit.publishableTxs.commitTx.tx.txid == tx.txid) {
-        // our commit got confirmed, so any htlc that is in their commitment but not ours will never reach the chain
-        val htlcsInRemoteCommit = remoteCommit.spec.htlcs ++ nextRemoteCommit_opt.map(_.spec.htlcs).getOrElse(Set.empty)
+        // our commit got confirmed, so any htlc that is in their commitment but not in ours will never reach the chain
+        val htlcsInRemoteCommit = remoteCommit.spec.htlcs ++ nextRemoteCommit_opt.flatMap(_.spec.htlcs)
         // NB: from the p.o.v of remote, their incoming htlcs are our outgoing htlcs
         htlcsInRemoteCommit.collect(incoming) -- localCommit.spec.htlcs.collect(outgoing)
       } else if (remoteCommit.txid == tx.txid) {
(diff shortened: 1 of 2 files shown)


pm47:
> If a channel closes when we've received an UpdateFailHtlc, signed it but not yet received our peer's revocation, we need to fail the htlc upstream.

I don't think we can. What if the peer publishes its previous commitment (that isn't revoked) and uses the preimage?

t-bast:
> What if the peer publishes its previous commitment 

He can't, we waited for our local commit to be confirmed ;)

pm47:
>> What if the peer publishes its previous commitment

>He can't, we waited for our local commit to be confirmed ;)

Well then yes. Alright, you got me I didn't go past the PR description

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -954,10 +954,10 @@ object Helpers {
       val remoteCommit = d.commitments.remoteCommit
       val nextRemoteCommit_opt = d.commitments.remoteNextCommitInfo.left.toOption.map(_.nextRemoteCommit)
       if (localCommit.publishableTxs.commitTx.tx.txid == tx.txid) {
-        // our commit got confirmed, so any htlc that we signed but they didn't sign will never reach the chain
-        val mostRecentRemoteCommit = nextRemoteCommit_opt.getOrElse(remoteCommit)
+        // our commit got confirmed, so any htlc that is in their commitment but not ours will never reach the chain
```
pm47, line 957:
```suggestion
        // our commit got confirmed, so any htlc that is in their commitment but not in ours will never reach the chain
```
maitre_capello.jpg

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -954,10 +954,10 @@ object Helpers {
...
+        val htlcsInRemoteCommit = remoteCommit.spec.htlcs ++ nextRemoteCommit_opt.map(_.spec.htlcs).getOrElse(Set.empty)
```
pm47, line 958: 😎 

```suggestion
        val htlcsInRemoteCommit = remoteCommit.spec.htlcs ++ nextRemoteCommit_opt.flatMap(_.spec.htlcs)
```

pm47 (commented):
When I see how complex that corner case is and how simple and elegant the fix _and_ the tests are, I think we can be proud of that codebase!

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1706/commits/e2119a400387f34134257c5975f884a23c0c3411

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1706/commits/e2119a400387f34134257c5975f884a23c0c3411

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Actually it doesn't build...

pm47 (approved):

PR #1709, opened 2021-02-26
Improve channel state tests

pm47:
Commits can be reviewed independently.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
@@ -98,29 +98,36 @@ trait StateTestsHelperMethods extends TestKitBase {
   }
 
   def reachNormal(setup: SetupFixture, tags: Set[String] = Set.empty): Unit = {
+    import com.softwaremill.quicklens._
     import setup._
+    val channelVersion = List(
+      ChannelVersion.STANDARD,
+      if (tags.contains(StateTestsTags.AnchorOutputs)) ChannelVersion.ANCHOR_OUTPUTS else ChannelVersion.ZEROES,
+      if (tags.contains(StateTestsTags.StaticRemoteKey)) ChannelVersion.STATIC_REMOTEKEY else ChannelVersion.ZEROES,
+    ).reduce(_ | _)
     val channelFlags = if (tags.contains(StateTestsTags.ChannelsPublic)) ChannelFlags.AnnounceChannel else ChannelFlags.Empty
-    val pushMsat = if (tags.contains(StateTestsTags.NoPushMsat)) 0.msat else TestConstants.pushMsat
-    val (aliceParams, bobParams, channelVersion) = if (tags.contains(StateTestsTags.AnchorOutputs)) {
-      val features = Features(Set(ActivatedFeature(Features.StaticRemoteKey, FeatureSupport.Mandatory), ActivatedFeature(Features.AnchorOutputs, FeatureSupport.Optional)))
-      (Alice.channelParams.copy(features = features), Bob.channelParams.copy(features = features), ChannelVersion.ANCHOR_OUTPUTS)
-    } else if (tags.contains(StateTestsTags.StaticRemoteKey)) {
-      val features = Features(Set(ActivatedFeature(Features.StaticRemoteKey, FeatureSupport.Optional)))
-      val aliceParams = Alice.channelParams.copy(features = features, walletStaticPaymentBasepoint = Some(Helpers.getWalletPaymentBasepoint(wallet)))
-      val bobParams = Bob.channelParams.copy(features = features, walletStaticPaymentBasepoint = Some(Helpers.getWalletPaymentBasepoint(wallet)))
-      (aliceParams, bobParams, ChannelVersion.STATIC_REMOTEKEY)
-    } else {
-      (Alice.channelParams, Bob.channelParams, ChannelVersion.STANDARD)
-    }
+    val aliceParams = Alice.channelParams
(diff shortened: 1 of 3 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -98,29 +98,36 @@ trait StateTestsHelperMethods extends TestKitBase {
   }
 
   def reachNormal(setup: SetupFixture, tags: Set[String] = Set.empty): Unit = {
+    import com.softwaremill.quicklens._
     import setup._
+    val channelVersion = List(
+      ChannelVersion.STANDARD,
+      if (tags.contains(StateTestsTags.AnchorOutputs)) ChannelVersion.ANCHOR_OUTPUTS else ChannelVersion.ZEROES,
+      if (tags.contains(StateTestsTags.StaticRemoteKey)) ChannelVersion.STATIC_REMOTEKEY else ChannelVersion.ZEROES,
+    ).reduce(_ | _)
```
t-bast, line 107: Nice use of `reduce`

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -98,29 +98,36 @@ trait StateTestsHelperMethods extends TestKitBase {
...
-      val features = Features(Set(ActivatedFeature(Features.StaticRemoteKey, FeatureSupport.Optional)))
-      val aliceParams = Alice.channelParams.copy(features = features, walletStaticPaymentBasepoint = Some(Helpers.getWalletPaymentBasepoint(wallet)))
-      val bobParams = Bob.channelParams.copy(features = features, walletStaticPaymentBasepoint = Some(Helpers.getWalletPaymentBasepoint(wallet)))
-      (aliceParams, bobParams, ChannelVersion.STATIC_REMOTEKEY)
-    } else {
-      (Alice.channelParams, Bob.channelParams, ChannelVersion.STANDARD)
-    }
+    val aliceParams = Alice.channelParams
+      .modify(_.features).setToIf(channelVersion.hasAnchorOutputs)(Features(Set(ActivatedFeature(Features.StaticRemoteKey, FeatureSupport.Mandatory), ActivatedFeature(Features.AnchorOutputs, FeatureSupport.Optional))))
+      .modify(_.features).setToIf(channelVersion.hasStaticRemotekey)(Features(Set(ActivatedFeature(Features.StaticRemoteKey, FeatureSupport.Optional))))
```
t-bast, line 111: That shouldn't work...because `ChannelVersion.ANCHOR_OUTPUTS` returns true for `hasStaticRemoteKey` so you'll override the previous line and remove anchor outputs from the feature bits.

t-bast: But if you invert these two lines I think it should work. It's strange it tests pass with that code (unless I'm wrong there).

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/b/WaitForFundingSignedStateSpec.scala
```diff
@@ -39,13 +41,30 @@ class WaitForFundingSignedStateSpec extends TestKitBaseClass with FixtureAnyFunS
...
+      (Btc(5).toSatoshi, TestConstants.pushMsat)
+    } else {
+      (TestConstants.fundingSatoshis, TestConstants.pushMsat)
+    }
+
+    val setup = init(aliceNodeParams, bobNodeParams)
+
     import setup._
     val aliceInit = Init(Alice.channelParams.features)
     val bobInit = Init(Bob.channelParams.features)
```
t-bast, line 64:
```suggestion
    val aliceInit = Init(aliceParams.features)
    val bobInit = Init(bobParams.features)
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
t-bast: Ok I understand why it works whereas it shouldn't.
The `init` use the wrong feature bits (no anchor output because overridden by `static_remote_key`).
But in `INPUT_INIT_XXX` you directly inject the `channelVersion` which is correct (contains anchor outputs).
I'd still prefer it if we switch the order of these two lines so that the feature bits are consistent.

pm47: Agree, but this whole thing is not great. Actually I considered adding the "wumbo" feature handling here and didn't do it for that same reason.

t-bast: True, we should rework how we set feature bits in tests and do something cleaner and more flexible, but maybe after the release?

pm47: Check this out: d25e70d2d.

The only issue is that we may have duplicate features for `StaticRemoteKey` if both tags `AnchorOutputs` and `StaticRemoteKey` are active. This makes me think that maybe we should have defined `Features.activated` as a `Map[Feature->FeatureSupport]`.

t-bast (approved):

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
t-bast: > This makes me think that maybe we should have defined Features.activated as a Map[Feature->FeatureSupport].

Good idea, that's a refactoring we can do right after the release :+1:

PR #1716, opened 2021-03-04
Fix potential race condition in node-relay

t-bast:
We previously relied on `context.child` to check whether we already had a relay handler for a given `payment_hash`.

Unfortunately this can return an actor that is currently stopping itself.
When that happens our relay command can end up in the dead letters and the payment will not be relayed, nor be failed upstream.

We fix that by maintaining the list of current relay handlers in the `NodeRelayer` and removing them from the list _before_ stopping them. This is similar to what's done in the `MultiPartPaymentFSM`.

This bug has been seen in the wild by the Electrum team with an eclair trampoline node that they run. Phoenix is *not* affected by this, because this bug only happens when the sender uses trampoline with multiple incoming HTLCs, but *without* letting the trampoline node aggregate the incoming MPP.

In their case, the `NodeRelayer` receives for example 2 HTLCs with the same `payment_hash`, but each HTLC sets `total_amount_msat` to its `amount_msat` (so the trampoline node cannot aggregate them and must relay them separately).

If the first HTLC fails downstream slightly before the second HTLC reaches our node, the `NodeRelayer` thinks it has a handler for the `payment_hash` of the second HTLC but in fact it's stopping itself, and when we send it `NodeRelay.Relay` with the second HTLC the actor is gone and we don't correctly fail the HTLC upstream.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
@@ -16,8 +16,6 @@
 
 package fr.acinq.eclair.payment.relay
 
-import java.util.UUID
-
 import akka.actor.ActorRef
 import akka.actor.typed.Behavior
 import akka.actor.typed.eventstream.EventStream
@@ -41,6 +39,7 @@ import fr.acinq.eclair.router.{BalanceTooLow, RouteCalculation, RouteNotFound}
 import fr.acinq.eclair.wire._
 import fr.acinq.eclair.{CltvExpiry, Features, Logs, MilliSatoshi, NodeParams, nodeFee, randomBytes32}
 
+import java.util.UUID
 import scala.collection.immutable.Queue
 
 /**
@@ -60,13 +59,13 @@ object NodeRelay {
   private case class WrappedPaymentFailed(paymentFailed: PaymentFailed) extends Command
   // @formatter:on
 
-  def apply(nodeParams: NodeParams, router: ActorRef, register: ActorRef, relayId: UUID, paymentHash: ByteVector32, fsmFactory: FsmFactory = new FsmFactory): Behavior[Command] =
+  def apply(nodeParams: NodeParams, parent: akka.actor.typed.ActorRef[NodeRelayer.Command], router: ActorRef, register: ActorRef, relayId: UUID, paymentHash: ByteVector32, fsmFactory: FsmFactory = new FsmFactory): Behavior[Command] =
     Behaviors.setup { context =>
       Behaviors.withMdc(Logs.mdc(
         category_opt = Some(Logs.LogCategory.PAYMENT),
         parentPaymentId_opt = Some(relayId), // for a node relay, we use the same identifier for the whole relay itself, and the outgoing payment
         paymentHash_opt = Some(paymentHash))) {
-        new NodeRelay(nodeParams, router, register, relayId, paymentHash, context, fsmFactory)()
+        new NodeRelay(nodeParams, parent, router, register, relayId, paymentHash, context, fsmFactory)()
       }
     }
 
@@ -136,6 +135,7 @@ object NodeRelay {
  * see https://doc.akka.io/docs/akka/current/typed/style-guide.html#passing-around-too-many-parameters
  */
 class NodeRelay private(nodeParams: NodeParams,
+                        parent: akka.actor.typed.ActorRef[NodeRelayer.Command],
                         router: ActorRef,
                         register: ActorRef,
                         relayId: UUID,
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -38,33 +38,42 @@ object NodeRelayer {
...
   def mdc: Command => Map[String, String] = {
-    case c: Relay => Logs.mdc(
-      paymentHash_opt = Some(c.nodeRelayPacket.add.paymentHash))
+    case c: Relay => Logs.mdc(paymentHash_opt = Some(c.nodeRelayPacket.add.paymentHash))
+    case c: RelayComplete => Logs.mdc(paymentHash_opt = Some(c.paymentHash))
+    case _: GetPendingPayments => Logs.mdc()
   }
 
-  def apply(nodeParams: NodeParams, router: ActorRef, register: ActorRef): Behavior[Command] =
+  def apply(nodeParams: NodeParams, router: akka.actor.ActorRef, register: akka.actor.ActorRef, children: Map[ByteVector32, ActorRef[NodeRelay.Command]] = Map.empty): Behavior[Command] =
```
pm47, line 51: You scraped that comment:
> // we index children by paymentHash, not relayId, because there is no concept of individual payment on LN

It is arguably a bit cryptic but maybe this one is better?
```suggestion
    /**
   * @param children a map of current in-process payments, indexed by payment hash and purposefully *not* by payment id,
   *                 because that is how we aggregate payment parts.
   */
  def apply(nodeParams: NodeParams, router: akka.actor.ActorRef, register: akka.actor.ActorRef, children: Map[ByteVector32, ActorRef[NodeRelay.Command]] = Map.empty): Behavior[Command] =
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -38,33 +38,42 @@ object NodeRelayer {
...
+              case None =>
+                val relayId = UUID.randomUUID()
+                context.log.debug(s"spawning a new handler with relayId=$relayId")
+                val handler = context.spawn(NodeRelay.apply(nodeParams, context.self, router, register, relayId, paymentHash), relayId.toString)
+                context.log.debug("forwarding incoming htlc to new handler")
+                handler ! NodeRelay.Relay(nodeRelayPacket)
+                apply(nodeParams, router, register, children + (paymentHash -> handler))
+            }
+          case RelayComplete(childHandler, paymentHash) =>
+            childHandler ! NodeRelay.Stop
```
pm47, line 71:
```suggestion
            // we do a back-and-forth between parent and child before stopping the child to prevent a race condition
            // where the parent would send a message to a child that just stopped
            childHandler ! NodeRelay.Stop
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -249,16 +250,27 @@ class NodeRelay private(nodeParams: NodeParams,
...
-          Behaviors.stopped
+          stopping()
       }
     }
 
+  /**
+   * Once the downstream payment is settled (fulfilled or failed), we reject new upstream payments while we wait for our parent to stop us.
+   */
+  private def stopping(): Behavior[Command] = {
+    parent ! NodeRelayer.RelayComplete(context.self, paymentHash)
```
pm47, line 267: I wonder if that pattern of doing an action before calling `Behaviors.receive*` is a good practice, but it sure is convenient to handle transition to this state from several places. I believe it works provided that `def stopping()` isn't recursive.

t-bast: If you "flatten" the operations, `stopping` is really just a function call, so I think it's similar to the old `onTransition` hook we had for FSMs

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
t-bast: That previous comment was really confusing, your proposal is much better!

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1716/commits/0ddbc2b02f8988c3c8f6634432d7c917193b037e

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1716/commits/0ddbc2b02f8988c3c8f6634432d7c917193b037e

pm47 (approved):

PR #1718, opened 2021-03-04
Configurable anchor output max commitment feerate

t-bast:
It makes sense to allow node operators to configure the value they want to use as a maximum threshold for anchor outputs commitment tx feerate.

This allows node operators to raise this value when mempools start getting full in anticipation of a potential rise of the min-relay-fee.

This value can also be overridden for specific nodes.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -117,7 +117,7 @@ eclair {
       ratio-high = 10.0 // will allow remote fee rates as high as 10 times our local feerate
       // when using anchor outputs, we only need to use a commitment feerate that allows the tx to propagate: we will use CPFP to speed up confirmation if needed.
       // the following value is the maximum feerate we'll use for our commit tx (in sat/byte)
-      anchor-output-max-commit-feerate = 15
+      anchor-output-max-commit-feerate = 10
     }
     override-feerate-tolerance = [ // optional per-node feerate tolerance
       #  {
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -115,13 +115,17 @@ eclair {
     feerate-tolerance {
       ratio-low = 0.5 // will allow remote fee rates as low as half our local feerate
       ratio-high = 10.0 // will allow remote fee rates as high as 10 times our local feerate
+      // when using anchor outputs, we only need to use a commitment feerate that allows the tx to propagate: we will use CPFP to speed up confirmation if needed.
+      // the following value is the maximum feerate we'll use for our commit tx (in sat/byte)
+      anchor-output-max-commit-feerate = 15
     }
     override-feerate-tolerance = [ // optional per-node feerate tolerance
       #  {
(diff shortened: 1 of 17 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -115,13 +115,17 @@ eclair {
...
       ratio-low = 0.5 // will allow remote fee rates as low as half our local feerate
       ratio-high = 10.0 // will allow remote fee rates as high as 10 times our local feerate
+      // when using anchor outputs, we only need to use a commitment feerate that allows the tx to propagate: we will use CPFP to speed up confirmation if needed.
+      // the following value is the maximum feerate we'll use for our commit tx (in sat/byte)
+      anchor-output-max-commit-feerate = 15
     }
     override-feerate-tolerance = [ // optional per-node feerate tolerance
       #  {
-      #    nodeid = "02aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+      #    nodeid = "02aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
pm47, line 124: 👓

eclair-core/src/main/resources/reference.conf
```diff
@@ -115,13 +115,17 @@ eclair {
     feerate-tolerance {
       ratio-low = 0.5 // will allow remote fee rates as low as half our local feerate
       ratio-high = 10.0 // will allow remote fee rates as high as 10 times our local feerate
+      // when using anchor outputs, we only need to use a commitment feerate that allows the tx to propagate: we will use CPFP to speed up confirmation if needed.
+      // the following value is the maximum feerate we'll use for our commit tx (in sat/byte)
+      anchor-output-max-commit-feerate = 15
```
pm47, line 120: Did you mean to change the default value from 10 to 15 sat/byte?

t-bast: Since the mempool purged at 11 sat/byte last week, I think it would make sense, wouldn't it?
Or should we leave it at 10 sat/byte by default and let node operators change it if necessary?

t-bast: Updated in https://github.com/ACINQ/eclair/pull/1718/commits/1d6c299d01c571fc3df30b9e306cb73e8a887aa8

pm47: I'd leave it to 11 sat/byte for now, and if we change the default let's make a PR just for that.

t-bast: Why 11 instead of 10 which is what lnd uses (and is more "round")?

pm47: Because! https://www.youtube.com/watch?v=uMSV4OteqBE

No really, I have no idea why I typed 11, I meant *10* sat/byte.

pm47 (dismissed):

pm47 (approved):

PR #1721, opened 2021-03-05
Clarify commit tx fee anchor cost

t-bast:
Since anchor outputs, we not only deduce the commit tx fee from the funder's main output but the cost of the anchors as well.
We rename the function that does that for more clarity as it came up in multiple reviews as something that was a bit confusing.

The change is split in two commits to make it easier to verify that there's no regression:

- the first commit renames the function
- the second commit adapts places where we actually only want the commit tx fee

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
@@ -226,24 +226,30 @@ object Transactions {
   /** Fee for an un-trimmed HTLC. */
   def htlcOutputFee(feeratePerKw: FeeratePerKw, commitmentFormat: CommitmentFormat): MilliSatoshi = weight2feeMsat(feeratePerKw, commitmentFormat.htlcOutputWeight)
 
+  /** Fee paid by the commit tx (depends on which HTLCs will be trimmed). */
+  def commitTxFeeMsat(dustLimit: Satoshi, spec: CommitmentSpec, commitmentFormat: CommitmentFormat): MilliSatoshi = {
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -226,28 +226,33 @@ object Transactions {
...
+    weight2feeMsat(spec.feeratePerKw, weight)
+  }
+
+  /**
+   * While on-chain amounts are generally computed in Satoshis (since this is the smallest on-chain unit), it may be
+   * useful in some cases to calculate it in MilliSatoshi to avoid rounding issues.
+   * If you are adding multiple fees together for example, you should always add them in MilliSatoshi and then round
+   * down to Satoshi.
+   */
+  def toDeduceFromFunderOutputMsat(dustLimit: Satoshi, spec: CommitmentSpec, commitmentFormat: CommitmentFormat): MilliSatoshi = {
```
pm47, line 243: I think I would have gone with `commitTxFeeMsat` and `commitTxTotalCostMsat`
```suggestion
  def commitTxTotalCostMsat(dustLimit: Satoshi, spec: CommitmentSpec, commitmentFormat: CommitmentFormat): MilliSatoshi = {
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1721/commits/856f427fd48ec80e7727c43386af72508a8cc796

pm47 (approved):

PR #1724, opened 2021-03-08
Add metadata to local_channels table

pm47:
(First commit is a refactor/cleanup and can be reviewed independently).

Here is the rationale for implementing channel metadata as additional
columns in the `local_channels` table of the `channels` db, as opposed
to a dedicated `channel_metadata` table of a `audit` db:

1) There is a migration to do (in the `local_channels` table, no less!),
but it's just a table migration, as opposed to a data migration, if we
had to populate a new table in a separate database.
2) We don't need to worry about creating a new metadata line when a new
channel is created (compared to doing add-or-update stuff). It's only
_updating_ optional columns in a best-effort manner.
3) We don't need to worry about inconsistencies between two tables
located in two separated databases (that's a big one).
4) We may want to use the metadata during operations, not just for
audit purposes. For example to close channels that have stayed unused
for a long time.
5) The audit db is an append-only log of events and shouldn't be used
for anything else. There is no `UPDATE` sql statement in
`*AuditDb.scala`. The `channel_metadata` would break that heuristic.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/DbEventHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/DbEventHandler.scala
@@ -32,7 +32,8 @@ import fr.acinq.eclair.payment._
  */
 class DbEventHandler(nodeParams: NodeParams) extends Actor with ActorLogging {
 
-  val db = nodeParams.db.audit
+  val auditDb: AuditDb = nodeParams.db.audit
+  val channelsDb: ChannelsDb = nodeParams.db.channels
 
   context.system.eventStream.subscribe(self, classOf[PaymentEvent])
   context.system.eventStream.subscribe(self, classOf[NetworkFeePaid])
@@ -46,7 +47,8 @@ class DbEventHandler(nodeParams: NodeParams) extends Actor with ActorLogging {
       PaymentMetrics.PaymentAmount.withTag(PaymentTags.Direction, PaymentTags.Directions.Sent).record(e.recipientAmount.truncateToSatoshi.toLong)
       PaymentMetrics.PaymentFees.withTag(PaymentTags.Direction, PaymentTags.Directions.Sent).record(e.feesPaid.truncateToSatoshi.toLong)
       PaymentMetrics.PaymentParts.withTag(PaymentTags.Direction, PaymentTags.Directions.Sent).record(e.parts.length)
-      db.add(e)
+      auditDb.add(e)
+      e.parts.foreach(p => channelsDb.updateChannelMeta(p.toChannelId, e))
 
     case _: PaymentFailed =>
       PaymentMetrics.PaymentFailed.withTag(PaymentTags.Direction, PaymentTags.Directions.Sent).increment()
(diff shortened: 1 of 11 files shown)


pm47:
@sstone @t-bast the PR isn't finished but I'd like to have your feedback before I go further. I did start by implementing this in the `audit` db, before changing for this approach.

t-bast:
That sounds reasonable, these are good reasons to go down that route.
As long as we use separate, optional columns and not modify the core `ChannelData` structures let's try and see where it takes us.

eclair-core/src/main/scala/fr/acinq/eclair/db/DbEventHandler.scala
```diff
@@ -65,49 +74,68 @@ class Auditor(nodeParams: NodeParams) extends Actor with ActorLogging {
           PaymentMetrics.PaymentParts.withTag(PaymentTags.Direction, PaymentTags.Directions.Sent).record(outgoing.length)
         case _: ChannelPaymentRelayed =>
       }
-      db.add(e)
+      auditDb.add(e)
```
t-bast, line 77: Why don't you update the channel meta here as well?

eclair-core/src/main/scala/fr/acinq/eclair/db/DbEventHandler.scala
```diff
@@ -65,49 +74,68 @@ class Auditor(nodeParams: NodeParams) extends Actor with ActorLogging {
...
       e match {
         case ChannelStateChanged(_, channelId, _, remoteNodeId, WAIT_FOR_FUNDING_LOCKED, NORMAL, Some(commitments: Commitments)) =>
           ChannelMetrics.ChannelLifecycleEvents.withTag(ChannelTags.Event, ChannelTags.Events.Created).increment()
-          db.add(ChannelLifecycleEvent(channelId, remoteNodeId, commitments.capacity, commitments.localParams.isFunder, !commitments.announceChannel, "created"))
+          val event = ChannelLifecycleEvent.EventType.Created
+          auditDb.add(ChannelLifecycleEvent(channelId, remoteNodeId, commitments.capacity, commitments.localParams.isFunder, !commitments.announceChannel, event))
+          channelsDb.updateChannelMeta(channelId, event)
         case ChannelStateChanged(_, _, _, _, WAIT_FOR_INIT_INTERNAL, _, _) =>
+        case ChannelStateChanged(_, channelId, _, _, OFFLINE, SYNCING, _) =>
+          channelsDb.updateChannelMeta(channelId, ChannelLifecycleEvent.EventType.Connected)
```
t-bast, line 99: We may want to have this event in the `auditDb` as well, don't we?

eclair-core/src/test/scala/fr/acinq/eclair/db/SqliteChannelsDbSpec.scala
```diff
@@ -69,7 +73,64 @@ class SqliteChannelsDbSpec extends AnyFunSuite {
...
+      db.updateChannelMeta(channel1.channelId, ChannelLifecycleEvent.EventType.Closed(null))
+      assert(getTimestamp(channel1.channelId, "closed_timestamp").nonEmpty)
+
+      // make sure all metadata are still empty for channel 2
+      assert(getTimestamp(channel2.channelId, "created_timestamp").isEmpty)
+      assert(getTimestamp(channel2.channelId, "last_payment_sent_timestamp").isEmpty)
+      assert(getTimestamp(channel2.channelId, "last_payment_received_timestamp").isEmpty)
+      assert(getTimestamp(channel2.channelId, "last_connected_timestamp").isEmpty)
+      assert(getTimestamp(channel2.channelId, "closed_timestamp").isEmpty)
+
```
t-bast, line 129: nit: empty line

eclair-core/src/main/scala/fr/acinq/eclair/db/DbEventHandler.scala
pm47: I don't think so, because the risk of spam is too high imo.

PR #1727, opened 2021-03-09
Trampoline fast fail

btcontract:
Fail relayed trampoline payment if we see that `payloadOut.outgoingCltv` is below chain height. It's not our problem if it is but it makes sense to not risk our funds for a payment which will be failed by rational receivers.


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -89,6 +89,8 @@ object NodeRelay {
       Some(TrampolineFeeInsufficient)
     } else if (upstream.expiryIn - payloadOut.outgoingCltv < nodeParams.expiryDelta) {
       Some(TrampolineExpiryTooSoon)
+    } else if (CltvExpiry(nodeParams.currentBlockHeight) > payloadOut.outgoingCltv) {
+      Some(IncorrectOrUnknownPaymentDetails(upstream.adds.head.amountMsat, nodeParams.currentBlockHeight))
```
t-bast, line 93: I don't think this should be an `IncorrectOrUnknownPaymentDetails`, this is really a final recipient error for payments that don't match what's in the payment database...
Why didn't you use `TrampolineExpiryTooSoon`?

t-bast (commented):
That's a small enough change, definitely worth it :+1:

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
btcontract: Mainly because it has a chain height hint, but if this breaks the rules too much `TrampolineExpiryTooSoon` seems good enough.

t-bast (approved):
Thanks!

PR #1729, opened 2021-03-12
Fix API regression

t-bast:
We incorrectly applied error handlers at each sub-route instead of applying it after grouping all sub-routes together. The result was that only `getinfo` (the first one tried) could actually be called.

The tests were changed to directly call the sub-routes so they didn't notice this issue. I fixed that by adding tests that failed with the previous code by simply using the top-level route.

I thought I had tested this E2E before merging #1690 but my (custom, poorly written) scripts in fact didn't correctly re-source my `.bash_aliases` so I was in fact testing `master` :facepalm: 

cc @tompro

### DIFF
--- a/eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
+++ b/eclair-node/src/main/scala/fr/acinq/eclair/api/Service.scala
@@ -47,9 +47,12 @@ trait Service extends EclairDirectives with WebSocket with Node with Channel wit
   implicit val mat: Materializer
 
   /**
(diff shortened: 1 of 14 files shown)


pm47 (approved):

tompro:
So routing was canceled after the first miss? Really sorry just tested the getinfo route with this commit as I was hunting for the json error formatting. Need to get this E2E test setup running to prevent my custom poor manual testing and be more careful next time.

t-bast:
No worries, my tests should have caught that and failed to do so as well!

PR #1731, opened 2021-03-18
Reorganize internal codecs (rework-closing-commit-published)

pm47:
Builds on #1728.

The goal is to separate internal codecs from protocol codecs, and to reduce risk during migrations.

Commits should be read one by one.


pm47:
Superseded by #1732.

PR #1732, opened 2021-03-19
Reorganize internal codecs (master)

pm47:
This PR is the same as #1731, but applied on master instead of on top of #1728.

The goal is to ensure maximum safety when migrating data.

To achieve this, we strictly segregate each codec version (to make sure
that we don't accidentally mix codec versions), while still letting
tests access each unitary codecs (which using inner `private classes`
would have prevented). Relevant tests only need to be moved to the
appropriate package.

The package structure is now:

```
wire
 |
 `-- internal
 |       |
 |       `-- channel
 |       |      |
 |       |      `-- version0
 |       |      |      |
 |       |      |      `-- ChannelCodecs0
 |       |      |
 |       |      `-- version1
 |       |      |      |
 |       |      |      `-- ChannelCodecs1
 |       |      |
 |       |      `-- ChannelCodecs
 |       |
 |       `-- CommandCodecs
 |
 `-- others
```


eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecs.scala
```diff
@@ -0,0 +1,87 @@
...
+       }
+
+       import Codecs._
+
+       val STATE_Codec: Codec[STATE] = ...
+ * }}}
+ *
+ * Notice that the outer class has a visibility restricted to package [[channel]], while the inner class has a
+ * visibility restricted to package [[version0]]. This guarantees that we strictly segregate each codec version,
+ * while still allowing unitary testing.
```
t-bast, line 52: Great trick :+1:

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecs.scala
```diff
@@ -0,0 +1,87 @@
...
+   */
+  val stateDataCodec: Codec[HasCommitments] = discriminated[HasCommitments].by(byte)
+    .typecase(1, discriminated[HasCommitments].by(uint16)
+      .typecase(0x20, ChannelCodecs1.DATA_WAIT_FOR_FUNDING_CONFIRMED_Codec)
+      .typecase(0x21, ChannelCodecs1.DATA_WAIT_FOR_FUNDING_LOCKED_Codec)
+      .typecase(0x22, ChannelCodecs1.DATA_NORMAL_Codec)
+      .typecase(0x23, ChannelCodecs1.DATA_SHUTDOWN_Codec)
+      .typecase(0x24, ChannelCodecs1.DATA_NEGOTIATING_Codec)
+      .typecase(0x25, ChannelCodecs1.DATA_CLOSING_Codec)
+      .typecase(0x26, ChannelCodecs1.DATA_WAIT_FOR_REMOTE_PUBLISH_FUTURE_COMMITMENT_Codec))
```
t-bast, line 74: Should we move this inside `ChannelCodecs1`? And same for the ones below (which should move to `ChannelCodecs0`?

That would allow us to have these versioned `DATA_*` codecs go inside the inner `private[versionX] object Codecs` to hide them even more.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version1/ChannelCodecs1.scala
```diff
@@ -0,0 +1,266 @@
...
+import fr.acinq.bitcoin.{ByteVector32, OutPoint, Transaction, TxOut}
+import fr.acinq.eclair.MilliSatoshi
+import fr.acinq.eclair.channel._
+import fr.acinq.eclair.crypto.ShaChain
+import fr.acinq.eclair.transactions.Transactions._
+import fr.acinq.eclair.transactions.{CommitmentSpec, DirectedHtlc, IncomingHtlc, OutgoingHtlc}
+import fr.acinq.eclair.wire.CommonCodecs.{bytes32, bytes64, mapCodec, publicKey, setCodec, varintoverflow, _}
+import fr.acinq.eclair.wire.LightningMessageCodecs.{channelAnnouncementCodec, channelUpdateCodec, closingSignedCodec, commitSigCodec, fundingCreatedCodec, fundingLockedCodec, fundingSignedCodec, lightningMessageCodec, shutdownCodec, updateAddHtlcCodec, _}
+import fr.acinq.eclair.wire.UpdateMessage
+import scodec.codecs.{bits, bool, byteAligned, bytes, discriminated, either, int64, listOfN, optional, provide, uint16, uint32, variableSizeBytesLong, _}
```
t-bast, line 13: nit:
```suggestion
import fr.acinq.eclair.wire.CommonCodecs._
import fr.acinq.eclair.wire.LightningMessageCodecs._
import fr.acinq.eclair.wire.UpdateMessage
import scodec.codecs._
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version1/ChannelCodecs1.scala
```diff
@@ -0,0 +1,266 @@
...
+import scodec.{Attempt, Codec}
+
+private[channel] object ChannelCodecs1 {
+
+  private[version1] object Codecs {
+
+    /**
+     * All LN protocol message must be stored as length-delimited, because they may have arbitrary trailing data
+     */
+    def lengthDelimited[T](codec: Codec[T]): Codec[T] = variableSizeBytesLong(varintoverflow, codec)
```
t-bast, line 23: Shouldn't this one be in the `CommonCodecs`?

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecs.scala
pm47: I don't think we should, because this would prevent us from having a global view of the migration path. It entirely depends on the order of the codecs and is fully described in a single place which I think is nice.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version1/ChannelCodecs1.scala
pm47: Weird, how come this isn't fixed by _optimize imports_?

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version1/ChannelCodecs1.scala
pm47: Yes this one could, but only that one I think.

pm47: Done with ee9050d46.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecs.scala
t-bast: Note that what I'm suggesting is:

```
val stateDataCodec: Codec[HasCommitments] = discriminated[HasCommitments].by(byte)
    .typecase(1, ChannelCodecs1.sateDataCodec)
    .typecase(0, ChannelCodecs0.sateDataCodec)
```

This ensures the order of codec that we rely on works, and leaves details of each migration internal to each versioned package. I think it's cleaner as we'll require each legacy codec to only expose a `stateDataCodec`, unless I'm missing something?

pm47: No, you're right, I like it.

pm47: Done with 719b7fb60.

t-bast (approved):
:+1:

PR #1733, opened 2021-03-22
Add high-level architecture diagrams

t-bast:
It can be useful for newcomers to have a high-level view of the main components in eclair. This will help them quickly find where they should start digging into actual code to achieve what they want.

We're leaving a lot of details out to ensure this document stays up-to-date and doesn't need to completely change every time we slightly rework internal details of our architecture.

### DIFF
--- a/docs/Architecture.md
+++ b/docs/Architecture.md
@@ -30,42 +30,78 @@ The entry point for `eclair-core` is in `Setup.scala`, where we start the actor 
 
 ## Actor system overview
 
-Here is a high-level view of the main actors in the system:
+Here is a high-level view of the hierarchy of some of the main actors in the system:
 
 ```ascii
-                                                             +--------+
-                          +--------------------------------->| Router |<------------------------------------+
-                          |                                  +--------+                                     |
-                          |                                                                                 |
-                          |           +---------+                                                           |
-                          | +-------->| Channel |<----------------+                                         |
-                          v |         +---------+                 |                                         |
-                        +------+      +---------+                 |                                         v
-      +---------------->| Peer |----->| Channel |<----------------+                                +------------------+
-      |                 +------+      +---------+                 |                    +-----------| PaymentInitiator |
-      |                     |         +---------+                 |                    |           +------------------+
-      |                     +-------->| Channel |<----------------+                    |
-      |                               +---------+                 |                    v
-+-------------+                                              +----------+         +---------+
-| Switchboard |                                              | Register |<--------| Relayer |
-+-------------+                                              +----------+         +---------+
-      |                               +---------+                 |                    |
-      |                     +-------->| Channel |<----------------+                    |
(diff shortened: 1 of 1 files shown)


docs/Architecture.md
```diff
@@ -0,0 +1,200 @@
...
+                          v |         +---------+                 |                                         |
+                        +------+      +---------+                 |                                         v
+      +---------------->| Peer |----->| Channel |<----------------+                                +------------------+
+      |                 +------+      +---------+                 |                    +-----------| PaymentInitiator |
+      |                     |         +---------+                 |                    |           +------------------+
+      |                     +-------->| Channel |<----------------+                    |
+      |                               +---------+                 |                    v
++-------------+                                              +----------+         +---------+
+| Switchboard |                                              | Register |<--------| Relayer |
++-------------+                                              +----------+         +---------+
```
pm47, line 51: Channels also directly send messages to the `Relayer`. On the other hand, the `PaymentInitiator` sends messages to the `Register` (but receive messages from the  `Relayer`.

Reading further I noticed that you detail the send/receive/relay scenarii below. Maybe in this diagram we should represent the actor hierarchy (it is a bit unclear what the arrows represent).

I wonder if it would be useful to also mention the `PeerConnection`/`TransportHandler`.

t-bast: I tried to keep this high-level, so that a newcomer can easily locate an entry point in the code based on what he's looking for. That's why I didn't mention `PeerConnection`/`TransportHandler`, if you're looking for them you'll figure out from the diagram that you should look into the `Switchboard` or the `Peer` and then the code will lead you to these lower-level actors.

I tried to show a simple message flow, but as you noticed it's actually incorrect and when you dive into the details it's always a bit more complex than we'd like...I'll try changing this diagram to expose the actor hierarchy as you suggest, it's probably easier.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1733/commits/5542fdf84852e89869db6f97759acb11e629d32d, it's a bit hard to decide where to stop to keep this high-level but detailed enough to give a good overall view...

docs/Architecture.md
```diff
@@ -0,0 +1,234 @@
...
++-------------+
+      |                               +---------+
+      |                     +-------->| Channel |
+      |                     |         +---------+
+      |                 +------+      +---------+
+      +---------------->| Peer |----->| Channel |
+                        +------+      +---------+
+                            |         +---------+
+                            +-------->| Channel |
+                                      +---------
```
pm47, line 56:
```suggestion
                                      +---------+
```

pm47 (commented):
I think it looks pretty good, just one nit.

docs/Architecture.md
t-bast: :detective:

pm47 (approved):

PR #1734, opened 2021-03-22
Clarify some comments and add tests

t-bast:
This commit clarifies some parts of the code on which we regularly have questions during pull requests.

We also add a test for an edge case in shutdown that was correctly handled, but not properly tested, to ensure non-regression.


pm47 (dismissed):

t-bast:
Rebased

pm47 (approved):

PR #1736, opened 2021-03-22
Move protocol codecs to their own package

t-bast:
This is a follow-up for #1732 

The codecs in the `protocol` package will always be backwards-compatible, as they are defined in the spec.


pm47 (approved):
LGTM.

We could also have renamed the parent package to `fr.acinq.eclair.types` as internal codecs do not go on the wire. But `types` is a bit generic. Well...

PR #1737, opened 2021-03-23
Add a globalbalance API call

pm47:
For now it just gives detailed info on the current pending mutual closes.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -17,11 +17,15 @@
 package fr.acinq.eclair
 
 import akka.actor.ActorRef
+import akka.actor.typed.scaladsl.AskPattern.Askable
+import akka.actor.typed.scaladsl.adapter.ClassicSchedulerOps
 import akka.pattern._
 import akka.util.Timeout
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{ByteVector32, ByteVector64, Crypto, Satoshi}
 import fr.acinq.eclair.TimestampQueryFilters._
+import fr.acinq.eclair.balance.CheckBalance.GlobalBalance
+import fr.acinq.eclair.balance.{BalanceActor, ChannelsListener}
 import fr.acinq.eclair.blockchain.OnChainBalance
 import fr.acinq.eclair.blockchain.bitcoind.BitcoinCoreWallet
 import fr.acinq.eclair.blockchain.bitcoind.BitcoinCoreWallet.WalletTransaction
@@ -39,12 +43,13 @@ import fr.acinq.eclair.payment.send.PaymentInitiator.{SendPayment, SendPaymentTo
 import fr.acinq.eclair.router.Router._
 import fr.acinq.eclair.router.{NetworkStats, RouteCalculation, Router}
 import fr.acinq.eclair.wire.protocol._
+import grizzled.slf4j.Logging
 import scodec.bits.ByteVector
 
 import java.nio.charset.StandardCharsets
 import java.util.UUID
 import scala.concurrent.duration._
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.{ExecutionContext, Future, Promise}
 import scala.reflect.ClassTag
 
 case class GetInfoResponse(version: String, nodeId: PublicKey, alias: String, color: String, features: Features, chainHash: ByteVector32, network: String, blockHeight: Int, publicAddresses: Seq[NodeAddress], instanceId: String)
@@ -59,9 +64,9 @@ case class VerifiedMessage(valid: Boolean, publicKey: PublicKey)
 
 object TimestampQueryFilters {
   /** We use this in the context of timestamp filtering, when we don't need an upper bound. */
-  val MaxEpochMilliseconds = Duration.fromNanos(Long.MaxValue).toMillis
+  val MaxEpochMilliseconds: Long = Duration.fromNanos(Long.MaxValue).toMillis
 
-  def getDefaultTimestampFilters(from_opt: Option[Long], to_opt: Option[Long]) = {
+  def getDefaultTimestampFilters(from_opt: Option[Long], to_opt: Option[Long]): TimestampQueryFilters = {
     // NB: we expect callers to use seconds, but internally we use milli-seconds everywhere.
     val from = from_opt.getOrElse(0L).seconds.toMillis
     val to = to_opt.map(_.seconds.toMillis).getOrElse(MaxEpochMilliseconds)
@@ -149,12 +154,14 @@ trait Eclair {
 
   def onChainTransactions(count: Int, skip: Int): Future[Iterable[WalletTransaction]]
 
+  def globalBalance()(implicit timeout: Timeout): Future[GlobalBalance]
+
   def signMessage(message: ByteVector): SignedMessage
 
   def verifyMessage(message: ByteVector, recoverableSignature: ByteVector): VerifiedMessage
 }
 
-class EclairImpl(appKit: Kit) extends Eclair {
+class EclairImpl(appKit: Kit) extends Eclair with Logging {
 
   implicit val ec: ExecutionContext = appKit.system.dispatcher
 
@@ -427,6 +434,14 @@ class EclairImpl(appKit: Kit) extends Eclair {
 
   override def usableBalances()(implicit timeout: Timeout): Future[Iterable[UsableBalance]] =
     (appKit.relayer ? GetOutgoingChannels()).mapTo[OutgoingChannels].map(_.channels.map(_.toUsableBalance))
+
+  override def globalBalance()(implicit timeout: Timeout): Future[GlobalBalance] = {
+    for {
+      ChannelsListener.GetChannelsResponse(channels) <- appKit.channelsListener.ask(ref => ChannelsListener.GetChannels(ref))(timeout, appKit.system.scheduler.toTyped)
+      globalBalance_try <- appKit.balanceActor.ask(res => BalanceActor.GetGlobalBalance(res, channels))(timeout, appKit.system.scheduler.toTyped)
+      globalBalance <- Promise[GlobalBalance]().complete(globalBalance_try).future
+    } yield globalBalance
+  }
 
   override def signMessage(message: ByteVector): SignedMessage = {
     val bytesToSign = SignedMessage.signedBytes(message)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -25,11 +25,12 @@ import akka.util.Timeout
 import com.softwaremill.sttp.okhttp.OkHttpFutureBackend
 import fr.acinq.bitcoin.{Block, ByteVector32, Satoshi}
 import fr.acinq.eclair.Setup.Seeds
-import fr.acinq.eclair.blockchain._
+import fr.acinq.eclair.balance.{BalanceActor, ChannelsListener}
 import fr.acinq.eclair.blockchain.bitcoind.rpc.{BasicBitcoinJsonRPCClient, BatchingBitcoinJsonRPCClient, ExtendedBitcoinClient}
 import fr.acinq.eclair.blockchain.bitcoind.zmq.ZMQActor
 import fr.acinq.eclair.blockchain.bitcoind.{BitcoinCoreWallet, ZmqWatcher}
-import fr.acinq.eclair.blockchain.fee._
+import fr.acinq.eclair.blockchain.fee.{ConstantFeeProvider, _}
+import fr.acinq.eclair.blockchain.{EclairWallet, _}
 import fr.acinq.eclair.channel.{Channel, Register}
 import fr.acinq.eclair.crypto.WeakEntropyPool
 import fr.acinq.eclair.crypto.keymanager.{LocalChannelKeyManager, LocalNodeKeyManager}
@@ -194,6 +195,7 @@ class Setup(val datadir: File,
       tcpBound = Promise[Done]()
       routerInitialized = Promise[Done]()
       postRestartCleanUpInitialized = Promise[Done]()
+      channelsListenerReady = Promise[Done]()
 
       defaultFeerates = {
         val confDefaultFeerates = FeeratesPerKB(
@@ -249,6 +251,9 @@ class Setup(val datadir: File,
       wallet = new BitcoinCoreWallet(bitcoin)
       _ = wallet.getReceiveAddress().map(address => logger.info(s"initial wallet address=$address"))
 
+      channelsListener = system.spawn(ChannelsListener(channelsListenerReady), name = "channels-listener")
+      _ <- channelsListenerReady.future
+
       _ = if (config.getBoolean("file-backup.enabled")) {
         nodeParams.db match {
           case fileBackup: FileBackup if config.getBoolean("file-backup.enabled") =>
@@ -283,6 +288,8 @@ class Setup(val datadir: File,
       paymentInitiator = system.actorOf(SimpleSupervisor.props(PaymentInitiator.props(nodeParams, PaymentInitiator.SimplePaymentFactory(nodeParams, router, register)), "payment-initiator", SupervisorStrategy.Restart))
       _ = for (i <- 0 until config.getInt("autoprobe-count")) yield system.actorOf(SimpleSupervisor.props(Autoprobe.props(nodeParams, router, paymentInitiator), s"payment-autoprobe-$i", SupervisorStrategy.Restart))
 
+      balanceActor = system.spawn(BalanceActor(nodeParams.db, extendedBitcoinClient, channelsListener, nodeParams.balanceCheckInterval), name = "balance-actor")
+
       kit = Kit(
         nodeParams = nodeParams,
         system = system,
@@ -294,6 +301,8 @@ class Setup(val datadir: File,
         switchboard = switchboard,
         paymentInitiator = paymentInitiator,
         server = server,
+        channelsListener = channelsListener,
+        balanceActor = balanceActor,
         wallet = wallet)
 
       zmqBlockTimeout = after(5 seconds, using = system.scheduler)(Future.failed(BitcoinZMQConnectionTimeoutException))
@@ -360,6 +369,8 @@ case class Kit(nodeParams: NodeParams,
                switchboard: ActorRef,
                paymentInitiator: ActorRef,
                server: ActorRef,
+               channelsListener: typed.ActorRef[ChannelsListener.Command],
+               balanceActor: typed.ActorRef[BalanceActor.Command],
                wallet: EclairWallet)
 
 object Kit {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
@@ -1,0 +1,139 @@
+package fr.acinq.eclair.balance
+
+import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
+import akka.actor.typed.{ActorRef, Behavior}
+import fr.acinq.bitcoin.ByteVector32
+import fr.acinq.eclair.balance.BalanceActor._
+import fr.acinq.eclair.balance.CheckBalance.GlobalBalance
+import fr.acinq.eclair.balance.Monitoring.{Metrics, Tags}
+import fr.acinq.eclair.blockchain.bitcoind.rpc.ExtendedBitcoinClient
+import fr.acinq.eclair.blockchain.bitcoind.rpc.ExtendedBitcoinClient.Utxo
+import fr.acinq.eclair.channel.HasCommitments
+import fr.acinq.eclair.db.Databases
+import grizzled.slf4j.Logger
+import org.json4s.JsonAST.JInt
+
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
+
+object BalanceActor {
+
+  // @formatter:off
+  sealed trait Command
+  private final case object TickBalance extends Command
+  final case class GetGlobalBalance(replyTo: ActorRef[Try[GlobalBalance]], channels: Map[ByteVector32, HasCommitments]) extends Command
+  private final case class WrappedChannels(wrapped: ChannelsListener.GetChannelsResponse) extends Command
+  private final case class WrappedGlobalBalance(wrapped: Try[GlobalBalance]) extends Command
+  private final case class WrappedUtxoInfo(wrapped: Try[UtxoInfo]) extends Command
+  // @formatter:on
+
+  def apply(db: Databases, extendedBitcoinClient: ExtendedBitcoinClient, channelsListener: ActorRef[ChannelsListener.GetChannels], interval: FiniteDuration)(implicit ec: ExecutionContext): Behavior[Command] = {
+    Behaviors.setup { context =>
+      Behaviors.withTimers { timers =>
+        timers.startTimerWithFixedDelay(TickBalance, interval)
+        new BalanceActor(context, db, extendedBitcoinClient, channelsListener).apply(refBalance_opt = None)
+      }
+    }
+  }
+
+  final case class UtxoInfo(utxos: Seq[Utxo], ancestorCount: Map[ByteVector32, Long])
+
+  def checkUtxos(extendedBitcoinClient: ExtendedBitcoinClient)(implicit ec: ExecutionContext): Future[UtxoInfo] = {
+
+    def getUnconfirmedAncestorCount(utxo: Utxo): Future[(ByteVector32, Long)] = extendedBitcoinClient.rpcClient.invoke("getmempoolentry", utxo.txid).map(json => {
+      val JInt(ancestorCount) = json \ "ancestorcount"
+      (utxo.txid, ancestorCount.toLong)
+    }).recover {
+      case ex: Throwable =>
+        // a bit hackish but we don't need the actor context for this simple log
+        val log = Logger(classOf[BalanceActor])
+        log.warn(s"could not retrieve unconfirmed ancestor count for txId=${utxo.txid} amount=${utxo.amount}:", ex)
+        (utxo.txid, 0)
+    }
+
+    def getUnconfirmedAncestorCountMap(utxos: Seq[Utxo]): Future[Map[ByteVector32, Long]] = Future.sequence(utxos.filter(_.confirmations == 0).map(getUnconfirmedAncestorCount)).map(_.toMap)
+
+    for {
+      utxos <- extendedBitcoinClient.listUnspent()
+      ancestorCount <- getUnconfirmedAncestorCountMap(utxos)
+    } yield UtxoInfo(utxos, ancestorCount)
+  }
+
+}
+
+private class BalanceActor(context: ActorContext[Command],
+                           db: Databases,
+                           extendedBitcoinClient: ExtendedBitcoinClient,
+                           channelsListener: ActorRef[ChannelsListener.GetChannels])(implicit ec: ExecutionContext) {
+
+  private val log = context.log
+
+  def apply(refBalance_opt: Option[GlobalBalance]): Behavior[Command] = Behaviors.receiveMessage {
+    case TickBalance =>
+      log.debug("checking balance...")
+      channelsListener ! ChannelsListener.GetChannels(context.messageAdapter[ChannelsListener.GetChannelsResponse](WrappedChannels))
+      context.pipeToSelf(checkUtxos(extendedBitcoinClient))(WrappedUtxoInfo)
+      Behaviors.same
+    case WrappedChannels(res) =>
+      context.pipeToSelf(CheckBalance.computeGlobalBalance(res.channels, db, extendedBitcoinClient))(WrappedGlobalBalance)
+      Behaviors.same
+    case WrappedGlobalBalance(res) =>
+      res match {
+        case Success(result) =>
+          log.info("current balance: total={} onchain.confirmed={} onchain.unconfirmed={} offchain={}", result.total.toDouble, result.onChain.confirmed.toDouble, result.onChain.unconfirmed.toDouble, result.offChain.total.toDouble)
+          log.debug("current balance details : {}", result)
+          Metrics.GlobalBalance.withoutTags().update(result.total.toMilliBtc.toDouble)
+          Metrics.GlobalBalanceDetailed.withTag(Tags.BalanceType, Tags.BalanceTypes.OnchainConfirmed).update(result.onChain.confirmed.toMilliBtc.toLong)
+          Metrics.GlobalBalanceDetailed.withTag(Tags.BalanceType, Tags.BalanceTypes.OnchainUnconfirmed).update(result.onChain.unconfirmed.toMilliBtc.toLong)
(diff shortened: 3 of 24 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -53,15 +57,31 @@ case class AuditResponse(sent: Seq[PaymentSent], received: Seq[PaymentReceived],
 
 case class TimestampQueryFilters(from: Long, to: Long)
 
+case class MutualCloseStatus(unpublished: Satoshi, unconfirmed: Satoshi, confirmed: Satoshi)
+
+case class GlobalBalance private(total: Btc, onchain: CorrectedOnchainBalance, offchain: BalanceResult, mutualClose: MutualCloseStatus)
+
+object GlobalBalance {
+
+  private def apply(total: Btc, onchain: CorrectedOnchainBalance, offchain: BalanceResult, mutualClose: MutualCloseStatus): GlobalBalance = ???
```
pm47, line 66: The goal of this is to have the `total` attribute in the class `GlobalBalance` automatically computed as opposed to set manually (this could be bypassed by a call to `copy` though.

Why not make `total` a method? Because I want it to be automatically serialized in the json API response. We could add it manually in the json serializer but it then forces us to manually serialize all other fields, which is really ugly. Unless I'm missing something?

eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
```diff
@@ -0,0 +1,66 @@
...
+import akka.actor.typed.scaladsl.adapter.ClassicActorRefOps
+import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
+import fr.acinq.bitcoin.ByteVector32
+import fr.acinq.eclair.balance.ChannelsListener.{ChannelData, ChannelDied, Command, GetChannels, GetChannelsResponse}
+import fr.acinq.eclair.channel.Helpers.Closing
+import fr.acinq.eclair.channel.{ChannelPersisted, ChannelRestored, HasCommitments}
+
+import scala.concurrent.Promise
+
+object ChannelsListener {
```
pm47, line 16: The goal of this actor is to maintain a local cache of all channel data, that can be used by the API, instead of querying thousands of channel actors. Surprisingly, this is turns out to not be noticeably faster than the `ask`+`Future.sequence` pattern.

Maybe we should remove this? Or we could keep it and use that for all remaining API calls, because it saves some burden of the channel actors? I'm not sure.

pm47:
Squashed and rebased.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -385,25 +386,14 @@ private class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client
   }
 
   def checkUtxos(): Future[Unit] = {
```
t-bast, line 388: I think this whole check should be removed from the `ZmqWatcher` and instead move to the `BalanceActor`, which should check our utxos at every block to update the metric. I always had a plan to remove it from here once we had some actor dedicated to monitoring our utxos/balance.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
```diff
@@ -322,7 +336,9 @@ object ExtendedBitcoinClient {
    */
   case class MempoolTx(txid: ByteVector32, vsize: Long, weight: Long, replaceable: Boolean, fees: Satoshi, ancestorCount: Int, ancestorFees: Satoshi, descendantCount: Int, descendantFees: Satoshi)
 
-  case class Utxo(txid: ByteVector32, vout: Long)
+  case class UnlockOutpointUtxo(txid: ByteVector32, vout: Long)
```
t-bast, line 339: nit: I find it less confusing if that doesn't even mention "utxo" (it's actually just an outpoint, where we use `txid` instead of the reversed `hash`)

```suggestion
  case class UnlockOutpoint(txid: ByteVector32, vout: Long)
```

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -351,6 +359,7 @@ object Setup {
 }
 
 case class Kit(nodeParams: NodeParams,
+               bitcoin: BitcoinJsonRPCClient,
```
t-bast, line 362: I'm not a big fan of exposing that to all plugins.

The only reason it's necessary is because you put some of the logic inside `Eclair.scala` (in the `globalBalance()` function). I think it wasn't a good choice, IMO the logic should be inside the balance actors, and `globalBalance()` should simply send a request to the actor and return its response (this way it doesn't make direct calls to the bitcoin client).

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
```diff
@@ -0,0 +1,83 @@
...
+object BalanceActor {
+
+  // @formatter:off
+  sealed trait Command
+  private final case object TickBalance extends Command
+  private final case class WrappedChannels(wrapped: ChannelsListener.GetChannelsResponse) extends Command
+  private final case class WrappedGlobalBalance(wrapped: Try[GlobalBalance]) extends Command
+  // @formatter:on
+
+  def apply(eclair: Eclair, channelsListener: ActorRef[ChannelsListener.GetChannels], interval: FiniteDuration)(implicit ec: ExecutionContext): Behavior[Command] = {
```
t-bast, line 22: As mentioned in another comment, I don't think actors should use the API, I think it should be the other way around: the `eclair: Eclair` API should be a thin wrapper around some logic implemented by actors.

eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
```diff
@@ -0,0 +1,66 @@
...
+  // @formatter:on
+
+  case class GetChannelsResponse(channels: Map[ByteVector32, HasCommitments])
+
+  def apply(ready: Promise[Done]): Behavior[Command] =
+    Behaviors.setup { context =>
+      context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[ChannelRestored](e => ChannelData(e.channelId, e.channel, e.data)))
+      context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[ChannelPersisted](e => ChannelData(e.channelId, e.channel, e.data)))
+      // since subscription is asynchronous, we send a fake event so we know when we are subscribed
+      context.system.eventStream ! EventStream.Publish(ChannelRestored(null, ByteVector32.Zeroes, null, null, null))
```
t-bast, line 32: This is dangerous, other actors who registered to this event will receive this fake event as well and it may break some of their internal logic...Unfortunately it looks like there's no built-in mechanism to check whether we've subscribed.

I'd prefer using a different `case class` than `ChannelRestored` or `ChannelPersisted` to check when we're subscribed. Can you define a new `case class CheckSubscribed(actor: ActorRef[Something])`, subscribe to this type of messages, send one where `actor = context.self` and wait to receive it through the event stream?

eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
t-bast: I think it's useful to have it, when we'll have more load on the channel actors it should be noticeably useful, don't you think?

eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  def running(channels: Map[ByteVector32, HasCommitments]): Behavior[Command] =
+    Behaviors.receiveMessage {
+      case ChannelData(channelId, channel, data) =>
+        Closing.isClosed(data, additionalConfirmedTx_opt = None) match {
+          case None =>
+            context.watchWith(channel.toTyped, ChannelDied(channelId))
+            running(channels + (channelId -> data))
+          case _ =>
+            // if channel is closed we remove early from the map because
```
t-bast, line 54: because...? That cliff-hanger! :smile:

eclair-core/src/main/scala/fr/acinq/eclair/balance/CheckBalance.scala
```diff
@@ -0,0 +1,295 @@
...
+   *
+   * The base assumption regarding htlcs is that they will all timeout. That means that we ignore incoming htlcs, and we
+   * count outgoing htlcs in our balance.
+   */
+  case class MainAndHtlcBalance(toLocal: Btc = 0.sat, htlcOut: Btc = 0.sat) {
+    val total: Btc = toLocal + htlcOut
+  }
+
+  /**
+   * In the closing staten some transactions may be published or even confirmed. They will be taken into account if we
```
t-bast, line 30: nit:
```suggestion
   * In the closing state some transactions may be published or even confirmed. They will be taken into account if we
```

eclair-core/src/main/scala/fr/acinq/eclair/balance/CheckBalance.scala
```diff
@@ -0,0 +1,295 @@
...
+
+  /**
+   * Mutual close transactions are always immediately published, and will be taken into account by bitcoin core, so we
+   * don't count them in the balance.
+   */
+  case class ClosingBalance(
+                             localCloseBalance: PossiblyPublishedBalance = PossiblyPublishedBalance(),
+                             remoteCloseBalance: PossiblyPublishedBalance = PossiblyPublishedBalance(),
+                             unknownCloseBalance: MainAndHtlcBalance = MainAndHtlcBalance()
+                           ) {
```
t-bast, line 50: nit: I really find the indentation weird (same for other case classes below), I don't get why the formatter does that...

eclair-core/src/main/scala/fr/acinq/eclair/balance/Monitoring.scala
```diff
@@ -0,0 +1,63 @@
...
+    val GlobalBalanceDiff: Metric.Gauge = Kamon.gauge("globalbalance.diff", "Global Balance Diff (Satoshi)")
+    val GlobalBalanceDetailed: Metric.Gauge = Kamon.gauge("globalbalance.detailed", "Global Balance Detailed (BTC)")
+  }
+
+  object Tags {
+    val BalanceType = "type"
+    val OffchainState = "state"
+    val DiffSign = "sign"
+
+    /** TODO: there is duplication with [[fr.acinq.eclair.blockchain.Monitoring]] */
```
t-bast, line 36: I think it's worth removing the `blockchain.Monitoring` one if we move the utxo check inside the `balance` package (as proposed in the comment inside `ZmqWatcher`).

eclair-core/src/main/scala/fr/acinq/eclair/balance/Monitoring.scala
```diff
@@ -0,0 +1,63 @@
...
+      val normal = "normal"
+      val shutdown = "shutdown"
+      val negotiating = "negotiating"
+      val closingLocal = "closing-local"
+      val closingRemote = "closing-remote"
+      val closingUnknown = "closing-unknown"
+      val waitForPublishFutureCommitment = "waitForPublishFutureCommitment"
+    }
+
+    /** we can't chart negative amounts in Kamon */
```
t-bast, line 55: sigh...

t-bast (commented):
First pass, I haven't looked in details in the actual balance computation, but I'm proposing a few architectural changes and found some nits.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -53,15 +57,31 @@ case class AuditResponse(sent: Seq[PaymentSent], received: Seq[PaymentReceived],
...
+
+  def apply(onchain: CorrectedOnchainBalance, offchain: BalanceResult, mutualClose: MutualCloseStatus): GlobalBalance = new GlobalBalance(
+    total = onchain.total + offchain.total,
+    onchain = onchain,
+    offchain = offchain,
+    mutualClose = mutualClose
+  )
+}
```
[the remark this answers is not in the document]
pm47: Because that would fail at runtime and I wanted something that fails at compilation.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -53,15 +57,31 @@ case class AuditResponse(sent: Seq[PaymentSent], received: Seq[PaymentReceived],
 
 case class TimestampQueryFilters(from: Long, to: Long)
 
+case class MutualCloseStatus(unpublished: Satoshi, unconfirmed: Satoshi, confirmed: Satoshi)
+
+case class GlobalBalance private(total: Btc, onchain: CorrectedOnchainBalance, offchain: BalanceResult, mutualClose: MutualCloseStatus)
```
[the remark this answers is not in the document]
pm47: I think lowercase looks better in the json, but I agree this is more consistent.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
pm47: Yes I wasn't satisfied either. Another idea was to put argument-specific classes inside a dedicated `Args` object.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
pm47: That's fair, I'll rework that.

eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
pm47: Should we submit a PR to akka (cc @sstone)? This limitation is really annoying, not only here but in tests as well.

eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
pm47: I think this is what I was about to write:
```suggestion
            // if channel is closed we remove early from the map because we want to minimize the window during which
            // there is potential duplication between on-chain and off-chain amounts (the channel actors stays alive
            // for a few seconds after the channel is closed)
```

eclair-core/src/main/scala/fr/acinq/eclair/balance/CheckBalance.scala
pm47: The formatter is innocent, I bear the sole responsibility I'm afraid. Would you prefer:

```scala
case class ClosingBalance(localCloseBalance: PossiblyPublishedBalance = PossiblyPublishedBalance(),
                          remoteCloseBalance: PossiblyPublishedBalance = PossiblyPublishedBalance(),
                          unknownCloseBalance: MainAndHtlcBalance = MainAndHtlcBalance()) {
```

eclair-core/src/main/scala/fr/acinq/eclair/balance/Monitoring.scala
pm47: 🤷‍♂️

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -53,15 +57,31 @@ case class AuditResponse(sent: Seq[PaymentSent], received: Seq[PaymentReceived],
 
 case class TimestampQueryFilters(from: Long, to: Long)
 
+case class MutualCloseStatus(unpublished: Satoshi, unconfirmed: Satoshi, confirmed: Satoshi)
```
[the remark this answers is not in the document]
pm47: Actually, this is just additional information about mutual close transactions, useful to track evictions. The mutual close balance is tracked as part of onchain balance, see:

```scala
  /**
   * Mutual close transactions are always immediately published, and will be taken into account by bitcoin core, so we
   * don't count them in the balance.
   */
  case class ClosingBalance
```

But now that I think of it, we should probably treat mutual closes like other transactions, in the offchain balance, and prune it afterwards. The goal of pruning was initially to remove duplicates for transactions that were published (and possibly confirmed but less than mindepth), but it should also nicely handle evicted transactions.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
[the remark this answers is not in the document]
t-bast: But right now what you have fails at runtime as well since you have an `apply(...) = ???`...
Or is it a leftover that you wanted to remove?

eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
t-bast: Good idea, it's worth a shot and we'll learn a lot from their feedback.

eclair-core/src/main/scala/fr/acinq/eclair/balance/CheckBalance.scala
t-bast: Yes that looks better!

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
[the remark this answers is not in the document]
pm47: The method is defined as `private` the goal was to override the default one and make it inaccessible from the outside.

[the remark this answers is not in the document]
t-bast: But why can't you simply remove it?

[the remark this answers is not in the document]
pm47: It's a default method for case classes. Really I think the problem here is with serialization, I can't believe there is no way of simply adding a field to the base serialization.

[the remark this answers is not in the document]
t-bast: I agree, I find our serialization code unsatisfying, but maybe it's simply because we haven't spent enough time improving it and use the library imperfectly? It's always been a part of the code we haven't bothered changing much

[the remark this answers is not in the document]
pm47: :nerd_face:

```scala
case class GlobalBalance (onchain: CorrectedOnchainBalance, offchain: BalanceResult, mutualClose: MutualCloseStatus) {
  val total: Btc = onchain.total + offchain.total
}

class GlobalBalanceSerializer extends CustomSerializer[GlobalBalance](_ => ( {
  null
}, {
  case o: GlobalBalance =>
    JObject(JField("total", JDecimal(o.total.toDouble))) merge Extraction.decompose(o)(DefaultFormats + new BtcSerializer)
}))
```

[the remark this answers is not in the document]
pm47: Done with 635b2ce.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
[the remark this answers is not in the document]
pm47: Done with 1fd52f5.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
pm47: See discussion https://github.com/ACINQ/eclair/pull/1737#discussion_r661306163.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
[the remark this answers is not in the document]
pm47: Ok, I remember now why the mutual close transactions were handled like this. The problem is that there can be several competing mutual closes, and in that case the simple pruning logic that we have to make sure we count the amount exactly once doesn't work. This would lead us to possibly count twice the amount in the competing-mutual-closes scenario.

On the other hand, what we had before meant that we simply didn't count the amount in the competing-mutual-closes scenario, and also in the case where the mutual close was evicted. Not perfect either.

In general I would say that it is better to undershoot rather than overshoot the balance computation, but here the trade-off makes sense for consistency, and also it is easy to tell when we are potentially counting a mutual close twice: the `mutualCloseBalance` should always be zero after pruning.

So I went ahead and did it with 4b7839203, with a detailed comment on what is going on.

eclair-core/src/main/scala/fr/acinq/eclair/balance/CheckBalance.scala
pm47: Done with 4b78392.

eclair-core/src/main/scala/fr/acinq/eclair/balance/Monitoring.scala
pm47: Done in 2b60dc9. I have kept the previous metrics because they offer a deeper level of details.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/ExtendedBitcoinClient.scala
pm47: Done with 2400e7e.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
pm47: Done with 3606f4f.

pm47: Done in 3606f4f.

eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
pm47: I tried a slightly more elegant, but still hackish version here: 0f933fe96 in the shorter term.

eclair-node/src/main/scala/fr/acinq/eclair/api/serde/JsonSerializers.scala
```diff
@@ -68,6 +75,12 @@ class UInt64Serializer extends CustomSerializer[UInt64](_ => ( {
   case x: UInt64 => JInt(x.toBigInt)
 }))
 
+class BtcSerializer extends CustomSerializer[Btc](_ => ( {
```
t-bast, line 78: Definitely not for this PR, but I'm realizing only now that it would probably be much better if our serializers for bitcoin amounts returned an object that specified the unit, something like:
```json
{
  "amount": 0.3,
  "unit": "btc"
}
```
```json
{
  "amount": 0.3,
  "unit": "msat"
}
```

I'll create an issue to suggest that.

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
```diff
@@ -0,0 +1,139 @@
...
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
+
+object BalanceActor {
+
+  // @formatter:off
+  sealed trait Command
+  private final case object TickBalance extends Command
+  final case class GetGlobalBalance(replyTo: ActorRef[Future[GlobalBalance]], channels: Map[ByteVector32, HasCommitments]) extends Command
```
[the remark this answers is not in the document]
pm47: Indeed it is a bit strange, but we have to manage compatibility between typed ask and the existing api code. The alternative is:

```diff
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
index 44dd7aff4..ae29fec4c 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -438,7 +438,8 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
   override def globalBalance()(implicit timeout: Timeout): Future[GlobalBalance] = {
     for {
       ChannelsListener.GetChannelsResponse(channels) <- appKit.channelsListener.ask(ref => ChannelsListener.GetChannels(ref))(timeout, appKit.system.scheduler.toTyped)
-      globalBalance <- appKit.balanceActor.ask(res => BalanceActor.GetGlobalBalance(res, channels))(timeout, appKit.system.scheduler.toTyped).flatten
+      globalBalance_try <- appKit.balanceActor.ask(res => BalanceActor.GetGlobalBalance(res, channels))(timeout, appKit.system.scheduler.toTyped)
+      globalBalance = globalBalance_try.get
     } yield globalBalance
   }

diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala b/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
index f1963fe37..29b553c7d 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
@@ -1,5 +1,6 @@
 package fr.acinq.eclair.balance

+import akka.actor.Status
 import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
 import akka.actor.typed.{ActorRef, Behavior}
 import fr.acinq.bitcoin.ByteVector32
@@ -22,7 +23,7 @@ object BalanceActor {
   // @formatter:off
   sealed trait Command
   private final case object TickBalance extends Command
-  final case class GetGlobalBalance(replyTo: ActorRef[Future[GlobalBalance]], channels: Map[ByteVector32, HasCommitments]) extends Command
+  final case class GetGlobalBalance(replyTo: ActorRef[Try[GlobalBalance]], channels: Map[ByteVector32, HasCommitments]) extends Command
   private final case class WrappedChannels(wrapped: ChannelsListener.GetChannelsResponse) extends Command
   private final case class WrappedGlobalBalance(wrapped: Try[GlobalBalance]) extends Command
   private final case class WrappedUtxoInfo(wrapped: Try[UtxoInfo]) extends Command
@@ -112,7 +113,7 @@ private class BalanceActor(context: ActorContext[Command],
           Behaviors.same
       }
     case GetGlobalBalance(replyTo, channels) =>
-      replyTo ! CheckBalance.computeGlobalBalance(channels, pendingCommandsDb, extendedBitcoinClient)
+      CheckBalance.computeGlobalBalance(channels, pendingCommandsDb, extendedBitcoinClient) onComplete (replyTo ! _)
       Behaviors.same
     case WrappedUtxoInfo(res) =>
       res match {
```

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
```diff
@@ -0,0 +1,139 @@
...
+                           extendedBitcoinClient: ExtendedBitcoinClient,
+                           channelsListener: ActorRef[ChannelsListener.GetChannels])(implicit ec: ExecutionContext) {
+
+  private val log = context.log
+
+  def apply(refBalance_opt: Option[GlobalBalance]): Behavior[Command] = Behaviors.receiveMessage {
+    case TickBalance =>
+      log.debug("checking balance...")
+      channelsListener ! ChannelsListener.GetChannels(context.messageAdapter[ChannelsListener.GetChannelsResponse](WrappedChannels))
+      context.pipeToSelf(checkUtxos(extendedBitcoinClient))(WrappedUtxoInfo)
```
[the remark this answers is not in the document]
pm47: No because the `WrappedUtxoInfo` takes a try?

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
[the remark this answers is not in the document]
t-bast: I think this is slightly better. Instead of a `for` comprehension (which restricts us a bit here), you could also simply write the `map/flatMap` yourself and explicitly handle the failure case (which gets rid of the `.get`).

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
```diff
@@ -0,0 +1,139 @@
...
+              val normalizedValue = 100 + (if (refBalance.total.toSatoshi.toLong > 0) (result.total.toSatoshi.toLong - refBalance.total.toSatoshi.toLong) * 1000D / refBalance.total.toSatoshi.toLong else 0)
+              val diffValue = result.total.toSatoshi.toLong - refBalance.total.toSatoshi.toLong
+              log.info("relative balance: current={} reference={} normalized={} diff={}", result.total.toDouble, refBalance.total.toDouble, normalizedValue, diffValue)
+              Metrics.GlobalBalanceNormalized.withoutTags().update(normalizedValue)
+              Metrics.GlobalBalanceDiff.withTag(Tags.DiffSign, Tags.DiffSigns.plus).update(diffValue.max(0))
+              Metrics.GlobalBalanceDiff.withTag(Tags.DiffSign, Tags.DiffSigns.minus).update((-diffValue).max(0))
+              Behaviors.same
+            case None =>
+              log.info("using balance={} as reference", result.total.toDouble)
+              apply(Some(result))
```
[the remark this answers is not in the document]
pm47: We do compute absolute and relative balance. The relative balance is useful to troubleshoot issues with... the computing of the balance itself.

eclair-core/src/test/scala/fr/acinq/eclair/balance/CheckBalanceSpec.scala
```diff
@@ -0,0 +1,114 @@
...
+
+import java.io.File
+import java.sql.DriverManager
+import scala.collection.immutable.Queue
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, ExecutionContext, Future}
+
+class CheckBalanceSpec extends AnyFunSuite {
+
+  ignore("compute from eclair.sqlite") {
```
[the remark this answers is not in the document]
pm47: It may come in handy because I get the feeling we're not quite done with this balance computation.

eclair-core/src/test/scala/fr/acinq/eclair/balance/CheckBalanceSpec.scala
```diff
@@ -0,0 +1,114 @@
...
+    val dbFile = new File("eclair.sqlite")
+    val sqliteConfig = new SQLiteConfig()
+    sqliteConfig.setReadOnly(true)
+    val sqlite = DriverManager.getConnection(s"jdbc:sqlite:$dbFile", sqliteConfig.toProperties)
```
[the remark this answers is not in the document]
pm47: It's not compatible with the read-only flag, due to how we handle version.

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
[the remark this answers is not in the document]
pm47: 🤓 

```scala
globalBalance <- Promise[GlobalBalance]().complete(globalBalance_try).future
```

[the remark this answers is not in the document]
t-bast: witchcraft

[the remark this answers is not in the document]
pm47: Applied in 3ffbe81.

t-bast (approved):
LGTM. I haven't looked in details at the actual balance computation details, real-world experience is the best judge for whether it works or not ;)

And we can easily implement several balance computation algorithms and run them in parallel, and see which one mirrors more closely reality.

pm47:
Rebased and squashed.

t-bast (approved):

PR #1738, opened 2021-03-23
Handle aggregated anchor outputs htlc txs

t-bast:
An interesting side-effect of anchor outputs is that htlc txs can be merged when they have the same lockTime (thanks to sighash flags).

We're not currently doing that, but our peers may do it, so we need to handle it in the revoked commit tx case and correctly claim multiple outputs if necessary.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
@@ -563,7 +563,7 @@ object Transactions {
   def makeClaimP2WPKHOutputTx(commitTx: Transaction, localDustLimit: Satoshi, localPaymentPubkey: PublicKey, localFinalScriptPubKey: ByteVector, feeratePerKw: FeeratePerKw): Either[TxGenerationSkipped, ClaimP2WPKHOutputTx] = {
     val redeemScript = Script.pay2pkh(localPaymentPubkey)
     val pubkeyScript = write(pay2wpkh(localPaymentPubkey))
-    findPubKeyScriptIndex(commitTx, pubkeyScript, amount_opt = None) match {
+    findPubKeyScriptIndex(commitTx, pubkeyScript) match {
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -755,21 +756,30 @@ object Transactions {
...
-    val toLocalOutput = findPubKeyScriptIndex(tx, localScriptPubKey, None).map(index => OutputInfo(index, toLocalAmount, localScriptPubKey)).toOption
+    val toLocalOutput = findPubKeyScriptIndex(tx, localScriptPubKey).map(index => OutputInfo(index, toLocalAmount, localScriptPubKey)).toOption
     ClosingTx(commitTxInput, tx, toLocalOutput)
   }
 
-  def findPubKeyScriptIndex(tx: Transaction, pubkeyScript: ByteVector, amount_opt: Option[Satoshi]): Either[TxGenerationSkipped, Int] = {
-    val outputIndex = tx.txOut
-      .zipWithIndex
-      .indexWhere { case (txOut, _) => amount_opt.forall(_ == txOut.amount) && txOut.publicKeyScript == pubkeyScript }
+  def findPubKeyScriptIndex(tx: Transaction, pubkeyScript: ByteVector): Either[TxGenerationSkipped, Int] = {
```
pm47, line 763: What changed that makes the `amount` not needed anymore? We added this in 449c3071327d1c47b0bca07862afd0e67dd07d04 to handle multiple similar htlcs.

t-bast: We don't need to deduplicate anymore since we now which output maps to which HTLC.
I think we stopped depending on this with https://github.com/ACINQ/eclair/pull/790 where we introduced the `CommitmentOutputLink`.

pm47 (approved):

PR #1740, opened 2021-03-24
Attempt to clarify units in docs

### LINKED ISSUE #1739: channelReserve unit unclear from documentation
Expected: API doc clearly states whether `channelReserve` is sat or msat.

Current state: API doc has no such informtion. Quick look at the code suggests that it's most likely in sats. This seems to make sense looking at the data.

Steps to reproduce:

1. open https://acinq.github.io/eclair/#channels-2
2. look at channel reserve

Version and environment: Irrelevant, I believe

Kixunil:
This is a draft of specifying units of fields that are not obvious.

Closes #1739

I inferred the units from context, so please double-check if it's correct. Also please LMK what unit to specify for `htlcMinimum`.

### DIFF
--- a/source/index.html.md
+++ b/source/index.html.md
@@ -359,6 +359,8 @@ eclair-cli updaterelayfee \
 }
 ```
 
+Returned `feeBase` is in msat.
+
 Updates the fee policy for the specified _channelId_. A new update for this channel will be broadcast to the network.
 The endpoint supports receiving multiple channel id(s) or short channel id(s); to update multiple channels, you can use the parameters `channelIds` or `shortChannelIds` below.
 
@@ -422,6 +424,16 @@ curl -u :<eclair_api_password> -X POST "http://localhost:8080/channels"
 # with eclair-cli
 eclair-cli channels
 ```
+
+The units of returned fields that are not obvious from their names:
+
+field          | unit
+---------------|-----
+dustLimit      | sats
+channelReserve | sats
+htlcMinimum    | msats
+toSelfDelay    | blocks
+feeratePerKw   | sats
 
 > The above command returns:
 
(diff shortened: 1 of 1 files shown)


source/index.html.md
```diff
@@ -423,6 +425,16 @@ curl -u :<eclair_api_password> -X POST "http://localhost:8080/channels"
 eclair-cli channels
 ```
 
+The units of returned fields that are not obvious from their names:
+
+field          | unit
+---------------|-----
+dustLimit      | sats
+channelReserve | sats
+htlcMinimum    | ???
```
t-bast, line 434: This one is `msats`

source/index.html.md
```diff
@@ -423,6 +425,16 @@ curl -u :<eclair_api_password> -X POST "http://localhost:8080/channels"
...
+toSelfDelay    | blocks
+feeratePerKw   | sats
```
t-bast, line 436: This is `sat/kw` to be more precise (bitcoin feerate units can be very confusing - but it's getting better). This also applies to line 653.

source/index.html.md
```diff
@@ -359,6 +359,8 @@ eclair-cli updaterelayfee \
 }
 ```
 
+Returned `feeBase` is in msat.
```
t-bast, line 362: Not sure this is necessary, a couple lines below it's obvious that it's `msat`.

t-bast (commented):
Looks good, thanks!

source/index.html.md
Kixunil: Yeah, it's in name so I feared it'd make it look like sat/kw^2 but maybe it's too strange for anyone to seriously think about it?

source/index.html.md
Kixunil: True, shouldn't hurt though, right?

source/index.html.md
t-bast: Right, that makes sense, let's keep it like that.

source/index.html.md
t-bast: Agreed, let's keep it if you think it's useful.

source/index.html.md
```diff
@@ -630,6 +642,16 @@ curl -u :<eclair_api_password> -X POST -F channelId=<channel>  "http://localhost
 eclair-cli channel --channelId=<channel>
 ```
 
+The units of returned fields that are not obvious from their names:
+
+field          | unit
+---------------|-----
+dustLimit      | sats
+channelReserve | sats
+htlcMinimum    | ???
```
t-bast, line 651: You've missed that one, then we should be good to go.

Kixunil:
OK, I've fixed the other `???` and squashed.

t-bast (approved):

PR #1743, opened 2021-03-26
Refactor Postgres code

pm47:


eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
```diff
@@ -18,81 +18,174 @@ package fr.acinq.eclair.db.pg
...
+            statement.executeUpdate(s"CREATE TABLE IF NOT EXISTS $LeaseTable (id INTEGER PRIMARY KEY default(1), expires_at TIMESTAMP NOT NULL, instance VARCHAR NOT NULL, CONSTRAINT one_row CHECK (id = 1))")
+        }
+      }
 
-  case object NoLock extends DatabaseLock {
-    override def obtainExclusiveLock(implicit ds: DataSource): Unit = ()
+      private def acquireExclusiveTableLock()(implicit connection: Connection): Unit = {
+        using(connection.createStatement()) {
+          statement =>
+            statement.executeUpdate(s"SET lock_timeout TO '${LockTimeout.toSeconds}s'")
```
pm47, line 134: @rorp. Why not use `NOWAIT`  here?

pm47: I went ahead and changed it in 723d926, but I'm still interested in your rationale opting for a non-zero timeout.

rorp: AFAIR there was some weirdness in exceptions Pg driver threw in NOWAIT mode. Should work in both ways.

rorp (dismissed):
👍

eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
```diff
@@ -0,0 +1,81 @@
...
+    jdbcUrlFile.deleteOnExit()
+
+    implicit val system: ActorSystem = ActorSystem()
+
+    // @formatter:off
+  override val connection: Connection = pg.getPostgresDatabase.getConnection
+  override lazy val db: Databases = Databases.PostgresDatabases(hikariConfig, UUID.randomUUID(), lock, jdbcUrlFile_opt = Some(jdbcUrlFile))
+  override def getVersion(statement: Statement, db_name: String, currentVersion: Int): Int = PgUtils.getVersion(statement, db_name, currentVersion)
+  override def close(): Unit = pg.close()
+  // @formatter:on
```
t-bast, line 70: nit: indentation is weird

eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  def forAllDbs(f: TestDatabases => Unit): Unit = {
+    // @formatter:off
+  def using(dbs: TestDatabases)(g: TestDatabases => Unit): Unit = try g(dbs) finally dbs.close()
+  using(TestSqliteDatabases())(f)
+  using(TestPgDatabases())(f)
+  // @formatter:on
```
t-bast, line 78: Here as well

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+      intercept[LockFailureHandler.LockException] {
+        // this will fail because the database is already locked for a different instance id
+        Databases.postgres(config, UUID.randomUUID(), datadir, LockFailureHandler.logAndThrow)
+      }.lockFailure === LockFailure.AlreadyLocked(instanceId1))
+
+    // we can renew the lease at will
+    db.obtainExclusiveLock()
+
+    // we wait longer than the lease interval, and make sure that the lock is still there
+    Thread.sleep(20_000)
```
t-bast, line 37: Can't we configure a shorter lease interval in tests?

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+      intercept[LockFailureHandler.LockException] {
+        // this will fail because the database is already locked for a different instance id
+        Databases.postgres(config, UUID.randomUUID(), datadir, LockFailureHandler.logAndThrow)
+      }.lockFailure === LockFailure.AlreadyLocked(instanceId1))
+
+    // we close the first connection
+    db.dataSource.close()
+    while (!db.dataSource.isClosed) {
+      Thread.sleep(1000)
+    }
```
t-bast, line 48: Looks like you should rather extend `TestKitBase` and use `awaitCond`? Especially since you need an `ActorSystem`.

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+    Thread.sleep(11_000)
```
t-bast, line 49: Where does that number come from?

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
pm47: Seeing that lock errors seem to be handled fine in tests, I think we can go ahead with `NOWAIT`.

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
pm47: We already are using short values (see https://github.com/ACINQ/eclair/blob/e18e81aaba22b24714659a5e7ceb47eae1171a39/eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala#L112-L117). We could go a  bit further but I don't want to be too extreme because 5s renew interval is already pretty short. I'l try to reduce them further.

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
pm47: It's `lease.interval` + 1s

t-bast: Can we make that more explicit?

eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
pm47: Done with d506a4a.

eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
pm47: Done with d506a4a.

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
pm47: Done with 912ba4e.

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
pm47: Done with 912ba4e.

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
pm47: Done with 8a844fa.

I didn't find a better way than using `Thread.sleep`. It seems there is no standard non-blocking way of waiting in tests. I looked into scalatest's `Waiters` but it's for a different purpose. `awaitCond` and `eventually` are for cases where we expect a statement to eventually return a certain value. But here we just want to wait.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -109,127 +166,52 @@ object Databases extends Logging {
...
-        statement => {
-          statement.executeUpdate(s"backup to ${backupFile.getAbsolutePath}")
-        }
-      }
-
     }
   }
 
-  def setupPgDatabases(dbConfig: Config, instanceId: UUID, datadir: File, lockExceptionHandler: LockExceptionHandler): Databases with ExclusiveLock = {
+  def postgres(dbConfig: Config, instanceId: UUID, dbdir: File, lockExceptionHandler: LockFailureHandler = LockFailureHandler.logAndStop)(implicit system: ActorSystem): PostgresDatabases = {
```
t-bast, line 180: Why do you need to pull in an implicit ActorSystem?

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgPaymentsDb.scala
```diff
@@ -16,27 +16,26 @@
...
-import java.sql.ResultSet
-import java.util.UUID
-
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.MilliSatoshi
 import fr.acinq.eclair.db.Monitoring.Metrics.withMetrics
-import fr.acinq.eclair.db.pg.PgUtils.DatabaseLock
-import fr.acinq.eclair.db._
+import fr.acinq.eclair.db.{HopSummary, _}
```
t-bast, line 23: nit:

```suggestion
import fr.acinq.eclair.db._
```

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
t-bast: But this should work:

```
awaitCond(db.dataSource.isClosed)
```

Doesn't it? And it can be used in multiple places in this file.

I agree that you can't really replace the `Thread.sleep` that are waiting for the lease to expire on the postgres side but these ones can be improved.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
pm47: The locking mechanism needs it to schedule the periodic lease refresh.

NB: What we really need is a `Scheduler` and an `ExecutionContext`. Actually it could make sense to create a private independent `ExecutionContext` for that purpose. And if we go that way, since we would have a dedicated thread available, we may as well implement the scheduling part ourselves and remove the implicits altogether. What do you think?

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
pm47: Totally! Done with f243a119d.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgPaymentsDb.scala
pm47: How do you spot those? My import optimizer is totally oblivious.

t-bast (approved):

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
pm47: Okay I think I know why it was done that way. There can be a conflict between a regular SQL requests checking the lock, and the background task that regularly renew the lease. If those happen at the same time, one of them will fail which will make the app stop. Adding a 5 seconds timeout provided some leeway, one of the lock attempts would just wait after the first one is completed.

PR #1744, opened 2021-03-26
Introduce actor factories

t-bast:
We'd like to provide an abstraction for child actors creation, otherwise our actors become too tightly coupled and hard to test in isolation. But we don't want to move too much business logic outside of the actor (for instance the choice of child supervision strategy).

I'm going with a simple ad-hoc factory pattern in this PR, which removes unnecessary fields in actor constructors and allows more flexibility in tests. I must say I'm not entirely satisfied, but can't find a simple alternative (it's important to have a simple solution here, I don't want to make it harder for readers of the code).

If you have a better naming for the concrete Factory classes, let me know!


pm47:
I don't really like factories in general, but I don't know a better way of solving your problem, and I have to say it looks pretty clean.

Regarding names, I would have gone with `SimplePeerFactory`/`SimpleChannelFactory`.

t-bast:
> Regarding names, I would have gone with SimplePeerFactory/SimpleChannelFactory.

Harmonized in https://github.com/ACINQ/eclair/pull/1744/commits/52af73030cf5b1751988d460e00f0035d6739954

pm47 (approved):

PR #1745, opened 2021-03-29
Fix implementation of Dijkstra's algorithm

thomash-acinq:
Ensure that we only visit each node once in Dijkstra's algorithm.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -208,8 +208,8 @@ object Graph {
     while (toExplore.nonEmpty && !targetFound) {
       // node with the smallest distance from the target
       val current = toExplore.dequeue() // O(log(n))
-      if (current.key != sourceNode) {
-        val currentWeight = bestWeights(current.key) // NB: there is always an entry for the current in the 'bestWeights' map
+      targetFound = current.key == sourceNode
+      if (!targetFound && current.weight == bestWeights(current.key)) { // NB: there is always an entry for the current key in the 'bestWeights' map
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -208,8 +209,9 @@ object Graph {
     while (toExplore.nonEmpty && !targetFound) {
       // node with the smallest distance from the target
       val current = toExplore.dequeue() // O(log(n))
-      if (current.key != sourceNode) {
-        val currentWeight = bestWeights(current.key) // NB: there is always an entry for the current in the 'bestWeights' map
```
t-bast, line 212: Are you sure about that change?
At first glance getting the best weight from this map seems useful, doesn't it?

t-bast (commented):
Good catch!

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
thomash-acinq: It's exactly the same weight. We already have it in `current.weight`, there is no point in getting it from somewhere else.

Actually we could remove `bestWeights` completely, it's not needed by the algorithm, the comparison between the different weights for a node is done inside the priority queue.
Now it's just used to know when to update `bestEdges` and it sometimes prevents adding useless weights to the priority queue.

t-bast: > the comparison between the different weights for a node is done inside the priority queue.

Right, I glossed over that point, once you realize this it's obviously not needed.

t-bast (approved):

PR #1746, opened 2021-03-30
Add support for dual db backend

pm47:
We still use sqlite as the primary db, but all calls are replicated
asynchronously on postgres.

The goal is to prepare a smooth transition from sqlite to postgres
on a production server. This is a very specific use case and most users
shouldn't use it, which is why the new config `eclair.db.driver=dual` is
not documented.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
@@ -1,0 +1,378 @@
+package fr.acinq.eclair.db
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder
+import fr.acinq.bitcoin.{ByteVector32, Crypto, Satoshi}
+import fr.acinq.eclair.channel._
+import fr.acinq.eclair.db.Databases.{FileBackup, PostgresDatabases, SqliteDatabases}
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
```diff
@@ -0,0 +1,378 @@
...
+
+object DualDatabases extends Logging {
+
+  /** Run asynchronously and print errors */
+  def runAsync[T](f: => T)(implicit ec: ExecutionContext): Future[T] = Future {
+    Try(f) match {
+      case Success(res) => res
+      case Failure(t) =>
+        logger.error("postgres error:\n", t)
+        throw t
```
t-bast, line 55: Do we really want to throw here?
Since Sqlite is our primary DB, shouldn't we just swallow errors since this dual mode is only meant to let us monitor how postgres behaves in practice? This way we're sure it doesn't impact the running node, and can be deployed on mainnet safely (hopefully?).

pm47: It already works that way, because this is all within a future that we are not waiting on. We have to provide a return type or an exception, but the `throw` is effectively ignored.

t-bast (approved):

PR #1747, opened 2021-03-30
Shortest paths: prevent producing loops and remove some duplicated work

thomash-acinq:
Several improvements to `yenKshortestPaths`:
- Run `yenKshortestPaths` backwards like what is already done in `dijkstraShortestPath`. It allows to compute the correct weights directly.
- Prevent producing loops by excluding the root path nodes
- Do not run `dijkstraShortestPath` several times on the same input by only computing spur paths for new edges. `shortestPaths(k)` often has a lot of overlap with the previous `shortestPaths` so a lot of the poential spur nodes were already considered.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -103,7 +103,7 @@ object Graph {
       return Seq.empty // if we can't even find a single path, avoid returning a Seq(Seq.empty)
     }
 
-    case class PathWithSpur(p : WeightedPath, spurIndex : Int)
+    case class PathWithSpur(p: WeightedPath, spurIndex: Int)
     implicit object PathWithSpurComparator extends Ordering[PathWithSpur] {
       override def compare(x: PathWithSpur, y: PathWithSpur): Int = y.p.weight.compare(x.p.weight)
     }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -117,9 +117,9 @@ object Graph {
     // main loop
     for (k <- 1 until pathsToFind) {
       if (!allSpurPathsFound) {
-        val prevShortestPath = shortestPaths(k - 1).p.path
+        val PathWithSpur(WeightedPath(prevShortestPath, _), spurIndex) = shortestPaths(k - 1)
         // for every new edge in the path, we will try to find a different path before that edge
-        for (i <- shortestPaths(k - 1).spurIndex until prevShortestPath.length) {
+        for (i <- spurIndex until prevShortestPath.length) {
           // select the spur node as the i-th element from the target of the previous shortest path
           val spurNode = prevShortestPath(prevShortestPath.length - 1 - i).desc.b
           // select the sub-path from the spur node to the target
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -141,7 +141,6 @@ object Graph {
           if (spurPath.nonEmpty) {
             val completePath = spurPath ++ rootPathEdges
             val candidatePath = WeightedPath(completePath, pathWeight(sourceNode, completePath, amount, currentBlockHeight, wr))
-            assert(boundaries(candidatePath.weight) && !shortestPaths.exists(_.p == candidatePath) && !candidates.exists(_.p == candidatePath) && validatePath(completePath, amount))
             candidates.enqueue(PathWithSpur(candidatePath, i))
           }
         }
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -98,49 +98,50 @@ object Graph {
...
-        val prevShortestPath = shortestPaths(k - 1).path
-        // for every edge in the path, we will try to find a different path after that edge
-        for (i <- prevShortestPath.indices) {
-          // select the spur node as the i-th element of the previous shortest path
-          val spurNode = prevShortestPath(i).desc.a
-          // select the sub-path from the source to the spur node
-          val rootPathEdges = prevShortestPath.take(i)
+        val prevShortestPath = shortestPaths(k - 1).p.path
+        // for every new edge in the path, we will try to find a different path before that edge
+        for (i <- shortestPaths(k - 1).spurIndex until prevShortestPath.length) {
```
t-bast, line 122: nit:

```suggestion
        val PathWithSpur(WeightedPath(prevShortestPath, _), spurIndex) = shortestPaths(k - 1)
        // for every new edge in the path, we will try to find a different path before that edge
        for (i <- spurIndex until prevShortestPath.length) {
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -98,49 +98,50 @@ object Graph {
...
+          // if for example the spur node is D, and we already found shortest paths ending with A->D->E and B->D->E,
+          // we want to ignore the A->D and B->D edges
+          // [...] --> A --+
+          //               |
+          // [...] --> B --+--> D --> E
+          //               |
+          // [...] --> C --+
+          val alreadyExploredEdges = shortestPaths.collect { case p if p.p.path.takeRight(i) == rootPathEdges => p.p.path(p.p.path.length - 1 - i).desc }.toSet
+          // we also want to ignore any vertex on the root path to prevent loops
+          val alreadyExploredVertices = rootPathEdges.map(_.desc.b).toSet
```
t-bast, line 137: Isn't this too aggressive? Aren't we going to miss edges from a previously explored vertex that lead to a new alternative route?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -175,7 +175,6 @@ object Graph {
    * @param wr                 ratios used to 'weight' edges when searching for the shortest path
    */
   private def dijkstraShortestPath(g: DirectedGraph,
-                                   sender: PublicKey,
```
t-bast, line 178: Clever, you've been able to remove the need for that field by reversing the flow of k-shortest path, I like it!

t-bast (commented):
Nice work!
I'll run some tests on a real-world graph.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
thomash-acinq: No, any spur path that passes through a node on the root path would create a loop since the same node is visited once in the spur path and then again in the root path. That's why it used to produce paths with loops.

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -1545,6 +1546,82 @@ class RouteCalculationSpec extends AnyFunSuite with ParallelTestExecution {
...
+  test("k-shortest paths must be distinct") {
+
+    def makeEdges(n: Int): Seq[GraphEdge] = {
+      val nodes = new Array[(PublicKey, PublicKey)](n)
+      for (i <- nodes.indices) {
+        nodes(i) = (randomKey.publicKey, randomKey.publicKey)
+      }
+      val q = new mutable.Queue[GraphEdge]
+      q.enqueue(makeEdge(1L, a, nodes(0)._1, 100 msat, 90))
+      q.enqueue(makeEdge(2L, a, nodes(0)._2, 100 msat, 100))
```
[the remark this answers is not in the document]
thomash-acinq: I wanted to ensure that all the shortest paths share as long a prefix as possible. By having one side more expensive, I guarantee that all k-shortest paths differ by at most one node from the shortest path.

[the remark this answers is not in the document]
t-bast: Ok, that's understandable, then maybe add a comment to explain that?
Do you think it's worth having both tests (once with same cost everywhere and one with one cheaper route to have long common prefixes)?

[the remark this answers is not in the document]
thomash-acinq: Yes, I've added the test with all paths costing the same.

t-bast (approved):
LGTM :+1:

PR #1749, opened 2021-03-31
Extract tx publishing from watchers

t-bast:
Introduce a `TxPublisher` actor whose responsibility is to publish channel transactions (from funding to closing).
Move logic from `ZmqWatcher` to this new actor, with minimal functional changes.

This is best reviewed commit-by-commit:

- The first commit moves code from watcher to the tx-publisher (almost no functional change)
- The second commit slightly refactors tx publishing and uses `Commitments` instead of `TxSigningKit`
- The third commit adds a few tests around cltv and csv delays

In future PRs, we will refactor and enrich this `TxPublisher`.


eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
```diff
@@ -0,0 +1,418 @@
...
+        val nextBlockCount = nodeParams.currentBlockHeight + 1
+        val cltvDelayedTxs1 = cltvDelayedTxs + (nextBlockCount -> (cltvDelayedTxs.getOrElse(nextBlockCount, Seq.empty) :+ p))
+        run(cltvDelayedTxs1, csvDelayedTxs)
+    }
+
+  implicit val ec: ExecutionContext = ExecutionContext.global
+
+  // NOTE: we use a single thread to publish transactions so that it preserves order.
+  // CHANGING THIS WILL RESULT IN CONCURRENCY ISSUES WHILE PUBLISHING PARENT AND CHILD TXS!
+  val singleThreadExecutionContext = ExecutionContext.fromExecutor(Executors.newSingleThreadExecutor())
```
pm47, line 246: We can't create a thread per channel, this should be defined in the _companion object_. All transaction publication will be serialized, but it's okay, and that's how it is currently.

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
```diff
@@ -0,0 +1,418 @@
...
+private class TxPublisher(nodeParams: NodeParams, watcher: akka.actor.ActorRef, client: ExtendedBitcoinClient, context: ActorContext[TxPublisher.Command]) {
+
+  import TxPublisher._
+
+  private case class TxWithRelativeDelay(childTx: PublishTx, parentTxIds: Set[ByteVector32])
+
+  val log = context.log
+  val keyManager = nodeParams.channelKeyManager
+  val feeEstimator = nodeParams.onChainFeeConf.feeEstimator
+  val feeTargets = nodeParams.onChainFeeConf.feeTargets
```
pm47, line 172:
```suggestion
  private val log = context.log
  private val keyManager = nodeParams.channelKeyManager
  private val feeEstimator = nodeParams.onChainFeeConf.feeEstimator
  private val feeTargets = nodeParams.onChainFeeConf.feeTargets
```

It feels weird to redefine those as `val`, I think `def` would be a better fit.

I would even go with:
```suggestion
  import context.log
  import nodeParams.{channelKeyManager => keyManager}
  import nodeParams.onChainFeeConf.feeEstimator
  import nodeParams.onChainFeeConf.feeTargets
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
```diff
@@ -0,0 +1,418 @@
...
+ * This actor ensures its parent channel's on-chain transactions confirm in a timely manner.
+ * It sets the fees, tracks confirmation progress and bumps fees if necessary.
+ */
+object TxPublisher {
+
+  // @formatter:off
+  sealed trait Command
+  sealed trait PublishTx extends Command {
+    /** Actor that should receive a [[WatchEventConfirmed]] once the transaction has been confirmed. */
+    def replyTo: ActorRef[WatchEvent]
```
pm47, line 51: This field looks unused, maybe let's defer introducing it before it's really used? Also why doesn't it take a `WatchEventConfirmed`?

```suggestion
    def replyTo: ActorRef[WatchEventConfirmed]
```

pm47 (commented):
I don't really like 05111f86c3ccca652c9ed8fb3b3920c209759b96, but I see it's meant as an intermediate step.

What do you mean by?
> We also now publish the commit tx and its anchor tx independently.

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: Good catch

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: I figured it's a good practice to define it right from the start, but maybe it's worth waiting to actually use it indeed.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1749/commits/d17b3d0ab1d33f504c04ad67a537b46649ee3230

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1749/commits/d17b3d0ab1d33f504c04ad67a537b46649ee3230

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1749/commits/d17b3d0ab1d33f504c04ad67a537b46649ee3230 except for `context.log` which somehow freezes the tests...?

t-bast:
> What do you mean by?
> > We also now publish the commit tx and its anchor tx independently.

What we previously had was a bit ugly: we sent one `PublishAsap` for the commit tx which in fact could end up publishing either only the commit tx or the commit tx and its anchor child, and it wasn't very intuitive what transaction the `txInfo` field referred to.

Now it is cleaner, we separate these into two messages.
We first call `PublishRawTx` with the commit tx, and then `SignAndPublishTx` with the anchor tx (where we check whether we really need to publish this or not depending on the current on-chain feerate).

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
pm47: I don't think we should make `singleThreadExecutionContext` a class-wide implicit execution context, because it may be picked up by other unrelated asynchronous tasks and will generate deadlocks. Instead I would not make it an implicit, and have the callers reference it directly from the object, so it is very clear what operations are serialized. Note that in the previous implementation it was actually provided explicitly and still is:

```scala
  private def publish(tx: Transaction): Future[ByteVector32] = {
    client.publishTransaction(tx)(singleThreadExecutionContext).recoverWith {
      case t: Throwable =>
        log.error("cannot publish tx: reason={} txid={}", t.getMessage, tx.txid)
        Future.failed(t)
    }
  }
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
```diff
@@ -0,0 +1,413 @@
...
+
+  // NOTE: we use a single thread to publish transactions so that it preserves order.
+  // CHANGING THIS WILL RESULT IN CONCURRENCY ISSUES WHILE PUBLISHING PARENT AND CHILD TXS!
+  val singleThreadExecutionContext = ExecutionContext.fromExecutor(Executors.newSingleThreadExecutor())
+
+  def apply(nodeParams: NodeParams, watcher: akka.actor.ActorRef, client: ExtendedBitcoinClient): Behavior[Command] =
+    Behaviors.setup { context =>
+      context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[CurrentBlockCount](cbc => WrappedCurrentBlockCount(cbc.blockCount)))
+      new TxPublisher(nodeParams, watcher, client, context)(singleThreadExecutionContext).run(SortedMap.empty, Map.empty)
+    }
```
pm47, line 74: One of the great advantages of making this publisher a child of the channel is that we can now contextualize all logs with an MDC, which is awesome for debugging. Here it would be `nodeId` and `channelId`.

Here is an example:
https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala#L53-L59

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: It feels a bit dangerous because we can easily forget to use it on one of the calls that require it (and only code review catches that).
Maybe if we force *all* calls to explicitly specify their execution context (no implicit global execution context as we had in previous `ZmqWatcher`) then it's better, I'll try that.

pm47: > we can easily forget to use it on one of the calls that require it

If there is no implicit execution context it won't compile. However I share the concern, maybe we could go with another architecture that prevents foot guns.

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: It's unfortunately not that simple due to the complexity of the `Channel` initialization...
We don't have the `channelId` available easily.

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: I simplified this in https://github.com/ACINQ/eclair/pull/1749/commits/989baaa254dc370d2369619a2efa00de534f9bb7
There's only one place where we call `client.publishTransaction` now and it's where we explicitly use the `singleThreadExecutionContext`.

Unless I'm missing something, I believe the only case where tx ordering matters is for the commit tx, that needs to be published first. There is not parent-child dependency elsewhere since we now publish htlc 3rd-stage txs only once htlc-txs are confirmed. Maybe we can do something more clever in the future where we special case the commit tx somehow and wait for it to be in the mempool before publishing other txs, TBD.

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
```diff
@@ -0,0 +1,413 @@
...
+            log.error("cannot publish tx: reason={} input={}:{} txid={}", t.getMessage, txInfo.input.outPoint.txid, txInfo.input.outPoint.index, p.tx.txid)
+            Future.failed(t)
+        }
+      case PublishRawTx(tx) =>
+        log.info("publishing tx: txid={} tx={}", tx.txid, tx)
+        publish(tx)
+    }
+  }
+
+  private def publish(tx: Transaction): Future[ByteVector32] = {
```
pm47, line 269:
```suggestion
  /**
   * This method uses a single thread to publish transactions so that it preserves the order of publication.
   * We need that to prevent concurrency issues while publishing parent and child transactions.
   */
  private def publish(tx: Transaction): Future[ByteVector32] = {
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
pm47: We could also use the `pipeTo` pattern and maintain a publication queue in the actor state, so we don't rely on a single thread, which is kind of hacky.

> Unless I'm missing something, I believe the only case where tx ordering matters is for the commit tx, that needs to be published first.

I'm embarrassed to say I can't tell where the ordering is required. Our 2nd stage transactions are delayed relative to the commit tx, so there shouldn't be a problem either. 🤔

t-bast: > We could also use the pipeTo pattern and maintain a publication queue in the actor state, so we don't rely on a single thread, which is kind of hacky.

I'll keep that in mind in a future PR where I'll make the tx publishing logic more modular (in particular to cleanly unlock utxos in case of failures), we may have an opportunity to improve that then.

> I'm embarrassed to say I can't tell where the ordering is required. Our 2nd stage transactions are delayed relative to the commit tx

The only cases where I think it could matter are:

- spending the anchor (which obviously needs the commit tx to be in the mempool)
- HTLC txs with the default commitment format (they don't have a csv 1, they can be in the same block as the commit tx)

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1749/commits/515c2fa945eae1872d70af7c418e9ada6e521090

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1749/commits/7239391f41248f24cc9d1d9933713ba76e7bdda9
It seems to work fine!

t-bast:
I added the `desc` field to `TransactionWithInputInfo` in https://github.com/ACINQ/eclair/pull/1749/commits/ee40547c48e9dafb1aac68b226ad3091029fdf18

This is something we had already discussed during #1728, I think it's worth doing (and the new `HtlcDelayedTx` type removes the ambiguity between the main output and 3rd-stage htlc outputs).

We still need to repeat these `desc` in `generateTx` in `Helpers.scala` because we want to handle `TxGenerationSkipped` cases. I think it's ok. I toyed with the idea of putting the `desc` field in companion objects of each `TransactionWithInputInfo` to have a static value, but it isn't very elegant and produces more code, so I kept the small string duplication instead.

Let me know what you think @pm47, I think it will make the logs much easier to analyze.

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
pm47: Doh! Htlcs txs are _not_ delayed relative to the commitment tx. Their _outputs_ are! Updated my comment.

t-bast:
> Most definitely. I submitted two additions on branch tx-publisher-pm. The duplication in Helpers.scala is still itching though!

I applied the first one, but I find the second one over-engineered to be honest.
I'd rather keep that duplication for now.

In order to automatically to CPFP on funding txs, we will re-work part of the `PublishRawTx`, and we'll have an opportunity to improve it then depending on what data changes are required to make that CPFP work.

pm47:
I knew you'd say that!

pm47 (approved):

PR #1750, opened 2021-04-01
Remove Electrum support

t-bast:
Electrum support was provided for mobile wallets, server nodes should always run a bitcoind node as this provides more control (especially for utxo management for anchor outputs channels).

Since wallets will use https://github.com/acinq/eclair-kmp instead of eclair, we can now remove Electrum and API fee providers from eclair.

10k lines of code removed, it's quite a feat! Eclair is now down to ~30k lines of code ;)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
```diff
@@ -43,7 +43,7 @@ sealed trait Watch {
  *
  * @param replyTo         actor to notify once the transaction is confirmed.
  * @param txId            txid of the transaction to watch.
- * @param publicKeyScript when using electrum, we need to specify a public key script; any of the output scripts should work.
+ * @param publicKeyScript output script of any of the transaction's outputs.
```
t-bast, line 46: Should I completely remove these fields? They were only necessary for Electrum.

eclair-core/src/main/resources/reference.conf
```diff
@@ -20,7 +20,7 @@ eclair {
   // override this with a script/exe that will be called everytime a new database backup has been created
   # backup-notify-script = "/absolute/path/to/script.sh"
 
-  watcher-type = "bitcoind" // other *experimental* values include "electrum"
+  watcher-type = "bitcoind"
```
pm47, line 23: We could remove this. Will we add a new watcher type in a near future?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/WatcherTypes.scala
pm47: I think yes.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1750/commits/447cba4187a56c7d974c73feb3e533d8f7369708

eclair-core/src/main/resources/reference.conf
t-bast: Agreed, we can remove it. I don't think we'll support a new watcher type in the near future.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1750/commits/9771ce6d4868ac7955f6d338dc9f70779d523f88

pm47 (approved):

pm47 (approved):

PR #1754, opened 2021-04-02
Make signing payment requests faster

sstone:
There was a much more efficient method to compute the pukey recovery id that was not used.


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -108,7 +108,7 @@ case class PaymentRequest(prefix: String, amount: Option[MilliSatoshi], timestam
    */
   def sign(priv: PrivateKey): PaymentRequest = {
     val sig64 = Crypto.sign(hash, priv)
-    val (pub1, _) = Crypto.recoverPublicKey(sig64, hash)
+    val pub1 = Crypto.recoverPublicKey(sig64, hash, 0.toByte)
```
pm47, line 111: This would be less confusing I think?

```suggestion
    // in order to tell what the recovery id is, we actually recover the pubkey ourselves and compare it to the real one
    val pub0 = Crypto.recoverPublicKey(sig64, hash, 0.toByte)
```

pm47 (approved):

t-bast (approved):
Nice, we indeed dived a bit into it when doing the `signMessage` API but didn't think to change it here.

PR #1757, opened 2021-04-07
Select channel relay at random

t-bast:
When relaying payments, we previously used the channel with the smallest balance, but this makes it easier to probe our channels and discover their balances. Selecting the channel at random among the candidates makes probing less predictable.

This was mentioned in particular in https://eprint.iacr.org/2021/384, and discussed when we initially setup that behavior.

Note however that this may be worse against spamming: the previous behavior ensured spammers would be directed towards our smaller channels first, thus not impacting our bigger channels. Protecting against spamming may be more important than protecting against probing...


t-bast:
Closing this for now, defending against spam is likely more important than defending against probing.
If users ask for this, we can support both modes in the future based on a configuration flag.

PR #1760, opened 2021-04-09
Catch all connection failures and reconnect

### LINKED ISSUE #1748: Automatic reconnection still not 100% reliable
I still have offline channels sometimes that have to be brought back online manually.

Logs when the connection died:

```
2021-03-27 20:40:16,760 INFO  f.a.e.i.ReconnectionTask CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connecting to /170.75.164.148:9735
2021-03-27 20:40:16,760 INFO  f.a.e.io.ClientSpawner - initiating new connection to nodeId=0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 origin=Actor[akka://eclair-node/deadLetters]
2021-03-27 20:40:16,760 INFO  f.acinq.eclair.io.Client CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connecting to 170.75.164.148:9735
2021-03-27 20:40:18,100 INFO  f.acinq.eclair.io.Client CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connection failed to 170.75.164.148:9735
2021-03-27 20:40:18,100 INFO  f.a.e.i.ReconnectionTask CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connection failed, next reconnection in 110 seconds

2021-03-27 20:42:08,931 INFO  f.a.e.i.ReconnectionTask CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connecting to /170.75.164.148:9735
2021-03-27 20:42:08,931 INFO  f.a.e.io.ClientSpawner - initiating new connection to nodeId=0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 origin=Actor[akka://eclair-node/deadLetters]
2021-03-27 20:42:08,931 INFO  f.acinq.eclair.io.Client CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connecting to 170.75.164.148:9735
2021-03-27 20:42:09,050 INFO  f.acinq.eclair.io.Client CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connected to 170.75.164.148:9735
2021-03-27 20:42:09,180 INFO  f.a.e.io.PeerConnection CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connection authenticated with 0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31@170.75.164.148:9735 direction=outgoing
2021-03-27 20:42:09,180 INFO  f.a.e.io.PeerConnection CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - using features=payment_secret:optional,var_onion_optin:optional,basic_mpp:optional,option_support_large_channel:optional,gossip_queries_ex:optional,option_data_loss_protect:optional,option_static_remotekey:optional,gossip_queries:optional

2021-03-27 20:42:19,199 WARN  f.a.e.io.PeerConnection CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - initialization timed out after 10 seconds
2021-03-27 20:42:19,199 INFO  f.a.e.c.TransportHandler CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - stopping (unackedReceived= unackedSent=None)
```

peer stayed offline until manual reconnection two days later:

```
2021-03-29 22:05:32,330 INFO  f.a.e.i.ReconnectionTask CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connecting to /170.75.164.148:9735
2021-03-29 22:05:32,330 INFO  f.a.e.io.ClientSpawner - initiating new connection to nodeId=0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 origin=Actor[akka://eclair-node/deadLetters]
2021-03-29 22:05:32,330 INFO  f.acinq.eclair.io.Client CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connecting to 170.75.164.148:9735
2021-03-29 22:05:32,408 INFO  f.a.eclair.router.Router SYN n:03f96dcd4861705f6c33a79c7569ba1b3d33f8177ee07bc96c606759bfc9940f80 - asking for the next slice of short_channel_ids (remaining=35/415)
2021-03-29 22:05:32,447 INFO  f.acinq.eclair.io.Client CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connected to 170.75.164.148:9735
2021-03-29 22:05:32,570 INFO  f.a.e.io.PeerConnection CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - connection authenticated with 0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31@170.75.164.148:9735 direction=outgoing
2021-03-29 22:05:32,571 INFO  f.a.e.io.PeerConnection CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - using features=payment_secret:optional,var_onion_optin:optional,basic_mpp:optional,option_support_large_channel:optional,gossip_queries_ex:optional,option_data_loss_protect:optional,option_static_remotekey:optional,gossip_queries:optional
2021-03-29 22:05:32,692 INFO  f.a.e.io.PeerConnection CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - peer is using features=payment_secret:optional,option_data_loss_protect:mandatory,var_onion_optin:optional,basic_mpp:optional,option_support_large_channel:optional,option_static_remotekey:optional,gossip_queries:optional (unknown=5), networks=
2021-03-29 22:05:32,692 INFO  f.a.e.io.PeerConnection CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - rebroadcast will be delayed by 51 seconds
2021-03-29 22:05:32,692 INFO  f.a.e.io.PeerConnection n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - sending sync channel range query with flags_opt=None replacePrevious=true
2021-03-29 22:05:32,692 INFO  f.a.eclair.router.Router n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - sending query_channel_range=QueryChannelRange(6fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d6190000000000,0,2147483647,TlvStream(List(),List()))
2021-03-29 22:05:32,692 INFO  f.a.e.i.ReconnectionTask CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 - peer is connected
2021-03-29 22:05:32,693 INFO  f.a.e.i.Peer CON n:0234227bdcd0361622f39447f961618480e7347ab1cfe6e74f561b976135343b31 c:df96b4b7537c973e5c06789bcc7fee4cae935114
```

pm47:
The `ReconnectionTask` was only catching
`ConnectionResult.Failure.ConnectionFailed`, which is a subset of
possible failures. It should instead have caught
`ConnectionResult.Failure`.

All authentication and initialization failures were not caught and
didn't trigger reconnections.

Fixes #1748.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
@@ -53,8 +53,8 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey) extends 
   startWith(IDLE, IdleData(Nothing))
 
   when(CONNECTING) {
-    case Event(_: PeerConnection.ConnectionResult.ConnectionFailed, d: ConnectingData) =>
-      log.info(s"connection failed, next reconnection in ${d.nextReconnectionDelay.toSeconds} seconds")
+    case Event(failure: PeerConnection.ConnectionResult.Failure, d: ConnectingData) =>
+      log.info(s"connection failed (reason=$failure), next reconnection in ${d.nextReconnectionDelay.toSeconds} seconds")
       setReconnectTimer(d.nextReconnectionDelay)
       goto(WAITING) using WaitingData(nextReconnectionDelay(d.nextReconnectionDelay, nodeParams.maxReconnectInterval))
 
--- a/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
@@ -30,12 +30,12 @@ import scala.concurrent.duration._
 
 class ReconnectionTaskSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with ParallelTestExecution {
 
-  val fakeIPAddress = NodeAddress.fromParts("1.2.3.4", 42000).get
-  val channels = Map(Peer.FinalChannelId(randomBytes32) -> system.deadLetters)
+  private val fakeIPAddress = NodeAddress.fromParts("1.2.3.4", 42000).get
+  private val channels = Map(Peer.FinalChannelId(randomBytes32) -> system.deadLetters)
 
-  val PeerNothingData = Peer.Nothing
-  val PeerDisconnectedData = Peer.DisconnectedData(channels)
-  val PeerConnectedData = Peer.ConnectedData(fakeIPAddress.socketAddress, system.deadLetters, null, null, channels.map { case (k: ChannelId, v) => (k, v) })
+  private val PeerNothingData = Peer.Nothing
+  private val PeerDisconnectedData = Peer.DisconnectedData(channels)
+  private val PeerConnectedData = Peer.ConnectedData(fakeIPAddress.socketAddress, system.deadLetters, null, null, channels.map { case (k: ChannelId, v) => (k, v) })
 
   case class FixtureParam(nodeParams: NodeParams, remoteNodeId: PublicKey, reconnectionTask: TestFSMRef[ReconnectionTask.State, ReconnectionTask.Data, ReconnectionTask], monitor: TestProbe)
 
@@ -166,6 +166,34 @@ class ReconnectionTaskSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike 
     // the auto reconnect kicks off again, but this time we pick up the reconnect delay where we left it
     val TransitionWithData(ReconnectionTask.IDLE, ReconnectionTask.WAITING, _, waitingData3: WaitingData) = monitor.expectMsgType[TransitionWithData]
     assert(waitingData3.nextReconnectionDelay === (waitingData0.nextReconnectionDelay * 8))
+  }
+
+  test("all kind of connection failures should be caught by the reconnection task", Tag("auto_reconnect")) { f =>
+    import f._
+
+    val peer = TestProbe()
+    nodeParams.db.peers.addOrUpdatePeer(remoteNodeId, fakeIPAddress)
+    peer.send(reconnectionTask, Peer.Transition(PeerNothingData, PeerDisconnectedData))
+    val TransitionWithData(ReconnectionTask.IDLE, ReconnectionTask.WAITING, _, _) = monitor.expectMsgType[TransitionWithData]
+    val TransitionWithData(ReconnectionTask.WAITING, ReconnectionTask.CONNECTING, _, connectingData: ReconnectionTask.ConnectingData) = monitor.expectMsgType[TransitionWithData]
+
+    val failures = List(
+      PeerConnection.ConnectionResult.ConnectionFailed(connectingData.to),
+      PeerConnection.ConnectionResult.NoAddressFound,
+      PeerConnection.ConnectionResult.InitializationFailed("incompatible features"),
+      PeerConnection.ConnectionResult.AuthenticationFailed("authentication timeout")
+    )
+
+    failures.foreach { failure =>
+      // we simulate a connection error
+      reconnectionTask ! failure
+      // a new reconnection task will be scheduled
+      val TransitionWithData(ReconnectionTask.CONNECTING, ReconnectionTask.WAITING, _, _) = monitor.expectMsgType[TransitionWithData]
+      // we send the tick manually so we don't wait
+      reconnectionTask ! ReconnectionTask.TickReconnect
+      // this triggers a reconnection
+      val TransitionWithData(ReconnectionTask.WAITING, ReconnectionTask.CONNECTING, _, _) = monitor.expectMsgType[TransitionWithData]
+    }
   }
 
   test("reconnect using the address from node_announcement") { f =>
--- a/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
@@ -207,7 +207,7 @@ class ReconnectionTaskSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike 
 
     // at this point, we are attempting to connect to the peer
     // let's assume that an incoming connection arrives from the peer right before our outgoing connection, but we haven't
-    // yes received the peer transition
+    // yet received the peer transition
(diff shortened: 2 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/ReconnectionTask.scala
```diff
@@ -53,8 +53,8 @@ class ReconnectionTask(nodeParams: NodeParams, remoteNodeId: PublicKey) extends
   startWith(IDLE, IdleData(Nothing))
 
   when(CONNECTING) {
-    case Event(_: PeerConnection.ConnectionResult.ConnectionFailed, d: ConnectingData) =>
-      log.info(s"connection failed, next reconnection in ${d.nextReconnectionDelay.toSeconds} seconds")
+    case Event(failure: PeerConnection.ConnectionResult.Failure, d: ConnectingData) =>
```
t-bast, line 56: Good catch!

But I don't think we want to trigger a reconnection on *all* failures, for example `AlreadyConnected` shouldn't trigger a reconnection otherwise we'd enter a loop...maybe we should define a `attemptReconnection` boolean on `ConnectionResult.Failure` or something like it?

pm47: Or we could make `AlreadyConnected` a `ConnectionResult.Success`, I remember flip-floping on that one.

t-bast: Good point, it may make more sense that way.

t-bast: In any case, a unit test for the `AlreadyConnected` case will be necessary :)

pm47: Actually I think we should leave it that way.

This handler is for an automated reconnections, that will be attempted if the peer is in `DISCONNECTED` state. So if we have a race here, it would be with a manual reconnect done in parallel (that's the only thing I can think of). In that case, the reconnection task will schedule a new reconnection attempt _with an increased delay_ (i.e. : no infinite loop), but then the `DISCONNECTED`->`CONNECTED` transition will make the reconnection actor move to `IDLE` and cancelling the reconnection attempt.

t-bast: Can't it simply happen because your peer is connecting to you while you're connecting to him?
In that case if they finished connecting to us, we don't want to break that connection by connecting to them after a delay, do we?

I don't feel good scheduling a new connection when our current status is `AlreadyConnected`, it sounds like something isn't right...

pm47: > we don't want to break that connection by connecting to them after a delay, do we?

We won't, because we will receive the transition in the `WAITING` state and cancel the scheduled reconnection. We will receive that transition for sure, because the only case where we send a `AlreadyConnected` is if we are indeed in the `CONNECTED` state.

> I don't feel good scheduling a new connection when our current status is AlreadyConnected, it sounds like something isn't right...

I agree, but then again, this whole reconnection logic is complicated and I'm afraid to create an obscure regression by making `AlreadyConnected` be a `Success` or creating special handling for that case.

t-bast: > We won't, because we will receive the transition in the WAITING state and cancel the scheduled reconnection. We will receive that transition for sure, because the only case where we send a AlreadyConnected is if we are indeed in the CONNECTED state.

Can you add a test to ensure that?
TBH it's really hard to ensure by just looking at the code.

pm47: Done with 098feba5d, but it's not very satisfactory, because the behavior of the peer is mocked. I'm not sure it will give you a clear view of what's happening here. In my defense, the other reconnection tests work that way, and assuming that the peer sends its state transitions is not too far fetched.

What this tests prove, is that we do not enter an infinite loop in the `AlreadyConnected` scenario. If for some reason we miss the `Disconnected`->`Connected` transition from the peer, then we have a bigger problem and all connection failures would lead us to an infinite loop (albeit with an exponential backoff delay).

eclair-core/src/test/scala/fr/acinq/eclair/io/ReconnectionTaskSpec.scala
```diff
@@ -168,6 +168,56 @@ class ReconnectionTaskSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike
...
+
+    val peer = TestProbe()
+    nodeParams.db.peers.addOrUpdatePeer(remoteNodeId, fakeIPAddress)
+    peer.send(reconnectionTask, Peer.Transition(PeerNothingData, PeerDisconnectedData))
+    val TransitionWithData(ReconnectionTask.IDLE, ReconnectionTask.WAITING, _, _) = monitor.expectMsgType[TransitionWithData]
+    val TransitionWithData(ReconnectionTask.WAITING, ReconnectionTask.CONNECTING, _, _: ReconnectionTask.ConnectingData) = monitor.expectMsgType[TransitionWithData]
+
+    // at this point, we are attempting to connect to the peer
+    // let's assume that an incoming connection arrives from the peer right before our outgoing connection, but we haven't
+    // yes received the peer transition
```
t-bast, line 210: nit:
```suggestion
    // yet received the peer transition
```

t-bast (approved):

t-bast (approved):

PR #1762, opened 2021-04-12
Posgres: fix concurrency in channels db

pm47:
The upsert pattern in `channelsDb.addOrUpdateChannel` is apparently not compatible with concurrent transactions when they are in `SERIALIZABLE` mode.

A simple test shows the issue and the following exception is thrown:
```
Canceled on identification as a pivot, during conflict out checking
```

I tried reducing the isolation level, but 1) I'm not sure what the consequences are in terms of locking the db and 2) it didn't entirely fix the issue.

As per the [postgres doc](https://www.postgresql.org/docs/current/plpgsql-control-structures.html#PLPGSQL-UPSERT-EXAMPLE), the recommended construct for the _upsert_ pattern is with the `ON CONFLICT` clause:

> It is recommended that applications use INSERT with ON CONFLICT DO
UPDATE rather than actually using this pattern.

I'm not sure about the performance of this (we mostly do updates, so we will run into the `ON CONFLICT` a lot), but let's first make it work.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgChannelsDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgChannelsDb.scala
@@ -71,7 +71,7 @@ class PgChannelsDb(implicit ds: DataSource, lock: PgLock) extends ChannelsDb wit
         """
           | INSERT INTO local_channels (channel_id, data, is_closed)
           | VALUES (?, ?, FALSE)
-          | ON CONFLICT ON CONSTRAINT local_channels_pkey
+          | ON CONFLICT (channel_id)
           | DO UPDATE SET data = EXCLUDED.data ;
           | """.stripMargin)) { statement =>
         statement.setString(1, state.channelId.toHex)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgChannelsDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgChannelsDb.scala
@@ -67,16 +67,16 @@ class PgChannelsDb(implicit ds: DataSource, lock: PgLock) extends ChannelsDb wit
   override def addOrUpdateChannel(state: HasCommitments): Unit = withMetrics("channels/add-or-update-channel", DbBackends.Postgres) {
     withLock { pg =>
       val data = stateDataCodec.encode(state).require.toByteArray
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgChannelsDb.scala
```diff
@@ -67,16 +67,16 @@ class PgChannelsDb(implicit ds: DataSource, lock: PgLock) extends ChannelsDb wit
...
-            statement.setString(1, state.channelId.toHex)
-            statement.setBytes(2, data)
-            statement.executeUpdate()
-          }
-        }
+      using(pg.prepareStatement(
+        """
+          | INSERT INTO local_channels (channel_id, data, is_closed)
+          | VALUES (?, ?, FALSE)
+          | ON CONFLICT ON CONSTRAINT local_channels_pkey
```
t-bast, line 74: Is it guaranteed that the primary key will be named `local_channels_pkey`, regardless of the postgres version?

t-bast (commented):
Why don't we have the same kind of issue with `addOrUpdatePeer`?

pm47:
> Why don't we have the same kind of issue with addOrUpdatePeer?

We probably do! Will do a pass on all other db calls.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgChannelsDb.scala
pm47: I'm _almost certain_ (famous last words) that it is the case. However we can remove all doubt and explicitly name the constraint:
```sql
CREATE TABLE IF NOT EXISTS local_channels (channel_id TEXT NOT NULL CONSTRAINT local_channels_pkey PRIMARY KEY, data BYTEA NOT NULL, ...  
```

pm47: NB: there is another variant that uses the column name, and I initially wanted to use that one, but couldn't figure out how to.

pm47: Done with 8e7409e3f.

pm47:
>> Why don't we have the same kind of issue with addOrUpdatePeer?

> We probably do! Will do a pass on all other db calls.

Done with bb93ee546.

t-bast (approved):

PR #1764, opened 2021-04-13
Make db errors fatal in channels

pm47:
We don't want a database error to cause force close of channels.
Database errors are more likely to happen when using Postgres, but can
also happen with Sqlite in case of e.g. full disk.

Since we always write to disk before sending messages, we should be able
to recover gracefully after the db issue is fixed and eclair is
restarted.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -33,14 +33,17 @@ import fr.acinq.eclair.channel.TxPublisher.{PublishRawTx, PublishTx, SetChannelI
 import fr.acinq.eclair.crypto.ShaChain
 import fr.acinq.eclair.crypto.keymanager.ChannelKeyManager
 import fr.acinq.eclair.db.PendingRelayDb
+import fr.acinq.eclair.db.pg.PgUtils.PgLock.logger
 import fr.acinq.eclair.io.Peer
 import fr.acinq.eclair.payment.PaymentSettlingOnChain
 import fr.acinq.eclair.router.Announcements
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2448,6 +2451,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       try {
         s(event)
       } catch {
+        case t: SQLException =>
+          log.error(t, "fatal database error\n")
+          sys.exit(-2)
```
t-bast, line 2456: Can we test that (probably not easy because of `sys.exit`, unless we somehow mock it)?
Are we sure that all DB errors will be wrapped in `SQLException`s?

t-bast (commented):
We need to have a way to detect that the node stopped (I'm not sure we currently do).
Perhaps we can define an alert on Kamon based on the absence of data (which indicates the node stopped)?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: I did test it manually by making the db call fail after 3 calls, and running one of the `NormalStateSpec`

pm47: Another way would be to use the same mechanism as for postgres lock:
https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala#L53-L84

But that would mean add that handler to the `ChannelFactory`, maybe a bit overkill.

t-bast (approved):

PR #1767, opened 2021-04-14
Add trampoline info to auditDB

thomash-acinq:
Add two columns for the recipient and amount sent to the recipient in case of trampoline relaying.
When using trampoline, the recipient may not be the next node on the path.


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -85,7 +85,7 @@ sealed trait PaymentRelayed extends PaymentEvent {
 
 case class ChannelPaymentRelayed(amountIn: MilliSatoshi, amountOut: MilliSatoshi, paymentHash: ByteVector32, fromChannelId: ByteVector32, toChannelId: ByteVector32, timestamp: Long = System.currentTimeMillis) extends PaymentRelayed
 
-case class TrampolinePaymentRelayed(paymentHash: ByteVector32, incoming: PaymentRelayed.Incoming, outgoing: PaymentRelayed.Outgoing, timestamp: Long = System.currentTimeMillis) extends PaymentRelayed {
+case class TrampolinePaymentRelayed(paymentHash: ByteVector32, incoming: PaymentRelayed.Incoming, outgoing: PaymentRelayed.Outgoing, recipientNodeId: PublicKey, recipientAmount: MilliSatoshi, timestamp: Long = System.currentTimeMillis) extends PaymentRelayed {
```
t-bast, line 88: To be more future-proof, I'd rename these new fields:

```suggestion
case class TrampolinePaymentRelayed(paymentHash: ByteVector32, incoming: PaymentRelayed.Incoming, outgoing: PaymentRelayed.Outgoing, nextTrampolineNodeId: PublicKey, nextTrampolineAmount: MilliSatoshi, timestamp: Long = System.currentTimeMillis) extends PaymentRelayed {
```

When doing a trampoline relay, we make what looks like from our point of view an E2E payment to that next trampoline node, but he may not be the real final recipient (he's only a final recipient from our limited point of view).

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -75,26 +76,38 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
       statement.executeUpdate("CREATE INDEX IF NOT EXISTS relayed_payment_hash_idx ON relayed(payment_hash)")
     }
 
+    def migration45(statement: Statement): Int = {
+      statement.executeUpdate("ALTER TABLE relayed ADD recipientNodeId BLOB")
+      statement.executeUpdate("ALTER TABLE relayed ADD recipientAmount INTEGER")
```
t-bast, line 81: IIRC there was a limitation of SQLite that prevented us from simply doing an `ALTER TABLE` to add new columns, which is why we've done this weird dance in other migrations to re-create the table and copy old rows.

I'm not sure if it's a limitation on *all* versions of SQLite or only the one we use on mobile, but it's safer to do something similar to previous migrations.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -38,9 +39,9 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
   import ExtendedResultSet._
 
   val DB_NAME = "audit"
-  val CURRENT_VERSION = 4
+  val CURRENT_VERSION = 5
 
-  case class RelayedPart(channelId: ByteVector32, amount: MilliSatoshi, direction: String, relayType: String, timestamp: Long)
+  case class RelayedPart(channelId: ByteVector32, amount: MilliSatoshi, direction: String, relayType: String, recipientNodeId: Option[PublicKey], recipientAmount: Option[MilliSatoshi], timestamp: Long)
```
t-bast, line 44: It bothers me a bit that:

- this is only set for trampoline and `None` for channel relays
- this is duplicated when using MPP (all parts in a trampoline payment will have the same value here)

IIRC we already hesitated in the past to add this info, but chose to omit it, @pm47 do you remember why?

thomash-acinq: I agree, I've changed the design to put all the new data in a separate table. Now there are no empty columns for channel relays and no duplication.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -282,11 +319,14 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
             case Some(RelayedPart(_, _, _, "channel", timestamp)) => incoming.zip(outgoing).map {
               case (in, out) => ChannelPaymentRelayed(in.amount, out.amount, paymentHash, in.channelId, out.channelId, timestamp)
             }
-            case Some(RelayedPart(_, _, _, "trampoline", timestamp)) => TrampolinePaymentRelayed(paymentHash, incoming, outgoing, timestamp) :: Nil
+            case Some(RelayedPart(_, _, _, "trampoline", timestamp)) =>
+              val (nextTrampolineAmount, nextTrampolineNodeId) = trampolineByHash.getOrElse(paymentHash, (0 msat, PublicKey(hex"020000000000000000000000000000000000000000000000000000000000000000")))
```
t-bast, line 323: nit: there's a `PlaceHolderPubKey` that you can use in `Transactions.scala`, you can either import it here or use `PrivateKey(ByteVector32.One).publicKey` directly if you prefer

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -38,18 +40,28 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
...
   val DB_NAME = "audit"
-  val CURRENT_VERSION = 4
+  val CURRENT_VERSION = 5
 
   case class RelayedPart(channelId: ByteVector32, amount: MilliSatoshi, direction: String, relayType: String, timestamp: Long)
 
   inTransaction { pg =>
     using(pg.createStatement()) { statement =>
+      def migration45(statement: Statement): Int = {
+        statement.executeUpdate("CREATE TABLE IF NOT EXISTS relayed_trampoline (payment_hash TEXT NOT NULL, amount_msat BIGINT NOT NULL, next_node_id TEXT NOT NULL, timestamp BIGINT NOT NULL)")
```
t-bast, line 50: @pm47 said something recently about not using `BIGINT` for timestamps in postgres, not sure if that's something you should change in this PR or whether he'll do a sweep of all the codebase to update it.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -256,7 +289,9 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
               case Some(RelayedPart(_, _, _, "channel", timestamp)) => incoming.zip(outgoing).map {
                 case (in, out) => ChannelPaymentRelayed(in.amount, out.amount, paymentHash, in.channelId, out.channelId, timestamp)
               }
-              case Some(RelayedPart(_, _, _, "trampoline", timestamp)) => TrampolinePaymentRelayed(paymentHash, incoming, outgoing, timestamp) :: Nil
+              case Some(RelayedPart(_, _, _, "trampoline", timestamp)) =>
+                val (nextTrampolineAmount, nextTrampolineNodeId) = trampolineByHash.getOrElse(paymentHash, (0 msat, PublicKey(hex"020000000000000000000000000000000000000000000000000000000000000000")))
```
t-bast, line 293: nit: same as sqlite, you can use a `PlaceHolderPubKey` here

t-bast (commented):
LGTM, only a few nits.
@pm47 has been working on the DB recently, let's wait to see for his review as well.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
pm47: The limitation was that below 3.25.0, sqlite didn't allow column renaming. But we use 3.27.2.1 so we should be good.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
pm47: Indeed, but it'll take some time so let's use `BIGINT` here and we'll migrate all of them in one pass.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
thomash-acinq: Done.

t-bast (approved):
:rocket:

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
thomash-acinq: It's unrelated and everything else still uses `BIGINT`. We should change all the timestamps together if we decide we want to do it.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
pm47: There is a potential source of confusion here because now we have a `relayed` table and `relayed_trampoline` table, but both of them contain data related to trampoline payments. The `relayed` table is more granular and refers to individual in/out transfers at the channel level, that make full relays when aggregated. Can't think of a better name though.

> IIRC we already hesitated in the past to add this info, but chose to omit it, @pm47 do you remember why?

🤷‍♂️

pm47 (approved):

PR #1770, opened 2021-04-15
Index trampoline payments by hash and secret

### LINKED ISSUE #1723: NodeRelayer should index payments by `payment_hash || payment_secret`
Right now, the `NodeRelayer` indexes child actors by `payment_hash` only.
It means that senders must know beforehand the total amount they will send through each trampoline node.
This works fine for single-trampoline scenarios, but will not work with multi-trampoline scenarios: if the payer splits his payment between several trampoline routes and one fails, he may end up sending more additional HTLCs to a previous route when retrying.
Right now, we will reject them because the `payment_secret` doesn't match, or we'll simply fail to relay them (if the `payment_secret` is reused) and wait for a timeout from the final recipient.

We should instead index by `payment_hash || payment_secret`, but that requires extracting the `payment_secret` a bit higher up in our call stack.

t-bast:
We need to group incoming HTLCs together by `payment_hash` *and* `payment_secret`, otherwise we will reject valid payments that are split into multiple distinct trampoline parts (same `payment_hash` but different `payment_secret`).

Thanks @ecdsa for pointing out this issue.

Fixes #1723

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
@@ -24,6 +24,7 @@ import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.eclair.channel.{CMD_FAIL_HTLC, CMD_FULFILL_HTLC}
 import fr.acinq.eclair.db.PendingRelayDb
+import fr.acinq.eclair.payment.IncomingPacket.NodeRelayPacket
 import fr.acinq.eclair.payment.Monitoring.{Metrics, Tags}
 import fr.acinq.eclair.payment.OutgoingPacket.Upstream
 import fr.acinq.eclair.payment._
@@ -75,13 +76,29 @@ object NodeRelay {
     }
   }
 
-  def apply(nodeParams: NodeParams, parent: akka.actor.typed.ActorRef[NodeRelayer.Command], register: ActorRef, relayId: UUID, paymentHash: ByteVector32, outgoingPaymentFactory: OutgoingPaymentFactory): Behavior[Command] =
+  def apply(nodeParams: NodeParams,
+            parent: akka.actor.typed.ActorRef[NodeRelayer.Command],
+            register: ActorRef,
+            relayId: UUID,
+            nodeRelayPacket: NodeRelayPacket,
+            paymentSecret: ByteVector32,
+            outgoingPaymentFactory: OutgoingPaymentFactory): Behavior[Command] =
     Behaviors.setup { context =>
+      val paymentHash = nodeRelayPacket.add.paymentHash
+      val totalAmountIn = nodeRelayPacket.outerPayload.totalAmount
       Behaviors.withMdc(Logs.mdc(
         category_opt = Some(Logs.LogCategory.PAYMENT),
         parentPaymentId_opt = Some(relayId), // for a node relay, we use the same identifier for the whole relay itself, and the outgoing payment
         paymentHash_opt = Some(paymentHash))) {
-        new NodeRelay(nodeParams, parent, register, relayId, paymentHash, context, outgoingPaymentFactory)()
+        context.log.info("relaying payment relayId={}", relayId)
+        val mppFsmAdapters = {
+          context.messageAdapter[MultiPartPaymentFSM.ExtraPaymentReceived[HtlcPart]](WrappedMultiPartExtraPaymentReceived)
+          context.messageAdapter[MultiPartPaymentFSM.MultiPartPaymentFailed](WrappedMultiPartPaymentFailed)
+          context.messageAdapter[MultiPartPaymentFSM.MultiPartPaymentSucceeded](WrappedMultiPartPaymentSucceeded)
+        }.toClassic
+        val incomingPaymentHandler = context.actorOf(MultiPartPaymentFSM.props(nodeParams, paymentHash, totalAmountIn, mppFsmAdapters))
+        new NodeRelay(nodeParams, parent, register, relayId, paymentHash, paymentSecret, context, outgoingPaymentFactory)
+          .receiving(Queue.empty, nodeRelayPacket.innerPayload, nodeRelayPacket.nextPacket, incomingPaymentHandler)
       }
     }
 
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -144,66 +161,37 @@ class NodeRelay private(nodeParams: NodeParams,
...
-  private def receiving(htlcs: Queue[UpdateAddHtlc], secret: ByteVector32, nextPayload: Onion.NodeRelayPayload, nextPacket: OnionRoutingPacket, handler: ActorRef): Behavior[Command] =
+  private def receiving(htlcs: Queue[UpdateAddHtlc], nextPayload: Onion.NodeRelayPayload, nextPacket: OnionRoutingPacket, handler: ActorRef): Behavior[Command] =
     Behaviors.receiveMessagePartial {
       case Relay(IncomingPacket.NodeRelayPacket(add, outer, _, _)) => outer.paymentSecret match {
+        // TODO: @pm: maybe those checks should be done by the mpp FSM?
         case None =>
-          context.log.warn("rejecting htlcId={}: missing payment secret", add.id)
+          context.log.warn("rejecting htlc #{} from channel {}: missing payment secret", add.id, add.channelId)
           rejectHtlc(add.id, add.channelId, add.amountMsat)
           Behaviors.same
```
pm47, line 186: Why keep this? We know it can't happen right?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -144,66 +161,37 @@ class NodeRelay private(nodeParams: NodeParams,
...
-        case Some(incomingSecret) if incomingSecret != secret =>
-          context.log.warn("rejecting htlcId={}: payment secret doesn't match other HTLCs in the set", add.id)
+        case Some(incomingSecret) if incomingSecret != paymentSecret =>
+          context.log.warn("rejecting htlc #{} from channel {}: payment secret doesn't match other HTLCs in the set", add.id, add.channelId)
           rejectHtlc(add.id, add.channelId, add.amountMsat)
           Behaviors.same
```
pm47, line 190: This cannot happen anymore either?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -48,34 +51,47 @@ object NodeRelayer {
...
-                context.log.debug(s"spawning a new handler with relayId=$relayId")
-                val outgoingPaymentFactory = NodeRelay.SimpleOutgoingPaymentFactory(nodeParams, router, register)
-                val handler = context.spawn(NodeRelay.apply(nodeParams, context.self, register, relayId, paymentHash, outgoingPaymentFactory), relayId.toString)
-                context.log.debug("forwarding incoming htlc to new handler")
-                handler ! NodeRelay.Relay(nodeRelayPacket)
-                apply(nodeParams, router, register, children + (paymentHash -> handler))
+                context.log.warn("rejecting htlc #{} from channel {}: missing payment secret", htlcIn.id, htlcIn.channelId)
+                val failureMessage = IncorrectOrUnknownPaymentDetails(htlcIn.amountMsat, nodeParams.currentBlockHeight)
+                val cmd = CMD_FAIL_HTLC(htlcIn.id, Right(failureMessage), commit = true)
+                PendingRelayDb.safeSend(register, nodeParams.db.pendingRelay, htlcIn.channelId, cmd)
```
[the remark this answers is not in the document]
t-bast: I think it would make sense, yes. It's been introduced in December 2019, so we can probably consider them mandatory now.
When we make that change, we should set the feature bit as mandatory. I added that to my TODO-list.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
```diff
@@ -29,16 +32,16 @@ import java.util.UUID
...
+ * It doesn't do the job itself, instead it dispatches each individual payment (which can be multi-in, multi-out) to a
+ * child actor of type [[NodeRelay]].
  */
 object NodeRelayer {
 
   // @formatter:off
   sealed trait Command
   case class Relay(nodeRelayPacket: IncomingPacket.NodeRelayPacket) extends Command
-  case class RelayComplete(childHandler: ActorRef[NodeRelay.Command], paymentHash: ByteVector32) extends Command
+  case class RelayComplete(childHandler: ActorRef[NodeRelay.Command], paymentHash: ByteVector32, paymentSecret: ByteVector32) extends Command
```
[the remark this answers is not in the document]
t-bast: I hesitated, it forces the `NodeRelay` child to know about a type that could be private to `NodeRelayer`...

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -75,13 +76,29 @@ object NodeRelay {
     }
   }
 
-  def apply(nodeParams: NodeParams, parent: akka.actor.typed.ActorRef[NodeRelayer.Command], register: ActorRef, relayId: UUID, paymentHash: ByteVector32, outgoingPaymentFactory: OutgoingPaymentFactory): Behavior[Command] =
+  def apply(nodeParams: NodeParams,
+            parent: akka.actor.typed.ActorRef[NodeRelayer.Command],
+            register: ActorRef,
+            relayId: UUID,
+            nodeRelayPacket: NodeRelayPacket,
+            paymentSecret: ByteVector32,
```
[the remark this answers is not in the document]
t-bast: Definitely, will be cleaned up in a future change. I'll discuss it with other implementations as well to see where they stand on making it mandatory.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -144,66 +161,37 @@ class NodeRelay private(nodeParams: NodeParams,
...
-    context.messageAdapter[PaymentSent](WrappedPaymentSent)
-    context.messageAdapter[PaymentFailed](WrappedPaymentFailed)
-  }.toClassic
-
-  def apply(): Behavior[Command] =
-    Behaviors.receiveMessagePartial {
-      // We make sure we receive all payment parts before forwarding to the next trampoline node.
-      case Relay(IncomingPacket.NodeRelayPacket(add, outer, inner, next)) => outer.paymentSecret match {
-        case None =>
-          // TODO: @pm: maybe those checks should be done later in the flow (by the mpp FSM?)
```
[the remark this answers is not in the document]
t-bast: IIUC that comment was alluding to the fact that the MPP FSM doesn't check the `payment_secret` at all, and assumes it has been checked earlier. We *could* change that assumption in the MPP FSM, but I'd rather make `payment_secret` mandatory and simplify this altogether.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelayer.scala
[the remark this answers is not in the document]
pm47: Yes, let's forget it.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
t-bast: Yes, but at the type level it could (if we change the `NodeRelayer` code without updating the `NodeRelay`), so it's kind of a defense-in-depth to keep this handler (for now, let's get rid of all of them soon!)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
t-bast: Same reason, it doesn't cost much to handle this here and is a defense-in-depth, this way we don't need to make assumptions about how our parent behaves

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: Another way to deal with that is to add a `require` when instantiating the actor, and then do `outer.paymentSecret.get` everywhere?

t-bast: I'll try that

t-bast: I'm not a big fan, it's less robust since we can cleanly handle this case in just a few lines of code without any risk of crashing the actor. I'd rather wait to make `payment_secret` mandatory to simplify this.

pm47: True, it's a bit dirty.

pm47 (approved):

PR #1771, opened 2021-04-15
Use current timestamp in invoice preimage

t-bast:
In case our random number generator completely fails, we may end up generating the same payment preimage for unrelated payments.

To mitigate that, we use the current time at the end of the preimage.
This still leaves at least 192 bits of randomness which is more than enough.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
@@ -19,7 +19,7 @@ package fr.acinq.eclair.payment
 import fr.acinq.bitcoin.Crypto.{PrivateKey, PublicKey}
 import fr.acinq.bitcoin.{Base58, Base58Check, Bech32, Block, ByteVector32, ByteVector64, Crypto}
 import fr.acinq.eclair.payment.PaymentRequest._
-import fr.acinq.eclair.{CltvExpiryDelta, FeatureSupport, Features, MilliSatoshi, MilliSatoshiLong, NodeParams, ShortChannelId, randomBytes32}
+import fr.acinq.eclair.{CltvExpiryDelta, FeatureSupport, Features, MilliSatoshi, MilliSatoshiLong, NodeParams, ShortChannelId, randomBytes, randomBytes32}
 import scodec.bits.{BitVector, ByteOrdering, ByteVector}
 import scodec.codecs.{list, ubyte}
 import scodec.{Codec, Err}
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -484,6 +484,17 @@ object PaymentRequest {
...
 
+  /**
+   * To avoid collision on payment preimages, we use a mix of randomness and timestamping.
+   * The end of the preimage contains a timestamp and the rest (at least 192 bits) comes from our RNG.
+   * This ensures that even if the randomness fails, it's unlikely that preimages will be reused.
+   */
+  def generatePreimage(): ByteVector32 = {
+    val timestamp = BigInt(System.currentTimeMillis()).toByteArray
+    val random = randomBytes(32 - timestamp.length)
+    ByteVector32(random ++ ByteVector.view(timestamp))
```
pm47, line 495: How about xoring instead of concatenating? That preserves the entropy, and avoids leaking data about when the preimage was generated.

```suggestion
    val timestamp = ByteVector.fromLong(System.nanoTime()).padTo(32)
    val random = randomBytes32
    ByteVector32(random xor timestamp)
```

t-bast: That was my first idea, but then I thought this is better for the following reasons:

1. If you `xor`, you completely lose the timestamp information, whereas it may be handy to spot instantly when looking at a preimage (it's not secret at all, the `timestamp` is included in the invoice so there's no reason to hide it)
2. 192 bits of entropy is already way more than enough for a payment preimage, we really don't need to preserve these bits of entropy
3. If you xor the timestamp with random bytes, there is a (small) chance of creating a collision whereas there's 0 chance of collision if you use the plain timestamp (unless your clock is stuck)

I'm not set in stone on that one though, but I currently prefer using a plain timestamp.

pm47: The timestamp is known to the sender, but not the intermediate nodes? Now they can tell when relaying a payment that it was generated 5 minutes or 2 hours ago.

I agree with you on entropy.

t-bast: You're right, we don't want to give intermediate node a way of knowing a payment was made to a specific wallet implementation by watching the `update_fulfill_htlc`.

t-bast:
Closing in favor of #1774 which should address entropy issues.

PR #1773, opened 2021-04-19
More database nits

pm47:
Commits are independent and are best reviewed separately.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
@@ -177,7 +177,7 @@ object PgUtils extends JdbcUtils {
         using(connection.createStatement()) {
           statement =>
             // allow only one row in the ownership lease table
-            statement.executeUpdate(s"CREATE TABLE IF NOT EXISTS $LeaseTable (id INTEGER PRIMARY KEY default(1), expires_at TIMESTAMP NOT NULL, instance VARCHAR NOT NULL, CONSTRAINT one_row CHECK (id = 1))")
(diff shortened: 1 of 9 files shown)


pm47:
Rebased.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
```diff
@@ -181,10 +183,15 @@ object PgUtils extends JdbcUtils {
...
       }
 
-      private def acquireExclusiveTableLock()(implicit connection: Connection): Unit = {
+      private def acquireExclusiveTableLock(lockTimeout: FiniteDuration)(implicit connection: Connection): Unit = {
         using(connection.createStatement()) {
           statement =>
-            statement.executeUpdate(s"LOCK TABLE $LeaseTable IN ACCESS EXCLUSIVE MODE NOWAIT")
+            // We use a timeout here, because we might not be able to get the lock right away due to concurrent access
+            // by other threads. That timeout gives time for other transactions to complete, then ours can take the lock
+            statement.executeUpdate(s"SET lock_timeout TO '${lockTimeout.toSeconds}s'")
```
t-bast, line 191: This is redone every time we re-acquire the lock, whereas we only need to do it once, right?
Does it make sense to do this somewhere else (when starting eclair) instead of redoing this (no-op) statement regularly?
If it's clearer to keep it here, don't bother fixing this.

pm47: Done with 7a5bf8836.

It is less intrusive and more performant, but less obvious because the cause and the effect are separated in two different files. I guess it's okay.

t-bast (approved):

PR #1774, opened 2021-04-19
Add additional PRNG

t-bast:
In case of catastrophic failures of the `SecureRandom` instance, we add a secondary randomness source that we mix into the random stream.

This is a somewhat weak random source and should not be used on its own, but it doesn't hurt to xor it with the output of `SecureRandom`.

The second commit should be reviewed independently: it simply adds parenthesis to functions that generate random things, since these have side-effects. It's trivial but painful to review since it touches a lot of test files.

NB: this feature make a lot of sense in https://github.com/acinq/lightning-kmp as well since mobile platforms may have more common randomness failures.


eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -30,11 +31,24 @@ package object eclair {
...
 
   def randomBytes(length: Int): ByteVector = {
     val buffer = new Array[Byte](length)
     secureRandom.nextBytes(buffer)
+    val weakBuffer = new Array[Byte](length)
+    weakRandom.nextBytes(weakBuffer)
+    for (i <- buffer.indices) {
+      buffer(i) = (buffer(i) ^ weakBuffer(i)).toByte
+    }
     ByteVector.view(buffer)
```
pm47, line 52:
```suggestion
    val rand1 = {
      val buffer = new Array[Byte](length)
      secureRandom.nextBytes(buffer)
      ByteVector.view(buffer)
    }
    val rand2 = {
      val weakBuffer = new Array[Byte](length)
      weakRandom.nextBytes(weakBuffer)
      ByteVector.view(weakBuffer)
    }
    rand1 xor rand2
```

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -44,6 +58,10 @@ package object eclair {
 
   def randomKey: PrivateKey = PrivateKey(randomBytes32)
 
+  def randomLong(): Long = {
```
pm47, line 61: It made sense to put flat method calls in the package object when we just called basic methods on `secureRandom`, but now I think we should probably define a class that encapsulates all those xor calls, and hides `secureRandom` and `weakRandom` which are not supposed to be called directly.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Autoprobe.scala
```diff
@@ -96,7 +96,7 @@ object Autoprobe {
     if (peers.isEmpty) {
       None
     } else {
-      peers.drop(secureRandom.nextInt(peers.size)).headOption
+      peers.drop(randomLong().toInt % peers.size).headOption
```
pm47, line 99: We should define a `randomInt(bound: Int)`.

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -30,11 +31,24 @@ package object eclair {
...
-  val secureRandom = new SecureRandom()
+  private val secureRandom = new SecureRandom()
+
+  def initRandom(): Unit = secureRandom.nextInt()
+
+  /**
+   * We're using an additional, weaker randomness source to protect against catastrophic failures of the SecureRandom
+   * instance.
+   */
+  val weakRandom = new WeakRandom()
```
pm47, line 42: We should find a way to not make it possible to access `weakRandom.nextBytes` so easily. I understand that `weakRandom.addEntropy()` needs to stay public.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Autoprobe.scala
t-bast: This is painful to do if we want an unbiased random, and to be honest we don't need it (the `Autoprobe` is unused or doesn't care about bias). I'd rather add it only if we ever need it...

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
t-bast: Easy, I can export different traits for each purpose

t-bast: Done in https://github.com/ACINQ/eclair/pull/1774/commits/0df44d9d5a3f6beac430640e213f8603185f57f0

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1774/commits/0df44d9d5a3f6beac430640e213f8603185f57f0

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
t-bast: This requires two byte array allocation for no reason, we can reuse at least one of them.
And xoring two `ByteVector` has to be less efficient than just iterating over the array to xor each byte individually.

t-bast: TBH I think this is really clear enough with the current code, and there's no risk of mutating the `ByteVector`, it's only applied after xor-ing everything.

pm47 (changes requested):

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -44,6 +40,8 @@ package object eclair {
 
   def randomKey: PrivateKey = PrivateKey(randomBytes32)
 
+  def randomLong: Long = randomGen.nextLong()
```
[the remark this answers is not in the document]
t-bast: That's right, I get turned off because IntelliJ prompts me to remove the parenthesis, but they do make sense.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/WeakEntropyPool.scala
```diff
@@ -0,0 +1,97 @@
...
+      case WrappedPeerConnected(nodeId) => collecting(collector, Some(collect(entropy_opt, nodeId.value ++ ByteVector.fromLong(System.currentTimeMillis()))))
+
+      case WrappedNodeUpdated(sig) => collecting(collector, Some(collect(entropy_opt, sig ++ ByteVector.fromLong(System.currentTimeMillis()))))
+
+      case WrappedChannelSignature(wtxid) => collecting(collector, Some(collect(entropy_opt, wtxid ++ ByteVector.fromLong(System.currentTimeMillis()))))
+    }
+  }
+
+  private def collect(entropy_opt: Option[ByteVector32], additional: ByteVector): ByteVector32 = {
+    Crypto.sha256(entropy_opt.map(_.bytes).getOrElse(ByteVector.empty) ++ additional)
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1774/commits/54e961e5ba0f6be72d73211c47092989d4c02df9

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
```diff
@@ -18,23 +18,19 @@ package fr.acinq
...
 import scala.util.{Failure, Success, Try}
 
 package object eclair {
 
-  /**
-   * We are using 'new SecureRandom()' instead of 'SecureRandom.getInstanceStrong()' because the latter can hang on Linux
-   * See http://bugs.java.com/view_bug.do?bug_id=6521844 and https://tersesystems.com/2015/12/17/the-right-way-to-use-securerandom/
-   */
-  val secureRandom = new SecureRandom()
+  def randomGen = new StrongRandom()
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1774/commits/54e961e5ba0f6be72d73211c47092989d4c02df9

eclair-core/src/main/scala/fr/acinq/eclair/package.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1774/commits/1e1037f13dc7744449dda518e0cb3fa0c280e6e6
I kept it in a separate commit that does only this, since it's trivial to review but touches a lot of test files (which is painful to read).

eclair-core/src/main/scala/fr/acinq/eclair/crypto/keymanager/LocalChannelKeyManager.scala
```diff
@@ -75,7 +75,7 @@ class LocalChannelKeyManager(seed: ByteVector, chainHash: ByteVector32) extends
   override def newFundingKeyPath(isFunder: Boolean): KeyPath = {
     val last = DeterministicWallet.hardened(if (isFunder) 1 else 0)
 
-    def next(): Long = secureRandom.nextInt() & 0xFFFFFFFFL
+    def next(): Long = randomLong & 0xFFFFFFFFL
```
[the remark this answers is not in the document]
sstone: LGTM

pm47 (dismissed):
LGTM.

According to [this](https://security.stackexchange.com/a/47461), XORing two pseudorandom data sources cannot reduce randomness, as long as the two sources are unrelated to each other. This is the case here because we know exactly how the `WeakRandom` generator works.

pm47:
Let's wait for a full review from @sstone before merging.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Random.scala
```diff
@@ -0,0 +1,172 @@
...
+      // NB: init internally resets the engine, no need to reset it explicitly ourselves.
+      stream.init(true, new ParametersWithIV(new KeyParameter(seed), new Array[Byte](12)))
+    }
+  }
+
+  def nextBytes(bytes: Array[Byte]): Unit = synchronized {
+    if (shouldSample()) {
+      sampleEntropy()
+    }
+    stream.processBytes(bytes, 0, bytes.length, bytes, 0)
```
sstone, line 126: Is it safe to use `bytes` here as both input and output to  `processBytes()` ?

t-bast: It should be how we do in-place stream encryption (without needing to allocate a new array for no reason).
But this is definitely an area where you should double-check that I didn't miss something!

t-bast: I can add a test vector to verify that this does what we expect.

t-bast: I added a test case in https://github.com/ACINQ/eclair/pull/1774/commits/9fb8417d1b8b85a9c63904768f599a5ad82fb5e2 to prove this works as expected.

sstone (approved):

PR #1775, opened 2021-04-19
Rework the db version management

pm47:
(This part of the code is sensitive, but not critical, because bugs
should be obvious, and their impact should be limited to the app
refusing to start.)

~~No SQL code has been modified, this is just a refactoring.~~
edit: second commit does remove `IF NOT EXISTS` statements.

The way our `getVersion`/`setVersion` worked was very unintuitive
and led to comments like the following in tests:
```scala
dbs.getVersion(statement, "network", 1) // this will set version to 1
```

The reason is that we treat unitialized databases and up-to-date
databases exactly the same way. That is why a `getVersion` will set the
version to the most up-to-date if none is found. It's also why we use
`CREATE IF NOT EXISTS` statements for tables and indices.

I think that is misleading and we should have a `getVersion` that
only, you know, _gets_ the version (or returns `None` if no version
exists). We would then treat unitialized databases and up-to-date
databases differently.

Also, there was some variations in the layout between files, which have
been fixed. There is also less repetition.

I'm curious to have your opinion on this @thomash-acinq, since you had to deal with
db migration recently.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
@@ -16,15 +16,15 @@
 
 package fr.acinq.eclair.db.jdbc
 
-import java.sql.{Connection, ResultSet, Statement}
-import java.util.UUID
-
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.eclair.MilliSatoshi
-import javax.sql.DataSource
+import org.sqlite.SQLiteConnection
 import scodec.Codec
 import scodec.bits.{BitVector, ByteVector}
 
+import java.sql.{Connection, ResultSet, Statement}
+import java.util.UUID
+import javax.sql.DataSource
 import scala.collection.immutable.Queue
 
 trait JdbcUtils {
@@ -57,6 +57,42 @@ trait JdbcUtils {
     } finally {
       if (inTransaction) statement.getConnection.setAutoCommit(autoCommit)
       if (statement != null) statement.close()
+    }
+  }
+
+  private def createVersionTable(statement: Statement): Unit = {
+    statement.executeUpdate("CREATE TABLE IF NOT EXISTS versions (db_name TEXT NOT NULL PRIMARY KEY, version INTEGER NOT NULL)")
+  }
+
+  /**
+   * Several logical databases (channels, network, peers) may be stored in the same physical sqlite database.
+   * We keep track of their respective version using a dedicated table. The version entry will be created if
+   * there is none but will never be updated here (use setVersion to do that).
+   */
+  def getVersion(statement: Statement, db_name: String): Option[Int] = {
+    createVersionTable(statement)
+    // if there was a previous version installed, this will return a different value from current version
+    val rs = statement.executeQuery(s"SELECT version FROM versions WHERE db_name='$db_name'")
+    if (rs.next()) Some(rs.getInt("version")) else None
+  }
+
+  /**
+   * Updates the version for a particular logical database, it will overwrite the previous version.
+   *
+   * NB: we could define this method in [[fr.acinq.eclair.db.sqlite.SqliteUtils]] and [[fr.acinq.eclair.db.pg.PgUtils]]
+   *     but it would make testing more complicated because we need to use one or the other depending on the backend.
+   */
+  def setVersion(statement: Statement, db_name: String, newVersion: Int): Unit = {
+    createVersionTable(statement)
+    statement.getConnection match {
+      case _: SQLiteConnection =>
+        // if there was no version for the current db, then insert the current version
+        statement.executeUpdate(s"INSERT OR IGNORE INTO versions VALUES ('$db_name', $newVersion)")
+        // if there was an existing version, then previous step was a no-op, now we overwrite the existing version
+        statement.executeUpdate(s"UPDATE versions SET version=$newVersion WHERE db_name='$db_name'")
+      case _ => // if it isn't an sqlite connection, we assume it's postgres
+        // insert or update the version
+        statement.executeUpdate(s"INSERT INTO versions VALUES ('$db_name', $newVersion) ON CONFLICT (db_name) DO UPDATE SET version = EXCLUDED.version ;")
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
@@ -65,7 +65,7 @@ trait JdbcUtils {
   }
 
   /**
-   * Several logical databases (channels, network, peers) may be stored in the same physical sqlite database.
+   * Several logical databases (channels, network, peers) may be stored in the same physical database.
    * We keep track of their respective version using a dedicated table. The version entry will be created if
    * there is none but will never be updated here (use setVersion to do that).
    */
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
@@ -46,18 +46,14 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
 
   inTransaction { pg =>
     using(pg.createStatement()) { statement =>
-      def migration45(statement: Statement): Int = {
+      def migration45(statement: Statement): Unit = {
         statement.executeUpdate("CREATE TABLE IF NOT EXISTS relayed_trampoline (payment_hash TEXT NOT NULL, amount_msat BIGINT NOT NULL, next_node_id TEXT NOT NULL, timestamp BIGINT NOT NULL)")
         statement.executeUpdate("CREATE INDEX IF NOT EXISTS relayed_trampoline_timestamp_idx ON relayed_trampoline(timestamp)")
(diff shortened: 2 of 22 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
```diff
@@ -60,6 +60,42 @@ trait JdbcUtils {
...
+    // if there was a previous version installed, this will return a different value from current version
+    val rs = statement.executeQuery(s"SELECT version FROM versions WHERE db_name='$db_name'")
+    if (rs.next()) Some(rs.getInt("version")) else None
+  }
+
+  /**
+   * Updates the version for a particular logical database, it will overwrite the previous version.
+   *
+   * NB: we could define this method in [[fr.acinq.eclair.db.sqlite.SqliteUtils]] and [[fr.acinq.eclair.db.pg.PgUtils]]
+   *     but it would make testing more complicated because we need to use one or the other depending on the backend.
```
pm47, line 83: This design choice is debatable but I think it's worthwhile, less duplication and simpler tests.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -46,18 +46,14 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
 
   inTransaction { pg =>
     using(pg.createStatement()) { statement =>
-      def migration45(statement: Statement): Int = {
+      def migration45(statement: Statement): Unit = {
         statement.executeUpdate("CREATE TABLE IF NOT EXISTS relayed_trampoline (payment_hash TEXT NOT NULL, amount_msat BIGINT NOT NULL, next_node_id TEXT NOT NULL, timestamp BIGINT NOT NULL)")
         statement.executeUpdate("CREATE INDEX IF NOT EXISTS relayed_trampoline_timestamp_idx ON relayed_trampoline(timestamp)")
         statement.executeUpdate("CREATE INDEX IF NOT EXISTS relayed_trampoline_payment_hash_idx ON relayed_trampoline(payment_hash)")
```
thomash-acinq, line 52: Don't you want to remove the `IF NOT EXISTS`? If they do exist then it's a bug.

pm47: Exactly, done here 945ada214. I didn't want to put SQL changes in the first commit.

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
```diff
@@ -60,6 +60,42 @@ trait JdbcUtils {
...
+    createVersionTable(statement)
+    statement.getConnection match {
+      case _: SQLiteConnection =>
+        // if there was no version for the current db, then insert the current version
+        statement.executeUpdate(s"INSERT OR IGNORE INTO versions VALUES ('$db_name', $newVersion)")
+        // if there was an existing version, then previous step was a no-op, now we overwrite the existing version
+        statement.executeUpdate(s"UPDATE versions SET version=$newVersion WHERE db_name='$db_name'")
+      case _ => // if it isn't an sqlite connection, we assume it's postgres
+        // insert or update the version
+        statement.executeUpdate(s"INSERT INTO versions VALUES ('$db_name', $newVersion) ON CONFLICT (db_name) DO UPDATE SET version = EXCLUDED.version ;")
```
thomash-acinq, line 95: Why did you change the SQL here? It was `s"UPDATE versions SET version=$newVersion WHERE db_name='$db_name'"`, I find the old version more readable.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -46,38 +46,38 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
 
   inTransaction { pg =>
     using(pg.createStatement()) { statement =>
-      def migration45(statement: Statement): Int = {
-        statement.executeUpdate("CREATE TABLE IF NOT EXISTS relayed_trampoline (payment_hash TEXT NOT NULL, amount_msat BIGINT NOT NULL, next_node_id TEXT NOT NULL, timestamp BIGINT NOT NULL)")
-        statement.executeUpdate("CREATE INDEX IF NOT EXISTS relayed_trampoline_timestamp_idx ON relayed_trampoline(timestamp)")
-        statement.executeUpdate("CREATE INDEX IF NOT EXISTS relayed_trampoline_payment_hash_idx ON relayed_trampoline(payment_hash)")
+      def migration45(statement: Statement): Unit = {
```
thomash-acinq, line 49: Since this is inside the `using` clause, we don't need to take `statement` as a parameter.

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
pm47: This replaces two statements:
```scala
    statement.executeUpdate(s"INSERT INTO versions VALUES ('$db_name', $currentVersion) ON CONFLICT DO NOTHING")
...
    statement.executeUpdate(s"UPDATE versions SET version=$newVersion WHERE db_name='$db_name'")
```

This is the recommended pattern according to postgres doc
(https://www.postgresql.org/docs/current/plpgsql-control-structures.html#PLPGSQL-UPSERT-EXAMPLE):

> It is recommended that applications use INSERT with ON CONFLICT DO
UPDATE rather than actually using this pattern.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
pm47: True, but I don't like having methods modify objects that are outside of their scope. Tbh I started by moving those methods outside of the enclosing `inTransaction` and put them back inside at the last minute.

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
thomash-acinq: Understood, previously `setVersion` assumed that the version was already there (but there still was a useless `CREATE TABLE IF NOT EXISTS`) and needed updating but now it may not be there in the first place and we need `INSERT`.

thomash-acinq (dismissed):

pm47:
Rebased.

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
```diff
@@ -60,6 +60,42 @@ trait JdbcUtils {
     }
   }
 
+  private def createVersionTable(statement: Statement): Unit = {
+    statement.executeUpdate("CREATE TABLE IF NOT EXISTS versions (db_name TEXT NOT NULL PRIMARY KEY, version INTEGER NOT NULL)")
+  }
+
+  /**
+   * Several logical databases (channels, network, peers) may be stored in the same physical sqlite database.
```
t-bast, line 68: nit:
```suggestion
   * Several logical databases (channels, network, peers) may be stored in the same physical database.
```

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
t-bast: Since this is a shared mechanism for all DB types, I think it's perfectly fine to have it here, especially if it makes testing easier.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgChannelsDb.scala
```diff
@@ -40,27 +40,29 @@ class PgChannelsDb(implicit ds: DataSource, lock: PgLock) extends ChannelsDb wit
...
+
+      def migration23(statement: Statement): Unit = {
+        statement.executeUpdate("ALTER TABLE local_channels ADD COLUMN created_timestamp BIGINT")
+        statement.executeUpdate("ALTER TABLE local_channels ADD COLUMN last_payment_sent_timestamp BIGINT")
+        statement.executeUpdate("ALTER TABLE local_channels ADD COLUMN last_payment_received_timestamp BIGINT")
+        statement.executeUpdate("ALTER TABLE local_channels ADD COLUMN last_connected_timestamp BIGINT")
+        statement.executeUpdate("ALTER TABLE local_channels ADD COLUMN closed_timestamp BIGINT")
+      }
+
+      getVersion(statement, DB_NAME) match {
```
t-bast, line 54: nit: the order of the `case` is a bit hard to read, what about using a consistent order in all files?
What makes the most sense to me is:

- list the supported previous versions (`case Some(1)`, `case Some(2)`, etc)
- then the current version (`case Some(CURRENT_VERSION)`)
- then catch unsupported versions (`case Some(unknownVersion)`)
- then `case None`

pm47: Nice catch! It's an oversight, I actually reordered all of them except that one. Fixed with 8e7e76fbd.

The only difference between your ordering and mine is that I put `None` first. The rationale is that it makes sense to have the full from-scratch init first, because it's what describes the current schema of the database.

t-bast: That makes sense, SGTM.

t-bast (dismissed):

pm47:
Rebased.

t-bast (approved):

PR #1776, opened 2021-04-19
Ignore tests with serversocket

pm47:


t-bast (approved):
Let's unblock current builds with this until we figure out how to fix these tests.

PR #1777, opened 2021-04-20
Reenable ServerSocket tests

t-bast:
There are issues with IPv6 loopback addresses on build machines, so we force IPv4 and use non-blocking socket servers.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
@@ -37,12 +37,13 @@ import org.scalatest.funsuite.FixtureAnyFunSuiteLike
 import org.scalatest.{Outcome, ParallelTestExecution, Tag}
 import scodec.bits.ByteVector
 
-import java.net.{ServerSocket, Socket}
-import java.util.concurrent.Executors
+import java.net.InetSocketAddress
+import java.nio.channels.ServerSocketChannel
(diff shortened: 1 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -379,3 +368,15 @@ class PeerSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with Paralle
...
   }
 }
+
+object PeerSpec {
+
+  def createMockServer(): (ServerSocketChannel, InetSocketAddress) = {
+    val mockServer = ServerSocketChannel.open()
+    // NB: we force 127.0.0.1 (IPv4) because there are issues on ubuntu build machines with IPv6 loopback
+    mockServer.bind(new InetSocketAddress("127.0.0.1", 0))
+    mockServer.configureBlocking(false)
```
pm47, line 378: `configureBlocking(false)` ❤️

pm47 (approved):

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
t-bast: I discovered that one while I was thinking that the issue was a deadlock on the server. It didn't fix the build, but it's still an improvement :smiley:

PR #1778, opened 2021-04-20
Use proper data type for timestamps in Postgres

pm47:
Did some refactoring in tests and introduced a new `migrationCheck`
helper method.

Note that the change of data type in sqlite for the `commitment_number`
field (from `BLOB` to `INTEGER`) is not a migration. If the table has
been created before, it will stay like it was. It doesn't matter due to
how sqlite stores data, and we make sure in tests that there is no
regression.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
@@ -29,7 +29,8 @@ import fr.acinq.eclair.transactions.Transactions.PlaceHolderPubKey
 import fr.acinq.eclair.{MilliSatoshi, MilliSatoshiLong}
 import grizzled.slf4j.Logging
 
-import java.sql.Statement
+import java.sql.{Statement, Timestamp}
+import java.time.Instant
 import java.util.UUID
 import javax.sql.DataSource
 import scala.collection.immutable.Queue
@@ -40,7 +41,7 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
   import ExtendedResultSet._
 
   val DB_NAME = "audit"
-  val CURRENT_VERSION = 5
+  val CURRENT_VERSION = 6
 
   case class RelayedPart(channelId: ByteVector32, amount: MilliSatoshi, direction: String, relayType: String, timestamp: Long)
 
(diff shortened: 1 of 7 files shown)


pm47:
I didn't add a specific v5->v6 migration test because existing tests already make sure that there are no issues with the `Long`/`Timestamp` conversion. The risk is essentially around the queries, which are tested in the `add/list events` test.  I feel like a new test would add a lot of verbosity for little to no value. Let me know if you disagree.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -52,15 +53,25 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
         statement.executeUpdate("CREATE INDEX relayed_trampoline_payment_hash_idx ON relayed_trampoline(payment_hash)")
       }
 
+      def migration56(statement: Statement): Unit = {
+        statement.executeUpdate("ALTER TABLE sent               ALTER COLUMN timestamp SET DATA TYPE TIMESTAMP WITH TIME ZONE USING timestamp with time zone 'epoch' + timestamp * interval '1 millisecond'")
```
pm47, line 57: This was provided as an example in the [postgres documentation(https://www.postgresql.org/docs/9.6/sql-altertable.html).

thomash-acinq:
> I didn't add a specific v5->v6 migration test because existing tests already make sure that there are no issues with the `Long`/`Timestamp` conversion. The risk is essentially around the queries, which are tested in the `add/list events` test. I feel like a new test would add a lot of verbosity for little to no value. Let me know if you disagree.

I agree that the migration is already covered by the other tests. However I find it awkward that the sqlite and postgres databases are now using different versions, I would bump the version for sqlite too just to stay consistent.

pm47:
> I find it awkward that the sqlite and postgres databases are now using different versions, I would bump the version for sqlite too just to stay consistent.

The reason is that sqlite has _not_ been migrated, it still stores timestamps as `INTEGER`. The rationale for not migrating sqlite is that it is much more difficult to do than on postgres (there is no `ALTER DATA TYPE`) and adds unnecessary risk on our production db. In the future we may significantly change the database format only in Postgres for optimization purposese (e.g. split serialized blobs into several columns), and I think it would make sense to have the version reflect the difference in the schema. Another argument: we could use a totally different backend (e.g. NOSQL) and the versioning would not be the same at all.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -189,17 +204,17 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
...
       }
     }
   }
 
   override def listSent(from: Long, to: Long): Seq[PaymentSent] =
     inTransaction { pg =>
-      using(pg.prepareStatement("SELECT * FROM sent WHERE timestamp >= ? AND timestamp < ?")) { statement =>
-        statement.setLong(1, from)
-        statement.setLong(2, to)
+      using(pg.prepareStatement("SELECT * FROM sent WHERE timestamp BETWEEN ? AND ? ORDER BY timestamp")) { statement =>
```
pm47, line 215: NB: added an `ORDER BY TIMESTAMP` here.

eclair-core/src/test/scala/fr/acinq/eclair/db/AuditDbSpec.scala
```diff
@@ -182,13 +183,20 @@ class AuditDbSpec extends AnyFunSuite {
     }
   }
 
-  test("handle migration version 1 -> 5") {
-    forAllDbs {
-      case _: TestPgDatabases => // no migration
-      case dbs: TestSqliteDatabases =>
-        import fr.acinq.eclair.db.sqlite.SqliteUtils.getVersion
-        val connection = dbs.connection
+  test("migrate sqlite audit database v1 -> v5/v6") {
```
t-bast, line 186: nit: why `v5/v6`, it's migrating to v5 since there's no v6?

eclair-core/src/test/scala/fr/acinq/eclair/db/ChannelsDbSpec.scala
```diff
@@ -175,85 +163,217 @@ class ChannelsDbSpec extends AnyFunSuite {
...
-          statement.executeUpdate("CREATE TABLE IF NOT EXISTS htlc_infos (channel_id BLOB NOT NULL, commitment_number BLOB NOT NULL, payment_hash BLOB NOT NULL, cltv_expiry INTEGER NOT NULL, FOREIGN KEY(channel_id) REFERENCES local_channels(channel_id))")
-          statement.executeUpdate("CREATE INDEX IF NOT EXISTS htlc_infos_idx ON htlc_infos(channel_id, commitment_number)")
-          setVersion(statement, "channels", 2)
-        }
+  }
+}
+
+object ChannelsDbSpec {
+
+  case class TestCase(
```
t-bast, line 330: nit: indent is weird here as well

t-bast (commented):
Ok I can see why having the version numbers diverge between sqlite and postgres make the migration tests quite harder to read...

But I think there's no way around it, we will want to make them evolve independently. I think it's ok to have completely separate migration tests for each DB type in the future. The important tests to share between DB types are the functional tests of the behavior of the latest DB version (e.g. `add/remove/list channels` for the channels DB).

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
pm47: The idea was to make similarities obvious between lines. Removed in 641abae.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
pm47: In fact we can't, because results are aggregated in a `Map`. I removed the `ORDER BY` instead in 6b20b25.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteChannelsDb.scala
```diff
@@ -63,9 +62,9 @@ class SqliteChannelsDb(sqlite: Connection) extends ChannelsDb with Logging {
 
     getVersion(statement, DB_NAME) match {
       case None =>
-        statement.executeUpdate("CREATE TABLE local_channels (channel_id BLOB NOT NULL PRIMARY KEY, data BLOB NOT NULL, is_closed BOOLEAN NOT NULL DEFAULT 0, created_timestamp INTEGER, last_payment_sent_timestamp INTEGER, last_payment_received_timestamp INTEGER, last_connected_timestamp INTEGER, closed_timestamp INTEGER)")
-        statement.executeUpdate("CREATE TABLE htlc_infos (channel_id BLOB NOT NULL, commitment_number BLOB NOT NULL, payment_hash BLOB NOT NULL, cltv_expiry INTEGER NOT NULL, FOREIGN KEY(channel_id) REFERENCES local_channels(channel_id))")
-        statement.executeUpdate("CREATE INDEX htlc_infos_idx ON htlc_infos(channel_id, commitment_number)")
+        statement.executeUpdate("CREATE TABLE IF NOT EXISTS local_channels (channel_id BLOB NOT NULL PRIMARY KEY, data BLOB NOT NULL, is_closed BOOLEAN NOT NULL DEFAULT 0, created_timestamp INTEGER, last_payment_sent_timestamp INTEGER, last_payment_received_timestamp INTEGER, last_connected_timestamp INTEGER, closed_timestamp INTEGER)")
```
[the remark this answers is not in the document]
pm47: Fixed with c97900a.

eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
```diff
@@ -77,4 +78,24 @@ object TestDatabases {
     // @formatter:on
   }
 
+  def migrationCheck(
```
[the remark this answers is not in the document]
pm47: Done with aba59ab.

eclair-core/src/test/scala/fr/acinq/eclair/db/AuditDbSpec.scala
pm47: Done with aba59ab.

eclair-core/src/test/scala/fr/acinq/eclair/db/ChannelsDbSpec.scala
pm47: Done with aba59ab.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
t-bast: It's still there in this file...?

pm47: Oops, forgot one file. Fixed with 1ff9b7b.

t-bast (approved):

pm47:
:love_you_gesture:

PR #1781, opened 2021-04-27
Don't log ClosingType object

t-bast:
It may contain a lot of data and pollute the logs.


pm47 (approved):

pm47 (commented):
NB: `EventType.Closed(closingType).label` would look better

pm47 (approved):

PR #1785, opened 2021-05-04
Remove bintray repository

pm47:
Bintray has been discontinued and artifacts have been migrated to Maven
Central.


t-bast (approved):
All the kamon artifacts seem to be available on Maven Central, so LGTM

PR #1787, opened 2021-05-06
Proper support for keysend tlv

pm47:
We should use the typed `KeySend` tlv instead of a custom one.
Also this makes `SendPaymentRequest` more consistent with
`SendMultiPartPayment`.


t-bast (commented):
This change is incomplete, since this is also used at the JSON API level.
If we really want to get rid of using a generic tlv for keysend, I don't think this is how we should do it, adding the possibility to insert any `OnionTlv` by senders is very footgun-ish.
We should instead explicitly provide a "keysend" hook.

pm47:
Agreed, this is ill-designed, and the status of keysend is unclear anyway.

PR #1788, opened 2021-05-07
Use secp256k1-kmp

sstone:
We get rid of our old JNI wrapper and switch to bitcoin-lib 0.18 which is based on secp256k1-kmp.
This will give us a consistent secp256k1 library on all our apps (since secp256k1-kmp also provides an Android library).
We also switch from spongycastle bouncycastle (bitcoin-lib 0.18 uses bouncycastle).


dpad85 (approved):
LGTM 👍

dpad85:
Tested successfully on device.

PR #1791, opened 2021-05-10
Fix ConnectionControlPlugin

btcontract:
Adding `forceReconnect` to `ReconnectionTask` was not enough to properly enable connection control by a plugin, without these changes a connection without normal channels will still be removed from db and node won't be able to automatically reconnect on restart.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -79,7 +79,7 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: EclairWa
       val h = d.channels.filter(_._2 == actor).keys
       log.info(s"channel closed: channelId=${h.mkString("/")}")
       val channels1 = d.channels -- h
-      if (channels1.isEmpty) {
+      if (!nodeParams.forceReconnect(remoteNodeId) && channels1.isEmpty) {
         // we have no existing channels, we can forget about this peer
         stopPeer()
       } else {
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -185,7 +185,7 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: EclairWa
         Logs.withMdc(diagLog)(Logs.mdc(category_opt = Some(Logs.LogCategory.CONNECTION))) {
           log.info("connection lost")
         }
-        if (d.channels.isEmpty) {
+        if (!nodeParams.forceReconnect(remoteNodeId) && d.channels.isEmpty) {
```
t-bast, line 188: I think you also need to update this condition: https://github.com/ACINQ/eclair/blob/a8d4e07bdde9b5f28ebad291ac3dbef9cc97c759/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala#L200 to keep the connection open even if there are no channels left?

btcontract: To think about it, `Peer` logic seems rather loaded as is, OTOH it occured to me that any plugin can easily request reconnection by itself on getting `PeerDisconnected` (for example by obtaining `Router` data and extracting address from there, or by maintaining its own db).

Maybe it makes sense to get rid of a whole `ConnectionControlPlugin` logic?

t-bast: If we can get rid of it without impacting plugins too much, then hell yes, less complexity to deal with inside the peer ;)

btcontract: For now the only plugin (that I'm aware of) using this is mine, and even that one is not released yet, so it's a good time to roll this feature back. I'll proceed with PR if that's OK.

t-bast: Perfect, that sounds good to me!

Issue #1789, opened 2021-05-08
Eclair can't keep up with Bitcoind


engenegr:
Eclair was constantly behind the tip in one of my setups.

I have 2 bitcoind full nodes. I can switch between them and Eclair runs VPS. Full node and Eclair node are in VPN network. One of these nodes is slow but it seems it can provide blocks via ZMQ reliable. Other is not and I currently have no idea why. I tried to use keep alive option for tcp sockets but it seems it didn't help or I did something wrong.

The point is, why couldn't Eclair actively request bitcoind node for block? Because I see such behaviour that ZMQ could probably silently dropped, Eclair finds out it is late on the tip and peers disconnect, then Eclair doesn't do anything, just waiting for blocks or restart to reconnect to ZMQ again.

engenegr:
While I was writing previous message Eclair was on the same height. Then it jumped from 682717 to 682722 in getinfo.

I attempted to measure python-zmq simple code vs Eclair.

Python recieved 682725 and 682726 headers with 20 sec delay (to block timestamp as it showed by Core) for 682726.
While I'm writing this my Eclair is still on 682724.

engenegr:
An output of Python zmq script with timestamps when it receives header.

```
- RAW BLOCK HEADER (39) - 2021-05-09 09:41:37.844825
b'0000c020214f40f0d9e1b51b9eafa74b942f676d8688e221bc820d000000000000000000124edcb793b847eeee8a1b70aeb264b675ff95ab3dc721a75b4f092f1b7b5cd739ae976063a80d17040675d7'
- RAW BLOCK HEADER (40) - 2021-05-09 09:41:43.497660
b'00e0ff27db405edafde0264e08075407d02ab1c77a9b44adc1e20900000000000000000033f35a99befa81b5c435961f8896fb21834a02f992adfcf1639e86216951ad5241ae976063a80d17014a0ed9'
```

An output of `getinfo` after that

```
{
  "version": "0.5.2-SNAPSHOT-90fbcd3",
  "nodeId": "02cd1b7bc418fac2dc99f0ba350d60fa6c45fde5ab6017ee14df6425df485fb1dd",
  "alias": "SATM",
  "color": "#b00b69",
  "features": {
    "activated": {
      "gossip_queries": "optional",
      "payment_secret": "optional",
      "option_data_loss_protect": "optional",
      "option_static_remotekey": "optional",
      "basic_mpp": "optional",
      "gossip_queries_ex": "optional",
[...]

PR #1793, opened 2021-05-11
ZMQ actors should subscribe to a single topic

t-bast:
We use one actor per topic, but each actor previously registered to multiple topics so we received duplicate events and consumed twice the necessary bandwidth.

Thanks @btcontract for noticing this in #1789


t-bast:
I agree that it's a bit strange to implement all topics whereas we will subscribe to a single one, but:

- There should really be two distinct actors to ensure parallelism and isolation (we really want two distinct tcp sockets)
- It's better and simpler to have all the code in the same place (instead of separating the logic in two distinct actor types)

If we opened a single socket I would totally agree with you, but given the separate sockets setup I think the existing code makes sense.

PR #1794, opened 2021-05-11
Fix computation of path weight

thomash-acinq:
There are two cases depending on whether you use weight ratios or not. They used to behave very differently.
- Without weight ratios, the weight would be the total amount to send (base amount + fees)
- With weight ratios, the ratios would apply to the total amount, not just the fees. The effect is that only the number of hops and then the weight factors matter as the fee itself is negligible compared to the total amount.

The code is now shared and the weight is now the sum of the fees (multiplied by a factor if using weight ratios).


t-bast (approved):
LGTM, good catch!

PR #1796, opened 2021-05-12
Update default path-finding weight ratios

t-bast:
* The two months window for channel age was too small for today's network (where most channels are at least 6 months old)
* CLTV is less of an issue nowadays: there are fewer stuck payments and we're encouraging nodes to increase their CLTV because of the recent mempool congestions, so when your payment is stuck it's very painful regardless of the path you took

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -274,9 +274,9 @@ object Graph {
       case Some(wr) =>
         import RoutingHeuristics._
 
-        // Every edge is weighted by funding block height where older blocks add less weight, the window considered is 2 months.
+        // Every edge is weighted by funding block height where older blocks add less weight. The window considered is 1 year.
         val channelBlockHeight = ShortChannelId.coordinates(edge.desc.shortChannelId).blockHeight
-        val ageFactor = normalize(channelBlockHeight, min = currentBlockHeight - BLOCK_TIME_TWO_MONTHS, max = currentBlockHeight)
+        val ageFactor = normalize(channelBlockHeight, min = currentBlockHeight - BLOCK_TIME_ONE_YEAR, max = currentBlockHeight)
 
         // Every edge is weighted by channel capacity, larger channels add less weight
         val edgeMaxCapacity = edge.capacity.toMilliSatoshi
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -336,8 +336,8 @@ object Graph {
 
   object RoutingHeuristics {
 
-    // Number of blocks in two months
-    val BLOCK_TIME_TWO_MONTHS = 8640
+    // Number of blocks in one year
+    val BLOCK_TIME_ONE_YEAR = 52560
```
thomash-acinq, line 340: I would write the formula, not just the result, it's easier to check.

```suggestion
    val BLOCK_TIME_ONE_YEAR = 365 * 24 * 6
```

thomash-acinq (commented):
The fact that most channels are at least 6 months is not really relevant here. What matters is: Is a 6-month-old channel more reliable than a 2-month-old one? And I'm not convinced that's the case.

thomash-acinq (approved):

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: Good idea, done in https://github.com/ACINQ/eclair/pull/1796/commits/5e668b22240a40118a66f38042763e2f622b20df

t-bast:
> What matters is: Is a 6-month-old channel more reliable than a 2-month-old one?

I believe that yes, the older a channel is, the more reliable it is, for the following reasons:

- If an HTLC was ever stuck in that channel, the channel would have been force-closed. So the older a channel is, the less buggy it probably is
- It shows that the channel is frequently used, otherwise its operator would have closed it to allocate its funds elsewhere. So it's frequently used and hasn't had a fatal bug, which is a good indication that it's reliable

The drawback is the barrier to entry for new channels / new nodes, maybe we should lower the ratio on channel age to offer a better opportunity for newcomers? But from the point of view of the sender, old is good if they don't care about the network as a market (and new channels can offer more competitive fees to make up for their young age).

Of course there's always the risk that a channel was very reliable for a long time, and then upgraded its software to a buggy version, but I don't think we should (or even can) account for that...

thomash-acinq (approved):

PR #1798, opened 2021-05-12
Add tlv stream to UpdateAddHtlc message

btcontract:
Storing some data in `UpdateAddHtlc` will in particular help an HC peer to correctly resolve its own in-flight payments when it resyncs with incomplete local data. What's needed from Eclair is to be aware that `UpdateAddHtlc` can have some tlv data so HC plugin can store a complete message.

Currently this fails some backwards-compat tests in `wire`, not sure why, some help would be appreciated.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/UpdateAddHtlcTlv.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/UpdateAddHtlcTlv.scala
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2019 ACINQ SAS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fr.acinq.eclair.wire.protocol
+
+import fr.acinq.eclair.UInt64
+import fr.acinq.eclair.wire.protocol.CommonCodecs._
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/UpdateAddHtlcTlv.scala
```diff
@@ -0,0 +1,35 @@
...
+package fr.acinq.eclair.wire.protocol
+
+import fr.acinq.eclair.UInt64
+import fr.acinq.eclair.wire.protocol.CommonCodecs._
+import fr.acinq.eclair.wire.protocol.TlvCodecs.tlvStream
+import scodec.Codec
+import scodec.bits.ByteVector
+import scodec.codecs._
+
+object UpdateAddHtlcTlv {
```
t-bast, line 26: nit: it's likely that some TLV will be shared between many `*Htlc` messages, so I would create only one file for all of them: `HtlcTlv`

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/UpdateAddHtlcTlv.scala
```diff
@@ -0,0 +1,35 @@
...
+  case class Data(data: ByteVector) extends Tlv
+
+  val codec: Codec[TlvStream[Data]] = tlvStream(
+    discriminated[Data].by(varint)
+      .typecase(UInt64(1L), variableSizeBytesLong(varintoverflow, bytes).as[Data])
+  )
```
t-bast, line 32: We can't reserve in `eclair` tlv type `1` which isn't spec-ed.
You should create a sealed trait without any concrete implementations for now.
See what's done in `ChannelTlv` for example.

btcontract:
Can we have some progress on this?

t-bast:
> Can we have some progress on this?

WDYM? This PR is still draft and the build isn't passing, do you need help figuring out what's wrong?

btcontract:
Yes, I still don't understand why specifically backwards-compatibility tests fail while otherwise it works, this is why it's still a draft.

t-bast:
I'll try to find some time to look at this today.

t-bast:
Closed in favor of #1891

PR #1799, opened 2021-05-12
Do not retry sending if payment gets confirmed on chain

btcontract:
This fixes a case where `MultiPartPaymentLifecycle` keeps retrying a payment even after it got confirmed on chain. Doing so is harmless but meaningless.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
@@ -16,10 +16,14 @@
 
 package fr.acinq.eclair.payment.send
 
+import java.util.UUID
+import java.util.concurrent.TimeUnit
+
 import akka.actor.{ActorRef, FSM, Props, Status}
 import akka.event.Logging.MDC
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.eclair.channel.{HtlcOverriddenByLocalCommit, HtlcsTimedoutDownstream, HtlcsWillTimeoutUpstream}
 import fr.acinq.eclair.db.{OutgoingPayment, OutgoingPaymentStatus, PaymentType}
 import fr.acinq.eclair.payment.Monitoring.{Metrics, Tags}
 import fr.acinq.eclair.payment.OutgoingPacket.Upstream
@@ -32,10 +36,6 @@ import fr.acinq.eclair.router.RouteCalculation
 import fr.acinq.eclair.router.Router._
 import fr.acinq.eclair.wire.protocol._
 import fr.acinq.eclair.{CltvExpiry, FSMDiagnosticActorLogging, Logs, MilliSatoshi, MilliSatoshiLong, NodeParams}
-import java.util.UUID
-import java.util.concurrent.TimeUnit
-
-import fr.acinq.eclair.channel.{HtlcOverriddenByLocalCommit, HtlcsTimedoutDownstream}
 
 /**
  * Created by t-bast on 18/07/2019.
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -377,8 +378,9 @@ object MultiPartPaymentLifecycle {
     SendPaymentToRoute(replyTo, Right(route), finalPayload)
   }
 
-  /** When we receive an error from the final recipient, we should fail the whole payment, it's useless to retry. */
-  private def isFinalRecipientFailure(pf: PaymentFailed, d: PaymentProgress): Boolean = pf.failures.collectFirst {
+  /** When we receive an error from the final recipient or payment gets confirmed on chain, we should fail the whole payment, it's useless to retry. */
+  private def doNotRetry(pf: PaymentFailed, d: PaymentProgress): Boolean = pf.failures.collectFirst {
+    case LocalFailure(_, _: HtlcsTimedoutDownstream | _: HtlcOverriddenByLocalCommit) => true
```
t-bast, line 383: This is `HtlcsWillTimeoutUpstream` as well.
You should probably create a dedicated method `isOnChainFail(f: PaymentFailure)`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -377,8 +378,9 @@ object MultiPartPaymentLifecycle {
     SendPaymentToRoute(replyTo, Right(route), finalPayload)
   }
 
-  /** When we receive an error from the final recipient, we should fail the whole payment, it's useless to retry. */
-  private def isFinalRecipientFailure(pf: PaymentFailed, d: PaymentProgress): Boolean = pf.failures.collectFirst {
+  /** When we receive an error from the final recipient or payment gets confirmed on chain, we should fail the whole payment, it's useless to retry. */
+  private def doNotRetry(pf: PaymentFailed, d: PaymentProgress): Boolean = pf.failures.collectFirst {
```
t-bast, line 382:
```suggestion
  /** When we receive an error from the final recipient or payment gets settled on chain, we should fail the whole payment, it's useless to retry. */
  private def abortPayment(pf: PaymentFailed, d: PaymentProgress): Boolean = pf.failures.collectFirst {
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -32,10 +32,11 @@ import fr.acinq.eclair.router.RouteCalculation
 import fr.acinq.eclair.router.Router._
 import fr.acinq.eclair.wire.protocol._
 import fr.acinq.eclair.{CltvExpiry, FSMDiagnosticActorLogging, Logs, MilliSatoshi, MilliSatoshiLong, NodeParams}
-
 import java.util.UUID
 import java.util.concurrent.TimeUnit
 
+import fr.acinq.eclair.channel.{HtlcOverriddenByLocalCommit, HtlcsTimedoutDownstream}
```
t-bast, line 38: nit: these aren't correctly placed, you should apply a `sort imports` refactoring

eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycleSpec.scala
```diff
@@ -384,6 +384,20 @@ class MultiPartPaymentLifecycleSpec extends TestKitBaseClass with FixtureAnyFunS
     assert(result.failures.length === 1)
   }
 
+  test("abort if payment gets confirmed on chain") { f =>
```
t-bast, line 387: nit: `confirmed` implies success in my mind, whereas most of the time it will be an htlc-timeout that will happen on-chain when we're sending a payment, so `settled` feels better.

```suggestion
  test("abort if payment gets settled on chain") { f =>
```

t-bast (commented):
Good catch!

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
btcontract: Seems I'd have to do exactly the same matching in a new method, what about keeping as is but expanding each case: https://github.com/ACINQ/eclair/pull/1799/commits/776d9dfc92906ca665c37fd12954fb58290b487c#diff-f28a1f15ee9515a9b2bb404c65cb4a01415b12d264f509420bdfb1251fd49afbR382-R388 ?

t-bast (approved):

PR #1800, opened 2021-05-13
Make new API structures reusable by plugins

btcontract:
New `AbstractService` can be used by a plugin to only define its custom routes and reuse everything else.

### DIFF
--- a/eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
+++ b/eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
@@ -69,14 +69,12 @@ object Boot extends App with Logging {
     val config = system.settings.config.getConfig("eclair")
     if (config.getBoolean("api.enabled")) {
       logger.info(s"json API enabled on port=${config.getInt("api.port")}")
-      implicit val materializer = ActorMaterializer()
       val apiPassword = config.getString("api.password") match {
         case "" => throw EmptyAPIPasswordException
         case valid => valid
       }
       val apiRoute = new Service {
         override val actorSystem = system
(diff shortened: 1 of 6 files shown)


t-bast:
I don't really like extending this by inheritance, I think we should use a more explicit system to allow multiple independent plugins to safely add handlers.

What about adding explicit methods that let plugins provide their route handlers, and have some explicit logic in `Boot.scala` to get these handlers for all plugins and inject them in the `Service`?

btcontract:
Currently this works by plugin establishing its own API endpoint on a different port which is independent from base API. `AbstractService` is only needed for code reuse, nothing is injected into base API.

I was thinking about actually injecting custom routes but:
- there is no `akka-http-dsl` module in `eclair-core` which would be needed for hypothetical `ApiPlugin` trait to declare `Route` type and I guess that module does not belong to core.
- there is a chance that base API and plugin API endpoints will start clashing, especially if many API-enabled plugins are to be used at some point.

So I've decided that letting plugins defining their own endpoints is a safer way.

t-bast:
But the `plugin` trait is in `eclair-node`, not `eclair-core`, so you can easily add methods to it without touching `eclair-core`?
This wouldn't be a `PluginParams` but rather dedicated methods on the `Plugin` trait?

On the contrary I think it makes a lot more sense to have plugin routes integrate in the same endpoint, we don't want to end up with a completely incoherent ecosystem that will be a nightmare for node operators. Plugin developers should IMO ensure their routes don't clash with core routes, and the core code should gracefully handle the case where two plugins try to register the same route (and refuse to start in that case), don't you think?

PR #1806, opened 2021-05-16
Use satoshi for htlc ordering

### LINKED ISSUE #1804: Satoshi amounts must be used for htlc comparison
Spec says that the equality is based on outputs, which are in sats:

> Lexicographic ordering: see BIP69. In the case of identical HTLC outputs, the outputs are ordered in increasing cltv_expiry order.

We're using the millisatoshi amount:

https://github.com/ACINQ/eclair/blob/898c17bc76c1af9077314958612ad96c019bd381/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala#L309-L314

This causes commitment sig mismatch and force closes.

Thanks @roasbeef for the debug!

pm47:
Reproduce and fix the spec violation. Updated the test vectors according to https://github.com/lightningnetwork/lightning-rfc/pull/872#issuecomment-841818906.

Fixes #1804.


t-bast (approved):
Good catch!

PR #1808, opened 2021-05-17
Check blockchain watchdogs regularly

### LINKED ISSUE #1803: BlockchainWatchdog doesn't work independently
I open this issue because of suggestion made by Anton.

The issue revealed as a result of other problems https://github.com/ACINQ/eclair/issues/1789.

In particular, I noticed that when ZMQ actor stops receiving blocks, BlockchainWatchdog also stops working and counting how many blocks node missed with regard to explorers.

t-bast:
We want to check secondary blockchain sources when we haven't received blocks in a while.
Thanks @engenegr and @btcontract for raising this.

Fixes #1803

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/BlockchainWatchdog.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/BlockchainWatchdog.scala
@@ -34,9 +34,10 @@ import scala.util.Random
 /** Monitor secondary blockchain sources to detect when we're being eclipsed. */
 object BlockchainWatchdog {
 
-  case class BlockHeaderAt(blockCount: Long, blockHeader: BlockHeader)
-
   // @formatter:off
+  case class BlockHeaderAt(blockCount: Long, blockHeader: BlockHeader)
+  case object NoBlockReceivedTimer
+
   sealed trait BlockchainWatchdogEvent
   /**
    * We are missing too many blocks compared to one of our blockchain watchdogs.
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/BlockchainWatchdog.scala
```diff
@@ -50,21 +51,29 @@ object BlockchainWatchdog {
...
+  def apply(chainHash: ByteVector32, maxRandomDelay: FiniteDuration, blockTimeout: FiniteDuration = 15 minutes): Behavior[Command] = {
     Behaviors.setup { context =>
       context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[CurrentBlockCount](cbc => WrappedCurrentBlockCount(cbc.blockCount)))
       Behaviors.withTimers { timers =>
+        // We start a timer to check blockchain watchdogs regularly even when we don't receive any block.
+        timers.startSingleTimer(NoBlockReceivedTimer, NoBlockReceivedSince(0), blockTimeout)
         Behaviors.receiveMessage {
+          case NoBlockReceivedSince(lastBlockCount) =>
+            context.self ! CheckLatestHeaders(lastBlockCount)
+            timers.startSingleTimer(NoBlockReceivedTimer, NoBlockReceivedSince(lastBlockCount), blockTimeout)
```
pm47, line 71: Doc:

> if a new timer with same key is started the previous is cancelled.

So `timers.startSingleTimer` effectively resets the timer. ✔️

pm47 (approved):

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/BlockchainWatchdog.scala
t-bast: Yes, that's why I chose the overload with a `key` ;)

PR #1810, opened 2021-05-17
Make payment_secret mandatory

t-bast:
This is a security feature that has been introduced a long time ago and is widely supported across the network.
We can safely make it mandatory which closes probing attack vectors.

A separate follow-up PR will remove some legacy code that handled cases where `payment_secret` could be missing.
I'd rather first release the activation bits and then remove that legacy code instead of bundling the two together.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -44,9 +44,9 @@ eclair {
     option_data_loss_protect = optional
     gossip_queries = optional
     gossip_queries_ex = optional
-    var_onion_optin = optional
+    var_onion_optin = mandatory
     option_static_remotekey = optional
-    payment_secret = optional
+    payment_secret = mandatory
     basic_mpp = optional
     option_support_large_channel = optional
     option_anchor_outputs = disabled
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -241,7 +241,8 @@ object NodeParams extends Logging {
     def validateFeatures(features: Features): Unit = {
       val featuresErr = Features.validateFeatureGraph(features)
       require(featuresErr.isEmpty, featuresErr.map(_.message))
-      require(features.hasFeature(Features.VariableLengthOnion), s"${Features.VariableLengthOnion.rfcName} must be enabled")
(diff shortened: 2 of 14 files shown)


pm47:
cc @dpad85 may have some impact on old Eclair Mobile IIUC

pm47 (commented):
Should we make our onion type stricter and simplify the relayers at the same time? https://github.com/ACINQ/eclair/pull/1770#discussion_r625853126. Or do you think we should wait because there may be pre-existing payment requests that have not yet expired?

t-bast:
> Should we make our onion type stricter and simplify the relayers at the same time? #1770 (comment). Or do you think we should wait because there may be pre-existing payment requests that have not yet expired?

I'd rather do it in a separate PR, in the release *after* we make `payment_secret` mandatory.
I started removing all legacy code related to `var_onion_optin` and `payment_secret`, and I think it's quite a lot of subtle changes so I prefer doing it after the release, for the next one.

pm47 (approved):

t-bast:
> cc @dpad85 may have some impact on old Eclair Mobile IIUC

Support for `payment_secret` was added to eclair mobile in the [v0.4.7 release](https://github.com/ACINQ/eclair-mobile/releases/tag/v0.4.7-MAINNET) in July 2019. Only wallets older than that will not be able to connect to our node anymore and will need to update. A lot has happened since then, so it's really recommended that they upgrade anyway!

joostjager:
@t-bast could it be that `payment_secret` is mandatory on the ACINQ node, but that its feature bits are still communicated as optional?

I am running into `INVALID_ONION_PAYLOAD` when probing the ACINQ node without a payment secret.

t-bast:
Yes that's exactly it.

We set the feature bit as mandatory in invoices (because that's where this feature bit makes sense) but not in our `init` / `node_announcement`. Since probing is done without an invoice, that's why you're not getting this information.

Depending on what library you're probing with, this may already be fixed if you upgrade (see https://github.com/ACINQ/eclair/issues/1913).

joostjager:
Why don't you set the feature bit as mandatory in the `init` / `node_announcement` too?

t-bast:
Because there's no reason to yet (as a relaying-only node you don't care at all about `payment_secret`, you don't even need to set it) and some of our peers don't have it in their `init` which would break our connection to them for no good reason.

PR #1814, opened 2021-05-19
Stop masking feature bits

t-bast:
Remove a legacy work-around for very old eclair-mobile users.

NB: this is a PR to the `android` branch. It should then be merged in the `android-phoenix` branch as well.


sstone (approved):

dpad85 (approved):
Note, for eclair mobile we will use the following features when connecting to not-acinq nodes:

https://github.com/ACINQ/eclair-mobile/blob/mainnet/app/src/main/resources/application.conf#L6-L12

PR #1815, opened 2021-05-19
Add cost per hop and bias to weight ratios

thomash-acinq:
- Add a bias weight ratio
  Allows to give more weight to the fee itself. Setting the bias to 1 is like not using weight ratios.
- Add cost per hop
  Allows prioritizing short paths. The cost per hop has a base component and a proportional one, similar to channel fees.


t-bast:
Sorry I didn't look at this one earlier, can you rebase it?
I'll review this in details soon, it looks like a worthy change (more flexibility in tuning the path-finding algorithm to each node's specific needs).

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -45,8 +45,8 @@ object Graph {
    * We use heuristics to calculate the weight of an edge based on channel age, cltv delta and capacity.
    * We favor older channels, with bigger capacity and small cltv delta.
    */
-  case class WeightRatios(cltvDeltaFactor: Double, ageFactor: Double, capacityFactor: Double) {
-    require(0 < cltvDeltaFactor + ageFactor + capacityFactor && cltvDeltaFactor + ageFactor + capacityFactor <= 1, "The sum of heuristics ratios must be between 0 and 1 (included)")
+  case class WeightRatios(biasFactor: Double, cltvDeltaFactor: Double, ageFactor: Double, capacityFactor: Double, hopCostBase: MilliSatoshi, hopCostMillionths: Long) {
```
t-bast, line 48: You can also update the comment on this class to explain that we favor shorter routes when the `biasFactor` (which needs to be renamed to mention what it biases on) is greater than 0 and `hopCostBase` and `hopCostMillionths` are as well.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -309,10 +309,12 @@ object Router {
                         searchMaxFeePct: Double,
                         searchMaxRouteLength: Int,
                         searchMaxCltv: CltvExpiryDelta,
-                        searchHeuristicsEnabled: Boolean,
+                        searchRatioBias: Double,
                         searchRatioCltv: Double,
                         searchRatioChannelAge: Double,
                         searchRatioChannelCapacity: Double,
+                        searchHopCostBase: MilliSatoshi,
+                        searchHopCostMillionths: Long,
```
t-bast, line 317: Should we validate that these two are positive values?

Interestingly, if we use negative values here we favor longer routes, which some privacy-aware users may want, so not sure whether it would be a bug or a feature...if we don't constrain it to positive values we probably need tests to verify the behavior with negative values.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -269,31 +269,27 @@ object Graph {
...
+    // Every edge is weighted by its cltv-delta value, normalized
+    val cltvFactor = normalize(edge.update.cltvExpiryDelta.toInt, CLTV_LOW, CLTV_HIGH)
 
-        // NB we're guaranteed to have weightRatios and factors > 0
-        (cltvFactor * wr.cltvDeltaFactor) + (ageFactor * wr.ageFactor) + (capFactor * wr.capacityFactor)
-    }
-    val totalWeight = prev.weight + fee.toLong * factor
+    // NB we're guaranteed to have weightRatios and factors > 0
+    val factor = weightRatios.biasFactor + (cltvFactor * weightRatios.cltvDeltaFactor) + (ageFactor * weightRatios.ageFactor) + (capFactor * weightRatios.capacityFactor)
+    val totalWeight = prev.weight + (fee + hopCost).toLong * factor
```
t-bast, line 292: Intuitively, I would have chosen a slightly different formula, where I would have kept that new hop factor separate from the other factors:
```
val totalWeight = prev.weight + hopCost * weightRatios.biasFactor + fee * ((cltvFactor * weightRatios.cltvDeltaFactor) + (ageFactor * weightRatios.ageFactor) + (capFactor * weightRatios.capacityFactor))
```

What do you think are the pros/cons of each formula? Is my suggestion dumb?

t-bast (commented):
Thanks, I think it's a really useful heuristic to add, biasing towards shorter routes is often a good choice. Sorry I let this PR slip!

eclair-core/src/main/resources/reference.conf
```diff
@@ -199,13 +199,18 @@ eclair {
...
+      // the following parameters can be used to ask the router to use heuristics to find i.e: 'cltv-optimized' routes, **the sum of the four ratios must be 1**
+      ratio-bias = 0.0
       ratio-cltv = 0.05             // when computing the weight for a channel, consider its CLTV delta in this proportion
       ratio-channel-age = 0.4       // when computing the weight for a channel, consider its AGE in this proportion
       ratio-channel-capacity = 0.55 // when computing the weight for a channel, consider its CAPACITY in this proportion
 
+      // Virtual fee for additional hops
+      // Corresponds to how much you are willing to pay to get one less hop in the payment path
+      hop-cost-base-msat = 500
+      hop-cost-millionths = 200
```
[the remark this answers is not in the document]
thomash-acinq: I've just made them up from what I thought was reasonable. It just corresponds to how much we're willing to pay to make the route one hop shorter. It should be completely independent from the state of the network.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
thomash-acinq: You've misunderstood what the `biasFactor` does, it is not related to the `hopCost`. It biases towards just using the fee as the weight.

eclair-core/src/main/resources/reference.conf
```diff
@@ -199,13 +199,18 @@ eclair {
       fee-threshold-sat = 21       // if fee is below this value we skip the max-fee-pct check
       max-fee-pct = 0.03           // route will be discarded if fee is above this value (in percentage relative to the total payment amount); doesn't apply if fee < fee-threshold-sat
 
-      // channel 'weight' is computed with the following formula: channelFee * (cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity)
-      // the following parameters can be used to ask the router to use heuristics to find i.e: 'cltv-optimized' routes, **the sum of the three ratios must be > 0 and <= 1**
-      heuristics-enable = true      // if true uses heuristics for path-finding
+      // channel 'weight' is computed with the following formula: (channelFee + hop-cost) * (ratio-bias + cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity)
+      // the following parameters can be used to ask the router to use heuristics to find i.e: 'cltv-optimized' routes, **the sum of the four ratios must be 1**
+      ratio-bias = 0.0
```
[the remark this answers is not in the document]
thomash-acinq: `ratio-bias` is not about prioritizing smaller routes, that's what the hop cost is for. It doesn't bias any property, exactly the opposite, it biases toward not using any property.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: Got it, the formula makes much more sense now! That explains why mine was different, since I was trying to apply a ratio to the `hopCost`.

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
t-bast: Ok, they kinda match the default fees we set, which makes sense.
We need to properly document this to help node operators tweak it to their liking.

t-bast (commented):
LGTM!

I think we just need to add more documentation inside `reference.conf` to give concrete examples of how to set all these knobs depending on what you want to optimize for, and we should be good to go.

Don't hesitate to add as many detailed examples as you can, I'm sure many users will find it useful!

thomash-acinq:
Added some examples.

t-bast (approved):
Neat, thanks!

PR #1817, opened 2021-05-21
Various improvements and fixes

t-bast:
This PR contains many small commits that fix various issues I found looking at our node's logs for the past month.
All commits are independent and should be reviewed separately; each commit description details what is improved/fixed.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2125,11 +2133,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   private def handleLocalError(cause: Throwable, d: Data, msg: Option[Any]) = {
     cause match {
       case _: ForcedLocalCommit => log.warning(s"force-closing channel at user request")
+      case _ if stateName == WAIT_FOR_OPEN_CHANNEL => log.info(s"${cause.getMessage} while processing msg=${msg.getOrElse("n/a").getClass.getSimpleName} in state=$stateName")
```
pm47, line 2136: This was long overdue. How about putting it in `WARN` though?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2125,11 +2133,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
       case _ => log.error(s"${cause.getMessage} while processing msg=${msg.getOrElse("n/a").getClass.getSimpleName} in state=$stateName")
     }
     cause match {
       case _: ChannelException => ()
-      case _ => log.error(cause, s"msg=${msg.getOrElse("n/a")} stateData=$stateData ")
+      case _ => log.error(cause, s"msg=${msg.getOrElse("n/a")} in state=$stateName ")
```
pm47, line 2141: Did you see that happening a lot? Everyting should be a `ChannelException`, we may want more details in case of unexpected errors?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -70,7 +70,12 @@ class PostRestartHtlcCleaner(nodeParams: NodeParams, register: ActorRef, initial
     val nonStandardRelayedOutHtlcs: Map[Origin, Set[(ByteVector32, Long)]] = nodeParams.pluginParams.collect { case p: CustomCommitmentsPlugin => p.getHtlcsRelayedOut(htlcsIn, nodeParams, log) }.flatten.toMap
     val relayedOut: Map[Origin, Set[(ByteVector32, Long)]] = getHtlcsRelayedOut(channels, htlcsIn) ++ nonStandardRelayedOutHtlcs
 
-    val notRelayed = htlcsIn.filterNot(htlcIn => relayedOut.keys.exists(origin => matchesOrigin(htlcIn.add, origin)))
+    val notRelayed = htlcsIn.filterNot(htlcIn => {
+      // If an HTLC has been relayed and then settled downstream, it will not have a matching entry in relayedOut.
+      // When that happens, there will be an HTLC settlement command in the pendingRelay DB, and we will let the channel
+      // replay it instead of sending a conflicting command.
+      relayedOut.keys.exists(origin => matchesOrigin(htlcIn.add, origin)) || nodeParams.db.pendingCommands.listSettlementCommands(htlcIn.add.channelId).exists(_.id == htlcIn.add.id)
+    })
```
pm47, line 78: This would avoid too many calls to the database (not that we are expecting a lot but still).

```suggestion
    val settledHtlcs: Set[(ByteVector32, Long)] = nodeParams.db.pendingCommands.listSettlementCommands().map { case (channelId, cmd) => (channelId, cmd.id) }.toSet

    val notRelayed = htlcsIn.filterNot { htlcIn =>
      // If an HTLC has been relayed and then settled downstream, it will not have a matching entry in relayedOut.
      // When that happens, there will be an HTLC settlement command in the pendingRelay DB, and we will let the channel
      // replay it instead of sending a conflicting command.
      relayedOut.keys.exists(origin => matchesOrigin(htlcIn.add, origin)) || settledHtlcs.contains(htlcIn.add.channelId, htlcIn.add.id)
    }
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -401,7 +401,11 @@ private class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client
     def getUnconfirmedAncestorCount(utxo: Utxo): Future[(ByteVector32, Long)] = client.rpcClient.invoke("getmempoolentry", utxo.txId).map(json => {
       val JInt(ancestorCount) = json \ "ancestorcount"
       (utxo.txId, ancestorCount.toLong)
-    })
+    }).recover {
+      case ex: Throwable =>
+        log.warn(s"could not retrieve unconfirmed ancestor count for txId=${utxo.txId} amount=${utxo.amount}", ex)
```
pm47, line 406:
```suggestion
        log.warn(s"could not retrieve unconfirmed ancestor count for txId=${utxo.txId} amount=${utxo.amount}: ", ex)
```

pm47 (commented):
LGTM, just nits.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: It's a bit noisy, but you're right, a warning would make more sense and it's easy to filter out these log lines if we don't want to see them.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: No I didn't see that one happen, but I looked for places where we logged the whole `stateData` and removed them. But you're right, since this is something that should never happen it makes sense to log all the state to help troubleshoot.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1817/commits/ee9a1e03f284e3e2250b086cffe369d3d5751017

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1817/commits/ee9a1e03f284e3e2250b086cffe369d3d5751017

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1817/commits/ee9a1e03f284e3e2250b086cffe369d3d5751017

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1817/commits/ee9a1e03f284e3e2250b086cffe369d3d5751017

pm47 (approved):

PR #1818, opened 2021-05-21
Don't reveal unannounced channels on relay failures

t-bast:
We may not want to include a `channel_update` for unannounced channel errors because it leaks the channel outpoint (and the fact that an unannounced channel exists between these nodes). An attacker could probe unannounced channels by sending probes and relying on non-strict forwarding.

There are downsides to this change though:

- we can't communicate to an honest payer that feerates have changed since we generated the invoice: they won't be able to retry with the right fees, a new invoice will be needed. But on the other hand, we can't distinguish an honest payer from a malicious one, so our only solution would be to *always* return a dummy channel update *with our up-to-date fees* regardless of whether a channel exists or not?
- even if we fix this, there's another simple way for malicious nodes to probe an unannounced channel between nodes `A` and `B`: send a payment with a random `payment_hash` to `B` via `A`, and if you receive an error from `A` there is indeed a channel between them. `B` would need `A` to be nice and convert that error to an `unknown_next_peer` to be protected, but `B` cannot verify that `A` correctly translates (it would be at best a gentlemen's agreement)...

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -690,7 +690,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       // note: spec would allow us to keep sending new htlcs after having received their shutdown (and not sent ours)
       // but we want to converge as fast as possible and they would probably not route them anyway
       val error = NoMoreHtlcsClosingInProgress(d.channelId)
-      handleAddHtlcCommandError(c, error, Some(d.channelUpdate))
+      handleAddHtlcCommandError(c, error, d, Some(d.channelUpdate))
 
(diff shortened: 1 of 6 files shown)


btcontract:
I'm not sure if this is generally the right way to go. 

What if instead: 
- revealing of something _like_ unannounced channel is always possible, but 
- payee invoice `nodeId` and route hints are always ephemeral, and 
- next to last node would not ever try to route through an unannounced `shortChannelId` which points to a real utxo?

Ephemeral `nodeId` is possible now, hints are solved with trampoline I believe and perhaps something can be done with current hints as well.

t-bast:
Yes a longer-term solution with spec changes is always possible, and route blinding and offers provide good long term solutions here. But I'm wondering whether in the short term we can/should do something or if it's a lost cause?

t-bast:
Closing this: unannounced channels exist merely to avoid polluting the routing graph with channels that won't relay payments, they aren't designed to offer strong privacy guarantees anyway.

To improve privacy we should rework this mechanism more thoroughly.

PR #1822, opened 2021-05-25
Rename pending_relay to pending_commands

pm47:
Naming was confusing because it led to believe messages were related to
htlcs that have not yet been relayed, whereas those are settlement
messages, meaning that htlcs have been relayed and are pending resolution
upstream.

The database has been renamed to a more generic `PendingCommandsDb`
because we may store other types of commands for which we need reliable
delivery.


eclair-core/src/test/scala/fr/acinq/eclair/db/PendingCommandsDbSpec.scala
```diff
@@ -0,0 +1,145 @@
...
+                statement.setString(1, testCase.channelId.toHex)
+                statement.setLong(2, testCase.cmd.id)
+                statement.setBytes(3, cmdCodec.encode(testCase.cmd).require.toByteArray)
+                statement.executeUpdate()
+              }
+            }
+          },
+          dbName = "pending_relay",
+          targetVersion = 2,
+          postCheck = _ =>
```
t-bast, line 100: Do we have a way of checking the table name in the `postCheck`?

t-bast (commented):
LGTM

eclair-core/src/test/scala/fr/acinq/eclair/db/PendingCommandsDbSpec.scala
pm47: We could by calling metadata tables, or even a `count` on the new table, but it feels a bit overkill, and calling `pendingCommands.listSettlementCommands()` does that indirectly.

t-bast (approved):

PR #1823, opened 2021-05-25
Make result set an iterable

pm47:
This allows us to use the full power of scala collections, to iterate
over results, convert to options, etc. while staying purely functional
and immutable.

There is a catch though: the iterator is lazy, it must be materialized
before the result set is closed, by converting the end result in a
collection or an option. In other words, database methods must never
return an `Iterable` or `Iterator`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
@@ -19,15 +19,16 @@ package fr.acinq.eclair.db.jdbc
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.eclair.MilliSatoshi
 import org.sqlite.SQLiteConnection
-import scodec.Codec
+import scodec.Decoder
 import scodec.bits.{BitVector, ByteVector}
 
 import java.sql.{Connection, ResultSet, Statement, Timestamp}
 import java.util.UUID
 import javax.sql.DataSource
-import scala.collection.immutable.Queue
 
 trait JdbcUtils {
+
+  import ExtendedResultSet._
 
   def withConnection[T](f: Connection => T)(implicit dataSource: DataSource): T = {
     val connection = dataSource.getConnection()
@@ -72,15 +73,16 @@ trait JdbcUtils {
   def getVersion(statement: Statement, db_name: String): Option[Int] = {
     createVersionTable(statement)
     // if there was a previous version installed, this will return a different value from current version
-    val rs = statement.executeQuery(s"SELECT version FROM versions WHERE db_name='$db_name'")
-    if (rs.next()) Some(rs.getInt("version")) else None
+    statement.executeQuery(s"SELECT version FROM versions WHERE db_name='$db_name'")
+      .map(rs => rs.getInt("version"))
+      .headOption
   }
 
   /**
    * Updates the version for a particular logical database, it will overwrite the previous version.
    *
    * NB: we could define this method in [[fr.acinq.eclair.db.sqlite.SqliteUtils]] and [[fr.acinq.eclair.db.pg.PgUtils]]
-   *     but it would make testing more complicated because we need to use one or the other depending on the backend.
+   * but it would make testing more complicated because we need to use one or the other depending on the backend.
    */
   def setVersion(statement: Statement, db_name: String, newVersion: Int): Unit = {
     createVersionTable(statement)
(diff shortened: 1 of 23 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
```diff
@@ -96,20 +98,25 @@ trait JdbcUtils {
...
+    /**
+     * Iterates over all rows of a result set.
+     *
+     * Careful: the iterator is lazy, it must be materialized before the [[ResultSet]] is closed, by converting the end
+     * result in a collection or an option.
+     */
+    override def iterator: Iterator[ResultSet] = {
+      // @formatter:off
+      new Iterator[ResultSet] {
+        def hasNext: Boolean = rs.next()
```
t-bast, line 112: Is it ok that `hasNext` actually has a side-effect?
This only works if the calling pattern is always:
```scala
while (it.hasNext) it.next()
```

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
```diff
@@ -182,8 +181,7 @@ class PgNetworkDb(implicit ds: DataSource) extends NetworkDb with Logging {
     inTransaction { pg =>
       using(pg.prepareStatement("SELECT short_channel_id from pruned WHERE short_channel_id=?")) { statement =>
         statement.setLong(1, shortChannelId.toLong)
-        val rs = statement.executeQuery()
-        rs.next()
+        statement.executeQuery().next()
```
t-bast, line 184: Not sure this behaves the same as before, since it's `hasNext` that actually calls `rs.next()` under the hood...

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteNetworkDb.scala
```diff
@@ -171,8 +170,7 @@ class SqliteNetworkDb(sqlite: Connection) extends NetworkDb with Logging {
   override def isPruned(shortChannelId: ShortChannelId): Boolean = withMetrics("network/is-pruned", DbBackends.Sqlite) {
     using(sqlite.prepareStatement("SELECT short_channel_id from pruned WHERE short_channel_id=?")) { statement =>
       statement.setLong(1, shortChannelId.toLong)
-      val rs = statement.executeQuery()
-      rs.next()
+      statement.executeQuery().next()
```
t-bast, line 173: Same here, is this correctly calling `hasNext` before calling `next()` internally?

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
pm47: Yes it works, because we always use this pattern to iterate once. Attempting to iterate twice on the same result set will yield no result on the second iteration, but that was already the case before this change.

However the abstraction does make it easier to shoot yourself in the foot because it hides away this kind of details.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
pm47: Yeah, I wasn't sure about this one either, but there is a unit test so I guess it works. How about:

```scala
statement.executeQuery().headOption.nonEmpty
```

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteNetworkDb.scala
pm47: Fixed with 305d0914d.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
pm47: Fixed with 305d0914d.

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
t-bast: Our tests should catch new code that doesn't comply with this pattern, so we're probably ok.

t-bast (approved):

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
pm47: To be clear, this change really is a trade-off. It makes some things safer but can be riskier in other ways, and I'm not sure our tests will catch everything.

pm47: 🤞

PR #1825, opened 2021-05-25
Add a random delay before processing blocks

pm47:
The goal is to reduce herd effect when there are lots of channels.


eclair-core/src/main/resources/reference.conf
```diff
@@ -84,6 +84,7 @@ eclair {
   // expiry-delta-blocks.
   fulfill-safety-before-timeout-blocks = 24
   min-final-expiry-delta-blocks = 30 // Bolt 11 invoice's min_final_cltv_expiry; must be strictly greater than fulfill-safety-before-timeout-blocks
+  max-block-processing-delay = 10 seconds // we add a random delay before processing blocks, capped at this value, to prevent herd effect
```
t-bast, line 87: I think a default of 30 seconds would be better if we want the effect to be really visible, don't you think? 10 seconds feels too small.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1783,8 +1786,13 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     // we only care about this event in NORMAL and SHUTDOWN state, and there may be cases where the task is not cancelled
     case Event(_: RevocationTimeout, _) => stay
 
+    // we reschedule with a random delay to prevent herd effect when there are a lot of channels
+    case Event(c@CurrentBlockCount(_), _) =>
```
t-bast, line 1790: nit:
```suggestion
    case Event(c: CurrentBlockCount, _) =>
```

eclair-core/src/main/resources/reference.conf
pm47:
```suggestion
  max-block-processing-delay = 30 seconds // we add a random delay before processing blocks, capped at this value, to prevent herd effect
```

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -111,6 +111,7 @@ object TestConstants {
       expiryDelta = CltvExpiryDelta(144),
       fulfillSafetyBeforeTimeout = CltvExpiryDelta(6),
       minFinalExpiryDelta = CltvExpiryDelta(18),
+      maxNewBlockDelay = 10 millis,
```
pm47, line 114:
```suggestion
      maxBlockProcessingDelay = 10 millis,
```

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -215,6 +216,7 @@ object TestConstants {
...
```
pm47, line 219:
```suggestion
      maxBlockProcessingDelay = 10 millis,
```

t-bast (dismissed):

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
```diff
@@ -77,7 +77,8 @@ abstract class IntegrationSpec extends TestKitBaseClass with BitcoindService wit
     "eclair.bitcoind.wallet" -> defaultWallet,
     "eclair.mindepth-blocks" -> 2,
     "eclair.max-htlc-value-in-flight-msat" -> 100000000000L,
-    "eclair.router.broadcast-interval" -> "2 second",
+    "eclair.max-block-processing-delay" -> "2 seconds",
```
t-bast, line 80: Good catch! We don't want the test to become crazy slow :)

t-bast (approved):

eclair-core/src/test/scala/fr/acinq/eclair/integration/IntegrationSpec.scala
pm47: Tests were actually failing because of timeouts...

PR #1826, opened 2021-05-25
Add metrics on channels processing time

pm47:
It's similar with built-in akka metrics, but with message-type granularity.


t-bast (approved):
Nice, this will be helpful in figuring out what bottlenecks appear in real-world settings!
It's going to create a lot of data though, I think that long-term it's something we may want to disable in production.

pm47:
> It's going to create a lot of data though

It should be properly sampled by Kamon AFAIK?

PR #1827, opened 2021-05-27
Unlock utxos when TxPublish fails

t-bast:
This PR refactors the `TxPublisher` and adds some new features:

- We now spawn a FSM for each tx publication: this makes the publish steps more clear and easier to extend
- We return a `TxPublishResult`, that will in the future allow the parent `TxPublisher` to manage the list of pending txs and RBF when needed
- We don't spend our anchor output when a commit tx is already confirmed
- We unlock utxos when tx publishing fails (which happens when we restart eclair, which re-sends channel txs to the publisher who will try to publish them but fail because they're already in the mempool)

It can be useful to review commit by commit to see each change individually.


pm47:
Regarding the first commit (9ed346e):
- have you considered handing the delayed txs in the `TxPublish` actor? Just my first reaction when looking at the code
- it seems the new asynchronous publication effectively bypasses the `singleThreadExecutionContext` and we lose the sequentiality guarantee

t-bast:
> have you considered handing the delayed txs in the TxPublish actor? Just my first reaction when looking at the code

Yes, all txs will be handled by the `TxPublish` actor in the future, but one step at a time (I start with the ones that *need* fee-bumping and will later add the ones for which it's optional).

> it seems the new asynchronous publication effectively bypasses the singleThreadExecutionContext and we lose the sequentiality guarantee

That's true. But I really think moving to dedicated actors is important before adding more features...
The only tx that needs to be broadcast first is the commit tx, maybe we can explicitly re-broadcast it in the first step?

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
```diff
@@ -239,192 +144,28 @@ private class TxPublisher(nodeParams: NodeParams, watcher: ActorRef[ZmqWatcher.C
...
-          case tx: HtlcTx => publishHtlcTx(tx, commitments)
+        toPublish.values.flatten.foreach(tx => publish(tx, channelInfo))
+        run(cltvDelayedTxs -- toPublish.keys, csvDelayedTxs, channelInfo)
+
+      case WrappedTxPublishResult(result) =>
+        result match {
+          case _: TxPublish.InsufficientFunds =>
+            // We retry when the next block has been found, we may have more funds available.
+            val nextBlockCount = nodeParams.currentBlockHeight + 1
+            val cltvDelayedTxs1 = cltvDelayedTxs + (nextBlockCount -> (cltvDelayedTxs.getOrElse(nextBlockCount, Seq.empty) :+ result.cmd))
```
pm47, line 155: That's a bit hacky, the transaction isn't really cltv-delayed here. Also, there is a potential risk of creating a herd effect if our wallet balance is low. That would be a nice stress test of our ability to concurrently fund many transactions and manages locks, but I'm not sure it's a good thing.

pm47:
> > it seems the new asynchronous publication effectively bypasses the singleThreadExecutionContext and we lose the sequentiality guarantee

> That's true. But I really think moving to dedicated actors is important before adding more features...
The only tx that needs to be broadcast first is the commit tx, maybe we can explicitly re-broadcast it in the first step and completely get rid of singleThreadExecutionContext?

IIRC, the most probable scenario would be local-commit and htlc-success

t-bast:
The only scenario where it's an issue is the local commit tx, it must be published before its children, otherwise the rest is CSV-delayed so it's not impacted.

eclair-core/src/main/scala/fr/acinq/eclair/channel/TxPublisher.scala
t-bast: That will be improved later, when we'll introduce deadlines (block heights at which we should republish/bump some txs), but that requires more work. In the meantime I'd rather err on the safety side and republish aggressively to ensure we don't forget to publish an important transaction because of an unknown error.

t-bast:
> it seems the new asynchronous publication effectively bypasses the singleThreadExecutionContext and we lose the sequentiality guarantee

This is fixed in https://github.com/ACINQ/eclair/pull/1827/commits/81a1b5179a8933c268f844d77e0638dd2c3ed68a
I chose to optimistically publish txs and react on failures by publishing the parent first.

t-bast:
Closed in favor of #1844

Issue #1829, opened 2021-05-28
Dangling proposed HTLCs on force-close


btcontract:
I'm not completely sure but it looks like `localCommit.proposed` HTLCs are left untreated currently specifically when channel gets force-closed (But they do get failed when channel gets offline before we have a chance to sign them). 

I think this will result in upstream force-close for routed HTLCs in such a state.

pm47:
Nice catch Anton!

PR #1832, opened 2021-05-31
Fail unsigned outgoing htlcs on force-close

pm47:
See #1829.

We should fail outgoing _unsigned_ htlcs on force close, just like we do when disconnected.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1790,7 +1790,16 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         case data: HasCommitments =>
           val replyTo = if (c.replyTo == ActorRef.noSender) sender else c.replyTo
           replyTo ! RES_SUCCESS(c, data.channelId)
-          handleLocalError(ForcedLocalCommit(data.channelId), data, Some(c))
+          val failure = ForcedLocalCommit(data.channelId)
+          // if we are in NORMAL state, we may have outgoing unsigned htlcs that we can fail right away
+          data match {
+            case _: DATA_NORMAL =>
+              data.commitments.localChanges.proposed.collect {
+              case add: UpdateAddHtlc => relayer ! RES_ADD_SETTLED(data.commitments.originChannels(add.id), add, HtlcResult.OnChainFail(failure))
+            }
+            case _ => ()
+          }
+          handleLocalError(failure, data, Some(c))
         case _ => handleCommandError(CommandUnavailableInThisState(d.channelId, "forceclose", stateName), c)
       }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1791,14 +1791,6 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           val replyTo = if (c.replyTo == ActorRef.noSender) sender else c.replyTo
           replyTo ! RES_SUCCESS(c, data.channelId)
           val failure = ForcedLocalCommit(data.channelId)
-          // if we are in NORMAL state, we may have outgoing unsigned htlcs that we can fail right away
-          data match {
-            case _: DATA_NORMAL =>
-              data.commitments.localChanges.proposed.collect {
-              case add: UpdateAddHtlc => relayer ! RES_ADD_SETTLED(data.commitments.originChannels(add.id), add, HtlcResult.OnChainFail(failure))
-            }
-            case _ => ()
-          }
           handleLocalError(failure, data, Some(c))
         case _ => handleCommandError(CommandUnavailableInThisState(d.channelId, "forceclose", stateName), c)
       }
@@ -2216,6 +2208,17 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     blockchain ! WatchTxConfirmed(self, closingTx.tx.txid, nodeParams.minDepthBlocks)
   }
 
+  /** We fail outgoing unsigned htlcs right away when transitioning from NORMAL to CLOSING */
+  private def failUnsignedProposedHtlcs(d: HasCommitments) = {
+    d match {
+      case dd: DATA_NORMAL =>
+        dd.commitments.localChanges.proposed.collect {
+          case add: UpdateAddHtlc => relayer ! RES_ADD_SETTLED(d.commitments.originChannels(add.id), add, HtlcResult.ChannelFailureBeforeCommitted)
+        }
+      case _ => ()
+    }
+  }
+
   private def spendLocalCurrent(d: HasCommitments) = {
     val outdatedCommitment = d match {
       case _: DATA_WAIT_FOR_REMOTE_PUBLISH_FUTURE_COMMITMENT => true
@@ -2226,6 +2229,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       log.warning("we have an outdated commitment: will not publish our local tx")
       stay
     } else {
+      failUnsignedProposedHtlcs(d)
       val commitTx = d.commitments.fullySignedLocalCommitTx(keyManager).tx
       val localCommitPublished = Helpers.Closing.claimCurrentLocalCommitTxOutputs(keyManager, d.commitments, commitTx, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)
       val nextData = d match {
@@ -2301,6 +2305,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     log.warning(s"they published their current commit in txid=${commitTx.txid}")
     require(commitTx.txid == d.commitments.remoteCommit.txid, "txid mismatch")
 
+    failUnsignedProposedHtlcs(d)
     val remoteCommitPublished = Helpers.Closing.claimRemoteCommitTxOutputs(keyManager, d.commitments, d.commitments.remoteCommit, commitTx, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)
     val nextData = d match {
       case closing: DATA_CLOSING => closing.copy(remoteCommitPublished = Some(remoteCommitPublished))
@@ -2332,6 +2337,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     val remoteCommit = d.commitments.remoteNextCommitInfo.left.get.nextRemoteCommit
     require(commitTx.txid == remoteCommit.txid, "txid mismatch")
 
+    failUnsignedProposedHtlcs(d)
     val remoteCommitPublished = Helpers.Closing.claimRemoteCommitTxOutputs(keyManager, d.commitments, remoteCommit, commitTx, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)
     val nextData = d match {
       case closing: DATA_CLOSING => closing.copy(nextRemoteCommitPublished = Some(remoteCommitPublished))
@@ -2361,6 +2367,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
   private def handleRemoteSpentOther(tx: Transaction, d: HasCommitments) = {
     log.warning(s"funding tx spent in txid=${tx.txid}")
+    failUnsignedProposedHtlcs(d)
     Helpers.Closing.claimRevokedRemoteCommitTxOutputs(keyManager, d.commitments, tx, nodeParams.db.channels, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets) match {
       case Some(revokedCommitPublished) =>
         log.warning(s"txid=${tx.txid} was a revoked commitment, publishing the penalty tx")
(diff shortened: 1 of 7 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -1915,6 +1915,24 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
+    alice ! WatchFundingDeeplyBuriedTriggered(400000, 42, null)
+    val update1a = alice2bob.expectMsgType[ChannelUpdate]
+    assert(Announcements.isEnabled(update1a.channelFlags))
+    val (_, htlc1) = addHtlc(10000 msat, alice, bob, alice2bob, bob2alice, sender.ref)
+    sender.expectMsgType[RES_SUCCESS[CMD_ADD_HTLC]]
+    val aliceData = alice.stateData.asInstanceOf[DATA_NORMAL]
+    assert(aliceData.commitments.localChanges.proposed.size == 1)
+
+    // actual test starts here
+    Thread.sleep(1100)
```
sstone, line 1930: Why do we need to wait here ?

btcontract:
Looks like this only works on user-issued force-close command currently? I'd guess that would be the least likely occurrence, more probable case is something like we are sending HTLCs and suddenly see remote commit published.

pm47:
> Looks like this only works on user-issued force-close command currently? I'd guess that would be the least likely occurrence, more probable case is something like we are sending HTLCs and suddenly see remote commit published.

I agree the fix should be apply and handle all cases of channel closing.

t-bast:
I think the right way to fix this is here: https://github.com/ACINQ/eclair/blob/dbecb28d96d780668fc006aeee2a043a1cc52d10/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1428

We should wait for the commit tx to be confirmed and then fail all these proposed HTLCs, either by modifying `overriddenOutgoingHtlcs` to return them or by explicitly handling them separately.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
pm47: For no good reason (it was a leftover from another test that involved `channel_update`s and second-precision timestamps. Fixed in 559dc3c.

pm47:
> We should wait for the commit tx to be confirmed and then fail all these proposed HTLCs, either by modifying overriddenOutgoingHtlcs to return them or by explicitly handling them separately.

I disagree, because in the case of an overridden htlc, that htlc was previously committed, which is why we need to make sure that a competing commitment won the race. Here, since the htlc has not been signed, it will be overridden as soon as we enter the closing procedure.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1036,12 +1036,13 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+      // if we have pending unsigned htlcs, then we cancel them and generate and update with the disabled flag set, that will be returned to the sender in a temporary channel failure
       val d1 = if (d.commitments.localChanges.proposed.collectFirst { case add: UpdateAddHtlc => add }.isDefined) {
         log.info("updating channel_update announcement (reason=disabled)")
         val channelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, d.channelUpdate.cltvExpiryDelta, d.channelUpdate.htlcMinimumMsat, d.channelUpdate.feeBaseMsat, d.channelUpdate.feeProportionalMillionths, d.commitments.capacity.toMilliSatoshi, enable = false)
+        // NB: the htlcs stay in the commitments.localChange, they will be cleaned up after reconnection
         d.commitments.localChanges.proposed.collect {
-          case add: UpdateAddHtlc => relayer ! RES_ADD_SETTLED(d.commitments.originChannels(add.id), add, HtlcResult.Disconnected(channelUpdate))
+          case add: UpdateAddHtlc => relayer ! RES_ADD_SETTLED(d.commitments.originChannels(add.id), add, HtlcResult.DisconnectedBeforeSigned(channelUpdate))
```
pm47, line 1045: NB: we can't factor that in the same transition handler as for `NORMAL`->`CLOSING` because there is a constraint on the  `channel_update`, which must have the disabled flag set.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -114,7 +114,10 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
         case HtlcResult.OnChainFail(cause) =>
           // if the outgoing htlc is being resolved on chain, we treat it like a local error but we cannot retry
           handleLocalFail(d, cause, isFatal = true)
-        case HtlcResult.Disconnected(_) =>
+        case HtlcResult.ChannelFailureBeforeSigned =>
+          // if the outgoing htlc is being resolved on chain, we treat it like a local error but we cannot retry
+          handleLocalFail(d, ChannelFailureException, isFatal = true)
```
t-bast, line 119: This one isn't fatal, we can retry with another channel since we don't need to wait to settle the htlc on-chain.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -1915,6 +1915,21 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
+    val sender = TestProbe()
+    val (_, htlc1) = addHtlc(10000 msat, alice, bob, alice2bob, bob2alice, sender.ref)
+    sender.expectMsgType[RES_SUCCESS[CMD_ADD_HTLC]]
+    val aliceData = alice.stateData.asInstanceOf[DATA_NORMAL]
+    assert(aliceData.commitments.localChanges.proposed.size == 1)
+
+    // actual test starts here
+    alice ! CMD_FORCECLOSE(sender.ref)
+    sender.expectMsgType[RES_SUCCESS[CMD_FORCECLOSE]]
+    val addSettled = relayerA.expectMsgType[RES_ADD_SETTLED[Origin, HtlcResult.ChannelFailureBeforeSigned.type ]]
```
t-bast, line 1929: nit:
```suggestion
    val addSettled = relayerA.expectMsgType[RES_ADD_SETTLED[Origin, HtlcResult.ChannelFailureBeforeSigned.type]]
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -2392,6 +2407,22 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
+    val (_, htlc1) = addHtlc(10000 msat, alice, bob, alice2bob, bob2alice, sender.ref)
+    sender.expectMsgType[RES_SUCCESS[CMD_ADD_HTLC]]
+    val aliceData = alice.stateData.asInstanceOf[DATA_NORMAL]
+    assert(aliceData.commitments.localChanges.proposed.size == 1)
+
+    // actual test starts here
+    // bob publishes his current commit tx
+    val bobCommitTx = bob.stateData.asInstanceOf[DATA_NORMAL].commitments.localCommit.publishableTxs.commitTx.tx
+    alice ! WatchFundingSpentTriggered(bobCommitTx)
+    val addSettled = relayerA.expectMsgType[RES_ADD_SETTLED[Origin, HtlcResult.ChannelFailureBeforeSigned.type ]]
```
t-bast, line 2422:
```suggestion
    val addSettled = relayerA.expectMsgType[RES_ADD_SETTLED[Origin, HtlcResult.ChannelFailureBeforeSigned.type]]
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -2651,6 +2728,20 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
+    import f._
+    val sender = TestProbe()
+    val (_, htlc1) = addHtlc(10000 msat, alice, bob, alice2bob, bob2alice, sender.ref)
+    sender.expectMsgType[RES_SUCCESS[CMD_ADD_HTLC]]
+    val aliceData = alice.stateData.asInstanceOf[DATA_NORMAL]
+    assert(aliceData.commitments.localChanges.proposed.size == 1)
+
+    // actual test starts here
+    alice ! Error(ByteVector32.Zeroes, "oops")
+    val addSettled = relayerA.expectMsgType[RES_ADD_SETTLED[Origin, HtlcResult.ChannelFailureBeforeSigned.type ]]
```
t-bast, line 2741:
```suggestion
    val addSettled = relayerA.expectMsgType[RES_ADD_SETTLED[Origin, HtlcResult.ChannelFailureBeforeSigned.type]]
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -2805,8 +2896,12 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
     // actual test starts here
     Thread.sleep(1100)
     alice ! INPUT_DISCONNECTED
-    assert(relayerA.expectMsgType[RES_ADD_SETTLED[Origin, HtlcResult.Disconnected]].htlc.paymentHash === htlc1.paymentHash)
-    assert(relayerA.expectMsgType[RES_ADD_SETTLED[Origin, HtlcResult.Disconnected]].htlc.paymentHash === htlc2.paymentHash)
+    val addSettled1 = relayerA.expectMsgType[RES_ADD_SETTLED[Origin, HtlcResult]]
+    assert(addSettled1.htlc == htlc1)
+    assert(addSettled1.result.isInstanceOf[HtlcResult.DisconnectedBeforeSigned])
```
t-bast, line 2901: nit: maybe we can also check that the channel update has the `disabled` bit set?

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
```diff
@@ -361,7 +361,7 @@ class ChannelRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("a
       TestCase(HtlcResult.RemoteFail(UpdateFailHtlc(channelId1, downstream_htlc.id, hex"deadbeef")), CMD_FAIL_HTLC(r.add.id, Left(hex"deadbeef"), commit = true)),
       TestCase(HtlcResult.RemoteFailMalformed(UpdateFailMalformedHtlc(channelId1, downstream_htlc.id, ByteVector32.One, FailureMessageCodecs.BADONION)), CMD_FAIL_MALFORMED_HTLC(r.add.id, ByteVector32.One, FailureMessageCodecs.BADONION, commit = true)),
       TestCase(HtlcResult.OnChainFail(HtlcOverriddenByLocalCommit(channelId1, downstream_htlc)), CMD_FAIL_HTLC(r.add.id, Right(PermanentChannelFailure), commit = true)),
-      TestCase(HtlcResult.Disconnected(u_disabled.channelUpdate), CMD_FAIL_HTLC(r.add.id, Right(TemporaryChannelFailure(u_disabled.channelUpdate)), commit = true))
+      TestCase(HtlcResult.DisconnectedBeforeSigned(u_disabled.channelUpdate), CMD_FAIL_HTLC(r.add.id, Right(TemporaryChannelFailure(u_disabled.channelUpdate)), commit = true))
```
t-bast, line 364: nit: also add one for `ChannelFailureBeforeSigned` for completeness?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1911,6 +1915,17 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       }
   }
 
+  /** Fail outgoing unsigned htlcs right away when transitioning from NORMAL to OFFLINE or CLOSING */
```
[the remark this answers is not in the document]
pm47:
```suggestion
  /** Fail outgoing unsigned htlcs right away when transitioning from NORMAL to  CLOSING */
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
pm47: Good catch!

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
pm47: It's already checked on line 2905:
```scala
assert(!Announcements.isEnabled(channelUpdateListener.expectMsgType[LocalChannelUpdate].channelUpdate.channelFlags))
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
pm47: Done in c6ed34f3d.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
pm47: Done in c6ed34f3d.

t-bast (approved):

pm47:
Rebased on master.

t-bast (approved):

PR #1833, opened 2021-06-01
Include routing hints in parseinvoice API call response

sstone:
JSON objects returned by parseinvoice do not include routing hints, which can be misleading when decoding payment requests created by mobile wallets.


pm47 (approved):
LGTM

PR #1835, opened 2021-06-03
(Minor) Fix eventually statements

pm47:
With the move to akka _typed_, we will be using more and more
scalatest's `eventually` as a replacement for akka's `awaitCond`
(which isn't available in `testkit.typed`).

But there is a catch:
- `awaitCond` expects a boolean
- `eventually` expects a non-failure

Which means that we must use `eventually(assert(cond))`, and not
`eventually(cond)`.


t-bast (approved):
Good catch, we need to get used to that change in behavior

PR #1836, opened 2021-06-03
Symmetrical HTLC limits (#1828)

sstone:
The spec defines `max_accepted_htlcs` and `max_htlc_value_in_flight_msat`
to let nodes reduce their exposure to pending HTLCs. This only applies to
received HTLCs, and we use the remote peer's values for outgoing HTLCs.

But when we're more restrictive than our peer, it makes sense to apply our
limits to outgoing HTLCs as well.


pm47 (approved):
AFAICT, this only affects outgoing htlcs, so the worst that can happen is that users cannot make payments. There is no risk of force close.

dpad85 (approved):

Issue #1830, opened 2021-05-29
Location of plugin-resources


engenegr:
Currently Eclair looks for `plugin-resources` directory from inside the directory `run-eclair` was invoked.

Is it possible for node to look for plugin configs relatively to `.eclair` dir or `eclair.conf`?

btcontract:
A bit to clarify here: `plugin-resources` is a directory I've started using when I realized that my HC plugin needs a config file of its own and also other file stuff (such as icon image). So I've decided to introduce a `./plugin-resources/<plugin-name>/` approach which is used in HC plugin and now in alarm bot plugin. 

I guess if this makes sense then at some point these approaches need to be solidified on some "plugin best practices" section or something like that.

engenegr:
Thank you very much @btcontract. I didn't realize this is only hosted-channels thing and did the same in Telegram warnings bot just because it is convenient to have separate config.

PR #1837, opened 2021-06-05
Make `Setup.datadir` a val

btcontract:
This will allow plugins to know where the Eclair datadir is and look for their files in that dir. Fixes https://github.com/ACINQ/eclair/issues/1830


t-bast (commented):
Are you sure this changes anything?
It seems to me that the existing code already makes it a `val`.

btcontract:
`Setup` is not a case class where all constructor parameters are `val`s by default, without adding a `val` explicitly `datadir` is not visible outside of `Setup`.

t-bast:
So it's for visibility, not immutability.
I thought it was to ensure that internally setup couldn't mutate that `datadir`.
For visibility that makes sense, LGTM.

t-bast (approved):

PR #1838, opened 2021-06-07
Alternate strategies for outdated commitments

pm47:
This PRs adds strategies to handle two typical disaster scenarios: outdated commitments and unhandled exceptions.

Default strategies may be the best choice for smaller loosely administered nodes, while alternative strategies may avoid unnecessary mass force-close (but are reserved for advanced users who closely monitor the node).

Strategies for outdated commitments:
- request the counterparty to close the channel (default).
- if the node was restarted less than 10 min ago, log an error message and stop the node

Strategies for unhandled exceptions:
- local force close of the channel (default)
- log an error message and stop the node

Default settings maintain the same behavior as before.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -95,6 +95,10 @@ eclair {
   max-block-processing-delay = 30 seconds // we add a random delay before processing blocks, capped at this value, to prevent herd effect
   max-tx-publish-retry-delay = 60 seconds // we add a random delay before retrying failed transaction publication
 
+  // see docs/Advanced.md for more information on the strategies
+  outdated-commitment-strategy = "remote-close" // remote-close or stop (NB: the app will only stop if it was recently restarted)
+  unhandled-exception-strategy = "local-close" // local-close or stop
+
   relay {
     fees {
       // Fees for public channels
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -127,6 +127,38 @@ object Channel {
   /** We don't immediately process [[CurrentBlockCount]] to avoid herd effects */
   case class ProcessCurrentBlockCount(c: CurrentBlockCount)
 
+  // @formatter:off
+  /** What do we do if we detect that our local commitment is outdated. */
+  sealed trait OutdatedCommitmentStrategy
+  object OutdatedCommitmentStrategy {
+    /**
+     * Ask our counterparty to close the channel, whenever our peer proves to us *or* simply tells us (could be lying)
+     * that we are using an outdated commitment.
+     * This may be the best choice for smaller loosely administered nodes.
+     */
+    case object RemoteClose extends OutdatedCommitmentStrategy
+    /**
+     * If the node was just restarted, just log an error and stop the app. The goal is to prevent unwanted mass
+     * force-close of channels if we accidentally restarted the node with an outdated backup. After a few minutes, we
+     * revert to the default behavior of requesting our peer to force close, otherwise this opens a huge attack vector
+     * where any peer can remotely stop our node.
+     * This strategy may be better suited for larger nodes, closely administered.
+     */
+    case object Stop extends OutdatedCommitmentStrategy
+  }
+  // @formatter:on
+
+  // @formatter:off
+  /** What do we do if we have a local unhandled exception. */
+  sealed trait UnhandledExceptionStrategy
+  object UnhandledExceptionStrategy {
+    /** Ask our counterparty to close the channel. This may be the best choice for smaller loosely administered nodes.*/
+    case object LocalClose extends UnhandledExceptionStrategy
+    /** Just log an error and stop the node. May be better for larger nodes, to prevent unwanted mass force-close.*/
+    case object Stop extends UnhandledExceptionStrategy
+  }
+  // @formatter:on
+
 }
 
 class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remoteNodeId: PublicKey, blockchain: typed.ActorRef[ZmqWatcher.Command], relayer: ActorRef, txPublisherFactory: Channel.TxPublisherFactory, origin_opt: Option[ActorRef] = None)(implicit ec: ExecutionContext = ExecutionContext.Implicits.global) extends FSM[ChannelState, ChannelData] with FSMDiagnosticActorLogging[ChannelState, ChannelData] {
@@ -1669,6 +1701,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
         case syncSuccess: SyncResult.Success =>
           var sendQueue = Queue.empty[LightningMessage]
           // normal case, our data is up-to-date
+
           if (channelReestablish.nextLocalCommitmentNumber == 1 && d.commitments.localCommit.index == 0) {
             // If next_local_commitment_number is 1 in both the channel_reestablish it sent and received, then the node MUST retransmit funding_locked, otherwise it MUST NOT
             log.debug("re-sending fundingLocked")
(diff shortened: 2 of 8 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2399,6 +2444,19 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     goto(ERR_INFORMATION_LEAK) calling doPublish(localCommitPublished, d.commitments) sending error
   }
 
+  private def handleOutdatedCommitment(channelReestablish: ChannelReestablish, d: HasCommitments) = {
+    nodeParams.outdatedCommitmentStrategy match {
```
t-bast, line 2448: That pattern match is incomplete, we don't handle the case `HaltIfJustRestarted` case after 10 minutes (so we'll throw a runtime exception).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2399,6 +2444,19 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+      case OutdatedCommitmentStrategy.HaltIfJustRestarted if ManagementFactory.getRuntimeMXBean.getUptime.millis < 10.minutes =>
+        log.error("we just restarted and may have an outdated commitment! stopping the node")
```
t-bast, line 2450: I'd put a longer, more thorough message, something like: "we just restarted and may have an outdated commitment: the channel should force-close but the node is stopping instead, please ensure your database is up-to-date before restarting"

t-bast (commented):
I don't think these two commits should be done in the same PR, I'm not sure the second one makes a lot of sense. In the long term it makes sense to change how we deal with remote errors and not always broadcast our local-commit, but that's probably better addressed by working on https://github.com/lightningnetwork/lightning-rfc/pull/834 and reviewing in which case we don't have a choice and really need to force-close.

As for the first commit, what do you do when the node stops and you do have your most up-to-date backup? If you restart it will stop again. So you'd need to change your config from "halt-if-just-restarted" to "always-request-remote-close" before restarting, otherwise a single faulty channel would DoS the whole node...

pm47:
> I don't think these two commits should be done in the same PR

Agreed, I got a bit carried away and kept experimenting.

> I'm not sure the second one makes a lot of sense. In the long term it makes sense to change how we deal with remote errors and not always broadcast our local-commit, but that's probably better addressed by working on lightningnetwork/lightning-rfc#834 and reviewing in which case we don't have a choice and really need to force-close.

IIUC lightningnetwork/lightning-rfc#834 deals with protocol errors, but here I'm trying to address the case where we have an internal error. Disk full, OOM exception, that sort of things.

> As for the first commit, what do you do when the node stops and you do have your most up-to-date backup? If you restart it will stop again. So you'd need to change your config from "halt-if-just-restarted" to "always-request-remote-close" before restarting, otherwise a single faulty channel would DoS the whole node...

Exactly, it requires human intervention, that's why this mode is for more advanced users. But it's less susceptible to blind mass force-close, which is what this PR aims to address. I feel like any other solution in-between would be a trade-off between convenience and security, that's how I ended with this rather basic mechanism.

t-bast:
> Exactly, it requires human intervention, that's why this mode is for more advanced users.

So the flow if the node stops because of an outdated commitment would be:

1. The node stops
2. Investigate your backups
3. If you had a DB issue, restart after fixing your DB
4. Otherwise if you find nothing and think you really are late, change your config to `always-request-remote-close` and restart, and accept losing channels

Is that right?
I think it should be documented clearly, probably a new document in the `docs` folder.

> I feel like any other solution in-between would be a trade-off between convenience and security, that's how I ended with this rather basic mechanism.

I agree, the other solutions I can think of are just proxies for almost the same behavior, so it's better to choose the simplest one.
I don't think it's an area where we'll want many different strategies though, so maybe the new configuration parameter should just be a boolean `stop-node-on-outdated-commitment-after-restart = false` or something similar?

pm47:
> I think it should be documented clearly, probably a new document in the docs folder.

Agree!

> maybe the new configuration parameter should just be a boolean stop-node-on-outdated-commitment-after-restart = false or something similar?

It was a left-over of my previous attempt, I flip-flopped around this and thought keeping an explicit strategy name would be easier to understand ... but it is probably not.

eclair-core/src/main/resources/reference.conf
```diff
@@ -86,6 +86,8 @@ eclair {
   fulfill-safety-before-timeout-blocks = 24
   min-final-expiry-delta-blocks = 30 // Bolt 11 invoice's min_final_cltv_expiry; must be strictly greater than fulfill-safety-before-timeout-blocks
   max-block-processing-delay = 30 seconds // we add a random delay before processing blocks, capped at this value, to prevent herd effect
+  outdated-commitment-strategy = "always-request-remote-close" // always-request-remote-close or halt-if-just-restarted
+  unhandled-exception-strategy = "local-force-close" // local-force-close or log-and-stop
```
pm47, line 90: >> maybe the new configuration parameter should just be a boolean stop-node-on-outdated-commitment-after-restart = false or something similar?

> It was a left-over of my previous attempt, I flip-flopped around this and thought keeping an explicit strategy name would be easier to understand ... but it is probably not.

Revisiting this PR, I think explicit names are better because they are self-explaining. Also, there are actually four different strategies, so a boolean would be a bit ambiguous (`true` may mean remote-close or local-close):

```
  stop-on-outdated-commitment = false
  stop-on-unhandled-exception = false
```

Preferred alternative:
```
  outdated-commitment-strategy = "remote-close" // remote-close or stop (NB: the app will only stop if it was recently restarted)
  unhandled-exception-strategy = "local-close" // local-close or stop
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Compiler now catches this, since we have re-enabled the `fatal-warning` flags.

```
match may not be exhaustive.
It would fail on the following input: Stop
    nodeParams.outdatedCommitmentStrategy match {

```

eclair-core/src/main/resources/reference.conf
pm47: Done with a25bc6a.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Done with a25bc6a.

pm47:
>> I don't think these two commits should be done in the same PR,

> Agreed, I got a bit carried away and kept experimenting.

On 2nd thought, I kept both changes in the same PR, I think it makes sense since those are related topics.

> I'm not sure the second one makes a lot of sense. In the long term it makes sense to change how we deal with remote errors and not always broadcast our local-commit, but that's probably better addressed by working on lightningnetwork/lightning-rfc#834 and reviewing in which case we don't have a choice and really need to force-close.

I could add a third strategy, to not force-close even if our peer sends an `error`, like `lnd` does (used to do?). But that would'nt be spec compliant, especially since we have `warning` messages now.

> I think it should be documented clearly, probably a new document in the docs folder.

Done with 93d07c4ec.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2256,14 +2285,15 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-      case _ if stateName == WAIT_FOR_OPEN_CHANNEL => log.warning(s"${cause.getMessage} while processing msg=${msg.getOrElse("n/a").getClass.getSimpleName} in state=$stateName")
-      case _ => log.error(s"${cause.getMessage} while processing msg=${msg.getOrElse("n/a").getClass.getSimpleName} in state=$stateName")
-    }
-    cause match {
-      case _: ChannelException => ()
-      case _ => log.error(cause, s"msg=${msg.getOrElse("n/a")} stateData=$stateData")
+      case _: ForcedLocalCommit =>
+        log.warning(s"force-closing channel at user request")
+      case _: ChannelException =>
+        log.error(s"${cause.getMessage} while processing msg=${msg.getOrElse("n/a").getClass.getSimpleName} in state=$stateName")
```
t-bast, line 2291: Why did you remove the specific `WAIT_FOR_OPEN_CHANNEL` case? It was quite handy to avoid polluting the error logs with invalid parameters from remote peers in `open_channel`.

t-bast (commented):
The end result feels much cleaner than previously, it's easier to simply read through the functions in `Helpers.Syncing` to check that all cases are properly handled.

The behavior looks good to me, I just added many comments to help clarify the code.

t-bast (commented):
Don't forget to also add an entry in the release notes for this, it's really newsworthy ;)

leexmario-oss:
Hey @t-bast — could you review **“Alternate strategies for outdated commitments”** when you get a chance? It hasn’t been updated since October 27, 2021. Thanks!

PR #1839, opened 2021-06-07
Use bitcoin-lib 0.19

sstone:
There are no functional changes, but bitcoin-lib 0.19 is based on secp256k1-kmp (instead of our own fork of secp256k1's JNI wrapper) which is cleaner, easier to maintain and used in our mobile apps.


t-bast (approved):
I did a few tests on regtest and combed through the logs, everything looks good :rocket:

PR #1844, opened 2021-06-18
Rework TxPublisher

t-bast:
This PR is an alternative approach to #1827

It contains more architectural changes (and more code to review) but is a cleaner design and will make automatic RBF much simpler to add.

The high-level architecture is the following:

```
+---------+
| Channel |---+
+---------+   |             +--------------------+
              | PublishTx   |    TxPublisher     |
              +------------>| - stores txs and   |---+                      +-----------------+
                            |   deadlines        |   | create child actor   |    TxPublish    |
                            | - create child     |   | ask it to publish    | - preconditions |
                            |   actors that fund |   | at a given feerate   | - (funding)     |
                            |   and publish txs  |   +--------------------->| - (signing)     |
                            +--------------------+                          | - publishing    |
                                      ^                                     | - waiting       |
                                      |                                     | - stopping      |
                                      |                                     +-----------------+
                                      |                                              |
                                      +----------------------------------------------+
                                        result:
                                         - transaction confirmed
                                         - transaction rejected
```

There is one `TxPublisher` per channel. The channel tells it to publish transactions (and in the future will give it deadline instructions). The `TxPublisher` creates child actors (`RawTxPublisher` / `ReplaceableTxPublisher`) for each transaction publishing attempts and waits for them to report the result and schedule a retry if necessary.

A `RawTxPublisher` simply publishes a complete transaction without modifying it.
A `ReplaceableTxPublisher` will add inputs to match the desired feerate and publish the resulting transaction.
Internally, they use a `TxTimeLockMonitor` that checks CLTV and CSV delays, and a `MempoolTxMonitor` that regularly checks if its transaction confirms or is rejected from the mempool.

NB: it's worth turning on `debug` logs for the `fr.acinq.eclair.channel.publish` package to troubleshoot potential issues with this new publishing mechanism.


eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/RawTxPublisher.scala
```diff
@@ -0,0 +1,190 @@
...
+    }
+  }
+
+  def checkParentPublished(replyTo: ActorRef[TxPublisher.PublishTxResult], cmd: TxPublisher.PublishRawTx): Behavior[Command] = {
+    cmd.parentTx_opt match {
+      case Some(parentTx) =>
+        context.self ! CheckParentTx
+        Behaviors.receiveMessagePartial {
+          case CheckParentTx =>
+            context.pipeToSelf(bitcoinClient.getTxConfirmations(parentTx.txid)) {
```
pm47, line 100: Why not use `WatchParentTxConfirmed` like in `TxTimeLocksMonitor`? Probably because it wouldn't give you the granularity needed.

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
```diff
@@ -0,0 +1,194 @@
...
+  def publish(replyTo: ActorRef[TxResult], tx: Transaction, input: OutPoint): Behavior[Command] = {
+    context.pipeToSelf(bitcoinClient.publishTransaction(tx)) {
+      case Success(_) => PublishOk
+      case Failure(reason) => PublishFailed(reason)
+    }
+    Behaviors.receiveMessagePartial {
+      case PublishOk =>
+        log.debug("txid={} was successfully published, waiting for confirmation...", tx.txid)
+        waitForConfirmation(replyTo, tx, input)
+      case PublishFailed(reason) if reason.getMessage.contains("rejecting replacement") =>
```
pm47, line 89: Shouldn't we match on the code? It's not really a problem as long as we have non-reg tests

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
```diff
@@ -0,0 +1,194 @@
...
+        } else if (status.spentUnconfirmed) {
+          log.info("could not publish tx: a conflicting mempool transaction is already in the mempool")
+          sendResult(replyTo, TxRejected(TxPublisher.ConflictingTxUnconfirmed))
+        } else {
+          log.info("could not publish tx: one of our wallet inputs is not available")
+          sendResult(replyTo, TxRejected(TxPublisher.WalletInputGone))
+        }
+      case CheckInputFailed(reason) =>
+        log.error("could not check input status", reason)
+        sendResult(replyTo, TxRejected(TxPublisher.TxSkipped(retryNextBlock = true))) // we act as if the input is potentially still spendable
```
pm47, line 113: Should the `InputStatus` part be handled in a separate state?

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+
+  private def computeInputStatus(parentConfirmations: Option[Int], spendableMempoolExcluded: Boolean, spendableMempoolIncluded: Boolean): InputStatus = {
+    parentConfirmations match {
+      case Some(0) => InputStatus(spentConfirmed = false, spentUnconfirmed = !spendableMempoolIncluded)
+      case Some(_) => InputStatus(!spendableMempoolExcluded, !spendableMempoolIncluded)
+      case None => InputStatus(spentConfirmed = false, spentUnconfirmed = false)
+    }
+  }
```
pm47, line 192: That one is tricky. Not to self @pm47 : do a second pass

pm47 (commented):
Part of what makes this PR interesting is that the seemingly simple task of funding and publishing transactions is much -much- more complicated than it appears. That's why I believe this architecture is right even if it looks over-engineered at first glance (especially the `*Monitor` actors).

The implementation of some of the actors is pretty complex (looking at `ReplaceableTxPublisher`), but I don't see a way around that, and we still manage to have a well defined, simple interface with those actors.

I'm going to do a review-by-testing now.

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/RawTxPublisher.scala
t-bast: Yes, and because here we know the parent should appear in the mempool very soon, so it's lighter to check ourselves than go through the watcher (which makes several calls to bitcoin core to fetch more information than we need).

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
t-bast: There are a few different cases why a transaction fails to replace an existing mempool transaction:

- it doesn't pay enough fees to replace the current tx
- it adds a new unconfirmed input
- it forces the eviction of too many children

All of these use the same error code `TX_MEMPOOL_POLICY` in `bitcoind`, but other errors share that same error code (such as `mempool min fee not met`). So unfortunately I didn't find a way to only catch these errors reliably...

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
t-bast: We could, but I think it would make the code more complex (because there are two states where we check input status, so I'm not sure how we'd easily specify to what state we should go on success).

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/TxPublisher.scala
```diff
@@ -0,0 +1,272 @@
...
+  //                                       |                                     | - stopping      |
+  //                                       |                                     +-----------------+
+  //                                       |                                              |
+  //                                       +----------------------------------------------+
+  //                                         result:
+  //                                          - transaction confirmed
+  //                                          - transaction rejected
+
+  // @formatter:off
+  sealed trait Command
```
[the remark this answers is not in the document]
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/1844/commits/2ffafbbf5cf5b5bbc07e54adab8cf9cf52a17668

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/TxPublisher.scala
```diff
@@ -0,0 +1,272 @@
...
+        }
+    }
+  }
+
+  private def stopAttempts(attempts: Seq[PublishAttempt]): Unit = attempts.foreach(stopAttempt)
+
+  private def stopAttempt(attempt: PublishAttempt): Unit = attempt match {
+    case RawAttempt(_, _, actor) => actor ! RawTxPublisher.Stop
+    case ReplaceableAttempt(_, _, _, actor) => actor ! ReplaceableTxPublisher.Stop
+  }
```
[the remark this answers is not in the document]
t-bast: I don't have a strong opinion on this, but I liked defining the `Attempt` trait inside this private class since it's an internal implementation detail. If we move this function in the `object`, we also need to move the `Attempt` trait...not sure if it's good or not.

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/TxPublisher.scala
```diff
@@ -0,0 +1,272 @@
...
+  private case class RawAttempt(id: UUID, cmd: PublishRawTx, actor: ActorRef[RawTxPublisher.Command]) extends PublishAttempt
+  private case class ReplaceableAttempt(id: UUID, cmd: PublishReplaceableTx, feerate: FeeratePerKw, actor: ActorRef[ReplaceableTxPublisher.Command]) extends PublishAttempt
+  // @formatter:on
+
+  private def run(pending: Map[OutPoint, Seq[PublishAttempt]], retryNextBlock: Seq[PublishTx], channelInfo: ChannelInfo): Behavior[Command] = {
+    Behaviors.receiveMessage {
+      case cmd: PublishRawTx =>
+        val attempts = pending.getOrElse(cmd.input, Seq.empty)
+        val alreadyPublished = attempts.exists {
+          case a: RawAttempt if a.cmd.tx.txid == cmd.tx.txid => true
```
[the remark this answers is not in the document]
t-bast: It has to be a `RawAttempt`, I can't apply your first proposal.
The reason is that we need a different `ActorRef[XXX]` in the `replyTo` depending on whether it's a raw publish or a replaceable publish, that's why I had to add these two encapsulating classes.

I applied the second proposal in https://github.com/ACINQ/eclair/pull/1844/commits/42f8350a14b62ebec56fabc1465695624f24a979

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/TxPublisher.scala
```diff
@@ -0,0 +1,272 @@
...
+
+  case class WrappedCurrentBlockCount(currentBlockCount: Long) extends Command
+  case class SetChannelId(remoteNodeId: PublicKey, channelId: ByteVector32) extends Command
+  // @formatter:on
+
+  // @formatter:off
+  case class ChannelInfo(remoteNodeId: PublicKey, channelId_opt: Option[ByteVector32])
+  case class TxPublishInfo(id: UUID, remoteNodeId: PublicKey, channelId_opt: Option[ByteVector32]) {
+    def mdc(): Map[String, String] = Logs.mdc(remoteNodeId_opt = Some(remoteNodeId), channelId_opt = channelId_opt, paymentId_opt = Some(id))
+  }
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1844/commits/6c98a548e9ea7b1f471ab2519e3b678b862671b1

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/TxPublisher.scala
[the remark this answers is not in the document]
pm47: Fair enough

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/RawTxPublisher.scala
```diff
@@ -0,0 +1,190 @@
...
+      case Stop =>
+        timeLocksChecker ! TxTimeLocksMonitor.Stop
+        Behaviors.stopped
+    }
+  }
+
+  def checkParentPublished(replyTo: ActorRef[TxPublisher.PublishTxResult], cmd: TxPublisher.PublishRawTx): Behavior[Command] = {
+    cmd.parentTx_opt match {
+      case Some(parentTx) =>
+        context.self ! CheckParentTx
```
pm47, line 97: Why don't we always check the parent tx and retry after a delay? We could use the inputs of the tx to be published?

t-bast: It would be a bit wasteful and almost never necessary.
With anchor outputs we never need this because there's always a CSV 1.
It's only necessary for non-anchor htlc txs (and only because there's a race condition on publishing), but these should be gone in the not-so-distant future.

pm47: Ok, but still I don't think we need to provide the full parent tx, since we only use the txid to make the check. If we always did the check (agreed, not optimal), we could simplify the API a little bit by removing parentTx_opt entirely. Not a big deal though.

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
pm47: If the spending tx is confirmed, we want `InputStatus(spentConfirmed = true, spentUnconfirmed = false)` right? Then I suppose it should be:

```suggestion
  private def computeInputStatus(parentConfirmations: Option[Int], spendableMempoolExcluded: Boolean, spendableMempoolIncluded: Boolean): InputStatus = {
    parentConfirmations match {
      case Some(0) => InputStatus(spentConfirmed = false, spentUnconfirmed = !spendableMempoolIncluded)
      case Some(_) => InputStatus(!spendableMempoolExcluded, spendableMempoolExcluded && !spendableMempoolIncluded)
      case None => InputStatus(spentConfirmed = false, spentUnconfirmed = false)
    }
  }
```

pm47: Also, I think you could completely remove the handler for `Some(0)`, it is redundant (even with my proposed modification).

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/RawTxPublisher.scala
t-bast: I changed it to only contain the parent `txid` in https://github.com/ACINQ/eclair/pull/1844/commits/eb4ddfcd362ad867feb5a3d1a198c27c69cf198d to make it lighter

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
t-bast: > If the spending tx is confirmed, we want InputStatus(spentConfirmed = true, spentUnconfirmed = false) right? 

You're right, that would make more sense. Done in https://github.com/ACINQ/eclair/pull/1844/commits/2ca2811f6fc503436e8e555f54094f52fc548fd4.

> Also, I think you could completely remove the handler for Some(0), it is redundant (even with my proposed modification).

Unfortunately we can't. In the case `Some(0)`, it means the parent is unconfirmed, so `spendableMempoolExcluded` will always be `false` (that input cannot be spendable since it doesn't even exist when you exclude the mempool). So we'd have `spentConfirmed = true` which is incorrect.

pm47 (approved):

pm47:
🎉

PR #1845, opened 2021-06-22
Schedule backup at regular interval

pm47:
This is a bit less trigger happy than previously, and the implementation
is simpler.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -34,6 +34,7 @@ import fr.acinq.eclair.channel.{Channel, Register}
 import fr.acinq.eclair.crypto.WeakEntropyPool
 import fr.acinq.eclair.crypto.keymanager.{LocalChannelKeyManager, LocalNodeKeyManager}
 import fr.acinq.eclair.db.Databases.FileBackup
+import fr.acinq.eclair.db.FileBackupHandler.FileBackupParams
 import fr.acinq.eclair.db.{Databases, DbEventHandler, FileBackupHandler}
 import fr.acinq.eclair.io.{ClientSpawner, Peer, Server, Switchboard}
 import fr.acinq.eclair.payment.receive.PaymentHandler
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/FileBackupHandler.scala
```diff
@@ -16,80 +16,100 @@
...
-// this notification is sent when we have completed our backup process (our backup file is ready to be used)
-case object BackupCompleted extends BackupEvent
+  def normal(willBackup: Boolean): Behavior[Command] =
+    Behaviors.receiveMessagePartial {
+      case _: WrappedChannelPersisted =>
+        context.log.debug("will perform backup")
+        normal(willBackup = true)
+      case TickBackup => if (willBackup) {
+        context.log.debug("performing backup")
+        doBackup()
```
t-bast, line 83: IIRC we had a bounded mailbox because `doBackup` can be slow, during which time the mailbox is getting full (if there is a lot of channel churn) which can cause issues. Do we want to do something about it? Or should we just direct users with enough load towards postgres which disables this component?

eclair-core/src/main/scala/fr/acinq/eclair/db/FileBackupHandler.scala
```diff
@@ -16,80 +16,100 @@
...
-      .withMailbox("eclair.backup-mailbox")
-      .withDispatcher("eclair.backup-dispatcher")
-}
+  private def doBackup(): Unit = {
+    KamonExt.time(Metrics.FileBackupDuration.withoutTags()) {
+      val tmpFile = new File(backupParams.targetFile.getAbsolutePath.concat(".tmp"))
+      databases.backup(tmpFile)
+
+      // this will throw an exception if it fails, which is possible if the backup file is not on the same filesystem
+      // as the temporary file
```
t-bast, line 96: Do we want to catch that and log an error instead of letting the actor crash?
Now that we moved to akka typed, the actor will not be restarted if it crashes since you didn't provide a `SupervisorStrategy` in `Setup`.

eclair-core/src/main/scala/fr/acinq/eclair/db/FileBackupHandler.scala
pm47: That's a good point, I knew I was missing something...

pm47: Fixed with d2a3998fc.

eclair-core/src/main/scala/fr/acinq/eclair/db/FileBackupHandler.scala
pm47: Fixed with d2a3998fc.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -248,15 +249,15 @@ class Setup(val datadir: File,
...
-              new File(chaindir, "eclair.sqlite.bak"),
-              if (config.hasPath("backup-notify-script")) Some(config.getString("backup-notify-script")) else None),
-            "backuphandler", SupervisorStrategy.Resume))
+          case fileBackup: FileBackup if config.getBoolean("file-backup.enabled") =>
+            val fileBackupParams = FileBackupParams(
+              interval = FiniteDuration(config.getDuration("file-backup.interval").getSeconds, TimeUnit.SECONDS),
+              targetFile = new File(chaindir, config.getString("file-backup.target-file")),
+              script_opt = if (config.hasPath("file-backup.notify-script")) Some(config.getString("file-backup.notify-script")) else None
+            )
+            system.spawn(FileBackupHandler(fileBackup, fileBackupParams), name = "backuphandler")
```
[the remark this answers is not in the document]
pm47: Agreed, done with 83ed67249. I'm not sure what the best place is for the call to `Behaviors.supervise`: in the `apply()`, or when calling `spawn`? Not sure what happens to the event stream subscription if the supervisor is set up below in the hierarchy.

I noticed there is a `Behaviors.restartWithBackoff` which could come in handy for other use cases. Maybe the reconnection task?

t-bast (approved):

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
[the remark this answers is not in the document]
t-bast: > I'm not sure what the best place is for the call to Behaviors.supervise: in the apply(), or when calling spawn?

I like to define that in the parent (so inside `spawn`), I feel like it's usually the parent's responsibility to decide how to supervise children, but I'm not set on this rule at all.

PR #1846, opened 2021-06-23
Implement option-upfront-shutdown-script

sstone:


sstone:
This PR also fixes a bug in how we handle invalid `Shutdown` messages: we would fore-close the channel instead of failing the connection.
Tested with c-lightning v0.9.3 and lnd v0.13.0-beta

sstone:
Rebased on top of #1849

t-bast:
Can you separate the channel codecs changes from the actual `upfront_shutdown_script` feature? IMO we should first focus on just the channel codec changes and migration and integrate it into #1849, which will already be quite a big PR on its own.

Then once the channel codecs are changed on `master`, we can review a separate PR that implements `upfront_shutdown_script` and stores the remote script in the new placeholder you added in `RemoteParams`.

sstone:
Rebased on top of #1849

eclair-core/src/main/resources/reference.conf
```diff
@@ -38,6 +38,7 @@ eclair {
   // see https://github.com/lightningnetwork/lightning-rfc/blob/master/09-features.md
   features {
     option_data_loss_protect = optional
+    option_upfront_shutdown_script = optional
```
t-bast, line 41: I think this should be disabled by default.
We should also add an option to provide the final script in the `open` API.
Finally, we should add a big comment here explaining the behavior:
- If users activate this feature, they will need to either provide a script when opening the channel, or accept that the funds will go back to the bitcoin core wallet that's loaded at the moment of opening (via our default final script)

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -869,84 +872,88 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-        // are there pending signed changes on either side? we need to have received their last revocation!
-        if (d.commitments.hasNoPendingHtlcsOrFeeUpdate) {
-          // there are no pending signed changes, let's go directly to NEGOTIATING
-          if (d.commitments.localParams.isFunder) {
-            // we are funder, need to initiate the negotiation by sending the first closing_signed
-            val (closingTx, closingSigned) = Closing.makeFirstClosingTx(keyManager, d.commitments, localShutdown.scriptPubKey, remoteShutdown.scriptPubKey, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)
-            goto(NEGOTIATING) using DATA_NEGOTIATING(d.commitments, localShutdown, remoteShutdown, List(List(ClosingTxProposed(closingTx, closingSigned))), bestUnpublishedClosingTx_opt = None) storing() sending sendList :+ closingSigned
+      d.commitments.getRemoteShutdownScript(remoteScriptPubKey) match {
+        case Left(e) =>
+          log.warning("they sent an invalid closing script")
```
t-bast, line 894: We should also send a `Warning` message explaining that now that we have this option ;)
Note that in this case, we should delay the `peer ! Peer.Disconnect(remoteNodeId)` to give a chance to our warning to be sent (see https://github.com/ACINQ/eclair/blob/ca51a2d16883d277da41258e527a4ac7a6894bb6/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1709).

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -86,7 +88,9 @@ trait StateTestsHelperMethods extends TestKitBase {
                           relayerA: TestProbe,
                           relayerB: TestProbe,
                           channelUpdateListener: TestProbe,
-                          wallet: EclairWallet) {
+                          wallet: EclairWallet,
+                          alicePeer: TestProbe,
+                          bobPeer: TestProbe) {
```
t-bast, line 93: Nice addition, it will be useful to test that we disconnect appropriately in some cases.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
```diff
@@ -176,6 +176,32 @@ class WaitForAcceptChannelStateSpec extends TestKitBaseClass with FixtureAnyFunS
     awaitCond(alice.stateName == WAIT_FOR_FUNDING_INTERNAL)
   }
 
+  test("recv AcceptChannel (upfront shutdown script)", Tag(StateTestsTags.OptionUpfrontShutdownScript)) { f =>
+    import f._
+    val accept = bob2alice.expectMsgType[AcceptChannel]
+    accept.tlvStream.get[ChannelTlv.UpfrontShutdownScript].contains(ChannelTlv.UpfrontShutdownScript(Bob.channelParams.defaultFinalScriptPubKey))
+    bob2alice.forward(alice, accept)
+    awaitCond(alice.stateName == WAIT_FOR_FUNDING_INTERNAL)
```
t-bast, line 184: We should also verify that the script is correctly stored in alice's `remoteParams`?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
```diff
@@ -176,6 +176,32 @@ class WaitForAcceptChannelStateSpec extends TestKitBaseClass with FixtureAnyFunS
...
+  }
+
+  test("recv AcceptChannel (empty upfront shutdown script)", Tag(StateTestsTags.OptionUpfrontShutdownScript)) { f =>
+    import f._
+    val accept = bob2alice.expectMsgType[AcceptChannel]
+    accept.tlvStream.get[ChannelTlv.UpfrontShutdownScript].contains(ChannelTlv.UpfrontShutdownScript(Bob.channelParams.defaultFinalScriptPubKey))
+    val accept1 = accept.copy(tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScript(ByteVector.empty)))
+    bob2alice.forward(alice, accept1)
+    awaitCond(alice.stateName == WAIT_FOR_FUNDING_INTERNAL)
```
t-bast, line 193: Same, we should check alice's `remoteParams` script field.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -1918,6 +1919,40 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
     awaitCond(alice.stateName == NORMAL)
   }
 
+  test("recv CMD_CLOSE (with a script that does not match our upfront shutdown script)", Tag(StateTestsTags.OptionUpfrontShutdownScript)) { f =>
+    import f._
+    val sender = TestProbe()
+    awaitCond(alice.stateData.asInstanceOf[DATA_NORMAL].localShutdown.isEmpty)
```
t-bast, line 1925: Why do you need that line?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -1918,6 +1919,40 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
+    val shutdownScript = Script.write(Script.pay2wpkh(randomKey().publicKey))
+    alice ! CMD_CLOSE(sender.ref, Some(shutdownScript))
+    sender.expectMsgType[RES_FAILURE[CMD_CLOSE, InvalidFinalScript]]
+  }
+
+  test("recv CMD_CLOSE (with a script that does match our upfront shutdown script)", Tag(StateTestsTags.OptionUpfrontShutdownScript)) { f =>
+    import f._
+    val sender = TestProbe()
+    awaitCond(alice.stateData.asInstanceOf[DATA_NORMAL].localShutdown.isEmpty)
```
t-bast, line 1934: Same, this is probably unnecessary?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -1918,6 +1919,40 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
+    val shutdown = alice2bob.expectMsgType[Shutdown]
+    assert(shutdown.scriptPubKey == alice.stateData.asInstanceOf[DATA_NORMAL].commitments.localParams.defaultFinalScriptPubKey)
+    awaitCond(alice.stateName == NORMAL)
+    awaitCond(alice.stateData.asInstanceOf[DATA_NORMAL].localShutdown.isDefined)
+  }
+
+  test("recv CMD_CLOSE (upfront shutdown script)", Tag(StateTestsTags.OptionUpfrontShutdownScript)) { f =>
+    import f._
+    val sender = TestProbe()
+    awaitCond(alice.stateData.asInstanceOf[DATA_NORMAL].localShutdown.isEmpty)
```
t-bast, line 1947: Same here.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -2002,19 +2037,23 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
   test("recv Shutdown (with invalid final script)") { f =>
     import f._
     bob ! Shutdown(ByteVector32.Zeroes, hex"00112233445566778899")
-    bob2alice.expectMsgType[Error]
-    bob2blockchain.expectMsgType[PublishTx]
-    bob2blockchain.expectMsgType[PublishTx]
-    bob2blockchain.expectMsgType[WatchTxConfirmed]
-    awaitCond(bob.stateName == CLOSING)
+    bob2alice.expectNoMessage(3 seconds)
```
t-bast, line 2040: 3 seconds feels a lot, 500 millis should be more than enough, isn't it? Same for the tests below.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/g/NegotiatingStateSpec.scala
```diff
@@ -139,6 +153,10 @@ class NegotiatingStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike
     testFeeConverge(f)
   }
 
+  test("recv ClosingSigned (theirCloseFee == ourCloseFee) (fee 1, upfront shutdown script)", Tag("fee1"), Tag(StateTestsTags.OptionUpfrontShutdownScript)) { f =>
```
t-bast, line 156: I think the "fee1" tag doesn't exist, it's the default behavior.
```suggestion
  test("recv ClosingSigned (theirCloseFee == ourCloseFee) (fee 1, upfront shutdown script)", Tag(StateTestsTags.OptionUpfrontShutdownScript)) { f =>
```

eclair-core/src/main/resources/reference.conf
sstone: Activating this feature was definitely not a good idea, I'll leave it disabled by default. 
I'm not sure yet we should provide a way for users to specify a custom shutdown script: we rely on bitcoin core to manage our node's wallet, and for example we use wallet points when static_remote_key is enabled.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -339,13 +340,17 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   when(WAIT_FOR_OPEN_CHANNEL)(handleExceptions {
     case Event(open: OpenChannel, d@DATA_WAIT_FOR_OPEN_CHANNEL(INPUT_INIT_FUNDEE(_, localParams, _, remoteInit, channelConfig, channelFeatures))) =>
       log.info("received OpenChannel={}", open)
+      val allowAnySegwit = Features.canUseFeature(localParams.initFeatures, remoteInit.features, Features.ShutdownAnySegwit)
```
t-bast, line 343: You're not using this anywhere, remove?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -869,84 +872,88 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-          // there are no pending signed changes, let's go directly to NEGOTIATING
-          if (d.commitments.localParams.isFunder) {
-            // we are funder, need to initiate the negotiation by sending the first closing_signed
-            val (closingTx, closingSigned) = Closing.makeFirstClosingTx(keyManager, d.commitments, localShutdown.scriptPubKey, remoteShutdown.scriptPubKey, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)
-            goto(NEGOTIATING) using DATA_NEGOTIATING(d.commitments, localShutdown, remoteShutdown, List(List(ClosingTxProposed(closingTx, closingSigned))), bestUnpublishedClosingTx_opt = None) storing() sending sendList :+ closingSigned
+      d.commitments.getRemoteShutdownScript(remoteScriptPubKey) match {
+        case Left(e) =>
+          log.warning("they sent an invalid closing script")
+          peer ! Peer.Disconnect(remoteNodeId)
+          stay()
```
t-bast, line 896: nit: we use `stay` (without parenthesis) everywhere else

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -78,10 +78,24 @@ object Helpers {
...
+  def extractShutdownScript(channelId: ByteVector32, channelFeatures: ChannelFeatures, upfrontShutdownScript_opt: Option[ByteVector]): Either[ChannelException, Option[ByteVector]] =
+    extractShutdownScript(channelId, channelFeatures.hasFeature(Features.OptionUpfrontShutdownScript), channelFeatures.hasFeature(Features.ShutdownAnySegwit), upfrontShutdownScript_opt)
+
+  def extractShutdownScript(channelId: ByteVector32, hasOptionUpfrontShutdownScript: Boolean, allowAnySegwit: Boolean, upfrontShutdownScript_opt: Option[ByteVector]): Either[ChannelException, Option[ByteVector]] = {
+    (hasOptionUpfrontShutdownScript, upfrontShutdownScript_opt) match {
+      case (true, None) => Left(MissingUpfrontShutdownScript(channelId))
+      case (true, Some(script)) if script.isEmpty => Right(None) // but the provided script can be empty
+      case (true, Some(script)) if !Closing.isValidFinalScriptPubkey(script, allowAnySegwit) => Left(InvalidFinalScript(channelId))
+      case (true, Some(script)) => Right(Some(script))
+      case (false, Some(script)) if !script.isEmpty => Left(InvalidFinalScript(channelId))
```
t-bast, line 90: I think this is incorrect, the spec says that even if `option_upfront_shutdown_script` isn't activated, senders "MAY include `upfront_shutdown_script`".

I think that in this case we should run the script through `Closing.isValidFinalScriptPubkey` (and return `Left(InvalidFinalScript(script))` when it's invalid `Right(None)` if it's valid).

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
t-bast: I think this isn't addressed in the latest commits (same for the test below).

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
```diff
@@ -176,6 +176,32 @@ class WaitForAcceptChannelStateSpec extends TestKitBaseClass with FixtureAnyFunS
     awaitCond(alice.stateName == WAIT_FOR_FUNDING_INTERNAL)
   }
 
+  test("recv AcceptChannel (upfront shutdown script)", Tag(StateTestsTags.OptionUpfrontShutdownScript)) { f =>
```
t-bast, line 179: We should also add tests to `WaitForOpenChannelStateSpec.scala`

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -2014,19 +2045,23 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
+    }
   }
 
   test("recv Shutdown (with unsupported native segwit script)") { f =>
     import f._
     bob ! Shutdown(ByteVector32.Zeroes, hex"51050102030405")
-    bob2alice.expectMsgType[Error]
-    bob2blockchain.expectMsgType[PublishTx]
-    awaitCond(bob.stateName == CLOSING)
+    bob2alice.expectNoMessage(3 seconds)
```
t-bast, line 2059: nit:
```suggestion
    bob2alice.expectNoMessage(500 millis)
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
sstone: done  in https://github.com/ACINQ/eclair/pull/1846/commits/f324eae12481bcda0992c902603a01ccde580a35

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
```diff
@@ -176,6 +176,31 @@ class WaitForAcceptChannelStateSpec extends TestKitBaseClass with FixtureAnyFunS
...
+    import f._
+    val accept = bob2alice.expectMsgType[AcceptChannel]
```
t-bast, line 181: Why did you remove the checks on the message tlv? I think they are still useful, because right now if no script is sent (`None`) your test will pass.

I would add:
```suggestion
    val accept = bob2alice.expectMsgType[AcceptChannel]
    assert(accept.upfrontShutdownScript_opt === Some(Bob.channelParams.defaultFinalScriptPubKey))
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForOpenChannelStateSpec.scala
```diff
@@ -216,6 +216,31 @@ class WaitForOpenChannelStateSpec extends TestKitBaseClass with FixtureAnyFunSui
     awaitCond(bob.stateName == WAIT_FOR_FUNDING_CREATED)
   }
 
+  test("recv OpenChannel (upfront shutdown script)", Tag(StateTestsTags.OptionUpfrontShutdownScript)) { f =>
+    import f._
+    val open = alice2bob.expectMsgType[OpenChannel]
```
t-bast, line 221: I would test the contents of the tlv here as well, otherwise if no script is sent your test still passes:
```suggestion
    val open = alice2bob.expectMsgType[OpenChannel]
    assert(open.upfrontShutdownScript_opt === Some(Alice.channelParams.defaultFinalScriptPubKey))
```

t-bast (commented):
LGTM, I would just add a few test assertions and address [this comment](https://github.com/ACINQ/eclair/pull/1846#discussion_r671098496) and we should be good to go

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
sstone: done in https://github.com/ACINQ/eclair/pull/1846/commits/e3174057030eb52428df1f77937f56568db4dfe4

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForOpenChannelStateSpec.scala
sstone: done in https://github.com/ACINQ/eclair/pull/1846/commits/e3174057030eb52428df1f77937f56568db4dfe4

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
sstone: done in https://github.com/ACINQ/eclair/pull/1846/commits/e3174057030eb52428df1f77937f56568db4dfe4

t-bast (approved):
LGTM :rocket:

PR #1849, opened 2021-06-24
Only store remote sigs for the local commit

pm47:
That way, even if the node database or a backup is compromised, the attacker won't be able to force close channels from the outside.

This is best reviewed commit by commit.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -39,8 +39,8 @@ case class RemoteChanges(proposed: List[UpdateMessage], acked: List[UpdateMessag
   def all: List[UpdateMessage] = proposed ++ signed ++ acked
 }
 case class Changes(ourChanges: LocalChanges, theirChanges: RemoteChanges)
-case class HtlcTxAndSigs(txinfo: HtlcTx, localSig: ByteVector64, remoteSig: ByteVector64)
-case class PublishableTxs(commitTx: CommitTx, htlcTxsAndSigs: List[HtlcTxAndSigs])
+case class HtlcTxAndSigs(txinfo: HtlcTx, remoteSig: ByteVector64)
+case class PublishableTxs(commitTx: CommitTx, remoteSig: ByteVector64, htlcTxsAndSigs: List[HtlcTxAndSigs])
```
t-bast: Should we rename the commitTx field to make it very explicit that it's unsigned (and prevent someone changing it in the future)?
```scala
case class PublishableTxs(unsignedCommitTx: CommitTx, remoteSig: ByteVector64, htlcTxsAndSigs: List[HtlcTxAndSigs])
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -144,9 +144,17 @@ case class Commitments(channelVersion: ChannelVersion,
     localCommit.spec.htlcs.collect(incoming).filter(nearlyExpired)
   }
 
-  def addLocalProposal(proposal: UpdateMessage): Commitments = Commitments.addLocalProposal(this, proposal)
-
-  def addRemoteProposal(proposal: UpdateMessage): Commitments = Commitments.addRemoteProposal(this, proposal)
+  /**
+   * Return a fully signed commit tx, that can be published as-is.
+   */
+  def fullySignedLocalCommitTx(keyManager: ChannelKeyManager): CommitTx = {
```
t-bast, line 150: Why not name this `signLocalCommitTx`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -144,9 +144,17 @@ case class Commitments(channelVersion: ChannelVersion,
...
+    val unsignedCommitTx = localCommit.publishableTxs.commitTx
+    val localSig = keyManager.sign(unsignedCommitTx, keyManager.fundingPublicKey(localParams.fundingKeyPath), TxOwner.Local, commitmentFormat)
+    val remoteSig = localCommit.publishableTxs.remoteSig
+    val commitTx = Transactions.addSigs(unsignedCommitTx, keyManager.fundingPublicKey(localParams.fundingKeyPath).publicKey, remoteParams.fundingPubKey, localSig, remoteSig)
+    Transactions.checkSpendable(commitTx).get
```
t-bast, line 155: Are we using this `.get` as some kind of `require`? We should have verified that the signed commit tx was ok when receiving their `commit_sig`, so it's here only to catch developer mistakes, right?

What about:
```scala
// We verify the remote signature when receiving their commit_sig, so this check should always pass.
require(Transactions.checkSpendable(commitTx).isSuccess, "commit signatures are invalid")
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: I did that with 644e3f8, the goal was to minimize the changeset for easier review.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: I could, but then it would make sense to provide the commit tx in the argument (and then we would have to make sure it is consistent with the commitment object itself. I also flip-flopped with moving this to `Helpers`. We're kind of in-between.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: It's actually a left-over from my testing, I think I should remove it.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: :+1: I like `require` approach, it's better than a rename

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: You're right, as a method of the `Commitments` object your naming makes more sense.
Let's keep it like that, and I think it makes sense inside the `Commitments` object (at least for now).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: What about keeping it in a `require` to ensure we catch potential issues?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -39,9 +39,9 @@ case class RemoteChanges(proposed: List[UpdateMessage], acked: List[UpdateMessag
   def all: List[UpdateMessage] = proposed ++ signed ++ acked
 }
 case class Changes(ourChanges: LocalChanges, theirChanges: RemoteChanges)
-case class HtlcTxAndSigs(txinfo: HtlcTx, localSig: ByteVector64, remoteSig: ByteVector64)
-case class PublishableTxs(commitTx: CommitTx, htlcTxsAndSigs: List[HtlcTxAndSigs])
-case class LocalCommit(index: Long, spec: CommitmentSpec, publishableTxs: PublishableTxs)
+case class HtlcTxAndRemoteSig(txinfo: HtlcTx, remoteSig: ByteVector64)
+case class CommitTxAndRemoteSig(commitTx: CommitTx, remoteSig: ByteVector64, htlcTxsAndSigs: List[HtlcTxAndRemoteSig])
```
t-bast, line 43: It's a bit weird that now the name doesn't reflect the fact that it contains htlc txs as well.
What it actually contains are all the txs that require both a local and a remote sig, with the corresponding remote sigs.
What about `CommitmentTxsAndRemoteSigs` (I'm not 100% convinced, we can probably do better)?

Or a somewhat different design would be to keep naming this `PublishableTxs(commitTx: CommitTx, remoteSig: ByteVector64, htlcTxsAndSigs: List[HtlcTxAndRemoteSig])` and move the methods to create the fully signed commit tx and fully signed htlc txs inside this class? Because this does contain publishable txs, we just need to add our local sig (which we should always be able to do).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -598,36 +606,33 @@ object Commitments {
     val channelKeyPath = keyManager.keyPath(commitments.localParams, commitments.channelVersion)
     val localPerCommitmentPoint = keyManager.commitmentPoint(channelKeyPath, commitments.localCommit.index + 1)
     val (localCommitTx, htlcTxs) = makeLocalTxs(keyManager, channelVersion, localCommit.index + 1, localParams, remoteParams, commitInput, localPerCommitmentPoint, spec)
-    val sig = keyManager.sign(localCommitTx, keyManager.fundingPublicKey(commitments.localParams.fundingKeyPath), TxOwner.Local, commitmentFormat)
 
     log.info(s"built local commit number=${localCommit.index + 1} toLocalMsat=${spec.toLocal.toLong} toRemoteMsat=${spec.toRemote.toLong} htlc_in={} htlc_out={} feeratePerKw=${spec.feeratePerKw} txid=${localCommitTx.tx.txid} tx={}", spec.htlcs.collect(incoming).map(_.id).mkString(","), spec.htlcs.collect(outgoing).map(_.id).mkString(","), localCommitTx.tx)
 
     // no need to compute htlc sigs if commit sig doesn't check out
```
t-bast, line 612: You can remove that comment, we're not computing htlc sigs anymore.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -598,36 +606,33 @@ object Commitments {
...
     }
-    val htlcSigs = sortedHtlcTxs.map(keyManager.sign(_, keyManager.htlcPoint(channelKeyPath), localPerCommitmentPoint, TxOwner.Local, commitmentFormat))
+
     val remoteHtlcPubkey = Generators.derivePubKey(remoteParams.htlcBasepoint, localPerCommitmentPoint)
-    // combine the sigs to make signed txes
-    val htlcTxsAndSigs = (sortedHtlcTxs, htlcSigs, commit.htlcSignatures).zipped.toList.collect {
-      case (htlcTx: HtlcTimeoutTx, localSig, remoteSig) =>
-        if (Transactions.checkSpendable(Transactions.addSigs(htlcTx, localSig, remoteSig, commitmentFormat)).isFailure) {
+    val htlcTxsAndSigs = sortedHtlcTxs.zip(commit.htlcSignatures).toList.collect {
+      case (htlcTx: HtlcTimeoutTx, remoteSig) =>
```
t-bast, line 624: We now handle `HtlcTimeoutTx` and `HtlcSuccessTx` the same way, so I don't think we need the collect anymore, we can combine the code for both cases.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -528,7 +528,7 @@ class ClosingStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
     // note that bob doesn't receive the new sig!
     // then we make alice unilaterally close the channel
     val closingState = localClose(alice, alice2blockchain)
-    assert(closingState.commitTx === aliceCommitTx)
+    assert(closingState.commitTx.txid === aliceCommitTx.txid)
```
t-bast, line 531: I think we need to test somewhere (not necessarily here, I'm not sure which test is the best place for that), that the commit txs sent to the `alice2blockchain`/`bob2blockchain` are fully signed. We don't need to repeat this check in every test, matching on the `txid` is ok in most places, but we need to test the codepaths where we're supposed to sign the commit tx (and that applies to htlc txs as well).

t-bast (commented):
I haven't reviewed the migration / codecs, I just focused on the logic changes for now.
I think the migration stuff won't be very hard (tedious, but not hard).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: If we had decided to not keep the unsigned transactions, we could have removed one level and be consistent with the `RemoteCommit` class:

```scala
case class LocalCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, remoteCommitSig: ByteVector64, remoteHtlcSigs: List[ByteVector64])
case class RemoteCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, remotePerCommitmentPoint: PublicKey)
```

That would have been pretty neat. We could do:

```scala
case class LocalCommit(index: Long, spec: CommitmentSpec, commitTxAndRemoteSig: CommitTxAndRemoteSig, htlcTxAndRemoteSig: HtlcTxAndRemoteSig)
case class RemoteCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, remotePerCommitmentPoint: PublicKey)
```

t-bast: > That would have been pretty neat. We could do:
>
> ```scala
> case class LocalCommit(index: Long, spec: CommitmentSpec, commitTxAndRemoteSig: CommitTxAndRemoteSig, > htlcTxAndRemoteSig: HtlcTxAndRemoteSig)
> case class RemoteCommit(index: Long, spec: CommitmentSpec, txid: ByteVector32, remotePerCommitmentPoint: PublicKey)
> ```

That sounds good, yes!

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
pm47: I asked myself the same question. A few lines above, we could totally do:
```scala
val aliceCommitTx = alice.stateData.asInstanceOf[DATA_NORMAL].commitments.fullySignedLocalCommitTx(alice.underlyingActor.nodeParams.channelKeyManager)
```

And then keep the full equality check. This is what we do in the `TxPublisher` test. That said, in the state tests it wouldn't actually check that transaction are properly signed, because it would requires an actual bitcoind. We do that in the integration tests.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: Nice catch

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: Done with 996355480.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: Done with 3b2d10d76.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: Done with 3b2d10d76.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: Done with 2adf25e0e.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
```diff
@@ -143,19 +145,33 @@ private[channel] object ChannelCodecs0 {
...
-      ("txinfo" | txWithInputInfoCodec.downcast[HtlcTx]) ::
-        ("localSig" | variableSizeBytes(uint16, sig64OrDERCodec)) :: // we store as variable length for historical purposes (we used to store as DER encoded)
-        ("remoteSig" | variableSizeBytes(uint16, sig64OrDERCodec))).as[HtlcTxAndSigs].decodeOnly
-
-    val publishableTxsCodec: Codec[PublishableTxs] = (
-      ("commitTx" | (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec)).as[CommitTx]) ::
-        ("htlcTxsAndSigs" | listOfN(uint16, htlcTxAndSigsCodec))).as[PublishableTxs].decodeOnly
-
-    val localCommitCodec: Codec[LocalCommit] = (
+    val htlcTxsAndRemoteSigsCodec: Codec[HtlcTxAndRemoteSig] = (
```
t-bast, line 148: I took a different approach in my migration, which I think is more future-proof:

1. Keep `HtlcTxAndSigs` and the previous `Commitments` object in `ChannelTypes0`
2. Provide a `migrate` function on that type that returns the new `Commitments` class with the right data in it
3. Use it in `ChannelCodecs0`, `ChannelCodecs1` and `ChannelCodecs2` (similar to what was done for the `XxxCommitPublished` [here](https://github.com/ACINQ/eclair/blob/45204e238052ca1691f9f6486eaf1863ae8836b2/eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala#L245))

It makes the changes to the `ChannelCodecsN` files minimal (just add a call to `.migrate()`) and puts the logic in `ChannelTypes0` which makes the `migrate()` function very easy to unit test (instead of having to provide blobs of old serialized data).

I think this is the only way we'll be able to really scale changes to the channel classes with full backwards-compatibility. If for example we decide in version 12 of eclair to change again the format of the `Commitments` case class, we'd simply have to "snapshot" the current `Commitments` in `ChannelTypes3.scala` (and it would be a different snapshot than the one in `ChannelTypes0.scala`). In the `ChannelTypes0.Commitments.migrate()` function we would only need to append a second `.migrate()` call (that would call `ChannelTypes3.Commitments.migrate()`) instead of duplicating some code. WDYT?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
t-bast: Instead of calling `commitments.fullySignedLocalCommitTx` I would instead grab the commit tx we intercept with the `alice2blockchain` actor, and call `CorrectlySpends(fundingTx)` on it, don't you think?

And as mentioned in another comment. we should verify in some tests that what we store is unsigned.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
pm47: As much as I like hacking around with `scodec`, this sounds like a more reasonable approach indeed, will give it a try.

There is another unsolved issue by the way: for this migration we probably want to actively migrate all data, as opposed to doing it next time the channel state is updated. I'm not sure what is the best way to do that.

t-bast: > for this migration we probably want to actively migrate all data

In that case we should probably add a DB migration, that simply reads then writes without changing the SQL schemas (or maybe we can also make schema changes if there are pending ones that we postponed?)

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
pm47: Done with 6f96eae.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -39,9 +39,13 @@ case class RemoteChanges(proposed: List[UpdateMessage], acked: List[UpdateMessag
   def all: List[UpdateMessage] = proposed ++ signed ++ acked
 }
 case class Changes(ourChanges: LocalChanges, theirChanges: RemoteChanges)
-case class HtlcTxAndSigs(txinfo: HtlcTx, localSig: ByteVector64, remoteSig: ByteVector64)
-case class PublishableTxs(commitTx: CommitTx, htlcTxsAndSigs: List[HtlcTxAndSigs])
-case class LocalCommit(index: Long, spec: CommitmentSpec, publishableTxs: PublishableTxs)
+case class HtlcTxAndRemoteSig(htlcTx: HtlcTx, remoteSig: ByteVector64) {
+  require(htlcTx.tx.txIn.forall(_.witness.stack.isEmpty), "htlc tx must not be signed")
```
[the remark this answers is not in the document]
pm47: Partially done with 33101bb, test to be written.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -598,36 +611,25 @@ object Commitments {
...
-    val htlcTxsAndSigs = (sortedHtlcTxs, htlcSigs, commit.htlcSignatures).zipped.toList.collect {
-      case (htlcTx: HtlcTimeoutTx, localSig, remoteSig) =>
-        if (Transactions.checkSpendable(Transactions.addSigs(htlcTx, localSig, remoteSig, commitmentFormat)).isFailure) {
-          return Left(InvalidHtlcSignature(commitments.channelId, htlcTx.tx))
-        }
-        HtlcTxAndSigs(htlcTx, localSig, remoteSig)
-      case (htlcTx: HtlcSuccessTx, localSig, remoteSig) =>
-        // we can't check that htlc-success tx are spendable because we need the payment preimage; thus we only check the remote sig
-        // we verify the signature from their point of view, where it is a remote tx
+    val htlcTxsAndRemoteSigs = sortedHtlcTxs.zip(commit.htlcSignatures).toList.collect {
```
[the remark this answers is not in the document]
pm47: Done with ca66e4b.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
pm47: Done with e4a0ae9. It looks pretty good, but I'm not sure if it would make an additional migration on the same class easier.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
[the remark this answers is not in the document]
pm47: Done with 4f31344.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -0,0 +1,324 @@
...
+      .typecase(0x01, htlcSuccessTxCodec)
+      .typecase(0x02, htlcTimeoutTxCodec)
+
+    val claimHtlcTxCodec: Codec[ClaimHtlcTx] = discriminated[ClaimHtlcTx].by(uint8)
+      .typecase(0x01, claimHtlcSuccessTxCodec)
+      .typecase(0x02, claimHtlcTimeoutTxCodec)
+
+    val htlcTxsAndRemoteSigsCodec: Codec[HtlcTxAndRemoteSig] = (
+      ("txinfo" | htlcTxCodec) ::
+        ("remoteSig" | lengthDelimited(bytes64))).as[HtlcTxAndRemoteSig]
```
t-bast, line 155: nit: isn't it an opportunity to remove the `lengthDelimited` around `bytes64` now that we don't have any ambiguity between der and compact sigs?

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -0,0 +1,324 @@
...
+
+    val commitTxAndRemoteSigCodec: Codec[CommitTxAndRemoteSig] = (
+      ("commitTx" | commitTxCodec) ::
+        ("remoteSig" | lengthDelimited(bytes64))).as[CommitTxAndRemoteSig]
```
t-bast, line 159: nit: same here, we could remove the `lengthDelimited`

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -281,14 +281,17 @@ trait StateTestsHelperMethods extends TestKitBase {
 
   def localClose(s: TestFSMRef[State, Data, Channel], s2blockchain: TestProbe): LocalCommitPublished = {
     // an error occurs and s publishes its commit tx
-    val commitTx = s.stateData.asInstanceOf[DATA_NORMAL].commitments.localCommit.publishableTxs.commitTx.tx
+    val commitTx = s.stateData.asInstanceOf[DATA_NORMAL].commitments.localCommit.commitTxAndRemoteSig.commitTx.tx
```
t-bast, line 284: I think this function is the right place to verify that our txs are stored unsigned. We could add here:
```scala
commitTx.txIn.foreach(txIn => assert(txIn.witness.isNull))
```

And below, the same thing for HTLC txs?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -281,14 +281,17 @@ trait StateTestsHelperMethods extends TestKitBase {
...
     s ! Error(ByteVector32.Zeroes, "oops")
     awaitCond(s.stateName == CLOSING)
     val closingState = s.stateData.asInstanceOf[DATA_CLOSING]
     assert(closingState.localCommitPublished.isDefined)
     val localCommitPublished = closingState.localCommitPublished.get
 
-    assert(s2blockchain.expectMsgType[TxPublisher.PublishRawTx].tx == commitTx)
+    val publishedLocalCommitTx = s2blockchain.expectMsgType[TxPublisher.PublishRawTx].tx
+    assert(publishedLocalCommitTx.txid == commitTx.txid)
+    val commitInput = s.stateData.asInstanceOf[DATA_CLOSING].commitments.commitInput
```
t-bast, line 293: nit:
```suggestion
    val commitInput = closingState.commitments.commitInput
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -1370,13 +1370,18 @@ class ClosingStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
+      case HtlcTxAndRemoteSig(txInfo: HtlcSuccessTx, remoteSig) if txInfo.htlcId == fulfilledHtlc.id =>
         assert(fulfilledHtlc.paymentHash === txInfo.paymentHash)
+        val localSig = bobKeyManager.sign(txInfo, bobKeyManager.htlcPoint(bobChannelKeyPath), bobPerCommitmentPoint, TxOwner.Local, commitmentFormat)
         Transactions.addSigs(txInfo, localSig, remoteSig, preimage, commitmentFormat)
     }.get
-    val bobHtlcTimeoutTx = bobRevokedTxs.htlcTxsAndSigs.collectFirst {
-      case HtlcTxAndSigs(txInfo: HtlcTimeoutTx, localSig, remoteSig) if txInfo.htlcId == failedHtlc.id =>
+    val bobHtlcTimeoutTx = bobRevokedCommit.htlcTxsAndRemoteSigs.collectFirst {
+      case HtlcTxAndRemoteSig(txInfo: HtlcTimeoutTx, remoteSig) if txInfo.htlcId == failedHtlc.id =>
+        val localSig = bobKeyManager.sign(txInfo, bobKeyManager.htlcPoint(bobChannelKeyPath), bobPerCommitmentPoint, TxOwner.Local, commitmentFormat)
```
t-bast, line 1384: You don't need all this logic, these transactions don't need to be signed with valid signatures, they only need to contain a script witness with the right format (and the preimage).

You can remove all this and change the code to simply do:
```scala
// HTLC-success
Transactions.addSigs(txInfo, ByteVector64.Zeroes, ByteVector64.Zeroes, preimage, commitmentFormat)
// HTLC-timeout
Transactions.addSigs(txInfo, ByteVector64.Zeroes, ByteVector64.Zeroes, commitmentFormat)
```

Which will help us focus on the important parts of the test.
It's my bad, I wrote that code and could have made it clear that we don't need the actual signatures, just something that looks like a signature.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -1475,12 +1481,17 @@ class ClosingStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
+    val bobPerCommitmentPoint = bobKeyManager.commitmentPoint(bobChannelKeyPath, bobRevokedCommit.index.toInt)
+    val bobHtlcTxs = bobRevokedCommit.htlcTxsAndRemoteSigs.collect {
+      case HtlcTxAndRemoteSig(txInfo: HtlcSuccessTx, remoteSig) =>
         val preimage = revokedCloseFixture.htlcsAlice.collectFirst { case (add, preimage) if add.id == txInfo.htlcId => preimage }.get
         assert(Crypto.sha256(preimage) === txInfo.paymentHash)
+        val localSig = bobKeyManager.sign(txInfo, bobKeyManager.htlcPoint(bobChannelKeyPath), bobPerCommitmentPoint, TxOwner.Local, commitmentFormat)
         Transactions.addSigs(txInfo, localSig, remoteSig, preimage, AnchorOutputsCommitmentFormat)
-      case HtlcTxAndSigs(txInfo: HtlcTimeoutTx, localSig, remoteSig) =>
+      case HtlcTxAndRemoteSig(txInfo: HtlcTimeoutTx, remoteSig) =>
+        val localSig = bobKeyManager.sign(txInfo, bobKeyManager.htlcPoint(bobChannelKeyPath), bobPerCommitmentPoint, TxOwner.Local, commitmentFormat)
```
t-bast, line 1494: Same here, you don't need to actually sign, you can simply put `ByteVector64.Zeroes` as signatures.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
pm47: I'm not completely sure because we have been bitten when we tried to be too smart. This seems harmless, but I wonder why we didn't do it during the 0->1 transition.

That comment is odd:
```
// we store as variable length for historical purposes (we used to store as DER encoded)
```

t-bast: It's true that this is quite harmless since it's only 2 bytes. And thinking about this more, if later in the future we have potentially different signature sizes (schnorr?) we'd be happy that we length-prefixed it. So let's keep it like that for now.

pm47: To be honest, if we introduce a new type of signature we will probably prefix them by a tag. I'm tempted to go ahead and remove that length prefix

t-bast: That's true, we would likely need a change anyway when we introduce a different signature format.
It's your call then, we can remove that!

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
pm47: Done in 2e074f710 (squashed).

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
pm47: Done in 2e074f710 (squashed).

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
pm47: Done in 2e074f710 (squashed).

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
pm47: Done in 2e074f710 (squashed).

t-bast (commented):
ACK https://github.com/ACINQ/eclair/pull/1849/commits/2e074f710fb02af9105709fcb63cfdcff0d895f5

I'll rebase #1848 on top of this branch.

t-bast (dismissed):
LGTM :rocket:

pm47 (commented):
Rebased on #1860.

eclair-core/src/test/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecsSpec.scala
```diff
@@ -117,21 +116,23 @@ class ChannelCodecsSpec extends AnyFunSuite {
...
-      assert(newbin.startsWith(hex"020002"))
+      assert(newbin.startsWith(hex"030002"))
       // make sure that round-trip yields the same data
       val newnormal = stateDataCodec.decode(newbin.bits).require.value
       assert(newnormal === oldnormal)
     }
   }
 
-  test("backward compatibility DATA_NORMAL_COMPAT_03_Codec (integrity)") {
+  test("backward compatibility older codecs (integrity)") {
```
pm47, line 126: @t-bast I started generalizing this test.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -476,61 +485,11 @@ final case class RemoteParams(nodeId: PublicKey,
                               paymentBasepoint: PublicKey,
                               delayedPaymentBasepoint: PublicKey,
                               htlcBasepoint: PublicKey,
-                              features: Features)
+                              features: Features,
```
pm47, line 488: For clarity (same in `LocalParams`)

```suggestion
                              initFeatures: Features,
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
```diff
@@ -0,0 +1,66 @@
...
+ * Created by t-bast on 24/06/2021.
+ */
+
+/**
+ * Subset of Bolt 9 features used to configure a channel and applicable over the lifetime of that channel.
+ * Even if one of these features is later disabled at the connection level, it will still apply to the channel until the
+ * channel is upgraded or closed.
+ */
+case class ChannelFeatures(features: Features) {
+
```
pm47, line 33: Should we add a `require` that all features are mandatory?

t-bast: We can, but that's not enforced when you `copy`, so it would only be a partial protection

pm47: How about changing the signature to:

```scala
case class ChannelFeatures(activated: Set[Feature], unknown: Set[UnknownFeature])
```

It removes all the ambiguity and makes the choice between optional/mandatory something to only deal with at the codec level.

eclair-core/src/test/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecsSpec.scala
```diff
@@ -148,6 +149,9 @@ class ChannelCodecsSpec extends AnyFunSuite {
       val newjson = Serialization.write(newnormal)(JsonSerializers.formats)
       //.replace(""","unknownFields":""""", "") // if we update the json serializers, we may need to filter out some fields
 
+      println(Serialization.writePretty(newnormal)(JsonSerializers.formats))
+      println(Serialization.write(newnormal)(JsonSerializers.formats))
```
t-bast, line 153: nit: remove

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -109,35 +110,44 @@ trait StateTestsHelperMethods extends TestKitBase {
...
     val channelFlags = if (tags.contains(StateTestsTags.ChannelsPublic)) ChannelFlags.AnnounceChannel else ChannelFlags.Empty
-    val aliceParams = setChannelFeatures(Alice.channelParams, tags)
-      .modify(_.walletStaticPaymentBasepoint).setToIf(channelVersion.paysDirectlyToWallet)(Some(Helpers.getWalletPaymentBasepoint(wallet)))
-      .modify(_.maxHtlcValueInFlightMsat).setToIf(tags.contains(StateTestsTags.NoMaxHtlcValueInFlight))(UInt64.MaxValue)
-      .modify(_.maxHtlcValueInFlightMsat).setToIf(tags.contains(StateTestsTags.AliceLowMaxHtlcValueInFlight))(UInt64(150000000))
-    val bobParams = setChannelFeatures(Bob.channelParams, tags)
-      .modify(_.walletStaticPaymentBasepoint).setToIf(channelVersion.paysDirectlyToWallet)(Some(Helpers.getWalletPaymentBasepoint(wallet)))
-      .modify(_.maxHtlcValueInFlightMsat).setToIf(tags.contains(StateTestsTags.NoMaxHtlcValueInFlight))(UInt64.MaxValue)
-    val initialFeeratePerKw = if (tags.contains(StateTestsTags.AnchorOutputs)) TestConstants.anchorOutputsFeeratePerKw else TestConstants.feeratePerKw
+    val initialFeeratePerKw = TestConstants.feeratePerKw
```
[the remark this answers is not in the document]
pm47: they didn't run last commit is wip and ci is disabled ^^

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
```diff
@@ -109,35 +110,44 @@ trait StateTestsHelperMethods extends TestKitBase {
     SetupFixture(alice, bob, alice2bob, bob2alice, alice2blockchain, bob2blockchain, router, relayerA, relayerB, channelUpdateListener, wallet)
   }
 
-  def setChannelFeatures(defaultChannelParams: LocalParams, tags: Set[String]): LocalParams = {
-    import com.softwaremill.quicklens._
+  def computeFeatures(setup: SetupFixture, tags: Set[String]): (LocalParams, ChannelFeatures, LocalParams, ChannelFeatures) = {
+    import setup._
+
+    val aliceInitFeatures = Alice.channelParams.features
```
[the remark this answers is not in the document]
pm47: They point to the same value: https://github.com/ACINQ/eclair/blob/only-store-remote-sigs/eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala#L171-L173.

But I agre that this makes more sense.

eclair-core/src/test/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecsSpec.scala
pm47: Done in 51b6bd7.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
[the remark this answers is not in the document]
pm47: Done in 51b6bd7.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/StateTestsHelperMethods.scala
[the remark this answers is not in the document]
pm47: Done in 51b6bd7.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
pm47: Done in 51b6bd7.

eclair-core/src/test/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecsSpec.scala
pm47: Done in 51b6bd7.

t-bast (approved):
LGTM, there's only this comment that hasn't been addressed: https://github.com/ACINQ/eclair/pull/1849#discussion_r667711111

eclair-core/src/test/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecsSpec.scala
pm47: Done in d74d93cad (for real this time).

t-bast (approved):

t-bast (dismissed):

t-bast (approved):

t-bast (approved):

PR #1852, opened 2021-06-29
Separate internal channel config from features

t-bast:
Our current ChannelVersion field mixes two unrelated concepts: channel features (as defined in Bolt 9) and channel internals (such as custom key derivation). It is more future-proof to separate these two unrelated concepts and will make it easier to implement channel types (see https://github.com/lightningnetwork/lightning-rfc/pull/880).

The second commit adds an optional shutdown script to the channel's remote params, which will let us support `upfront_shutdown_script` later.

NB: this is PR to the `only-store-remote-sigs` branch, to ensure we squash all channel codec changes into a single commit on `master`.


sstone:
Did you consider modifying `ChannelVersion` to include both BOLT-specific and eclair-specific fields (i.e `ChannelType` and `ChannelConfigOption`) ? It seems that they are related (as they have to be consistent and will be created together) and we often need both fields at the same time ?

t-bast:
> Did you consider modifying ChannelVersion to include both BOLT-specific and eclair-specific fields (i.e ChannelType and ChannelConfigOption) ? It seems that they are related (as they have to be consistent and will be created together) and we often need both fields at the same time ?

Yes I did, but I then decided to separate them. In most cases they are really unrelated and you only need one of them.

The only benefit to keep them encapsulated in a `ChannelVersion` would be to make less changes, but I believe we have to do these changes at some point, so I'd rather do them now (these changes are quite trivial). And it's really confusing that this was named `ChannelVersion`, and will be even more confusing once channel types are an official term in the spec.

sstone:
Ok. One of my concerns was that to implement custom extensions we would need both fields, and that it may even be true for parameter validation, but overall this PR LGTM.

t-bast:
IMO most custom extensions behave like channel features, so they should use only the channel type part, not the channel internal config.

sstone (approved):
I've rebased https://github.com/ACINQ/eclair/pull/1846 on top of this (see https://github.com/ACINQ/eclair/tree/implement-upfront-shutdown-script-channel-version-migration) and everything LGTM.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
```diff
@@ -0,0 +1,77 @@
...
+  def standard: ChannelConfigOptions = ChannelConfigOptions(activated = Set(FundingPubKeyBasedChannelKeyPath))
+
+  def apply(options: ChannelConfigOption*): ChannelConfigOptions = ChannelConfigOptions(Set.from(options))
+
+  def apply(bytes: ByteVector): ChannelConfigOptions = {
+    val activated: Set[ChannelConfigOption] = bytes.bits.toIndexedSeq.reverse.zipWithIndex.collect {
+      case (true, 0) => FundingPubKeyBasedChannelKeyPath
+    }.toSet
+    ChannelConfigOptions(activated)
+  }
```
pm47, line 64:
```suggestion
  def apply(bytes: ByteVector): ChannelConfigOptions = {
    val activated: Set[ChannelConfigOption] = bytes.bits.toIndexedSeq.reverse.zipWithIndex.collect {
      resolveOption(FundingPubKeyBasedChannelKeyPath)
    }.toSet
    ChannelConfigOptions(activated)
  }

  def resolveOption(option: ChannelConfigOption): PartialFunction[(Boolean, Int), ChannelConfigOption] = {
    case (true, bit) if bit == option.supportBit => option
  }
```
Then we could do:
```scala
  def apply(bytes: ByteVector): ChannelConfigOptions = {
    val activated: Set[ChannelConfigOption] = bytes.bits.toIndexedSeq.reverse.zipWithIndex.collect {
      resolveOption(Feature1) orElse
      resolveOption(Feature2) orElse
      resolveOption(Feature3)
    }.toSet
    ChannelConfigOptions(activated)
  }
```

image

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelType.scala
```diff
@@ -0,0 +1,69 @@
...
+    features.hasFeature(Features.StaticRemoteKey) && !features.hasFeature(Features.AnchorOutputs)
+  }
+
+}
+
+object ChannelTypes {
+
+  val standard = ChannelType(Features.empty)
+  val staticRemoteKey = ChannelType(Features(StaticRemoteKey -> Optional))
+  val anchorOutputs = ChannelType(Features(StaticRemoteKey -> Optional, AnchorOutputs -> Optional))
```
pm47, line 54: Nit to remove those pesky intellij warnings:
```suggestion
  val standard: ChannelType = ChannelType(Features.empty)
  val staticRemoteKey: ChannelType = ChannelType(Features(StaticRemoteKey -> Optional))
  val anchorOutputs: ChannelType = ChannelType(Features(StaticRemoteKey -> Optional, AnchorOutputs -> Optional))
```

eclair-node/src/main/scala/fr/acinq/eclair/api/serde/JsonSerializers.scala
```diff
@@ -135,10 +135,10 @@ class PrivateKeySerializer extends CustomSerializer[PrivateKey](_ => ( {
   case _: PrivateKey => JString("XXX")
 }))
 
-class ChannelVersionSerializer extends CustomSerializer[ChannelVersion](_ => ( {
+class ChannelConfigSerializer extends CustomSerializer[ChannelConfigOptions](_ => ( {
   null
 }, {
-  case x: ChannelVersion => JString(x.bits.toBin)
+  case x: ChannelConfigOptions => JString(x.bytes.toBin)
```
pm47, line 141: A resolution to feature names (like we do in `FeaturesSerializer`) would be nice. Something like:

```
{
  "channelConfig": [
    "FundingPubKeyBasedChannelKeyPath"
  ],
  "channelType": "AnchorOutputs"
}
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelType.scala
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * True if our main output in the remote commitment is directly sent (without any delay) to one of our wallet addresses.
+   */
+  def paysDirectlyToWallet: Boolean = {
+    features.hasFeature(Features.StaticRemoteKey) && !features.hasFeature(Features.AnchorOutputs)
+  }
+
+}
+
+object ChannelTypes {
```
pm47, line 50:
```suggestion
object ChannelType {
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelType.scala
```diff
@@ -0,0 +1,69 @@
...
+import fr.acinq.eclair.transactions.Transactions.{AnchorOutputsCommitmentFormat, CommitmentFormat, DefaultCommitmentFormat}
+
+/**
+ * Created by t-bast on 24/06/2021.
+ */
+
+/**
+ * A channel type is a specific combination of Bolt 9 features, defined in the RFC (Bolt 2).
+ */
+case class ChannelType(features: Features) {
```
pm47, line 31: How about something with more types?

```scala
/**
 * A channel type is a specific combination of Bolt 9 features, defined in the RFC (Bolt 2).
 */
sealed trait ChannelType {
  def features: Features

  def commitmentFormat: CommitmentFormat = {
    if (features.hasFeature(AnchorOutputs)) {
      AnchorOutputsCommitmentFormat
    } else {
      DefaultCommitmentFormat
    }
  }

  /**
   * True if our main output in the remote commitment is directly sent (without any delay) to one of our wallet addresses.
   */
  def paysDirectlyToWallet: Boolean = {
    features.hasFeature(Features.StaticRemoteKey) && !features.hasFeature(Features.AnchorOutputs)
  }
}

object ChannelTypes {

  // @formatter:off
  case object Standard extends ChannelType { override def features: Features = Features.empty }
  case object StaticRemoteKey extends ChannelType { override def features: Features = Features(Features.StaticRemoteKey -> Optional) }
  case object AnchorOutputs extends ChannelType { override def features: Features = Features(Features.StaticRemoteKey -> Optional, Features.AnchorOutputs -> Optional) }
  // @formatter:on

  /**
   * Pick the channel type that should be applied based on features alone (in case our peer doesn't support explicit channel type negotiation).
   */
  def pickChannelType(localFeatures: Features, remoteFeatures: Features): ChannelType = {
    if (Features.canUseFeature(localFeatures, remoteFeatures, Features.AnchorOutputs)) {
      AnchorOutputs
    } else if (Features.canUseFeature(localFeatures, remoteFeatures, Features.StaticRemoteKey)) {
      StaticRemoteKey
    } else {
      Standard
    }
  }

}
```

pm47 (commented):
The clash between `ChannelType` and `ChannelTypes` is unfortunate. I would rename (maybe in a separate PR):
- `State` -> `ChannelState`
- `Data` -> `ChannelData`
- `ChannelTypes.scala` -> `ChannelData.scala`

eclair-node/src/main/scala/fr/acinq/eclair/api/serde/JsonSerializers.scala
t-bast: I completely agree for the `channelConfig`, I'll add names.
For the `channelType`, it already displays the list of human-readable features (`"option_anchor_outputs": "optional") so we should be good.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelType.scala
t-bast: I wanted to be consistent with what we do for features, we have a `Feature` case class and a `Features` object which is a kind of enum that contains `Feature` instances, I think it makes sense?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelType.scala
t-bast: Good idea, I like it!

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
t-bast: Well, I'm oscillating between "this is really over-engineered" and "I feel mysteriously attracted to it" :laughing:

t-bast: What would be really great would be something that fails at compile time if we haven't handled all the possible `ChannelConfigOption`s (to make sure we don't forget to add a line here when we add a new option) but I'm not sure it's possible since here we're doing the reverse mapping (int to config options)...

t-bast:
> The clash between ChannelType and ChannelTypes is unfortunate. I would rename (maybe in a separate PR)

I agree, it's probably better to do it in a separate PR that only does that and no logic change.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelType.scala
sstone: It woks very well when features are exclusive (either A or B or C) but probably not that well with features that can be activate with any other combination of features (like `option_upfront_shutdown_script` or `peer_backup`) ?

t-bast: True, that's why I didn't do it at first, to keep the flexibility of orthogonal features without having to list the explosion of different channel types explicitly (as this comment shows https://github.com/lightningnetwork/lightning-rfc/pull/880#discussion_r655962589). Having a `sealed trait` is nice when we're pattern matching on the channel type, but I think it's not really necessary here, so I'll keep it like it was for now.

Note that I don't think `option_upfront_shutdown_script` nor `peer_backup` should be in the channel features, they aren't features that affect the channel's behavior (and we don't need them here to implement their behavior).

eclair-node/src/main/scala/fr/acinq/eclair/api/serde/JsonSerializers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1852/commits/aba4fa59782c83993358a87899344f1f7b3c324b

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelType.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1852/commits/9a199645acc87b7e632df6476851c200e717a1cc

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
```diff
@@ -0,0 +1,77 @@
...
+    val indices = activated.map(_.supportBit)
+    if (indices.isEmpty) {
+      ByteVector.empty
+    } else {
+      // NB: when converting from BitVector to ByteVector, scodec pads right instead of left, so we make sure we pad to bytes *before* setting bits.
+      var buffer = BitVector.fill(indices.max + 1)(high = false).bytes.bits
+      indices.foreach(i => buffer = buffer.set(i))
+      buffer.reverse.bytes
+    }
+  }
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1852/commits/c69f55e889b214bcceb514635a71c88f2920f61c

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
pm47: Use a `map` instead of a `collect`?

t-bast: A `map` would fail at runtime, but wouldn't help at compile time...at least with the current way I'm checking it, but I'm thinking that I should "reverse" the patter matching, I'll try something.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelType.scala
pm47: Those are two different topics though:
1) types vs values
2) exclusive vs additive

More precisely on 2): do we define a limited number of types that are a combination of certain features, or should we define all types that are a combination of all supported features?

If the answer to 2) is _no_, then I believe types are a better implementation than values. The codec would be a mapping between the whole feature bit and an `Unsupported` type would be used as a catch all.

Now, if the answer to 2) is _yes_, then neither types nor values are a good implementation, because we would have to define a lot of them. But I believe this is not the spirit of the spec, otherwise we would match two bitvectors at channel opening, instead of two _arrays_ of bitvectors. In this case, the codec would operate at the bit level.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
pm47: Gotcha, this is because the space for the `supportBit` is the whole range of integers right?

t-bast: Yes exactly, I'm struggling with reflection to at least have a unit test that checks we're exhaustive. Apparently Scala 3 makes this better with macros, but for now there's no good solution...

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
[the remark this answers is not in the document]
pm47: Sorry if this wasn't clear: I meant only the codec part. It should be part of `ChannelCodecs` imo. The actual `ChannelConfigOptions` should stay in the `channel` package. Does it make sense?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
pm47: Would this be the missing bit (pun intended):

```scala
trait ChannelConfigOption {
  // @formatter:off
  def supportBit: ChannelOptionBit
  def name: String
  // @formatter:on
}

sealed trait ChannelOptionBit { def bit: Int }
case object FundingPubKeyBasedChannelKeyPathBit extends ChannelOptionBit { override def bit = 0 }
```

pm47: This would still require narrowing the space somewhere around `bytes.bits.toIndexedSeq.reverse.zipWithIndex` though.

t-bast: I added a reflection test in https://github.com/ACINQ/eclair/pull/1852/commits/297d81aa5a2472145f7c9fad5cbcac76e7c7f2fc to make up for this (and I fixed the similar test we had in `FeatureSpec` which was completely broken :see_no_evil: ).

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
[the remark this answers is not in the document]
t-bast: I did misunderstand it :)
I'll fix that.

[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1852/commits/974194d8b25328d522d727284992b5e2a8e9fa35, I'll rebase/squash to clean-up this first wave of reviews.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelType.scala
t-bast: I remove `ChannelType` from this PR entirely, we now simply store the channel bolt 9 features (from which we can later extract the channel type, once spec-ed).

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
```diff
@@ -225,21 +225,20 @@ private[channel] object ChannelCodecs0 {
...
+            ("localChanges" | localChangesCodec) ::
+            ("remoteChanges" | remoteChangesCodec) ::
+            ("localNextHtlcId" | uint64overflow) ::
+            ("remoteNextHtlcId" | uint64overflow) ::
+            ("originChannels" | originsMapCodec) ::
+            ("remoteNextCommitInfo" | either(bool, waitingForRevocationCodec, publicKey)) ::
+            ("commitInput" | inputInfoCodec) ::
+            ("remotePerCommitmentSecrets" | ShaChain.shaChainCodec) ::
+            ("channelId" | bytes32))
+      }).as[ChannelTypes0.Commitments].decodeOnly.map[Commitments](_.migrate()).decodeOnly
```
pm47, line 241: The outermost parentheses can be removed (same in other file).

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -42,9 +43,26 @@ private[channel] object ChannelCodecs3 {
...
+      val indices = cfg.activated.map(_.supportBit)
+      if (indices.isEmpty) {
+        ByteVector.empty
+      } else {
+        // NB: when converting from BitVector to ByteVector, scodec pads right instead of left, so we make sure we pad to bytes *before* setting bits.
+        var buffer = BitVector.fill(indices.max + 1)(high = false).bytes.bits
+        indices.foreach(i => buffer = buffer.set(i))
+        buffer.reverse.bytes
+      }
+    })
```
pm47, line 61: This codec is a bit 🤢, but unfortunately it seems `scodec` doesn't work well with bitfields. 

Since we use bitfields at several places, it's probably worth it to he define a generic byte-aligned, right-to-left, bitfield codec? I'll give it a try.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Created by t-bast on 24/06/2021.
+ */
+
+/**
+ * Internal configuration option impacting the channel's structure or behavior.
+ * This must be set when creating the channel and cannot be changed afterwards.
+ */
+trait ChannelConfigOption {
```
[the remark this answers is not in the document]
t-bast: Why? It's really a configuration option...it's not a complete configuration so `ChannelConfig` feels weird IMO, and `ChannelOption` is on the contrary a bit too vague, isn't it?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
```diff
@@ -0,0 +1,58 @@
...
+
+case class ChannelConfigOptions(activated: Set[ChannelConfigOption]) {
+
+  def hasOption(option: ChannelConfigOption): Boolean = activated.contains(option)
+
+}
+
+object ChannelConfigOptions {
+
+  def standard: ChannelConfigOptions = ChannelConfigOptions(activated = Set(FundingPubKeyBasedChannelKeyPath))
```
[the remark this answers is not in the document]
t-bast: No reason, this should be a `val` indeed.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
```diff
@@ -154,10 +154,10 @@ private[channel] object ChannelCodecs0 {
       ("commitTx" | (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec)).as[CommitTx]) ::
         ("htlcTxsAndSigs" | listOfN(uint16, htlcTxAndSigsCodec))).as[PublishableTxs].decodeOnly
 
-    def localCommitCodec(remoteFundingPubKey: PublicKey): Codec[LocalCommit] = (
+    val localCommitCodec: Codec[ChannelTypes0.LocalCommit] = (
```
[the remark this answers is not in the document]
t-bast: I don't think you can without pulling the rest of the changes. I think it's ok, we should merge this inside `only-store-remote-sigs` and then review just the complete diff with `master`, don't you think?

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
t-bast: Agreed, it's useful so that we don't unintentionally go back to storing something that isn't byte-aligned!

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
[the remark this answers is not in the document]
pm47: It's just that everywhere we are using: `channelConfig: ChannelConfigOptions`

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
pm47: I noticed that when comparing both PRs to master

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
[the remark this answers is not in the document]
t-bast: But it's different, it's because it's a `ChannelConfigOptions`, which is a collection of `ChannelConfigOption`.
I agree this one could be renamed `ChannelConfigOptions` -> `ChannelConfig`, but not the trait on individual options.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1852/commits/4e0c81413694e47662e5a21ed4d8a53d19d5e983

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1852/commits/4e0c81413694e47662e5a21ed4d8a53d19d5e983

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelConfigOptions.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1852/commits/4e0c81413694e47662e5a21ed4d8a53d19d5e983

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -42,9 +43,26 @@ private[channel] object ChannelCodecs3 {
...
+    }, cfg => {
+      val indices = cfg.options.map(_.supportBit)
+      if (indices.isEmpty) {
+        ByteVector.empty
+      } else {
+        // NB: when converting from BitVector to ByteVector, scodec pads right instead of left, so we make sure we pad to bytes *before* setting bits.
+        var buffer = BitVector.fill(indices.max + 1)(high = false).bytes.bits
+        indices.foreach(i => buffer = buffer.set(i))
+        buffer.reverse.bytes
+      }
```
pm47, line 60:
```suggestion
      val indices = cfg.activated.map(_.supportBit)
      val maxLength = indices.maxOption.getOrElse(-1) + 1
      var buffer = BitVector.low(maxLength)
      indices.foreach(i => buffer = buffer.set(i))
      buffer.bytes.bits.reverse.bytes
      }
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelTypes.scala
```diff
@@ -476,61 +485,11 @@ final case class RemoteParams(nodeId: PublicKey,
                               paymentBasepoint: PublicKey,
                               delayedPaymentBasepoint: PublicKey,
                               htlcBasepoint: PublicKey,
-                              features: Features)
+                              features: Features,
+                              shutdownScript: Option[ByteVector])
```
pm47, line 489: The naming is inconsistent `LocalParams.defaultFinalScriptPubKey` and `RemoteParams.shutdownScript`

Either:
- `LocalParams.defaultFinalScriptPubKey` and `RemoteParams.upfrontFinalScriptPubKey`
or
- `LocalParams.defaultShutdownScript` and `RemoteParams.upfrontShutdownScript`

I would keep the `upfront` in any case.

pm47: Let's revisit the naming afterwards.

pm47 (approved):

PR #1853, opened 2021-06-30
Trampoline payments use per-channel fee and cltv

thomash-acinq:
Make sure trampoline users pay at least as much as the other users.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
@@ -120,12 +120,13 @@ object NodeRelay {
 
   /** Compute route params that honor our fee and cltv requirements. */
   def computeRouteParams(nodeParams: NodeParams, amountIn: MilliSatoshi, expiryIn: CltvExpiry, amountOut: MilliSatoshi, expiryOut: CltvExpiry): RouteParams = {
-    val routeMaxCltv = expiryIn - expiryOut - nodeParams.expiryDelta
-    val routeMaxFee = amountIn - amountOut - nodeFee(nodeParams.feeBase, nodeParams.feeProportionalMillionth, amountOut)
+    val routeMaxCltv = expiryIn - expiryOut
+    val routeMaxFee = amountIn - amountOut
     RouteCalculation.getDefaultRouteParams(nodeParams.routerConf).copy(
       maxFeeBase = routeMaxFee,
       routeMaxCltv = routeMaxCltv,
-      maxFeePct = 0 // we disable percent-based max fee calculation, we're only interested in collecting our node fee
+      maxFeePct = 0, // we disable percent-based max fee calculation, we're only interested in collecting our node fee
+      isRelay = true,
     )
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
@@ -126,7 +126,7 @@ object NodeRelay {
       maxFeeBase = routeMaxFee,
       routeMaxCltv = routeMaxCltv,
       maxFeePct = 0, // we disable percent-based max fee calculation, we're only interested in collecting our node fee
-      isRelay = true,
+      includeLocalChannelCost = true,
     )
   }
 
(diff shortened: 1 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -95,10 +96,11 @@ object Graph {
                         pathsToFind: Int,
                         wr: Option[WeightRatios],
                         currentBlockHeight: Long,
-                        boundaries: RichWeight => Boolean): Seq[WeightedPath] = {
+                        boundaries: RichWeight => Boolean,
+                        isRelay: Boolean): Seq[WeightedPath] = {
```
t-bast, line 100: I would rename it to something that makes what it does more clear, for example `includeLocalChannel(Cost|Weight)` or `countLocalChannel(Cost|Weight)` (or on the contrary, if you find it more clear, that parameter could be for *ignoring* local channel costs: `ignoreLocalChannel(Cost|Weight)`)

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -432,7 +432,7 @@ object Router {
 
   case class MultiPartParams(minPartAmount: MilliSatoshi, maxParts: Int)
 
-  case class RouteParams(randomize: Boolean, maxFeeBase: MilliSatoshi, maxFeePct: Double, routeMaxLength: Int, routeMaxCltv: CltvExpiryDelta, ratios: Option[WeightRatios], mpp: MultiPartParams) {
+  case class RouteParams(randomize: Boolean, maxFeeBase: MilliSatoshi, maxFeePct: Double, routeMaxLength: Int, routeMaxCltv: CltvExpiryDelta, ratios: Option[WeightRatios], mpp: MultiPartParams, isRelay: Boolean) {
```
t-bast, line 435: Same here, I would change the name of `isRelay` to something that more clearly explains that it governs whether we'll count local channel fees or not.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -120,12 +120,13 @@ object NodeRelay {
 
   /** Compute route params that honor our fee and cltv requirements. */
   def computeRouteParams(nodeParams: NodeParams, amountIn: MilliSatoshi, expiryIn: CltvExpiry, amountOut: MilliSatoshi, expiryOut: CltvExpiry): RouteParams = {
```
t-bast, line 122: NB: we still verify in `validateRelay` above that the fee is at least our current default fee, so we ignore the fact that we may have channels that require lower fees and could relay this payment.

But I think it's the right behavior to have, trampoline is more computationally expensive than standard channel-relay so it makes sense to have a higher threshold than our cheapest channels.

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -1662,6 +1662,14 @@ class RouteCalculationSpec extends AnyFunSuite with ParallelTestExecution {
     val fees = routes.map(_.fee)
     assert(fees.forall(_ == fees.head))
   }
+
+  test("can't relay if fee is not sufficient") {
+    val g = DirectedGraph(List(
+      makeEdge(1L, a, b, 1000 msat, 7000),
+    ))
+
+    val Failure(_) = findRoute(g, a, b, 10000000 msat, 10000 msat, numRoutes = 3, routeParams = DEFAULT_ROUTE_PARAMS.copy(isRelay = true), currentBlockHeight = 400000)
```
t-bast, line 1671: nit: we can be a little more precise:
```
assert(findRoute(g, a, b, 10000000 msat, 10000 msat, numRoutes = 3, routeParams = DEFAULT_ROUTE_PARAMS.copy(isRelay = true), currentBlockHeight = 400000) === Failure(RouteNotFound)
```

And I would also add the positive case for completeness:

```
assert(findRoute(g, a, b, 10000000 msat, 100000 msat, numRoutes = 3, routeParams = DEFAULT_ROUTE_PARAMS.copy(isRelay = true), currentBlockHeight = 400000).isSuccess)
```

t-bast (commented):
Thanks, this is indeed simple enough to include asap ;)

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
thomash-acinq: Renamed to includeLocalChannelCost, I agree it's a better name.

t-bast (approved):
LGTM!

Make sure to choose the "Squash and merge" option when merging, and provide a good commit title and message.

PR #1854, opened 2021-06-30
Decode warning messages

t-bast:
Add support for logging warning messages as introduced in https://github.com/lightningnetwork/lightning-rfc/pull/834
Support for sending warning messages instead of current errors will be added in a later PR.
Note that c-lightning already sends these messages, so it will be useful to have them correctly logged.
Even if the spec PR is eventually changed, this will not introduce backwards-compatibility issues so IMO it's worth merging on master now.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
@@ -34,7 +34,7 @@ import fr.acinq.eclair.io.Monitoring.Metrics
 import fr.acinq.eclair.io.PeerConnection.KillReason
 import fr.acinq.eclair.remote.EclairInternalsSerializer.RemoteTypes
 import fr.acinq.eclair.wire.protocol
-import fr.acinq.eclair.wire.protocol.{Error, HasChannelId, HasTemporaryChannelId, LightningMessage, NodeAddress, RoutingMessage, UnknownMessage}
+import fr.acinq.eclair.wire.protocol.{Error, HasChannelId, HasTemporaryChannelId, LightningMessage, NodeAddress, RoutingMessage, UnknownMessage, Warning}
 import scodec.bits.ByteVector
 
 import java.net.InetSocketAddress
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -49,6 +49,10 @@ object LightningMessageCodecs {
     ("channelId" | bytes32) ::
       ("data" | varsizebinarydata)).as[Error]
 
+  val warningCodec: Codec[Warning] = (
+    ("channelId" | bytes32) ::
+      ("data" | varsizebinarydata)).as[Warning]
+
```
pm47, line 55: Nit: I would move it above the `errorCodec`.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -105,6 +105,12 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: EclairWa
         d.peerConnection forward msg
         stay
 
+      case Event(warning: Warning, _: ConnectedData) =>
+        log.warning("peer sent warning: channelId={} message={}", warning.channelId, warning.toAscii)
```
pm47, line 109: The `channel_id` will be in the MDC:

```suggestion
        log.warning("peer sent warning: message={}", warning.channelId, warning.toAscii)
```

t-bast: Good catch, fixed in https://github.com/ACINQ/eclair/pull/1854/commits/ab67e7ae956ac80267dcc80506559e9a17e48bc5

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
t-bast: Agreed, fixed in https://github.com/ACINQ/eclair/pull/1854/commits/ab67e7ae956ac80267dcc80506559e9a17e48bc5

pm47 (approved):

PR #1857, opened 2021-07-02
Fix default file backup config

t-bast:
The `file-backup` section should be inside `eclair`, otherwise startup fails.


pm47 (approved):
Embarrassing. Here is a micro post mortem:
- the bug wasn't caught by the integration tests because we explicitly disable db backups for them (using the correct config path)
- the bug wasn't caught by the `StartupSpec` test either because the config isn't handled by `NodeParams`.

Issue #1850, opened 2021-06-26
Update error message when testnet estimatesmartfee fails


engenegr:
I'm experiencing crash with

Logs
```
2021-06-26 16:32:26,737 WARN  fr.acinq.eclair.Setup - cannot retrieve feerates: estimatesmartfee failed: Insufficient data or no feerate found
```
Error message
```
fatal error: cannot retrieve feerates: channels may be at risk
```
Actual local mempool statistics

```
{
  "loaded": true,
  "size": 44,
  "bytes": 7473,
  "usage": 55808,
  "total_fee": 0.00321238,
  "maxmempool": 50000000,
  "mempoolminfee": 0.00000000,
  "minrelaytxfee": 0.00000000,
  "unbroadcastcount": 0
}
```
Eclair version `master` commit af618bc44fe6054957f65d77000a460c14b41f30.

Actually just recently the node was fine and worked. I do not know what exactly happened which might lead to such results.

engenegr:
Hint by Anton that it is somehow related to bitcoind. I indeed have 
```
{
  "errors": [
    "Insufficient data or no feerate found"
  ],
  "blocks": 0
}
```
from my bitcoind instance after `bitcoin-cli -testnet estimatesmartfee 6`

engenegr:
Sorry, it works now.

It seems Bitcoin Core collected enough transactions and now able to serve fees. It is up on you to close this issue. In my opinion node shouldn't crash with such error. 

BTW you have another issue with the similar error.

t-bast:
We could change the error message (to hint that maybe you're on `testnet`/`regtest` and should just wait for a few test blocks), but apart from that I'm not sure what we should do differently? If we cannot retrieve feerates, the node is really at risk. Crashing the node is the best way to alert the operator and ensure they can fix this (by fixing their bitcoind node).

This happens on `regtest` and `testnet` because you usually don't run test nodes like mainnet ones, and `bitcoind` doesn't have enough data to estimate the fees. It never happens on `mainnet` (unless there is something really wrong with your node), so I'm not sure it's worth fixing. Do you have a suggestion of what you'd change?

engenegr:
Well, yes. Maybe changing error message on kind-of-alert "wait and restart node" makes sense in this situation.

t-bast:
Can you try submitting a PR to fix that?

engenegr:
Sure, why not.

PR #1859, opened 2021-07-02
changed message

engenegr:
A bit more verbose message but it contains clear instruction now.

Closes https://github.com/ACINQ/eclair/issues/1850

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
@@ -27,7 +27,7 @@ trait FeeProvider {
   def getFeerates: Future[FeeratesPerKB]
 }
 
-case object CannotRetrieveFeerates extends RuntimeException("cannot retrieve feerates: channels may be at risk")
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
```diff
@@ -27,7 +27,7 @@ trait FeeProvider {
   def getFeerates: Future[FeeratesPerKB]
 }
 
-case object CannotRetrieveFeerates extends RuntimeException("cannot retrieve feerates: channels may be at risk")
+case object CannotRetrieveFeerates extends RuntimeException("cannot retrieve feerates: ensure bitcoind serves rates and restart eclair")
```
t-bast, line 30: nit: I would still keep `channels may be at risk`, to make sure node operators understand this is quite a serious issue (if it happens on mainnet).
```suggestion
case object CannotRetrieveFeerates extends RuntimeException("cannot retrieve feerates, channels may be at risk: ensure bitcoind estimatesmartfee correctly returns feerates and restart eclair")
```

t-bast (commented):
Thanks!

pm47 (approved):

PR #1860, opened 2021-07-05
Reduce Postgres transaction isolation level

### LINKED ISSUE #1856: Node crash during payment
I've tried to make payment (Mainnet) from other CL node.

My Eclair just crashed meanwhile with such error log:

```
2021-06-30 16:52:31,074 ERROR f.a.e.channel.Channel n:02c16cca44562b590dd279c942200bdccfd4f990c3a69fad620c10ef2f8228eaff c:96f97fc3414dc324636fc08365d50fc096b2dc24acc2faf6dbda3ca49513d4a7 - fatal database error
org.postgresql.util.PSQLException: ERROR: could not serialize access due to concurrent update
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2533)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2268)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:313)
	at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:448)
	at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:369)
	at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:159)
	at org.postgresql.jdbc.PgPreparedStatement.executeUpdate(PgPreparedStatement.java:125)
	at com.zaxxer.hikari.pool.ProxyPreparedStatement.executeUpdate(ProxyPreparedStatement.java:61)
	at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.executeUpdate(HikariProxyPreparedStatement.java)
	at fr.acinq.eclair.db.pg.PgChannelsDb.$anonfun$addOrUpdateChannel$3(PgChannelsDb.scala:97)
	at fr.acinq.eclair.db.pg.PgChannelsDb.$anonfun$addOrUpdateChannel$3$adapted(PgChannelsDb.scala:94)
	at fr.acinq.eclair.db.jdbc.JdbcUtils.using(JdbcUtils.scala:50)
	at fr.acinq.eclair.db.jdbc.JdbcUtils.using$(JdbcUtils.scala:46)
	at fr.acinq.eclair.db.pg.PgUtils$.using(PgUtils.scala:32)
	at fr.acinq.eclair.db.pg.PgChannelsDb.$anonfun$addOrUpdateChannel$2(PgChannelsDb.scala:94)
	at fr.acinq.eclair.db.pg.PgChannelsDb.$anonfun$addOrUpdateChannel$2$adapted(PgChannelsDb.scala:86)
	at fr.acinq.eclair.db.pg.PgUtils$PgLock$LeaseLock.$anonfun$withLock$1(PgUtils.scala:123)
	at fr.acinq.eclair.db.pg.PgUtils$.inTransactionInternal(PgUtils.scala:255)
	at fr.acinq.eclair.db.pg.PgUtils$.$anonfun$inTransaction$1(PgUtils.scala:270)
	at fr.acinq.eclair.db.jdbc.JdbcUtils.withConnection(JdbcUtils.scala:35)
	at fr.acinq.eclair.db.jdbc.JdbcUtils.withConnection$(JdbcUtils.scala:32)
	at fr.acinq.eclair.db.pg.PgUtils$.withConnection(PgUtils.scala:32)
	at fr.acinq.eclair.db.pg.PgUtils$.inTransaction(PgUtils.scala:269)
	at fr.acinq.eclair.db.pg.PgUtils$PgLock$LeaseLock.withLock(PgUtils.scala:122)
```

After quick node restart it did even finalized payment.

pm47:
I was able to reproduce #1856 by replaying the "concurrent channel
updates" test with hardcoded additional delays in the database code. It
was due to a conflict between `addOrUpdateChannel` and
`updateChannelMetaTimestampColumn`. The two calls run in parallel and
the latter completed before the former, causing it to fail. Reducing
the isolation level makes the problem disappear.

We reduce the transaction isolation level from `SERIALIZABLE` to
`READ_COMMITTED`. Note that [*]:

> Read Committed is the default isolation level in PostgreSQL.

I'm not sure why we were using a stricter isolation level than the
default one, since we only use very basic queries. Doc does say that:

> This behavior makes Read Committed mode unsuitable for commands that involve complex search conditions; however, it is just right for simpler cases

To make sure this didn't cause regression withe the locking
mechanism, I wrote an additional test specifically on the `withLock`
method.

Here is what the doc says on the `INSERT ON CONFLICT DO UPDATE`
statement, which we use for `addOrUpdateChannel`:

> INSERT with an ON CONFLICT DO UPDATE clause behaves similarly. In Read Committed mode, each row proposed for insertion will either insert or update. Unless there are unrelated errors, one of those two outcomes is guaranteed. If a conflict originates in another transaction whose effects are not yet visible to the INSERT, the UPDATE clause will affect that row, even though possibly no version of that row is conventionally visible to the command.

In the scenario described above, the `addOrUpdate` will update the row
which timestamp was updated in parallel by the
`updateChannelMetaTimestampColumn`, and it's exactly what we want.

Fixes #1856.

[*] https://www.postgresql.org/docs/13/transaction-iso.html

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
@@ -268,7 +268,7 @@ object PgUtils extends JdbcUtils {
 
   def inTransaction[T](f: Connection => T)(implicit dataSource: DataSource): T = {
     withConnection { connection =>
-      inTransactionInternal(IsolationLevel.TRANSACTION_SERIALIZABLE)(connection)(f)
+      inTransactionInternal(IsolationLevel.TRANSACTION_READ_COMMITTED)(connection)(f)
     }
   }
 
--- a/eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
@@ -2,19 +2,19 @@ package fr.acinq.eclair.db
 
 import com.opentable.db.postgres.embedded.EmbeddedPostgres
 import com.typesafe.config.{Config, ConfigFactory}
-import fr.acinq.eclair.db.pg.PgUtils.{JdbcUrlChanged, migrateTable, using}
+import fr.acinq.eclair.db.pg.PgUtils.ExtendedResultSet._
 import fr.acinq.eclair.db.pg.PgUtils.PgLock.{LockFailure, LockFailureHandler}
+import fr.acinq.eclair.db.pg.PgUtils.{JdbcUrlChanged, migrateTable, using}
 import fr.acinq.eclair.{TestKitBaseClass, TestUtils}
-import grizzled.slf4j.Logging
-import org.postgresql.PGConnection
-import org.postgresql.jdbc.PgConnection
 import grizzled.slf4j.{Logger, Logging}
+import org.postgresql.jdbc.PgConnection
+import org.postgresql.util.PGInterval
 import org.scalatest.concurrent.Eventually
 import org.scalatest.funsuite.AnyFunSuiteLike
-import fr.acinq.eclair.db.pg.PgUtils.ExtendedResultSet._
 
 import java.io.File
 import java.util.UUID
+import javax.sql.DataSource
 
 class PgUtilsSpec extends TestKitBaseClass with AnyFunSuiteLike with Eventually {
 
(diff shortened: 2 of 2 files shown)


t-bast (approved):
LGTM, I don't see why we'd need a stronger isolation either.
Let's run with that for a while and see how it behaves.

pm47:
Rebased on #1862.

pm47:
Rebased on master.

t-bast (approved):

PR #1863, opened 2021-07-06
Use warning messages for connection issues

t-bast:
https://github.com/lightningnetwork/lightning-rfc/pull/834 recommends sending warning messages instead of connection-level errors in some cases, which avoids unnecessary channel closure.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1701,16 +1701,18 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     // a channel_reestablish when reconnecting a channel that recently got confirmed, and instead send a funding_locked
     // first and then go silent. This is due to a race condition on their side, so we trigger a reconnection, hoping that
     // we will eventually receive their channel_reestablish.
-    case Event(_: FundingLocked, _) =>
+    case Event(_: FundingLocked, d) =>
       log.warning("received funding_locked before channel_reestablish (known lnd bug): disconnecting...")
+      send(Warning(d.channelId, "spec violation: you sent funding_locked before channel_reestablish"))
       peer ! Peer.Disconnect(remoteNodeId)
       stay
 
     // This handler is a workaround for an issue in lnd similar to the one above: they sometimes send announcement_signatures
     // before channel_reestablish, which is a minor spec violation. It doesn't halt the channel, we can simply postpone
     // that message.
-    case Event(remoteAnnSigs: AnnouncementSignatures, _) =>
+    case Event(remoteAnnSigs: AnnouncementSignatures, d) =>
       log.warning("received announcement_signatures before channel_reestablish (known lnd bug): delaying...")
+      send(Warning(d.channelId, "spec violation: you sent announcement_signatures before channel_reestablish"))
       context.system.scheduler.scheduleOnce(5 seconds, self, remoteAnnSigs)
       stay
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1703,18 +1703,17 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     // we will eventually receive their channel_reestablish.
     case Event(_: FundingLocked, d) =>
       log.warning("received funding_locked before channel_reestablish (known lnd bug): disconnecting...")
-      send(Warning(d.channelId, "spec violation: you sent funding_locked before channel_reestablish"))
-      peer ! Peer.Disconnect(remoteNodeId)
-      stay
+      // NB: we use a small delay to ensure we've sent our warning before disconnecting.
+      context.system.scheduler.scheduleOnce(2 second, peer, Peer.Disconnect(remoteNodeId))
+      stay sending Warning(d.channelId, "spec violation: you sent funding_locked before channel_reestablish")
 
     // This handler is a workaround for an issue in lnd similar to the one above: they sometimes send announcement_signatures
     // before channel_reestablish, which is a minor spec violation. It doesn't halt the channel, we can simply postpone
     // that message.
     case Event(remoteAnnSigs: AnnouncementSignatures, d) =>
       log.warning("received announcement_signatures before channel_reestablish (known lnd bug): delaying...")
-      send(Warning(d.channelId, "spec violation: you sent announcement_signatures before channel_reestablish"))
       context.system.scheduler.scheduleOnce(5 seconds, self, remoteAnnSigs)
-      stay
+      stay sending Warning(d.channelId, "spec violation: you sent announcement_signatures before channel_reestablish")
 
     case Event(ProcessCurrentBlockCount(c), d: HasCommitments) => handleNewBlock(c, d)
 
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1701,16 +1701,18 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     // a channel_reestablish when reconnecting a channel that recently got confirmed, and instead send a funding_locked
     // first and then go silent. This is due to a race condition on their side, so we trigger a reconnection, hoping that
     // we will eventually receive their channel_reestablish.
-    case Event(_: FundingLocked, _) =>
+    case Event(_: FundingLocked, d) =>
       log.warning("received funding_locked before channel_reestablish (known lnd bug): disconnecting...")
+      send(Warning(d.channelId, "spec violation: you sent funding_locked before channel_reestablish"))
       peer ! Peer.Disconnect(remoteNodeId)
       stay
```
pm47, line 1708: I'm guessing we are going to send a few warnings 😜 

The preferred way would be:
```suggestion
      val warning = Warning(d.channelId, "spec violation: you sent funding_locked before channel_reestablish")
      peer ! Peer.Disconnect(remoteNodeId)
      stay sending warning
```

The goal is to have side effects occur outside the handlers themselves.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1701,16 +1701,18 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
 
     // This handler is a workaround for an issue in lnd similar to the one above: they sometimes send announcement_signatures
     // before channel_reestablish, which is a minor spec violation. It doesn't halt the channel, we can simply postpone
     // that message.
-    case Event(remoteAnnSigs: AnnouncementSignatures, _) =>
+    case Event(remoteAnnSigs: AnnouncementSignatures, d) =>
       log.warning("received announcement_signatures before channel_reestablish (known lnd bug): delaying...")
+      send(Warning(d.channelId, "spec violation: you sent announcement_signatures before channel_reestablish"))
       context.system.scheduler.scheduleOnce(5 seconds, self, remoteAnnSigs)
       stay
```
pm47, line 1717:
```suggestion
      val warning = Warning(d.channelId, "spec violation: you sent announcement_signatures before channel_reestablish")
      context.system.scheduler.scheduleOnce(5 seconds, self, remoteAnnSigs)
      stay sending warningy
```

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -195,18 +194,24 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
...
+              val latency = System.currentTimeMillis - timestamp
+              log.debug(s"received pong with latency=$latency")
+              cancelTimer(PingTimeout.toString())
+              // we don't need to call scheduleNextPing here, the next ping was already scheduled when we received that pong
+            } else {
+              log.warning(s"ignoring invalid pong with length=${data.length}")
+              d.transport ! Warning(s"invalid pong length (${data.length})")
+            }
           case None =>
             log.debug(s"received unexpected pong with size=${data.length}")
```
pm47, line 216: Maybe we could send a warning for unexpected pongs too? Or you could have factored all cases in the `case None`, sending a generic "invalid pong with length x" ?

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -325,6 +331,7 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
               d.behavior.copy(fundingTxAlreadySpentCount = d.behavior.fundingTxAlreadySpentCount + 1)
             } else {
               log.warning(s"peer sent us too many channel announcements with funding tx already spent (count=${d.behavior.fundingTxAlreadySpentCount + 1}), ignoring network announcements for $IGNORE_NETWORK_ANNOUNCEMENTS_PERIOD")
+              d.transport ! Warning("too many channel announcements with funding tx already spent, please check your bitcoin node")
```
pm47, line 334: Nice one!

pm47 (commented):
I see what you did there! 😉

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: But then we'll disconnect before sending the warning, won't we?

pm47: True. I believe a comment is in order!

There is always a risk though, because sending is asynchronous, but it was much less likely before.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1863/commits/bf10f462f50cd45c865e59370b572a42f736cd39

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1863/commits/bf10f462f50cd45c865e59370b572a42f736cd39

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: I added a delay before the `Peer.Disconnect` in https://github.com/ACINQ/eclair/pull/1863/commits/bf10f462f50cd45c865e59370b572a42f736cd39 this way we can send the warning outside of the handler with `sending`.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -207,8 +207,9 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
             log.debug(s"received pong with latency=$latency")
             cancelTimer(PingTimeout.toString())
           // we don't need to call scheduleNextPing here, the next ping was already scheduled when we received that pong
-          case None =>
+          case _ =>
             log.debug(s"received unexpected pong with size=${data.length}")
+            d.transport ! Warning(s"invalid pong length (${data.length})")
```
pm47, line 212: Nit² (it's not necessarily an issue with the length, it could be  an unexpected pong)

```suggestion
            log.debug(s"received unexpected pong with length=${data.length}")
            d.transport ! Warning(s"invalid pong with length=${data.length}")
```

pm47 (approved):

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1863/commits/3c7c0eed9489e471062615ff40bd3a966f5a5857

pm47 (approved):

PR #1865, opened 2021-07-07
Add json columns in Postgres

pm47:
A json column has been added to the few tables that contains an
opaque serialized blob:
- `local_channels.data`
- `nodes.data`
- `channels.channel_announcement`, `channels.channel_update_x`

We can now access all the individual data fields from SQL.

For the serialization, we use the same serializers than the one
that were previously used by the API. They have been moved to the
`eclair-core` module and simplified a bit.

There are two json data types in Postgres: `JSON` and `JSONB`. We use
the latter one, which is more recent, and allows indexing.

An alternative to this PR would have been to use columns, but:
- there would have been a *lot* of columns for the channel data
- every modification of our types would have required a db migration

NB: to handle non-backwards compatible changes in the json serializersi,
 all the json columns can be recomputed on restart by setting
`eclair.db.reset-json-columns=true`.


eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgChannelsDb.scala
```diff
@@ -62,37 +64,68 @@ class PgChannelsDb(implicit ds: DataSource, lock: PgLock) extends ChannelsDb wit
...
-          | VALUES (?, ?, FALSE)
+          | INSERT INTO local_channels (channel_id, data, json, is_closed)
+          | VALUES (?, ?, ?::JSONB, FALSE)
           | ON CONFLICT (channel_id)
-          | DO UPDATE SET data = EXCLUDED.data ;
+          | DO UPDATE SET data = EXCLUDED.data, json = EXCLUDED.json ;
           | """.stripMargin)) { statement =>
         statement.setString(1, state.channelId.toHex)
         statement.setBytes(2, data)
+        statement.setString(3, serialization.writePretty(state))
```
t-bast, line 128: This will be bad for our performance as JSON is notoriously slow (and this happens at every channel update).
But I think at that point we're not chasing perf at all, the convenience and ease of administration it provides is worth it.

In the (long term) future, I think the "main" DB shouldn't have this column at all, only replicas used for administration should.

eclair-core/src/test/scala/fr/acinq/eclair/db/ChannelsDbSpec.scala
```diff
@@ -324,7 +324,20 @@ class ChannelsDbSpec extends AnyFunSuite {
...
+    val dbs = TestPgDatabases()
+    val db = dbs.channels
+    val channel = ChannelCodecsSpec.normal
+    db.addOrUpdateChannel(channel)
+    dbs.connection.execSQLUpdate("UPDATE local_channels SET json='{}'")
+    db.asInstanceOf[PgChannelsDb].resetJsonColumns(dbs.connection)
+    assert({
+      val res = dbs.connection.execSQLQuery("SELECT * FROM local_channels")
+      res.next()
+      res.getString("json").length > 100
```
t-bast, line 339: Can't we test equality properly?

eclair-core/src/test/scala/fr/acinq/eclair/db/NetworkDbSpec.scala
```diff
@@ -260,9 +223,193 @@ class NetworkDbSpec extends AnyFunSuite {
...
+      t.update_1_opt.foreach(db.updateChannel)
+      t.update_2_opt.foreach(db.updateChannel)
+    }
+    dbs.connection.execSQLUpdate("UPDATE nodes SET json='{}'")
+    dbs.connection.execSQLUpdate("UPDATE channels SET channel_announcement_json='{}',channel_update_1_json=NULL,channel_update_2_json=NULL")
+    db.asInstanceOf[PgNetworkDb].resetJsonColumns(dbs.connection)
+    assert({
+      val res = dbs.connection.execSQLQuery("SELECT * FROM nodes")
+      res.next()
+      res.getString("json").length > 100
```
t-bast, line 344: Same, can't we test equality (same below)?

t-bast (commented):
First review for `5a8d76d`

t-bast (commented):
Reviewing `e5cb371`, could you also fix `ChannelCodecsSpec` where we copy-pasted the serializers because we couldn't reference them once they were in `eclair-node`, but now we can directly reference them instead of duplicating.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgChannelsDb.scala
pm47: I don't think the performance impact would be that big, because 1) we can parallelize the `scodec`/`json` serialization and 2) the serialization time is much smaller than the actual db write.

We could also make the json column optional.

> In the (long term) future, I think the "main" DB shouldn't have this column at all, only replicas used for administration should.

This would create complications, because then we would have to manually take care of updating the json column. It can't be delegated to db triggers since it requires custom json serializers.

eclair-core/src/main/resources/reference.conf
```diff
@@ -224,6 +224,7 @@ eclair {
       username = ""
       password = ""
       readonly-user = "" // if defined, this user will be granted read-only access to all tables in the database
+      reset-json-columns = false // in case of a json format change, this allows a full re-serialization of json data
```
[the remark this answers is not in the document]
pm47: > I think people will almost always forget to set this to true when needed.

They may, but I don't think it's a big deal.

I believe we should make no guarantee whatsoever on json backwards compatibility, and accept that this column is here for analytics/advanced admin purposes. Otherwise very change to the codecs will be hard. In my view, we shouldn't even bother with that and only trigger a json reset when some SQL query starts to fail.

NB: the full json rewrite is fast, it's only a few 10k lines, even on our large node.

eclair-core/src/test/scala/fr/acinq/eclair/db/ChannelsDbSpec.scala
pm47: It's not straightforward because with the `JSONB` type, Postgres mutates the data:

> The json data type stores an exact copy of the input text, which processing functions must reparse on each execution; while jsonb data is stored in a decomposed binary format that makes it slightly slower to input due to added conversion overhead, but significantly faster to process, since no reparsing is needed

https://www.postgresql.org/docs/13/datatype-json.html

In theory we could deserialize then reserialize but... we don't support deserialization. I think it's okay if we assume that this test is only meant to test that the data is actually reset?

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
```diff
@@ -98,6 +99,35 @@ trait JdbcUtils {
...
+      while (rs.next()) {
+        migrate(rs, insertStatement)
+        insertStatement.addBatch()
+        batchCount = batchCount + 1
+        if (batchCount % batchSize == 0) {
+          inserted = inserted + insertStatement.executeBatch().sum
+          batchCount = 0
+        }
+      }
+      inserted = inserted + insertStatement.executeBatch().sum
```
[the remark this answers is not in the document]
pm47: I added a test in b6773af2b.

eclair-core/src/test/scala/fr/acinq/eclair/db/ChannelsDbSpec.scala
t-bast: Yes, that sounds good enough!

pm47:
> Reviewing e5cb371, could you also fix ChannelCodecsSpec where we copy-pasted the serializers because we couldn't reference them once they were in eclair-node, but now we can directly reference them instead of duplicating.

Good catch, done with c5804ada4. Note that this test is pretty subtle and brings strong non-reg guarantees, see the commit message.

t-bast:
ACK https://github.com/ACINQ/eclair/pull/1865/commits/c5804ada4888e66dcaae89d9b3d1c24a34e76e33

Very nice to improve these tests, adding json serialization will indeed make that kind of tests more readable in the future, we should leverage it more.

pm47:
Rebased on parent PR.

pm47:
Parent PR merged, rebased on master.

t-bast (approved):

PR #1867, opened 2021-07-07
Explicit channel type in channel open

t-bast:
Add support for https://github.com/lightningnetwork/lightning-rfc/pull/880

This lets node operators open a channel with different features than what the implicit choice based on activated features would use.

This has been tested successfully against https://github.com/lightningnetwork/lnd/pull/5373

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -195,7 +195,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   startWith(WAIT_FOR_INIT_INTERNAL, Nothing)
 
   when(WAIT_FOR_INIT_INTERNAL)(handleExceptions {
-    case Event(initFunder@INPUT_INIT_FUNDER(temporaryChannelId, fundingSatoshis, pushMsat, initialFeeratePerKw, fundingTxFeeratePerKw, localParams, remote, _, channelFlags, channelConfig, channelFeatures), Nothing) =>
+    case Event(initFunder@INPUT_INIT_FUNDER(temporaryChannelId, fundingSatoshis, pushMsat, initialFeeratePerKw, fundingTxFeeratePerKw, localParams, remote, remoteInit, channelFlags, channelConfig, channelType), Nothing) =>
       context.system.eventStream.publish(ChannelCreated(self, peer, remoteNodeId, isFunder = true, temporaryChannelId, initialFeeratePerKw, Some(fundingTxFeeratePerKw)))
       activeConnection = remote
       txPublisher ! SetChannelId(remoteNodeId, temporaryChannelId)
@@ -203,7 +203,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       val channelKeyPath = keyManager.keyPath(localParams, channelConfig)
       // In order to allow TLV extensions and keep backwards-compatibility, we include an empty upfront_shutdown_script if this feature is not used
       // See https://github.com/lightningnetwork/lightning-rfc/pull/714.
-      val localShutdownScript = if (channelFeatures.hasFeature(Features.OptionUpfrontShutdownScript)) localParams.defaultFinalScriptPubKey else ByteVector.empty
+      val localShutdownScript = if (Features.canUseFeature(localParams.initFeatures, remoteInit.features, Features.OptionUpfrontShutdownScript)) localParams.defaultFinalScriptPubKey else ByteVector.empty
       val open = OpenChannel(nodeParams.chainHash,
         temporaryChannelId = temporaryChannelId,
         fundingSatoshis = fundingSatoshis,
@@ -222,7 +222,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         htlcBasepoint = keyManager.htlcPoint(channelKeyPath).publicKey,
         firstPerCommitmentPoint = keyManager.commitmentPoint(channelKeyPath, 0),
         channelFlags = channelFlags,
-        tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScript(localShutdownScript)))
+        tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScript(localShutdownScript), ChannelTlv.ChannelType(channelType.features)))
       goto(WAIT_FOR_ACCEPT_CHANNEL) using DATA_WAIT_FOR_ACCEPT_CHANNEL(initFunder, open) sending open
 
     case Event(inputFundee@INPUT_INIT_FUNDEE(_, localParams, remote, _, _, _), Nothing) if !localParams.isFunder =>
@@ -337,18 +337,17 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   })
 
   when(WAIT_FOR_OPEN_CHANNEL)(handleExceptions {
-    case Event(open: OpenChannel, d@DATA_WAIT_FOR_OPEN_CHANNEL(INPUT_INIT_FUNDEE(_, localParams, _, remoteInit, channelConfig, channelFeatures))) =>
-      log.info("received OpenChannel={}", open)
-      Helpers.validateParamsFundee(nodeParams, localParams.initFeatures, channelFeatures, open, remoteNodeId) match {
+    case Event(open: OpenChannel, d@DATA_WAIT_FOR_OPEN_CHANNEL(INPUT_INIT_FUNDEE(_, localParams, _, remoteInit, channelConfig, channelType))) =>
+      Helpers.validateParamsFundee(nodeParams, localParams.initFeatures, channelType, open, remoteNodeId, remoteInit.features) match {
         case Left(t) => handleLocalError(t, d, Some(open))
-        case Right(remoteShutdownScript) =>
+        case Right((channelFeatures, remoteShutdownScript)) =>
           context.system.eventStream.publish(ChannelCreated(self, peer, remoteNodeId, isFunder = false, open.temporaryChannelId, open.feeratePerKw, None))
           val fundingPubkey = keyManager.fundingPublicKey(localParams.fundingKeyPath).publicKey
           val channelKeyPath = keyManager.keyPath(localParams, channelConfig)
           val minimumDepth = Helpers.minDepthForFunding(nodeParams, open.fundingSatoshis)
           // In order to allow TLV extensions and keep backwards-compatibility, we include an empty upfront_shutdown_script if this feature is not used.
           // See https://github.com/lightningnetwork/lightning-rfc/pull/714.
-          val localShutdownScript = if (channelFeatures.hasFeature(Features.OptionUpfrontShutdownScript)) localParams.defaultFinalScriptPubKey else ByteVector.empty
+          val localShutdownScript = if (Features.canUseFeature(localParams.initFeatures, remoteInit.features, Features.OptionUpfrontShutdownScript)) localParams.defaultFinalScriptPubKey else ByteVector.empty
           val accept = AcceptChannel(temporaryChannelId = open.temporaryChannelId,
             dustLimitSatoshis = localParams.dustLimit,
             maxHtlcValueInFlightMsat = localParams.maxHtlcValueInFlightMsat,
@@ -363,7 +362,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
             delayedPaymentBasepoint = keyManager.delayedPaymentPoint(channelKeyPath).publicKey,
             htlcBasepoint = keyManager.htlcPoint(channelKeyPath).publicKey,
             firstPerCommitmentPoint = keyManager.commitmentPoint(channelKeyPath, 0),
-            tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScript(localShutdownScript)))
+            tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScript(localShutdownScript), ChannelTlv.ChannelType(channelType.features)))
           val remoteParams = RemoteParams(
             nodeId = remoteNodeId,
             dustLimit = open.dustLimitSatoshis,
@@ -391,11 +390,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
   })
 
   when(WAIT_FOR_ACCEPT_CHANNEL)(handleExceptions {
-    case Event(accept: AcceptChannel, d@DATA_WAIT_FOR_ACCEPT_CHANNEL(INPUT_INIT_FUNDER(temporaryChannelId, fundingSatoshis, pushMsat, initialFeeratePerKw, fundingTxFeeratePerKw, localParams, _, remoteInit, _, channelConfig, channelFeatures), open)) =>
-      log.info(s"received AcceptChannel=$accept")
-      Helpers.validateParamsFunder(nodeParams, channelFeatures, open, accept) match {
+    case Event(accept: AcceptChannel, d@DATA_WAIT_FOR_ACCEPT_CHANNEL(INPUT_INIT_FUNDER(temporaryChannelId, fundingSatoshis, pushMsat, initialFeeratePerKw, fundingTxFeeratePerKw, localParams, _, remoteInit, _, channelConfig, channelType), open)) =>
+      Helpers.validateParamsFunder(nodeParams, channelType, localParams.initFeatures, remoteInit.features, open, accept) match {
         case Left(t) => handleLocalError(t, d, Some(accept))
-        case Right(remoteShutdownScript) =>
+        case Right((channelFeatures, remoteShutdownScript)) =>
           val remoteParams = RemoteParams(
             nodeId = remoteNodeId,
             dustLimit = accept.dustLimitSatoshis,
@@ -1681,7 +1679,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           val shutdownInProgress = d.localShutdown.nonEmpty || d.remoteShutdown.nonEmpty
           if (d.commitments.localParams.isFunder && !shutdownInProgress) {
             val currentFeeratePerKw = d.commitments.localCommit.spec.feeratePerKw
-            val networkFeeratePerKw = nodeParams.onChainFeeConf.getCommitmentFeerate(remoteNodeId, d.commitments.channelFeatures, d.commitments.capacity, None)
+            val networkFeeratePerKw = nodeParams.onChainFeeConf.getCommitmentFeerate(remoteNodeId, d.commitments.channelType, d.commitments.capacity, None)
             if (nodeParams.onChainFeeConf.shouldUpdateFee(currentFeeratePerKw, networkFeeratePerKw)) {
               self ! CMD_UPDATE_FEE(networkFeeratePerKw, commit = true)
             }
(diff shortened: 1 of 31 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/ChannelTlv.scala
```diff
@@ -34,14 +34,25 @@ object ChannelTlv {
...
 
+  val upfrontShutdownScriptCodec: Codec[UpfrontShutdownScript] = variableSizeBytesLong(varintoverflow, bytes).as[UpfrontShutdownScript]
+
+  /** A channel type is a set of even feature bits that represent persistent features which affect channel operations. */
+  case class ChannelType(features: Features) extends OpenChannelTlv with AcceptChannelTlv
+
+  val channelTypeCodec: Codec[ChannelType] = variableSizeBytesLong(varintoverflow, bytes).xmap(
+    b => ChannelType(Features(b)),
+    ct => ct.features.toByteVector
+  )
```
pm47, line 45: I think we should:
- suffix tlvs by `Tlv` as a convention, this is consistent with the naming of traits, and removes conflicts between `ChannelTlv.ChannelType`/`channel.ChannelType`
- not expose the underlying `features`, which imo are a serialization detail
- make `ChannelTypes.fromFeatures` return a `Try` (this would make `Attempt.fromOption` move to `Attempt.fromTry`) (*)

```suggestion
  case class ChannelTypeTlv(channelType: ChannelType) extends OpenChannelTlv with AcceptChannelTlv

  val channelTypeCodec: Codec[ChannelTypeTlv] = variableSizeBytesLong(varintoverflow, bytes).exmap(
    b => Attempt.fromOption(ChannelTypes.fromFeatures(Features(b)).map(ChannelTypeTlv), ifNone = Err(s"unknown channel type with features=${Features(b)}")),
    ct => Attempt.successful(ct.channelType.features.toByteVector)
  )
```

(*) on 2nd thought: an unknown channel type will trigger a deserialization error and a connection close. The spec requires a *channel close* only:
> The receiving node MUST fail the channel if:
> - It supports `channel_type`, `channel_type` was set, and the `type` is not suitable.

(by the way there is an ambiguity between *fail* and *reject* the channel in the spec)

I think a cleaner approach would be to define an `UnsupportedChannelType` along `Standard`/`StaticRemoteKey`/`AnchorOutputs`. There could still be deserialization errors if, for example, odd bits are used, but in case it's just an unknown channel type, we could handle it properly. Just ignoring unknown channel types doesn't work because then we can't discriminate between "unknown channel type" and "no channel type".

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
```diff
@@ -55,18 +62,64 @@ object ChannelFeatures {
...
+  }
+  // @formatter:on
+
+  // NB: Bolt 2: features must exactly match in order to identify a channel type.
+  def fromFeatures(features: Features): Option[ChannelType] = features match {
+    case f if f == AnchorOutputs.features => Some(AnchorOutputs)
+    case f if f == StaticRemoteKey.features => Some(StaticRemoteKey)
+    case f if f == Standard.features => Some(Standard)
+    case _ => None
+  }
```
pm47, line 112:
```suggestion
  def fromFeatures(features: Features): Try[ChannelType] = features match {
    case f if f == AnchorOutputs.features => Success(AnchorOutputs)
    case f if f == StaticRemoteKey.features => Success(StaticRemoteKey)
    case f if f == Standard.features => Success(Standard)
    case _ => Failure(new RuntimeException(s"unknown channel type with features=$features"))
  }
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -103,6 +103,7 @@ case class OpenChannel(chainHash: ByteVector32,
                        channelFlags: Byte,
                        tlvStream: TlvStream[OpenChannelTlv] = TlvStream.empty) extends ChannelMessage with HasTemporaryChannelId with HasChainHash {
   val upfrontShutdownScript_opt: Option[ByteVector] = tlvStream.get[ChannelTlv.UpfrontShutdownScript].map(_.script)
+  val channelType_opt: Option[Features] = tlvStream.get[ChannelTlv.ChannelType].map(_.features)
```
pm47, line 106:
```suggestion
  val channelType_opt: Option[ChannelType] = tlvStream.get[ChannelTlv.ChannelTypeTlv].map(_.channelType)
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -121,6 +122,7 @@ case class AcceptChannel(temporaryChannelId: ByteVector32,
                          firstPerCommitmentPoint: PublicKey,
                          tlvStream: TlvStream[AcceptChannelTlv] = TlvStream.empty) extends ChannelMessage with HasTemporaryChannelId {
   val upfrontShutdownScript_opt: Option[ByteVector] = tlvStream.get[ChannelTlv.UpfrontShutdownScript].map(_.script)
+  val channelType_opt: Option[Features] = tlvStream.get[ChannelTlv.ChannelType].map(_.features)
```
pm47, line 125:
```suggestion
  val channelType_opt: Option[ChannelType] = tlvStream.get[ChannelTlv.ChannelTypeTlv].map(_.channelType)
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -137,28 +137,45 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: EclairWa
...
+                channel ! msg
+                stay() using d.copy(channels = d.channels + (TemporaryChannelId(temporaryChannelId) -> channel))
+              case None =>
+                msg.channelType_opt.foreach(proposedChannelType => {
+                  log.warning(s"ignoring open_channel with invalid channel type=$proposedChannelType")
+                  val ourDefaultChannelType = ChannelTypes.pickChannelType(d.localFeatures, d.localFeatures)
+                  d.peerConnection ! Error(msg.temporaryChannelId, InvalidChannelType(msg.temporaryChannelId, ourDefaultChannelType.features, proposedChannelType).getMessage)
+                })
+                stay()
+            }
```
pm47, line 178: Here I would only deal with channel types, not features. Following up with my idea about `UnsupportedChannelTypes`, this is what I get:

```suggestion
            val channelType_either: Either[ChannelException, ChannelType] = msg.channelType_opt match {
              case Some(unsupportedChannelType: UnsupportedChannelType) =>
                  log.warning(s"ignoring open_channel with invalid channel type=$unsupportedChannelType")
                  val ourDefaultChannelType = ChannelTypes.pickChannelType(d.localFeatures, d.localFeatures)
                  Left(InvalidChannelType(msg.temporaryChannelId, ourDefaultChannelType.features, unsupportedChannelType).getMessage))
              case Some(channelType) => Right(channelType)
              case None => Right(ChannelTypes.pickChannelType(d.localFeatures, d.remoteFeatures))
            }
            channelType_either match {
              case Right(channelType) =>
                val (channel, localParams) = createNewChannel(nodeParams, d.localFeatures, channelType, funder = false, fundingAmount = msg.fundingSatoshis, origin_opt = None)
                val temporaryChannelId = msg.temporaryChannelId
                log.info(s"accepting a new channel with type=$channelType temporaryChannelId=$temporaryChannelId localParams=$localParams")
                channel ! INPUT_INIT_FUNDEE(temporaryChannelId, localParams, d.peerConnection, d.remoteInit, channelConfig, channelType)
                channel ! msg
                stay() using d.copy(channels = d.channels + (TemporaryChannelId(temporaryChannelId) -> channel))
              case Left(error) =>
                  d.peerConnection ! error
                stay()
            }
```

I'm not sure if we should include a channel type tlv in the `accept_channel`, if there was none in the `open_channel`. The spec does allow both behaviors.

pm47 (commented):
This is a first pass, let me know if you like where it's going before I dig deeper.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/ChannelTlv.scala
t-bast: > I think a cleaner approach would be to define an UnsupportedChannelType along Standard/StaticRemoteKey/AnchorOutputs

Good idea, I'll try that.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
```diff
@@ -55,18 +62,64 @@ object ChannelFeatures {
...
+    ChannelFeatures(allFeatures: _*)
+  }
+
+}
+
+/** A channel type is a specific set of even feature bits that represent persistent channel features as defined in Bolt 2. */
+sealed trait ChannelType {
+  // @formatter:off
+  /** Features representing that channel type. */
+  def features: Features
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1867/commits/03bf4c363efa65184c39a2af079fb8976f389b77

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -222,7 +222,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         htlcBasepoint = keyManager.htlcPoint(channelKeyPath).publicKey,
         firstPerCommitmentPoint = keyManager.commitmentPoint(channelKeyPath, 0),
         channelFlags = channelFlags,
-        tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScript(localShutdownScript)))
+        tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScript(localShutdownScript), ChannelTlv.ChannelType(channelType.features)))
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1867/commits/03bf4c363efa65184c39a2af079fb8976f389b77

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/ChannelTlv.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1867/commits/03bf4c363efa65184c39a2af079fb8976f389b77

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
t-bast: Done slightly differently in https://github.com/ACINQ/eclair/pull/1867/commits/03bf4c363efa65184c39a2af079fb8976f389b77

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1867/commits/03bf4c363efa65184c39a2af079fb8976f389b77

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1867/commits/03bf4c363efa65184c39a2af079fb8976f389b77

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1867/commits/03bf4c363efa65184c39a2af079fb8976f389b77
I think we should always send our `channel_type`, it's harmless and more explicit so it's always a win.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -144,13 +147,25 @@ object Helpers {
...
-    extractShutdownScript(open.temporaryChannelId, channelFeatures, open.upfrontShutdownScript_opt)
+    val channelFeatures = ChannelFeatures(channelType, initFeatures, remoteFeatures)
+    extractShutdownScript(open.temporaryChannelId, initFeatures, remoteFeatures, open.upfrontShutdownScript_opt).map(script_opt => (channelFeatures, script_opt))
   }
 
   /**
    * Called by the funder
    */
-  def validateParamsFunder(nodeParams: NodeParams, channelFeatures: ChannelFeatures, open: OpenChannel, accept: AcceptChannel): Either[ChannelException, Option[ByteVector]] = {
+  def validateParamsFunder(nodeParams: NodeParams, channelType: SupportedChannelType, initFeatures: Features, remoteFeatures: Features, open: OpenChannel, accept: AcceptChannel): Either[ChannelException, (ChannelFeatures, Option[ByteVector])] = {
```
pm47, line 157:
```suggestion
  def validateParamsFunder(nodeParams: NodeParams, channelType: SupportedChannelType, localFeatures: Features, remoteFeatures: Features, open: OpenChannel, accept: AcceptChannel): Either[ChannelException, (ChannelFeatures, Option[ByteVector])] = {
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -96,7 +99,7 @@ object Helpers {
   /**
    * Called by the fundee
    */
-  def validateParamsFundee(nodeParams: NodeParams, initFeatures: Features, channelFeatures: ChannelFeatures, open: OpenChannel, remoteNodeId: PublicKey): Either[ChannelException, Option[ByteVector]] = {
+  def validateParamsFundee(nodeParams: NodeParams, initFeatures: Features, channelType: SupportedChannelType, open: OpenChannel, remoteNodeId: PublicKey, remoteFeatures: Features): Either[ChannelException, (ChannelFeatures, Option[ByteVector])] = {
```
pm47, line 102: Nit (or `localInitFeatures`/`remoteInitFeatures`?)

```suggestion
  def validateParamsFundee(nodeParams: NodeParams, channelType: SupportedChannelType, localFeatures: Features, open: OpenChannel, remoteNodeId: PublicKey, remoteFeatures: Features): Either[ChannelException, (ChannelFeatures, Option[ByteVector])] = {
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -137,28 +138,48 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: EclairWa
...
-            channel ! INPUT_INIT_FUNDEE(temporaryChannelId, localParams, d.peerConnection, d.remoteInit, channelConfig, channelFeatures)
-            channel ! msg
-            stay() using d.copy(channels = d.channels + (TemporaryChannelId(temporaryChannelId) -> channel))
+            val defaultChannelType = ChannelTypes.pickChannelType(d.localFeatures, d.remoteFeatures)
+            val chosenChannelType: Either[InvalidChannelType, SupportedChannelType] = msg.channelType_opt match {
+              case None => Right(defaultChannelType)
+              case Some(proposedChannelType: UnsupportedChannelType) => Left(InvalidChannelType(msg.temporaryChannelId, defaultChannelType, proposedChannelType))
+              case Some(proposedChannelType: SupportedChannelType) =>
+                // We ensure that we support the features necessary for this channel type.
+                val featuresSupported = proposedChannelType.features.forall(f => d.localFeatures.hasFeature(f))
```
pm47, line 163: So we cannot choose a channel type that depends on a feature that we didn't explicitly activate in our init after all?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -137,28 +138,48 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: EclairWa
...
           case None =>
             val channelConfig = ChannelConfig.standard
-            val channelFeatures = ChannelFeatures.pickChannelFeatures(d.localFeatures, d.remoteFeatures)
-            val (channel, localParams) = createNewChannel(nodeParams, d.localFeatures, channelFeatures, funder = false, fundingAmount = msg.fundingSatoshis, origin_opt = None)
-            val temporaryChannelId = msg.temporaryChannelId
-            log.info(s"accepting a new channel with temporaryChannelId=$temporaryChannelId localParams=$localParams")
-            channel ! INPUT_INIT_FUNDEE(temporaryChannelId, localParams, d.peerConnection, d.remoteInit, channelConfig, channelFeatures)
-            channel ! msg
-            stay() using d.copy(channels = d.channels + (TemporaryChannelId(temporaryChannelId) -> channel))
+            val defaultChannelType = ChannelTypes.pickChannelType(d.localFeatures, d.remoteFeatures)
```
pm47, line 157: Nice change.

I wonder if we should have done this validation in `Helpers.validateFundee`?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: For now no, we will introduce the ability to accept channel types without activating the corresponding features later.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: We cannot because of `static_remotekey`, we need to know the channel type here to correctly derive the payment basepoint (based on `channelType.paysDirectlyToWallet`). That's definitely something that bothered me, and it would be nice to change this in a later PR, but I suspect it's not going to be that simple.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1867/commits/439643cbf3f8215cd913f072021d5259903fa5c8

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1867/commits/439643cbf3f8215cd913f072021d5259903fa5c8

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -137,28 +138,48 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: EclairWa
...
+            val chosenChannelType: Either[InvalidChannelType, SupportedChannelType] = msg.channelType_opt match {
+              case None => Right(defaultChannelType)
```
sstone, line 159: Does this mean that we'll add a `channel_type` to our `accept` message even there was none in our peer's `open` message ? If so it looks like a spec violation ?

pm47:
I've put some test improvements on branch https://github.com/ACINQ/eclair/tree/channel-types-pm.

pm47 (commented):
I've put some test improvements on branch https://github.com/ACINQ/eclair/tree/channel-types-pm.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: Yes, we always send back a `channel_type`. It's not a spec violation, the only thing the spec requires is:

```
  - if it sets `channel_type`:
    - MUST set it to the `channel_type` from `open_channel`
```

If it wasn't set in `open_channel`, we're free to do what we want, and it's an odd message type so it's always ok to include.

sstone: But then it will be ignored, so what is the point ?

pm47: It doesn't hurt, is future proof, and we don't have to manage an option?

pm47: Also, it can be viewed as "FYI I'll be using that channel type"?

t-bast: Exactly, it's simpler in the flow to always use `channel_type`s internally, because otherwise you can't as easily distinguish in this piece of code the case where no channel type was sent and the case where an invalid one was sent (it would need a bit more code).

pm47 (dismissed):
LGTM

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
sstone: I guess that what bothered me is that it's not "symmetric" (we set it but they did not) but it does work.

pm47 (approved):

PR #1871, opened 2021-07-12
Enable WAL mode on Sqlite

pm47:
[Write-Ahead Logging](https://sqlite.org/wal.html) is both much more performant in general, and more suited to our particular access patterns.

With a simple throughput performance test, it improves performance by a factor of 5-20x depending on the sync flag.

 version                                 | throughput
-------------------------------|-------------
mode=journal sync=normal (*)| 11 htlc/s
mode=journal sync=full| 7 htlc/s
mode=wal sync=normal| 248 htlc/s
mode=wal sync=full (**)| 62 htlc/s 

(*) baseline (current master)
(**) new default

I went with a conservative new default of wal+full sync, which is both 5x more performant, and more secure than what we had before. There is an argument to be made that a more relaxed default of wal+normal sync could be more appropriate in a production settings, where OS crash / power failure is less of a concern and the disaster scenario is centered on the complete loss of the server: in that scenario we rely on the db backup, which means we typically already accept up to 1 min worth of data loss. Therefore I tend to think that we could recommend:
- raspberry pi home server: wal+full sync
- real server with redundant power: wal+normal sync

> In WAL mode when synchronous is NORMAL (1), the WAL file is synchronized before each checkpoint and the database file is synchronized after each completed checkpoint and the WAL file header is synchronized when a WAL file begins to be reused after a checkpoint, but no sync operations occur during most transactions. With synchronous=FULL in WAL mode, an additional sync operation of the WAL file happens after each transaction commit. The extra WAL sync following each transaction help ensure that transactions are durable across a power loss. Transactions are consistent with or without the extra syncs provided by synchronous=FULL. If durability is not a concern, then synchronous=NORMAL is normally all one needs in WAL mode.

Full details on the trade-offs here: https://www.sqlite.org/pragma.html#pragma_synchronous.

@joostjager I couldn't help but notice that despite using a different test setting, I get exactly the [same result as you](https://bottlepay.com/blog/bitcoin-lightning-node-performance/) with the baseline, on my laptop. Is there any chance you could re-run your test with this branch? Maybe even a 2nd run with `eclair.db.sqlite.sync=normal`?

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
@@ -28,7 +28,6 @@ import fr.acinq.eclair.payment.{IncomingPacket, PaymentReceived, PaymentRequest}
 import fr.acinq.eclair.wire.protocol._
 import fr.acinq.eclair.{Features, Logs, MilliSatoshi, NodeParams, randomBytes32}
 
-import scala.concurrent.{ExecutionContextExecutor, Future}
 import scala.util.{Failure, Success, Try}
 
 /**
@@ -52,13 +51,9 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
   /** Can be overridden to do custom post-processing on successfully received payments. */
   def postFulfill(paymentReceived: PaymentReceived)(implicit log: LoggingAdapter): Unit = ()
 
-  case class PaymentRequestResult(replyTo: ActorRef, res: Try[PaymentRequest])
-
   override def handle(implicit ctx: ActorContext, log: DiagnosticLoggingAdapter): Receive = {
     case ReceivePayment(amount_opt, desc, expirySeconds_opt, extraHops, fallbackAddress_opt, paymentPreimage_opt, paymentType) =>
-      implicit val ec: ExecutionContextExecutor = ctx.dispatcher
-      val replyTo = ctx.sender()
-      Future {
+      Try {
         val paymentPreimage = paymentPreimage_opt.getOrElse(randomBytes32())
         val paymentHash = Crypto.sha256(paymentPreimage)
         val expirySeconds = expirySeconds_opt.getOrElse(nodeParams.paymentRequestExpiry.toSeconds)
@@ -73,12 +68,9 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
         log.debug("generated payment request={} from amount={}", PaymentRequest.write(paymentRequest), amount_opt)
         db.addIncomingPayment(paymentRequest, paymentPreimage, paymentType)
         paymentRequest
-      }.onComplete(res => ctx.self ! PaymentRequestResult(replyTo, res))
-
-    case PaymentRequestResult(replyTo, res) =>
-      res match {
-        case Success(paymentRequest) => replyTo ! paymentRequest
-        case Failure(exception) => replyTo ! Status.Failure(exception)
+      } match {
+        case Success(paymentRequest) => ctx.sender ! paymentRequest
+        case Failure(exception) => ctx.sender ! Status.Failure(exception)
       }
 
     case p: IncomingPacket.FinalPacket if doHandle(p.add.paymentHash) =>
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
@@ -28,6 +28,7 @@ import fr.acinq.eclair.payment.{IncomingPacket, PaymentReceived, PaymentRequest}
 import fr.acinq.eclair.wire.protocol._
 import fr.acinq.eclair.{Features, Logs, MilliSatoshi, NodeParams, randomBytes32}
 
+import scala.concurrent.{ExecutionContextExecutor, Future}
 import scala.util.{Failure, Success, Try}
 
 /**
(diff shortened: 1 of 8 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -51,9 +52,13 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
...
   def postFulfill(paymentReceived: PaymentReceived)(implicit log: LoggingAdapter): Unit = ()
 
+  case class PaymentRequestResult(replyTo: ActorRef, res: Try[PaymentRequest])
+
   override def handle(implicit ctx: ActorContext, log: DiagnosticLoggingAdapter): Receive = {
     case ReceivePayment(amount_opt, desc, expirySeconds_opt, extraHops, fallbackAddress_opt, paymentPreimage_opt, paymentType) =>
-      Try {
+      implicit val ec: ExecutionContextExecutor = ctx.dispatcher
+      val replyTo = ctx.sender()
+      Future {
```
t-bast, line 61: I think it would be cleaner (more actor-like) to spawn a very simple actor for each payment request generation.

eclair-core/src/test/scala/fr/acinq/eclair/integration/PerformanceIntegrationSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+import fr.acinq.eclair.router.Router
+
+import java.util.UUID
+import java.util.concurrent.Executors
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.jdk.CollectionConverters._
+
+/**
+ * Created by PM on 15/03/2017.
```
t-bast, line 37: I think this is a lie. :see_no_evil:

eclair-core/src/test/scala/fr/acinq/eclair/integration/PerformanceIntegrationSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+class PerformanceIntegrationSpec extends IntegrationSpec {
+
+  test("start eclair nodes") {
+    val commonPerfTestConfig = ConfigFactory.parseMap(Map(
+      "eclair.max-funding-satoshis" -> 100_000_000,
+      "eclair.max-accepted-htlcs" -> 30,
```
t-bast, line 45: For perf tests we could make this bigger, don't you think? We could even go to the 483 limit (in case they're added more quickly than they're settled)?

eclair-core/src/test/scala/fr/acinq/eclair/integration/PerformanceIntegrationSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    // confirming the funding tx
+    generateBlocks(6)
+
+    within(60 seconds) {
+      eventListener.expectMsgType[ChannelStateChanged](60 seconds).currentState == NORMAL
+    }
+  }
+
+  test("wait for channels balance") {
```
t-bast, line 71: The name is a bit misleading, shouldn't it just be:
```suggestion
  test("wait for channels") {
```

eclair-core/src/test/scala/fr/acinq/eclair/integration/PerformanceIntegrationSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+      sender.send(nodes("A").router, Router.GetRoutingState)
+      val routingState = sender.expectMsgType[Router.RoutingState]
+      routingState.channels.nonEmpty
+    }, 60 seconds)
+  }
+
+  def sendPayment()(implicit ec: ExecutionContext): Future[PaymentSent] = Future {
+    val sender = TestProbe()
+    val amountMsat = 100_000.msat
+    // first we retrieve a payment hash from D
```
t-bast, line 84:
```suggestion
    // first we retrieve a payment hash from B
```

eclair-core/src/test/scala/fr/acinq/eclair/integration/PerformanceIntegrationSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+import java.util.concurrent.Executors
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.jdk.CollectionConverters._
+
+/**
+ * Created by PM on 15/03/2017.
+ */
+
+class PerformanceIntegrationSpec extends IntegrationSpec {
```
t-bast, line 40: Can we avoid running this in our CI? You can modify the `main.yml` to skip it like we did for external API tests.

t-bast (commented):
Nice, this is a very good find!

eclair-core/src/test/scala/fr/acinq/eclair/integration/PerformanceIntegrationSpec.scala
pm47: Indeed, and this just caused the current CI run to fail.

eclair-core/src/test/scala/fr/acinq/eclair/integration/PerformanceIntegrationSpec.scala
pm47: Or just:
```suggestion
@Ignore
class PerformanceIntegrationSpec extends IntegrationSpec {
```

pm47: Done in ded5108.

eclair-core/src/test/scala/fr/acinq/eclair/integration/PerformanceIntegrationSpec.scala
pm47: Fixed in ded5108.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
pm47: Also, best would probably to put this in a separate PR, I'll do that.

joostjager:
>@joostjager I couldn't help but notice that despite using a different test setting, I get exactly the same result as you with the baseline, on my laptop. Is there any chance you could re-run your test with this branch? Maybe even a 2nd run with eclair.db.sqlite.sync=normal

If you provide a PR on the repo with the exact change, I am happy to run it on our rig.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
pm47: Reverted in 0672cdc.

t-bast (approved):

PR #1874, opened 2021-07-15
Refactor global balance tests

t-bast:
I was OCD-ing over [this comment](https://github.com/ACINQ/eclair/pull/1737/files#r662959421) and it was quick to fix, so here it is :smile: 
I'm moving balance check tests to their own test file instead of making the tests in `NormalStateSpec` more complex than they were.


pm47 (dismissed):
It's a bit awkward because some tests do not use the fixture, but still better than before.

t-bast:
Rebased on master.

pm47 (approved):

PR #1876, opened 2021-07-15
Use psbt to fund and sign transactions

sstone:


t-bast:
While reading this PR, I realized that:

- the separation between `ExtendedBitcoinClient` and `BitcoinCoreWallet` doesn't make sense anymore, so I removed it in #1939 (which should be a pre-requisite for this PR)
- I added the address conversion utilities to bitcoin-lib in https://github.com/ACINQ/bitcoin-lib/pull/55 (with compatibility with future segwit versions)
- handling psbts is a bit messy: we only have a single `Psbt` type which doesn't reflect the state of the psbt, so we don't know for example if it's safe to call `extract().get`: I'll spend some time in `bitcoin-lib` to try to improve this, I think it's important to nail down the right abstraction for psbt before we start using it widely inside eclair

sstone:
Superseded by https://github.com/ACINQ/eclair/pull/2120

PR #1878, opened 2021-07-16
Parallel payment request generation

pm47:
Delegate the payment request generation, signature and db write to a short-lived child actor.

There is small (~5%) gain in performance in `PerformanceIntegrationSpec` because what matters is the db write, and it is not parallelized in WAL mode.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
@@ -17,6 +17,9 @@
 package fr.acinq.eclair.payment.receive
 
 import akka.actor.Actor.Receive
+import akka.actor.typed.Behavior
+import akka.actor.typed.scaladsl.Behaviors
+import akka.actor.typed.scaladsl.adapter.ClassicActorContextOps
 import akka.actor.{ActorContext, ActorRef, PoisonPill, Status}
 import akka.event.{DiagnosticLoggingAdapter, LoggingAdapter}
 import fr.acinq.bitcoin.{ByteVector32, Crypto}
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -52,26 +56,9 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
...
-        val paymentRequest = PaymentRequest(nodeParams.chainHash, amount_opt, paymentHash, nodeParams.privateKey, desc, nodeParams.minFinalExpiryDelta, fallbackAddress_opt, expirySeconds = Some(expirySeconds), extraHops = extraHops, features = features)
-        log.debug("generated payment request={} from amount={}", PaymentRequest.write(paymentRequest), amount_opt)
-        db.addIncomingPayment(paymentRequest, paymentPreimage, paymentType)
-        paymentRequest
-      } match {
-        case Success(paymentRequest) => ctx.sender ! paymentRequest
-        case Failure(exception) => ctx.sender ! Status.Failure(exception)
-      }
+    case receivePayment: ReceivePayment =>
+      val child = ctx.spawn(CreateInvoiceActor(nodeParams), name = UUID.randomUUID().toString)
```
t-bast, line 60: nit: you don't need to specify a name, you can use `spawnAnonymous`

t-bast (approved):

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
pm47: Will fix in a subsequent related PR.

PR #1880, opened 2021-07-16
Fetch incoming payments in parallel

pm47:
This is a simpler approach to completely parallelizing the handling of payments (attempted in branch [parallel-payment-handler](https://github.com/ACINQ/eclair/tree/parallel-payment-handler)), where we simply parallelize the fetch from the database.

The advantage is that it is a much smaller change, that already brings a ~30% performance improvement in performance in `PerformanceIntegrationSpec`.


t-bast (approved):
:rocket:

PR #1885, opened 2021-07-16
Update dependencies and enable fatal warnings

pm47:


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1659,9 +1660,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
           if (d.commitments.announceChannel) {
             // we will re-enable the channel after some delay to prevent flappy updates in case the connection is unstable
-            setTimer(Reconnected.toString, BroadcastChannelUpdate(Reconnected), 10 seconds, repeat = false)
+            startSingleTimer(Reconnected.toString, BroadcastChannelUpdate(Reconnected), 10 seconds)
           } else {
-            // except for private channels where our peer is likely a mobile wallet: they will stay online only for a short period of time,
+            // except for private channels where our peer is likely a mobile wallet: they will stay() online only for a short period of time,
```
t-bast, line 1665: Looks like someone replace `stay` with `stay()` in this file :smile: 

```suggestion
            // except for private channels where our peer is likely a mobile wallet: they will stay online only for a short period of time,
```

t-bast (dismissed):

sstone (approved):

PR #1889, opened 2021-07-19
Update akka-http-json4s dependency

pm47:
Fixes:

```
021-07-19 12:03:13,504 WARN  akka.util.ManifestInfo - You are using version 10.2.4 of Akka HTTP, but it appears you (perhaps indirectly) also depend on older versions of related artifacts. You can solve this by adding an explicit dependency on version 10.2.4 of the [akka-http] artifacts to your project. Here's a complete collection of detected artifacts: (10.2.0, [akka-http]), (10.2.4, [akka-http-core, akka-parsing]). See also: https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html#mixed-versioning-is-not-allowed
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,9 @@
         <developer>
             <id>sstone</id>
         </developer>
+        <developer>
+            <id>t-bast</id>
+        </developer>
     </developers>
 
     <properties>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -58,6 +58,9 @@
         <developer>
             <id>sstone</id>
         </developer>
+        <developer>
+            <id>t-bast</id>
```
t-bast, line 62: simpsons-wouhou

t-bast (approved):

PR #1890, opened 2021-07-19
Set relay fees per node and save them to database

thomash-acinq:
- Fees are set per node instead of per channel (setting different fees for different channels to the same node is most probably an error)
- Fees are saved to a database so that we can keep a trace of historic fees and new channels with a known node use the fee that we set and not the default fee.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -285,17 +285,17 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           context.system.eventStream.publish(ShortChannelIdAssigned(self, normal.channelId, normal.channelUpdate.shortChannelId, None))
 
           // we rebuild a new channel_update with values from the configuration because they may have changed while eclair was down
-          // NB: we don't update the routing fees, because we don't want to overwrite manual changes made with CMD_UPDATE_RELAY_FEE
-          // Since CMD_UPDATE_RELAY_FEE is handled even when being offline, that's the preferred solution to update routing fees
+          val defaultFees = nodeParams.relayParams.defaultFees(data.commitments.announceChannel)
+          val (feeBase, feeProportionalMillionth) = nodeParams.db.relayFees.getFees(remoteNodeId).getOrElse((defaultFees.feeBase, defaultFees.feeProportionalMillionth.toLong))
```
pm47, line 289: This should be factored in `NodeParams` exactly like per-node features and we should use the new `RelayFees` class:

```scala
class NodeParams(...) {
  def featuresFor(nodeId: PublicKey): Features = overrideFeatures.getOrElse(nodeId, features)
  def relayFeesFor(nodeId: PublicKey, announceChannel: Boolean): RelayFees = ???
}
```

Re: the change in behavior (deleted comment), I understand why you went that way, but it means we will have to be careful and always edit `eclair.conf` after calling `./eclair-cli updaterelayfee`.

Another way, closer to the origin behavior, would be to use the hardcoded fee if there are no existing channel, and otherwise keep the existing fee. I think it would be easier from an operational perspective. That way, the order of precedence would be clear (api > eclair.conf), whereas in your proposal it depends on when eclair was last restarted. WDYT?

eclair-core/src/main/scala/fr/acinq/eclair/db/RelayFeesDb.scala
```diff
@@ -0,0 +1,30 @@
...
+import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.eclair.MilliSatoshi
+
+import java.io.Closeable
+
+trait RelayFeesDb extends Closeable {
+
+  def addOrUpdateFees(nodeId: PublicKey, feeBase: MilliSatoshi, feeProportionalMillionths: Long): Unit
+
+  def getFees(nodeId: PublicKey): Option[(MilliSatoshi, Long)]
```
pm47, line 28: This method is unused, would remove.

eclair-core/src/main/scala/fr/acinq/eclair/db/RelayFeesDb.scala
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package fr.acinq.eclair.db
+
+import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.eclair.MilliSatoshi
+
+import java.io.Closeable
+
+trait RelayFeesDb extends Closeable {
```
pm47, line 24: What is the rationale for not adding this to the `AuditDb`? It would fit right in I think.

pm47 (commented):
Nice cleanup.

This PR should be rebased on top of master and use the newly introduced `RelayFees` class instead of tuples.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: > Another way, closer to the origin behavior, would be to use the hardcoded fee if there are no existing channel, and otherwise keep the existing fee. I think it would be easier from an operational perspective. That way, the order of precedence would be clear (api > eclair.conf), whereas in your proposal it depends on when eclair was last restarted. WDYT?

Thinking about this, I see the complication with getting the feerate used by other channels to the same peer. This would have to be handled in the `Peer` class. OTOH, it would make total sense given the intent to make relay fees peer-centric and not channel-centric.

eclair-core/src/main/scala/fr/acinq/eclair/db/RelayFeesDb.scala
thomash-acinq: `AuditDb` should contain information for auditing only and should not be required for running the node, it should be write only. This new database is required for running the node as eclair reads from it.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
thomash-acinq: There is no need to edit `eclair.conf` all the time. The API always gets precedence over `eclair.conf`.

- In case we manually change fees using the API, then it will be recorded in `RelayFeesDb` and it will survive restarts.
- In case we never changed the fees manually, then it will use the default fees from `eclair.conf`. This means that if we edit `eclair.conf`, all channels with the default fees will be updated without the need for manual intervention.

The rationale is that for most nodes we will use the default fees and only for a few important nodes we will manually change it.

pm47: I see now. I thought the per-node setting was in eclair.conf but it was actually retrieved from the db, making the behavior the same as now.

eclair-core/src/main/scala/fr/acinq/eclair/db/RelayFeesDb.scala
pm47: > eclair reads from it

I missed that! In that case, it belongs to the `PeersDb` I think. There is an ambiguity though, because we are also using that same table as an append-only, timestamped, event log. How about a very basic key-value table in `PeersDb` and an event log table in `AuditDb`? Yes there is duplication, but it's consistent with what we have between `PaymentsDb` and `AuditDb`.

We could also just put the relay fee changes in the `channel_event` table, or introduce a new `peer_event` table.

t-bast (commented):
Nice, I'm glad we can get rid of the ugly `initialRelayFees_opt` I added :)
You mention in the PR description that historical fees are kept, but that's not the case in the latest iteration, is it?

eclair-core/src/test/scala/fr/acinq/eclair/db/PeersDbSpec.scala
```diff
@@ -85,4 +86,26 @@ class PeersDbSpec extends AnyFunSuite {
...
+
+      val a = randomKey().publicKey
+      val b = randomKey().publicKey
+
+      assert(db.getFees(a) === None)
+      assert(db.getFees(b) === None)
+      db.addOrUpdateFees(a, RelayFees(1 msat, 123))
+      assert(db.getFees(a) === Some(RelayFees(1 msat, 123)))
+      assert(db.getFees(b) === None)
+      Thread.sleep(2)
```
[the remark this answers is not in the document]
thomash-acinq: It is not needed, I've removed it.

t-bast (approved):
LGTM. Let's wait for a final review from @pm47 before merging as he has more experience actually managing node fees than I do ;)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -112,10 +112,19 @@ object Relayer extends Logging {
     Props(new Relayer(nodeParams, router, register, paymentHandler, initialized))
 
   // @formatter:off
-  case class RelayFees(feeBase: MilliSatoshi, feeProportionalMillionth: Int)
+  case class RelayFees(feeBase: MilliSatoshi, feeProportionalMillionths: Long)
```
pm47, line 115: 👌

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -177,15 +177,18 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
     (appKit.switchboard ? Peer.Disconnect(nodeId)).mapTo[String]
   }
 
-  override def open(nodeId: PublicKey, fundingAmount: Satoshi, pushAmount_opt: Option[MilliSatoshi], fundingFeeratePerByte_opt: Option[FeeratePerByte], initialRelayFees_opt: Option[(MilliSatoshi, Int)], flags_opt: Option[Int], openTimeout_opt: Option[Timeout])(implicit timeout: Timeout): Future[ChannelOpenResponse] = {
+  override def open(nodeId: PublicKey, fundingAmount: Satoshi, pushAmount_opt: Option[MilliSatoshi], fundingFeeratePerByte_opt: Option[FeeratePerByte], relayFees_opt: Option[(MilliSatoshi, Int)], flags_opt: Option[Int], openTimeout_opt: Option[Timeout])(implicit timeout: Timeout): Future[ChannelOpenResponse] = {
```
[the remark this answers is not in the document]
pm47: I would completely remove the ability to set the relay fees in this command, to remove this ambiguity. Instead we can call `updaterelayfee` at any time (before or after) the channel is created which will have the same effect IIUC.

pm47 (commented):
Like @t-bast I'm also wondering why you have dropped the historical fees, do you feel like it's not worth it anymore? Can also be added later.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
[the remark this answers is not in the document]
t-bast: I thought about this and was bothered by the case where you open your first channel with a given peer, and don't want to use the default relay fees, but the way this has been implemented you can call `update_relay_fees` before having channel with a peer, which is great!

So I agree, I'm in favor of completely removing this option from `open` and we'll document that you can call `updaterelayfees` before opening the channel to achieve that result.

eclair-core/src/main/scala/fr/acinq/eclair/db/RelayFeesDb.scala
thomash-acinq: Done, I've added a `channel_updates` table in `AuditDb`.

thomash-acinq:
I've added the historical fees back. They're now in `AuditDb`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -285,24 +285,24 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+            fees.feeBase,
+            fees.feeProportionalMillionths,
             normal.commitments.capacity.toMilliSatoshi,
             enable = Announcements.isEnabled(normal.channelUpdate.channelFlags))
           val channelUpdate1 = if (Announcements.areSame(candidateChannelUpdate, normal.channelUpdate)) {
             // if there was no configuration change we keep the existing channel update
             normal.channelUpdate
           } else {
             log.info("refreshing channel_update due to configuration changes old={} new={}", normal.channelUpdate, candidateChannelUpdate)
+            nodeParams.db.audit.addChannelUpdate(data.channelId, candidateChannelUpdate)
```
pm47, line 305: The usual pattern is to rely on events to update the `audit` db, and make the calls from within the `DbEventHandler`. Main reason was to minimize performance impacts. Here there is no real perf concern since relay fees are not updated often (except if we update all channels at once), but it's still better to have as little blocking db calls as possible in the channel, and it would still be better for consistency.

In any case I would factor it in the `onTransition`. We actually already emit a `LocalChannelUpdate` event (see https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1881-L1904). However the event is also emitted if we simply enable/disable the channel. The `DbEventHandler` could listen to that event and only store the update in the audit db when the channel is enabled.

This would also mean that if the channel is being created, we would store the event in the audit db when the channel is ready and not before. Same if the channel is offline. I think that's ok.

thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -94,18 +102,23 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
...
+          }
+          if (v < 6) {
+            migration56(statement)
+          }
+          if (v < 7) {
+            migration67(statement)
+          }
+          if (v < 8) {
+            migration78(statement)
+          }
```
pm47, line 121: Totally understand the need to improve this, but let's do that in a separate PR and fix all db files at once.

Otherwise LGTM

thomash-acinq: reverted that part

pm47 (approved):

eclair-core/src/main/scala/fr/acinq/eclair/db/DbEventHandler.scala
```diff
@@ -116,6 +118,11 @@ class DbEventHandler(nodeParams: NodeParams) extends Actor with ActorLogging {
       auditDb.add(ChannelEvent(e.channelId, e.commitments.remoteParams.nodeId, e.commitments.commitInput.txOut.amount, e.commitments.localParams.isFunder, !e.commitments.announceChannel, event))
       channelsDb.updateChannelMeta(e.channelId, event)
 
+    case u: LocalChannelUpdate =>
```
t-bast, line 121: NB: this will trigger a lot of noise, we resend this every time a peer reconnects...but maybe it's ok since it doesn't contain that much data

engenegr:
Are you sure you haven't broken old way of setting fees? @t-bast 
Because this is what i get in response to my old command with channel ids
```
The form field 'nodeId(s)' was malformed:
Must specify nodeId or nodeIds (but not both)
```

engenegr:
That's for sure, you have broke old interface. 
Then please maybe update documentation.

pm47:
> That's for sure, you have broke old interface.

There has been a backward-incompatible change, yes. Need to use `nodeId(s)` instead of `channelId(s)`. 

> Then please maybe update documentation.

I think @t-bast does that right before release. In the meantime:

```shell
curl -u :<eclair_api_password> -X POST -F nodeId=<node> \
     -F feeBaseMsat=<feebase> -F feeProportionalMillionths=<feeproportional> \
     "http://localhost:8080/updaterelayfee"
```

t-bast:
> Then please maybe update documentation.

The documentation always reflects the latest release.
If you're running `master`, you can't expect the documentation to be up-to-date (otherwise we would need to have different documentation for every commit, which is way too much work).

I'm trying to improve that with #1951 so that release notes are written as `master` progresses.
That means that when you're running `master`, you can always check the draft release notes for the next release to see what has changed since the latest version.

PR #1892, opened 2021-07-19
Fix pay-to-open below minimum when using MPP

pm47:
_This is a PR to branch android-phoenix_

Compare the *sum* of all pay-to-open parts against the min pay-to-open
amount.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartPaymentFSMSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartPaymentFSMSpec.scala
@@ -22,7 +22,7 @@ import akka.testkit.{TestActorRef, TestProbe}
 import fr.acinq.bitcoin.{Block, ByteVector32}
 import fr.acinq.eclair.payment.receive.MultiPartPaymentFSM
 import fr.acinq.eclair.payment.receive.MultiPartPaymentFSM._
-import fr.acinq.eclair.wire.{IncorrectOrUnknownPaymentDetails, PayToOpenRequest, UnknownNextPeer, UpdateAddHtlc}
+import fr.acinq.eclair.wire.{IncorrectOrUnknownPaymentDetails, PayToOpenRequest, UpdateAddHtlc}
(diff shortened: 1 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartPaymentFSMSpec.scala
```diff
@@ -167,6 +181,22 @@ class MultiPartPaymentFSMSpec extends TestKitBaseClass with AnyFunSuiteLike {
     f.eventListener.expectNoMsg(50 millis)
   }
 
+  test("fulfil all if total pay-to-open is above minimum") {
```
t-bast, line 184: nit:
```suggestion
  test("fulfill all if total pay-to-open is above minimum") {
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartPaymentFSMSpec.scala
```diff
@@ -110,6 +111,19 @@ class MultiPartPaymentFSMSpec extends TestKitBaseClass with AnyFunSuiteLike {
     f.eventListener.expectNoMsg(50 millis)
   }
 
+  test("fail all if total pay-to-open is below minimum") {
+    val f = createFixture(250 millis, 20000000 msat)
```
t-bast, line 115: It would be useful to always have an `assert` that the `minPayToOpenAmount` is higher than `4000000 msat` but lower than `20000000 msat` since it's not explicitly set by this test (to make sure that if we change the internals of `createPayToOpenPart` we can't regress here without detecting it).

pm47: Done in c50d7f7.

eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartPaymentFSMSpec.scala
pm47: Done in c50d7f7.

t-bast (approved):

PR #1893, opened 2021-07-21
Different default relay fees for announced and unannounced channels

thomash-acinq:
Split the settings for the default relay fee in two:
- fees for announced channels
- fees for unannounced channels
- minimum fee for trampoline relays (we don't even try to relay trampoline payments with lower fees)

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -90,8 +90,23 @@ eclair {
   max-block-processing-delay = 30 seconds // we add a random delay before processing blocks, capped at this value, to prevent herd effect
   max-tx-publish-retry-delay = 60 seconds // we add a random delay before retrying failed transaction publication
 
-  fee-base-msat = 1000
-  fee-proportional-millionths = 100 // fee charged per transferred satoshi in millionths of a satoshi (100 = 0.01%)
+  relay.fees {
+      // Fees for public channels
+      public-channels {
+        fee-base-msat = 1000
+        fee-proportional-millionths = 200 // fee charged per transferred satoshi in millionths of a satoshi (200 = 0.02%)
+      }
+      // Fees for private channels
+      private-channels {
+        fee-base-msat = 1000
+        fee-proportional-millionths = 100
+      }
+      // Minimum fees for trampoline relays
+      min-trampoline {
+        fee-base-msat = 1000
+        fee-proportional-millionths = 100
+      }
+   }
 
   on-chain-fees {
     min-feerate = 1 // minimum feerate in satoshis per byte
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -26,6 +26,7 @@ import fr.acinq.eclair.crypto.Noise.KeyPair
 import fr.acinq.eclair.crypto.keymanager.{ChannelKeyManager, NodeKeyManager}
 import fr.acinq.eclair.db._
 import fr.acinq.eclair.io.PeerConnection
+import fr.acinq.eclair.payment.relay.Relayer.{RelayFees, RelayParams}
 import fr.acinq.eclair.router.Router.RouterConf
 import fr.acinq.eclair.tor.Socks5ProxyParams
 import fr.acinq.eclair.wire.protocol.{Color, EncodingType, NodeAddress}
@@ -69,8 +70,7 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
                       toRemoteDelay: CltvExpiryDelta,
                       maxToLocalDelay: CltvExpiryDelta,
                       minDepthBlocks: Int,
-                      feeBase: MilliSatoshi,
-                      feeProportionalMillionth: Int,
+                      relayParams: RelayParams,
                       reserveToFundingRatio: Double,
                       maxReserveToFundingRatio: Double,
                       db: Databases,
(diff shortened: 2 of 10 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -90,8 +90,17 @@ eclair {
...
+  fee-base-msat-public = 1000
+  fee-proportional-millionths-public = 200 // fee charged per transferred satoshi in millionths of a satoshi (200 = 0.02%)
+
+  // Fees for private channels
+  fee-base-msat-private = 1000
+  fee-proportional-millionths-private = 100
+
+  // Minimum fees for trampoline relays
+  min-trampoline-fee-base-msat = 1000
+  min-trampoline-fee-proportional-millionths = 100
```
pm47, line 103: I would go with a more hierarchical organization. We have been going in that direction with `router`, `peer-connection`, `db` and I'd like to move channel-related parameters to a `channel` section at some point.

Those fee params are kind of in-between `channel` and `relay` but since there are trampoline-related settings, it makes sense to put them under `relay` I think.

```suggestion
  relay.fees {
    // Fees for public channels
    public-channels {
      fee-base-msat = 1000
      fee-proportional-millionths = 200 // fee charged per transferred satoshi in millionths of a satoshi (200 = 0.02%)
    }
    // Fees for private channels
    private-channels {
      fee-base-msat = 1000
      fee-proportional-millionths = 100
    }
    // Minimum fees for trampoline relays
    min-trampoline {
      fee-base-msat = 1000
      fee-proportional-millionths = 100
    }
```

This structure makes it possible to factorize parsing too.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -69,8 +69,12 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
...
                       maxToLocalDelay: CltvExpiryDelta,
                       minDepthBlocks: Int,
-                      feeBase: MilliSatoshi,
-                      feeProportionalMillionth: Int,
+                      feeBasePublic: MilliSatoshi,
+                      feeProportionalMillionthPublic: Int,
+                      feeBasePrivate: MilliSatoshi,
+                      feeProportionalMillionthPrivate: Int,
+                      minTrampolineFeeBase: MilliSatoshi,
+                      minTrampolineFeeProportionalMillionth: Int,
```
pm47, line 77: Let's define some classes in the `Relayer` companion object:

```scala
case class RelayFees(feeBase: MilliSatoshi, feeProportionalMillionth: Int)
case class RelayParams(publicChannelFees: RelayFees,
                       privateChannelFees: RelayFees,
                       minTrampolineFees: RelayFees)
``` 

```suggestion
                      relayParams: RelayParams,
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -663,7 +663,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       blockchain ! WatchFundingDeeplyBuried(self, commitments.commitInput.outPoint.txid, ANNOUNCEMENTS_MINCONF)
       context.system.eventStream.publish(ShortChannelIdAssigned(self, commitments.channelId, shortChannelId, None))
       // we create a channel_update early so that we can use it to send payments through this channel, but it won't be propagated to other nodes since the channel is not yet announced
-      val (feeBase, feeProportionalMillionths) = initialRelayFees_opt.getOrElse((nodeParams.feeBase, nodeParams.feeProportionalMillionth))
+      val defaultFees =
+        if (commitments.announceChannel)
+          (nodeParams.feeBasePublic, nodeParams.feeProportionalMillionthPublic)
+        else
+          (nodeParams.feeBasePrivate, nodeParams.feeProportionalMillionthPrivate)
```
pm47, line 670: As a convention we always add curly brackets on `if`s.

pm47 (commented):
Concept ACK, some nits

thomash-acinq:
Relay fees are now grouped in their own struct

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -349,8 +352,11 @@ object NodeParams extends Logging {
       toRemoteDelay = CltvExpiryDelta(config.getInt("to-remote-delay-blocks")),
       maxToLocalDelay = CltvExpiryDelta(config.getInt("max-to-local-delay-blocks")),
       minDepthBlocks = config.getInt("mindepth-blocks"),
-      feeBase = feeBase,
-      feeProportionalMillionth = config.getInt("fee-proportional-millionths"),
+      relayParams = RelayParams(
+        getRelayFees(config.getConfig("relay.fees.public-channels")),
+        getRelayFees(config.getConfig("relay.fees.private-channels")),
+        getRelayFees(config.getConfig("relay.fees.min-trampoline")),
```
pm47, line 358: When types are the same I like to explicitly name the arguments to prevent future footguns:

```suggestion
        publicChannelFees = getRelayFees(config.getConfig("relay.fees.public-channels")),
        privateChannelFees = getRelayFees(config.getConfig("relay.fees.private-channels")),
        minTrampolineFees = getRelayFees(config.getConfig("relay.fees.min-trampoline")),
```

pm47 (commented):
One last nit, otherwise LGTM.

In a later iteration we can add a per-node override, like we do for features and on-chain fee tolerance.

pm47 (approved):

thomash-acinq:
> In a later iteration we can add a per-node override, like we do for features and on-chain fee tolerance.

That's the purpose of #1890 but the overrides are stored in a database instead of a configuration file. It allows changing the fees without restarting eclair and we keep track of old fees.

PR #1894, opened 2021-08-03
Balance: do not deduplicate identical htlcs amounts

pm47:
HTLCs are stored in a `Set`, and doing `htlcs.map(_.amount).sum` will
deduplicate identical amounts, resulting in an erroneous balance
calculation. The `Set` must be converted to a `List` first.

Identical htlcs amount are quite frequent due to splitting for AMP.


eclair-core/src/main/scala/fr/acinq/eclair/balance/CheckBalance.scala
```diff
@@ -1,7 +1,8 @@
 package fr.acinq.eclair.balance
 
 import com.softwaremill.quicklens._
-import fr.acinq.bitcoin.{Btc, ByteVector32, SatoshiLong}
+import fr.acinq.bitcoin.{Btc, ByteVector32, Satoshi, SatoshiLong}
+import fr.acinq.eclair.{MilliSatoshi, _}
```
t-bast, line 5: nit:
```suggestion
import fr.acinq.eclair._
```

t-bast (approved):

t-bast (approved):

PR #1897, opened 2021-08-08
Backup JSON API call

rorp:
This PR introduces `backupdatabase` JSON API endpoint and CLI command. It allows to perform scheduled database backups.


t-bast:
Can you elaborate on why you feel this feature is useful?

rorp:
- The database contains a various kind of data, but it get's backed up only when a channel is changed.
- Predictable backup timing. You don't have to wait for a channel update to perform a backup.
- You don't need to pay (one way or another) if you want to force backup.
- Helps in development/testing of backup notification scripts.

rorp:
Personally, I need it because of the last point.

t-bast:
I'm still not convinced that it's useful...

> The database contains various kinds of data, but it gets backed up only when a channel is changed.

But only the channel data really needs to be backed up.
The rest can be recreated automatically from the network.

> Predictable backup timing. You don't have to wait for a channel update to perform a backup.

Same as previous point, you only need to backup channel data, so you don't need predictable backup timing, it should be backed up when channels change.

> You don't need to pay (one way or another) if you want to force a backup.

Same.

> Helps in development/testing of backup notification scripts.

This isn't a good enough IMHO reason to add code that needs to be maintained on `master`, a test branch is enough.

pm47:
@rorp aren't you using postgres? If so, our preferred backup setup is a frequent pg_dump on a read replica. It is far superior to what we have for sqlite, because it allows for a very up-to-date backup, with no impact on the main database. The fact that the backup is done on an asynchronously replicated copy of the db is at odds with a synchronous api call.

PR #1900, opened 2021-08-09
Lower minimum remote dust limit

t-bast:
Implement https://github.com/lightningnetwork/lightning-rfc/pull/894

We can safely accept dust limits as small as 354 sats, as long as we explicitly handle the case where a closing tx is generated with an output below dust (when that happens, we force-close to get our funds back).

The spec PR received broad agreement during yesterday's spec meeting, so we can safely merge this change now.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -518,12 +518,34 @@ object Helpers {
         Left(InvalidCloseFee(commitments.channelId, remoteClosingFee))
       } else {
         val (closingTx, closingSigned) = makeClosingTx(keyManager, commitments, localScriptPubkey, remoteScriptPubkey, ClosingFees(remoteClosingFee, remoteClosingFee, remoteClosingFee))
-        val signedClosingTx = Transactions.addSigs(closingTx, keyManager.fundingPublicKey(commitments.localParams.fundingKeyPath).publicKey, remoteParams.fundingPubKey, closingSigned.signature, remoteClosingSig)
-        Transactions.checkSpendable(signedClosingTx) match {
-          case Success(_) => Right(signedClosingTx, closingSigned)
-          case _ => Left(InvalidCloseSignature(commitments.channelId, signedClosingTx.tx))
+        if (checkClosingDustAmounts(closingTx)) {
+          val signedClosingTx = Transactions.addSigs(closingTx, keyManager.fundingPublicKey(commitments.localParams.fundingKeyPath).publicKey, remoteParams.fundingPubKey, closingSigned.signature, remoteClosingSig)
+          Transactions.checkSpendable(signedClosingTx) match {
+            case Success(_) => Right(signedClosingTx, closingSigned)
+            case _ => Left(InvalidCloseSignature(commitments.channelId, signedClosingTx.tx))
+          }
+        } else {
+          Left(InvalidCloseAmountBelowDust(commitments.channelId, closingTx.tx))
         }
       }
+    }
+
+    /**
+     * Check that all closing outputs are above bitcoin's dust limit for their script type, otherwise there is a risk
+     * that the closing transaction will not be relayed to miners' mempool and will not confirm.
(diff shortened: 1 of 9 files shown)


sstone:
If our peer uses non-segwit shutdown scripts and we lower the dust limit below what is acceptable for non-segwit outputs (546 instead of 330), then we may end up creating mutual close transactions with dust outputs ?

t-bast:
I started a discussion on the spec regarding this issue: https://github.com/lightningnetwork/lightning-rfc/issues/905
Let's see what direction we want to go in and I'll update this PR accordingly.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -518,14 +518,36 @@ object Helpers {
...
+     */
+    def checkClosingDustAmounts(closingTx: ClosingTx): Boolean = {
+      closingTx.tx.txOut.forall(txOut => {
+        Try(Script.parse(txOut.publicKeyScript)) match {
+          case Success(OP_DUP :: OP_HASH160 :: OP_PUSHDATA(pubkeyHash, _) :: OP_EQUALVERIFY :: OP_CHECKSIG :: Nil) if pubkeyHash.size == 20 => txOut.amount >= 546.sat
+          case Success(OP_HASH160 :: OP_PUSHDATA(scriptHash, _) :: OP_EQUAL :: Nil) if scriptHash.size == 20 => txOut.amount >= 540.sat
+          case Success(OP_0 :: OP_PUSHDATA(pubkeyHash, _) :: Nil) if pubkeyHash.size == 20 => txOut.amount >= 294.sat
+          case Success(OP_0 :: OP_PUSHDATA(scriptHash, _) :: Nil) if scriptHash.size == 32 => txOut.amount >= 330.sat
+          case Success((OP_1 | OP_2 | OP_3 | OP_4 | OP_5 | OP_6 | OP_7 | OP_8 | OP_9 | OP_10 | OP_11 | OP_12 | OP_13 | OP_14 | OP_15 | OP_16) :: OP_PUSHDATA(program, _) :: Nil) if 2 <= program.length && program.length <= 40 => txOut.amount >= 354.sat
+          case _ => txOut.amount >= 546.sat
```
pm47, line 546: Trusting you on this...

pm47 (commented):
I was expecting a check on the script with a force close branch here: https://github.com/ACINQ/eclair/blob/ca2c87f1990bf98a07fa2bc2a3a2cc91db2d9444/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L886 but you are doing it differently. You let the mutual close process reach the point where a sig is exchanged, an only then do we check amounts and maybe force-close. Is that correct?

I was also expecting a check there for the case where we initiate the mutual close:
https://github.com/ACINQ/eclair/blob/ca2c87f1990bf98a07fa2bc2a3a2cc91db2d9444/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L869 and I understand why you may want to fail earlier in the API to return a proper error, but that doesn't remove the need for proper handling of `CMD_CLOSE` to prevent unnecessary force closes if used directly (e.g. from a plugin).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Heh, this will be in the spec and has been independently verified, hopefully we're good!

t-bast:
> You let the mutual close process reach the point where a sig is exchanged, an only then do we check amounts and maybe force-close. Is that correct?

Yes, that was the least invasive option, because we need to have decided on a fee before we can force-close (since the fee impacts the output amounts).

> I was also expecting a check there for the case where we initiate the mutual close:

I don't think it's worth explicitly rejecting here.
In almost all cases, it's actually ok if non-segwit is used.
The edge case where we need to force-close is close to impossible to get into (see details here: https://github.com/lightningnetwork/lightning-rfc/issues/905).

But this was a good opportunity to promote segwit-only at the API level to encourage our users to migrate (if not already done), because there's just no reason to use non-segwit today.

pm47 (approved):

PR #1901, opened 2021-08-11
Refactor db migration

thomash-acinq:
Simplify the migration by not duplicating code. Should be less error-prone when adding new migration.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/db/AuditDbSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/db/AuditDbSpec.scala
@@ -228,7 +228,7 @@ class AuditDbSpec extends AnyFunSuite {
           statement.executeUpdate()
         }
       },
-      dbName = "audit",
+      dbName = SqliteAuditDb.DB_NAME,
       targetVersion = SqliteAuditDb.CURRENT_VERSION,
       postCheck = connection => {
         // existing rows in the 'sent' table will use id=00000000-0000-0000-0000-000000000000 as default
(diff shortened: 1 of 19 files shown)


t-bast (dismissed):

pm47:
Another thing that would make db updates a little less of a hassle would be to move `val CURRENT_VERSION = 8` to the companion object. This would save updating test files at multiple place.

I may have another idea for the migration process, will let you know.

pm47 (approved):

thomash-acinq:
I've moved `CURRENT_VERSION` to the companion object.

eclair-core/src/test/scala/fr/acinq/eclair/db/AuditDbSpec.scala
```diff
@@ -229,7 +229,7 @@ class AuditDbSpec extends AnyFunSuite {
         }
       },
       dbName = "audit",
-      targetVersion = 6,
+      targetVersion = SqliteAuditDb.CURRENT_VERSION,
```
pm47, line 232: The line above could also be updated to `SqliteAuditDb.DB_NAME`.

pm47 (commented):
The concern I have now is that nothing prevents a database file to mistakenly import the wrong db object.

thomash-acinq:
I don't get your concern about importing the wrong db object, can you give an example?

pm47:
> I don't get your concern about importing the wrong db object, can you give an example?

If we mistakenly import e.g. `import PgAuditDb._` from `PgChannels`, then the migration will be completely broken. Tests should catch that though.

pm47 (approved):

t-bast (approved):
:+1:

PR #1904, opened 2021-08-12
Do not log channel updates in AuditDb if they haven't changed

thomash-acinq:


t-bast (commented):
While this solution prevents the DB from growing too much, we still call the DB every time and now potentially call it twice (read then insert), which I'd like to avoid (am I being too extreme about minimizing DB calls?).

What about making the same kind of check but inside the `DbEventHandler` actor, against a local cache of each channel's last `channel_update`? Since we need to keep less than 100 bytes of data per `channel_update` in memory to check equality, even assuming 100 000 channels it would only use 10Mb of RAM which we can easily afford.

The way I would do it would be by adding a private map channelId -> lastChannelUpdate inside the `DbEventHandler` actor and check it before calling `auditDb.addChannelUpdate(u)`. I don't think we need to initialize this map by reading all `channel_update`s from the DB when the actor starts, it's ok if we potentially duplicate `channel_update`s once when the node restarts. Ideally this map wouldn't contain the whole `channel_update` but just the fields we use to check whether the new `channel_update` is worth storing (to save some space).

WDYT?

thomash-acinq:
> > What about making the same kind of check but inside the DbEventHandler actor, against a local cache of each channel's last channel_update?
> 
> We could simply modify the event to include the previous `channel_update`. We have access to both in the `onTransition` and already do comparisons like `d1.channelUpdate != d2.channelUpdate`.

I didn't realize that we had access to the previous channel update, that's a much better solution.

thomash-acinq:
Done in https://github.com/ACINQ/eclair/pull/1906

PR #1905, opened 2021-08-12
Better handling of remote commit confirmation in TxPublisher

### LINKED ISSUE #1898: Abandon anchor tx when competing commit tx has been confirmed
We use `fundrawtransaction` to create our `claim-anchor` transactions, which means `bitcoind` will not check whether our parent (the commit tx) exists. If we try to claim the anchor of our local commit but the remote commit gets confirmed first, our claim anchor tx will stay in our `bitcoind` wallet (hoping that the parent tx will eventually appear in the mempool or in the blockchain).

When we detect that the other commit tx has won the race, we should call `bitcoind`'s `abandontransaction` RPC to clean up the anchor tx that will never complete.

t-bast:
This PR is split in two independent commits.

In the first commit, we fix a (harmless) issue where we kept spawning `TxPublisher` actors to spend anchor outputs even when the commit tx was confirmed if the feerate was already high enough. This was spamming the logs for no reason (and spawning actors that immediately stopped themselves), we should instead check the status of the commit tx as the first precondition to spending anchor outputs.

In the second commit, we call `bitcoind`'s `abandontransaction` when 2nd-stage txs are evicted from the mempool because their parent is replaced (e.g. the remote commit tx is confirmed instead of our local commit tx). This ensures the `bitcoind` wallet will not keep these obsolete transactions in its DB and reserve their inputs for no good reason.

Fixes #1898

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
@@ -24,7 +24,7 @@ import fr.acinq.eclair.blockchain.bitcoind.ZmqWatcher
 import fr.acinq.eclair.blockchain.bitcoind.rpc.ExtendedBitcoinClient
 import fr.acinq.eclair.blockchain.bitcoind.rpc.ExtendedBitcoinClient.FundTransactionOptions
 import fr.acinq.eclair.blockchain.fee.FeeratePerKw
-import fr.acinq.eclair.channel.publish.TxPublisher.TxPublishLogContext
+import fr.acinq.eclair.channel.publish.TxPublisher.{TxPublishLogContext, TxRejectedReason}
 import fr.acinq.eclair.channel.publish.TxTimeLocksMonitor.CheckTx
 import fr.acinq.eclair.channel.{Commitments, HtlcTxAndRemoteSig}
 import fr.acinq.eclair.transactions.Transactions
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
```diff
@@ -329,6 +329,14 @@ private class ReplaceableTxPublisher(nodeParams: NodeParams,
...
       case WrappedTxResult(MempoolTxMonitor.TxConfirmed) => sendResult(replyTo, TxPublisher.TxConfirmed(cmd, tx))
       case WrappedTxResult(MempoolTxMonitor.TxRejected(reason)) =>
+        reason match {
+          case TxRejectedReason.WalletInputGone =>
+            // The transaction now has an unknown input from bitcoind's point of view, so it will keep it in the wallet in
+            // case that input appears later in the mempool or the blockchain. In our case, we know it won't happen so we
+            // abandon that transaction and will retry with a different set of inputs (if it still makes sense to publish).
+            bitcoinClient.abandonTransaction(tx.txid)
+          case _ => // nothing to do
+        }
```
pm47, line 339: Shouldn't there be a check here that the remote commit has been published, instead of only relying on the error code from bitcoind?

t-bast: I don't think so, it's more general than that.
This case can also happen for htlc txs that spend an unconfirmed output of another htlc tx which is then replaced.
I think that in all cases where one of our wallet input disappears like this, we should abandon the transaction to free the inputs, and we'll retry later if needed anyway so there's no risk that we forget to claim an output.

pm47 (approved):

PR #1906, opened 2021-08-12
Don't add channel update to auditDb if it hasn't changed

thomash-acinq:


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1871,20 +1871,25 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         case _ => ()
       }
 
+      val previousChannelUpdate = stateData match {
```
pm47, line 1874: The `_opt` suffix is a convention we have for options.

```suggestion
      val previousChannelUpdate_opt = stateData match {
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
```diff
@@ -38,7 +38,7 @@ case class ChannelIdAssigned(channel: ActorRef, remoteNodeId: PublicKey, tempora
 
 case class ShortChannelIdAssigned(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, previousShortChannelId: Option[ShortChannelId]) extends ChannelEvent
 
-case class LocalChannelUpdate(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, channelAnnouncement_opt: Option[ChannelAnnouncement], channelUpdate: ChannelUpdate, commitments: AbstractCommitments) extends ChannelEvent
+case class LocalChannelUpdate(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, channelAnnouncement_opt: Option[ChannelAnnouncement], channelUpdate: ChannelUpdate, commitments: AbstractCommitments, previousChannelUpdate: Option[ChannelUpdate] = None) extends ChannelEvent
```
pm47, line 41: I would have inserted the new field before, and not used a default (those are treacherous):

```suggestion
case class LocalChannelUpdate(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, channelAnnouncement_opt: Option[ChannelAnnouncement], channelUpdate: ChannelUpdate, previousChannelUpdate_opt: Option[ChannelUpdate], commitments: AbstractCommitments) extends ChannelEvent
```

eclair-core/src/main/scala/fr/acinq/eclair/db/DbEventHandler.scala
```diff
@@ -119,8 +118,14 @@ class DbEventHandler(nodeParams: NodeParams) extends Actor with ActorLogging {
...
-      if (Announcements.isEnabled(u.channelUpdate.channelFlags)) {
-        auditDb.addChannelUpdate(u)
+      u.previousChannelUpdate match {
+        case Some(previous) if
+          u.channelUpdate.feeBaseMsat == previous.feeBaseMsat &&
+          u.channelUpdate.feeProportionalMillionths == previous.feeProportionalMillionths &&
+          u.channelUpdate.cltvExpiryDelta == previous.cltvExpiryDelta &&
+          u.channelUpdate.htlcMinimumMsat == previous.htlcMinimumMsat &&
+          u.channelUpdate.htlcMaximumMsat == previous.htlcMaximumMsat => ()
+        case _ => auditDb.addChannelUpdate(u)
```
pm47, line 128: Have you seen `Announcements.areSame`? It's not exactly what you need here (flags should be ignored) though.

t-bast (commented):
Great, that's much better! Nothing to add to the existing comments on my side.

eclair-core/src/main/scala/fr/acinq/eclair/db/DbEventHandler.scala
thomash-acinq: Yes I've seen it but it didn't match exactly what I needed and I thought it would be better to explicitly list the fields we want to compare.

t-bast (approved):

pm47 (approved):

PR #1908, opened 2021-08-13
Nicer feerate string representation

t-bast:
Override `toString()` for the various feerate classes.


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
```diff
@@ -43,6 +45,7 @@ case class FeeratePerKB(feerate: Satoshi) extends Ordered[FeeratePerKB] {
   def max(other: FeeratePerKB): FeeratePerKB = if (this > other) this else other
   def min(other: FeeratePerKB): FeeratePerKB = if (this < other) this else other
   def toLong: Long = feerate.toLong
+  override def toString: String = s"$feerate/kb"
```
pm47, line 48:
```suggestion
  override def toString: String = s"$feerate/kB"
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1908/commits/04abe683d66077ad8414969bf6e8ea3a61c180e2

pm47 (approved):

PR #1910, opened 2021-08-16
Switch ZMQ to block hash and improve resiliency

### LINKED ISSUE #1895: ZMQ died silently
ZMQ silently died again without signal about disconnecting with node.
Last signal I got on my node was from
```
2021-08-02 10:27:36,964 INFO  f.a.e.b.b.zmq.ZMQActor - connected to tcp://192.168.2.7:29000
```
then I used it for a while and on 6th Aug node didn't signal about ZMQ disruptions while Watchdog indicated eclipsing.

```
2021-08-06 12:39:20,703 WARN  f.a.e.b.w.BlockchainWatchdog - blockcypher.com: we are 10 blocks late: we may be eclipsed from the bitcoin network
2021-08-06 12:54:10,281 WARN  f.a.e.b.w.BlockchainWatchdog - blockstream.info: we are 13 blocks late: we may be eclipsed from the bitcoin network
2021-08-06 12:54:10,998 WARN  f.a.e.b.w.BlockchainWatchdog - mempool.space: we are 13 blocks late: we may be eclipsed from the bitcoin network
2021-08-06 12:54:11,860 WARN  f.a.e.b.w.BlockchainWatchdog - bitcoinheaders.net: we are 9 blocks late: we may be eclipsed from the bitcoin network
2021-08-06 12:54:20,448 WARN  f.a.e.b.w.BlockchainWatchdog - blockcypher.com: we are 13 blocks late: we may be eclipsed from the bitcoin network
```
Restarting BITCOIND didn't help that's why I think ZMQ actor died.
version `0.6.2-SNAPSHOT-cafaeed`

t-bast:
This PR contains 3 commits that can be reviewed separately:

- the first one disables the ZMQ high watermark to ensure no messages are dropped (as bitcoind's sample client does)
- the second one switches to block hash instead of raw block to save bandwidth
- the last one ensure we check the block count every 15 minutes in case we haven't received anything on the zmq socket

Each commit contains more detailed descriptions.

NB: this requires a change to `bitcoin.conf` to send block hash events instead of raw blocks over ZMQ. That will need to be properly documented in the next eclair release to ensure users correctly reconfigure their bitcoin node.

Closes #1895


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/zmq/ZMQActor.scala
```diff
@@ -116,7 +117,7 @@ object ZMQActor {
   // @formatter:on
 
   object Topics {
-    val RawBlock: String = "rawblock"
+    val HashBlock: String = "hashblock"
```
pm47, line 120: This feels really weird, should probably be:
```suggestion
    val BlockHash: String = "blockhash"
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -161,11 +164,13 @@ object ZmqWatcher {
...
   def apply(chainHash: ByteVector32, blockCount: AtomicLong, client: ExtendedBitcoinClient): Behavior[Command] =
     Behaviors.setup { context =>
-      context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[NewBlock](b => ProcessNewBlock(b.block)))
+      context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[NewBlock](b => ProcessNewBlock(b.blockHash)))
       context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[NewTransaction](t => ProcessNewTransaction(t.tx)))
       Behaviors.withTimers { timers =>
         // we initialize block count
         timers.startSingleTimer(TickNewBlock, TickNewBlock, 1 second)
+        // we start a timer in case we don't receive ZMQ block events
+        timers.startSingleTimer(TickBlockTimeout, TickBlockTimeout, blockTimeout)
```
pm47, line 173: Nit (does the same):

```suggestion
        timers.startSingleTimer(TickNewBlock, 1 second)
        // we start a timer in case we don't receive ZMQ block events
        timers.startSingleTimer(TickBlockTimeout, blockTimeout)
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -229,19 +234,32 @@ private class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client
...
 
-      case ProcessNewBlock(block) =>
-        log.debug("received blockid={}", block.blockId)
+      case ProcessNewBlock(blockHash) =>
+        log.debug("received blockhash={}", blockHash)
         log.debug("scheduling a new task to check on tx confirmations")
+        // we have received a block, so we can reset the block timeout timer
+        timers.startSingleTimer(TickBlockTimeout, TickBlockTimeout, blockTimeout)
         // we do this to avoid herd effects in testing when generating a lots of blocks in a row
         timers.startSingleTimer(TickNewBlock, TickNewBlock, 2 seconds)
```
pm47, line 243:
```suggestion
        timers.startSingleTimer(TickBlockTimeout, blockTimeout)
        // we do this to avoid herd effects in testing when generating a lots of blocks in a row
        timers.startSingleTimer(TickNewBlock, 2 seconds)
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -229,19 +234,32 @@ private class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client
...
         Behaviors.same
 
+      case TickBlockTimeout =>
+        // we haven't received a block in a while, we check whether we're behind and restart the timer.
+        timers.startSingleTimer(TickBlockTimeout, TickBlockTimeout, blockTimeout)
```
pm47, line 248:
```suggestion
        timers.startSingleTimer(TickBlockTimeout, blockTimeout)
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -229,19 +234,32 @@ private class ZmqWatcher(chainHash: ByteVector32, blockCount: AtomicLong, client
...
+        val currentBlockCount = blockCount.get()
+        client.getBlockCount.map { count =>
+          if (count > currentBlockCount) {
+            context.log.warn("new block wasn't received via ZMQ, you should verify your bitcoind node")
+            context.self ! TickNewBlock
+          }
+        }
```
pm47, line 255: A simpler logic would have been to just send `TickNewBlock` unconditionnally.

pm47 (commented):
Nice optimization! Saves us from parsing blocks too.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/zmq/ZMQActor.scala
pm47: Wait, the official topic name is actually `hashblock`, nevermind.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: That would be more expensive, because we would go through the `checkConfirmed` code for every `WatchConfirmed` for no reason...and I like being able to have a log line for that case where ZMQ seems to be behind, so once we have that `if` it makes sense to use it to trigger work as well :slightly_smiling_face:

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1910/commits/46935832f41f5a6b4a1c8ce4d57e7acf0844815d

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1910/commits/46935832f41f5a6b4a1c8ce4d57e7acf0844815d

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1910/commits/46935832f41f5a6b4a1c8ce4d57e7acf0844815d

pm47 (approved):

PR #1911, opened 2021-08-16
MPP scale min part amount based on total amount

t-bast:
Some users are experiencing path-finding issues when trying to send big payments.
This is because the way MPP path-finding currently works is:

- Find N routes with Yen's algorithm for a small arbitrary amount (set by the node operator, by default 15 000 sats)
- Split the total amount to send between the routes we found based on their capacity (and some randomization)

If this set of routes cannot carry the whole payment, the path-finding fails (and returns `RouteNotFound`).

For this very simple algorithm to work, instead of providing an arbitrary, static amount to Yen's algorithm, we should instead use `total_amount / num_routes` (which is what this PR does). The reason we didn't do this earlier was because this code was shared with wallets (such as Phoenix) and we wanted to be able to empty all our channels to a given peer (even channels with a very small balance): that's why we were actually *lowering* `minPartAmount` when we had direct channels to the destination instead of *raising* it (which this PR now does).

As `eclair` now primarily focuses on routing nodes / lightning service providers (wallets should use https://github.com/acinq/lightning-kmp which will have a different path-finding algorithm, optimized for wallet use-cases) I don't think it's very important to explicitly try to include channels with a low balance in MPP payments, and this change should raise the success rate of bigger payments (especially when we have direct channels to the destination).

#1815 will nicely complement this PR by penalizing longer routes.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -246,7 +246,7 @@ eclair {
 
       mpp {
         min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
-        max-parts = 6 // maximum number of HTLCs sent per payment: increasing this value will impact performance
+        max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
       }
     }
   }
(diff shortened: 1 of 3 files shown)


pm47 (dismissed):

t-bast:
Rebased.

pm47 (approved):

PR #1912, opened 2021-08-17
Gracefully release Postgres lock on shutdown

### LINKED ISSUE #1896: Graceful stop and DB unlock
I have sometimes
```
2021-08-06 13:06:42,501 ERROR f.a.e.d.p.PgUtils - cannot obtain lock on the database 
```
even when I do `kill -SIGTERM` which means that process should be stopped gracefully.

Isn't there any way to close connections to Postrgre so the node could be restarted immediately?

pm47:
This allows instant restart of eclair when using Postgres backend.

Fixes #1896.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
@@ -120,7 +120,7 @@ object Databases extends Logging {
               logger.info("cancelling the pg lock renew task...")
               leaseLockTask.cancel()
               logger.info("releasing the curent pg lock...")
-              l.releaseExclusiveLock
+              l.releaseExclusiveLock(ds)
               logger.info("closing the connection pool...")
               ds.close()
               Done
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
@@ -16,7 +16,8 @@
 
 package fr.acinq.eclair.db
 
-import akka.actor.ActorSystem
+import akka.Done
+import akka.actor.{ActorSystem, CoordinatedShutdown}
 import com.typesafe.config.Config
 import com.zaxxer.hikari.{HikariConfig, HikariDataSource}
 import fr.acinq.eclair.db.pg.PgUtils.PgLock.LockFailureHandler
@@ -29,6 +30,7 @@ import java.io.File
 import java.nio.file._
 import java.sql.Connection
 import java.util.UUID
+import scala.concurrent.Future
 import scala.concurrent.duration._
 
 trait Databases {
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -111,7 +113,19 @@ object Databases extends Logging {
...
           import system.dispatcher
-          system.scheduler.scheduleWithFixedDelay(l.leaseRenewInterval, l.leaseRenewInterval)(() => l.obtainExclusiveLock(ds))
+          val leaseLockTask = system.scheduler.scheduleWithFixedDelay(l.leaseRenewInterval, l.leaseRenewInterval)(() => l.obtainExclusiveLock(ds))
+
+          CoordinatedShutdown(system).addTask(CoordinatedShutdown.PhaseActorSystemTerminate, "release-postgres-lock") { () =>
+            Future {
+              logger.info("cancelling the pg lock renew task...")
+              leaseLockTask.cancel()
+              logger.info("releasing the curent pg lock...")
+              l.releaseExclusiveLock
```
t-bast, line 123: nit: it's quite weird that this is called like a pure function, whereas it has (strong) side-effects, shouldn't this be `l.releaseExclusiveLock(ds)`?

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
```diff
@@ -119,6 +123,14 @@ object PgUtils extends JdbcUtils {
         }
       }
 
+      override def releaseExclusiveLock(implicit ds: DataSource): Unit = {
+        // put a new lease that expires right away: same as releasing the lock
+        obtainDatabaseLease(instanceId, 1 millisecond) match {
```
t-bast, line 128: Interesting, there is no cleaner mechanism to release a lock?

eclair-core/src/test/scala/fr/acinq/eclair/db/PgUtilsSpec.scala
```diff
@@ -99,6 +100,28 @@ class PgUtilsSpec extends TestKitBaseClass with AnyFunSuiteLike with Eventually
...
   }
 
+  test("lock release utility method") {
+    val pg = EmbeddedPostgres.start()
+    implicit val ds: DataSource = pg.getPostgresDatabase
+
+    val lock1 = LeaseLock(UUID.randomUUID(), 2 minutes, 1 minute, LockFailureHandler.logAndThrow)
+    val lock2 = LeaseLock(UUID.randomUUID(), 2 minutes, 1 minute, LockFailureHandler.logAndThrow)
+
+    lock1.obtainExclusiveLock
```
t-bast, line 110: nit: it's really weird to call these without the parenthesis...

pm47: Done with 61352e250.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
pm47: Done with 61352e250.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgUtils.scala
pm47: In theory we could also delete the line in the lease table, but it's not better, because (1) we lose information (identifier of the last lock holder), (2) deleting from the lease table is a new scenario (currently we only insert or update) and (3) it's more code.

t-bast (approved):

PR #1915, opened 2021-08-18
Update README.md

engenegr:
Documented some best practices I've discovered while working with node software.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,10 @@ Depending on the actual hardware configuration, it may be useful to provide incr
 dbcache=2048
 rpcworkqueue=128
 rpcthreads=128
-rpctimeout=220
+
+# How many seconds bitcoin will wait for a complete RPC HTTP request.
+# after the HTTP connection is established.
+rpcclienttimeout=30
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -59,14 +59,13 @@ zmqpubrawblock=tcp://127.0.0.1:29000
 zmqpubrawtx=tcp://127.0.0.1:29000
 ```
 
-Depending on the actual hardware configuration, it may be useful to provide increased `dbcache` parameter value for faster verification and `rpcthreads` parameter value for better handling of multiple simultaneous API requests from `bitcoind` side. 
+Depending on the actual hardware configuration, it may be useful to provide increased `dbcache` parameter value for faster verification and `rpcworkqueue` parameter value for better handling of API requests on `bitcoind` side. 
 
 ```conf
 # UTXO database cache size, in MiB
 dbcache=2048
 # Number of allowed pending RPC requests (default is 16)
 rpcworkqueue=128
-rpcthreads=128
 
 # How many seconds bitcoin will wait for a complete RPC HTTP request.
 # after the HTTP connection is established.
--- a/README.md
+++ b/README.md
@@ -62,7 +62,9 @@ zmqpubrawtx=tcp://127.0.0.1:29000
 Depending on the actual hardware configuration, it may be useful to provide increased `dbcache` parameter value for faster verification and `rpcthreads` parameter value for better handling of multiple simultaneous API requests from `bitcoind` side. 
 
 ```conf
+# UTXO database cache size, in MiB
 dbcache=2048
+# Number of allowed pending RPC requests (default is 16)
 rpcworkqueue=128
 rpcthreads=128
 
--- a/README.md
+++ b/README.md
@@ -59,6 +59,16 @@ zmqpubrawblock=tcp://127.0.0.1:29000
 zmqpubrawtx=tcp://127.0.0.1:29000
 ```
 
+Depending on the actual hardware configuration, it may be useful to provide increased `dbcache` parameter value for faster verification and `rpcthreads` parameter value for better handling of multiple simultaneous API requests from `bitcoind` side. 
+
+```conf
+dbcache=2048
+rpcworkqueue=128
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,6 +59,16 @@ zmqpubrawblock=tcp://127.0.0.1:29000
 zmqpubrawtx=tcp://127.0.0.1:29000
 ```
 
+Depending on the actual hardware configuration, it may be useful to provide increased `dbcache` parameter value for faster verification and `rpcthreads` parameter value for better handling of multiple simultaneous API requests from `bitcoind` side. 
+
+```conf
+dbcache=2048
+rpcworkqueue=128
```
t-bast, line 66: nit: it's helpful to add small comments to explain what each configuration option does, since these are somewhat advanced tweaks.

```suggestion
# UTXO database cache size, in MiB
dbcache=2048
# Number of allowed pending RPC requests (default is 16)
rpcworkqueue=128
```

README.md
```diff
@@ -59,6 +59,16 @@ zmqpubrawblock=tcp://127.0.0.1:29000
...
+rpcthreads=128
+rpctimeout=220
```
t-bast, line 68: I couldn't find this one in the bitcoind codebase, can you point me to where it's documented?

README.md
```diff
@@ -59,6 +59,16 @@ zmqpubrawblock=tcp://127.0.0.1:29000
 zmqpubrawtx=tcp://127.0.0.1:29000
 ```
 
+Depending on the actual hardware configuration, it may be useful to provide increased `dbcache` parameter value for faster verification and `rpcthreads` parameter value for better handling of multiple simultaneous API requests from `bitcoind` side. 
+
+```conf
+dbcache=2048
+rpcworkqueue=128
+rpcthreads=128
```
t-bast, line 67: I really wouldn't recommend that. If you want the best performance, you shouldn't have more threads than your number of virtual CPUs. If you put that setting on a hardware with a small number of vCPUs, you'll be able to keep more requests pending in memory, but you'll strongly degrade your performance (context-switching all the time can be expensive and is unnecessary here) so you'll take a lot more time handling these requests.

Raising `rpcworkqueue` is a much better choice here, it gives you the benefits you want without the drawbacks.

README.md
```diff
@@ -117,7 +127,11 @@ Any BTC found in the wallet can be used to fund the channels you choose to open.
 
 #### Java Environment Variables
 
-Some advanced parameters can be changed with java environment variables. Most users won't need this and can skip this section.
+Some advanced parameters can be changed with java environment variables. Most users won't need this and can skip this section. However, if Eclair instance fails to start with Java heap size error, try to increase JVM memory allocation via
```
t-bast, line 130: I'd be even more precise:
```suggestion
Some advanced parameters can be changed with java environment variables. Most users won't need this and can skip this section.

However, if you're seeing Java heap size errors, you can try increasing the maximum memory allocated to the JVM with the `-Xmx` parameter.

You can for example set it to use up to 512 MB (or any value that fits the amount of RAM on your machine) with:
```

t-bast (commented):
Thanks for sharing those, I agree they're helpful to have in the README.

README.md
engenegr: That's the problem there are not so much documentation around these options. They are out of reference config. Should I care and dig deeper? I left it on behalf of Eclair's user.

README.md
engenegr: > I really wouldn't recommend that. If you want the best performance, you shouldn't have more threads than your number of virtual CPUs. If you put that setting on a hardware with a small number of vCPUs, you'll be able to keep more requests pending in memory, but you'll strongly degrade your performance (context-switching all the time can be expensive and is unnecessary here) so you'll take a lot more time handling these requests.

You are likely get it wrong `rpcthreads` is not related to CPU threads. It is about RPC threads in a sense of accepted requests. Default is 4 I'm testing some more on various hardware setups. No results yet.

t-bast: Interesting, I didn't know that! I'll check the bitcoin code in more details to better understand that parameter.

README.md
t-bast: I need to look deeper, but are you sure this is really taken into account by bitcoind?

t-bast: I see you've removed it from the last commit, all good then since I'm not sure this parameter really exists in bitcoin.

README.md
t-bast: Are you sure about that? I'm looking at https://github.com/bitcoin/bitcoin/blob/b784ab10f555b043c77a2356faff492c35a3404c/src/httpserver.cpp#L417, I'm not 100% sure but it really looks like these are `std::thread` objects, which are OS threads AFAIK?

engenegr: Let maybe keep it around 4? I do not see clear recommendations from documentation.
https://en.cppreference.com/w/cpp/thread/thread

t-bast: I think we should remove it from the recommendations in that case, it's the same default value as what bitcoin core sets.

engenegr: I just tried 128. I can't report about meaningful difference also because my node is not under heavy load. I'd suggested to leave it but you are free to remove it of course.

t-bast: I don't think we should keep a parameter for which we don't have a good recommendation, can you please remove it and we should be good to merge?
And don't forget to remove it from the line above as well (where you can instead mention `rpcworkqueue`.

engenegr: Done

t-bast (approved):

PR #1917, opened 2021-08-19
Zmq resilence

engenegr:
In addition to proposed by @t-bast ZMQ options following an example in Core Python code, I've add two more lines.

For two days Eclair is following Core without meaningful lag (before it was like 5-6 blocks). Maximum lag I observed was about 2 blocks and probably related to more frequent  blocks due to increasing hashrate.

I did little research collecting timestamps as they pop-up in Eclair logs and comparing them with block timestamps.

zmq-blocks

zmq-delta

I'm continuing to look after node health but I hope these two little strings address the issue.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/zmq/ZMQActor.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/zmq/ZMQActor.scala
@@ -41,6 +41,8 @@ class ZMQActor(address: String, topic: String, connected: Option[Promise[Done]] 
   subscriber.monitor("inproc://events", ZMQ.EVENT_CONNECTED | ZMQ.EVENT_DISCONNECTED)
   subscriber.setRcvHWM(0) // disable high watermark to ensure we never drop messages
   subscriber.setTCPKeepAlive(1) // enable tcp keep-alive
+  subscriber.setTCPKeepAliveIdle(300)
+  subscriber.setTCPKeepAliveInterval(300)
   subscriber.subscribe(topic.getBytes(ZMQ.CHARSET))
   subscriber.connect(address)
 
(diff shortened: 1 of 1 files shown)


t-bast:
Cool, let's give it more time and if everything looks good, we'll merge this.
I'll update my PR to set these values and I'll make them configurable by the node operator to allow more experimentation.

engenegr:
For 4 days I haven't noticed any significant lagging.

t-bast:
I don't think we should merge the changes to `setTCPKeepAliveIdle` and `setTCPKeepAliveInterval`.
There's no inherently good value to provide here, it all depends on the user's setup and observed behavior.
My hope is that with #1910 everything works fine and there's no need to fine-tune these parameters.

engenegr:
> There's no inherently good value to provide here

I think you are wrong. Good value I guess is around half of expected block interval.
I'm feeling nervous without these changes since i'm testing with them for a couple of weeks and the node goes fine with them.

pm47:
We could make those settings configurable. Not in `NodeParams`, but directly through `context.system.settings`.

t-bast:
> I think you are wrong. Good value I guess is around half of expected block interval.

Definitely not, this config is used for both blocks and transactions, and transactions come all the time (even though they're not critical to receive via ZMQ).

What you're setting here is a TCP-level configuration, it's completely different from higher level protocols.
The only thing that would impact the values you should put here is the quality of the connection (latency and bandwidth).

We could make this user-configurable, but I'm afraid it won't help us find what default value to put here, that needs more thorough investigation on what exactly is done with these parameters internally to ZMQ.

engenegr:
> What you're setting here is a TCP-level configuration, it's completely different from higher level protocols.

I do not insist that it is similar or different. I argue that it pushes ZMQ lib to properly maintain connection. And apparently connections are different from subscription to subscription. Otherwise we wouldn't have such long story with ZMQActor.

engenegr:
Okay. There were many other changes. This PR may be not merged and I will try master without these parameters and see what happens. Maybe those other changes in block updates and processing logic will influence somehow node behaviour in that regard. Otherwise I will just make my own patch.

t-bast:
Thanks, let me see if the switch to block hashes fixed the issue.
If not, then we can investigate these parameters deeper to understand them enough to provide good default value recommendations.

PR #1918, opened 2021-08-23
Fix duplicate channel_updates in auditDb when restarting

thomash-acinq:
Because the previous channel update is never part of the channel state when restarting eclair, we always think that they have changed and add them to the database.
We now check that the channel update has changed when restoring it.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -304,15 +304,12 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
             log.info("refreshing channel_update due to configuration changes old={} new={}", normal.channelUpdate, candidateChannelUpdate)
             candidateChannelUpdate
           }
-
-          context.system.eventStream.publish(LocalChannelUpdate(self, normal.commitments.channelId, normal.shortChannelId, normal.commitments.remoteParams.nodeId, normal.channelAnnouncement, channelUpdate1, Some(normal.channelUpdate), normal.commitments))
-
           // we need to periodically re-send channel updates, otherwise channel will be considered stale and get pruned by network
           // we take into account the date of the last update so that we don't send superfluous updates when we restart the app
           val periodicRefreshInitialDelay = Helpers.nextChannelUpdateRefresh(channelUpdate1.timestamp)
           context.system.scheduler.scheduleWithFixedDelay(initialDelay = periodicRefreshInitialDelay, delay = REFRESH_CHANNEL_UPDATE_INTERVAL, receiver = self, message = BroadcastChannelUpdate(PeriodicRefresh))
 
-          goto(OFFLINE) using normal.copy(channelUpdate = channelUpdate1)
+          goto(OFFLINE) using normal.copy(channelUpdate = channelUpdate1, channelUpdateBeforeRestore_opt = Some(normal.channelUpdate))
 
         case funding: DATA_WAIT_FOR_FUNDING_CONFIRMED =>
           watchFundingTx(funding.commitments)
@@ -669,7 +666,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       val initialChannelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, shortChannelId, nodeParams.expiryDelta, d.commitments.remoteParams.htlcMinimum, fees.feeBase, fees.feeProportionalMillionths, commitments.capacity.toMilliSatoshi, enable = Helpers.aboveReserve(d.commitments))
       // we need to periodically re-send channel updates, otherwise channel will be considered stale and get pruned by network
       context.system.scheduler.scheduleWithFixedDelay(initialDelay = REFRESH_CHANNEL_UPDATE_INTERVAL, delay = REFRESH_CHANNEL_UPDATE_INTERVAL, receiver = self, message = BroadcastChannelUpdate(PeriodicRefresh))
-      goto(NORMAL) using DATA_NORMAL(commitments.copy(remoteNextCommitInfo = Right(nextPerCommitmentPoint)), shortChannelId, buried = false, None, initialChannelUpdate, None, None) storing()
+      goto(NORMAL) using DATA_NORMAL(commitments.copy(remoteNextCommitInfo = Right(nextPerCommitmentPoint)), shortChannelId, buried = false, None, initialChannelUpdate, None, None, None) storing()
 
     case Event(remoteAnnSigs: AnnouncementSignatures, d: DATA_WAIT_FOR_FUNDING_LOCKED) if d.commitments.announceChannel =>
       log.debug("received remote announcement signatures, delaying")
@@ -1885,10 +1882,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       (state, nextState, stateData, nextStateData) match {
         // ORDER MATTERS!
         case (WAIT_FOR_INIT_INTERNAL, OFFLINE, _, normal: DATA_NORMAL) =>
-          // LocalChannelUpdate is already published when restoring the channel
           Logs.withMdc(diagLog)(Logs.mdc(category_opt = Some(Logs.LogCategory.CONNECTION))) {
             log.debug("re-emitting channel_update={} enabled={} ", normal.channelUpdate, Announcements.isEnabled(normal.channelUpdate.channelFlags))
           }
+          context.system.eventStream.publish(LocalChannelUpdate(self, normal.commitments.channelId, normal.shortChannelId, normal.commitments.remoteParams.nodeId, normal.channelAnnouncement, normal.channelUpdate, normal.channelUpdateBeforeRestore_opt, normal.commitments))
         case (_, _, d1: DATA_NORMAL, d2: DATA_NORMAL) if d1.channelUpdate == d2.channelUpdate && d1.channelAnnouncement == d2.channelAnnouncement =>
           // don't do anything if neither the channel_update nor the channel_announcement didn't change
           ()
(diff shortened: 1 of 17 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -304,6 +304,10 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
             log.info("refreshing channel_update due to configuration changes old={} new={}", normal.channelUpdate, candidateChannelUpdate)
             candidateChannelUpdate
           }
+
+          val hasChanged = !Announcements.areSameWithoutFlags(candidateChannelUpdate, normal.channelUpdate)
+          context.system.eventStream.publish(LocalChannelUpdate(self, normal.commitments.channelId, normal.shortChannelId, normal.commitments.remoteParams.nodeId, normal.channelAnnouncement, normal.channelUpdate, hasChanged, normal.commitments))
```
t-bast, line 309: Shouldn't this be:

```suggestion
          context.system.eventStream.publish(LocalChannelUpdate(self, normal.commitments.channelId, normal.shortChannelId, normal.commitments.remoteParams.nodeId, normal.channelAnnouncement, channelUpdate1, hasChanged, normal.commitments))
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1879,25 +1883,24 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
 
-      val previousChannelUpdate_opt = stateData match {
-        case data: DATA_NORMAL => Some(data.channelUpdate)
-        case _ => None
-      }
-
       (state, nextState, stateData, nextStateData) match {
         // ORDER MATTERS!
         case (WAIT_FOR_INIT_INTERNAL, OFFLINE, _, normal: DATA_NORMAL) =>
+          // LocalChannelUpdate is already published when restoring the channel
```
t-bast, line 1889: I preferred having it here, it was clearer when it was explicitly done in the transition phases...

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
```diff
@@ -38,7 +38,7 @@ case class ChannelIdAssigned(channel: ActorRef, remoteNodeId: PublicKey, tempora
 
 case class ShortChannelIdAssigned(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, previousShortChannelId: Option[ShortChannelId]) extends ChannelEvent
 
-case class LocalChannelUpdate(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, channelAnnouncement_opt: Option[ChannelAnnouncement], channelUpdate: ChannelUpdate, previousChannelUpdate_opt: Option[ChannelUpdate], commitments: AbstractCommitments) extends ChannelEvent
+case class LocalChannelUpdate(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, channelAnnouncement_opt: Option[ChannelAnnouncement], channelUpdate: ChannelUpdate, hasChanged: Boolean, commitments: AbstractCommitments) extends ChannelEvent
```
t-bast, line 41: This is a very confusing name, because the `commitments` object may have changed greatly while the `channelUpdate` didn't, so this will be false in cases where important data inside this class has changed, downstream components may misinterpret this field.

t-bast (commented):
This is a dangerous change, because it impacts downstream components and how we re-emit `channel_update`s to our peers, which may lead to spec violations if we make a mistake.

pm47:
I still don't get what caused the issue precisely. Didn't the following code ensure that we wouldn't emit duplicate updates?

```scala
val channelUpdate1 = if (Announcements.areSame(candidateChannelUpdate, normal.channelUpdate)) {
  // if there was no configuration change we keep the existing channel update
  normal.channelUpdate
} else {
  log.info("refreshing channel_update due to configuration changes old={} new={}", normal.channelUpdate, candidateChannelUpdate)
  candidateChannelUpdate
}
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
thomash-acinq: Absolutely

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
thomash-acinq: I understand but the problem is that at this place we don't have enough information to know if the channel update changed or not.

thomash-acinq:
> I still don't get what caused the issue precisely. Didn't the following code ensure that we wouldn't emit duplicate updates?
> 
> ```scala
> val channelUpdate1 = if (Announcements.areSame(candidateChannelUpdate, normal.channelUpdate)) {
>   // if there was no configuration change we keep the existing channel update
>   normal.channelUpdate
> } else {
>   log.info("refreshing channel_update due to configuration changes old={} new={}", normal.channelUpdate, candidateChannelUpdate)
>   candidateChannelUpdate
> }
> ```

This code only reuses the previous channel update if it is still valid. But after that, without knowing what was the previous update we can't know if it has changed or not.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
thomash-acinq: I've reverted to including the full previous channel update.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
thomash-acinq: I've added a `channelUpdateBeforeRestore_opt` in `DATA_NORMAL` so that we have access to it in the transition function and put everything back there.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -302,6 +302,7 @@ private[channel] object ChannelCodecs3 {
         ("buried" | bool8) ::
         ("channelAnnouncement" | optional(bool8, lengthDelimited(channelAnnouncementCodec))) ::
         ("channelUpdate" | lengthDelimited(channelUpdateCodec)) ::
+        ("channelUpdateBeforeRestore_opt" | provide[Option[ChannelUpdate]](None)) ::
```
t-bast, line 305: Unfortunately it's not that simple, because what you're doing here means you're never actually saving it in the DB, so it will always be `None` after a restart.

There's a simple trick to migrate codecs in that case (see what's done in https://github.com/ACINQ/eclair/pull/1768 where I have the same kind of needs). However I'm not sure we want to go in that direction for this issue, give me some time to review this PR more in depth before we go further.

thomash-acinq: I know but I think it's fine not to save it. Anytime we restart, this value will be ignored and overwritten anyway.

pm47:
Here is an idea. We move the "update `channel_update` according to conf" code (https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L289-L306) to the `SYNCING` state (there is already related code here: https://github.com/ACINQ/eclair/blob/c5046582649832248bc858522e4ebc18502e5759/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1668-1676).

This has the following benefits:
- the change in `channel_update` would be caught by the `onTransition` handler
- we do not change the general logic of emitting `channel_update`s (I agree with @t-bast's concerns)
- it prevents us to generate thousands of new `channel_update` at startup when we update the conf and will instead to it at the first reconnection.

OTOH, it also means that we would run this code at each reconnection.

thomash-acinq:
>     * the change in `channel_update` would be caught by the `onTransition` handler 
>     * we do not change the general logic of emitting `channel_update`s (I agree with @t-bast's concerns)
This would emit an additional `LocalChannelUpdate` compared to the current situation. Unless we stop emitting it on WAIT_FOR_INIT_INTERNAL -> OFFLINE which would actually simplify things.

pm47:
>> the change in channel_update would be caught by the onTransition handler
>> we do not change the general logic of emitting channel_updates (I agree with @t-bast's concerns)

>This would emit an additional LocalChannelUpdate compared to the current situation.

Only if the conf changes, right? Alternatively, instead of updating the `channel_update` manually, we could have the channel send a `CMD_UPDATE_RELAY_FEE` to itself, which ensure we go to the normal flow and is simpler to reason about. There are details though (e.g. need to add the cltv expiry delta to the command), and we still need to update `channel_update`s inline, e.g. in `handleAddDisconnected`.

> Unless we stop emitting it on WAIT_FOR_INIT_INTERNAL -> OFFLINE which would actually simplify things.

That is a more risky change, even if sending a `LocalChannelUpdate` at this transition does feel hacky. At first glance, removing this shouldn't impact direct peers, since we are disconnected at this step. The router is another story though: `LocalChannelUpdate`s are how it learns about private channels.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
t-bast: You're completely right, I really didn't look hard enough at this PR, we only care about the `channelUpdate` field being stored!
Let's investigate the other changes proposed by pm47 though, we'll come back to this option later if we can't find a more satisfying one.

thomash-acinq:
> > This would emit an additional LocalChannelUpdate compared to the current situation.
> 
> Only if the conf changes, right? 

Yes, only if it changes. And in that case it will emit a first channel update that is outdated and then the correct one.

If we want to not change the current flow between states, I think adding `channelUpdateBeforeRestore_opt` to `DATA_NORMAL` is the best solution. Anything else will probably change some behavior of the system.

thomash-acinq:
I've tried multiple ways doesn't seem possible to make it work by updating the channelUpdate in the SYNCING state.
The problem is the transition WAIT_FOR_INIT_INTERNAL -> OFFLINE. If we stop emitting the LocalChannelUpdate from there, then if there is no change, no LocalChannelUpdate will be emitted at all which is bad for the router. If we continue emitting the LocalChannelUpdate from there, then it may be an outdated one if there was a config change which leads to useless work.
Also we do additional work every time we reconnect even though it's only needed after a restart. Here is how that would look like : https://github.com/ACINQ/eclair/pull/1920

PR #1921, opened 2021-08-25
Split `SendPayment` in `SendPaymentToRoute` and `SendPaymentToNode`

thomash-acinq:
Split `SendPayment` in `SendPaymentToRoute` and `SendPaymentToNode`
This allow us to ensure that when sending to a specific route, we never try another route.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
@@ -135,7 +135,7 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
         router ! RouteRequest(nodeParams.nodeId, data.c.targetNodeId, data.c.finalPayload.amount, sendPaymentToNode.getMaxFee(nodeParams), data.c.assistedRoutes, ignore1, sendPaymentToNode.routeParams, paymentContext = Some(cfg.paymentContext))
         goto(WAITING_FOR_ROUTE) using WaitingForRoute(data.c, data.failures :+ failure, ignore1)
       case (_: SendPaymentToRoute) =>
-        log.error("retrying SendPaymentToRoute")
+        log.error("unexpected retry during SendPaymentToRoute")
         stop(FSM.Normal)
     }
   }
@@ -160,7 +160,7 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
     }
   }
 
-  private def handleRemoteFail(d: WaitingForComplete, fail: UpdateFailHtlc): State = {
+  private def handleRemoteFail(d: WaitingForComplete, fail: UpdateFailHtlc) = {
     import d._
     (Sphinx.FailurePacket.decrypt(fail.reason, sharedSecrets) match {
       case success@Success(e) =>
@@ -203,31 +203,30 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
       case Success(e@Sphinx.DecryptedFailurePacket(nodeId, failureMessage: Update)) =>
         log.info(s"received 'Update' type error message from nodeId=$nodeId, retrying payment (failure=$failureMessage)")
         val failure = RemoteFailure(cfg.fullRoute(route), e)
-        val ignore1 = if (Announcements.checkSig(failureMessage.update, nodeId)) {
+        if (Announcements.checkSig(failureMessage.update, nodeId)) {
           val assistedRoutes1 = handleUpdate(nodeId, failureMessage, d)
           val ignore1 = PaymentFailure.updateIgnored(failure, ignore)
           // let's try again, router will have updated its state
           c match {
-            case (c: SendPaymentToRoute) =>
-              c.route.fold(
-                hops => router ! FinalizeRoute(c.finalPayload.amount, hops, c.assistedRoutes, paymentContext = Some(cfg.paymentContext)),
-                route => self ! RouteResponse(route :: Nil)
-              )
+            case (_: SendPaymentToRoute) =>
+              log.error("unexpected retry during SendPaymentToRoute")
+              stop(FSM.Normal)
             case (c: SendPaymentToNode) =>
               router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, c.getMaxFee(nodeParams), assistedRoutes1, ignore1, c.routeParams, paymentContext = Some(cfg.paymentContext))
+              goto(WAITING_FOR_ROUTE) using WaitingForRoute(c, failures :+ failure, ignore1)
           }
-          ignore1
         } else {
           // this node is fishy, it gave us a bad sig!! let's filter it out
           log.warning(s"got bad signature from node=$nodeId update=${failureMessage.update}")
           c match {
-            case (_: SendPaymentToRoute) => return stop(FSM.Normal)
+            case (_: SendPaymentToRoute) =>
+              log.error("unexpected retry during SendPaymentToRoute")
+              stop(FSM.Normal)
             case (c: SendPaymentToNode) =>
               router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, c.getMaxFee(nodeParams), c.assistedRoutes, ignore + nodeId, c.routeParams, paymentContext = Some(cfg.paymentContext))
+              goto(WAITING_FOR_ROUTE) using WaitingForRoute(c, failures :+ failure, ignore + nodeId)
           }
-          ignore + nodeId
         }
-        goto(WAITING_FOR_ROUTE) using WaitingForRoute(c, failures :+ failure, ignore1)
       case Success(e@Sphinx.DecryptedFailurePacket(nodeId, failureMessage)) =>
         log.info(s"received an error message from nodeId=$nodeId, trying to use a different channel (failure=$failureMessage)")
         val failure = RemoteFailure(cfg.fullRoute(route), e)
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -130,9 +129,15 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
-    router ! RouteRequest(nodeParams.nodeId, data.c.targetNodeId, data.c.finalPayload.amount, data.c.getMaxFee(nodeParams), data.c.assistedRoutes, ignore1, data.c.routeParams, paymentContext = Some(cfg.paymentContext))
-    goto(WAITING_FOR_ROUTE) using WaitingForRoute(data.c, data.failures :+ failure, ignore1)
+    data.c match {
+      case (sendPaymentToNode: SendPaymentToNode) =>
+        val ignore1 = PaymentFailure.updateIgnored(failure, data.ignore)
+        router ! RouteRequest(nodeParams.nodeId, data.c.targetNodeId, data.c.finalPayload.amount, sendPaymentToNode.getMaxFee(nodeParams), data.c.assistedRoutes, ignore1, sendPaymentToNode.routeParams, paymentContext = Some(cfg.paymentContext))
+        goto(WAITING_FOR_ROUTE) using WaitingForRoute(data.c, data.failures :+ failure, ignore1)
+      case (_: SendPaymentToRoute) =>
+        log.error("retrying SendPaymentToRoute")
+        stop(FSM.Normal)
```
thomash-acinq, line 139: This should never happen as `maxAttempts` would be 1 and we wouldn't call `retry`.

t-bast: Then maybe the log message should be `unexpected retry during SendPaymentToRoute`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -202,12 +207,24 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
           val assistedRoutes1 = handleUpdate(nodeId, failureMessage, d)
           val ignore1 = PaymentFailure.updateIgnored(failure, ignore)
           // let's try again, router will have updated its state
-          router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, c.getMaxFee(nodeParams), assistedRoutes1, ignore1, c.routeParams, paymentContext = Some(cfg.paymentContext))
+          c match {
+            case (c: SendPaymentToRoute) =>
+              c.route.fold(
+                hops => router ! FinalizeRoute(c.finalPayload.amount, hops, c.assistedRoutes, paymentContext = Some(cfg.paymentContext)),
```
t-bast, line 213: In case the updated channel comes from the invoice and not the public graph, you need to pass it to `FinalizeRoute`:

```suggestion
                hops => router ! FinalizeRoute(c.finalPayload.amount, hops, assistedRoutes1, paymentContext = Some(cfg.paymentContext)),
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -202,12 +207,24 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
+                route => self ! RouteResponse(route :: Nil)
```
t-bast, line 214: In that case we should abort, shouldn't we? Or we should update `route` to take into account the changed `channel_update`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -202,12 +207,24 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
+            case (c: SendPaymentToNode) =>
+              router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, c.getMaxFee(nodeParams), assistedRoutes1, ignore1, c.routeParams, paymentContext = Some(cfg.paymentContext))
+          }
           ignore1
         } else {
           // this node is fishy, it gave us a bad sig!! let's filter it out
           log.warning(s"got bad signature from node=$nodeId update=${failureMessage.update}")
-          router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, c.getMaxFee(nodeParams), c.assistedRoutes, ignore + nodeId, c.routeParams, paymentContext = Some(cfg.paymentContext))
+          c match {
+            case (_: SendPaymentToRoute) => return stop(FSM.Normal)
```
t-bast, line 224: I don't like using `return`, it breaks the flow and is similar to `goto` (with the same pitfalls).
Can we avoid it with a clever refactoring?

t-bast (commented):
Good idea, can you add some tests of the new behavior that is introduced for `SendToRoute`?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -198,19 +203,30 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
-        val ignore1 = if (Announcements.checkSig(failureMessage.update, nodeId)) {
+        if (Announcements.checkSig(failureMessage.update, nodeId)) {
           val assistedRoutes1 = handleUpdate(nodeId, failureMessage, d)
           val ignore1 = PaymentFailure.updateIgnored(failure, ignore)
           // let's try again, router will have updated its state
-          router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, c.getMaxFee(nodeParams), assistedRoutes1, ignore1, c.routeParams, paymentContext = Some(cfg.paymentContext))
-          ignore1
+          c match {
+            case (_: SendPaymentToRoute) =>
+              log.error("unexpected retry during SendPaymentToRoute")
```
t-bast, line 212: This log line doesn't make sense, it's completely ok to receive a remote error here, we're not retrying.
Maybe just replace it with:
```suggestion
              log.info("not retrying SendPaymentToRoute")
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -198,19 +203,30 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
+              goto(WAITING_FOR_ROUTE) using WaitingForRoute(c, failures :+ failure, ignore1)
+          }
         } else {
           // this node is fishy, it gave us a bad sig!! let's filter it out
           log.warning(s"got bad signature from node=$nodeId update=${failureMessage.update}")
-          router ! RouteRequest(nodeParams.nodeId, c.targetNodeId, c.finalPayload.amount, c.getMaxFee(nodeParams), c.assistedRoutes, ignore + nodeId, c.routeParams, paymentContext = Some(cfg.paymentContext))
-          ignore + nodeId
+          c match {
+            case (_: SendPaymentToRoute) =>
+              log.error("unexpected retry during SendPaymentToRoute")
```
t-bast, line 223: Same here.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
```diff
@@ -713,7 +713,7 @@ class PaymentLifecycleSpec extends BaseRouterSpec {
     import payFixture._
     import cfg._
 
-    val request = SendPayment(sender.ref, d, Onion.createSinglePartPayload(defaultAmountMsat, defaultExpiry, defaultInvoice.paymentSecret.get), 3)
```
t-bast, line 716: We need to add tests to this file to verify that we're not retrying `SendPaymentToRoute`.

t-bast (commented):
> The cases I was trying to fix could not actually happen since there is a check that we never retry for SendPaymentToRoute

Good catch, the code is nicely simplified now.

However, I do think we should retry `SendPaymentToRoute` when we receive a `channel_update` from one of the hops that tells us that the fees or cltv have changed: when that happens we should retry the same route but with the latest fees/cltv applied (and we should retry only in this specific error case).

But this isn't a priority at all, we can do that in a follow up PR.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
thomash-acinq: It is fine to receive a remote error but this is still a form of retry. There is already this case that prevents retries and takes care of such remote errors:

```scala
case res if failures.size + 1 >= c.maxAttempts =>
```

For `SendPaymentToRoute`, `maxAttempts` is set to 1 so we should never reach this retry.

thomash-acinq:
> However, I do think we should retry `SendPaymentToRoute` when we receive a `channel_update` from one of the hops that tells us that the fees or cltv have changed: when that happens we should retry the same route but with the latest fees/cltv applied (and we should retry only in this specific error case).

I can put that in an other PR.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
t-bast: Good catch, you're right!

t-bast (approved):

PR #1922, opened 2021-08-25
Send `CMD_UPDATE_RELAY_FEES` to self at restore 

pm47:
This is an alternative to #1918 and #1920. It's very close to the
latter, except that we do check the conf only once in the
`WAIT_FOR_INIT_INTERNAL` state, as opposed to at each reconnection in
`SYNCING`.

We do not change the `channel_update` in `WAIT_FOR_INIT_INTERNAL`, which
allows us to set `previousChannelUpdate_opt=Some(normal.channelUpdate)`
in the transition and fix the duplicate bug in the audit db.

If there is a change in conf, there will be an additional
`LocalChannelUpdate` emitted, but only at reconnection, and following
the regular update flow, which should protect us against regressions.

We do handle `CMD_UPDATE_RELAY_FEES` in both `OFFLINE` and `SYNCING`,
because there may be a race between `CMD_UPDATE_RELAY_FEES` and
`ChannelRestablish` if the conf change at restore. And there was no good
reason to behave differently in those states anyway.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2148,6 +2129,15 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     }
   }
 
+  private def handleUpdateRelayFeeDisconnected(c: CMD_UPDATE_RELAY_FEE, d: DATA_NORMAL) = {
+    val channelUpdate1 = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, c.cltvExpiryDelta_opt.getOrElse(d.channelUpdate.cltvExpiryDelta), d.channelUpdate.htlcMinimumMsat, c.feeBase, c.feeProportionalMillionths, d.commitments.capacity.toMilliSatoshi, enable = false)
```
pm47, line 2133: It probably doesn't matter a lot, but I believe this is a bug:

```suggestion
    val channelUpdate1 = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, d.shortChannelId, c.cltvExpiryDelta_opt.getOrElse(d.channelUpdate.cltvExpiryDelta), d.channelUpdate.htlcMinimumMsat, c.feeBase, c.feeProportionalMillionths, d.commitments.capacity.toMilliSatoshi, enable = Announcements.isEnabled(d.channelUpdate.channelFlags))
```

thomash-acinq (commented):
Overall it looks like a clean solution.
The only downside is that it can behave differently than the current code by emitting an outdated LocalChannelUpdate. If you think it's fine, I'm OK with it. I'm not sure how this LocalChannelUpdate is used later, maybe it's not a problem to emit a bad one followed not too long after by a good one.

There should be some tests that the correct LocalChannelUpdate are emitted when we expect and won't produce duplicate rows in the channel_updates table.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2148,6 +2129,15 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
+    log.info(s"updating relay fees: prev={} next={}", d.channelUpdate.toStringShort, channelUpdate1.toStringShort)
+    val replyTo = if (c.replyTo == ActorRef.noSender) sender() else c.replyTo
+    replyTo ! RES_SUCCESS(c, d.channelId)
+    // we're in OFFLINE state, we don't broadcast the new update right away, we will do that when next time we go to NORMAL state
+    stay() using d.copy(channelUpdate = channelUpdate1) storing()
```
pm47, line 2138: So... this doesn't work, as evidenced by 2c8db6e.

By using a `stay()` to defer the sending of the event, we lose the reference to the previous `channel_update` :man_facepalming:

pm47:
Closed in favor of #1934

PR #1924, opened 2021-08-26
Extend `ChannelRestored` event

btcontract:
Previosuly `ChannelRestored` event had `commitments: AbstractCommitments` field, since https://github.com/ACINQ/eclair/pull/1737/files#diff-4dac524d45d2e6f5b4c59d0a105f689ae1dfc8459eafdadf4d6d79cd4d46165bR35 it was replaced with `data: HasCommitments` which breaks compatibility with hosted channels plugin since it uses this event, but can't use `HasCommitments` by design.

What's proposed here is to extend that field to `data: Either[HasCommitments, AbstractCommitments]`. Currently `data` is mostly used by `ChannelsListener` to calculate global balance and this change enables to take non-standard channel balances into account in future (they are omitted for now).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/balance/ChannelsListener.scala
@@ -9,7 +9,7 @@ import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.eclair.balance.ChannelsListener._
 import fr.acinq.eclair.channel.Helpers.Closing
-import fr.acinq.eclair.channel.{ChannelPersisted, ChannelRestored, HasCommitments}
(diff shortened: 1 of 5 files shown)


t-bast:
IIRC this was changed back because `AbstractCommitments` wasn't a well-designed abstractions and this broke some of our internal feature branches...I'll let @pm47 chime in on that one.

btcontract:
Interesting, hope we can arrive at a better abstraction then. 

For now `ChannelRestored` event is the only breaking point and as far as I can see it's justified specifically in `ChannelsListener` because that actor is interested not only in commitments data, but whether channel itself is closing, too.

pm47:
I can confirm that there is no other reason that for computing the balance. `data: Either[HasCommitments, AbstractCommitments]` would work but it's really too ugly.

Why do you have to use `ChannelRestored`? Your code must be the one emitting and listening to those events, since we only emit `ChannelRestored` for non-virtual channels. What stops you from using another class, e.g. `HostedChannelRestored`? `ChannelEvent` isn't a `sealed trait` (which it should, btw)?

PR #1927, opened 2021-08-31
Add `AbstractChannelRestored` event trait

btcontract:
This is an alternative and hopefully more principled solution to https://github.com/ACINQ/eclair/pull/1924. It makes `Register` in base Eclair to rely on new `ChannelAvailable` trait rather than on concrete `ChannelRestored` class so non-standard channels can be injected into `Register` without affecting other parts of system.

> https://github.com/ACINQ/eclair/pull/1924#issuecomment-908417301
> Why do you have to use ChannelRestored? Your code must be the one emitting and listening to those events, since we only emit ChannelRestored for non-virtual channels.

`ChannelRestored` connectes HC to the rest of the system by injecting it into `Register` so routings like `Normal channel -> HC`, `HC -> Normal channel`, `HC -> HC` become possible while reusing all of the existing base code. 

This is one of the few places where HC must make base system to know about itself and it was doing that by emitting `ChannelRestored` just like a normal channel would. With this change it emits a special `HostedChannelRestored extends ChannelAvailable`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
@@ -32,14 +32,15 @@ trait ChannelEvent
 
 case class ChannelCreated(channel: ActorRef, peer: ActorRef, remoteNodeId: PublicKey, isFunder: Boolean, temporaryChannelId: ByteVector32, initialFeeratePerKw: FeeratePerKw, fundingTxFeeratePerKw: Option[FeeratePerKw]) extends ChannelEvent
 
-trait ChannelAvailable extends ChannelEvent {
+// This trait can be used by non-standard channels to inject themselves into Register actor and thus make them usable for routing
+trait AbstractChannelRestored extends ChannelEvent {
   val channel: ActorRef
   val channelId: ByteVector32
   val peer: ActorRef
   val remoteNodeId: PublicKey
 }
 
-case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, data: HasCommitments) extends ChannelAvailable
+case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, data: HasCommitments) extends AbstractChannelRestored
 
 case class ChannelIdAssigned(channel: ActorRef, remoteNodeId: PublicKey, temporaryChannelId: ByteVector32, channelId: ByteVector32) extends ChannelEvent
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
@@ -32,7 +32,14 @@ trait ChannelEvent
 
 case class ChannelCreated(channel: ActorRef, peer: ActorRef, remoteNodeId: PublicKey, isFunder: Boolean, temporaryChannelId: ByteVector32, initialFeeratePerKw: FeeratePerKw, fundingTxFeeratePerKw: Option[FeeratePerKw]) extends ChannelEvent
 
-case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, data: HasCommitments) extends ChannelEvent
+trait ChannelAvailable extends ChannelEvent {
+  val channel: ActorRef
+  val channelId: ByteVector32
+  val peer: ActorRef
+  val remoteNodeId: PublicKey
+}
+
+case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, data: HasCommitments) extends ChannelAvailable
 
 case class ChannelIdAssigned(channel: ActorRef, remoteNodeId: PublicKey, temporaryChannelId: ByteVector32, channelId: ByteVector32) extends ChannelEvent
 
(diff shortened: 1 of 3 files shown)


t-bast:
That sounds much better, I really like this solution!

To ensure that this abstract mechanism keeps working with future changes in the eclair codebase, you should create a `RegisterSpec` and add tests that verify that various kinds of `ChannelAvailable` events are correctly taken into account in the register maps. You should also add comments to the `ChannelAvailable` trait to explain how it can be used to inject custom channel types to `eclair` via plugins.

Finally, does that mean we can completely get rid of the `AbstractCommitments` abstraction in `eclair`? It would be great, I think it was misleading and dangerous (easy to misuse).

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
```diff
@@ -32,7 +32,14 @@ trait ChannelEvent
...
 
-case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, data: HasCommitments) extends ChannelEvent
+trait ChannelAvailable extends ChannelEvent {
+  val channel: ActorRef
+  val channelId: ByteVector32
+  val peer: ActorRef
+  val remoteNodeId: PublicKey
+}
+
+case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, data: HasCommitments) extends ChannelAvailable
```
pm47, line 42: For consistency with `Commitment`/`AbstractCommitments`, and because `Available` has another meaning for channels.

```suggestion
trait AbstractChannelRestored extends ChannelEvent {
  val channel: ActorRef
  val channelId: ByteVector32
  val peer: ActorRef
  val remoteNodeId: PublicKey
}

case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey, data: HasCommitments) extends AbstractChannelRestored
```

btcontract:
> Finally, does that mean we can completely get rid of the AbstractCommitments abstraction in eclair? It would be great, I think it was misleading and dangerous (easy to misuse)

It's still needed in a few places, off the top of my head: in `ChannelRelayer` to calculate balances and use HC in non-strict routing if applicable, in `PostRestartHtlcCleaner` to correctly process HTLC leftovers in HCs.

pm47 (approved):
LGTM

t-bast:
> It's still needed in a few places, off the top of my head: in ChannelRelayer to calculate balances and use HC in non-strict routing if applicable, in PostRestartHtlcCleaner to correctly process HTLC leftovers in HCs.

Damn, too bad. I really think we should aim to remove this abstraction though, it's a bad one.
We could rework the events used to ensure we don't need a complete `Commitments` (or `AbstractCommitments`) object but rather a more narrow object that could be more easily injected by plugins.

eclair-core/src/test/scala/fr/acinq/eclair/channel/RegisterSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+import akka.testkit.TestProbe
+import fr.acinq.bitcoin.ByteVector32
+import fr.acinq.bitcoin.Crypto.PublicKey
+import org.scalatest.funsuite.AnyFunSuite
+import akka.pattern.ask
+import akka.util.Timeout
+
+import scala.concurrent.Await
+
+class RegisterSpec extends AnyFunSuite {
```
t-bast, line 16: You should use the same pattern as other test suites in the repo, see `PostRestartHtlcCleanerSpec` for example:

```suggestion
class RegisterSpec extends AnyFunSuite extends TestKitBaseClass with AnyFunSuiteLike with ParallelTestExecution {
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/RegisterSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  test("register processes custom restored events") {
+    implicit val timeout: Timeout = Timeout(3.seconds)
+    implicit val system: ActorSystem = ActorSystem("test")
+
+    val registerRef = system.actorOf(Props(new Register))
+    val customRestoredEvent = CustomChannelRestored(TestProbe().ref, randomBytes32(), TestProbe().ref, randomKey().publicKey)
+    registerRef ! customRestoredEvent
+
+    assert(Await.result(registerRef ? Symbol("channels"), 3.seconds).asInstanceOf[Map[ByteVector32, ActorRef]].keys.head == customRestoredEvent.channelId)
```
t-bast, line 28: You should use `awaitCond` to avoid blocking with `Await`. This is done in many other tests if you need examples.

And your test should be more thorough: you should wait until the register returns a map which contains `customRestoredEvent.channelId` as key, AND that the value is the right actor ref.

eclair-core/src/test/scala/fr/acinq/eclair/channel/RegisterSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import scala.concurrent.Await
+
+class RegisterSpec extends AnyFunSuite {
+
+  case class CustomChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: ActorRef, remoteNodeId: PublicKey) extends AbstractChannelRestored
+
+  test("register processes custom restored events") {
+    implicit val timeout: Timeout = Timeout(3.seconds)
+    implicit val system: ActorSystem = ActorSystem("test")
```
t-bast, line 22: Unnecessary if you extends the right traits as suggested above.

t-bast (changes requested):

btcontract:
> I really think we should aim to remove this abstraction though, it's a bad one.

I'm not against but if this can be postponed I'd rather it will since I want to launch something on mainnet ASAP.

t-bast:
> I'm not against but if this can be postponed I'd rather it will since I want to launch something on mainnet ASAP.

Of course, there's absolutely no rush for that!

t-bast (approved):

Issue #1926, opened 2021-08-30
Channel Disappeared


Mark8673:
Please help, I am a noob and can't figure out why my 500,000sat's are nowhere to be found.

My node pubkey is 02cf138e5dc277f1017d31ac8b661f6034f04fbecd773053e7307fda2dad8a69cd

I opened a channel to lightningconductor.net with Eclair v0.6.2, it said "waiting on funding, confirmed", checked a couple hours later and the transaction was still in the mempool, I increased the transaction fee hoping it would speed up the process and a couple hours after that the channel disappeared from my Eclair window.

Transaction 1 details:

Status: conflicted with a transaction with 33 confirmations
Date: 8/27/2021 12:28
To: bc1ql3l3rny8wp0j09yux42gpcgcj9zdn689yaaxqka2pq52f2ssj4mqyak7k2
Debit: -500 000 sat
Transaction fee: -429 sat
Net amount: -500 429 sat
Transaction ID: 7957b189f8e95dcd335aedb6622174181373affd8806f6a9348ba6cae556f579
Transaction total size: 826 bytes
Transaction virtual size: 424 bytes
Output index: 1

Transaction 2 details with increased transaction fee:

Status: 33 confirmations
Date: 8/27/2021 15:27
To: bc1ql3l3rny8wp0j09yux42gpcgcj9zdn689yaaxqka2pq52f2ssj4mqyak7k2
Debit: -500 000 sat
Transaction fee: -2549 sat
Net amount: -502 549 sat
Transaction ID: 098ca1deb20725b7e880da41a28e9a0fcef216491dc596110b7405a82a3295f1
Transaction total size: 826 bytes
Transaction virtual size: 424 bytes
Output index: 1

I have attached the Eclair log, please tell me how to fix.

Thanks in advance for your help!!!

Mark

[eclair.log](https://github.com/ACINQ/eclair/files/7079133/eclair.log)

t-bast:
> I increased the transaction fee

What do you mean by that? Eclair doesn't let you increase the transaction fee (for very good reasons).
Did you RBF the funding transaction outside of eclair?

Mark8673:
I am using Eclair in conjunction with Bitcoin Core and upped the transaction fee in Bitcoin Core.

t-bast:
You must never RBF a funding transaction outside of eclair...you can do CPFP on it, but never RBF.
That broke the channel, and your funds are now in a multisig 2-of-2 output shared between you and lightningconductor.net.

[...]

PR #1929, opened 2021-08-31
Add warning about GUI deprecation

t-bast:
The GUI has been deprecated a long time ago, and doesn't inform users about potential risks, such as RBF-ing funding txs.

We should instead incentivize users to use the CLI and read the documentation, which ensures they will know about potential pitfalls. I have added a warning about RBF to the wiki documentation [here](https://github.com/ACINQ/eclair/wiki/Usage).

This is related to #1926

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,22 +10,24 @@
 This software follows the [Lightning Network Specifications (BOLTs)](https://github.com/lightningnetwork/lightning-rfc). Other implementations include [c-lightning](https://github.com/ElementsProject/lightning), [lnd](https://github.com/LightningNetwork/lnd), [electrum](https://github.com/spesmilo/electrum/), and [rust-lightning](https://github.com/rust-bitcoin/rust-lightning).
 
 ---
-  * [Lightning Network Specification Compliance](#lightning-network-specification-compliance)
-  * [JSON API](#json-api)
-  * [Documentation](#documentation)
-  * [Installation](#installation)
-    * [Prerequisite: Bitcoin Core](#prerequisite-bitcoin-core)
-    * [Installing Eclair](#installing-eclair)
-  * [Configuration](#configuration)
-    * [Configuration file](#configuration-file)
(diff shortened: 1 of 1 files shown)


Kixunil:
We also had a bunch of issues in `electrs` because of people reading outdated/low-quality guides, although thankfully not as catastrophic as #1926. We ended up adding big warnings to several places saying people shouldn't blindly follow other guides than our own. I can't say how effective it is because we only added it a few days ago but I think adding such warning here shouldn't hurt.

t-bast:
Good idea, I'll add that!

README.md
```diff
@@ -97,6 +98,12 @@ eclair-node-<version>-<commit_id>/bin/eclair-node.sh
 
 You can then control your node via the [eclair-cli](https://github.com/ACINQ/eclair/wiki/Usage) or the [API](https://github.com/ACINQ/eclair/wiki/API).
 
+This repository also contains a sample GUI ([eclair-node-gui](https://github.com/ACINQ/eclair/tree/master/eclair-node-gui)).
+
+:warning: The sample GUI should only be used for prototyping and demo purposes: it is unsafe to use it on mainnet.
```
pm47, line 103:
```suggestion
:rotating_light: The sample GUI should only be used for prototyping and demo purposes: using it on mainnet is unsafe.
```

pm47 (commented):
How about completely dropping the gui in eclair 0.7?

t-bast:
> How about completely dropping the gui in eclair 0.7?

I'm down with that! Kill the old witch!
Then I'll only keep the warning about obsolete guides in this PR.

README.md
t-bast: Done in https://github.com/ACINQ/eclair/pull/1929/commits/7474d3ce32e45f162e7b82ada0ab05396158eb9e

pm47 (approved):

PR #1930, opened 2021-08-31
AB testing

thomash-acinq:
Enable doing AB testing to find the best path finding parameters.

When a new payment needs to use path finding, it will be assigned randomly an experiment (which may be the control experiment) and use its set of parameters. When the payment finishes (either with a success or failure), we record metrics such as the fee we paid and the time the payment took.
After sending enough payments we can compute the success rate, average fee and average duration of payments for a given set of parameters and choose the best one.

Each experiments is assigned some percentage of the traffic. It is advised to start low with new experiments in case the new parameters are terrible but should then be increased to collect enough data timely. Experiments can be at 0% in which case they will never be used by live relay traffic but can still be used manually using the API.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -254,6 +254,7 @@ eclair {
       // One section per experiment
       // Each experiment must have a percentage of the traffic it affects (must sum to 100 for all experiments) and may
       // have other values that override the default ones.
+        // The control experiment is mandatory
         control {
           percentage = 98
         }
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -247,8 +247,10 @@ eclair {
             channel-age = 0
             channel-capacity = 0
           }
-          hop-cost-base-msat = 0
-          hop-cost-millionths = 0
+          hop-cost {
+            fee-base-msat = 0
+            fee-proportional-millionths = 0
+          }
         }
 
         // To optimize for shorter paths:
@@ -259,8 +261,11 @@ eclair {
             channel-age = 0
             channel-capacity = 0
           }
-          hop-cost-base-msat = 10000 // High hop cost penalizes strongly longer paths
-          hop-cost-millionths = 10000
+          hop-cost {
+            // High hop cost penalizes strongly longer paths
+            fee-base-msat = 10000
+            fee-proportional-millionths = 10000
+          }
         }
 
         // To optimize for successful payments:
@@ -271,8 +276,11 @@ eclair {
             channel-age = 0.5 // Old channels should have less risk of failures
             channel-capacity = 0.5 // High capacity channels are more likely to have enough liquidity to relay our payment
           }
-          hop-cost-base-msat = 1000 // Less hops means less chances of failures
-          hop-cost-millionths = 1000
+          hop-cost {
+            // Less hops means less chances of failures
+            fee-base-msat = 1000
+            fee-proportional-millionths = 1000
+          }
         }
 
         // To optimize for fast payments:
@@ -283,8 +291,11 @@ eclair {
             channel-age = 0.3 // Older channels are more likely to run smoothly
             channel-capacity = 0
           }
-          hop-cost-base-msat = 10000 // Shorter paths should be faster
-          hop-cost-millionths = 10000
+          hop-cost {
+            // Shorter paths should be faster
+            fee-base-msat = 10000
+            fee-proportional-millionths = 10000
+          }
         }
       }
     }
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -249,6 +249,29 @@ eclair {
         min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
         max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
       }
+
+      ab-testing {
+      // One section per experiment
+      // Each experiment must have a percentage of the traffic it affects (must sum to 100 for all experiments) and may
+      // have other values that override the default ones.
+        control {
+          percentage = 98
+        }
+        test-no-weights {
+          percentage = 1
+
+          ratio-base = 1
+          ratio-cltv = 0
+          ratio-channel-age = 0
+          ratio-channel-capacity = 0
+        }
+        test-no-hop-cost {
+          percentage = 1
+
+          hop-cost-base-msat = 0
+          hop-cost-millionths = 0
+        }
+      }
     }
   }
 
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -191,89 +191,86 @@ eclair {
       channel-query-chunk-size = 100 // max number of short_channel_ids in query_short_channel_ids *do not change this unless you know what you are doing*
     }
 
-    default-path-finding-params {
-      randomize-route-selection = true // when computing a route for a payment we randomize the final selection
-
-      max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
-      max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
-      fee-threshold-sat = 21       // if fee is below this value we skip the max-fee-pct check
-      max-fee-pct = 0.03           // route will be discarded if fee is above this value (in percentage relative to the total payment amount); doesn't apply if fee < fee-threshold-sat
-
-      // channel 'weight' is computed with the following formula: (channelFee + hop-cost) * (ratio-base + cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity)
-      // the following parameters can be used to ask the router to use heuristics to find i.e: 'cltv-optimized' routes, **the sum of the four ratios must be 1**
-      ratio-base = 0.0
-      ratio-cltv = 0.05             // when computing the weight for a channel, consider its CLTV delta in this proportion
-      ratio-channel-age = 0.4       // when computing the weight for a channel, consider its AGE in this proportion
-      ratio-channel-capacity = 0.55 // when computing the weight for a channel, consider its CAPACITY in this proportion
-
-      // Virtual fee for additional hops
-      // Corresponds to how much you are willing to pay to get one less hop in the payment path
-      hop-cost-base-msat = 500
-      hop-cost-millionths = 200
-
-      // Examples:
-      // To optimize for fees only:
-      // ratio-base = 1
-      // ratio-cltv = 0
-      // ratio-channel-age = 0
-      // ratio-channel-capacity = 0
-      // hop-cost-base-msat = 0
-      // hop-cost-millionths = 0
-      //
-      // To optimize for shorter paths:
-      // ratio-base = 1
-      // ratio-cltv = 0
-      // ratio-channel-age = 0
-      // ratio-channel-capacity = 0
-      // hop-cost-base-msat = 10000 // High hop cost penalizes strongly longer paths
-      // hop-cost-millionths = 10000
-      //
-      // To optimize for successful payments:
-      // ratio-base = 0
-      // ratio-cltv = 0
-      // ratio-channel-age = 0.5 // Old channels should have less risk of failures
-      // ratio-channel-capacity = 0.5 // High capacity channels are more likely to have enough liquidity to relay our payment
-      // hop-cost-base-msat = 1000 // Less hops means less chances of failures
-      // hop-cost-millionths = 1000
-      //
-      // To optimize for fast payments:
-      // ratio-base = 0.2
-      // ratio-cltv = 0.5 // In case of failure we want our funds back as fast as possible
-      // ratio-channel-age = 0.3 // Older channels are more likely to run smoothly
-      // ratio-channel-capacity = 0
-      // hop-cost-base-msat = 10000 // Shorter paths should be faster
-      // hop-cost-millionths = 10000
-
-      mpp {
-	    min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
-	    max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
-      }
-    }
-
     // The values below will be used to perform route searching
     // One section per experiment
     // Each experiment must have a percentage of the traffic it affects (must sum to 100 for all experiments).
     // After starting an experiment, if you want to modify it you should give it a new name. If you don't, then you'll
     // end up with metrics for different sets of parameters under the same name.
     path-finding {
-      control = ${eclair.router.default-path-finding-params} {
-        percentage = 98
-      }
+      default {
+        randomize-route-selection = true // when computing a route for a payment we randomize the final selection
 
-      test-no-weights = ${eclair.router.default-path-finding-params} {
-        percentage = 2
+        max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
+        max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
+        fee-threshold-sat = 21       // if fee is below this value we skip the max-fee-pct check
+        max-fee-pct = 0.03           // route will be discarded if fee is above this value (in percentage relative to the total payment amount); doesn't apply if fee < fee-threshold-sat
 
-        ratio-base = 1
-        ratio-cltv = 0
-        ratio-channel-age = 0
-        ratio-channel-capacity = 0
-      }
+        // channel 'weight' is computed with the following formula: (channelFee + hop-cost) * (ratio-base + cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity)
+        // the following parameters can be used to ask the router to use heuristics to find i.e: 'cltv-optimized' routes, **the sum of the four ratios must be 1**
+        ratio-base = 0.0
+        ratio-cltv = 0.05             // when computing the weight for a channel, consider its CLTV delta in this proportion
+        ratio-channel-age = 0.4       // when computing the weight for a channel, consider its AGE in this proportion
+        ratio-channel-capacity = 0.55 // when computing the weight for a channel, consider its CAPACITY in this proportion
 
-      test-no-hop-cost = ${eclair.router.default-path-finding-params} {
+        // Virtual fee for additional hops
+        // Corresponds to how much you are willing to pay to get one less hop in the payment path
+        hop-cost-base-msat = 500
+        hop-cost-millionths = 200
+
+        mpp {
+          min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
+          max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
+        }
+
         percentage = 0
+      }
 
-        hop-cost-base-msat = 0
-        hop-cost-millionths = 0
+      experiments {
+        control = ${eclair.router.path-finding.default} {
+          percentage = 100 // All traffic use the default configuration
+        }
+
+        // Examples of other configs as 0% experiments:
+
+        // To optimize for fees only:
+        test-fees-only = ${eclair.router.path-finding.default} {
+          ratio-base = 1
+          ratio-cltv = 0
+          ratio-channel-age = 0
+          ratio-channel-capacity = 0
+          hop-cost-base-msat = 0
+          hop-cost-millionths = 0
+        }
+
+        // To optimize for shorter paths:
+        test-short-paths = ${eclair.router.path-finding.default} {
+          ratio-base = 1
+          ratio-cltv = 0
+          ratio-channel-age = 0
+          ratio-channel-capacity = 0
+          hop-cost-base-msat = 10000 // High hop cost penalizes strongly longer paths
+          hop-cost-millionths = 10000
+        }
+
+        // To optimize for successful payments:
+        test-pay-safe = ${eclair.router.path-finding.default} {
+          ratio-base = 0
+          ratio-cltv = 0
+          ratio-channel-age = 0.5 // Old channels should have less risk of failures
+          ratio-channel-capacity = 0.5 // High capacity channels are more likely to have enough liquidity to relay our payment
+          hop-cost-base-msat = 1000 // Less hops means less chances of failures
+          hop-cost-millionths = 1000
+        }
+
+        // To optimize for fast payments:
+        test-pay-fast = ${eclair.router.path-finding.default} {
+          ratio-base = 0.2
+          ratio-cltv = 0.5 // In case of failure we want our funds back as fast as possible
+          ratio-channel-age = 0.3 // Older channels are more likely to run smoothly
+          ratio-channel-capacity = 0
+          hop-cost-base-msat = 10000 // Shorter paths should be faster
+          hop-cost-millionths = 10000
+        }
       }
     }
   }
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -203,8 +203,8 @@ eclair {
         boundaries {
           max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
           max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
-          // the route must have a fee that's below max-fee-sat OR below max-fee-proportional * total-amount
-          max-fee-sat = 21
+          // the route must have a fee that's below max-fee-fixed-sat OR below max-fee-proportional * total-amount
+          max-fee-fixed-sat = 21
           max-fee-proportional = 0.03
         }
 
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -219,8 +219,10 @@ eclair {
 
         // Virtual fee for additional hops
         // Corresponds to how much you are willing to pay to get one less hop in the payment path
-        hop-cost-base-msat = 500
-        hop-cost-millionths = 200
+        hop-cost {
+          fee-base-msat = 500
+          fee-proportional-millionths = 200
+        }
 
         mpp {
           min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -191,87 +191,87 @@ eclair {
       channel-query-chunk-size = 100 // max number of short_channel_ids in query_short_channel_ids *do not change this unless you know what you are doing*
     }
 
-    // the values below will be used to perform route searching
+    // The values below will be used to perform route searching
+    // One section per experiment
+    // Each experiment must have a percentage of the traffic it affects (must sum to 100 for all experiments).
+    // After starting an experiment, if you want to modify it you should give it a new name. If you don't, then you'll
+    // end up with metrics for different sets of parameters under the same name.
     path-finding {
-      randomize-route-selection = true // when computing a route for a payment we randomize the final selection
+      control {
+        percentage = 98
 
-      max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
-      max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
-      fee-threshold-sat = 21       // if fee is below this value we skip the max-fee-pct check
-      max-fee-pct = 0.03           // route will be discarded if fee is above this value (in percentage relative to the total payment amount); doesn't apply if fee < fee-threshold-sat
+        randomize-route-selection = true // when computing a route for a payment we randomize the final selection
 
-      // channel 'weight' is computed with the following formula: (channelFee + hop-cost) * (ratio-base + cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity)
-      // the following parameters can be used to ask the router to use heuristics to find i.e: 'cltv-optimized' routes, **the sum of the four ratios must be 1**
-      ratio-base = 0.0
-      ratio-cltv = 0.05             // when computing the weight for a channel, consider its CLTV delta in this proportion
-      ratio-channel-age = 0.4       // when computing the weight for a channel, consider its AGE in this proportion
-      ratio-channel-capacity = 0.55 // when computing the weight for a channel, consider its CAPACITY in this proportion
+        max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
+        max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
+        fee-threshold-sat = 21       // if fee is below this value we skip the max-fee-pct check
+        max-fee-pct = 0.03           // route will be discarded if fee is above this value (in percentage relative to the total payment amount); doesn't apply if fee < fee-threshold-sat
 
-      // Virtual fee for additional hops
-      // Corresponds to how much you are willing to pay to get one less hop in the payment path
-      hop-cost-base-msat = 500
-      hop-cost-millionths = 200
+        // channel 'weight' is computed with the following formula: (channelFee + hop-cost) * (ratio-base + cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity)
+        // the following parameters can be used to ask the router to use heuristics to find i.e: 'cltv-optimized' routes, **the sum of the four ratios must be 1**
+        ratio-base = 0.0
+        ratio-cltv = 0.05             // when computing the weight for a channel, consider its CLTV delta in this proportion
+        ratio-channel-age = 0.4       // when computing the weight for a channel, consider its AGE in this proportion
+        ratio-channel-capacity = 0.55 // when computing the weight for a channel, consider its CAPACITY in this proportion
 
-      // Examples:
-      // To optimize for fees only:
-      // ratio-base = 1
-      // ratio-cltv = 0
-      // ratio-channel-age = 0
-      // ratio-channel-capacity = 0
-      // hop-cost-base-msat = 0
-      // hop-cost-millionths = 0
-      //
-      // To optimize for shorter paths:
-      // ratio-base = 1
-      // ratio-cltv = 0
-      // ratio-channel-age = 0
-      // ratio-channel-capacity = 0
-      // hop-cost-base-msat = 10000 // High hop cost penalizes strongly longer paths
-      // hop-cost-millionths = 10000
-      //
-      // To optimize for successful payments:
-      // ratio-base = 0
-      // ratio-cltv = 0
-      // ratio-channel-age = 0.5 // Old channels should have less risk of failures
-      // ratio-channel-capacity = 0.5 // High capacity channels are more likely to have enough liquidity to relay our payment
-      // hop-cost-base-msat = 1000 // Less hops means less chances of failures
-      // hop-cost-millionths = 1000
-      //
-      // To optimize for fast payments:
-      // ratio-base = 0.2
-      // ratio-cltv = 0.5 // In case of failure we want our funds back as fast as possible
-      // ratio-channel-age = 0.3 // Older channels are more likely to run smoothly
-      // ratio-channel-capacity = 0
-      // hop-cost-base-msat = 10000 // Shorter paths should be faster
-      // hop-cost-millionths = 10000
+        // Virtual fee for additional hops
+        // Corresponds to how much you are willing to pay to get one less hop in the payment path
+        hop-cost-base-msat = 500
+        hop-cost-millionths = 200
 
-      mpp {
-        min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
-        max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
+        // Examples:
+        // To optimize for fees only:
+        // ratio-base = 1
+        // ratio-cltv = 0
+        // ratio-channel-age = 0
+        // ratio-channel-capacity = 0
+        // hop-cost-base-msat = 0
+        // hop-cost-millionths = 0
+        //
+        // To optimize for shorter paths:
+        // ratio-base = 1
+        // ratio-cltv = 0
+        // ratio-channel-age = 0
+        // ratio-channel-capacity = 0
+        // hop-cost-base-msat = 10000 // High hop cost penalizes strongly longer paths
+        // hop-cost-millionths = 10000
+        //
+        // To optimize for successful payments:
+        // ratio-base = 0
+        // ratio-cltv = 0
+        // ratio-channel-age = 0.5 // Old channels should have less risk of failures
+        // ratio-channel-capacity = 0.5 // High capacity channels are more likely to have enough liquidity to relay our payment
+        // hop-cost-base-msat = 1000 // Less hops means less chances of failures
+        // hop-cost-millionths = 1000
+        //
+        // To optimize for fast payments:
+        // ratio-base = 0.2
+        // ratio-cltv = 0.5 // In case of failure we want our funds back as fast as possible
+        // ratio-channel-age = 0.3 // Older channels are more likely to run smoothly
+        // ratio-channel-capacity = 0
+        // hop-cost-base-msat = 10000 // Shorter paths should be faster
+        // hop-cost-millionths = 10000
+
+        mpp {
+          min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
+          max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
+        }
       }
 
-      ab-testing {
-      // One section per experiment
-      // Each experiment must have a percentage of the traffic it affects (must sum to 100 for all experiments) and may
-      // have other values that override the default ones.
-        // The control experiment is mandatory
-        control {
-          percentage = 98
-        }
-        test-no-weights {
-          percentage = 1
+      test-no-weights = ${eclair.router.path-finding.control} {
+        percentage = 2
 
-          ratio-base = 1
-          ratio-cltv = 0
-          ratio-channel-age = 0
-          ratio-channel-capacity = 0
-        }
-        test-no-hop-cost {
-          percentage = 1
+        ratio-base = 1
+        ratio-cltv = 0
+        ratio-channel-age = 0
+        ratio-channel-capacity = 0
+      }
 
-          hop-cost-base-msat = 0
-          hop-cost-millionths = 0
-        }
+      test-no-hop-cost = ${eclair.router.path-finding.control} {
+        percentage = 0
+
+        hop-cost-base-msat = 0
+        hop-cost-millionths = 0
       }
     }
   }
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -203,9 +203,9 @@ eclair {
         boundaries {
           max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
           max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
-          // the route must have a fee that's below max-fee-fixed-sat OR below max-fee-proportional * total-amount
-          max-fee-fixed-sat = 21
-          max-fee-proportional = 0.03
+          // the route must have a fee that's below max-fee-flat-sat OR below max-fee-proportional-percent / 100 * total-amount
+          max-fee-flat-sat = 21
+          max-fee-proportional-percent = 3
         }
 
         // channel 'weight' is computed with the following formula: (channelFee + hop-cost) * (ratio-base + cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity)
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -200,17 +200,22 @@ eclair {
       default {
         randomize-route-selection = true // when computing a route for a payment we randomize the final selection
 
-        max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
-        max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
-        fee-threshold-sat = 21       // if fee is below this value we skip the max-fee-pct check
-        max-fee-pct = 0.03           // route will be discarded if fee is above this value (in percentage relative to the total payment amount); doesn't apply if fee < fee-threshold-sat
+        boundaries {
+          max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
+          max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
+          // the route must have a fee that's below max-fee-sat OR below max-fee-proportional * total-amount
+          max-fee-sat = 21
+          max-fee-proportional = 0.03
+        }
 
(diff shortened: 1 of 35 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -122,7 +122,7 @@ object NodeRelay {
   def computeRouteParams(nodeParams: NodeParams, amountIn: MilliSatoshi, expiryIn: CltvExpiry, amountOut: MilliSatoshi, expiryOut: CltvExpiry): RouteParams = {
     val routeMaxCltv = expiryIn - expiryOut
     val routeMaxFee = amountIn - amountOut
-    RouteCalculation.getDefaultRouteParams(nodeParams.routerConf.pathFindingConf).copy(
+    RouteCalculation.getDefaultRouteParams(nodeParams.routerConf.pathFindingExperimentConf.get()).copy(
```
pm47, line 125: The way Phoenix works is that it will make several retries with increased fees if no route can be found. I think it make sense to use the same algo parameters for the subsequent retries, otherwise interpretation might be difficult.

We could derive the experiment from the `payment_hash`, or a combination of `payment_hash` and `payment secret`, if we want separate payment attempts (initiated by the user) to use different parameters.

pm47 (commented):
I like that it's a pretty simple change. Haven't reviewed the metrics+db part yet.

Note to self: requires updating `eclair-front`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -230,4 +230,6 @@ object PaymentFailure {
     }))
   }
 
-}
\ No newline at end of file
+}
+
+case class ExperimentMetrics(amount: MilliSatoshi, fees: MilliSatoshi, success: Boolean, duration: Long, timestamp: Long, experimentName: String)
```
pm47, line 235: The implicit assumption is that experiments are immutable, since we only persist the experiment name.

The name is too generic though, how about:

```suggestion
case class PathFindingExperimentMetrics(amount: MilliSatoshi, fees: MilliSatoshi, success: Boolean, duration: Long, timestamp: Long, experimentName: String)
```

eclair-core/src/main/scala/fr/acinq/eclair/db/AuditDb.scala
```diff
@@ -41,6 +41,8 @@ trait AuditDb extends Closeable {
 
   def addChannelUpdate(channelUpdateParametersChanged: ChannelUpdateParametersChanged): Unit
 
+  def addExperimentMetrics(metrics: ExperimentMetrics): Unit
```
pm47, line 44:
```suggestion
  def addPathFindingExperimentMetrics(metrics: PathFindingExperimentMetrics): Unit
```

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -83,6 +83,13 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
         statement.executeUpdate("CREATE INDEX channel_updates_timestamp_idx ON audit.channel_updates(timestamp)")
       }
 
+      def migration89(statement: Statement): Unit = {
+        statement.executeUpdate("CREATE TABLE audit.metrics (amount_msat BIGINT NOT NULL, fees_msat BIGINT NOT NULL, success BOOLEAN NOT NULL, duration_ms BIGINT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL, experiment_name TEXT NOT NULL)")
```
pm47, line 87: The name is too generic:
```suggestion
        statement.executeUpdate("CREATE TABLE audit.path_finding_metrics (amount_msat BIGINT NOT NULL, fees_msat BIGINT NOT NULL, success BOOLEAN NOT NULL, duration_ms BIGINT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL, experiment_name TEXT NOT NULL)")
```

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -91,6 +91,13 @@ class SqliteAuditDb(sqlite: Connection) extends AuditDb with Logging {
       statement.executeUpdate("CREATE INDEX channel_updates_timestamp_idx ON channel_updates(timestamp)")
     }
 
+    def migration67(statement: Statement): Unit = {
+      statement.executeUpdate("CREATE TABLE metrics (amount_msat INTEGER NOT NULL, fees_msat INTEGER NOT NULL, success INTEGER NOT NULL, duration_ms INTEGER NOT NULL, timestamp INTEGER NOT NULL, experiment_name TEXT NOT NULL)")
```
pm47, line 95:
```suggestion
      statement.executeUpdate("CREATE TABLE path_finding_metrics (amount_msat INTEGER NOT NULL, fees_msat INTEGER NOT NULL, success INTEGER NOT NULL, duration_ms INTEGER NOT NULL, timestamp INTEGER NOT NULL, experiment_name TEXT NOT NULL)")
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -362,6 +362,7 @@ object PaymentInitiator {
                                paymentRequest: Option[PaymentRequest],
                                storeInDb: Boolean, // e.g. for trampoline we don't want to store in the DB when we're relaying payments
                                publishEvent: Boolean,
+                               recordMetrics: Boolean,
```
pm47, line 365: Need to clarify why we may not want to publish metrics (probably same reason as for `publishEvent`? needs a scaladoc).

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -444,7 +446,7 @@ object Router {
 
   case class MultiPartParams(minPartAmount: MilliSatoshi, maxParts: Int)
 
-  case class RouteParams(randomize: Boolean, maxFeeBase: MilliSatoshi, maxFeePct: Double, routeMaxLength: Int, routeMaxCltv: CltvExpiryDelta, ratios: WeightRatios, mpp: MultiPartParams, includeLocalChannelCost: Boolean) {
+  case class RouteParams(randomize: Boolean, maxFeeBase: MilliSatoshi, maxFeePct: Double, routeMaxLength: Int, routeMaxCltv: CltvExpiryDelta, ratios: WeightRatios, mpp: MultiPartParams, includeLocalChannelCost: Boolean, experimentName: String) {
```
pm47, line 449: I'm not sure what the relationship between `PathFindingConf` and `RouteParams` is?

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
pm47: Also let's use pg types:

```suggestion
        statement.executeUpdate("CREATE TABLE audit.metrics (amount_msat BIGINT NOT NULL, fees_msat BIGINT NOT NULL, success BOOLEAN NOT NULL, duration INTERVAL NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL, experiment_name TEXT NOT NULL)")
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
thomash-acinq: We can add the experiment name in relevant parts of the logs to help debugging. I don't think that using different experiments for subsequent retries is problematic. I'm afraid that always forcing the same experiment for a given payment hash may distort the metrics if someone tries the same payment many times and it keeps failing, if we choose the experiment randomly each time then the effect of such a payment would be spread to all experiments.

eclair-core/src/main/resources/reference.conf
```diff
@@ -191,63 +191,87 @@ eclair {
...
-      // ratio-cltv = 0.5 // In case of failure we want our funds back as fast as possible
-      // ratio-channel-age = 0.3 // Older channels are more likely to run smoothly
-      // ratio-channel-capacity = 0
-      // hop-cost-base-msat = 10000 // Shorter paths should be faster
-      // hop-cost-millionths = 10000
-
-      mpp {
-        min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
-        max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
+      control {
```
pm47, line 200: I think not defining a `default` makes things harder to understand. It's not obvious that the `control` is also the `default`. Also that means experiments _may_ override settings, but _must_ override the percentage.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -83,6 +84,13 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
         statement.executeUpdate("CREATE INDEX channel_updates_timestamp_idx ON audit.channel_updates(timestamp)")
       }
 
+      def migration89(statement: Statement): Unit = {
+        statement.executeUpdate("CREATE TABLE audit.path_finding_metrics (amount_msat BIGINT NOT NULL, fees_msat BIGINT NOT NULL, success BOOLEAN NOT NULL, duration INTERVAL NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL, experiment_name TEXT NOT NULL)")
```
pm47: Using the `interval` data type was a bad suggestion, sorry. For example both queries below return `0`:

```sql
select extract('millisecond' from interval '60 seconds')
select date_part('milliseconds', interval '60 seconds')
```

eclair-core/src/main/scala/fr/acinq/eclair/router/PathFindingExperimentConf.scala
```diff
@@ -0,0 +1,21 @@
+package fr.acinq.eclair.router
+
+import fr.acinq.eclair.router.Router.RouteParams
+
+import scala.util.Random
+
+case class PathFindingExperimentConf(experiments: Map[String, RouteParams]) {
+  private val confByPercentile: Array[RouteParams] = experiments.values.flatMap(e => Array.fill(e.experimentPercentage)(e)).toArray
```
pm47, line 8: Nice one!

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
thomash-acinq: Changed the names. But kept `duration_ms` as intervals are too inconvenient.

eclair-core/src/main/resources/reference.conf
```diff
@@ -249,6 +249,30 @@ eclair {
         min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
         max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
       }
+
+      ab-testing {
+      // One section per experiment
+      // Each experiment must have a percentage of the traffic it affects (must sum to 100 for all experiments) and may
+      // have other values that override the default ones.
+        // The control experiment is mandatory
```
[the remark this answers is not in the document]
thomash-acinq: I've reorganized it to use substitution, putting all the defaults together and the experiments override some of the defaults.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -444,7 +422,16 @@ object Router {
...
+  case class RouteParams(randomize: Boolean,
+                         maxFeeBase: MilliSatoshi,
+                         maxFeePct: Double,
+                         routeMaxLength: Int,
+                         routeMaxCltv: CltvExpiryDelta,
+                         ratios: WeightRatios,
+                         mpp: MultiPartParams,
+                         includeLocalChannelCost: Boolean,
+                         experimentName: String,
+                         experimentPercentage: Int) {
```
pm47, line 434: We can't just merge `RouteParams` and `PathFindingConf`, because they do not have the same meaning, as the special handling of `includeLocalChannelCost` shows. `PathFindingConf` is static but `RouteParams` depends on a particular route request.

How about (cc @t-bast interested in your opinion):

```scala
  case class PathFindingConf(randomizeRouteSelection: Boolean,
                             boundaries: SearchBoundaries,
                             weights: WeightRatios,
                             mpp: MultiPartParams)

  // new class that groups some parameters, not sure about Search
  case class SearchBoundaries(maxFeeBase: MilliSatoshi,
                              maxFeePct: Double,
                              maxRouteLength: Int,
                              maxCltv: CltvExpiryDelta)

  case class RouteParams(conf: PathFindingConf,
                         includeLocalChannelCost: Boolean)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: @t-bast WDYT using log context to track experiments? Never debugged the algo so not sure what is more convenient

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: If we go with this, it would make sense to create subsections:

```
    defaults {
      randomize-route-selection = true // when computing a route for a payment we randomize the final selection

      boundaries {
        max-route-length = 6
        max-cltv = 1008
        fee-threshold-sat = 21
        max-fee-pct = 0.03
      }

      ratios {
        base = 0.0
        cltv = 0.05
        channel-age = 0.4
        channel-capacity = 0.55
        hop-cost-base-msat = 500
        hop-cost-millionths = 200
      }

      mpp {
	    min-amount-satoshis = 15000
	    max-parts = 5
      }
    }
```

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
thomash-acinq: Done, there is now a default that is not an experiment itself. The default has `percentage = 0` so experiments that don't override `experiment` will just be 0% experiments (can be triggered manually but doesn't impact live traffic).

eclair-core/src/main/resources/reference.conf
```diff
@@ -191,63 +191,87 @@ eclair {
...
-      // ratio-channel-age = 0.3 // Older channels are more likely to run smoothly
-      // ratio-channel-capacity = 0
-      // hop-cost-base-msat = 10000 // Shorter paths should be faster
-      // hop-cost-millionths = 10000
-
-      mpp {
-        min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
-        max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
+      control {
+        percentage = 98
```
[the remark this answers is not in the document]
thomash-acinq: Done, control has 100%

eclair-core/src/main/resources/reference.conf
```diff
@@ -191,63 +191,87 @@ eclair {
...
+        // hop-cost-base-msat = 1000 // Less hops means less chances of failures
+        // hop-cost-millionths = 1000
+        //
+        // To optimize for fast payments:
+        // ratio-base = 0.2
+        // ratio-cltv = 0.5 // In case of failure we want our funds back as fast as possible
+        // ratio-channel-age = 0.3 // Older channels are more likely to run smoothly
+        // ratio-channel-capacity = 0
+        // hop-cost-base-msat = 10000 // Shorter paths should be faster
+        // hop-cost-millionths = 10000
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
thomash-acinq: I've changed `PathFindingConf` to be a nested case class as you suggested but I kept `RouteParams` as a separate flat case class because many of its fields are modified in many places and it's not possible to modify nested fields in a nice way.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -282,24 +277,37 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
+      val fees = result match {
+        case paymentSent: PaymentSent => paymentSent.feesPaid
+        case _ => 0 msat
+      }
+      val success = result.isInstanceOf[PaymentSent]
+      val now = System.currentTimeMillis
+      val duration = now - start
+      val experimentName = request match {
+        case SendPaymentToNode(_, _, _, _, _, routeParams) => routeParams.experimentName
+        case SendPaymentToRoute(_, _, _, _) => ""
```
t-bast, line 302: Why don't you explicitly skip these and only publish events in the `SendPaymentToNode` case?
Since in other places you explicitly try to disable it, it would make more sense, wouldn't it?

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/PathFinding.scala
```diff
@@ -29,26 +29,26 @@ trait PathFinding {
   import fr.acinq.eclair.api.serde.JsonSupport.{formats, marshaller, serialization}
 
   val findRoute: Route = postRequest("findroute") { implicit t =>
-    formFields(invoiceFormParam, amountMsatFormParam.?) {
-      case (invoice@PaymentRequest(_, Some(amount), _, nodeId, _, _), None) =>
-        complete(eclairApi.findRoute(nodeId, amount, invoice.routingInfo))
-      case (invoice, Some(overrideAmount)) =>
-        complete(eclairApi.findRoute(invoice.nodeId, overrideAmount, invoice.routingInfo))
+    formFields(invoiceFormParam, amountMsatFormParam.?, "experimentName".as[String].?) {
```
t-bast, line 32: It's probably worth adding a couple tests to `ApiServiceTests` just to check that we correctly forward the given `experimentName` to the eclair instance?

t-bast (commented):
Great feature, I really like the way it's presented and configurable in `reference.conf`.
There are important details about what exact events we should measure that we need to discuss, the details can be subtle and may highly influence the resulting data.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
t-bast: I think it would be hard to ensure we use the same experiment for retries, we'd need some state in the `NodeRelayer` which I'd like to avoid...

I think it's not an issue that the retries use a different experiment, as long as our events contain the "fee budget" allocated to the path-finding. This way we can compare experiments that have approximately the same fee budget (it indeed doesn't make much sense to compare two events if the fees allocated to the path-finding were widely different).

eclair-core/src/main/resources/reference.conf
```diff
@@ -191,63 +191,99 @@ eclair {
...
-        max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
+      default {
+        randomize-route-selection = true // when computing a route for a payment we randomize the final selection
+
+        boundaries {
+          max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
+          max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
+          // the route must have a fee that's below max-fee-sat OR below max-fee-proportional * total-amount
+          max-fee-sat = 21
+          max-fee-proportional = 0.03
```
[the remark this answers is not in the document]
thomash-acinq: I disagree, there is no percent involved here.
The old name was actually `max-fee-pct` and I thought that `max-fee-pct = 0.03` meant a maximum fee of 0.03% of the amount, but no it's actually 3%, not 0.03%. That's why I've changed the name of the field to reflect what it actually is.
I think the unit should always be visible (either in the name or in the type) but in this case this is a unit-less quantity.

We could also decide to use hundredths (a.k.a. percents) or millionths but then the value should be multiplied by 100 or 1000000 to match the unit.

eclair-core/src/main/resources/reference.conf
```diff
@@ -191,63 +191,99 @@ eclair {
...
+          base = 0.0
+          cltv = 0.05             // when computing the weight for a channel, consider its CLTV delta in this proportion
+          channel-age = 0.4       // when computing the weight for a channel, consider its AGE in this proportion
+          channel-capacity = 0.55 // when computing the weight for a channel, consider its CAPACITY in this proportion
+        }
+
+        // Virtual fee for additional hops
+        // Corresponds to how much you are willing to pay to get one less hop in the payment path
+        hop-cost-base-msat = 500
+        hop-cost-millionths = 200
```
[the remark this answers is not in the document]
thomash-acinq: I'm actually doing it in the next PR, I'll put it in this one too.

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
pm47: If only we had used [bips](https://en.wikipedia.org/wiki/Basis_point) in Lightning we wouldn't be in this mess. Although I can think of another confusion that would have created ^^

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -305,21 +317,34 @@ object NodeParams extends Logging {
...
+      experimentName = name,
+      experimentPercentage = config.getInt("percentage"))
+
+
+    def getPathFindingExperimentConf(config: Config): PathFindingExperimentConf = {
+      val experiments = for ((name -> _) <- config.root.asScala;
+                             experimentConfig = config.getConfig(name))
+      yield (name -> getPathFindingConf(experimentConfig, name))
+
+      PathFindingExperimentConf(experiments.toMap)
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -444,16 +446,25 @@ object Router {
 
   case class MultiPartParams(minPartAmount: MilliSatoshi, maxParts: Int)
 
-  case class RouteParams(randomize: Boolean, maxFeeBase: MilliSatoshi, maxFeePct: Double, routeMaxLength: Int, routeMaxCltv: CltvExpiryDelta, ratios: WeightRatios, mpp: MultiPartParams, includeLocalChannelCost: Boolean) {
+  case class RouteParams(randomize: Boolean,
+                         maxFee: MilliSatoshi,
```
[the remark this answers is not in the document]
thomash-acinq: I disagree, I've found it very confusing to have `maxFeeBase`, the "base" doesn't mean anything here and it behaves very differently from all of the other things that are called `feeBase` as it is not added to the proportional fee.
But I agree that it's not a very good name as it's only part of the maxFee with `maxFeeProportional`. I've renamed it to `maxFeeFixed`.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -444,16 +446,25 @@ object Router {
...
       // The payment fee must satisfy either the flat fee or the percentage fee, not necessarily both.
-      maxFeeBase.max(amount * maxFeePct)
+      maxFee.max(amount * maxFeeProportional)
     }
   }
 
   case class Ignore(nodes: Set[PublicKey], channels: Set[ChannelDesc]) {
     // @formatter:off
     def +(ignoreNode: PublicKey): Ignore = copy(nodes = nodes + ignoreNode)
+
```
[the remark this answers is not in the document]
thomash-acinq: removed

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -112,19 +112,19 @@ trait Eclair {
 
   def receivedInfo(paymentHash: ByteVector32)(implicit timeout: Timeout): Future[Option[IncomingPayment]]
 
-  def send(externalId_opt: Option[String], amount: MilliSatoshi, invoice: PaymentRequest, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None)(implicit timeout: Timeout): Future[UUID]
+  def send(externalId_opt: Option[String], amount: MilliSatoshi, invoice: PaymentRequest, maxAttempts_opt: Option[Int] = None, feeThresholdSat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None, experimentName: Option[String] = None)(implicit timeout: Timeout): Future[UUID]
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -279,13 +279,23 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
...
+      case Some(name) => appKit.nodeParams.routerConf.pathFindingExperimentConf.getByName(name).map(_.getDefaultRouteParams)
+    }
+  }
+
+  override def findRouteBetween(sourceNodeId: PublicKey, targetNodeId: PublicKey, amount: MilliSatoshi, experimentName: Option[String], assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty)(implicit timeout: Timeout): Future[RouteResponse] = {
+    getRouteParams(experimentName) match {
+      case Some(routeParams) =>
+        val maxFee = routeParams.getMaxFee(amount)
+        (appKit.router ? RouteRequest(sourceNodeId, targetNodeId, amount, maxFee, assistedRoutes, routeParams = routeParams)).mapTo[RouteResponse]
+      case None => Future.failed(new IllegalArgumentException(s"Experiment $experimentName does not exist."))
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -306,33 +316,36 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
...
+
+        externalId_opt match {
+          case Some(externalId) if externalId.length > externalIdMaxLength => Left(new IllegalArgumentException(s"externalId is too long: cannot exceed $externalIdMaxLength characters"))
+          case _ if invoice.isExpired => Left(new IllegalArgumentException("invoice has expired"))
+          case _ => invoice.minFinalCltvExpiryDelta match {
+            case Some(minFinalCltvExpiryDelta) => Right(SendPaymentToNode(amount, invoice, maxAttempts, minFinalCltvExpiryDelta, externalId_opt, assistedRoutes = invoice.routingInfo, routeParams = routeParams))
+            case None => Right(SendPaymentToNode(amount, invoice, maxAttempts, externalId = externalId_opt, assistedRoutes = invoice.routingInfo, routeParams = routeParams))
+          }
+        }
+      case None => Left(new IllegalArgumentException(s"Experiment $experimentName does not exist."))
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -341,15 +354,18 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
...
-    (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
+    getRouteParams(experimentName) match {
+      case Some(defaultRouteParams) =>
+        val routeParams = defaultRouteParams.copy(
+          maxFeeProportional = maxFeePct_opt.getOrElse(defaultRouteParams.maxFeeProportional),
+          maxFee = feeThreshold_opt.map(_.toMilliSatoshi).getOrElse(defaultRouteParams.maxFee)
+        )
+        val sendPayment = SendSpontaneousPayment(amount, recipientNodeId, paymentPreimage, maxAttempts, externalId_opt, routeParams)
+        (appKit.paymentInitiator ? sendPayment).mapTo[UUID]
+      case None => Future.failed(new IllegalArgumentException(s"Experiment $experimentName does not exist."))
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -208,40 +207,50 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
-      Metrics.RetryFailedChannelsResult.withTag(Tags.Success, event.isRight).increment()
+    if (cfg.recordMetrics) {
+      val fees = event match {
+        case Left(paymentFailed) => 0 msat
+        case Right(paymentSent) => paymentSent.feesPaid
+      }
+      val success = event.isRight
+      val now = System.currentTimeMillis
+      val duration = now - start
+      context.system.eventStream.publish(PathFindingExperimentMetrics(cfg.recipientAmount, fees, success, duration, now, request.routeParams.experimentName))
```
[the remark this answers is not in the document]
thomash-acinq: In the end the user doesn't care what kind of failure happened. The reason I've chosen these fields and not others is that it's all the end user can see: Did it work? How much did it cost? How long did it take?
We could add metrics but I'm not sure they would be useful. The goal of AB testing is not to help us debug individual payments but to tell us if on average a given set of parameters is better than another.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
pm47: Closing this thread, follow up on "fee budget" is https://github.com/ACINQ/eclair/pull/1930/files#r704265635.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -230,4 +230,6 @@ object PaymentFailure {
     }))
   }
 
-}
\ No newline at end of file
+}
+
+case class PathFindingExperimentMetrics(amount: MilliSatoshi, fees: MilliSatoshi, success: Boolean, duration: Long, timestamp: Long, experimentName: String)
```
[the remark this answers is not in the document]
thomash-acinq: In AB testing we don't care about individual payments, we're only interested in the metrics averaged on many payments. If you want to manually try individual payments with different settings, that will be made easy with 0% experiments. But here the goal is to answer the question "I have two sets of parameters, which one should I use?" with minimal manual by just trying both on large enough samples.

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
t-bast: I agree in theory @thomash-acinq (the formula in the comment clearly shows how it should be used) but unfortunately we need to deal with context here. In lightning `proportional` is usually `proportional-millionths`, so it's misleading if we don't add more details and users will get bitten by it (and that most important thing when choosing configuration names is to minimize the risk that users get it wrong).

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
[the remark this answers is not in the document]
t-bast: I agree that `maxFeeBase` as also a bad name :grimacing: 
Something like `maxFeeFixed` or `maxFeeFlat` sounds good to me, thanks!

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
t-bast: I would agree if we had a lot of volume with a nice amount distribution, but that's not the case.

I'm afraid that just looking at the overall "it failed / it succeeded" signal without more information will make us take very misguided decisions because we would compare two experiments based on data that's not really comparable. I'm pretty sure that if we do just that, we will mostly do this loop:

- "Cool, experiment B has much better results than experiment A, let's switch it to use more than 50%"
- "Damn now it's experiment A that's better, let's switch back to experiment A using more than 50%"
- Rinse & Repeat

And in the meantime users would see failed payments that could be avoided...

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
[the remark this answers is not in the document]
t-bast: I think the issue in your argument lies in `on large enough samples`: we don't have that and won't have it in the foreseeable future. This discussion is similar to https://github.com/ACINQ/eclair/pull/1930#discussion_r704291804, let's continue on this thread to avoid splitting the discussion in two.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
t-bast: I don't know if that's unfortunate or not, but I see this A/B testing mechanism useful in the short-term *with manual work afterwards* if we have enough data in the DB. Yes it's not ideal, but I really think that's the reality of the data sets we'll work with for now.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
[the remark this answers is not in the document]
thomash-acinq: Changed fixed to flat

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
thomash-acinq: I've changed it to use percents instead.

eclair-core/src/main/resources/reference.conf
```diff
@@ -191,63 +191,101 @@ eclair {
...
+          fee-base-msat = 500
+          fee-proportional-millionths = 200
+        }
+
+        mpp {
+          min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
+          max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
+        }
+
+        percentage = 0
```
pm47, line 232:
```suggestion
        percentage = 0 // integer between 0-100 (0 = disabled)
```

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -199,7 +201,17 @@ object NodeParams extends Logging {
...
+      "router.path-finding.max-route-length" -> "router.path-finding.default.boundaries.max-route-length",
+      "router.path-finding.max-cltv" -> "router.path-finding.default.boundaries.max-cltv",
+      "router.path-finding.fee-threshold-sat" -> "router.path-finding.default.boundaries.max-fee-flat-sat",
+      "router.path-finding.max-fee-pct" -> "router.path-finding.default.boundaries.max-fee-proportional-percent",
+      "router.path-finding.ratio-base" -> "router.path-finding.default.ratios.base",
+      "router.path-finding.ratio-cltv" -> "router.path-finding.default.ratios.cltv",
+      "router.path-finding.ratio-channel-age" -> "router.path-finding.default.ratios.channel-age",
+      "router.path-finding.ratio-channel-capacity" -> "router.path-finding.default.ratios.channel-capacity",
+      "router.path-finding.hop-cost-base-msat" -> "router.path-finding.default.hop-cost-base-msat",
+      "router.path-finding.hop-cost-millionths" -> "router.path-finding.default.hop-cost-millionths",
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/resources/reference.conf
```diff
@@ -191,63 +191,101 @@ eclair {
...
-        max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
+      default {
+        randomize-route-selection = true // when computing a route for a payment we randomize the final selection
+
+        boundaries {
+          max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
+          max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
+          // the route must have a fee that's below max-fee-flat-sat OR below max-fee-proportional-percent / 100 * total-amount
+          max-fee-flat-sat = 21
+          max-fee-proportional-percent = 3
```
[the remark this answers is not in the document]
thomash-acinq: Added more explanations

eclair-core/src/main/resources/reference.conf
thomash-acinq: Added more explanations

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/PathFinding.scala
thomash-acinq: Added some tests

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
thomash-acinq: I've added `isMultiPart`. I don't think I can add the different types of failures: because of retries a single payment can fail in several different ways, how would you deal with that?

eclair-core/src/main/resources/reference.conf
```diff
@@ -191,63 +191,101 @@ eclair {
...
+        // To optimize for fees only:
+        test-fees-only = ${eclair.router.path-finding.default} {
+          ratios {
+            base = 1
+            cltv = 0
+            channel-age = 0
+            channel-capacity = 0
+          }
+          hop-cost-base-msat = 0
+          hop-cost-millionths = 0
```
t-bast, line 251: Don't you need to update the format of `hop-cost` to reflect what's in `default`?
Same for other experiments below.

```suggestion
          hop-cost {
            fee-base-msat = 0
            fee-proportional-millionths = 0
          }
```

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -149,8 +149,8 @@ class EclairImplSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with I
     assert(send3.recipientNodeId === nodePrivKey.publicKey)
     assert(send3.recipientAmount === 123.msat)
     assert(send3.paymentHash === ByteVector32.Zeroes)
-    assert(send3.routeParams.maxFeeBase === 123000.msat) // conversion sat -> msat
-    assert(send3.routeParams.maxFeePct === 4.20)
+    assert(send3.routeParams.maxFeeFlat === 123000.msat) // conversion sat -> msat
+    assert(send3.routeParams.maxFeeProportional === 4.20)
```
t-bast, line 153: This was before your PR, but I'm realizing now that the value for this test is completely absurd (420%) and users may find it very misleading...could you update that to a "normal" percentage, such as `0.15` (15%)?

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Payment.scala
```diff
@@ -38,16 +38,16 @@ trait Payment {
   }
 
   val payInvoice: Route = postRequest("payinvoice") { implicit t =>
-    formFields(invoiceFormParam, amountMsatFormParam.?, "maxAttempts".as[Int].?, "feeThresholdSat".as[Satoshi].?, "maxFeePct".as[Double].?, "externalId".?, "blocking".as[Boolean].?) {
-      case (invoice@PaymentRequest(_, Some(amount), _, nodeId, _, _), None, maxAttempts, feeThresholdSat_opt, maxFeePct_opt, externalId_opt, blocking_opt) =>
+    formFields(invoiceFormParam, amountMsatFormParam.?, "maxAttempts".as[Int].?, "feeThresholdSat".as[Satoshi].?, "maxFeePct".as[Double].?, "externalId".?, "blocking".as[Boolean].?, "pathFindingExperimentName".?) {
```
t-bast, line 41: nit: I'm realizing now that since `maxFeePct` is a misleading name (we expect `0.05` if the user wants to set if to 5%) we should verify here that it is between `0.0` and `1.0`. This is completely unrelated to this PR though, I can fix that in a separate follow-up PR if you want.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
t-bast: Unless I'm missing something, it should be pretty simple to add with the way you currently emit the events since in `MultiPartPaymentLifecycle` you set `emitMetrics = false` to the child `PaymentLifecycle` and only emit the event when the payment is complete (failed or succeeded) after all retries have happened.

We need to do the following:

- When the payment failed, we should populate the `fees` field with the fee budget (instead of setting it to `0 msat`), which you can easily do using `routeParams.getMaxFee(recipientAmount)`
- When the payment failed, we should check all the `PaymentFailure` objects it contains, and if any of these comes from the final recipient (which you know by checking if it's a `RemoteFailure` where `e.originNode` matches the recipient) we should **not** emit the event

I think that once we've done that, the data we'll have should be relevant.

Can you also add the `targetNodeId` to the `PathFindingMetrics`? I really think it can be useful, it doesn't cost much to add, we can ignore it if we don't care about it, but if for some reason we need it it will be available.

eclair-core/src/main/resources/reference.conf
thomash-acinq: You're right. That's a problem with overriding a default conf, if you make a typo it won't complain and will silently use the default.

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Payment.scala
thomash-acinq: I've changed it so that `maxFeePct` is treated as a percentage as the name implies, so `5` would actually be 5%, which is `0.05`.

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
thomash-acinq: I've changed it so that `maxFeePct` is treated as a percentage.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Take a look at `quicklens`:

```scala
import com.softwaremill.quicklens.ModifyPimp
val aliceParams2 = Alice.nodeParams.modify(_.routerConf.requestNodeAnnouncements).setTo(true)
```

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Payment.scala
pm47: Can we add the `includeLocalChannelCost` param so we can manually test in a trampoline setting (which is our primary use case for route computation)?

I think we need to rename `feeThresholdSat` and `maxFeePct` to be consistent with `reference.conf`.

t-bast: > I've changed it so that maxFeePct is treated as a percentage as the name implies, so 5 would actually be 5%, which is 0.05.

Thanks. I added this to my note of breaking changes that we really need to document in the next release!

eclair-core/src/main/resources/reference.conf
```diff
@@ -191,63 +191,112 @@ eclair {
...
+
+        boundaries {
+          max-route-length = 6         // max route length for the 'first pass', if none is found then a second pass is made with no limit
+          max-cltv = 1008              // max acceptable cltv expiry for the payment (1008 ~ 1 week)
+          // the route must have a fee that's below max-fee-flat-sat OR below max-fee-proportional-percent / 100 * total-amount
+          max-fee-flat-sat = 21
+          max-fee-proportional-percent = 3 // that's 3%
+        }
+
+        // channel 'weight' is computed with the following formula: (channelFee + hop-cost) * (ratio-base + cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity)
```
pm47, line 207: Noob question. Why isn't the formula:

`(channelFee + hop-cost) + cltvDelta * ratio-cltv + channelAge * ratio-channel-age + channelCapacity * ratio-channel-capacity`

Just summing the weights.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
thomash-acinq: Added all that.

eclair-core/src/main/resources/reference.conf
thomash-acinq: It could be, that probably makes more sense actually.

t-bast: I think the reason we currently do that is because we want the fee to have a scaling impact even when ratios for other things are high. We internally always bias towards the fee as the most important parameter (and I think it's ok).

We could have an experiment where we change that formula though!

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -282,24 +277,49 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
...
-    Metrics.SentPaymentDuration
-      .withTag(Tags.MultiPart, if (cfg.id != cfg.parentId) Tags.MultiPartType.Child else Tags.MultiPartType.Disabled)
-      .withTag(Tags.Success, value = true)
-      .record(System.currentTimeMillis - start, TimeUnit.MILLISECONDS)
-  }
-
-  private def onFailure(replyTo: ActorRef, result: PaymentFailed): Unit = {
-    if (cfg.storeInDb) paymentsDb.updateOutgoingPayment(result)
-    replyTo ! result
+  def myStop(request: SendPayment, result: PaymentEvent): State = {
```
t-bast, line 280: nit: similarly to `MultiPartPaymentLifecycle`, you could here take a `result: Either[PaymentFailed, PaymentSent]` to avoid the `INTERNAL_ERROR` case and the `case _ => ()`

t-bast (approved):
LGTM :rocket:

eclair-core/src/main/resources/reference.conf
pm47: > We could have an experiment where we change that formula though!

Right, using an additional parameter to select the weight calculation mode. I guess let's revisit this later and not delay this PR further then

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
thomash-acinq: Done

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Payment.scala
pm47: > Can we add the includeLocalChannelCost param so we can manually test in a trampoline setting (which is our primary use case for route computation)?

Done in #1942

pm47 (approved):

t-bast (approved):

PR #1932, opened 2021-09-01
Implement anchor outputs zero fee htlc txs

t-bast:
Implement https://github.com/lightningnetwork/lightning-rfc/pull/824

This is a small variation of the original anchor outputs proposal, where the fees of the htlc 2nd-stage txs are set to 0 to fix a fee inflating attack when you're fundee (see https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-September/002796.html for details).

The first commit creates the corresponding feature bit and lets peers negotiate its activation.
The second commit actually sets the htlc txs fee to 0 when this new channel type is negotiated.

I've tested cross-compatibility with lnd 0.13.1-beta (their latest release).
I verified that we were able to open channels and send payments to each other.
I also verified that HTLCs that were exactly at the dust limit were included in the commit tx (so the trimmed to dust calculation is compatible and correctly uses a 0-fee htlc tx) and htlcs just below the dust limit were not included in the commit tx.


eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -193,6 +193,11 @@ object Features {
     val mandatory = 20
   }
 
+  case object AnchorOutputsZeroFeeHtlcTxs extends Feature {
```
pm47, line 196:
```suggestion
  case object AnchorOutputsZeroFeeHtlcTx extends Feature {
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
```diff
@@ -31,26 +30,20 @@ import fr.acinq.eclair.{Feature, FeatureSupport, Features}
...
-    } else if (hasFeature(StaticRemoteKey)) {
+    } else if (hasFeature(Features.StaticRemoteKey)) {
       ChannelTypes.StaticRemoteKey
     } else {
       ChannelTypes.Standard
     }
   }
 
   val paysDirectlyToWallet: Boolean = channelType.paysDirectlyToWallet
+  val commitmentFormat: CommitmentFormat = channelType.commitmentFormat
```
pm47, line 46: For the sake of discussing: why not define two commitment formats for anchor, instead of passing around the `channelType`? Things like below make me doubt, because the commitment format should be what we use to deal with commitment-related stuff:

```scala
val commitFees = commitTxTotalCostMsat(remoteParams.dustLimit, reduced, channelType)
```

Could be:

```scala
sealed trait AnchorOutputsCommitmentFormat

case object UnsafeLegacyAnchorOutputsCommitmentFormat extends AnchorOutputsCommitmentFormat

case object ZeroFeeHtlcAnchorOutputsCommitmentFormat extends AnchorOutputsCommitmentFormat
```

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
```diff
@@ -70,11 +72,17 @@ case class IncomingHtlc(add: UpdateAddHtlc) extends DirectedHtlc
 
 case class OutgoingHtlc(add: UpdateAddHtlc) extends DirectedHtlc
 
-final case class CommitmentSpec(htlcs: Set[DirectedHtlc], feeratePerKw: FeeratePerKw, toLocal: MilliSatoshi, toRemote: MilliSatoshi) {
+final case class CommitmentSpec(htlcs: Set[DirectedHtlc], commitTxFeerate: FeeratePerKw, toLocal: MilliSatoshi, toRemote: MilliSatoshi) {
```
pm47, line 75: My gut feeling is that this class should depend on the `CommitmentFormat` (or `ChannelType` but you get the idea).

Probably okay for now.

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
```diff
@@ -239,11 +239,11 @@ private class ReplaceableTxPublisher(nodeParams: NodeParams,
   }
 
   def checkHtlcPreconditions(replyTo: ActorRef[TxPublisher.PublishTxResult], cmd: TxPublisher.PublishReplaceableTx, htlcTx: HtlcTx, targetFeerate: FeeratePerKw): Behavior[Command] = {
-    val commitFeerate = cmd.commitments.localCommit.spec.feeratePerKw
+    val htlcFeerate = cmd.commitments.localCommit.spec.htlcTxFeerate(cmd.commitments.channelType)
```
pm47, line 242: So, this is where the magic is happening right? But I don't get how it works. Why don't we go to the `fund()` branch when `targetFeerate != htlcFeerate`?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
```diff
@@ -45,6 +46,8 @@ case class FeerateTolerance(ratioLow: Double, ratioHigh: Double, anchorOutputMax
         proposedFeerate < networkFeerate * ratioLow || networkFeerate * ratioHigh < proposedFeerate
       case ChannelTypes.AnchorOutputs =>
         proposedFeerate < networkFeerate * ratioLow || anchorOutputMaxCommitFeerate * ratioHigh < proposedFeerate
+      case ChannelTypes.AnchorOutputsZeroFeeHtlcTxs =>
+        proposedFeerate < networkFeerate * ratioLow || anchorOutputMaxCommitFeerate * ratioHigh < proposedFeerate
```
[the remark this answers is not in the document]
t-bast: Nice, I'll try it out!

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
t-bast: I view `commitmentFormat` as `the format of the commit tx`, and it's the same in both cases...but I agree that the `channelType` felt a bit redundant, I'll try that.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/CommitmentSpec.scala
t-bast: The reason I didn't go that way is because changing this class means we need to change the channel codecs (and a migration)...and I really wanted to avoid that if that wasn't really necessary!

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
t-bast: I don't understand your question. What we're doing here is simply checking the current feerate of the htlc tx as it was signed by both parties during channel operation. If that feerate is >= targetFeerate, we don't need to add inputs/outputs and use the simpler flow of just signing the htlc tx and broadcasting it.

With `anchor_output_zero_fee_htlc_tx` the `htlcFeerate` here will always be `0`, so we'll always go to the `fund()` branch, but for anchor outputs without `zero_fee_htlc_tx` we may directly broadcast the htlc tx without any funding (which is better for our utxo set management).

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1932/commits/c537fcb1c68e708e0c258f7787194ee4d33ead45, it provides a cleaner separation between the `transactions` package and the `channel` package, transactions don't need to receive a channel type.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: Done.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeEstimator.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1932/commits/c537fcb1c68e708e0c258f7787194ee4d33ead45

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
pm47: Nevermind 🙃

pm47 (commented):
LGTM, just needs a rebase.

pm47 (approved):

PR #1934, opened 2021-09-02
Keep former `channel_update` in channel state

pm47:
This is very similar to #1918 (ef31de1), but we use the internal
actor state instead of the `ChannelData`.

Pros:
- conceptually simple, low risk of regression
- `ChannelData` stays untouched

Cons:
- now there is a `var` in the channel

Those are the minimal changes to have the simplest diff, but we can
include some improvements made in ef31de1.


thomash-acinq (approved):

pm47:
And.... still doesn't work 😒. For the same reasons as https://github.com/ACINQ/eclair/pull/1922#discussion_r700913262.

If we send multiple `CMD_UPDATE_RELAY_FEE` in `OFFLINE`, the `previousChannelUpdate_opt` will be overwritten.

The root cause of all our problems is that we should not reuse `LocalChannelUpdate`. The semantics of `LocalChannelUpdate` revolves around channel availability (connections/disconnections) and trying to minimize events in order to save bandwidth, by deferring some events. It's a very different from tracking changes in parameters.

PR #1935, opened 2021-09-02
Dedicated event for `channel_update` modifications

pm47:
We use an event `ChannelUpdateParametersChanged` for the sole purpose of tracking changes to `channel_update`.

First two commits are a rebase on improvements made in previous PRs (mostly simplify the restore flow by including it in the general "update while offline" scenario from #1922).

🤞


eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
```diff
@@ -38,7 +38,9 @@ case class ChannelIdAssigned(channel: ActorRef, remoteNodeId: PublicKey, tempora
 
 case class ShortChannelIdAssigned(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, previousShortChannelId: Option[ShortChannelId]) extends ChannelEvent
 
-case class LocalChannelUpdate(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, channelAnnouncement_opt: Option[ChannelAnnouncement], channelUpdate: ChannelUpdate, previousChannelUpdate_opt: Option[ChannelUpdate], commitments: AbstractCommitments) extends ChannelEvent
+case class LocalChannelUpdate(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, channelAnnouncement_opt: Option[ChannelAnnouncement], channelUpdate: ChannelUpdate, commitments: AbstractCommitments) extends ChannelEvent
+
+case class ChannelUpdateParametersChanged(channel: ActorRef, channelId: ByteVector32, shortChannelId: ShortChannelId, remoteNodeId: PublicKey, channelUpdate: ChannelUpdate) extends ChannelEvent
```
thomash-acinq, line 43: We don't need `channel: ActorRef` or `shortChannelId: ShortChannelId`.

pm47: True, I put those fields for consistency with other `channel_update`-related events, and because plugins may use them (we learnt with #1924 that some do listen to channel events).

thomash-acinq (dismissed):

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -424,6 +422,14 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
     Future.foldLeft(commands)(Map.empty[ApiTypes.ChannelIdentifier, Either[Throwable, T]])(_ + _)
   }
 
+  /** Send a request to multiple channels using node ids */
+  private def sendToNodes[T: ClassTag](nodeids: List[PublicKey], request: Any)(implicit timeout: Timeout): Future[Map[ApiTypes.ChannelIdentifier, Either[Throwable, T]]] = {
+    for {
+      channelIds <- (appKit.register ? Symbol("channelsTo")).mapTo[Map[ByteVector32, PublicKey]].map(_.filter(kv => nodeids.contains(kv._2)).keys)
```
t-bast, line 428: Note that this has a side-effect compared to calling `Router.GetLocalChannels`: you won't send to channels that are offline.

IIUC that's done on purpose here, because for the case of updating fees, we've stored it in the DB and the channel will update automatically on restore.

But I think it should be mentioned in a comment, otherwise we'll get bitten by this in the future when we use that function hoping that it sends to channels regardless of their state.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -287,32 +287,20 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
...
-            enable = Announcements.isEnabled(normal.channelUpdate.channelFlags))
-          val channelUpdate1 = if (Announcements.areSame(candidateChannelUpdate, normal.channelUpdate)) {
-            // if there was no configuration change we keep the existing channel update
-            normal.channelUpdate
-          } else {
-            log.info("refreshing channel_update due to configuration changes old={} new={}", normal.channelUpdate, candidateChannelUpdate)
-            candidateChannelUpdate
+          if (fees.feeBase != normal.channelUpdate.feeBaseMsat ||
+            fees.feeProportionalMillionths != normal.channelUpdate.feeProportionalMillionths ||
+            nodeParams.expiryDelta != normal.channelUpdate.cltvExpiryDelta) {
```
t-bast, line 294: Interesting, so now we're updating `cltv_expiry_delta` globally. I think that makes sense, I don't see why you'd have different values for different channels, and I think it's valuable to be able to update it (we previously couldn't, could we?).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1911,6 +1889,14 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
           context.system.eventStream.publish(LocalChannelDown(self, normal.commitments.channelId, normal.shortChannelId, normal.commitments.remoteParams.nodeId))
         case _ => ()
       }
+
+      (stateData, nextStateData) match {
+        // NORMAL->NORMAL, NORMAL->SYNCING, SYNCING->NORMAL
```
pm47, line 1894:
```suggestion
        // NORMAL->NORMAL, NORMAL->OFFLINE, SYNCING->NORMAL
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1816,6 +1796,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
 
     case Event(c: CMD_UPDATE_RELAY_FEE, d) => handleCommandError(CommandUnavailableInThisState(d.channelId, "updaterelayfee", stateName), c)
 
+    // at restore, if the configuration has changed, the channel will send a command to itself to update the relay fees
+    case Event(RES_SUCCESS(_: CMD_UPDATE_RELAY_FEE, channelId), d: DATA_NORMAL) if channelId == d.channelId => stay()
```
t-bast, line 1800: Can't we set `replyTo` to the deadletters actor instead of `ActorRef.noSender` to avoid that?

thomash-acinq (dismissed):

eclair-core/src/test/scala/fr/acinq/eclair/channel/RestoreSpec.scala
```diff
@@ -0,0 +1,229 @@
...
+  override def withFixture(test: OneArgTest): Outcome = {
+    val setup = test.tags.contains("disable-offline-mismatch") match {
+      case false => init()
+      case true => init(nodeParamsA = Alice.nodeParams.copy(onChainFeeConf = Alice.nodeParams.onChainFeeConf.copy(closeOnOfflineMismatch = false)))
+    }
+    import setup._
+    within(30 seconds) {
+      reachNormal(setup)
+      awaitCond(alice.stateName == NORMAL)
+      awaitCond(bob.stateName == NORMAL)
```
t-bast, line 39: nit: these aren't needed, they're already part of `reachNormal`

eclair-core/src/test/scala/fr/acinq/eclair/channel/RestoreSpec.scala
```diff
@@ -0,0 +1,229 @@
...
+    sender.send(bob, INPUT_DISCONNECTED)
+    awaitCond(alice.stateName == OFFLINE)
+    awaitCond(bob.stateName == OFFLINE)
+
+    // and we terminate Alice
+    alice.stop()
+
+    // we restart Alice with a different configuration
+    val newFees = RelayFees(765 msat, 2345)
+    val newConfig = Alice.nodeParams.copy(relayParams = RelayParams(newFees, newFees, newFees))
```
t-bast, line 211: nit: it would be interested to also change the `cltv_expiry_delta`

eclair-core/src/test/scala/fr/acinq/eclair/channel/RestoreSpec.scala
```diff
@@ -0,0 +1,229 @@
...
+    val newAlice: TestFSMRef[ChannelState, ChannelData, Channel] = TestFSMRef(new Channel(newConfig, wallet, Bob.nodeParams.nodeId, alice2blockchain.ref, relayerA.ref, FakeTxPublisherFactory(alice2blockchain)), alicePeer.ref)
+    newAlice ! INPUT_RESTORED(oldStateData)
+
+    val u1 = channelUpdateListener.expectMsgType[ChannelUpdateParametersChanged]
+    assert(!Announcements.areSameIgnoreFlags(u1.channelUpdate, oldStateData.channelUpdate))
```
t-bast, line 216: Maybe assert that we use the correct values as well (765 msat and 2345 ppm)?

t-bast (commented):
Concept ACK, a couple small comments

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
pm47: It's the opposite actually: offline channels will be returned by the `register`, but not by the `router`.

> IIUC that's done on purpose here, because for the case of updating fees, we've stored it in the DB and the channel will update automatically on restore.

It would indeed have been on _restore_, not on _reconnection_, which would have been really annoying. But that's not what happens, because we actually update channels the same way, whether they are `NORMAL`, `OFFLINE` or `SYNCING`. Where it differs is whether or not a `LocalChannelUpdate` will be sent, but that's not our concern anymore with this PR.

> But I think it should be mentioned in a comment, otherwise we'll get bitten by this in the future when we use that function hoping that it sends to channels regardless of their state.

Sending to all channels that correspond to whatever identifier(s) is/are provided is indeed the default expectation, which is why there is no specific comment. I think that would be confusing?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: No, we couldn't, it is a side effect new feature.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: We could, but we would get `WARN` logs from the `Deadletters` handlers. I think it's better to handle them explicitly, since we know they will happen, and leave `Deadletters` for unexpected messages.

eclair-core/src/test/scala/fr/acinq/eclair/channel/RestoreSpec.scala
pm47:
```suggestion
```

eclair-core/src/test/scala/fr/acinq/eclair/channel/RestoreSpec.scala
pm47: Done with 290ec5731.

eclair-core/src/test/scala/fr/acinq/eclair/channel/RestoreSpec.scala
pm47: Done with 290ec5731.

eclair-core/src/test/scala/fr/acinq/eclair/channel/RestoreSpec.scala
thomash-acinq: The behavior we're interested in should happen even if only one of the parameters change. It feels weird to test if it happens when all of them change, this test would also pass if we had a `&&` instead of a `||`.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: > Sending to all channels that correspond to whatever identifier(s) is/are provided is indeed the default expectation, which is why there is no specific comment. I think that would be confusing?

You're right, if it does send to all channels then no need for a comment, I incorrectly thought offline channels were ignored, but looking at the `ChannelRestored` event we're all good.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: I was hoping it wouldn't log anything in that case, but then let's keep it like that!

t-bast (dismissed):

eclair-core/src/test/scala/fr/acinq/eclair/channel/RestoreSpec.scala
pm47: That's a good point, I came up with 8135d11 with the goal of not making the test too complicated by testing all combinations.

pm47: Replaced randomization by iteration over various confs in be80a5b51.

thomash-acinq (approved):

PR #1936, opened 2021-09-02
Make publising of onion addresses configurable

rorp:
Currently onion addresses generated by Tor hidden service are always published in node announcements. 

In some cases it make sense to keep them private.


t-bast:
> In some cases it make sense to keep them private.

Please provide more details to motivate such PRs.

I'm having troubles understanding this: you would publish a clearnet IP address in your node announcement but don't want to publish the onion address? What's the point if people can reach you through a clearnet IP?

Or do you want to send a `node_announcement` with an empty address list because you don't want to receive any incoming connection?

rorp:
> Or do you want to send a `node_announcement` with an empty address list because you don't want to receive any incoming connection?

Yes. That's the main reason. One would be able to set up a "stealth" Tor-only node.

t-bast:
> Yes. That's the main reason. One would be able to set up a "stealth" Tor-only node, that accept connection only from nodes that know the URI.

Ok that makes sense.

Concept ACK, I see why that's needed (since Tor address is automatically resolved inside eclair and injected into `publicAddresses` without giving control to the user).

t-bast (approved):

PR #1939, opened 2021-09-06
Clean up inconsistency between bitcoin client and wallet

t-bast:
The separation between `ExtendedBitcoinClient` and `BitcoinCoreWallet` has become very blurry since anchor ouputs: eclair now requires fee bumping utilities from the underlying bitcoin wallet, and it's not yet clear what the interface should be. The on-chain utility methods that were added to the eclair API also made it awkward to cleanly separate concerns.

We completely remove the `BitcoinCoreWallet` in this PR and merge it inside the bitcoin client. We may in the future re-introduce a cleaner on-chain wallet abstraction, but that can only happen once we have stable fee bumping mechanisms (and once we can directly leverage bitcoin core's `bumpfee` RPC when it allows non-wallet inputs).

It's likely that the wallet traits that we'll (re)introduce in the future will be based on PSBTs only.


sstone (approved):
Nice cleanup of electrum leftovers!

PR #1941, opened 2021-09-07
Remove  `messageFlags` from `ChannelUpdate`

pm47:
We make it a serialization detail, which it should be. The `derive`
method from `scodec` makes it very easy to do. We should probably always
use a dedicated class to handle flags, instead of using the `byte` codec
and binary operators.

This allows to remove the `require` in the `ChannelUpdate` definition,
which recently bit us in testing.

The only annoying thing is that we still need to expose a `messageFlags`
method in order to populate the `ChannelDisabled` error message.

Took the opportunity to typeify channel flags instead of passing around a `Byte`.


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -220,35 +221,46 @@ object LightningMessageCodecs {
...
-      })
+      })).derive[MessageFlags].from {
+      // same comment above
+      case _ :: _ :: _ :: _ :: _ :: _ :: _ :: _ :: htlcMaximumMsat_opt :: _ :: HNil => MessageFlags(optionChannelHtlcMax = htlcMaximumMsat_opt.isDefined)
+    }
 
   val channelUpdateCodec: Codec[ChannelUpdate] = (
     ("signature" | bytes64) ::
-      channelUpdateWitnessCodec).as[ChannelUpdate]
+      (channelUpdateWitnessCodec)).as[ChannelUpdate]
```
pm47, line 263:
```suggestion
      channelUpdateWitnessCodec).as[ChannelUpdate]
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -253,15 +253,15 @@ case class ChannelUpdate(signature: ByteVector64,
...
                          channelFlags: Byte,
                          cltvExpiryDelta: CltvExpiryDelta,
                          htlcMinimumMsat: MilliSatoshi,
                          feeBaseMsat: MilliSatoshi,
                          feeProportionalMillionths: Long,
                          htlcMaximumMsat: Option[MilliSatoshi],
                          tlvStream: TlvStream[ChannelUpdateTlv] = TlvStream.empty) extends RoutingMessage with AnnouncementMessage with HasTimestamp with HasChainHash {
-  require(((messageFlags & 1) != 0) == htlcMaximumMsat.isDefined, "htlcMaximumMsat is not consistent with messageFlags")
+
+  def messageFlags: Byte = if (htlcMaximumMsat.isDefined) 1 else 0
```
pm47, line 264: We need this to populate the `ChannelDisabled` error message. 🙄

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -253,21 +252,41 @@ case class ChannelUpdate(signature: ByteVector64,
...
-  require(((messageFlags & 1) != 0) == htlcMaximumMsat.isDefined, "htlcMaximumMsat is not consistent with messageFlags")
 
-  def isNode1 = Announcements.isNode1(channelFlags)
+  def messageFlags: Byte = if (htlcMaximumMsat.isDefined) 1 else 0
 
   def toStringShort: String = s"cltvExpiryDelta=$cltvExpiryDelta,feeBase=$feeBaseMsat,feeProportionalMillionths=$feeProportionalMillionths"
 }
 
+object ChannelUpdate {
+  case class ChannelFlags(disable: Boolean, direction: Boolean) {
```
pm47, line 269: Those booleans could have been flattened directly in the `ChannelUpdate`, that would have generated less boiler plate but was a bit less close to the "spirit of the spec". Let me know what you think!

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -215,7 +215,7 @@ object HtlcResult {
   case class RemoteFailMalformed(fail: UpdateFailMalformedHtlc) extends Fail
   case class OnChainFail(cause: Throwable) extends Fail
   case object ChannelFailureBeforeSigned extends Fail
-  case class DisconnectedBeforeSigned(channelUpdate: ChannelUpdate) extends Fail { assert(!Announcements.isEnabled(channelUpdate.channelFlags), "channel update must have disabled flag set") }
+  case class DisconnectedBeforeSigned(channelUpdate: ChannelUpdate) extends Fail { assert(channelUpdate.channelFlags.disable, "channel update must have disabled flag set") }
```
t-bast, line 218: nit: why do we use `assert` instead of `require` here?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
```diff
@@ -70,7 +70,7 @@ object ChannelRelay {
       case (_: InsufficientFunds, Some(channelUpdate)) => TemporaryChannelFailure(channelUpdate)
       case (_: TooManyAcceptedHtlcs, Some(channelUpdate)) => TemporaryChannelFailure(channelUpdate)
       case (_: FeerateTooDifferent, Some(channelUpdate)) => TemporaryChannelFailure(channelUpdate)
-      case (_: ChannelUnavailable, Some(channelUpdate)) if !Announcements.isEnabled(channelUpdate.channelFlags) => ChannelDisabled(channelUpdate.messageFlags, channelUpdate.channelFlags, channelUpdate)
+      case (_: ChannelUnavailable, Some(channelUpdate)) if !channelUpdate.channelFlags.isEnabled => ChannelDisabled(channelUpdate.messageFlags, channelUpdate.channelFlags, channelUpdate)
```
t-bast, line 73:
```suggestion
      case (_: ChannelUnavailable, Some(channelUpdate)) if channelUpdate.channelFlags.disable => ChannelDisabled(channelUpdate.messageFlags, channelUpdate.channelFlags, channelUpdate)
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
t-bast: I agree, it's better to keep the "spirit of the spec" and keep the codec easy to read, so that we immediately see whether it correctly matches the spec or not.

I find it slightly weird though that `direction` is a `Boolean`...

t-bast:
The `derive` is pretty cool! Even though hopefully we shouldn't need it very often now that we use a tlv stream to enrich existing messages.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
pm47: Have no idea, will fix

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
pm47: Now that I think of it, even if we stick to 'spirit of the spec" we have no reason to keep the same types and names in our classes. That's what we do everywhere else, but since those are booleans, weirdly we kept them as-is, and defined redundant more-meaningful methods to interpret them. We could (should I think) define `ChannelFlags` as:

```scala
case class ChannelFlags(isEnabled: Boolean, isNode1: Boolean)
```

pm47: Done in 130f44772.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
pm47: Done in 130f44772.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
pm47: Deprecated in 130f44772.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -220,31 +221,48 @@ object LightningMessageCodecs {
     ("signature" | bytes64) ::
       nodeAnnouncementWitnessCodec).as[NodeAnnouncement]
 
+  private case class MessageFlags(optionChannelHtlcMax: Boolean)
+
+  private val messageFlagsCodec = ("messageFlags" | (ignore(7) :: bool)).as[MessageFlags]
+
+  val reverseBool: Codec[Boolean] = bool.xmap[Boolean](b => !b, b => !b)
```
t-bast, line 228: Is this a `loob` then? :trollface:

t-bast (approved):
LGTM

PR #1942, opened 2021-09-08
Add success probabilities in path finding

thomash-acinq:
Add a new way of scoring paths that estimates the success probabilities of paths and put a cost on failed payment attempts.

For now the success probabilities are estimated in a very simple way based on capacity: I assume that the balance is a random variable uniformly distributed between 0 and capacity and multiply probabilities for all the channels on the path.
The formula for estimating the probability could be changed easily, that's not the focus of this PR. The goal of this PR is to introduce a new scoring of paths that uses this probability and can be evaluated using the new AB testing framework from #1930.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -215,6 +215,7 @@ eclair {
           channel-age = 0.4       // when computing the weight for a channel, consider its AGE in this proportion
           channel-capacity = 0.55 // when computing the weight for a channel, consider its CAPACITY in this proportion
         }
+        use-ratios = true
 
         // Virtual fee for additional hops
         // Corresponds to how much you are willing to pay to get one less hop in the payment path
@@ -238,6 +239,20 @@ eclair {
       experiments {
         control = ${eclair.router.path-finding.default} {
           percentage = 100 // 100% of the traffic use the default configuration
+        }
+
+        // alternative routing heuristics (replaces ratios)
+        test-failure-cost = ${eclair.router.path-finding.default} {
+          use-ratios = false
+
+          locked-funds-risk = 1e-15 // msat per msat locked per block. It should be your expected interest rate per block multiplied by the probability that something goes wrong and your funds stay locked.
+
+          // Virtual fee for failed payments
+          // Corresponds to how much you are willing to pay to get one less failed payment attempt
+          failure-cost {
+            fee-base-msat = 2000
+            fee-proportional-millionths = 500
+          }
         }
 
         // Examples of other configs as 0% experiments:
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -245,7 +245,7 @@ eclair {
         test-failure-cost = ${eclair.router.path-finding.default} {
           use-ratios = false
 
-          locked-funds-risk = 1e-15 // msat per msat locked per block. It should be your expected interest rate per block multiplied by the probability that something goes wrong and your funds stay locked.
+          locked-funds-risk = 1e-10 // msat per msat locked per block. It should be your expected interest rate per block multiplied by the probability that something goes wrong and your funds stay locked.
 
           // Virtual fee for failed payments
           // Corresponds to how much you are willing to pay to get one less failed payment attempt
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -245,7 +245,8 @@ eclair {
         test-failure-cost = ${eclair.router.path-finding.default} {
           use-ratios = false
 
-          locked-funds-risk = 1e-10 // msat per msat locked per block. It should be your expected interest rate per block multiplied by the probability that something goes wrong and your funds stay locked.
+          locked-funds-risk = 1e-8 // msat per msat locked per block. It should be your expected interest rate per block multiplied by the probability that something goes wrong and your funds stay locked.
+          // 1e-8 corresponds to an interest rate of ~5% per year (1e-6 per block) and a probability of 1% that the channel will fail and our funds will be locked.
 
           // Virtual fee for failed payments
           // Corresponds to how much you are willing to pay to get one less failed payment attempt
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -17,7 +17,7 @@
 package fr.acinq.eclair.router
 
 import fr.acinq.bitcoin.Crypto.PublicKey
-import fr.acinq.bitcoin.{Btc, MilliBtc, Satoshi, SatoshiLong}
+import fr.acinq.bitcoin.{Btc, ByteVector32, MilliBtc, Satoshi, SatoshiLong}
 import fr.acinq.eclair._
 import fr.acinq.eclair.payment.relay.Relayer.RelayFees
 import fr.acinq.eclair.router.Graph.GraphStructure.{DirectedGraph, GraphEdge}
@@ -30,18 +30,20 @@ import scala.collection.mutable
 
 object Graph {
 
-  // @formatter:off
   /**
    * The cumulative weight of a set of edges (path in the graph).
    *
-   * @param cost   amount to send to the recipient + each edge's fees
-   * @param length number of edges in the path
-   * @param cltv   sum of each edge's cltv
-   * @param weight cost multiplied by a factor based on heuristics (see [[WeightRatios]]).
+   * @param amount             amount to send to the recipient + each edge's fees
+   * @param length             number of edges in the path
+   * @param cltv               sum of each edge's cltv
+   * @param successProbability estimate of the probability that the payment would succeed using this path
+   * @param fees               total fees of the path
+   * @param weight             cost multiplied by a factor based on heuristics (see [[WeightRatios]]).
    */
-  case class RichWeight(cost: MilliSatoshi, length: Int, cltv: CltvExpiryDelta, weight: Double) extends Ordered[RichWeight] {
+  case class RichWeight(amount: MilliSatoshi, length: Int, cltv: CltvExpiryDelta, successProbability: Double, fees: MilliSatoshi, weight: Double) extends Ordered[RichWeight] {
     override def compare(that: RichWeight): Int = this.weight.compareTo(that.weight)
   }
+
   /**
    * We use heuristics to calculate the weight of an edge based on channel age, cltv delta, capacity and a virtual hop cost to keep routes short.
    * We favor older channels, with bigger capacity and small cltv delta.
@@ -53,9 +55,21 @@ object Graph {
     require(ageFactor >= 0.0, "ratio-channel-age must be nonnegative")
     require(capacityFactor >= 0.0, "ratio-channel-capacity must be nonnegative")
   }
+
+  /**
+   * We use heuristics to calculate the weight of an edge.
+   * The fee for a failed attempt and the fee per hop are never actually spent, they are used to incentivize shorter
+   * paths or path with higher success probability.
+   *
+   * @param lockedFundsRisk cost of having funds locked in htlc in msat per msat per block
+   * @param failureCost     fee for a failed attempt
+   * @param hopCost         virtual fee per hop (how much we're willing to pay to make the route one hop shorter)
+   */
+  case class HeuristicsConstants(lockedFundsRisk: Double, failureCost: RelayFees, hopCost: RelayFees)
+
   case class WeightedNode(key: PublicKey, weight: RichWeight)
+
   case class WeightedPath(path: Seq[GraphEdge], weight: RichWeight)
-  // @formatter:on
 
   /**
    * This comparator must be consistent with the "equals" behavior, thus for two weighted nodes with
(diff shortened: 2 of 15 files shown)


t-bast:
When you're ready for feedback, can you update the description of this PR with details about how you plan on estimating success probabilities? I think it's worth reviewing the _concept_ before reviewing _actual code_, to ensure you don't implement something that will be a concept NACK.

thomash-acinq:
This PR is not ready yet because it depends on #1930 which is still being reviewed.

I estimate probabilities in a very simple way based on capacity: I assume that the balance is a random variable uniformly distributed between 0 and capacity and multiply probabilities for all the channels on the path.
But how I compute the probability doesn't really matter, we can change it easily. The goal of this PR is to introduce a new scoring of paths that uses this probability.

t-bast:
Ok thanks, please update the description with that info, it's important to know!

People will be looking at our PRs but they will mostly read the description, not the code, and we don't want them to get the wrong idea on what eclair actually does inside its path-finding.

Because based on the title of the PR alone, you could have started to store historic information about failed payments, and that would have been a concept NACK for me.

eclair-core/src/main/resources/reference.conf
```diff
@@ -212,6 +212,7 @@ eclair {
           channel-age = 0.4       // when computing the weight for a channel, consider its AGE in this proportion
           channel-capacity = 0.55 // when computing the weight for a channel, consider its CAPACITY in this proportion
         }
+        use-ratios = true
```
t-bast, line 215: A boolean is a bit limiting, it only allows choosing between two options.
What about having one sub-section here per cost formula?
And one enum field that selects which "formula" to use?

It would look something like:

```
scoring-heuristic = "ratios" // can be ratios, failure-cost or some-other-way-of-scoring

ratios {
  // contains all parameters for scoring based on ratios
}

failure-cost {
  // contains all parameters for scoring based on wasted capital in case of failure
}

some-other-way-of-scoring {
  // contains parameter for some other way of computing a score
}
```

This feels more future-proof: it makes it easy to introduce widely different scoring formulas/heuristics.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -99,12 +113,12 @@ object Graph {
                         ignoredVertices: Set[PublicKey],
                         extraEdges: Set[GraphEdge],
                         pathsToFind: Int,
-                        wr: WeightRatios,
+                        wr: Either[WeightRatios, HeuristicsConstants],
```
t-bast, line 116: I think we should introduce some kind of `sealed trait ScoringHeuristics` to allow future extensibility (and figure out a way of making `WeightRatios` and `HeuristicsConstants` fit below this new sealed trait).

Ideally in the longer term we'll figure out the right abstract methods this trait should have to avoid switching between different heuristics directly inside the dijkstra logic. Sorry if this is very hand-wavy, it's a bit foggy for now but I think it's a good general direction to go towards.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -53,9 +55,21 @@ object Graph {
...
+  /**
+   * We use heuristics to calculate the weight of an edge.
+   * The fee for a failed attempt and the fee per hop are never actually spent, they are used to incentivize shorter
+   * paths or path with higher success probability.
+   *
+   * @param lockedFundsRisk cost of having funds locked in htlc in msat per msat per block
+   * @param failureCost     fee for a failed attempt
+   * @param hopCost         virtual fee per hop (how much we're willing to pay to make the route one hop shorter)
+   */
+  case class HeuristicsConstants(lockedFundsRisk: Double, failureCost: RelayFees, hopCost: RelayFees)
```
t-bast, line 68: I don't think `HeuristicsConstants` is a good name for this, especially if we introduce a `sealed trait` as described in my other comment. This class should reflect the fact that the heuristic it uses is the cost of failure (`FundsLockedParams` or `FundsLockedHeuristics` maybe?).

t-bast (commented):
First pass on the general design / future extensibility.
Let me know if it's too vague, and we can discuss it to refine the proposal.

eclair-core/src/main/resources/reference.conf
thomash-acinq: In the long term I don't think we should support several scoring formulas. The idea here is to try a different formula, get some numbers and then delete the worst formula.

t-bast: Ok, interesting, in that case we can do something more ad-hoc like you did

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -274,28 +288,45 @@ object Graph {
...
+        val cltvFactor = normalize(edge.update.cltvExpiryDelta.toInt, CLTV_LOW, CLTV_HIGH)
+
+        // NB we're guaranteed to have weightRatios and factors > 0
+        val factor = weightRatios.baseFactor + (cltvFactor * weightRatios.cltvDeltaFactor) + (ageFactor * weightRatios.ageFactor) + (capFactor * weightRatios.capacityFactor)
+        val totalWeight = prev.weight + (fee + hopCost).toLong * factor
+        RichWeight(totalAmount, prev.length + 1, totalCltv, 1.0, totalFees, totalWeight)
+      case Right(heuristicsConstants) =>
+        val hopCost = nodeFee(heuristicsConstants.hopCost.feeBase, heuristicsConstants.hopCost.feeProportionalMillionths, prev.amount)
+        val cltv = if (edge.desc.a == sender && !includeLocalChannelCost) CltvExpiryDelta(0) else edge.update.cltvExpiryDelta
+        val totalCltv = prev.cltv + cltv
```
t-bast, line 320: These are already calculated above, remove?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -274,28 +288,45 @@ object Graph {
...
+        val riskCost = totalAmount.toLong * cltv.toInt * heuristicsConstants.lockedFundsRisk
```
t-bast, line 321: That will very strongly bias towards smaller routes, because for each hop you're re-applying the risk to the total amount.
This makes sense to me (I do think this is how it should be computed) but then the additional `hopCost` mechanism is redundant, isn't it?

We could simplify this heuristic by only applying this funds-locked-risk and nothing else, wouldn't that be better?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -274,28 +288,45 @@ object Graph {
...
+        // If the edge was added by the invoice, it is assumed that it can route the payment.
+        // If we know the balance of the channel, then we will check separately that it can relay the payment.
+        val successProbability = if (edge.update.chainHash == ByteVector32.Zeroes || edge.balance_opt.nonEmpty) 1.0 else 1.0 - totalAmount.toLong.toDouble / (1000 * edge.capacity.toLong.toDouble)
```
t-bast, line 324: You should use `edge.capacity.toMilliSatoshi` instead of multiplying by 1000, it would make it easier to see that we're comparing apples to apples.

t-bast: Side note: it would actually be interesting to do A/B testing on the probability distribution we use for the channel balance in the future, if we think this new heuristic is likely to replace the old one

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -274,28 +288,45 @@ object Graph {
...
+        val totalSuccessProbability = prev.successProbability * successProbability
+        val failureCost = nodeFee(heuristicsConstants.failureCost.feeBase, heuristicsConstants.failureCost.feeProportionalMillionths, totalAmount)
+        val weight = totalFees.toLong + hopCost.toLong + riskCost * totalCltv.toInt * totalAmount.toLong + failureCost.toLong / totalSuccessProbability
```
t-bast, line 327: `riskCost` already contains a `cltv` and the `totalAmount` components, why are you multiplying again?

eclair-core/src/main/resources/reference.conf
```diff
@@ -237,6 +238,20 @@ eclair {
           percentage = 100 // 100% of the traffic use the default configuration
         }
 
+        // alternative routing heuristics (replaces ratios)
+        test-failure-cost = ${eclair.router.path-finding.default} {
+          use-ratios = false
```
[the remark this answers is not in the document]
thomash-acinq: That's how I did it first but I changed it felt cleaner. In particular what bothered me is that we work by overwriting a default config and the "introspect" solution cannot be overwritten.

thomash-acinq (commented):
I've also fixed to the `hopCost` computation. I was only counting the last hop.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: > Actually I don't think it will. It would bias toward shorter routes if all channels had a similar cltv delta but that's not the case, both 40 and 144 are common values. The riskCost will be lower for a route of length 3 with cltv deltas of 40 than a route of size 1 with a cltv delta of 144.

Good point, it's true that we're currently seeing widely different values of cltv-expiry-delta, so this isn't a proxy heuristic for route length. I'm convinced, we should keep `hopCost` here as well!

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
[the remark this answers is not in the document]
t-bast: I agree both are hard to name...let's keep it like this for now then

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
t-bast: True, I don't see any easy elegant solution here, let's go for this simple one.
We can always refine later when we have more hands-on experience with configuring various experiments.

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -1762,18 +1762,69 @@ class RouteCalculationSpec extends AnyFunSuite with ParallelTestExecution {
...
+      makeEdge(2L, a, c, 400 msat, 500, capacity = (DEFAULT_AMOUNT_MSAT * 3).truncateToSatoshi),
+      makeEdge(3L, c, d, 400 msat, 500, capacity = (DEFAULT_AMOUNT_MSAT * 3).truncateToSatoshi),
+      makeEdge(4L, d, b, 400 msat, 500, capacity = (DEFAULT_AMOUNT_MSAT * 3).truncateToSatoshi),
+    ))
+
+    val hc = HeuristicsConstants(
+      lockedFundsRisk = 0.0,
+      failureCost = RelayFees(1000 msat, 500),
+      hopCost = RelayFees(0 msat, 0),
+    )
```
t-bast, line 1796: It feels to me that this configuration is essentially the same as the normal weight ratios with a big factor for the channel capacity component, isn't it? But maybe it's mostly because for now we use a uniform distribution for the likely balance of the channel...

eclair-core/src/main/resources/reference.conf
```diff
@@ -240,6 +241,20 @@ eclair {
...
+
+          locked-funds-risk = 1e-10 // msat per msat locked per block. It should be your expected interest rate per block multiplied by the probability that something goes wrong and your funds stay locked.
```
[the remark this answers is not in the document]
thomash-acinq: I've changed it to 1e-8 with a detailed explanation of how I came up with this number.

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/PathFinding.scala
```diff
@@ -33,26 +33,28 @@ trait PathFinding {
   private implicit def ec: ExecutionContext = actorSystem.dispatcher
 
   val findRoute: Route = postRequest("findroute") { implicit t =>
-    formFields(invoiceFormParam, amountMsatFormParam.?, "pathFindingExperimentName".?, routeFormat.?) {
-      case (invoice@PaymentRequest(_, Some(amount), _, nodeId, _, _), None, pathFindingExperimentName_opt, routeFormat) =>
-        complete(eclairApi.findRoute(nodeId, amount, pathFindingExperimentName_opt, invoice.routingInfo).map(r => RouteFormat.format(r, routeFormat)))
-      case (invoice, Some(overrideAmount), pathFindingExperimentName_opt, routeFormat) =>
-        complete(eclairApi.findRoute(invoice.nodeId, overrideAmount, pathFindingExperimentName_opt, invoice.routingInfo).map(r => RouteFormat.format(r, routeFormat)))
+    formFields(invoiceFormParam, amountMsatFormParam.?, "pathFindingExperimentName".?, routeFormat.?, "includeLocalChannelCost".as[Boolean].?) {
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
thomash-acinq: The goal is the same indeed but the result should be very different. With the current capacity ratio, a channel of 1BTC gets a score five times bigger than a channel of 0.2BTC, but when you want to pay 10k sats, both capacities are vastly larger than the payment amount so they should both get the maximum score (for the capacity part).

t-bast (approved):

PR #1943, opened 2021-09-08
Add 'shortChannelId' output format for findroute* API calls

rorp:
By default, `findroute*` API calls return routes as lists of node IDs. This PR adds an ability to request routes as lists short channel IDs with `--format` form parameter. That allows the user to find routes compatible (both `nodeId` and `shortChannelId` hops) with `sendtoroute` API call.


eclair-node/src/main/scala/fr/acinq/eclair/api/directives/RouteFormat.scala
```diff
@@ -0,0 +1,53 @@
...
+
+package fr.acinq.eclair.api.directives
+
+import fr.acinq.eclair.router.Router.RouteResponse
+
+sealed trait RouteFormat
+
+case object NodeIdRouteFormat extends RouteFormat
+
+case object ShortChannelIdRouteFormat extends RouteFormat
```
pm47, line 25: Nit:

```suggestion
// @formatter:off
sealed trait RouteFormat
case object NodeIdRouteFormat extends RouteFormat
case object ShortChannelIdRouteFormat extends RouteFormat
// @formatter:on
```

pm47 (commented):
LGTM, just needs a rebase.

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/RouteFormat.scala
```diff
@@ -0,0 +1,53 @@
...
+
+object RouteFormat {
+
+  val NODE_ID = "nodeId"
+  val SHORT_CHANNEL_ID = "shortChannelId"
+
+  def fromString(s: String): RouteFormat = s match {
+    case NODE_ID => NodeIdRouteFormat
+    case SHORT_CHANNEL_ID => ShortChannelIdRouteFormat
+    case _ => throw new IllegalArgumentException(s"invalid route format, possible values are ($NODE_ID, $SHORT_CHANNEL_ID)")
```
t-bast, line 35: Can we avoid throwing here and return an `Option` or an `Either` instead?

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/RouteFormat.scala
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  def format(route: RouteResponse, format_opt: Option[RouteFormat]): Seq[String] = format(route, format_opt.getOrElse(NodeIdRouteFormat))
```
t-bast, line 38: nit: linked to the previous comment, I think the format shouldn't be an option here, the caller should ensure it always provides a format

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/PathFinding.scala
```diff
@@ -19,36 +19,47 @@ package fr.acinq.eclair.api.handlers
...
   import fr.acinq.eclair.api.serde.JsonSupport.{formats, marshaller, serialization}
 
   val findRoute: Route = postRequest("findroute") { implicit t =>
-    formFields(invoiceFormParam, amountMsatFormParam.?, "pathFindingExperimentName".?) {
-      case (invoice@PaymentRequest(_, Some(amount), _, nodeId, _, _), None, pathFindingExperimentName_opt) =>
-        complete(eclairApi.findRoute(nodeId, amount, pathFindingExperimentName_opt, invoice.routingInfo))
-      case (invoice, Some(overrideAmount), pathFindingExperimentName_opt) =>
-        complete(eclairApi.findRoute(invoice.nodeId, overrideAmount, pathFindingExperimentName_opt, invoice.routingInfo))
+
+    implicit val ec: ExecutionContext = actorSystem.dispatcher
```
t-bast, line 35: Why is that needed? Is it because of the `map` call?
Can we lift this outside of each `val` directly in the `trait` or is it impossible?

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -940,6 +941,85 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
...
+        assert(entityAs[JArray](Json4sSupport.unmarshaller, ClassTag(classOf[ErrorResponse])) == JArray(List(
+          JString(mockHop1.nodeId.toString()),
+          JString(mockHop2.nodeId.toString()),
+          JString(mockHop3.nodeId.toString()),
+          JString(mockHop3.nextNodeId.toString())
+        )))
+        eclair.findRoute(PublicKey.fromBin(ByteVector.fromValidHex("036ded9bb8175d0c9fd3fad145965cf5005ec599570f35c682e710dc6001ff605e")), 456.msat, any, any)(any[Timeout]).wasCalled(twice)
+      }
+
+    Post("/findroute", FormData("invoice" -> invoice, "amountMsat" -> "456", "format" -> "shortChannelId")) ~>
```
t-bast, line 1007: You should also test the case where an invalid format is provided.

t-bast (commented):
Concept ACK, good idea!

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/RouteFormat.scala
rorp: Bit it throws here, for example: https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala#L44

There will to much boilerplate to handle `Option` or `Either` here.

t-bast: Ok fine, let's skip it for now, we'll refactor the whole API to avoid throwing exceptions in a separate PR.

rorp: This is how it looks like in the CLI:
```bash
$ ./eclair-cli findroutetonode --format= --nodeId=0321fb8a1edde1e0d1721e726989b5c28264d56796ba968a4529c19f05cbf63746 --amountMsat=1000 
The form field 'format' was malformed:
invalid route format, possible values are (nodeId, shortChannelId)
```

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/RouteFormat.scala
rorp: But non-optional format field will make the API backwards incompatible.

t-bast (approved):

PR #1945, opened 2021-09-09
Fix the build

pm47:
#1912 has caused a regression due to the use of akka's coordinated shutdown it caused the test jvm to halt and tests weren't passing. Hence the very low coverage since a week ago.


t-bast (approved):
The test run looks better, all tests seem to have run (with one failing though).

PR #1949, opened 2021-09-13
Replace bitcoin-lib with bitcoin-kmp

sstone:
Replacing bitcoin-lib with bitcoin-kmp on eclair means that we don't need to maintain 2 very similar code bases anymore. bitcoin-kmp is already used on iOS, soon on Android, and is much easier to use from Java/Kotlin/Clojure... than bitcoin-lib. Besides, a simple bitcoin library does not benefit as much from Scala's powerful type system as eclair does.

The goal is of this PR is

- to make sure that bitcoin-kmp  is usable from Scala
- to increase bitcoin-kmp's test coverage (it already passes all core reference tests, as well as all bitcoin-lib tests which we ported to kotlin)

Since bitcoin-kmp does not use any Scala types (including collection) its use is sometimes a bit less "fluent" (use of `new`, conversion to/from java collections and limited pattern matching, use of `plus`, `minus` on `Satoshi` instances instead of `+`, `-` ...)


Mikegarylee01:
> Replacing bitcoin-lib with bitcoin-kmp on eclair means that we don't need to maintain 2 very similar code bases anymore. bitcoin-kmp is already used on iOS, soon on Android, and is much easier to use from Java/Kotlin/Clojure... than bitcoin-lib. Besides, a simple bitcoin library does not benefit as much from Scala's powerful type system as eclair does.
> 
> The goal is of this PR is
> 
> - to make sure that bitcoin-kmp  is usable from Scala
> - to increase bitcoin-kmp's test coverage (it already passes all core reference tests, as well as all bitcoin-lib tests which we ported to kotlin)
> 
> Since bitcoin-kmp does not use any Scala types (including collection) its use is sometimes a bit less "fluent" (use of `new`, conversion to/from java collections and limited pattern matching, use of `plus`, `minus` on `Satoshi` instances instead of `+`, `-` ...)

:ɪ •ᴡɪʟʟ •ɪɴᴛʀᴏᴅᴜᴄᴇ •ʏᴏᴜ •on• how •to •make• higher  profit•through•ɪɴᴠᴇꜱᴛᴍᴇɴᴛ No lose Text' me' on' •ᴡʜᴀᴛꜱᴀᴘᴘ••
+13234525476.

PR #1957, opened 2021-09-17
Relay onion messages

thomash-acinq:
Basic support for relaying onion messages
https://github.com/lightningnetwork/lightning-rfc/pull/759


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -430,6 +464,8 @@ object Peer {
     def apply(uri: NodeURI): Connect = new Connect(uri.nodeId, Some(uri.address))
   }
 
+  case class SendMessage(nodeId: PublicKey, message: LightningMessage)
```
t-bast, line 467: I'm not sure that this belongs to the `Peer`, this is more of a `Switchboard` message (since you want to potentially trigger opening a new connection if we don't have one with the target).

I'm also not sure it should be that generic (containing any `LightningMessage`).
Let's start with just onion messages (as it's the only case for which we're using it) and later we can make it more generic if needed.

t-bast (commented):
A couple high-level comments, overall it's going in the right direction.

I'll be working on route blinding this week, this will probably overlap with some of the changes you made, once it's ready you should be able to rebase on top of my branch. I'm not sure yet where the TLVs and codecs should live, I need to dive back into the spec proposal to figure this out.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/io/RateLimiter.scala
```diff
@@ -0,0 +1,22 @@
+package fr.acinq.eclair.io
+
+import scala.concurrent.duration.FiniteDuration
+
+class RateLimiter(maxCount: Int, timeWindow: FiniteDuration) {
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -245,6 +250,16 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
         d.channels.values.toSet[ActorRef].foreach(_ ! INPUT_DISCONNECTED) // we deduplicate with toSet because there might be two entries per channel (tmp id and final id)
         gotoConnected(connectionReady, d.channels)
 
+      case Event(msg: OnionMessage, d: ConnectedData) =>
+        if (nodeParams.features.hasFeature(Features.OnionMessages) && messageRelayRateLimiter.ok()) {
+          relayOnionMessage(msg)
+        }
```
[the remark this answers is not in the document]
thomash-acinq: I don't see any reason to return anything, the sending node isn't waiting for a reply from us anyway. It's an odd message so it should just be ignored.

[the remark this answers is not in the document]
thomash-acinq: From the spec:

```
A receiving node:
  - upon receiving a message of _odd_, unknown type:
    - MUST ignore the received message.
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/Onion.scala
```diff
@@ -123,63 +124,76 @@ Notes:
 case class OnionRoutingPacket(version: Int, publicKey: ByteVector, payload: ByteVector, hmac: ByteVector32)
 
 /** Tlv types used inside onion messages. */
-sealed trait OnionTlv extends Tlv
+sealed trait OnionPaymentPayloadTlv extends Tlv
 
-object OnionTlv {
+sealed trait OnionMessagePayloadTlv extends Tlv
+
+sealed trait BlindedTlv extends Tlv
```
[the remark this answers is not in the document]
t-bast: We should really stay very close to spec names in our tlv traits, otherwise it's really hard to verify we correctly follow the spec. I need to spend time on the spec PR to see if that name needs to be clarified at the spec level or not.

eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala
```diff
@@ -71,6 +71,11 @@ class Switchboard(nodeParams: NodeParams, peerFactory: Switchboard.PeerFactory)
         case None => sender() ! Status.Failure(new RuntimeException("peer not found"))
       }
 
+    case s@Peer.SendOnionMessage(nodeId, _) =>
+      val peer = createOrGetPeer(nodeId, offlineChannels = Set.empty)
+      peer ! Peer.Connect(nodeId, None)
+      peer forward s
```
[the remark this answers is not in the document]
rustyrussell: In c-lightning we don't opportunistically connect like this *for forwards* (we do if we're the originator).  Long term I think that's the right thing, but it might be nice in the short term if nodes were try autoconnect like this even for forwarding (with some limit!).

It's overkill for the first implementation though, IMHO.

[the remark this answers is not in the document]
t-bast: If it's not too hard we'd like to have it sooner rather than later, because we'd like to ensure that Phoenix users don't have to dox their IP address when they use offers (if they aren't using Tor)...we'd rather have them send onion messages through our node and have our node connect on-the-fly. Ideally if Phoenix could make our node connect to the next-to-last node in the onion route (instead of the final recipient) it would be even better for privacy!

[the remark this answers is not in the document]
t-bast: But you're right that we can split some of the connection handling to a separate PR to avoid delaying and get a first version of onion messages out! I'll be reviewing it this week.

thomash-acinq:
@t-bast I was able to rebase it on top of your refactoring

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -0,0 +1,108 @@
...
+        val tlvs = padding.map(EncryptedRecipientDataTlv.Padding(_) :: Nil).getOrElse(Nil) ++ recipientSecret.map(EncryptedRecipientDataTlv.RecipientSecret(_) :: Nil).getOrElse(Nil)
+        val lastPayload = finalBlindedTlvCodec.encode(FinalBlindedTlv(TlvStream(tlvs))).require.bytes
+        Sphinx.RouteBlinding.create(blindingSecret, intermediateNodes.map(_.nodeId) :+ nodeId, intermediatePayloads :+ lastPayload)
+      case Right(route) =>
+        val Sphinx.RouteBlinding.BlindedRoute(introductionNodeId, blindingKey, blindedNodes) = Sphinx.RouteBlinding.create(blindingSecret, intermediateNodes.map(_.nodeId), intermediatePayloads)
+        Sphinx.RouteBlinding.BlindedRoute(introductionNodeId, blindingKey, blindedNodes ++ route.blindedNodes)
+    }
+  }
+
+  def buildMessage(sessionKey: PrivateKey,
```
t-bast, line 46: This function needs scaladoc comments to explain what exactly should be inside `intermediateNodes` and `content`

eclair-core/src/test/scala/fr/acinq/eclair/message/OnionMessagesSpec.scala
```diff
@@ -0,0 +1,218 @@
...
+      case OnionMessages.RelayMessage(nextNodeId, onionForBob) =>
+        assert(nextNodeId == bob.publicKey)
+        OnionMessages.process(bob, onionForBob) match {
+          case OnionMessages.RelayMessage(nextNodeId, onionForCarol) =>
+            assert(nextNodeId == carol.publicKey)
+            OnionMessages.process(carol, onionForCarol) match {
+              case OnionMessages.RelayMessage(nextNodeId, onionForDave) =>
+                assert(nextNodeId == dave.publicKey)
+                OnionMessages.process(dave, onionForDave) match {
+                  case OnionMessages.ReceiveMessage(_, _) => ()
```
t-bast, line 127: Isn't there a `pathId` we can verify here?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -243,6 +258,20 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
         d.channels.values.toSet[ActorRef].foreach(_ ! INPUT_DISCONNECTED) // we deduplicate with toSet because there might be two entries per channel (tmp id and final id)
         gotoConnected(connectionReady, d.channels)
 
+      case Event(msg: OnionMessage, _: ConnectedData) =>
+        if (nodeParams.features.hasFeature(Features.OnionMessages)) {
+          OnionMessages.process(nodeParams.privateKey, msg) match {
+            case OnionMessages.DropMessage(_) => () // We ignore bad messages
```
t-bast, line 264: We should log the error.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -243,6 +258,20 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
...
+            case OnionMessages.RelayMessage(nextNodeId, dataToRelay) => context.parent ! Peer.SendOnionMessage(nextNodeId, dataToRelay)
+            case OnionMessages.ReceiveMessage(_, _) => () // We only relay messages
```
t-bast, line 266: Can you add an `info` log here for now with the content of the message?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -94,6 +96,19 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
       stay() using d.copy(channels = d.channels + (FinalChannelId(channelId) -> channel))
 
     case Event(_: LightningMessage, _) => stay() // we probably just got disconnected and that's the last messages we received
+
+    case Event(Peer.SendOnionMessage(toNodeId, msg), d: DisconnectedData) if toNodeId == remoteNodeId =>
+      // We may drop a previous messageToRelay but that's fine. If we receive several messages to relay while trying to connect, we're probably getting spammed.
```
t-bast, line 101: Smart, I like this simple mechanism.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -94,6 +96,19 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
...
+      stay() using d.copy(messageToRelay = Some(msg))
+
+    case Event(_ : PeerConnection.ConnectionResult.Failure, d: DisconnectedData) =>
+      if (d.channels.isEmpty) {
+        // we don't have any channel with this peer and we can't connect to it so we just drop it
+        stopPeer()
+      } else {
+        stay() using d.copy(messageToRelay = None)
+      }
```
t-bast, line 110: This addition looks unrelated to onion messages, isn't it?
Why weren't we handling that event here before?
To be honest the `Peer` is subtle and I'm not sure I'm able to properly review it, it would be nice if @pm47 could review the changes to this file (no need to review the rest of the PR, it's been covered).

t-bast (commented):
Good stuff, the logic looks good to me :+1: 

We are missing the following tests though:

- We should create a `MessageOnionCodecsSpec.scala` and test serialization of each individual tlv, I've always found these extremely useful to quickly find codec bugs. In particular, it should contain tests for the fact that the onion message codec allows onions of arbitrary length.
- We should extend `PeerSpec` to test the newly added logic of relaying onion messages once connected

I've made annoying comments about multiple fields to rename / file organization to match `PaymentOnion.scala`, but since these are annoying I've also prepared a commit that applies them so you don't have to go through the pain of this refactoring in https://github.com/ACINQ/eclair/pull/2057

Apart from that, LGTM we should be able to merge that soon.
I'll discuss what final values we should set for each tlv field with rusty today and we can amend the PR with the latest values then.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OnionRouting.scala
```diff
@@ -40,10 +40,20 @@ object OnionRoutingCodecs {
...
+  case class ForbiddenTlv(tag: UInt64) extends Err {
+    // @formatter:off
+    val failureMessage: FailureMessage = InvalidOnionPayload(tag, 0)
+    override def message = failureMessage.message
+    override def context: List[String] = Nil
+    override def pushContext(ctx: String): Err = this
+    // @formatter:on
+  }
+
+  def onionRoutingPacketCodec(payloadLength: Codec[Int]): Codec[OnionRoutingPacket] = (
```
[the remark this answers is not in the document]
t-bast: Updated in https://github.com/ACINQ/eclair/pull/2057

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/MessageOnion.scala
```diff
@@ -0,0 +1,91 @@
...
+import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.eclair.UInt64
+import fr.acinq.eclair.crypto.Sphinx.RouteBlinding.{BlindedNode, BlindedRoute}
+import fr.acinq.eclair.wire.protocol.EncryptedRecipientDataCodecs.encryptedRecipientDataCodec
+import fr.acinq.eclair.wire.protocol.OnionRoutingCodecs.{ForbiddenTlv, MissingRequiredTlv, onionRoutingPacketCodec}
+import scodec.bits.ByteVector
+
+sealed trait OnionMessagePayloadTlv extends Tlv
+
+object MessageTlv {
```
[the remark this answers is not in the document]
t-bast: Updated in https://github.com/ACINQ/eclair/pull/2057

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -0,0 +1,108 @@
...
+    } else {
+      payloadSize.toInt
+    }
+    val Sphinx.PacketAndSecrets(packet, _) = Sphinx.create(sessionKey, packetSize, route.blindedNodes.map(_.blindedPublicKey), payloads, None)
+    (route.introductionNodeId, OnionMessage(blindingSecret.publicKey, packet))
+  }
+
+  sealed trait Action
+
+  case class DropMessage(reason: String) extends Action
```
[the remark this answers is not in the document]
t-bast: Updated in https://github.com/ACINQ/eclair/pull/2057

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OnionRouting.scala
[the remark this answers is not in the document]
thomash-acinq: I don't think that's possible with scodec. How would you do it?

thomash-acinq:
Thank you for the improvements in #2057, I've merged them.
I've duplicated the onion packet codec so that it doesn't affect the routing part.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
thomash-acinq: We didn't handle this before because we would never initiate a connection before. Connections were initiated manually with the API or by other peers connecting to us.
Now if we want to relay to a new peer, we first create the `Peer` object and then try to establish a connection. If the connection fails for a peer that we've just created that way, we should get rid of it.

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -0,0 +1,108 @@
...
+
+object OnionMessages {
+
+  case class IntermediateNode(nodeId: PublicKey, padding: Option[ByteVector] = None)
+
+  case class Recipient(nodeId: PublicKey, secret: Option[ByteVector], padding: Option[ByteVector] = None)
+
+  def buildRoute(blindingSecret: PrivateKey,
+                 intermediateNodes: Seq[IntermediateNode],
+                 destination: Either[Recipient, Sphinx.RouteBlinding.BlindedRoute]): Sphinx.RouteBlinding.BlindedRoute = {
```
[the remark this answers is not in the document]
thomash-acinq: I think it makes sense to share the code, there is a lot of common code and also these two functions would be used in the same place.

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -0,0 +1,108 @@
...
+    val last = destination match {
+      case Left(Recipient(nodeId, _, _)) => EncryptedRecipientDataTlv.OutgoingNodeId(nodeId) :: Nil
```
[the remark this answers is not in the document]
thomash-acinq: This tlv is not sent to the recipient but to the node just before the recipient tell it who is the next node. The padding for the recipient is added later in the tlv stream that will be sent to the recipient.

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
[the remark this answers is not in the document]
thomash-acinq: Padding adds some complexity, but if we don't have it we can't check the test vectors if they test padding.

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -0,0 +1,108 @@
...
+      case Right(Sphinx.RouteBlinding.BlindedRoute(nodeId, blindingKey, _)) => EncryptedRecipientDataTlv.OutgoingNodeId(nodeId) :: EncryptedRecipientDataTlv.NextBlinding(blindingKey) :: Nil
+    }
+    val intermediatePayloads =
+      if (intermediateNodes.isEmpty) {
+        Nil
+      } else {
+        (intermediateNodes.tail.map(node => EncryptedRecipientDataTlv.OutgoingNodeId(node.nodeId) :: Nil) :+ last)
```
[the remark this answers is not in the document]
thomash-acinq: I use `tail` because I want the next node ids. I'm zipping `intermediateNodes.tail` and `intermediateNodes` to get a tuple (next node, current node).
`intermediateNodes` contains all the nodes that are between us and the destination. It can be empty if we contact the destination directly.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala

eclair-core/src/test/scala/fr/acinq/eclair/message/OnionMessagesSpec.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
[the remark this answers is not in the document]
t-bast: Right, good point, then let's keep it like that.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: Good point! It looks good to me.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/EncryptedRecipientDataTlv.scala
```diff
@@ -49,18 +54,21 @@ object EncryptedRecipientDataCodecs {
...
+  private val pathId: Codec[PathId] = variableSizeBytesLong(varintoverflow, "path_id" | bytes).as[PathId]
+  private val nextBlinding: Codec[NextBlinding] = (("length" | constant(hex"21")) :: ("blinding" | publicKey)).as[NextBlinding]
 
   private val encryptedRecipientDataTlvCodec = discriminated[EncryptedRecipientDataTlv].by(varint)
     .typecase(UInt64(1), padding)
     .typecase(UInt64(2), outgoingChannelId)
     .typecase(UInt64(4), outgoingNodeId)
-    .typecase(UInt64(6), recipientSecret)
+    .typecase(UInt64(6), pathId)
+    .typecase(UInt64(12), nextBlinding)
```
t-bast, line 71: I discussed it with rusty earlier today, and we should use 8 for this tlv, no reason to leave a gap (I'll update the route blinding spec PR):
```suggestion
    .typecase(UInt64(8), nextBlinding)
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/MessageOnion.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  private val replyHopCodec: Codec[BlindedNode] = (("nodeId" | publicKey) :: ("encryptedData" | variableSizeBytes(uint16, bytes))).as[BlindedNode]
+
+  private val replyPathCodec: Codec[ReplyPath] = variableSizeBytesLong(varintoverflow, ("firstNodeId" | publicKey) :: ("blinding" | publicKey) :: ("path" | list(replyHopCodec).xmap[Seq[BlindedNode]](_.toSeq, _.toList))).as[BlindedRoute].as[ReplyPath]
+
+  private val encryptedDataCodec: Codec[EncryptedData] = variableSizeBytesLong(varintoverflow, bytes).as[EncryptedData]
+
+  private val onionTlvCodec = discriminated[OnionMessagePayloadTlv].by(varint)
+    .typecase(UInt64(2), replyPathCodec)
+    .typecase(UInt64(10), encryptedDataCodec)
```
t-bast, line 90: As seen with Rusty earlier today, we should use 4 here, no need for a gap:
```suggestion
    .typecase(UInt64(4), encryptedDataCodec)
```

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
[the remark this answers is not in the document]
t-bast: Right, that makes sense.

t-bast (dismissed):
LGTM, let's get this first version on `master`, well done!

t-bast (approved):

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: We just discussed this offline, I would have created a special-purpose actor for the task of connecting to the peer and sending messages. This would be consistent with what we do for relaying payments (and also waking up peers in feature branches) and would reduce added complexity to the `Peer`.

t-bast: Can we update that in a separate PR where we rework the peer aspect more in-depth?

pm47: Sure, it's not a blocker.

t-bast (approved):

PR #1958, opened 2021-09-17
Make Tor optional for blockchain watchdogs

rorp:
Fixes https://github.com/ACINQ/eclair/discussions/1947

### DIFF
--- a/docs/Tor.md
+++ b/docs/Tor.md
@@ -175,3 +175,9 @@ eclair.blockchain-watchdog.sources = [
   "mempool.space"
 ]
 ```
+
+Also, you can disable Tor for all watchdog sources altogether using:
+
+```s
+eclair.socks5.use-for-watchdogs = false
+```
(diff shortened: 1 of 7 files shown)


t-bast (commented):
@fiatjaf could you have a look at this?
I'm actually not completely sure how you use the socks5 configuration.
Do you want the watchdog calls to go through your socks5 proxy (which IIUC is not using Tor)?

It feels to me that @rorp may have taken a shortcut in thinking that the socks5 proxy would only be used with Tor. But you may be using a socks5 proxy with something else than Tor, so you would want the watchdog calls to go through that proxy, but without using their onion address (since the proxy isn't Tor).

But I may be completely over-thinking this, let me know which it is.

fiatjaf:
I only want to use socks5 proxy for Tor.

If I understood this well I think it is perfect. If one is using a socks5 proxy with something else than Tor they would be able to do it and even route watchdog calls through it, so it's all good.

Now I just must remember setting `socks5.use-for-watchdogs` to `false` here.

t-bast:
Cool, thanks for the clarification.

t-bast (approved):

PR #1960, opened 2021-09-20
Forward incompatibility with Bitcoin Core

dscotese:
Installation instructions now make it clear that future versions of Bitcoin Core are not supported in the current version of Eclair.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ You will find detailed guides and frequently asked questions there.
 
 ### Prerequisite: Bitcoin Core
 
-:warning: Eclair requires Bitcoin Core 0.20.1 or 0.21.1. If you are upgrading an existing wallet, you may need to create a new address and send all your funds to that address.
+:warning: Eclair requires Bitcoin Core 0.20.1 or 0.21.1. **Do not update to a newer version of Bitcoin Core software** until you have upgraded Eclair to a verstion that supports it.  If you are upgrading an existing wallet, you may need to create a new address and send all your funds to that address.
 
 Eclair needs a _synchronized_, _segwit-ready_, **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node.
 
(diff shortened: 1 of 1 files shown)


t-bast:
Why would that be needed?
The sentence already clearly says what versions of Bitcoin Core are supported, so if you're running with a different version it should be obvious that this isn't officially tested by us, isn't it?

dscotese:
The sentence is useful for a great many users who expect required software to be backward compatible enough to support other software that depends on it.  Bitcoin Core is a requirement for many Lightning Nodes.  There are features and functions of Bitcoin Core that are reasonably expected not to change.  Most changes in required software are introduced using deprecation (of APIs, semantics, etc.) to give the dependent software time to adjust.

I completely understand the position that forward compatibility is impossible.  My impression of the Bitcoin Core dev team is that they could mitigate this to some extent, but I don't pay enough attention to them to have an idea about how well they do that.  Limiting Eclair's use of the changing parts of Bitcoin Core's features and functions would also help, but then make Eclair less efficient and/or less capable.

You asked why the change is needed, but I think this is the wrong question.  The right question is: How does this help a user of Eclair?  When we limit our documentation to what's needed, we necessarily limit the audience for whom the documentation is useful to some subset of its potential userbase.  In any case, if the goal of Eclair is to provide only that documentation needed by whatever audience the devs believe will use Eclair, then leaving out my addition makes sense.  In my view, despite my general ability to understand more than average, this approach to documentation makes Eclair less useful overall.  So, not I don't think it's needed, but I believe it would be useful.

t-bast:
I thought that the wording was clear enough: there are only two versions of bitcoin core we recommend, they are explicitly listed, use other versions at your own risk. But since you're opening this PR, it means it wasn't clear enough. But I'm not a fan of your proposed change, if the previous wording wasn't enough we should instead stress that only these listed versions are recommended (instead of focusing on future versions). I would put something like:

```
:warning: Eclair requires Bitcoin Core 0.20.1 or 0.21.1 (other versions of Bitcoin Core are *not* actively tested - use at your own risk). If you are upgrading an existing wallet, you may need to create a new address and send all your funds to that address.
```

dscotese:
I like your warning better than my instruction, so I updated the file.  Thanks!

t-bast (approved):

PR #1962, opened 2021-09-21
Add low-level route blinding features

t-bast:
Implement the low-level cryptographic utilities necessary for https://github.com/lightningnetwork/lightning-rfc/pull/765
This PR doesn't do any integration with existing onion payloads yet, nor lets us use route blinding for receiving payments.


eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -372,5 +372,81 @@ object Sphinx extends Logging {
...
+     * Compute the blinded private key that must be used to decrypt an incoming blinded onion.
+     *
+     * @param privateKey           this node's private key.
+     * @param blindingEphemeralKey unblinding ephemeral key.
+     * @return this node's blinded private key.
+     */
+    def derivePrivateKey(privateKey: PrivateKey, blindingEphemeralKey: PublicKey): PrivateKey = {
+      val sharedSecret = computeSharedSecret(blindingEphemeralKey, privateKey)
+      privateKey.multiply(PrivateKey(generateKey("blinded_node_id", sharedSecret)))
+    }
```
thomash-acinq, line 431: How should this be used? I expected it to be used in the next function(`decryptPayload`) but it's not.

t-bast: See https://github.com/lightningnetwork/lightning-rfc/blob/route-blinding/proposals/route-blinding.md#sending-to-a-blinded-route

When a node is part of a blinded route, the onion it will receive isn't encrypted to its `node_id` so the node cannot decrypt the onion. Instead the onion has been encrypted using the `blinded_node_id`, so the node must be able to derive the corresponding private key *before* decrypting the onion. This function does exactly that.

t-bast: I'll add a test case that describes the complete onion encryption / decryption steps (for the HTLC case), I think it will make it clearer (and it's useful to have).

t-bast: Done in the latest force-push.
I also updated the spec PR, I think you should be able to build upon this for onion messages.

thomash-acinq (dismissed):

PR #1963, opened 2021-09-21
Count local fees in path finding metrics

thomash-acinq:
The fee that's recorded in the path-finding metrics should count the local channel. Without it we record failing payments with a fee budget larger than the fee recorded for the successful payment to the same node.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
@@ -249,7 +249,17 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, 
     if (cfg.recordPathFindingMetrics) {
       val fees = event match {
         case Left(_) => request.routeParams.getMaxFee(cfg.recipientAmount)
-        case Right(paymentSent) => paymentSent.feesPaid
+        case Right(paymentSent) =>
+          val localFees = cfg.upstream match {
+            case _: Upstream.Local => 0.msat // no local fees when we are the origin of the payment
+            case _: Upstream.Trampoline =>
+              // in case of a relayed payment, we need to take into account the fee of the first channels
+              paymentSent.parts.collect {
+                // NB: the route attribute will always be defined here
+                case p@PartialPayment(_, _, _, _, Some(route), _) => route.head.fee(p.amountWithFees)
+              }.sum
+          }
+          paymentSent.feesPaid + localFees
       }
       context.system.eventStream.publish(PathFindingExperimentMetrics(cfg.recipientAmount, fees, status, duration, now, isMultiPart = true, request.routeParams.experimentName, cfg.recipientNodeId))
     }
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -249,7 +249,7 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
     if (cfg.recordPathFindingMetrics) {
       val fees = event match {
         case Left(_) => request.routeParams.getMaxFee(cfg.recipientAmount)
-        case Right(paymentSent) => paymentSent.feesPaid
+        case Right(paymentSent) => paymentSent.feesPaid + paymentSent.localFees
```
pm47, line 252: This is incorrect because we want to record the local fees if, and only if, this is a trampoline relayed payment. The info is available in the the `upstream`, which can be either `Local` or `Trampoline`.

```suggestion
        case Right(paymentSent) => 
          val localFees = cfg.upstream match {
            case _: Upstream.Local => 0.msat // no local fees when we are the origin of the payment
            case _: Upstream.Trampoline =>
              // in case of a relayed payment, we need to take into account the fee of the first channels
              paymentSent.parts.collect {
                // NB: the route attribute will always be defined here
                case p@PartialPayment(_, _, _, _, Some(route), _) => route.head.fee(p.amountWithFees)
              }.sum
          }
          paymentSent.feesPaid + localFees
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala
```diff
@@ -305,7 +305,7 @@ class PaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig, router: A
     val duration = now - start
     if (cfg.recordPathFindingMetrics) {
       val fees = result match {
-        case Right(paymentSent: PaymentSent) => paymentSent.feesPaid
+        case Right(paymentSent: PaymentSent) => paymentSent.feesPaid + paymentSent.localFees
         case Left(_) => request match {
           case s: SendPaymentToNode => s.routeParams.getMaxFee(cfg.recipientAmount)
           case _: SendPaymentToRoute => 0 msat
```
pm47, line 311: What is the rationale for setting `0 msat` here?

thomash-acinq: It's actually not possible to reach this case (because for `SendPaymentToRoute`, `recordPathFindingMetrics` is set to false) but the compiler doesn't know it.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -71,6 +72,7 @@ object PaymentSent {
   case class PartialPayment(id: UUID, amount: MilliSatoshi, feesPaid: MilliSatoshi, toChannelId: ByteVector32, route: Option[Seq[Hop]], timestamp: Long = System.currentTimeMillis) {
     require(route.isEmpty || route.get.nonEmpty, "route must be None or contain at least one hop")
     val amountWithFees: MilliSatoshi = amount + feesPaid
+    def localFees: MilliSatoshi = route.map(_.head.fee(amountWithFees)).getOrElse(MilliSatoshi(0))
```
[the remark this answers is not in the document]
thomash-acinq: Done what was suggested below.

PR #1965, opened 2021-09-23
Minor changes

pm47:
A few minor changes and refactoring. Best reviewed commit-by-commit.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -21,6 +21,7 @@ import akka.actor.typed.scaladsl.AskPattern.Askable
 import akka.actor.typed.scaladsl.adapter.ClassicSchedulerOps
 import akka.pattern._
 import akka.util.Timeout
+import com.softwaremill.quicklens.ModifyPimp
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{ByteVector32, ByteVector64, Crypto, Satoshi}
 import fr.acinq.eclair.TimestampQueryFilters._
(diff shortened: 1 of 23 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -155,18 +154,10 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
...
-                val featuresSupported = proposedChannelType.features.forall(f => d.localFeatures.hasFeature(f))
-                if (featuresSupported) {
-                  Right(proposedChannelType)
-                } else {
-                  Left(InvalidChannelType(msg.temporaryChannelId, defaultChannelType, proposedChannelType))
-                }
+            val defaultChannelType = ChannelTypes.defaultFromFeatures(d.localFeatures, d.remoteFeatures)
+            val chosenChannelType: Either[InvalidChannelType, SupportedChannelType] = ChannelTypes.negotiateChannelType(d.localFeatures, defaultChannelType, msg.channelType_opt) match {
+              case Left(rejectedChannelType) => Left(InvalidChannelType(msg.temporaryChannelId, defaultChannelType, rejectedChannelType))
+              case Right(r) => Right(r)
```
pm47, line 160: @t-bast I wonder why we don't handle the `channel_type` negotiation, as fundee, in `Helpers.validateParamsFundee`? That would be more consistent with the funder workflow, even if we would create a channel actor for nothing (that's what we do for all other parameters).

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -316,15 +320,13 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
...
-      case Some(defaultRouteParams) =>
-        val routeParams = defaultRouteParams.copy(
-          maxFeeProportional = maxFeePct_opt.getOrElse(defaultRouteParams.maxFeeProportional) / 100.0,
-          maxFeeFlat = maxFeeFlat_opt.map(_.toMilliSatoshi).getOrElse(defaultRouteParams.maxFeeFlat)
-        )
-
+      case Right(defaultRouteParams) =>
+        val routeParams = defaultRouteParams
+          .modify(_.boundaries.maxFeeProportional).setToIfDefined(maxFeePct_opt.map(_ / 100))
+          .modify(_.boundaries.maxFeeFlat).setToIfDefined(maxFeeFlat_opt.map(_.toMilliSatoshi))
```
pm47, line 329: cc @thomash-acinq pretty neat right?

thomash-acinq (dismissed):

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
```diff
@@ -135,4 +135,18 @@ object ChannelTypes {
     }
   }
 
+  /** If remote is requesting a non-default channel type, we accept it only if we support required features. */
+  def negotiateChannelType(localFeatures: Features, localChannelType: SupportedChannelType, remoteChannelType_opt: Option[ChannelType]): Either[ChannelType, SupportedChannelType] = remoteChannelType_opt match {
```
t-bast, line 139: Why it `remoteChannelType` optional? Shouldn't this function be used only when the remote provides a `channel_type`? And the switch to use `defautFromFeatures` otherwise should be done by the caller?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: We have to do it here, because we need to know the channel type *before* creating the channel (calling `createNewChannel`), since we need to know whether we must ask `bitcoind` for a static payment point (`static_remotekey`).

It could be refactored though to move the creation of the payment basepoint to the `Channel`, but I'm not sure it's trivial given how we instantiate `ChannelParams` (and if we do that, we should move the funder case as well).

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
pm47: Done with 2a6e652dc. I went with repeating the call to `ChannelTypes.defaultFromFeatures`, which makes it less confusing and looks closer to what we have in `Helpers.validateParamsFunder`.

t-bast: With the latest change, it becomes more apparent that what bothered me is rather that this isn't a negotiation at all: the function either returns `Left(remoteChannelType)` or `Right(remoteChannelType)`, just based on whether we're ok with their proposal or not.

It feels like this function should instead be:

```
def canUseChannelType(localFeatures: Features, remoteChannelType: ChannelType): Boolean
```

It becomes more similar to `canUseFeature` and makes more sense here, don't you think?

pm47: It was going great, except that `negotiateChannelType` includes a cast from `ChannelType` to `SupportedChannelType` in the success case, which we lose with a `Boolean` return type. This doesn't compile:

```scala
val chosenChannelType: Either[InvalidChannelType, SupportedChannelType] = msg.channelType_opt match {
  // remote doesn't specify a channel type: we use spec-defined defaults
  case None => Right(ChannelTypes.defaultFromFeatures(d.localFeatures, d.remoteFeatures))
  // remote explicitly specifies a channel type: we negotiate
  case Some(remoteChannelType) if ChannelTypes.canUseChannelType(d.localFeatures, remoteChannelType) => Right(remoteChannelType)
  case Some(rejectedChannelType) => Left(InvalidChannelType(msg.temporaryChannelId, ChannelTypes.defaultFromFeatures(d.localFeatures, d.remoteFeatures), rejectedChannelType))
}

def canUseChannelType(localFeatures: Features, remoteChannelType: ChannelType): Boolean = remoteChannelType match {
    case _: UnsupportedChannelType => false
    // We ensure that we support the features necessary for this channel type.
    case proposedChannelType: SupportedChannelType => proposedChannelType.features.forall(f => localFeatures.hasFeature(f))
  }

```

Making the return type an `Option[SupportedChannelType]` isn't great either, because it doesn't convey the success/failure idea.

t-bast: Right, that's a good point...
It's not ideal, but I would prefer the solution where we return an `Option[SupportedChannelType]` to the `Either` case.
What about:

```scala
def supportChannelType(localFeatures: Features, remoteChannelType: ChannelType): Option[SupportedChannelType]
```

Which we would use like this:

```scala
case Some(remoteChannelType) => ChannelTypes.supportChannelType(d.localFeatures, remoteChannelType) match {
    case Some(...) => ...
    case None => ...
}
```

pm47: Done with 355799ddc. I went with `areCompatible` instead of `supportChannelType`.

t-bast (approved):

PR #1966, opened 2021-09-23
Log payment failure summary

t-bast:
It's quite cumbersome to investigate complex payment failures.
We need to grep on the parent ID, then group together logs for each child payment, and then we're ready for some analysis.
    
Most of the time, a quick look at the breakdown of all intermediate failures is all we need to diagnose the problem.
This PR adds such a log line, which hopefully will mean less time wasted grepping stuff.

I had to add an `amount` field to all `PaymentFailure`s: you will notice that in a few places, I didn't have the exact information available, let me know if you think the choices I made were wrong.

The resulting log line looks like this:

```
12:14:35.958 INFO  f.a.e.p.s.MultiPartPaymentLifecycle PAY n:025fc0c757afafb9fbae3b39f68fbbdbe81d94a7b10b5884b783c12e583e1d879f h:1e2fa9f432cee805c19fffa49e4aa840e62923d4104bd68eebfaac706ebd60ba p:9770a624 i:9770a624 - failed payment attempts details: {"paymentHash":"1e2fa9f432cee805c19fffa49e4aa840e62923d4104bd68eebfaac706ebd60ba","totalAmount":1000000,"pathFindingExperiment":"my-test-experiment","failures":[{"amount":500000,"route":["02dacc61d8023fbbcf003628f780a0f310c6f059600e7474e8d4e7efdad9952f60","029daed1bd433c0aea37116cc19029cd5e4e761bfc4a4e505bc074da245b24d8d0","039f872e043823050480f74fca31cf1aab02355b3baad582ca6998c393dae9ee52"],"message":"unreadable remote failure"},{"amount":500000,"route":["02dacc61d8023fbbcf003628f780a0f310c6f059600e7474e8d4e7efdad9952f60","02b7508b0be9ff145b4f2f618009167f812790146544dac6cda3c0b7108b795d2b","039f872e043823050480f74fca31cf1aab02355b3baad582ca6998c393dae9ee52"],"message":"unreadable remote failure"},{"amount":1000000,"route":[],"message":"payment attempts exhausted without success"},{"amount":500000,"route":["02dacc61d8023fbbcf003628f780a0f310c6f059600e7474e8d4e7efdad9952f60","029daed1bd433c0aea37116cc19029cd5e4e761bfc4a4e505bc074da245b24d8d0","039f872e043823050480f74fca31cf1aab02355b3baad582ca6998c393dae9ee52"],"message":"039f872e043823050480f74fca31cf1aab02355b3baad582ca6998c393dae9ee52 returned: the complete payment amount was not received within a reasonable time"}]}
```

If you'd like a different formatting, let me know!

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
@@ -26,6 +26,7 @@ import fr.acinq.eclair.wire.protocol.{ChannelDisabled, ChannelUpdate, Node, Temp
 import fr.acinq.eclair.{MilliSatoshi, ShortChannelId}
 
 import java.util.UUID
+import scala.util.{Failure, Success, Try}
 
 /**
  * Created by PM on 01/02/2017.
@@ -114,17 +115,20 @@ object PaymentReceived {
 case class PaymentSettlingOnChain(id: UUID, amount: MilliSatoshi, paymentHash: ByteVector32, timestamp: Long = System.currentTimeMillis) extends PaymentEvent
 
 sealed trait PaymentFailure {
+  // @formatter:off
+  def amount: MilliSatoshi
   def route: Seq[Hop]
+  // @formatter:on
 }
 
 /** A failure happened locally, preventing the payment from being sent (e.g. no route found). */
-case class LocalFailure(route: Seq[Hop], t: Throwable) extends PaymentFailure
+case class LocalFailure(amount: MilliSatoshi, route: Seq[Hop], t: Throwable) extends PaymentFailure
 
 /** A remote node failed the payment and we were able to decrypt the onion failure packet. */
-case class RemoteFailure(route: Seq[Hop], e: Sphinx.DecryptedFailurePacket) extends PaymentFailure
+case class RemoteFailure(amount: MilliSatoshi, route: Seq[Hop], e: Sphinx.DecryptedFailurePacket) extends PaymentFailure
 
 /** A remote node failed the payment but we couldn't decrypt the failure (e.g. a malicious node tampered with the message). */
-case class UnreadableRemoteFailure(route: Seq[Hop]) extends PaymentFailure
+case class UnreadableRemoteFailure(amount: MilliSatoshi, route: Seq[Hop]) extends PaymentFailure
 
 object PaymentFailure {
 
@@ -144,8 +148,16 @@ object PaymentFailure {
    */
   def transformForUser(failures: Seq[PaymentFailure]): Seq[PaymentFailure] = {
     failures match {
-      case previousFailures :+ LocalFailure(_, RouteNotFound) if previousFailures.nonEmpty => previousFailures
+      case previousFailures :+ LocalFailure(_, _, RouteNotFound) if previousFailures.nonEmpty => previousFailures
       case other => other
+    }
+  }
+
+  def jsonSummary(paymentHash: ByteVector32, totalAmount: MilliSatoshi, pathFindingExperiment: String, paymentFailed: PaymentFailed): String = {
+    import fr.acinq.eclair.json.{JsonSerializers, PaymentFailedSummary}
+    Try(JsonSerializers.serialization.write(PaymentFailedSummary(paymentHash, totalAmount, pathFindingExperiment, paymentFailed))(JsonSerializers.formats)) match {
+      case Failure(e) => s"json serialization failed: ${e.getMessage}"
+      case Success(json) => json
     }
   }
 
@@ -155,7 +167,7 @@ object PaymentFailure {
    */
   def hasAlreadyFailedOnce(nodeId: PublicKey, failures: Seq[PaymentFailure]): Boolean =
     failures
-      .collectFirst { case RemoteFailure(_, Sphinx.DecryptedFailurePacket(origin, u: Update)) if origin == nodeId => u.update }
+      .collectFirst { case RemoteFailure(_, _, Sphinx.DecryptedFailurePacket(origin, u: Update)) if origin == nodeId => u.update }
       .isDefined
 
   /** Ignore the channel outgoing from the given nodeId in the given route. */
@@ -170,12 +182,12 @@ object PaymentFailure {
 
   /** Update the set of nodes and channels to ignore in retries depending on the failure we received. */
   def updateIgnored(failure: PaymentFailure, ignore: Ignore): Ignore = failure match {
-    case RemoteFailure(hops, Sphinx.DecryptedFailurePacket(nodeId, _)) if nodeId == hops.last.nextNodeId =>
+    case RemoteFailure(_, hops, Sphinx.DecryptedFailurePacket(nodeId, _)) if nodeId == hops.last.nextNodeId =>
       // The failure came from the final recipient: the payment should be aborted without penalizing anyone in the route.
       ignore
-    case RemoteFailure(_, Sphinx.DecryptedFailurePacket(nodeId, _: Node)) =>
+    case RemoteFailure(_, _, Sphinx.DecryptedFailurePacket(nodeId, _: Node)) =>
       ignore + nodeId
-    case RemoteFailure(hops, Sphinx.DecryptedFailurePacket(nodeId, failureMessage: Update)) =>
+    case RemoteFailure(_, hops, Sphinx.DecryptedFailurePacket(nodeId, failureMessage: Update)) =>
       if (Announcements.checkSig(failureMessage.update, nodeId)) {
         val shouldIgnore = failureMessage match {
           case _: TemporaryChannelFailure => true
@@ -192,13 +204,13 @@ object PaymentFailure {
         // This node is fishy, it gave us a bad signature, so let's filter it out.
         ignore + nodeId
       }
-    case RemoteFailure(hops, Sphinx.DecryptedFailurePacket(nodeId, _)) =>
+    case RemoteFailure(_, hops, Sphinx.DecryptedFailurePacket(nodeId, _)) =>
       ignoreNodeOutgoingChannel(nodeId, hops, ignore)
-    case UnreadableRemoteFailure(hops) =>
+    case UnreadableRemoteFailure(_, hops) =>
       // We don't know which node is sending garbage, let's blacklist all nodes except the one we are directly connected to and the final recipient.
       val blacklist = hops.map(_.nextNodeId).drop(1).dropRight(1).toSet
       ignore ++ blacklist
-    case LocalFailure(hops, _) => hops.headOption match {
+    case LocalFailure(_, hops, _) => hops.headOption match {
       case Some(hop: ChannelHop) =>
         val faultyChannel = ChannelDesc(hop.lastUpdate.shortChannelId, hop.nodeId, hop.nextNodeId)
         ignore + faultyChannel
@@ -216,7 +228,7 @@ object PaymentFailure {
     // We're only interested in the last channel update received per channel.
     val updates = failures.foldLeft(Map.empty[ShortChannelId, ChannelUpdate]) {
       case (current, failure) => failure match {
-        case RemoteFailure(_, Sphinx.DecryptedFailurePacket(_, f: Update)) => current.updated(f.update.shortChannelId, f.update)
+        case RemoteFailure(_, _, Sphinx.DecryptedFailurePacket(_, f: Update)) => current.updated(f.update.shortChannelId, f.update)
         case _ => current
       }
     }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
@@ -20,6 +20,7 @@ import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.crypto.Sphinx
 import fr.acinq.eclair.payment.PaymentRequest.ExtraHop
+import fr.acinq.eclair.payment.send.PaymentInitiator.SendPaymentConfig
 import fr.acinq.eclair.router.Announcements
 import fr.acinq.eclair.router.Router.{ChannelDesc, ChannelHop, Hop, Ignore}
 import fr.acinq.eclair.wire.protocol.{ChannelDisabled, ChannelUpdate, Node, TemporaryChannelFailure}
@@ -153,9 +154,11 @@ object PaymentFailure {
     }
   }
 
-  def jsonSummary(paymentHash: ByteVector32, totalAmount: MilliSatoshi, pathFindingExperiment: String, paymentFailed: PaymentFailed): String = {
-    import fr.acinq.eclair.json.{JsonSerializers, PaymentFailedSummary}
-    Try(JsonSerializers.serialization.write(PaymentFailedSummary(paymentHash, totalAmount, pathFindingExperiment, paymentFailed))(JsonSerializers.formats)) match {
+  case class PaymentFailedSummary(cfg: SendPaymentConfig, pathFindingExperiment: String, paymentFailed: PaymentFailed)
+
+  def jsonSummary(cfg: SendPaymentConfig, pathFindingExperiment: String, paymentFailed: PaymentFailed): String = {
+    import fr.acinq.eclair.json.JsonSerializers
+    Try(JsonSerializers.serialization.write(PaymentFailedSummary(cfg, pathFindingExperiment, paymentFailed))(JsonSerializers.formats)) match {
       case Failure(e) => s"json serialization failed: ${e.getMessage}"
       case Success(json) => json
     }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
@@ -16,9 +16,6 @@
 
 package fr.acinq.eclair.payment.send
 
-import java.util.UUID
-import java.util.concurrent.TimeUnit
-
 import akka.actor.{ActorRef, FSM, Props, Status}
 import akka.event.Logging.MDC
 import fr.acinq.bitcoin.ByteVector32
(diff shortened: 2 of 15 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -230,13 +230,14 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
     event match {
       case Left(paymentFailed) =>
         log.warning("multi-part payment failed")
+        log.info(s"failed payment attempts details:\n    + ${paymentFailed.failures.map(f => PaymentFailure.print(f)).mkString("\n    + ")}")
```
pm47, line 233: I would split the string first, and then `foreach(log.info(...))`. That way all lines are prefixed by the MDC and can be grepped at once.

An alternative to this would be to print a compact (meaning: no line breaks) json representation of an object. That way the formatting would be defined by the object structure, instead of clever indenting. Another advantage is that we can store and parse this debug info in db, and also we can encapsulate mutliple payment attempts in a higher-level object. WDYT?

pm47 (commented):
> I had to add an amount field to all PaymentFailures: you will notice that in a few places, I didn't have the exact information available, let me know if you think the choices I made were wrong.

Can you point out exactly where you had to put approximate data?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -92,14 +92,14 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
...
-        val (toSend, maxFee) = remainingToSend(nodeParams, d.request, d.pending.values)
+        val (toSend, maxFee) = remainingToSend(d.request, d.pending.values)
         log.debug("retry sending {} with maximum fee {} without ignoring channels ({})", toSend, maxFee, d.ignore.channels.map(_.shortChannelId).mkString(","))
         val routeParams = d.request.routeParams.copy(randomize = true) // we randomize route selection when we retry
         router ! createRouteRequest(nodeParams, toSend, maxFee, routeParams, d, cfg).copy(ignore = d.ignore.emptyChannels())
         retriedFailedChannels = true
         stay() using d.copy(remainingAttempts = (d.remainingAttempts - 1).max(0), ignore = d.ignore.emptyChannels())
       } else {
-        val failure = LocalFailure(Nil, t)
+        val failure = LocalFailure(d.request.totalAmount - d.pending.values.map(_.amount).sum, Nil, t)
```
t-bast, line 102: The amount we set here may be inaccurate. I don't think it matters much though.
Here is the scenario where it may be inaccurate:

- we have 3 child payments of 100 sats each
- 1 of them fails: we ask for a new route for an amount of 100 sats
- a second one fails before we receive a response from the router: that will trigger a second route request for 100 sats
- we receive a router error for the first route request: we will over-estimate the amount and count 200 sats in the `LocalFailure`

I think this is ok, as:

- this is an edge case
- the amount in `PaymentFailure` shouldn't be used for business logic, just for reporting/logs

Let me know what you think!

t-bast:
> Can you point out exactly where you had to put approximate data?

Sure, it's here: https://github.com/ACINQ/eclair/pull/1966#discussion_r714937306

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
t-bast: > I would split the string first, and then foreach(log.info(...)). That way all lines are prefixed by the MDC and can be grepped at once.

That's what I actually want to avoid, because then we need to grep (because unrelated log lines may have been inserted), and each individual log line is confusing when taken outside of the context of the others...I really want this to be in a single log, like exceptions stack traces, so that it's really easy to just read in a log file.

> An alternative to this would be to print a compact (meaning: no line breaks) json representation of an object. 

I considered that, and I'm hesitating. I chose against it because my first need is readability / quick understanding of the payment breakdown, similarly to printed stack traces. But a JSON object could easily be copy-pasted in a `.json` file and easy to read, it's one extra step but it's not too bad.

I don't know what to decide, let me know what you prefer if you can help me tilt the decision towards one or the other!

pm47: > That's what I actually want to avoid, because then we need to grep (because unrelated log lines may have been inserted), and each individual log line is confusing when taken outside of the context of the others...I really want this to be in a single log, like exceptions stack traces, so that it's really easy to just read in a log file.

You mean in a separate log file? But there will still be concurrency between several independent payments and logs will be intertwinned (altough less frequently, granted). I think there is a case for putting the MDC even if we go for a separate file.

> I considered that, and I'm hesitating. I chose against it because my first need is readability / quick understanding of the payment breakdown, similarly to printed stack traces. But a JSON object could easily be copy-pasted in a .json file and easy to read, it's one extra step but it's not too bad.
> I don't know what to decide, let me know what you prefer if you can help me tilt the decision towards one or the other!

Here is another argument in favor of json: we could return it in our API response when some debug flag is enabled.

t-bast: > You mean in a separate log file?

No, I meant a single `log` call to ensure it's atomic and easy to read in one go.
This way you're sure that this will not be intertwined with other, unrelated logs.

> Here is another argument in favor of json: we could return it in our API response when some debug flag is enabled.

Ok let's go for a JSON structure then.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1966/commits/26d4a7699f78eb3a6e41ce1ede21bb704c31f549

pm47: > I meant a single log call to ensure it's atomic and easy to read in one go.
> This way you're sure that this will not be intertwined with other, unrelated logs.

Oh I see. I thought multi-line logs would be intertwined but maybe you are right!

t-bast: Actually I'm not sure at all, I suppose they're not but it's hard to actually verify!

eclair-core/src/test/scala/fr/acinq/eclair/db/PaymentsDbSpec.scala
```diff
@@ -481,7 +481,7 @@ class PaymentsDbSpec extends AnyFunSuite {
       db.updateOutgoingPayment(PaymentFailed(s3.id, s3.paymentHash, Nil, 310))
       val ss3 = s3.copy(status = OutgoingPaymentStatus.Failed(Nil, 310))
       assert(db.getOutgoingPayment(s3.id) === Some(ss3))
-      db.updateOutgoingPayment(PaymentFailed(s4.id, s4.paymentHash, Seq(LocalFailure(Seq(hop_ab), new RuntimeException("woops")), RemoteFailure(Seq(hop_ab, hop_bc), Sphinx.DecryptedFailurePacket(carol, UnknownNextPeer))), 320))
+      db.updateOutgoingPayment(PaymentFailed(s4.id, s4.paymentHash, Seq(LocalFailure(500 msat, Seq(hop_ab), new RuntimeException("woops")), RemoteFailure(400 msat, Seq(hop_ab, hop_bc), Sphinx.DecryptedFailurePacket(carol, UnknownNextPeer))), 320))
```
thomash-acinq, line 484: Where do these values come from? I can't link them to any of the other values above.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -144,18 +147,40 @@ object PaymentFailure {
...
+  def summary(paymentFailed: PaymentFailed): String = {
+    import org.json4s.JsonAST.{JArray, JField, JObject, JString}
+    import org.json4s.{DefaultFormats, jackson}
+    val failures = paymentFailed.failures.map(failure => {
+      val details = failure match {
+        case LocalFailure(_, _, t) => t.getMessage
+        case RemoteFailure(_, _, Sphinx.DecryptedFailurePacket(origin, failureMessage)) => s"$origin returned: ${failureMessage.message}"
+        case _: UnreadableRemoteFailure => "unreadable remote failure"
+      }
+      val route = JArray(failure.route.map(r => JString(r.nextNodeId.value.toHex)).toList)
```
pm47, line 167: We're missing the last hop here:
```suggestion
      val route = JArray((failure.route.map(_.nodeId) ++ failure.route.lastOption.map(_.nextNodeId).toList).map(r => JString(r.nextNodeId.value.toHex)).toList)
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/1966/commits/fd216f2abbd0a1163197e7915eca8cccb25c2465

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -92,14 +92,14 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
         // Channels are mostly ignored for temporary reasons, likely because they didn't have enough balance to forward
         // the payment. When we're retrying an MPP split, it may make sense to retry those ignored channels because with
         // a different split, they may have enough balance to forward the payment.
-        val (toSend, maxFee) = remainingToSend(nodeParams, d.request, d.pending.values)
+        val (toSend, maxFee) = remainingToSend(d.request, d.pending.values)
```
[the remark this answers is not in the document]
t-bast: Good idea, done in https://github.com/ACINQ/eclair/pull/1966/commits/fd216f2abbd0a1163197e7915eca8cccb25c2465

eclair-core/src/test/scala/fr/acinq/eclair/db/PaymentsDbSpec.scala
t-bast: I clarified it in https://github.com/ACINQ/eclair/pull/1966/commits/fd216f2abbd0a1163197e7915eca8cccb25c2465 (I used the amount from the right payment)

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
pm47: Does this means it fixes the approximation described here: https://github.com/ACINQ/eclair/pull/1966#discussion_r714937306?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
pm47: We're also missing the total amount, don't you think? Maybe we should define a purpose built `PaymentFailureSummary` class, that could also contain the experiment name, etc. (or even the whole payment config)?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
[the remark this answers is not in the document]
t-bast: No, it just avoids duplicating the code that computes the amount, but it can still be inaccurate during race condition.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
t-bast: Good catch, the total amount would be useful to include.
And why not include the experiment name as well, it doesn't cost anything to add!

t-bast: Added in https://github.com/ACINQ/eclair/pull/1966/commits/66f0a71e2ef8deebc7f680f274fbde554f17ec31

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -232,6 +235,31 @@ object PaymentFailure {
...
+      case _: UnreadableRemoteFailure => "unreadable remote failure"
+    }
+    PaymentFailureSummary(failure.amount, route, message)
+  }
+}
+
+case class PaymentFailedSummary(totalAmount: MilliSatoshi, pathFindingExperiment: String, failures: Seq[PaymentFailureSummary]) {
+  def json(): String = {
+    import fr.acinq.eclair.json.JsonSerializers
+    JsonSerializers.serialization.write(this)(JsonSerializers.formats)
```
pm47, line 255: This needs to be in a `Try` (json serialization errors do happen):

```suggestion
    import scala.util.Try
    Try(JsonSerializers.serialization.write(this)(JsonSerializers.formats)).getOrElse("json serialization error")
```

(or make the method return the `Try` and handle the error above)

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -232,6 +235,31 @@ object PaymentFailure {
 
 }
 
+case class PaymentFailureSummary(amount: MilliSatoshi, route: Seq[PublicKey], message: String)
```
pm47, line 238: After rebase, I suggest we move this to the `json` serializers as `PaymentFailureJson`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -232,6 +235,31 @@ object PaymentFailure {
...
+    val message = failure match {
+      case LocalFailure(_, _, t) => t.getMessage
+      case RemoteFailure(_, _, Sphinx.DecryptedFailurePacket(origin, failureMessage)) => s"$origin returned: ${failureMessage.message}"
+      case _: UnreadableRemoteFailure => "unreadable remote failure"
+    }
+    PaymentFailureSummary(failure.amount, route, message)
+  }
+}
+
+case class PaymentFailedSummary(totalAmount: MilliSatoshi, pathFindingExperiment: String, failures: Seq[PaymentFailureSummary]) {
```
pm47, line 252: Why not include the whole config? We can always filter out the relevant attributes in the serializer.

```suggestion
case class PaymentFailedSummary(config: SendPaymentConfig, failures: Seq[PaymentFailureSummary]) {
```

pm47 (commented):
Needs rebase.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
t-bast: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
t-bast: I moved it to the `json` serializers but didn't rename it `PaymentFailureJson` and didn't provide an implicit conversion, because in most cases we don't want it (we want the websocket and API to return a "full" `PaymentFailure`).

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
t-bast: Meh, it doesn't really contain the interesting bits, more than half of the fields of `SendPaymentConfig` are things we really don't care about

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
pm47: I'm not a fan: in my mind the classes in `JsonSerializers` are serialization details and should be `private`. Now we are in-between. Also we don't have to add the serializers to the `JsonSerializers.formats` if we don't want it in the general case.

This is what I mean: 8a9298d1a56634d355d39f5389f63d3b3e6f233d. Also I would have gone with a slighly different format for failure summary: 4bab67ebdafa6a92bcc984178ce8f836233a583f.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
pm47: True, but selecting the relevant fields could be considered a serialization detail. For example, in 4bab67ebdafa6a92bcc984178ce8f836233a583f I added the destination node id and had to update the definition of `PaymentFailedSummary`.

pm47 (approved):
We have probably bikeshed enough already, so I'm acking this as-is and let you decide if you take my additional suggestions.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
t-bast: Ok, I didn't understand what you previously meant, I thought you didn't want to have any `case class` in the `PaymentEvents` file anymore. I'll apply what you suggest.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
t-bast: Right, why not in that case.

t-bast: Done in https://github.com/ACINQ/eclair/pull/1966/commits/eb9f63309a905c81e3ef31d30b76302495af0d72

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/1966/commits/eb9f63309a905c81e3ef31d30b76302495af0d72, but I didn't include all the additional data you wanted in `PaymentFailure`, I'm really not sure it's useful. We'll see when actually using these logs, we can add later if necessary.

pm47 (approved):

PR #1969, opened 2021-09-26
Additional parameters for findroute* API calls

rorp:
This PR adds new parameters for `findroute*` API calls useful automated route calculations for third-party applications:

`ignoreNodeIds` - specifies ignored nodes
`ignoreChannelIds` - specifies ignored channels 
`maxFeeMsat` - sets max fee for route calculations

Also it adds `full` format that returns all information that `Router` can provide about the found routes.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -49,6 +49,7 @@ import scodec.bits.ByteVector
 
 import java.nio.charset.StandardCharsets
 import java.util.UUID
+import scala.collection.immutable.SortedMap
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future, Promise}
 import scala.reflect.ClassTag
@@ -123,9 +124,9 @@ trait Eclair {
 
   def sendOnChain(address: String, amount: Satoshi, confirmationTarget: Long): Future[ByteVector32]
 
-  def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, pathFindingExperimentName_opt: Option[String], assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, includeLocalChannelCost: Boolean = false)(implicit timeout: Timeout): Future[RouteResponse]
+  def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, pathFindingExperimentName_opt: Option[String], assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, includeLocalChannelCost: Boolean = false, ignoredNodeIds: Seq[PublicKey] = Seq.empty, ignoredShortChannelIds: Seq[ShortChannelId] = Seq.empty, maxFee_opt: Option[MilliSatoshi] = None)(implicit timeout: Timeout): Future[RouteResponse]
 
-  def findRouteBetween(sourceNodeId: PublicKey, targetNodeId: PublicKey, amount: MilliSatoshi, pathFindingExperimentName_opt: Option[String], assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, includeLocalChannelCost: Boolean = false)(implicit timeout: Timeout): Future[RouteResponse]
+  def findRouteBetween(sourceNodeId: PublicKey, targetNodeId: PublicKey, amount: MilliSatoshi, pathFindingExperimentName_opt: Option[String], assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, includeLocalChannelCost: Boolean = false, ignoredNodeIds: Seq[PublicKey] = Seq.empty, ignoredShortChannelIds: Seq[ShortChannelId] = Seq.empty, maxFee_opt: Option[MilliSatoshi] = None)(implicit timeout: Timeout): Future[RouteResponse]
 
   def sendToRoute(amount: MilliSatoshi, recipientAmount_opt: Option[MilliSatoshi], externalId_opt: Option[String], parentId_opt: Option[UUID], invoice: PaymentRequest, finalCltvExpiryDelta: CltvExpiryDelta, route: PredefinedRoute, trampolineSecret_opt: Option[ByteVector32] = None, trampolineFees_opt: Option[MilliSatoshi] = None, trampolineExpiryDelta_opt: Option[CltvExpiryDelta] = None, trampolineNodes_opt: Seq[PublicKey] = Nil)(implicit timeout: Timeout): Future[SendPaymentToRouteResponse]
 
@@ -280,8 +281,8 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
     }
   }
 
-  override def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, pathFindingExperimentName_opt: Option[String], assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, includeLocalChannelCost: Boolean = false)(implicit timeout: Timeout): Future[RouteResponse] =
-    findRouteBetween(appKit.nodeParams.nodeId, targetNodeId, amount, pathFindingExperimentName_opt, assistedRoutes, includeLocalChannelCost)
+  override def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, pathFindingExperimentName_opt: Option[String], assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, includeLocalChannelCost: Boolean = false, ignoredNodeIds: Seq[PublicKey] = Seq.empty, ignoredShortChannelIds: Seq[ShortChannelId] = Seq.empty, maxFee_opt: Option[MilliSatoshi] = None)(implicit timeout: Timeout): Future[RouteResponse] =
+    findRouteBetween(appKit.nodeParams.nodeId, targetNodeId, amount, pathFindingExperimentName_opt, assistedRoutes, includeLocalChannelCost, ignoredNodeIds, ignoredShortChannelIds, maxFee_opt)
 
   private def getRouteParams(pathFindingExperimentName_opt: Option[String]): Either[IllegalArgumentException, RouteParams] = {
     pathFindingExperimentName_opt match {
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -123,9 +124,9 @@ trait Eclair {
 
   def sendOnChain(address: String, amount: Satoshi, confirmationTarget: Long): Future[ByteVector32]
 
-  def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, pathFindingExperimentName_opt: Option[String], assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, includeLocalChannelCost: Boolean = false)(implicit timeout: Timeout): Future[RouteResponse]
+  def findRoute(targetNodeId: PublicKey, amount: MilliSatoshi, pathFindingExperimentName_opt: Option[String], assistedRoutes: Seq[Seq[PaymentRequest.ExtraHop]] = Seq.empty, includeLocalChannelCost: Boolean = false, ignoredNodeIds: Seq[PublicKey] = Seq.empty, ignoredShortChannelIds: Seq[ShortChannelId] = Seq.empty, maxFee_opt: Option[MilliSatoshi] = None)(implicit timeout: Timeout): Future[RouteResponse]
```
t-bast, line 127: nit: I would call these `ignoreNodeIds` and `ignoreShortChannelIds` (instead of `ignoredXXX`), it's more consistent.

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
```diff
@@ -221,13 +221,19 @@ class ColorSerializer extends CustomSerializerOnly[Color](_ => {
...
-    val nodeIds = route.routes.head.hops match {
-      case rest :+ last => rest.map(_.nodeId) :+ last.nodeId :+ last.nextNodeId
-      case Nil => Nil
-    }
-    JArray(nodeIds.toList.map(n => JString(n.toString)))
+class RouteSerializer extends CustomSerializerOnly[Route](_ => {
+  case route: Route =>
+    JObject(
+      ("amount", JLong(route.amount.toLong)),
+      ("hops", JArray(route.hops.map(hop => Extraction.decompose(hop)(
```
t-bast, line 228: Does that mean you're printing the whole `channel_update` for each hop?
Is that really necessary?
Most of the fields don't make sense (e.g. signature, chainhash), it would be better to have a more fine-grained choice of what exactly we print.

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/RouteFormat.scala
```diff
@@ -16,38 +16,47 @@
...
-    case ShortChannelIdRouteFormat =>
-      val shortChannelIds = route.routes.head.hops.map(_.lastUpdate.shortChannelId)
-      shortChannelIds.map(_.toString)
+    case FULL => FullRouteFormat
+    case _ => throw new IllegalArgumentException(s"invalid route format, possible values are ($NODE_ID, $SHORT_CHANNEL_ID, $FULL)")
   }
 
+  def format(route: RouteResponse, format_opt: Option[RouteFormat]): HttpResponse = format(route, format_opt.getOrElse(NodeIdRouteFormat))
+
+  def format(route: RouteResponse, format: RouteFormat): HttpResponse =
```
t-bast, line 46: Does this really need to return an `HttpResponse`? Can't it be a `String` and akka-http takes care of putting that inside an http response?

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -1039,6 +1039,27 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
         )))
         eclair.findRoute(PublicKey.fromBin(ByteVector.fromValidHex("036ded9bb8175d0c9fd3fad145965cf5005ec599570f35c682e710dc6001ff605e")), 456.msat, any, any)(any[Timeout]).wasCalled(threeTimes)
       }
+    Post("/findroute", FormData("format" -> "full", "invoice" -> invoice, "amountMsat" -> "456")) ~>
```
t-bast, line 1042: nit: add line before

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -1039,6 +1039,27 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
...
+        assert(responseArray.arr.head.isInstanceOf[JObject])
+        val route = responseArray.arr.head.asInstanceOf[JObject]
+        assert(route.obj.head == ("amount", JInt(456)))
+        assert(route.obj.last._1 == "hops")
+        val hops = route.obj.last._2.asInstanceOf[JArray]
+        assert(hops.arr.size == 3)
+        val (hop1 :: hop2 :: hop3 :: Nil) = hops.arr
+        assert(hop1.asInstanceOf[JObject].obj.head == ("nodeId", JString(mockHop1.nodeId.toString())))
+        assert(hop2.asInstanceOf[JObject].obj.head == ("nodeId", JString(mockHop2.nodeId.toString())))
+        assert(hop3.asInstanceOf[JObject].obj.head == ("nodeId", JString(mockHop3.nodeId.toString())))
```
t-bast, line 1060: Please compare to a complete JSON with `matchTestJson`

t-bast (commented):
Can you update the release notes as well since this PR updates some APIs?

And please post detailed results of how you manage to use this for rebalancing.

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
rorp: 1. that's the only way to get a channel ID. 
2. channel updates provide fee rates which are used for figuring out economical viability of found routes
3. one can get channel updates using `allupdates`, but without a channel ID it's unclear how to get an update when two nodes have more than one channel between each other. 
4. since there is an update for every hop, there's no need to call `allupdates` at all

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/RouteFormat.scala
rorp: String produces `text/plain`, most of JSON-RPC client libraries require `application/json`

rorp:
> Can you update the release notes as well since this PR updates some APIs?

I don't know how to it

rorp:
> And please post detailed results of how you manage to use this for rebalancing.

```python
        ignore_node_ids = [self.get_own_pubkey()]

        if isinstance(ignored_nodes, list):
            ignore_node_ids = ignore_node_ids + ignored_nodes

        params = {
            'sourceNodeId': first_hop_pubkey,
            'targetNodeId': last_hop_pubkey,
            'amountMsat': int(amount * 1000),
            'format': 'full',
            'ignoreNodeIds': ignore_node_ids,
            'ignoreChannelIds': ignore_channel_ids,
            'maxFeeMsat': fee_limit,
        }
        found_routes = self.call_eclair("findroutebetweennodes", params)
        if isinstance(found_routes, list):
            routes = []
            for found_route in found_routes:
                amount_msat = found_route['amount']
                hops = [self.route_to_hop(hop, amount_msat) for hop in found_route['hops']]
                if first_hop_channel:
                    hops.insert(0, first_hop_channel.to_hop(amount_msat, 0, first=True))
                if last_hop_channel:
                    hops.append(
                        last_hop_channel.to_hop(amount_msat, self.calc_fees_msat(amount_msat, last_hop_channel.chan_id),
                                                first=False))
                routes.append(Route(amount_msat, hops))
            return routes
```

t-bast:
> I don't know how to it

Simply update this section: https://github.com/ACINQ/eclair/blob/master/docs/release-notes/eclair-vnext.md#api-changes

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
t-bast: Let's wait for https://github.com/ACINQ/eclair/pull/1979 to be integrated, this will give you an opportunity to more easily fine-tune the serialization you want to use for this case.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -490,4 +495,15 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
     val pubKeyFromSignature = Crypto.recoverPublicKey(signature, signedBytes, recoveryId)
     VerifiedMessage(valid = true, pubKeyFromSignature)
   }
+
+  private def getChannelDescs(shortChannelIds: Set[ShortChannelId])(implicit timeout: Timeout): Future[Set[ChannelDesc]] = {
```
t-bast, line 499: If `shortChannelIds` is empty, you should immediately return an empty set instead of making a call to the `Router` for no reason.

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -1053,6 +1053,33 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
...
+//        val route = responseArray.arr.head.asInstanceOf[JObject]
+//        assert(route.obj.head == ("amount", JInt(456)))
+//        assert(route.obj.last._1 == "hops")
+//        val hops = route.obj.last._2.asInstanceOf[JArray]
+//        assert(hops.arr.size == 3)
+//        val (hop1 :: hop2 :: hop3 :: Nil) = hops.arr
+//        assert(hop1.asInstanceOf[JObject].obj.head == ("nodeId", JString(mockHop1.nodeId.toString())))
+//        assert(hop2.asInstanceOf[JObject].obj.head == ("nodeId", JString(mockHop2.nodeId.toString())))
+//        assert(hop3.asInstanceOf[JObject].obj.head == ("nodeId", JString(mockHop3.nodeId.toString())))
+//        eclair.findRoute(PublicKey.fromBin(ByteVector.fromValidHex("036ded9bb8175d0c9fd3fad145965cf5005ec599570f35c682e710dc6001ff605e")), 456.msat, any, any)(any[Timeout]).wasCalled(fourTimes)
```
t-bast, line 1081: Remove?

t-bast (commented):
Concept ACK, we should integrate this soon (we're planning a release and it would be nice to have this PR be part of it).

Two things to consider:

- https://github.com/ACINQ/eclair/pull/1979 will be merged soon and provides more flexibility for JSON serialization: I think you can leverage this to more cleanly apply the right serialization based on the requested route format
- https://github.com/ACINQ/eclair/pull/1974 will change the internal logic of ignoring channels. When you ask to ignore a channel between Bob and Carol, we will ignore ALL channels between them (because non-strict forwarding means that if one channel cannot be used, none of them can unless the node is very buggy). The only exception is obviously for local channels. IIUC for rebalancing you only need to use this with local channels, right? If that's the case it should work properly.

Regarding release notes, it would be great if you could add a new "Circular rebalancing" section to https://github.com/ACINQ/eclair/blob/master/docs/release-notes/eclair-vnext.md where you explain how the eclair API can be used to rebalance your channels. You should describe the example of a triangular topology and show the concrete `eclair-cli` commands that should be used.

rorp:
> Regarding release notes, it would be great if you could add a new "Circular rebalancing" section to https://github.com/ACINQ/eclair/blob/master/docs/release-notes/eclair-vnext.md where you explain how the eclair API can be used to rebalance your channels. You should describe the example of a triangular topology and show the concrete `eclair-cli` commands that should be used.

I think it should be in a separate doc

t-bast:
> I think it should be in a separate doc

Good idea, a separate doc would be very useful.

PR #1974, opened 2021-09-29
Ignore redundant channels

thomash-acinq:
If we ignore a remote channel (because it couldn't route a payment previously or because we've already used it for another part of the payment), we should ignore all of its siblings too (other channels between the same nodes). Nodes can ignore the channel id and route through any channel to the next node so trying another channel id is unlikely to succeed.
However for local channels, we have perfect information. We ignore only individual local channels so that we can optimize their use.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -20,7 +20,7 @@ import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{Btc, ByteVector32, MilliBtc, Satoshi}
 import fr.acinq.eclair._
 import fr.acinq.eclair.payment.relay.Relayer.RelayFees
-import fr.acinq.eclair.router.Graph.GraphStructure.{DirectedGraph, GraphEdge}
+import fr.acinq.eclair.router.Graph.GraphStructure.{DirectedGraph, GraphEdge, IgnoredEdges}
 import fr.acinq.eclair.router.Router._
 import fr.acinq.eclair.wire.protocol.ChannelUpdate
 
@@ -109,7 +109,7 @@ object Graph {
                         sourceNode: PublicKey,
                         targetNode: PublicKey,
                         amount: MilliSatoshi,
-                        ignoredEdges: Set[ChannelDesc],
+                        ignoredEdges: IgnoredEdges,
                         ignoredVertices: Set[PublicKey],
                         extraEdges: Set[GraphEdge],
                         pathsToFind: Int,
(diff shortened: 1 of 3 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -22,7 +22,7 @@ import fr.acinq.bitcoin.{Block, ByteVector32, ByteVector64, Satoshi, SatoshiLong
 import fr.acinq.eclair.payment.PaymentRequest.ExtraHop
 import fr.acinq.eclair.payment.relay.Relayer.RelayFees
 import fr.acinq.eclair.router.Graph.GraphStructure.DirectedGraph.graphEdgeToHop
-import fr.acinq.eclair.router.Graph.GraphStructure.{DirectedGraph, GraphEdge}
+import fr.acinq.eclair.router.Graph.GraphStructure.{DirectedGraph, GraphEdge, IgnoredEdges}
```
t-bast, line 25: Can you add a test to this file for this new behavior?
It looks like we don't have tests where we have multiple channels between two nodes and ignore one of them.
The new test should verify that all channels are actually ignored.

It would also be great to test that local channels are exempted.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -418,6 +418,34 @@ object Graph {
 
     }
 
+    case class IgnoredEdges(ignoredLocalEdges: Set[ShortChannelId], ignoredRemoteEdges: Set[(PublicKey, PublicKey)]) {
```
t-bast, line 421: Can you document this class? It's worth explaining why instead of ignoring a single channel, we ignore all channels between the two nodes (because we assume they're using non-strict forwarding, so none of their channels would work).

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -418,6 +418,34 @@ object Graph {
...
+        edges.iterator.foldLeft(this)((ignored, e) => ignored.add(e))
+      }
+    }
+
+    object IgnoredEdges {
+      val empty: IgnoredEdges = IgnoredEdges(Set.empty, Set.empty)
+
+      def apply(descs: Set[ChannelDesc]): IgnoredEdges = {
+        IgnoredEdges(descs.map(d => d.shortChannelId), descs.map(d => (d.a, d.b)))
+      }
```
t-bast, line 446: This is a bit hackish, since an edge should only be in one of the two sets...why not also include our own `nodeId` in that constructor, so that we know which channels are local (by comparing to `d.a`)?

Our `nodeId` can always be found in our `nodeParams`, so it should be easy to propagate it all the way to here.

t-bast (commented):
I was expecting you to change `PaymentLifecycle` and the `Router` to implement this, but it's actually smarter to directly make that change in the path-finding like you did, good choice!

I think it's a great improvement for our retry logic and should provide visible improvements to our payment success rate.

However, I'm a tiny bit afraid, because _it works only if remote nodes correctly do non-strict forwarding_. If a remote node doesn't do non-strict forwarding, we'll penalize all of their channels for no reason. But I think all nodes should implement non-strict forwarding for the network to work well, so I'm ok with penalizing them if they don't.

There's another side-effect I'm more afraid of: the `ExcludeChannel` mechanism.
Whenever we use it, we will now exclude all channels between these two nodes for 5 minutes, _for all payment attempts (even for unrelated payments)_.

Here is the scenario that will create issues:

- Bob and Carol have 5 channels, where Bob's balance is below 50 000 sats
- Alice tries sending a big payment to Carol (300 000 sats)
- Bob will return a `TemporaryChannelFailure`, because he doesn't have the liquidity to route this payment
- Alice will ban the Bob <-> Carol edge for all payments for 5 minutes
- Alice tries to send a smaller payment to Carol (10 000 sats): it cannot find a route because we excluded Bob, whereas the payment would have succeeded

Note that this issue existed before your PR, because that router-wide channel exclusion doesn't take the amount into account. But this PR makes it worse, because now it penalizes all channels, and I'm afraid it will have a visible negative impact.

I think the culprit is our `ExcludeChannel` mechanism. We probably shouldn't exclude so aggressively when receiving a `TemporaryChannelFailure` (but we do it for `ChannelDisabled`). Or maybe we should do it on `TemporaryChannelFailure` only when the amount is below a given threshold?

I'm curious to have your feedback on how we can improve this.

PR #1975, opened 2021-09-29
Ignore channels without capacity

thomash-acinq:
Channels with 0 capacity shouldn't exist. But if it happens somehow we should ignore them, not pretend that they are big channels.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -17,7 +17,7 @@
 package fr.acinq.eclair.router
 
 import fr.acinq.bitcoin.Crypto.PublicKey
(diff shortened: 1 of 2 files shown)


thomash-acinq:
@t-bast Do you know why this code was written? I can't find how the capacity could be 0.

t-bast:
> @t-bast Do you know why this code was written? I can't find how the capacity could be 0.

Good catch, it was necessary when the mobile wallets (eclair-mobile / phoenix) used the `eclair` codebase under the hood, because they shared the path-finding algorithm code but didn't fetch channels from the blockchain (see https://github.com/ACINQ/eclair/pull/1446). But we could have put that hack only in the `android` branch in the first place, it would have been cleaner.

We stopped merging `eclair/master` to the `android` branch used for eclair-mobile / phoenix and instead only cherry-pick important patches until wallets are migrated to [lightning-kmp](https://github.com/acinq/lightning-kmp).

So I believe we can safely remove it now.

t-bast (approved):
:100:

PR #1976, opened 2021-09-29
Store published txs in AuditDb

t-bast:
We previously computed the on-chain fees paid by us after the fact, when receiving a notification that a transaction was confirmed. This worked because lightning transactions had a single input, which we stored in our channel data to allow us to compute the fee.

With anchor outputs, this mechanism doesn't work anymore. Some txs have their fees paid by a child tx, and may have more than one input.

We completely change our model to store every transaction we publish, along with the fee we're paying for this transaction. We then separately store every transaction that confirms, which lets us join these two data sets to compute how much on-chain fees we paid.

This has the added benefit that we can now audit every transaction that we tried to publish, which lets node operators audit the anchor outputs internal RBF mechanism and all the on-chain footprint of a given channel.

### DIFF
--- a/docs/release-notes/eclair-vnext.md
+++ b/docs/release-notes/eclair-vnext.md
@@ -118,8 +118,8 @@ See the [spec discussions](https://github.com/lightningnetwork/lightning-rfc/pul
 
 ### Audit trail for published transactions
 
-Eclair now records every transaction it publishes in the `audit` database, in a new `transaction-published` table.
-It also stores confirmed transactions that have an impact on existing channels (including transactions made by your peer) in a new `transaction-confirmed` table.
+Eclair now records every transaction it publishes in the `audit` database, in a new `transactions_published` table.
+It also stores confirmed transactions that have an impact on existing channels (including transactions made by your peer) in a new `transactions_confirmed` table.
 
 This lets you audit the complete on-chain footprint of your channels and the on-chain fees paid.
 This information is exposed through the `networkfees` API (which was already available in previous versions).
--- a/docs/release-notes/eclair-vnext.md
+++ b/docs/release-notes/eclair-vnext.md
@@ -116,6 +116,16 @@ We still support receiving non-segwit remote scripts, but will force-close if th
 
 See the [spec discussions](https://github.com/lightningnetwork/lightning-rfc/pull/894) for more details.
 
+### Audit trail for published transactions
+
+Eclair now records every transaction it publishes in the `audit` database, in a new `transaction-published` table.
+It also stores confirmed transactions that have an impact on existing channels (including transactions made by your peer) in a new `transaction-confirmed` table.
+
+This lets you audit the complete on-chain footprint of your channels and the on-chain fees paid.
+This information is exposed through the `networkfees` API (which was already available in previous versions).
+
+We removed the previous `network_fees` table which achieved the same result but contained less details.
+
 ### Sample GUI removed
 
 We previously included code for a sample GUI: `eclair-node-gui`.
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -248,6 +248,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
           log.info("we have nothing at stake, going straight to CLOSED")
           goto(CLOSED) using closing
         case closing: DATA_CLOSING =>
+          val isFunder = closing.commitments.localParams.isFunder
           // we don't put back the WatchSpent if the commitment tx has already been published and the spending tx already reached mindepth
           val closingType_opt = Closing.isClosingTypeAlreadyKnown(closing)
           log.info(s"channel is closing (closingType=${closingType_opt.map(c => EventType.Closed(c).label).getOrElse("UnknownYet")})")
@@ -256,7 +257,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
           // - there is no need to attempt to publish transactions for other type of closes
           closingType_opt match {
             case Some(c: Closing.MutualClose) =>
-              doPublish(c.tx)
+              doPublish(c.tx, isFunder)
             case Some(c: Closing.LocalClose) =>
               doPublish(c.localCommitPublished, closing.commitments)
             case Some(c: Closing.RemoteClose) =>
@@ -268,7 +269,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
             case None =>
               // in all other cases we need to be ready for any type of closing
               watchFundingTx(data.commitments, closing.spendingTxs.map(_.txid).toSet)
-              closing.mutualClosePublished.foreach(doPublish)
+              closing.mutualClosePublished.foreach(mcp => doPublish(mcp, isFunder))
               closing.localCommitPublished.foreach(lcp => doPublish(lcp, closing.commitments))
               closing.remoteCommitPublished.foreach(doPublish)
               closing.nextRemoteCommitPublished.foreach(doPublish)
@@ -560,8 +561,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
           def publishFundingTx(): Unit = {
             wallet.commit(fundingTx).onComplete {
               case Success(true) =>
-                // NB: funding tx isn't confirmed at this point, so technically we didn't really pay the network fee yet, so this is a (fair) approximation
-                feePaid(fundingTxFee, fundingTx, "funding", commitments.channelId)
+                context.system.eventStream.publish(TransactionPublished(commitments.channelId, remoteNodeId, fundingTx, fundingTxFee, "funding"))
                 channelOpenReplyToUser(Right(ChannelOpenResponse.ChannelOpened(channelId)))
               case Success(false) =>
                 channelOpenReplyToUser(Left(LocalError(new RuntimeException("couldn't publish funding tx"))))
@@ -610,6 +610,8 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
         case Success(_) =>
           log.info(s"channelId=${commitments.channelId} was confirmed at blockHeight=$blockHeight txIndex=$txIndex")
           blockchain ! WatchFundingLost(self, commitments.commitInput.outPoint.txid, nodeParams.minDepthBlocks)
+          if (!d.commitments.localParams.isFunder) context.system.eventStream.publish(TransactionPublished(commitments.channelId, remoteNodeId, fundingTx, 0 sat, "funding"))
+          context.system.eventStream.publish(TransactionConfirmed(commitments.channelId, remoteNodeId, fundingTx))
           val channelKeyPath = keyManager.keyPath(d.commitments.localParams, commitments.channelConfig)
           val nextPerCommitmentPoint = keyManager.commitmentPoint(channelKeyPath, 1)
           val fundingLocked = FundingLocked(commitments.channelId, nextPerCommitmentPoint)
@@ -1450,7 +1452,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
       // we can then use these preimages to fulfill origin htlcs
       log.info(s"processing bitcoin output spent by txid=${tx.txid} tx=$tx")
       val extracted = Closing.extractPreimages(d.commitments.localCommit, tx)
-      extracted foreach { case (htlc, preimage) =>
+      extracted.foreach { case (htlc, preimage) =>
         d.commitments.originChannels.get(htlc.id) match {
           case Some(origin) =>
             log.info(s"fulfilling htlc #${htlc.id} paymentHash=${htlc.paymentHash} origin=$origin")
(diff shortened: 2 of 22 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -100,10 +106,11 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
...
           statement.executeUpdate("CREATE TABLE audit.channel_updates (channel_id TEXT NOT NULL, node_id TEXT NOT NULL, fee_base_msat BIGINT NOT NULL, fee_proportional_millionths BIGINT NOT NULL, cltv_expiry_delta BIGINT NOT NULL, htlc_minimum_msat BIGINT NOT NULL, htlc_maximum_msat BIGINT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL)")
           statement.executeUpdate("CREATE TABLE audit.path_finding_metrics (amount_msat BIGINT NOT NULL, fees_msat BIGINT NOT NULL, status TEXT NOT NULL, duration_ms BIGINT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL, is_mpp BOOLEAN NOT NULL, experiment_name TEXT NOT NULL, recipient_node_id TEXT NOT NULL)")
+          statement.executeUpdate("CREATE TABLE audit.transaction_published (tx_id TEXT NOT NULL PRIMARY KEY, channel_id TEXT NOT NULL, node_id TEXT NOT NULL, fee_sat BIGINT NOT NULL, tx_type TEXT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL)")
+          statement.executeUpdate("CREATE TABLE audit.transaction_confirmed (tx_id TEXT NOT NULL PRIMARY KEY, channel_id TEXT NOT NULL, node_id TEXT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL)")
```
thomash-acinq, line 113: Are we going to look at this table without joining it with `transaction_published`? If not then `channel_id` and `node_id` are redundant.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -92,6 +91,13 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
         statement.executeUpdate("CREATE INDEX metrics_name_idx ON audit.path_finding_metrics(experiment_name)")
       }
 
+      def migration910(statement: Statement): Unit = {
+        statement.executeUpdate("CREATE TABLE audit.transaction_published (tx_id TEXT NOT NULL PRIMARY KEY, channel_id TEXT NOT NULL, node_id TEXT NOT NULL, fee_sat BIGINT NOT NULL, tx_type TEXT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL)")
+        statement.executeUpdate("CREATE TABLE audit.transaction_confirmed (tx_id TEXT NOT NULL PRIMARY KEY, channel_id TEXT NOT NULL, node_id TEXT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL)")
+        statement.executeUpdate("CREATE INDEX transaction_published_timestamp_idx ON audit.transaction_published(timestamp)")
+        statement.executeUpdate("CREATE INDEX transaction_confirmed_timestamp_idx ON audit.transaction_confirmed(timestamp)")
+      }
```
thomash-acinq, line 99: Having the old table and the new ones together makes it a lot more complicated to query historical data.
Instead you should migrate the data to the new tables, you just need to copy `network_fees` to both `transaction_published` and `transaction_confirmed`.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
t-bast: We also store transactions here that we didn't publish ourselves but are important for the channel state (for example a transaction where they spend an htlc output). So we should consider them separate, and not always overlapping.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
t-bast: That's a good point, I didn't realize that the migration was in fact quite easy. I'll work on that!

t-bast: Done in https://github.com/ACINQ/eclair/pull/1976/commits/efba393128d9dd28c36440cc6f6036c8967464cd

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -92,6 +91,17 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
...
+        statement.executeUpdate("CREATE TABLE audit.transaction_confirmed (tx_id TEXT NOT NULL PRIMARY KEY, channel_id TEXT NOT NULL, node_id TEXT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL)")
+        statement.executeUpdate("CREATE INDEX transaction_published_timestamp_idx ON audit.transaction_published(timestamp)")
+        statement.executeUpdate("CREATE INDEX transaction_confirmed_timestamp_idx ON audit.transaction_confirmed(timestamp)")
+        // Migrate data from the network_fees table (which only stored data about confirmed transactions).
+        statement.executeUpdate("INSERT INTO audit.transaction_published (tx_id, channel_id, node_id, fee_sat, tx_type, timestamp) SELECT tx_id, channel_id, node_id, fee_sat, tx_type, timestamp FROM audit.network_fees ON CONFLICT DO NOTHING")
+        statement.executeUpdate("INSERT INTO audit.transaction_confirmed (tx_id, channel_id, node_id, timestamp) SELECT tx_id, channel_id, node_id, timestamp FROM audit.network_fees ON CONFLICT DO NOTHING")
```
pm47, line 101: We do have a handful of duplicates here on our mainnet node. Seems harmless.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -610,6 +610,8 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
         case Success(_) =>
           log.info(s"channelId=${commitments.channelId} was confirmed at blockHeight=$blockHeight txIndex=$txIndex")
           blockchain ! WatchFundingLost(self, commitments.commitInput.outPoint.txid, nodeParams.minDepthBlocks)
+          if (!d.commitments.localParams.isFunder) context.system.eventStream.publish(TransactionPublished(commitments.channelId, remoteNodeId, fundingTx, 0 sat, "funding"))
+          context.system.eventStream.publish(TransactionConfirmed(commitments.channelId, remoteNodeId, fundingTx))
```
pm47, line 614: `0 sat`: I see what you did there!

So we are also tracking txes that we didn't publish ourselves. It's a change of semantics, but probably okay, and I see how it can be useful.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -565,6 +565,13 @@ object Helpers {
       }
     }
 
+    /** Compute the fee paid by a commitment transaction. */
+    def commitTxFee(commitInput: InputInfo, commitTx: Transaction, isFunder: Boolean): Satoshi = {
```
pm47, line 569: So this small method allowed you to remove the huge `def networkFeePaid()` one? Where is the trade-off? Is it only that we store published *and* confirmed transactions, instead of just confirmed? Seems too good to be true 😅 . And on top of that we are more precise IIUC. And we remove the duplicate tx desc hardcoded in `networkFeePaid`. Gosh.

pm47 (commented):
image

docs/release-notes/eclair-vnext.md
```diff
@@ -116,6 +116,16 @@ We still support receiving non-segwit remote scripts, but will force-close if th
 
 See the [spec discussions](https://github.com/lightningnetwork/lightning-rfc/pull/894) for more details.
 
+### Audit trail for published transactions
+
+Eclair now records every transaction it publishes in the `audit` database, in a new `transaction-published` table.
+It also stores confirmed transactions that have an impact on existing channels (including transactions made by your peer) in a new `transaction-confirmed` table.
```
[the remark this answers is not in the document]
t-bast: :eagle: :eyes: done in https://github.com/ACINQ/eclair/pull/1976/commits/1ac5cc01aefb1d673b5ab9b7d1aedebd7107680d

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
```diff
@@ -60,9 +60,12 @@ case class ChannelSignatureReceived(channel: ActorRef, commitments: Commitments)
 
 case class ChannelErrorOccurred(channel: ActorRef, channelId: ByteVector32, remoteNodeId: PublicKey, data: ChannelData, error: ChannelOpenError, isFatal: Boolean) extends ChannelEvent
 
-case class NetworkFeePaid(channel: ActorRef, remoteNodeId: PublicKey, channelId: ByteVector32, tx: Transaction, fee: Satoshi, txType: String) extends ChannelEvent
+// NB: the fee should be set to 0 when we're not paying it.
+case class TransactionPublished(channelId: ByteVector32, remoteNodeId: PublicKey, tx: Transaction, fee: Satoshi, desc: String) extends ChannelEvent
 
-// NB: this event is only sent when the channel is available
+case class TransactionConfirmed(channelId: ByteVector32, remoteNodeId: PublicKey, tx: Transaction) extends ChannelEvent
```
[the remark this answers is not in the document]
t-bast: I'm not sure, `TransactionPublished` could go in `channel.publish` but not `TransactionConfirmed` since it's sent also for transactions we didn't publish, and I'd like to keep both of them in the same file...we can consider that all of these are channel events, in the sense that they are events related to the channel's lifecycle?

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -92,6 +91,17 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
         statement.executeUpdate("CREATE INDEX metrics_name_idx ON audit.path_finding_metrics(experiment_name)")
       }
 
+      def migration910(statement: Statement): Unit = {
+        statement.executeUpdate("CREATE TABLE audit.transaction_published (tx_id TEXT NOT NULL PRIMARY KEY, channel_id TEXT NOT NULL, node_id TEXT NOT NULL, fee_sat BIGINT NOT NULL, tx_type TEXT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL)")
+        statement.executeUpdate("CREATE TABLE audit.transaction_confirmed (tx_id TEXT NOT NULL PRIMARY KEY, channel_id TEXT NOT NULL, node_id TEXT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL)")
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/1976/commits/1ac5cc01aefb1d673b5ab9b7d1aedebd7107680d
I agree that `network_fees` was confusing, it's good that it's gone!
I think it makes sense to keep it very declarative like it is, it can then be used to extract the fees but also to just check on transactions published/confirmed that the channel should be concerned with.

> should we go with mining_fees

Do you mean for the field name inside the table (currently `fee`)?
Or the table name?
I think the table name is better as `transactions_published`, but I could rename the field to `miner_fee`.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
t-bast: Yes, I think it's harmless and hard to avoid with our model, so the primary key is a good way to avoid storing duplicates for no reason.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Exactly, the idea is to track transactions that are important for the channel logic, whether we published them or not.
So we're storing their 2nd-stage txs as well once they confirm, which can make it easy to see which outputs they got and which we got.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Yes, the trade-off is that we're storing published transactions that may never confirm, but I think it's actually a good auditing feature so I believe it's a net win!

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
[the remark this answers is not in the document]
pm47: > Do you mean for the field name inside the table (currently fee)?

I mean the field. Did a quick poll here, people prefer `mining_fee`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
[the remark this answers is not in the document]
pm47: Fair enough!

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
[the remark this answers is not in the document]
t-bast: Totally agree, it's more explicit. Done in https://github.com/ACINQ/eclair/pull/1976/commits/1bd8031a9c6bb2d47867c7ed7b175d97e6465763

pm47 (approved):

PR #1977, opened 2021-09-30
Log payment failure summary (alternative)

pm47:
This build upon #1966 and is best reviewed commit-by-commit.

Changes in json serializers and timestamp formats were rabbit-holey, so I only did the minimum as a proof of concept that should be generalized in a separate PR.

NB: the new timestamp serialization breaks API backwards compatibility, requires API doc update and an addition to the release note.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Timestamp.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Timestamp.scala
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2019 ACINQ SAS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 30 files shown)


t-bast:
It feels like this is mostly orthogonal to #1966, not a replacement.
You're making changes to the json serializers that are unrelated to #1966, and then using those changes instead of a "hand-written" serializer.

So it feels like these are two separate PRs, aren't they? That can even be applied in any order.

BTW really good idea to type the timestamps, so that we finally don't have any ambiguity between milliseconds and seconds :)

pm47:
> So it feels like these are two separate PRs, aren't they? That can even be applied in any order.

I think you could cherry-pick the first three commits though, because they really are an implementation alternative to #1966, and let's leave out the timestamp types for another PR?

PR #1978, opened 2021-09-30
Fix race condition in 'stream updates to front' test

pm47:
As usual, the race condition is due to a delay before subscribing to the event stream.


t-bast (approved):
:rocket:

PR #1979, opened 2021-09-30
Json serializers refactoring

pm47:
Best reviewed commit-by-commit.


eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
```diff
@@ -230,48 +250,43 @@ class RouteResponseSerializer extends CustomSerializerOnly[RouteResponse](_ => {
...
-    DefaultFormats +
-      new ByteVector32Serializer +
-      new ByteVectorSerializer +
-      new PublicKeySerializer +
-      new MilliSatoshiSerializer +
-      new CltvExpirySerializer))))
-})
+// @formatter:off
+private case class DirectedHtlcJson(direction: String, add: UpdateAddHtlc)
+object DirectedHtlcSerializer extends ConvertClassSerializer[DirectedHtlc](h => DirectedHtlcJson(direction = h.direction, add = h.add))
```
pm47, line 272: Note here how we don't import the lower level serializers. We only need to take care of the actual `DirectedHtlcJson` class.

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
```diff
@@ -203,9 +230,12 @@ object OutPointKeySerializer extends MinimalKeySerializer({
...
 
-object InputInfoSerializer extends MinimalSerializer({
-  case x: InputInfo => JObject(("outPoint", JString(s"${x.outPoint.txid}:${x.outPoint.index}")), ("amountSatoshis", JInt(x.txOut.amount.toLong)))
-})
+// @formatter:off
+private case class InputInfoJson(outPoint: String, amountSatoshis: Long)
+object InputInfoSerializer extends ConvertClassSerializer[InputInfo](i =>
+  InputInfoJson(outPoint = s"${i.outPoint.txid}:${i.outPoint.index}", amountSatoshis = i.txOut.amount.toLong)
+)
+// @formatter:on
```
t-bast, line 238: Since we already have a serializer for `OutPoint` and one for `MilliSatoshi`, it feels like we should be able to do:

```scala
private case class InputInfoJson(outPoint: OutPoint, amountSatoshis: Satoshi)
object InputInfoSerializer extends ConvertClassSerializer[InputInfo](i => InputInfoJson(i.outPoint, i.txOut.amount))
```

Can't we?

t-bast (commented):
ACK, I like it! It's clearly easier to read and feels more natural.

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
pm47: Absolutely! And this is indeed a perfect show case of this new technique. Done in 62c9e4151.

t-bast (approved):

t-bast (approved):

PR #1980, opened 2021-10-01
Update anchor outputs feerate tolerance

t-bast:
Allow any feerate when we're fundee and are using anchor outputs: this will completely prevent unwanted channel closure.

However this can be unsafe in a high fee environment if the commit tx is below the propagation threshold.
It could be argued that when we discover it it's too late anyway, so our only option is to wait for package relay to save the day.
But I feel a bit uneasy with a completely unbounded restriction on the lower bound...
Shall we keep the check on the lower bound until package relay lands or is it useless?

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -124,6 +124,7 @@ class Setup(val datadir: File,
     // @formatter:off
     override def getFeeratePerKb(target: Int): FeeratePerKB = feeratesPerKB.get().feePerBlock(target)
     override def getFeeratePerKw(target: Int): FeeratePerKw = feeratesPerKw.get().feePerBlock(target)
+    override def mempoolMinFee(): FeeratePerKw = feeratesPerKw.get().mempoolMinFee
     // @formatter:on
   }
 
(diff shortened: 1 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -124,6 +124,7 @@ class Setup(val datadir: File,
     // @formatter:off
     override def getFeeratePerKb(target: Int): FeeratePerKB = feeratesPerKB.get().feePerBlock(target)
     override def getFeeratePerKw(target: Int): FeeratePerKw = feeratesPerKw.get().feePerBlock(target)
+    override def mempoolMinFee(): FeeratePerKw = feeratesPerKw.get().mempoolMinFee
```
pm47, line 127: For consistency? Not pretty I know:

```suggestion
    override def getMempoolMinFeeratePerKw(): FeeratePerKw = feeratesPerKw.get().mempoolMinFee
```

t-bast: Consistency is good, done in https://github.com/ACINQ/eclair/pull/1980/commits/9574bb183cab6dcc7f4ba2e458286f7d670a50db

pm47 (approved):

PR #1981, opened 2021-10-01
Remove the GUI

t-bast:
So long old friend, and thanks for all the demos!


dpad85 (approved):
gif4452372289617956630

PR #1982, opened 2021-10-01
Add log file for important notifications

t-bast:
Add a new log file for important notifications that require an action from the node operator.

Using a separate log file makes it easier than grepping specific messages from the standard logs, and lets us use a different style of messaging, where we provide more information about what steps to take to resolve the issue.


btcontract:
Can this be made catchable by [AlarmBot plugin](https://github.com/engenegr/eclair-alarmbot-plugin) (maybe publish some kind of event to stream)? This use case fits it like a glove.

t-bast:
> Can this be made catchable by AlarmBot plugin (maybe publish some kind of event to stream)? This use case fits it like a glove.

Yes definitely. I planned on creating an event, but wasn't sure how it would be used.
Is a `case class NodeOperatorNotification(message: String)` enough for your alarm bot?

btcontract:
> Is a case class NodeOperatorNotification(message: String) enough for your alarm bot?

Yes.

PR #1983, opened 2021-10-03
Move path-finding examples to documentation

pm47:
Having basic documentation in-place by providing examples in
`eclair.conf` is great and very convenient, but in the case of
path-finding, defining experiments take so much space that it makes
the whole configuration file actually more complicated to understand.
And since we don't want to enable experiments by default, the user still
has to figure out what to change to actually enable AB-testing.

I believe a better trade-off is to move examples to the docs.

### DIFF
--- a/docs/Configure.md
+++ b/docs/Configure.md
@@ -3,35 +3,46 @@
 ---
 
 * [Configuration file](#configuration-file)
-* [Change your node's configuration](#change-your-nodes-configuration)
+  * [Changing the data directory](#changing-the-data-directory)
+  * [Splitting the configuration](#splitting-the-configuration)
 * [Options reference](#options-reference)
 * [Customize features](#customize-features)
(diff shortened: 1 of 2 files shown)


docs/Configure.md
```diff
@@ -1,21 +1,49 @@
...
+
+You can change the data directory with the `eclair.datadir` parameter:
+```sh
+eclair-node.sh -Declair.datadir="/path/to/custom/eclair/data/folder"
+```
+
+### Splitting the configuration
+
+Note that HOCON allows you to have files include other files. This allows you to split your configuration file into
+several logical files, for easier management. For example, you could defined a file `routing.conf` file with parameters
```
t-bast, line 44: nit:
```suggestion
several logical files, for easier management. For example, you could define a file `routing.conf` file with parameters
```

t-bast (approved):

PR #1984, opened 2021-10-03
`deleteinvoice` method

rorp:
This is needed for rebalancing too.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgPaymentsDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgPaymentsDb.scala
@@ -30,7 +30,7 @@ import scodec.Attempt
 import scodec.bits.BitVector
 import scodec.codecs._
 
-import java.sql.{ResultSet, Statement, Timestamp}
+import java.sql.{Connection, ResultSet, Statement, Timestamp}
 import java.time.Instant
 import java.util.UUID
 import javax.sql.DataSource
@@ -281,10 +281,7 @@ class PgPaymentsDb(implicit ds: DataSource, lock: PgLock) extends PaymentsDb wit
 
   override def getIncomingPayment(paymentHash: ByteVector32): Option[IncomingPayment] = withMetrics("payments/get-incoming", DbBackends.Postgres) {
     withLock { pg =>
-      using(pg.prepareStatement("SELECT * FROM payments.received WHERE payment_hash = ?")) { statement =>
-        statement.setString(1, paymentHash.toHex)
-        statement.executeQuery().map(parseIncomingPayment).headOption
-      }
+      getIncomingPaymentInternal(pg, paymentHash)
     }
   }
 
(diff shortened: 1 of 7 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgPaymentsDb.scala
```diff
@@ -401,5 +398,32 @@ class PgPaymentsDb(implicit ds: DataSource, lock: PgLock) extends PaymentsDb wit
...
+            case _: IncomingPaymentStatus.Received =>
+              throw new IllegalArgumentException("Cannot remove a received incoming payment")
+            case _: IncomingPaymentStatus =>
+              using(pg.prepareStatement("DELETE FROM payments.received WHERE payment_hash = ?")) { delete =>
+                delete.setString(1, paymentHash.toHex)
+                delete.executeUpdate()
+              }
+          }
+        case None =>
+          throw new IllegalArgumentException("Unknown incoming payment")
```
t-bast, line 415: I'm not sure we should throw here, I'd rather have this method be idempotent, it's much better.
If the payment doesn't exist in the DB, I would just return (it's conceptually a success, the payment isn't in the DB anymore).

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgPaymentsDb.scala
```diff
@@ -401,5 +398,32 @@ class PgPaymentsDb(implicit ds: DataSource, lock: PgLock) extends PaymentsDb wit
     }
   }
 
+  override def removeIncomingPayment(paymentHash: ByteVector32): Unit = withMetrics("payments/remove-incoming", DbBackends.Postgres) {
+    withLock { pg =>
+      getIncomingPaymentInternal(pg, paymentHash) match {
+        case Some(incomingPayment) =>
+          incomingPayment.status match {
+            case _: IncomingPaymentStatus.Received =>
+              throw new IllegalArgumentException("Cannot remove a received incoming payment")
```
t-bast, line 407: Please don't throw. Scala is full of better ways of signaling non-exceptional errors (this case is really not an exceptional error - the only places in this file where we throw correspond to what would be an internal bug).

Just change the signature of `removeIncomingPayment` to return a `Try[Unit]` instead.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqlitePaymentsDb.scala
```diff
@@ -390,6 +390,23 @@ class SqlitePaymentsDb(sqlite: Connection) extends PaymentsDb with Logging {
     }
   }
 
+  override def removeIncomingPayment(paymentHash: ByteVector32): Unit = withMetrics("payments/remove-incoming", DbBackends.Sqlite) {
```
t-bast, line 393: Same comments as the postgres version.

t-bast (commented):
Thanks for doing this!
Can you add this to the release notes?

PR #1986, opened 2021-10-04
Cookie auth

### LINKED ISSUE #1436: Feature request: Support cookie authentication for bitcoind RPC
It'd be great to be able to use cookie authentication for authenticating with bitcoind RPC. It's easier to work with in automated systems and [more secure](https://github.com/Kixunil/security_writings/blob/master/cookie_files.md).

Thoragh:
fixes #1436 

This pull request add support for bitcoind RPC cookie authentication. This is my first time writing scala code so it might not be the best solution.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -21,6 +21,8 @@ eclair {
   bitcoind {
     host = "localhost"
     rpcport = 8332
+    auth = "password" // "password" for user/password authentication, "safecookie" for cookie authentication
+    cookie = ${user.home}"/.bitcoin/.cookie"
     rpcuser = "foo"
     rpcpassword = "bar"
     // Name of the bitcoind wallet that should be used to fund channels.
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -51,6 +51,8 @@ import scodec.bits.ByteVector
 
 import java.io.File
 import java.net.InetSocketAddress
+import java.nio.charset.StandardCharsets
+import java.nio.file.{Files, Path}
 import java.util.UUID
 import java.util.concurrent.TimeUnit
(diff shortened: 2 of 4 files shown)


t-bast:
Thanks for the PR! Can you please add to the PR description:

- details on how to test this end-to-end
- links to the relevant `bitcoind` documentation on cookie auth
- a small summary of advantages/drawbacks compared to password auth

This will also need an update to the release notes. Can add a line to the following section saying that eclair now supports cookie auth with bitcoind?

https://github.com/ACINQ/eclair/blob/master/docs/release-notes/eclair-vnext.md#miscellaneous-improvements-and-bug-fixes

Thoragh:
Edited the first comment. Will add changes to the docs later today.

t-bast:
Thanks! We're currently finalizing a release, so we probably won't have time to review this PR right now, but don't worry we will get back to it as soon as we can.

t-bast:
I haven't had time yet to look at the actual code (but I read the links you shared, thanks!), but on the feature itself, here are the requirements based on what you shared:

- we need to be able to restart eclair or bitcoind independently
- we don't want to do a file read every time we make a request to bitcoind, that would hurt performance too much for no good reason

It seems to me that this PR should thus:

- read the cookie file at startup
- send requests to bitcoind as usual (without re-reading the cookie file each time)
- when a request to bitcoind fails, re-read the cookie file and retry the request with the updated cookie if it has changed (if the cookie hasn't changed, there's another issue so we can let the error bubble up)

That last point isn't trivial, but I think it's useful.
We may want to add a delay before retrying the request after a cookie change, to make sure bitcoind has correctly started.

Thoragh:
I've updated the PR to only read the cookie at startup and after failed authentication (when the RPC response is "Status code 401"). 

This PR replaces the user and password parameters for the `BasicBitcoinJsonRPCClient` class with an RPCAuthMethod parameter. RPCAuthMethod has two subclasses: `RPCSafeCookie(path: Path)` and `RPCPassword(user: String, password: String)` depending on which authentication method it should use.

PR #1989, opened 2021-10-05
Fix ZmqWatcher block timeout

t-bast:
Unfortunately, `context.log` is *not* thread safe and shouldn't be used in future continuation. We should instead use `pipeToSelf` when we want to act on the results of a `Future`.


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -247,27 +250,41 @@ private class ZmqWatcher(nodeParams: NodeParams, blockCount: AtomicLong, client:
...
         Behaviors.same
 
       case TickNewBlock =>
-        client.getBlockCount.map { count =>
-          log.debug("setting blockCount={}", count)
-          blockCount.set(count)
-          context.system.eventStream ! EventStream.Publish(CurrentBlockCount(count))
+        context.pipeToSelf(client.getBlockCount) {
+          case Failure(t) => GetBlockCountFailed(t)
+          case Success(count) => PublishBlockCount(count)
```
pm47, line 274: Nit: this partial function could be factorized.

pm47 (approved):

pm47 (approved):

PR #1994, opened 2021-10-08
Check serialization consistency in all channel tests

sstone:
We add a simple wrapper over the channels db used in all channel state unit tests, which will basically check
that deserialize(serialize(state)) == state.


sstone:
Yes, tests will fail if channel states cannot be read back consistently but these failures are a bit cryptic (less if you enable console logging with `-Dlogback.configurationFile`), I could not find a simple way to improve this

PR #2021, opened 2021-10-21
Offers

thomash-acinq:
Implements offers (https://github.com/lightningnetwork/lightning-rfc/pull/798)


thomash-acinq:
This code should work, however it currently has two big limitations:
- eclair does not yet support paying to blinded routes. This means that the offer can use blinded routes but the invoice must be to a non blinded node id or we won't be able to pay it.
- the spec does not use a payment_secret (possibly assuming that all payments will use blinded routes) but eclair requires one. I have added a payment secret to the invoice (TLV 58) which makes it incompatible with c-lightning.

PR #2027, opened 2021-10-22
Add nightly build with bitcoind master

t-bast:
We should regularly run against bitcoind master, to detect if a change to bitcoind is affecting us.
This will let us detect changes that may break lightning before bitcoind releases them, giving us a chance to rectify it.

This isn't a completely unrealistic scenario since we depend on some non-standard "hacks" in our transactions formats.
For example, a change to CSV behavior was recently proposed: https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-September/019400.html
If merged into bitcoin, lightning transactions wouldn't have been relayed by upgraded bitcoin nodes, so it would have been very important to catch it *before* a bitcoind release.

### DIFF
--- a/.github/workflows/latest-bitcoind.yml
+++ b/.github/workflows/latest-bitcoind.yml
@@ -26,7 +26,7 @@ jobs:
         working-directory: ./bitcoin
 
       - name: Configure bitcoind
-        run: ./configure
+        run: ./configure --with-zmq --without-gui --disable-shared --with-pic --disable-tests --disable-bench
         working-directory: ./bitcoin
 
       - name: Build bitcoind
--- a/.github/workflows/latest-bitcoind.yml
+++ b/.github/workflows/latest-bitcoind.yml
@@ -1,0 +1,56 @@
+name: Latest Bitcoin Core
+
+on:
+  # TODO: remove, keep only the bi-weekly trigger
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    # Run at midnight on Sunday and Wednesday.
+    - cron: '0 0 * * 0,3'
+
+jobs:
+
+  regression-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    steps:
+      - name: Checkout bitcoind master
+        uses: actions/checkout@v2
+        with:
+          repository: bitcoin/bitcoin
+          path: bitcoin
+
+      - name: Install bitcoind dependencies
+        run: sudo apt-get install build-essential libtool autotools-dev automake pkg-config bsdmainutils python3 libssl-dev libevent-dev libboost-system-dev libboost-filesystem-dev libboost-chrono-dev libboost-test-dev libboost-thread-dev libminiupnpc-dev libzmq3-dev libqt5gui5 libqt5core5a libqt5dbus5 qttools5-dev qttools5-dev-tools libprotobuf-dev protobuf-compiler git libsqlite3-dev ccache
+        working-directory: ./bitcoin
+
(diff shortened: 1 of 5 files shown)


.github/workflows/latest-bitcoind.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install bitcoind dependencies
+        run: sudo apt-get install build-essential libtool autotools-dev automake pkg-config bsdmainutils python3 libssl-dev libevent-dev libboost-system-dev libboost-filesystem-dev libboost-chrono-dev libboost-test-dev libboost-thread-dev libminiupnpc-dev libzmq3-dev libqt5gui5 libqt5core5a libqt5dbus5 qttools5-dev qttools5-dev-tools libprotobuf-dev protobuf-compiler git libsqlite3-dev ccache
+        working-directory: ./bitcoin
+
+      - name: Autogen bitcoind
+        run: ./autogen.sh
+        working-directory: ./bitcoin
+
+      - name: Configure bitcoind
+        run: ./configure
```
benthecarman, line 29: some options that should make this faster

`--with-zmq --without-gui --disable-shared --with-pic --disable-tests --disable-bench`

.github/workflows/latest-bitcoind.yml
```diff
@@ -0,0 +1,51 @@
...
+        working-directory: ./bitcoin
+
+      - name: Build bitcoind
+        run: make -j "$(($(nproc)))"
```
benthecarman, line 33: can do `"$(($(nproc)+1))"`

.github/workflows/latest-bitcoind.yml
t-bast: Thanks, good idea!

.github/workflows/latest-bitcoind.yml
t-bast: Can you explain? That would give it our number of logical CPUs + 1, right? Why does it make sense to add one?

benthecarman: tbh idk why but every compile bitcoin core from source tutorial I've seen uses +1

t-bast: I'd rather not include something I don't understand :wink: 
If you figure out why those tutorials do that, I'm interested!

benthecarman: Seems it has to do with I/O delays and other resource constraints

https://unix.stackexchange.com/a/483306/330110

t-bast: Thanks for finding this!

> You may have some anecdotal experience where doing core-count + 1 results in faster compile times.

This doesn't look like a clear-cut win for doing +1 and feels highly dependent on the actual platform.
I'd rather stick to `nproc` which is intuitively the best choice, as +1 can in theory also degrade performance, and I don't want to have to measure the difference regularly to choose the best one only to win a few minutes of CI time (I would do it if it was running on every PR or commit to master, but this is going to be a nightly build that runs a few times a week).

.github/workflows/latest-bitcoind.yml
t-bast: Thanks for the tip, done in https://github.com/ACINQ/eclair/pull/2027/commits/21efe18b3202dfd9e807a6b78f02baa6b7bfc770

sstone (approved):
Nice and compact! It's too bad that the build is taking so long but it's just a few times a week. I wonder if there is a way to speed things by using a script that starts a docker image instead of a locally compiled bitcoind but it could be done later.

PR #2038, opened 2021-10-28
Use bitcoin-kmp through bitcoin-lib

sstone:
We use a version of bitcoin-lib that uses bitcoin-kmp under the hood (see https://github.com/ACINQ/bitcoin-lib/pull/63)
Its API is the same as before, but its package is now fr.acinq.bitcoinscala.

This is an alternative to #1949 that is much less invasive and will allow us to mix `eclair` and `lightning-kmp`.

There is a significant breaking change in bitcoin-lib: we only support OSes that are compatible with secp256k1-kmp (i.e we don't fallback to bouncycastle anymore if it cannot be loaded). In practice it means that `eclair` won't run on some 32 bits OSes and possibly some linux distros.

The version of bitcoin-lib that is used here is 0.20-KMP-SNAPSHOT, which is published to https://oss.sonatype.org/content/repositories/snapshots/. This is why this is a draft PR. If it is accepted I'll publish a new bitcoin-lib release to maven central and update this PR.

EDIT: this PR is now based on bitcoin-lib 0.22 which is available on maven central, supports Linux Arm64 (through bitcoin-kmp 0.8.3) and can be merged as-is.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
@@ -16,12 +16,12 @@
 
 package fr.acinq.eclair.blockchain.bitcoind.rpc
 
-import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.eclair.KamonExt
 import fr.acinq.eclair.blockchain.Monitoring.{Metrics, Tags}
-import org.json4s.JsonAST.{JString, JValue}
+import fr.acinq.eclair.json.{ByteVector32KmpSerializer, ByteVector32Serializer}
+import org.json4s.DefaultFormats
+import org.json4s.JsonAST.JValue
 import org.json4s.jackson.Serialization
-import org.json4s.{CustomSerializer, DefaultFormats}
 import sttp.client3._
 import sttp.client3.json4s._
 import sttp.model.StatusCode
@@ -34,14 +34,8 @@ import scala.util.{Failure, Success, Try}
 
 class BasicBitcoinJsonRPCClient(rpcAuthMethod: BitcoinJsonRPCAuthMethod, host: String = "127.0.0.1", port: Int = 8332, ssl: Boolean = false, wallet: Option[String] = None)(implicit sb: SttpBackend[Future, _]) extends BitcoinJsonRPCClient {
 
-  // necessary to properly serialize ByteVector32 into String readable by bitcoind
-  object ByteVector32Serializer extends CustomSerializer[ByteVector32](_ => ( {
-    null
-  }, {
-    case x: ByteVector32 => JString(x.toHex)
-  }))
+  implicit val formats = DefaultFormats.withBigDecimal + ByteVector32Serializer + ByteVector32KmpSerializer
 
-  implicit val formats = DefaultFormats.withBigDecimal + ByteVector32Serializer
   private val scheme = if (ssl) "https" else "http"
   private val serviceUri = wallet match {
     case Some(name) => uri"$scheme://$host:$port/wallet/$name"
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BitcoinCoreClient.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BitcoinCoreClient.scala
@@ -16,8 +16,9 @@
 
 package fr.acinq.eclair.blockchain.bitcoind.rpc
 
-import fr.acinq.bitcoin.Crypto.PublicKey
-import fr.acinq.bitcoin._
+import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
+import fr.acinq.bitcoin.{Bech32, Block}
+import fr.acinq.bitcoin.scalacompat._
 import fr.acinq.eclair.ShortChannelId.coordinates
 import fr.acinq.eclair.blockchain.OnChainWallet
 import fr.acinq.eclair.blockchain.OnChainWallet.{MakeFundingTxResponse, OnChainBalance}
@@ -32,6 +33,7 @@ import org.json4s.JsonAST._
 import scodec.bits.ByteVector
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.jdk.CollectionConverters.ListHasAsScala
 import scala.util.{Failure, Success, Try}
 
 /**
@@ -130,7 +132,11 @@ class BitcoinCoreClient(val rpcClient: BitcoinJsonRPCClient) extends OnChainWall
    * @param outputIndex   index of the transaction output that has been spent.
    * @return the transaction spending the given output.
    */
-  def lookForSpendingTx(blockhash_opt: Option[ByteVector32], txid: ByteVector32, outputIndex: Int)(implicit ec: ExecutionContext): Future[Transaction] =
+  def lookForSpendingTx(blockhash_opt: Option[ByteVector32], txid: ByteVector32, outputIndex: Int)(implicit ec: ExecutionContext): Future[Transaction] = {
+    lookForSpendingTx(blockhash_opt.map(KotlinUtils.scala2kmp), KotlinUtils.scala2kmp(txid), outputIndex)
+  }
+
+  def lookForSpendingTx(blockhash_opt: Option[fr.acinq.bitcoin.ByteVector32], txid: fr.acinq.bitcoin.ByteVector32, outputIndex: Int)(implicit ec: ExecutionContext): Future[Transaction] =
     for {
       blockhash <- blockhash_opt match {
         case Some(b) => Future.successful(b)
@@ -138,10 +144,10 @@ class BitcoinCoreClient(val rpcClient: BitcoinJsonRPCClient) extends OnChainWall
       }
       // with a verbosity of 0, getblock returns the raw serialized block
       block <- rpcClient.invoke("getblock", blockhash, 0).collect { case JString(b) => Block.read(b) }
-      prevblockhash = block.header.hashPreviousBlock.reverse
-      res <- block.tx.find(tx => tx.txIn.exists(i => i.outPoint.txid == txid && i.outPoint.index == outputIndex)) match {
+      prevblockhash = block.header.hashPreviousBlock.reversed()
+      res <- block.tx.asScala.find(tx => tx.txIn.asScala.exists(i => i.outPoint.txid == txid && i.outPoint.index == outputIndex)) match {
         case None => lookForSpendingTx(Some(prevblockhash), txid, outputIndex)
-        case Some(tx) => Future.successful(tx)
+        case Some(tx) => Future.successful(KotlinUtils.kmp2scala(tx))
       }
     } yield res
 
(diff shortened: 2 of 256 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/ExplorerApi.scala
```diff
@@ -48,12 +49,14 @@ object ExplorerApi {
...
 
+  implicit def scala2kmp(input: ByteVector32): fr.acinq.bitcoin.ByteVector32 = new fr.acinq.bitcoin.ByteVector32(input.toArray)
+
   sealed trait Explorer {
     // @formatter:off
     /** Explorer friendly-name. */
     def name: String
     /** Map from chainHash to explorer API URI. */
-    def baseUris: Map[ByteVector32, Uri]
+    def baseUris: Map[fr.acinq.bitcoin.ByteVector32, Uri]
```
t-bast, line 59: Can we use a `ByteVector32` here instead? It would be nice to keep the `fr.acinq.bitcoin.ByteVector32` hidden away as much as possible, wouldn't it?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/ExplorerApi.scala
```diff
@@ -163,7 +166,7 @@ object ExplorerApi {
             val JInt(nonce) = block \ "nonce"
             val previousBlockHash = (block \ "prev_block").extractOpt[String].map(ByteVector32.fromValidHex(_).reverse).getOrElse(ByteVector32.Zeroes)
             val merkleRoot = (block \ "mrkl_root").extractOpt[String].map(ByteVector32.fromValidHex(_).reverse).getOrElse(ByteVector32.Zeroes)
-            val header = BlockHeader(version.toLong, previousBlockHash, merkleRoot, OffsetDateTime.parse(time).toEpochSecond, bits.toLong, nonce.toLong)
+            val header = new BlockHeader(version.toLong, previousBlockHash, merkleRoot, OffsetDateTime.parse(time).toEpochSecond, bits.toLong, nonce.toLong)
```
t-bast, line 169: I think we should have a scala `BlockHeader` type in `bitcoin-lib` to hide the kotlin part.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BitcoinCoreClient.scala
```diff
@@ -130,18 +132,24 @@ class BitcoinCoreClient(val rpcClient: BitcoinJsonRPCClient) extends OnChainWall
    * @param outputIndex   index of the transaction output that has been spent.
    * @return the transaction spending the given output.
    */
-  def lookForSpendingTx(blockhash_opt: Option[ByteVector32], txid: ByteVector32, outputIndex: Int)(implicit ec: ExecutionContext): Future[Transaction] =
+  def lookForSpendingTx(blockhash_opt: Option[ByteVector32], txid: ByteVector32, outputIndex: Int)(implicit ec: ExecutionContext): Future[Transaction] = {
```
t-bast, line 135: Can't we define the right wrapper types in `bitcoin-lib` to avoid juggling with kotlin types here?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/keymanager/LocalChannelKeyManager.scala
```diff
@@ -58,6 +58,8 @@ class LocalChannelKeyManager(seed: ByteVector, chainHash: ByteVector32) extends
       override def load(keyPath: KeyPath): ExtendedPublicKey = publicKey(privateKeys.get(keyPath))
     })
 
+  private implicit def longs2keypath(input: List[Long]): KeyPath = KeyPath(input)
```
t-bast, line 61: Shouldn't this be in `bitcoin-lib`?

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Scripts.scala
```diff
@@ -94,11 +97,13 @@ object Scripts {
...
   private def sequenceToBlockHeight(sequence: Long): Long = {
-    if ((sequence & TxIn.SEQUENCE_LOCKTIME_DISABLE_FLAG) != 0) {
+    if ((sequence & SEQUENCE_LOCKTIME_DISABLE_FLAG) != 0) {
       0
     } else {
-      require((sequence & TxIn.SEQUENCE_LOCKTIME_TYPE_FLAG) == 0, "CSV timeout must use block heights, not block times")
-      sequence & TxIn.SEQUENCE_LOCKTIME_MASK
+      require((sequence & SEQUENCE_LOCKTIME_TYPE_FLAG) == 0
+        , "CSV timeout must use block heights, not block times"
+      )
```
t-bast, line 105: nit: revert?

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
```diff
@@ -42,14 +44,17 @@ private[channel] object ChannelCodecs0 {
...
+    val keyPathCodec: Codec[KeyPath] = ("path" | listOfN(uint16, uint32)).xmap[KeyPath](l => KeyPath(l), keyPath => keyPath.path.toList).as[KeyPath].decodeOnly
 
     val extendedPrivateKeyCodec: Codec[ExtendedPrivateKey] = (
       ("secretkeybytes" | bytes32) ::
         ("chaincode" | bytes32) ::
         ("depth" | uint16) ::
         ("path" | keyPathCodec) ::
-        ("parent" | int64)).as[ExtendedPrivateKey].decodeOnly
+        ("parent" | int64)).xmap(
+      { case a :: b :: c :: d :: e :: HNil => ExtendedPrivateKey(a, b, c, d, e) },
```
t-bast, line 55: I'm not a big fan of depending on shapeless here, I'll try to find an alternative.

t-bast (commented):
Just a quick first pass, I need to play around more with the code

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/ExplorerApi.scala
sstone: There are 2 possible design choices here:

1. make the transition from bitcoin-lib as easy as possible: add many wrappers, conversion methods, ...
2. make the transition to bitcoin-kmp as easy as possible and start using bitcoin-kmp types as much as possible (and update bitcoin-kmp if needed, being easily usable from Scala and Java is one of its primary design goals)

I'm leaning towards 2, which is why there are no scala wrappers for blocks, headers, psbt, base58, bech32...
There is an exception for transactions and scripts because otherwise migration would have been too difficult and clumsy. 

But adding wrappers for blocks and headers is easy and would have no negative impacts that I can see (memory, performance...) so I'll wait a bit and see what the general feeling about this is.

t-bast: I think you're right, I'm starting to lean a bit more towards option 2, I guess I just need to spend more time playing around in the codebase and see how it impacts the dev experience.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BitcoinCoreClient.scala
sstone: it's linked to the comment above: without scala wrappers for `Block` we end up dealing with java lists, and since implicit conversions to Scala lists are deprecated we need to use `asScala` explicitly.

t-bast (commented):
I'm not seeing anything weird after playing with this version on regtest, and the minimal code changes look good to me.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/ExplorerApi.scala
```diff
@@ -48,12 +50,14 @@ object ExplorerApi {
...
 
+  import fr.acinq.bitcoinscala.KotlinUtils.scala2kmp
+
   sealed trait Explorer {
     // @formatter:off
     /** Explorer friendly-name. */
     def name: String
     /** Map from chainHash to explorer API URI. */
-    def baseUris: Map[ByteVector32, Uri]
+    def baseUris: Map[ByteVector32Kmp, Uri]
```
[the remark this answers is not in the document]
sstone: The pb is that the type of all genesis block hashes is `fr.acinq.bitcoin.ByteVector32`,  and we use them as keys to `baseUris`

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/HeadersOverDns.scala
```diff
@@ -45,7 +47,9 @@ object HeadersOverDns {
   private case class WrappedDnsFailed(cause: Throwable) extends Command
   // @formatter:on
 
-  def apply(chainHash: ByteVector32, currentBlockHeight: BlockHeight): Behavior[Command] = {
+  def apply(chainHash: ByteVector32, currentBlockHeight: BlockHeight): Behavior[Command] = apply(new bitcoin.ByteVector32(chainHash.toArray), currentBlockHeight)
+
+  def apply(chainHash: fr.acinq.bitcoin.ByteVector32, currentBlockHeight: BlockHeight): Behavior[Command] = {
```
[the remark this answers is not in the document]
sstone: I'm sorry I don't understand what you mean here ?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/ExplorerApi.scala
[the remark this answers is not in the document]
t-bast: It doesn't have to be, you can simply import `fr.acinq.bitcoinscala.Block` instead of `fr.acinq.bitcoin.Block` to get the genesis block hashes (which are then `ByteVector32` from the `bitcoinscala` package).

Just replace your import from `fr.acinq.bitcoin.Block` to `fr.acinq.bitcoinscala.Block` and you don't need to change this line (and the others mentioned in https://github.com/ACINQ/eclair/pull/2038#discussion_r803519256)

t-bast (commented):
ACK https://github.com/ACINQ/eclair/pull/2038/commits/5de05d5a76cddabe253d46537a403b76651dcacc

pm47 (changes requested):
I just put this blocker upstream: https://github.com/ACINQ/bitcoin-lib/pull/63#pullrequestreview-915456829

sstone:
We now use bitcoin-kmp 0.21-KMP, bitcoin-kmp 0.8.3 and secp256k1-kmp 0.6.3 (which includes JNI bindings for Linux Arm64).

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
```diff
@@ -41,7 +41,13 @@ class BasicBitcoinJsonRPCClient(rpcAuthMethod: BitcoinJsonRPCAuthMethod, host: S
     case x: ByteVector32 => JString(x.toHex)
   }))
 
-  implicit val formats = DefaultFormats.withBigDecimal + ByteVector32Serializer
+  object ByteVector32KmpSerializer extends CustomSerializer[fr.acinq.bitcoin.ByteVector32](_ => ( {
+    null
+  }, {
+    case x: fr.acinq.bitcoin.ByteVector32 => JString(x.toHex)
+  }))
```
t-bast, line 48: Wouldn't it make sense to move this serializer to `JsonSerializers.scala`?

t-bast (commented):
LGTM, just one nit

t-bast (approved):

pm47 (approved):

PR #2040, opened 2021-10-28
Implement cookie authentication for eclair json-rpc api

### LINKED ISSUE #1437: Feature request: implement cookie authentication for eclair RPC
Eclair could be [better secured](https://github.com/Kixunil/security_writings/blob/master/cookie_files.md) if it could generate its own cookie file and use it for authentication, just like bitcoind can.

Note that this is different from #1436, which is about Eclair -> bitcoind connection, while this one is about Lapp -> Eclair connection.

Thoragh:
fixes #1437 

This PR add cookie authentication for eclair API. It is enabled by default when the API is enabled and the default cookie file is `.cookie` in the datadir.
It is possible to have both cookie and password authentication enabled at the same time and the API can be enabled without password. I used a separate configuration option for flag because I don't like how it is done in bitcoind where cookie depends on if an rpcpassword is set.

It is also possible to to use `eclair.api.cookie-group-readable` to make the generated cookie file group readable on Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,9 +118,10 @@ name                         | description                                      
 -----------------------------|---------------------------------------------------------------------------------------|--------------
  eclair.chain                | Which blockchain to use: *regtest*, *testnet* or *mainnet*                            | mainnet
  eclair.server.port          | Lightning TCP port                                                                    | 9735
- eclair.api.enabled          | Enable/disable the API                                                                | false. By default the API is disabled. If you want to enable it, you must set a password.
+ eclair.api.enabled          | Enable/disable the API                                                                | false. By default the API is disabled.
  eclair.api.port             | API HTTP port                                                                         | 8080
- eclair.api.password         | API password (BASIC)                                                                  | "" (must be set if the API is enabled)
+ eclair.api.password         | API password (BASIC)                                                                  | "" (change to something else if you want to enable password authentication)
+ eclair.api.cookie-enabled   | API cookie authentication                                                             | true
  eclair.bitcoind.rpcuser     | Bitcoin Core RPC user                                                                 | foo
  eclair.bitcoind.rpcpassword | Bitcoin Core RPC password                                                             | bar
  eclair.bitcoind.zmqblock    | Bitcoin Core ZMQ block address                                                        | "tcp://127.0.0.1:29000"
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -13,8 +13,9 @@ eclair {
     enabled = false // disabled by default for security reasons
     binding-ip = "127.0.0.1"
     port = 8080
-    password = "" // password for basic auth, must be non empty if json-rpc api is enabled
+    password = "" // password for basic auth, must be non empty to enable password authentication
     cookie-enabled = true
+    cookie-group-readable = false
     cookie-path =   ${eclair.datadir}"/.cookie"
   }
 
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -16,7 +16,7 @@ eclair {
     password = "" // password for basic auth, must be non empty to enable password authentication
     cookie-enabled = true
     cookie-group-readable = false
-    cookie-path =   ${eclair.datadir}"/.cookie"
+    cookie-path = ${eclair.datadir}"/.cookie"
   }
 
   watch-spent-window = 1 minute // at startup watches will be put back within that window to reduce herd effect; must be > 0s
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -14,6 +14,8 @@ eclair {
     binding-ip = "127.0.0.1"
     port = 8080
     password = "" // password for basic auth, must be non empty if json-rpc api is enabled
+    cookie-enabled = true
+    cookie-path =   ${eclair.datadir}"/.cookie"
   }
 
   watch-spent-window = 1 minute // at startup watches will be put back within that window to reduce herd effect; must be > 0s
--- a/eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
+++ b/eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
@@ -134,7 +134,7 @@ object Boot extends App with Logging {
       aclView.setAcl(List(entry).asJava)
     })
 
-    Files.writeString(path, s"__COOKIE__:$hexPassword", StandardCharsets.UTF_8)
+    Files.writeString(path, s"${Service.CookieUserName}:$hexPassword", StandardCharsets.UTF_8)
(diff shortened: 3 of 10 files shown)


Thoragh:
I have tested that this pr work on Windows and Linux. It would be grate if someone could try on Mac OS since I can't.

README.md
```diff
@@ -118,9 +118,10 @@ name                         | description
...
  eclair.server.port          | Lightning TCP port                                                                    | 9735
- eclair.api.enabled          | Enable/disable the API                                                                | false. By default the API is disabled. If you want to enable it, you must set a password.
+ eclair.api.enabled          | Enable/disable the API                                                                | false. By default the API is disabled.
  eclair.api.port             | API HTTP port                                                                         | 8080
- eclair.api.password         | API password (BASIC)                                                                  | "" (must be set if the API is enabled)
+ eclair.api.password         | API password (BASIC)                                                                  | "" (change to something else if you want to enable password authentication)
+ eclair.api.cookie-enabled   | API cookie authentication                                                             | true
```
t-bast, line 124: We should be consistent with bitcoin core's naming:

```suggestion
 eclair.api.safecookie   | Enable API safe cookie authentication                                                             | true
```

eclair-core/src/main/resources/reference.conf
```diff
@@ -13,7 +13,10 @@ eclair {
     enabled = false // disabled by default for security reasons
     binding-ip = "127.0.0.1"
     port = 8080
-    password = "" // password for basic auth, must be non empty if json-rpc api is enabled
+    password = "" // password for basic auth, must be non empty to enable password authentication
+    cookie-enabled = true
```
t-bast, line 17:
```suggestion
    use-safecookie = true // NB: when using cookie authentication, password authentication will be automatically disabled
```

eclair-core/src/main/resources/reference.conf
```diff
@@ -13,7 +13,10 @@ eclair {
...
+    cookie-group-readable = false
```
t-bast, line 18: Please add a comment explaining that field.

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -68,12 +72,21 @@ object Boot extends App with Logging {
...
+        case "" => None
+        case valid => Some(valid)
+      }
+      val apiCookiePassword = if (config.getBoolean("api.cookie-enabled")) {
+        Some(generateCookie(config.getString("api.cookie-path"), config.getBoolean("api.cookie-group-readable")))
+      } else {
+        None
+      }
+      if (apiPassword.isEmpty && apiCookiePassword.isEmpty) {
+        throw new RuntimeException("json-rpc api requires that either password or cookie is enabled")
```
t-bast, line 84: Instead of an unnamed exception, you should replace the `EmptyAPIPasswordException` in `Setup.scala` with a `APICredentialsNotConfiguredException`

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -68,12 +72,21 @@ object Boot extends App with Logging {
...
       }
       val service: Service = new Service {
         override val actorSystem: ActorSystem = system
-        override val password: String = apiPassword
+        override val password: Option[String] = apiPassword
+        override val cookiePassword: Option[String] = apiCookiePassword
```
t-bast, line 89: We really shouldn't support both cookie and password auth at the same time.
The reason you would enable cookie authentication is to explicitly get rid of the user/password auth.
When cookie auth is enabled, this should automatically disable password auth.

I think we should force the user to clear its `password` field in `reference.conf`, like Bitcoin Core does, as this makes it very explicit for the user, and we should verify that they have done so.

Once that's done, I don't think we need to modify the `Service` class, there will still be a single `password` field that will be non-optional (containing either the user password or cookie password depending on which is enabled).

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -84,6 +97,46 @@ object Boot extends App with Logging {
...
+    Files.createFile(path)
+
+    val posixView = Files.getFileAttributeView(path, classOf[PosixFileAttributeView])
+    val aclView = Files.getFileAttributeView(path, classOf[AclFileAttributeView])
+    if (posixView != null) {
+      // Change permissions for the .cookie file on Linux/Mac OS
+      if (groupReadable) {
+        posixView.setPermissions(PosixFilePermissions.fromString("rw-r-----"))
+      } else {
+        posixView.setPermissions(PosixFilePermissions.fromString("rw-------"))
```
t-bast, line 117: Do you need to change it in that case? What are the default permissions if you don't do anything?

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -84,6 +97,46 @@ object Boot extends App with Logging {
...
+    } else if (aclView != null) {
+      // Change permissions for the .cookie file on Windows
+      val aclEntry = AclEntry.newBuilder()
+        .setPermissions(Set.from(AclEntryPermission.values()).asJava)
+        .setPrincipal(aclView.getOwner)
+        .setType(AclEntryType.ALLOW)
+        .build()
+      // setAcl() replaces all acl entries for the file.
+      // That means setting acl with an list containing only an entry that gives access to the owner, everyone else lose access to the file.
+      aclView.setAcl(List(aclEntry).asJava)
```
t-bast, line 128: Why is that necessary if you can't make the file group-readable on windows?
What exactly are you changing here compared to the default permissions?

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -84,6 +97,46 @@ object Boot extends App with Logging {
...
+      if (groupReadable) {
+        logger.warn("could not make the .cookie file group readable")
```
t-bast, line 130: This shouldn't be a warning, it should be an exception.
We should note in the `reference.conf` that this field cannot be set to `true` on windows.

eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala
```diff
@@ -84,6 +97,46 @@ object Boot extends App with Logging {
...
+      }
+    } else {
+      logger.warn("could not change the permissions of the .cookie file")
+    }
+
+    Files.writeString(path, s"${Service.CookieUserName}:$hexPassword", StandardCharsets.UTF_8)
```
t-bast, line 136: It feels like we should use ASCII instead of UTF8 here, we don't actually need UTF8, do we?

t-bast (commented):
Thanks for doing this!
I haven't tested it yet as I believe we should change some of the requirements first.

t-bast:
Closing this inactive PR.

PR #2046, opened 2021-11-02
Handle mutual close published from the outside

pm47:
If a _local_ mutual close transaction is published from outside of the actor state machine, the channel will fail to recognize it, and will move to the `ERR_INFORMATION_LEAK` state. We could instead log a warning and handle it gracefully, since no harm has been done.

This is different from a local force close, because we do not keep the fully-signed local commit tx anymore, so an unexpected published tx would indeed be very fishy in that case. But we do store the best fully-signed, read-to-publish mutual close tx in the channel data so we must be ready to handle the case where the operator manually publishes it for whatever reason.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1366,6 +1366,11 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
       // they can publish a closing tx with any sig we sent them, even if we are not done negotiating
       handleMutualClose(getMutualClosePublished(tx, d.closingTxProposed), Left(d))
 
+    case Event(WatchFundingSpentTriggered(tx), d: DATA_NEGOTIATING) if d.bestUnpublishedClosingTx_opt.map(_.tx).contains(tx) =>
+      log.warning(s"looks like a mutual close tx has been published from the outside of the channel: closingTxId=${tx.txid}")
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1366,6 +1366,11 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
       // they can publish a closing tx with any sig we sent them, even if we are not done negotiating
       handleMutualClose(getMutualClosePublished(tx, d.closingTxProposed), Left(d))
 
+    case Event(WatchFundingSpentTriggered(tx), d: DATA_NEGOTIATING) if d.bestUnpublishedClosingTx_opt.map(_.tx).contains(tx) =>
```
t-bast, line 1369: I think we should match on `txid`, not on the whole tx, it protects us against witness mutations.
Since this is spending a multisig utxo, they could chose to reorder the two signatures and your check on tx equality would fail.

pm47:
> I think we should match on txid, not on the whole tx, it protects us against witness mutations.

Good point, and also it wasn't consistent with the other handlers. Fixed with 0d4b436.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Good point, and also it wasn't consistent with the other handlers. Fixed with 0d4b436.

t-bast (approved):

PR #2052, opened 2021-11-04
Refactor sphinx payment packet

t-bast:
We previously created restrictions in `Sphinx.scala` to only allow using it for two types of onions: a 1300 bytes one for HTLCs and a 400 bytes one for trampoline.

This doesn't make sense anymore. The latest version of trampoline allows any onion size, and onion messages also allow any onion size. The Sphinx protocol doesn't care either about the size of the payload.

Another reason to remove it is that it wasn't working that well with pattern matching because of type erasure.

So now the caller must explicitly set the length of the payload, which is more flexible. Verifying that the correct length is used is deferred to higher level components.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
@@ -29,7 +29,6 @@ import scodec.bits.ByteVector
 import scodec.{Attempt, DecodeResult}
 
 import java.util.UUID
-import scala.reflect.ClassTag
 
 /**
  * Created by t-bast on 08/10/2019.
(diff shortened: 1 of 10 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -29,39 +29,51 @@ import scodec.bits.ByteVector
...
+          case Attempt.Successful(DecodeResult(perHopPayload: PaymentOnion.PaymentPacket, _)) => Right(DecodedOnionPacket(perHopPayload, nextPacket))
+          case Attempt.Failure(e: OnionRoutingCodecs.MissingRequiredTlv) => Left(e.failureMessage)
+          // Onion is correctly encrypted but the content of the per-hop payload couldn't be parsed.
+          // It's hard to provide tag and offset information from scodec failures, so we currently don't do it.
+          case Attempt.Failure(_) => Left(InvalidOnionPayload(UInt64(0), 0))
+        }
+      case Left(badOnion) => Left(badOnion)
+    }
+
+  private[payment] def decryptTrampolineOnion(paymentHash: ByteVector32, privateKey: PrivateKey, packet: OnionRoutingPacket)(implicit log: LoggingAdapter): Either[FailureMessage, DecodedOnionPacket[PaymentOnion.TrampolinePacket]] =
```
t-bast, line 71: I've chosen to duplicate this code to be able to get rid of the `unchecked` pattern matching, I think having the compiler verify that it's exhaustive is better than avoiding a few lines of duplication.

I was hoping I could get the best of both worlds, but haven't been able to. If you find a refactoring that achieves it, I'm interested!

thomash-acinq:
```scala
private[payment] def decryptOnion[T <: PaymentOnion.PacketType](perHopPayloadCodec : (Boolean) => Codec[T])(paymentHash: ByteVector32, privateKey: PrivateKey, packet: OnionRoutingPacket)(implicit log: LoggingAdapter): Either[FailureMessage, DecodedOnionPacket[T]] =
    Sphinx.peel(privateKey, paymentHash, packet) match {
      case Right(p@Sphinx.DecryptedPacket(payload, nextPacket, _)) =>
        perHopPayloadCodec(p.isLastPacket).decode(payload.bits) match {
          case Attempt.Successful(DecodeResult(perHopPayload, _)) => Right(DecodedOnionPacket(perHopPayload, nextPacket))
          case Attempt.Failure(e: OnionRoutingCodecs.MissingRequiredTlv) => Left(e.failureMessage)
          // Onion is correctly encrypted but the content of the per-hop payload couldn't be parsed.
          // It's hard to provide tag and offset information from scodec failures, so we currently don't do it.
          case Attempt.Failure(_) => Left(InvalidOnionPayload(UInt64(0), 0))
        }
      case Left(badOnion) => Left(badOnion)
    }
```
And then you use `decryptOnion(PaymentOnionCodecs.paymentOnionPerHopPayloadCodec)` and `decryptOnion(PaymentOnionCodecs.trampolineOnionPerHopPayloadCodec)`.

t-bast: Good catch! I had done almost exactly the same thing, with a tiny difference, I was doing:

```
case Attempt.Successful(DecodeResult(perHopPayload: T, _)) => Right(DecodedOnionPacket(perHopPayload, nextPacket))
```

But that wasn't working because of type erasure (I had to add the `unchecked`).
I didn't even try removing the `: T` type constraint because I thought it was this method's responsibility to enforce the type bounds, but with your proposal it's actually the `perHopPayloadCodec` function that enforces it and it's all fine.

Thanks!

t-bast: Done in https://github.com/ACINQ/eclair/pull/2052/commits/17ed855e8320d51e78793dc80824ed38f771097d

thomash-acinq (approved):

PR #2059, opened 2021-11-09
Clarify route blinding types

t-bast:
We rename the `EncryptedRecipientData` types. The data it contains is namespaced to usages for route blinding, so we
make that explicit. In onion messages, since route blinding can be used in the _forward_ direction, it's not always data encrypted by the recipient, so we drop that part.

This way if future scenarios use another kind of encrypted tlv stream we won't have name clashes (e.g. encrypted state backup).

We also update the route blinding official test vectors.


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
```diff
@@ -0,0 +1,99 @@
...
+import fr.acinq.eclair.{ShortChannelId, UInt64}
+import scodec.bits.ByteVector
+
+import scala.util.Try
+
+/**
+ * Created by t-bast on 19/10/2021.
+ */
+
+object RouteBlinding {
```
t-bast, line 30: What do you think of including it in a `RouteBlinding` object so that we use it as a namespace?
It has the advantage of being very explicitly scoped, so we're sure we won't have issues in the future, it's very explicit, but do you think it's too cumbersome?

thomash-acinq: It feels weird that it's the only thing inside `RouteBlinding`. Also I'm not sure route blinding has its place inside `wire.protocol`, route blinding is what it's used for but in `wire.protocol` we only put types and codecs independently of how they are used.

t-bast: I agree, I'm not entirely satisfied...What I want to somehow convey is that the trait (`EncryptedDataTlv`) and codecs here are really specific to route blinding, they will only ever be used in the context of route blinding.

Because later we will other other cases where we have an encrypted tlv stream, but it will be a completely separate namespace (for example for peer encrypted state backup). So my idea was that we would have `RouteBlinding.EncryptedDataTlv` for tlvs inside a route blinding encrypted blob, and `PeerBackup.EncryptedDataTlv` for tlvs inside an encrypted peer backup. Does that make sense?

The only alternative I see is to instead extend the names: `sealed trait RouteBlindingEncryptedDataTlv` and `sealed trait PeerBackupEncryptedDataTlv`, do you think it's better? It's even more explicit than an encapsulating object.

thomash-acinq: I think I prefer extending the name. It's a bit verbose but it's fine.

t-bast: Done in https://github.com/ACINQ/eclair/pull/2059/commits/f9e689d6fd527303d5806e025abafbadb9b0c8aa

thomash-acinq (approved):

PR #2060, opened 2021-11-09
Simplify onion message codec

t-bast:
This PR was quite a trip. It started because I realized after merging #1957 that since the sphinx packet doesn't really use a prefix, but rather a mix of prefix and suffix (since the mac is encoded after the payload), I actually wasn't sure the `_ - 66, _ + 66` trick was correct.

So I dived into the tests and added new ones playing with length more extensively, and I ended up with some failing tests. At some point I thought our `bytes32` codec was the culprit, so I added tests for that, but it turned out it wasn't the culprit (but I think it's worth keeping the tests, we didn't have enough coverage there). It took me quite a while to realize that it was actually my test that wasn't correctly written :face_with_head_bandage: 

Now everything is green, we have more coverage (in particular for the remaining bytes / not enough bytes cases) and I think the codec is simpler to understand, let me know what you think.

Note that the previous codec was correct, it did pass all tests, but I found it harder to understand (scodec wizardry).


thomash-acinq (approved):
Yes, it's a lot simpler. That's what I wanted to do but couldn't find a way to do it.

PR #2061, opened 2021-11-09
Relay onion messages

thomash-acinq:
Actually relay onion messages and add API to send onion messages.
Follows  #1957

### DIFF
--- a/docs/release-notes/eclair-vnext.md
+++ b/docs/release-notes/eclair-vnext.md
@@ -18,6 +18,10 @@ Node operators should watch this file very regularly.
 An event is also sent to the event stream for every such notification.
 This lets plugins notify the node operator via external systems (push notifications, email, etc).
 
+### Initial support for onion messages
+
+Eclair now supports the feature `option_onion_messages`. If this feature is enabled, eclair will relay onion messages, initiating or receiving onion messages is not supported yet. 
+
 ### API changes
 
 #### Timestamps
--- a/docs/release-notes/eclair-vnext.md
+++ b/docs/release-notes/eclair-vnext.md
@@ -20,7 +20,9 @@ This lets plugins notify the node operator via external systems (push notificati
 
 ### Initial support for onion messages
 
-Eclair now supports the feature `option_onion_messages`. If this feature is enabled, eclair will relay onion messages, initiating or receiving onion messages is not supported yet. 
+Eclair now supports the feature `option_onion_messages`. If this feature is enabled, eclair will relay onion messages.
+It can also send onion messages with the `sendonionmessage` API.
+Messages sent to Eclair will be ignored.
 
 ### API changes
 
@@ -63,6 +65,14 @@ Examples:
     }
   }
   ```
+  
+#### Sending onion messages
+
+You can now send onion messages with `sendonionmessage`.
+It expects `--route` a list of `nodeId` to send the message through, the last one being the recipient, and `--content` the content of the message as an encoded TLV stream in hexadecimal.
+It also accepts `--pathId` as an encoded TLV stream in hexadecimal.
+Sending to a blinded route (as a reply to a previous message) is not supported.
+
 
 This release contains many other API updates:
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -47,6 +47,7 @@ import fr.acinq.eclair.router.{NetworkStats, Router}
 import fr.acinq.eclair.wire.protocol._
 import grizzled.slf4j.Logging
 import scodec.bits.ByteVector
+import scodec.{Attempt, DecodeResult, codecs}
 
 import java.nio.charset.StandardCharsets
 import java.util.UUID
@@ -151,7 +152,7 @@ trait Eclair {
 
   def verifyMessage(message: ByteVector, recoverableSignature: ByteVector): VerifiedMessage
 
-  def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: PublicKey): String
+  def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: PublicKey, userCustomContent: ByteVector): String
 }
 
 class EclairImpl(appKit: Kit) extends Eclair with Logging {
@@ -505,18 +506,23 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
     }
   }
 
-  override def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: PublicKey): String = {
+  override def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: PublicKey, userCustomContent: ByteVector): String = {
     val sessionKey = randomKey()
     val blindingSecret = randomKey()
-    val (nextNodeId, message) =
-      OnionMessages.buildMessage(
-        sessionKey,
-        blindingSecret,
-        intermediateNodes.map(OnionMessages.IntermediateNode(_)),
-        Left(OnionMessages.Recipient(destination, None)),
-        Nil)
-    println("Sending message")
-    MessageRelay(appKit.switchboard, nextNodeId, message)
-    "sent"
+    codecs.list(TlvCodecs.genericTlv).decode(userCustomContent.bits) match {
+      case Attempt.Successful(DecodeResult(userCustomTlvs, _)) =>
+        val (nextNodeId, message) =
+        OnionMessages.buildMessage(
+          sessionKey,
+          blindingSecret,
+          intermediateNodes.map(OnionMessages.IntermediateNode(_)),
+          Left(OnionMessages.Recipient(destination, None)),
+          Nil,
+          userCustomTlvs)
+        MessageRelay(appKit.switchboard, nextNodeId, message)
+        "sent"
+      case Attempt.Failure(cause) => cause.message
+    }
+
   }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -523,9 +523,7 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
             Left(OnionMessages.Recipient(route.last, pathId)),
             Nil,
             userCustomTlvs)
-        val relay = appKit.system.spawnAnonymous(MessageRelay())
-        relay.ask((ref: typed.ActorRef[MessageRelay.Status]) => MessageRelay.RelayMessage(appKit.switchboard, nextNodeId, message, ref))
-          .map {
+        (appKit.switchboard ? OnionMessages.SendMessage(nextNodeId, message)).mapTo[MessageRelay.Status].map {
             case MessageRelay.Success => SendOnionMessageResponse(sent = true, None)
             case MessageRelay.Failure(f) => SendOnionMessageResponse(sent = false, Some(f.toString))
           }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -525,7 +525,7 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
             case MessageRelay.Success => SendOnionMessageResponse(sent = true, None)
             case MessageRelay.Failure(f) => SendOnionMessageResponse(sent = false, Some(f.toString))
           }
-      case Attempt.Failure(cause) => Future.successful(SendOnionMessageResponse(sent = false, Some(cause.message)))
+      case Attempt.Failure(cause) => Future.successful(SendOnionMessageResponse(sent = false, Some(s"the `content` field is invalid, it must contain encoded tlvs: ${cause.message}")))
     }
 
   }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -161,14 +161,16 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
 
   implicit val ec: ExecutionContext = appKit.system.dispatcher
 
-  implicit val typedSystem: typed.ActorSystem[Nothing] = appKit.system.toTyped
-
   // We constrain external identifiers. This allows uuid, long and pubkey to be used.
   private val externalIdMaxLength = 66
 
-  override def connect(target: Either[NodeURI, PublicKey])(implicit timeout: Timeout): Future[String] = target match {
-    case Left(uri) => appKit.switchboard.toTyped.ask((ref: typed.ActorRef[PeerConnection.ConnectionResult]) => Peer.Connect(uri, ref)).map(_.toString)
-    case Right(pubKey) => appKit.switchboard.toTyped.ask((ref: typed.ActorRef[PeerConnection.ConnectionResult]) => Peer.Connect(pubKey, None, ref)).map(_.toString)
+  override def connect(target: Either[NodeURI, PublicKey])(implicit timeout: Timeout): Future[String] = {
+    implicit val typedScheduler: typed.Scheduler = appKit.system.scheduler.toTyped
+    appKit.switchboard.toTyped[Peer.Connect]
+      .ask[PeerConnection.ConnectionResult](ref => target.fold(
+        uri => Peer.Connect(uri, ref),
+        nodeId => Peer.Connect(nodeId, None, ref))
+      ).map(_.toString)
   }
 
   override def disconnect(nodeId: PublicKey)(implicit timeout: Timeout): Future[String] = {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -17,8 +17,8 @@
 package fr.acinq.eclair
 
 import akka.actor.{ActorRef, typed}
-import akka.actor.typed.scaladsl.AskPattern.Askable
-import akka.actor.typed.scaladsl.adapter.ClassicSchedulerOps
+import akka.actor.typed.scaladsl.AskPattern.{Askable, schedulerFromActorSystem}
+import akka.actor.typed.scaladsl.adapter.{ClassicActorRefOps, ClassicActorSystemOps, ClassicSchedulerOps}
 import akka.pattern._
 import akka.util.Timeout
 import com.softwaremill.quicklens.ModifyPimp
@@ -159,12 +159,14 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
 
   implicit val ec: ExecutionContext = appKit.system.dispatcher
 
+  implicit val typedSystem: typed.ActorSystem[Nothing] = appKit.system.toTyped
+
   // We constrain external identifiers. This allows uuid, long and pubkey to be used.
   private val externalIdMaxLength = 66
 
   override def connect(target: Either[NodeURI, PublicKey])(implicit timeout: Timeout): Future[String] = target match {
-    case Left(uri) => appKit.switchboard.ask((ref: typed.ActorRef[PeerConnection.ConnectionResult]) => Peer.Connect(uri, ref)).map (_.toString)
-    case Right(pubKey) => appKit.switchboard.ask((ref: typed.ActorRef[PeerConnection.ConnectionResult]) => Peer.Connect(pubKey, None, ref)).map(_.toString)
+    case Left(uri) => appKit.switchboard.toTyped.ask((ref: typed.ActorRef[PeerConnection.ConnectionResult]) => Peer.Connect(uri, ref)).map(_.toString)
+    case Right(pubKey) => appKit.switchboard.toTyped.ask((ref: typed.ActorRef[PeerConnection.ConnectionResult]) => Peer.Connect(pubKey, None, ref)).map(_.toString)
   }
 
   override def disconnect(nodeId: PublicKey)(implicit timeout: Timeout): Future[String] = {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -35,7 +35,8 @@ import fr.acinq.eclair.channel._
 import fr.acinq.eclair.db.AuditDb.{NetworkFee, Stats}
 import fr.acinq.eclair.db.{IncomingPayment, OutgoingPayment}
 import fr.acinq.eclair.io.Peer.{GetPeerInfo, PeerInfo}
-import fr.acinq.eclair.io.{NodeURI, Peer, PeerConnection}
+import fr.acinq.eclair.io.{MessageRelay, NodeURI, Peer, PeerConnection}
+import fr.acinq.eclair.message.OnionMessages
 import fr.acinq.eclair.payment._
 import fr.acinq.eclair.payment.receive.MultiPartHandler.ReceivePayment
 import fr.acinq.eclair.payment.relay.Relayer.{GetOutgoingChannels, OutgoingChannels, RelayFees, UsableBalance}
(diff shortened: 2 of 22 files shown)


thomash-acinq:
@t-bast  Not ready yet but does the way I use actors look good?

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fr.acinq.eclair.io
+
+import akka.actor.{Actor, ActorRef, ActorRefFactory, Props}
```
t-bast, line 19: We try to use akka typed for all new actors, it provides stricter types for the messages sent and received (cleaner data contract that can be enforced by the compiler). See https://doc.akka.io/docs/akka/current/typed/index.html for the akka doc, it's always useful to regularly go through it.

A somewhat simple example of how to create a typed actor is the `BlockchainWatchdog.scala`. You have to define a `sealed trait Command` and subtypes for all messages you handle. You'll need to dive in a bit into the concept of message adapter to receive the `PeerConnectionResult` (see https://doc.akka.io/docs/akka/current/typed/interaction-patterns.html#adapted-response).

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Message.scala
```diff
@@ -39,5 +40,11 @@ trait Message {
     }
   }
 
-  val messageRoutes: Route = signMessage ~ verifyMessage
+  val sendOnionMessage: Route = postRequest("sendonionmessage") { implicit t =>
+    formFields("intermediateNodes".as[List[PublicKey]](pubkeyListUnmarshaller), "destination".as[PublicKey](publicKeyUnmarshaller)) { (intermediateNodes, destination) =>
```
t-bast, line 44: We should also take a hex-encoded payload for the recipient (which would be a tlv stream, not length-prefixed, that we should inject into the final payload). Probably no need to validate it at that point, it's for experimentation, so the caller should make sure they don't mess it up.

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
```diff
@@ -0,0 +1,45 @@
...
+class MessageRelay private(switchboard: ActorRef, nextNodeId: PublicKey, msg: OnionMessage) extends Actor {
+
+  import MessageRelay._
+
+  override def receive: Receive = {
+    case Relay =>
+      switchboard ! Peer.Connect(nextNodeId, None)
+    case PeerConnection.ConnectionResult.Connected | PeerConnection.ConnectionResult.AlreadyConnected =>
+      sender() ! msg
+      context.stop(self)
```
t-bast, line 41: I'm not sure yet, but I don't think we should stop ourselves immediately.
We should probably watch the peer actor and stop ourselves when they stop.
This way later we can implement a mechanism to have this `MessageRelay` actor actively stop the peer after a while if the `Peer` is inactive (not sure of the details of how we'll do that yet though).

t-bast (commented):
> @t-bast Not ready yet but does the way I use actors look good?

I think it's the right direction, yes!
I put a few comments to guide you to akka typed.

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Message.scala
thomash-acinq: The problem is that the final payload contains the `RouteBlindingEncryptedDataTlv`. If I let the user set this manually it will be completely unusable. The other solution is to duplicate the `buildMessage` function with some ugly ad-hoc code to build a TLV stream from a combination of type TLV and an untyped `ByteVector`.

t-bast: Have a look at what we do in `PaymentInitiator.scala` with `SendPaymentToNode`: we let users provide a list of `GenericTlv` that we insert in the tlv stream. I think it would work fine in our case, these should simply be added to the tlv stream of `OnionMessagePayloadTlv`, it shouldn't impact the blinded part AFAICT.

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
thomash-acinq: I don't think that's possible without some major changes. The problem is that there is no sender in Akka typed and `Switchboard`, `Peer` and `PeerConnection` all rely on it for the connection.

t-bast: I don't see how this is an issue, we've done that in multiple other actors without any problem?
See for example what we did in the `Channel`:

```
val replyTo = if (c.replyTo == ActorRef.noSender) sender() else c.replyTo
```

thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
thomash-acinq: Done

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Message.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -243,6 +245,19 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
         d.channels.values.toSet[ActorRef].foreach(_ ! INPUT_DISCONNECTED) // we deduplicate with toSet because there might be two entries per channel (tmp id and final id)
         gotoConnected(connectionReady, d.channels)
 
+      case Event(msg: OnionMessage, _: ConnectedData) =>
+        if (nodeParams.features.hasFeature(Features.OnionMessages)) {
+          OnionMessages.process(nodeParams.privateKey, msg) match {
+            case OnionMessages.DropMessage(reason) =>
+              log.debug(s"dropping message from ${remoteNodeId.value.toHex}: ${reason.toString}")
+            case OnionMessages.RelayMessage(nextNodeId, dataToRelay) =>
+              context.spawnAnonymous(MessageRelay(switchboard, nextNodeId, dataToRelay))
```
t-bast, line 254: From an architecture's point of view, I'm not sure the `Peer` should be spawning these actors. The issue with that is that if the `Peer` is stopped, all its children are stopped, which could terminate a message relay in the middle of its operations without any warning. This isn't an issue right now, but when we automatically stop idle peers with which we have no channels, there is a risk that this will bite us.

I think the `MessageRelay` actors should be spawned by the `Switchboard`, which is a singleton that should always stay alive. I think it's the `Switchboard` that should handle a `SendOnionMessage` request and create the child `MessageRelay` actor. Similarly, it means that `sendOnionMessage` in `Eclair.scala` shouldn't spawn the `MessageRelay` actor but instead just send a message to the `Switchboard`.

eclair-core/src/test/scala/fr/acinq/eclair/io/MessageRelaySpec.scala
```diff
@@ -0,0 +1,78 @@
...
+import org.scalatest.Outcome
+import org.scalatest.funsuite.FixtureAnyFunSuiteLike
+
+case class FakeChannelFactory(channel: TestProbe) extends ChannelFactory {
+  override def spawn(context: ActorContext, remoteNodeId: PublicKey, origin_opt: Option[ActorRef]): ActorRef = {
+    channel.ref
+  }
+}
+
+class MessageRelaySpec extends TestKitBaseClass with FixtureAnyFunSuiteLike {
```
t-bast, line 40: If you follow my suggestion to have the switchboard spawn the message relay actors, I think this test suite can be greatly simplified. It should simply spawn message relay actors with test probes in all the `ActorRef` parameters, send the messages the `MessageRelay` actor expects and check its responses on the test probes.

Separately of that unit test suite, you should add end to end tests in `PaymentIntegrationSpec` to verify that the complete flow of sending an onion message works.

t-bast (commented):
A couple of comments on the actor architecture itself.
I think that if you apply all these comments, it will both simplify the PR and make the system more testable and extensible in the future.

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
t-bast: It's true that we don't need it yet, we can introduce that when we implement automatically stopping these ephemeral peers.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -150,18 +151,22 @@ trait Eclair {
   def signMessage(message: ByteVector): SignedMessage
 
   def verifyMessage(message: ByteVector, recoverableSignature: ByteVector): VerifiedMessage
+
+  def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: PublicKey, userCustomContent: ByteVector): String
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
```diff
@@ -134,6 +134,6 @@ class Client(keyPair: KeyPair, socks5ProxyParams_opt: Option[Socks5ProxyParams],
 
 object Client {
 
-  def props(keyPair: KeyPair, socks5ProxyParams_opt: Option[Socks5ProxyParams], peerConnectionConf: PeerConnection.Conf, switchboard: ActorRef, router: ActorRef, address: InetSocketAddress, remoteNodeId: PublicKey, origin_opt: Option[ActorRef]): Props = Props(new Client(keyPair, socks5ProxyParams_opt, peerConnectionConf, switchboard, router, address, remoteNodeId, origin_opt))
+  def props(keyPair: KeyPair, socks5ProxyParams_opt: Option[Socks5ProxyParams], peerConnectionConf: PeerConnection.Conf, switchboard: ActorRef, router: ActorRef, address: InetSocketAddress, remoteNodeId: PublicKey, origin_opt: Option[typed.ActorRef[PeerConnection.ConnectionResult]]): Props = Props(new Client(keyPair, socks5ProxyParams_opt, peerConnectionConf, switchboard, router, address, remoteNodeId, origin_opt))
```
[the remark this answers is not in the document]
thomash-acinq: Converting a typed ActorRef to an untyped one is unsafe and you lose all the benefits of using typed actors.

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
```diff
@@ -0,0 +1,38 @@
...
+package fr.acinq.eclair.io
+
+import akka.actor.ActorRef
+import akka.actor.typed.Behavior
+import akka.actor.typed.scaladsl.Behaviors
+import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.eclair.wire.protocol.OnionMessage
+
+object MessageRelay {
+  def apply(switchboard: ActorRef, nextNodeId: PublicKey, msg: OnionMessage): Behavior[PeerConnection.ConnectionResult] = {
```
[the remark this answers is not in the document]
thomash-acinq: Done.
Although I strongly dislike this "best practice", it creates two big regressions:
- using `MessageRelay` is now more complicated, you need to call `MessageRelay()`, spawn a new actor and send it a command instead of just calling a single function
- the actor now has two states that accept different types of messages so we lose some of the benefits from using typed actors.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -243,6 +245,19 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
...
+            case msg: OnionMessages.ReceiveMessage =>
+              log.info(s"received message from ${remoteNodeId.value.toHex}: $msg")
```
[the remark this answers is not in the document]

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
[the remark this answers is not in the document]
t-bast: > using MessageRelay is now more complicated, you need to call MessageRelay(), spawn a new actor and send it a command instead of just calling a single function

Well, following the same reasoning, you could say that using multiple actors instead of a single actor that does everything is more complicated (from the caller's point of view), but I think you would agree that it's better engineering?

You should think about actor spawning as a constructor. You usually don't put your business logic inside a constructor, do you? You put it in methods on your object, right? Because that is separation of concern / modularity, which lets your component be extended more easily for unknown future use-cases, and be more easily unit tested.

eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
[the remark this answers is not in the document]
t-bast: It's a trade-off: if we follow your argument, we should potentially update `ActorRef`s across the whole codebase in a single PR, which isn't part of the main feature the PR focuses on and will make it longer to integrate it because there are more unrelated things to review.

Akka introduced these conversions explicitly to make it easy for codebases to update gradually, actor by actor to akka typed.

I guess for this specific PR it doesn't bleed into too many actors in the codebase, so maybe it's ok. @pm47 what is your recommendation as the akka-typed guru?

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala

eclair-core/src/test/scala/fr/acinq/eclair/io/MessageRelaySpec.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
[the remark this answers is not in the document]
thomash-acinq: It is normal to split the logic into multiple functions / actors but if the caller needs to know about it that's a design defect.

If the objects needs to be initialized, then I think this initialization logic should go into the constructor, not in a separate `init` function, it's the same thing as `var x; x = "toto";` vs `var x = "toto;"`, I think you agree that the latter is better and that's why modern languages don't allow the former.

eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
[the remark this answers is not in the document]
pm47: > I don't think you should modify Client, ClientSpawner, Peer and the other existing actors at all to introduce akka typed into them.

It's a tough one: in general I would keep the number of modified files to a minimum. In this case though, since `origin_opt` is used as a callback, using a typed actor as early as possible really brings an improvement. It's true that _akka_ specifically provides `messageAdapters` to handle conversions in typed actors. Those are useful when we can't modify other actors, but a major drawback is that any conversion error or missing adapter will be discovered at runtime. Here I would say that the change in `Client` is trivial enough that it makes it worthwhile.

OTOH, changing parts of the code not directly related to the PR at hand increases the risk of regressions (especially in `ReconnectionTask` which is a tricky one), so we must be careful with that kind of move.

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  case object Success extends Status
+
+  case class Failure(failure: PeerConnection.ConnectionResult.Failure) extends Status
+
+  sealed trait Command
+
+  case class RelayMessage(switchboard: ActorRef, nextNodeId: PublicKey, msg: OnionMessage, replyTo: typed.ActorRef[Status]) extends Command
+
+  case class WrappedConnectionResult(result: PeerConnection.ConnectionResult) extends Command
```
pm47, line 36: Always use formatting to organize the commands for better readability. Usually we put the `Command` on top.

```suggestion
  // @formatter:off
  sealed trait Command
  case class RelayMessage(switchboard: ActorRef, nextNodeId: PublicKey, msg: OnionMessage, replyTo: typed.ActorRef[Status]) extends Command
  case class WrappedConnectionResult(result: PeerConnection.ConnectionResult) extends Command

  sealed trait Status
  case object Success extends Status
  case class Failure(failure: PeerConnection.ConnectionResult.Failure) extends Status
  // @formatter:on
```

eclair-core/src/main/scala/fr/acinq/eclair/io/MessageRelay.scala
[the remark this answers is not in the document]
t-bast: I completely agree with both of your points, but I don't see how they apply to the current scenario though...?

> if the caller needs to know about it that's a design defect.

I agree with this general statement, but in our case the caller wants to create an actor and have this new actor relay a message. So all of this is part of what the caller expects, it doesn't expose to the caller something he shouldn't know about. The caller creates an actor, then asks it to relay a message: that looks like a perfectly normal calling pattern to me.

Maybe we're talking past each other because you're thinking about it from a functional programming point of view, as just a unique function call (and want to abstract away the fact that it has to create an actor to perform the task)? But I think actor programming is more of an object-oriented paradigm, where you create actors and then ask them to perform actions. The way I view it it feels like a better separation of concern, but maybe that's just me.

> I think this initialization logic should go into the constructor

I agree as well, but sending an outgoing message is not part of an initialization logic, it's an external side-effect. I think that probably pinpoints the core of our misunderstanding: if sending this outgoing message feels part of the internal initialization logic to you, your points make sense.

My point is then simply that calling another actor to trigger business logic doesn't feel part of the initialization logic of the actor to me.

I don't think it's worth arguing too much about this now, especially since the actor is currently so simple, but hopefully once we extend this actor with more initialization logic and more commands to handle, it will make sense why this separation of concern is better (e.g. when it may have to handle various first actions).

eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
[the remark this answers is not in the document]
t-bast: Ok interesting, I was a bit afraid of mostly the (seemingly unrelated) changes it required to `Eclair.scala` and `EclairInternalSerializers.scala`, but I'll review it in more details then.

eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
```diff
@@ -129,6 +130,8 @@ object EclairInternalsSerializer {
     (path: String) => system.provider.resolveActorRef(path),
     (actor: ActorRef) => Serialization.serializedActorPath(actor))
 
+  def typedActorRefCodec[T](system: ExtendedActorSystem):Codec[typed.ActorRef[T]] = actorRefCodec(system).xmap(_.toTyped, _.toClassic)
+
```
pm47, line 134: [According to akka devs](https://discuss.lightbend.com/t/akka-typed-serialization/4336), this should be instead:

```suggestion
  def typedActorRefCodec[T](system: typed.ActorSystem[Nothing]): Codec[typed.ActorRef[T]] = variableSizeBytes(uint16, utf8).xmap(
    (path: String) => ActorRefResolver.get(system).resolveActorRef(path),
    (actor: typed.ActorRef[T]) => ActorRefResolver.get(system).toSerializationFormat(actor))
```

NB: untested, and needs to be called with `system.toTyped`

pm47: Note to self: this will require synchronous update of `eclair-front`

eclair-core/src/main/scala/fr/acinq/eclair/io/Client.scala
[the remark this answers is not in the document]
pm47: I indeed missed those on my first pass. This is not so trivial and shifts the balance towards keeping untyped actors.

I suggest we stick to untyped actors for this PR, and we do a follow up PR that also types `Switchboard`, WDYT @thomash-acinq ?

eclair-core/src/test/scala/fr/acinq/eclair/io/MessageRelaySpec.scala
```diff
@@ -0,0 +1,93 @@
...
+import fr.acinq.eclair.{TestKitBaseClass, randomKey}
+import org.scalatest.funsuite.AnyFunSuiteLike
+
+case class FakeChannelFactory(channel: TestProbe) extends ChannelFactory {
+  override def spawn(context: ActorContext, remoteNodeId: PublicKey, origin_opt: Option[ActorRef]): ActorRef = {
+    channel.ref
+  }
+}
+
+class MessageRelaySpec extends TestKitBaseClass with AnyFunSuiteLike {
```
t-bast, line 37: This test class should use the typed actor test framework.
I've done that change in https://github.com/ACINQ/eclair/commit/ece7bc08504fb5aba991e72d4f3ea93b1928ac6e and beefed up the integration tests.

It took me a lot more time than I expected, because I wanted to test that disconnected nodes that knew each other's addresses from previous interactions would automatically reconnect on-the-fly, but it was made difficult because we aggressively prune our DBs (which is good):

- when we don't have any channels with a peer we remove them from the `PeersDb`
- when a node doesn't have any visible channel on the network we remove them from the `NetworkDb`

So the test setup is a bit more complex now, but it lets us test more cases that match real-world usage.

t-bast (commented):
Looking good, thanks for the changes!
A few minor comments and additional tests (in a separate commit in my branch) and we should be good to go.

eclair-core/src/test/scala/fr/acinq/eclair/io/MessageRelaySpec.scala
thomash-acinq: Done, I've merged your commit.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -502,4 +511,26 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
       }
     }
   }
+
+  override def sendOnionMessage(route: Seq[PublicKey], userCustomContent: ByteVector, pathId: Option[ByteVector])(implicit timeout: Timeout): Future[SendOnionMessageResponse] = {
+    val sessionKey = randomKey()
+    val blindingSecret = randomKey()
+    codecs.list(TlvCodecs.genericTlv).decode(userCustomContent.bits) match {
+      case Attempt.Successful(DecodeResult(userCustomTlvs, _)) =>
```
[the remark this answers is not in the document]
thomash-acinq: According to the doc of `codecs.list`: "When decoding, codec.decode is called repeatedly until there are no more remaining bits and the value result of each decode is returned in the list."
So the remainder has to be empty.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -502,4 +509,26 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
...
+            blindingSecret,
+            route.dropRight(1).map(OnionMessages.IntermediateNode(_)),
+            Left(OnionMessages.Recipient(route.last, pathId)),
+            Nil,
+            userCustomTlvs)
+        (appKit.switchboard ? OnionMessages.SendMessage(nextNodeId, message)).mapTo[MessageRelay.Status].map {
+            case MessageRelay.Success => SendOnionMessageResponse(sent = true, None)
+            case MessageRelay.Failure(f) => SendOnionMessageResponse(sent = false, Some(f.toString))
+          }
+      case Attempt.Failure(cause) => Future.successful(SendOnionMessageResponse(sent = false, Some(cause.message)))
```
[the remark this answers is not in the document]
thomash-acinq: Done

docs/release-notes/eclair-vnext.md
```diff
@@ -18,6 +18,10 @@ Node operators should watch this file very regularly.
 An event is also sent to the event stream for every such notification.
 This lets plugins notify the node operator via external systems (push notifications, email, etc).
 
+### Initial support for onion messages
+
+Eclair now supports the feature `option_onion_messages`. If this feature is enabled, eclair will relay onion messages, initiating or receiving onion messages is not supported yet.
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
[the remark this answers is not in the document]
t-bast: Good catch!

thomash-acinq:
I've tested it https://github.com/ElementsProject/lightning/pull/4921 can relay messages between eclair nodes.

PR #2063, opened 2021-11-12
Add support for `option_payment_metadata`

t-bast:
Add support for `option_payment_metadata` as specified in https://github.com/lightning/bolts/pull/912
This feature lets recipients generate "light" invoices without storing them in the local DB until they are paid.

The trick is to simply put the important data (mainly `payment_secret` and `total_amount`) in an encrypted blob inside the invoice, and verify its presence and integrity when receiving a payment that contains a `payment_metadata` field (and insert it in the DB on-the-fly).

This PR only implements support for the sender side, as recipients will only be able to leverage this features once all senders support it.

We do include a `payment_metadata` in all invoices we generate, which lets us know whether our payers support this feature: this is important data to have before we decide to make it mandatory and actually implement storage-less invoices.

We also take this opportunity to have our first invoice-only feature to implement feature bit filtering, as required by the spec (`IN9` flags as specified in [bolt 9](https://github.com/lightning/bolts/blob/master/09-features.md)): see the first commit of the PR.

The commits are somewhat independent and can be reviewed separately.


joostjager:
So for interop testing, we can only do eclair -> lnd?

t-bast:
> So for interop testing, we can only do eclair -> lnd?

Right now yes, as the description mentions including a `payment_metadata` in invoices is not implemented yet. But it will when I have time to get back to it.

joostjager:
Shall I get you a testnet invoice so that you can try to pay it?

t-bast:
I'm not ready yet for compat tests, I should address my first three todos before that, I'll let you know when I'm ready.

t-bast:
@joostjager this should now be ready for interop testing :tada:

joostjager:
Tried, but running into this when opening a channel from lnd:

```
2021-12-07 13:38:31,151 WARN  fr.acinq.eclair.io.Peer n:028a52ef0c0d4fffe72239d723d3969f779d19d4645b774b9df2d463ee264fee45 c:18b9b4ef528d25374188358d068c3798c843f83c611472c8d999f69743209a6f - ignoring open_channel: option_channel_type was negotiated but channel_type is missing
```

I thought this thing was solved?

t-bast:
> I thought this thing was solved?

Looks like it's not! Are you sure you're running an up-to-date lnd?
This log means that we both have activated the `option_channel_type` feature, but lnd didn't send one in the tlv stream.
The spec clearly says that it must send the tlv.

joostjager:
I also tried with vanilla `v0.14.1-beta`, same story.

joostjager:
One finding is that `eclair-cli parseinvoice` doesn't show the metadata of my test invoice `lnbcrt5u1ps67e6rpp53hg4urhja4t8xedl5xmvchru4gnu42ecte49p2ejrl9lqzt5crqqdqqmqgqypqxpq9cqzpgxqyz5vqsp59ywh2qdt07gkypxap72gwgalpnuafqgu7c5w4zpc3pc7p7a4ks4s9q2gqqqqqyssqfa22zca7m58e0xtrv4ny7p6fsqkkwsh2h8an3djl4n7mu3jd48qpvc5cye29e3jp4pu9ny0utfwzyrmlss6a9ahep03hc0xh0c0axncqy8afdn`

joostjager:
We now have this problem: https://github.com/lightning/bolts/pull/912#discussion_r764045143

joostjager:
Spec update to use tlv field 16. If you update this PR, I can do another attempt at interop.

t-bast:
> Spec update to use tlv field 16. If you update this PR, I can do another attempt at interop.

Done.

joostjager:
The interop test passed. I received payment metadata from eclair:

```
            "htlcs": [
                {
                    "chan_id": "20469607974371328",
                    "htlc_index": "4",
                    "amt_msat": "1000000",
                    "accept_height": 18636,
                    "accept_time": "1638951794",
                    "resolve_time": "1638951794",
                    "expiry_height": 18676,
                    "state": "SETTLED",
                    "custom_records": {
                    },
                    "mpp_total_amt_msat": "1000000",
                    "amp": null,
                    "metadata": "010203"
                }
            ],
```

t-bast:
Thanks for testing this Joost!

joostjager:
Tested LND->Eclair interop, seems to work too

`2022-01-03 21:10:39,999 INFO  f.a.e.p.r.PaymentHandler PAY h:1cfd29bfdf71bebc5b1b20ddb56dbc857db45c9f2db58551cc98deb6200a3e07 - received payment for amount=50000 msat totalAmount=50000 msat paymentMetadata=00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000`

t-bast:
Thanks for the tests @joostjager!

t-bast:
@thomash-acinq I put you as reviewer as this touches invoice stuff, so it may be loosely related to your work on offers (at least it may be useful for you to track this change). I'm interested in your feedback on this, it's ready to review as the spec PR has been merged recently.

thomash-acinq (commented):
Sorry for missing this PR earlier. It would be nice some integration tests, especially on how it interacts with trampoline.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -109,7 +109,8 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
 
   def currentBlockHeight: Long = blockCount.get
 
-  def featuresFor(nodeId: PublicKey): Features = overrideFeatures.getOrElse(nodeId, features)
+  /** Returns the features that should be used in our init message with the given peer. */
+  def featuresFor(nodeId: PublicKey): Features = overrideFeatures.getOrElse(nodeId, features).initFeatures()
```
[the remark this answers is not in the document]
t-bast: Agreed, done in https://github.com/ACINQ/eclair/pull/2063/commits/96ce001b35235e41cf80bb896e7850fc87df3125

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentRequest.scala
```diff
@@ -67,6 +66,11 @@ case class PaymentRequest(prefix: String, amount: Option[MilliSatoshi], timestam
     case PaymentRequest.DescriptionHash(h) => Right(h)
   }.get
 
+  /**
+   * @return the payment metadata
```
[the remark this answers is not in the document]
t-bast: I expanded it a bit in https://github.com/ACINQ/eclair/pull/2063/commits/96ce001b35235e41cf80bb896e7850fc87df3125 to mention the feature, this way we can search for it in the codebase and find the explanation of how it can be used

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -155,6 +155,12 @@ object OnionPaymentPayloadTlv {
   /** Id of the next node. */
   case class OutgoingNodeId(nodeId: PublicKey) extends OnionPaymentPayloadTlv
 
+  /**
+   * When payment metadata is included in a Bolt 9 invoice, we should send it as-is to the recipient.
```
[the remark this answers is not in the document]
t-bast: Good catch, no idea why I confused these two!
Done in https://github.com/ACINQ/eclair/pull/2063/commits/96ce001b35235e41cf80bb896e7850fc87df3125

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -213,7 +213,10 @@ object OutgoingPaymentPacket {
    *         - the trampoline onion to include in final payload of a normal onion
    */
   def buildTrampolineToLegacyPacket(invoice: PaymentRequest, hops: Seq[NodeHop], finalPayload: PaymentOnion.FinalPayload): (MilliSatoshi, CltvExpiry, Sphinx.PacketAndSecrets) = {
-    val (firstAmount, firstExpiry, payloads) = hops.drop(1).reverse.foldLeft((finalPayload.amount, finalPayload.expiry, Seq[PaymentOnion.PerHopPayload](finalPayload))) {
+    // NB: the final payload will never reach the recipient, since the next-to-last trampoline hop will convert that to a legacy payment
```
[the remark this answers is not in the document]
t-bast: The node just before the last node.
Is there a different comment that would make more sense to you?
I looked up the definition of "penultimate", but it says that it means "next to last" so I guess this should be more understandable?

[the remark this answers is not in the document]
thomash-acinq: I think my confusion mainly comes from the term "hop". A hop is between two nodes but you actually mean a specific node. And I don't think it's the penultimate trampoline node that does the conversion, it's the last trampoline node, isn't it?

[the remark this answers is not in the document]
t-bast: Right, I will replace that with node, it makes more sense.

It is the penultimate trampoline node that does the conversion, not the last one, because for now the last node is the recipient: we build a trampoline onion for the complete route to the recipient (even though we know it will be converted to a legacy payment instead of being relayed). This was mostly done to minimize code changes, because I hoped we would be able to get rid of the "trampoline to legacy" case quickly (I thought wallet would take less time to at least implement the trampoline onion decryption, but that hasn't seen a lot of traction yet...).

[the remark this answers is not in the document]
t-bast: Clarified in https://github.com/ACINQ/eclair/pull/2063/commits/fa82531fa61b6728c84b28212fdcfeaa0b02d4fd

We should change that behavior some day, but I've been delaying that to when we implement the "new" version of trampoline (where trampoline onion is variable-size, as the latest state of the spec PR recommends).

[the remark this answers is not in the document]
thomash-acinq: I wouldn't consider the last node to be a trampoline node. You can call it "the last trampoline node before the recipient" to remove all ambiguity.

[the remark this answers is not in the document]
t-bast: What's important here is that the recipient receives a trampoline payload in the onion, if that wasn't the case the comment would be inaccurate (or worse, the code would be incorrect). I used "trampoline node" to refer to "node that will receive a payload in the trampoline onion" for simplicity, I'll find a better way of expressing that.

thomash-acinq (approved):

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -71,7 +71,13 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
...
               PendingCommandsDb.safeSend(register, nodeParams.db.pendingCommands, p.add.channelId, cmdFail)
             case None =>
-              log.info("received payment for amount={} totalAmount={}", p.add.amountMsat, p.payload.totalAmount)
+              // We log whether the sender included the payment metadata field.
+              // We always set it in our invoices to test whether senders support it.
+              // Once all incoming payments correctly set that field, we can make it mandatory.
+              p.payload.paymentMetadata match {
+                case Some(paymentMetadata) => log.info("received payment for amount={} totalAmount={} paymentMetadata={}", p.add.amountMsat, p.payload.totalAmount, paymentMetadata.toHex)
+                case None => log.info("received payment for amount={} totalAmount={} without payment metadata", p.add.amountMsat, p.payload.totalAmount)
+              }
```
pm47, line 80: With less repetition:
```suggestion
              log.info("received payment for amount={} totalAmount={} paymentMetadata={}", p.add.amountMsat, p.payload.totalAmount, p.payload.paymentMetadata.map(_.toHex).getOrElse("none"))
```

Adding a metric would make it easier to track progress.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -355,6 +373,7 @@ object PaymentOnionCodecs {
     .typecase(UInt64(8), paymentData)
     .typecase(UInt64(10), encryptedRecipientData)
     .typecase(UInt64(12), blindingPoint)
+    .typecase(UInt64(16), paymentMetadata)
```
pm47, line 376: Yes, I will say it one more time: our tlv codecs are awesome.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -223,14 +229,25 @@ object MultiPartHandler {
               val paymentPreimage = paymentPreimage_opt.getOrElse(randomBytes32())
               val paymentHash = Crypto.sha256(paymentPreimage)
               val expirySeconds = expirySeconds_opt.getOrElse(nodeParams.paymentRequestExpiry.toSeconds)
-              val features = {
-                val f1 = Seq(Features.PaymentSecret.mandatory, Features.VariableLengthOnion.mandatory)
-                val allowMultiPart = nodeParams.features.hasFeature(Features.BasicMultiPartPayment)
-                val f2 = if (allowMultiPart) Seq(Features.BasicMultiPartPayment.optional) else Nil
-                val f3 = if (nodeParams.enableTrampolinePayment) Seq(Features.TrampolinePayment.optional) else Nil
-                PaymentRequest.PaymentRequestFeatures(f1 ++ f2 ++ f3: _*)
+              val paymentMetadata = Some(ByteVector64.Zeroes.bytes)
```
pm47, line 232: I would use the smallest possible vector, since our BOLT 11 invoices are already too big:
```suggestion
              val paymentMetadata = Some(ByteVector(0))
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -213,9 +213,12 @@ object OutgoingPaymentPacket {
    *         - the trampoline onion to include in final payload of a normal onion
    */
   def buildTrampolineToLegacyPacket(invoice: PaymentRequest, hops: Seq[NodeHop], finalPayload: PaymentOnion.FinalPayload): (MilliSatoshi, CltvExpiry, Sphinx.PacketAndSecrets) = {
-    val (firstAmount, firstExpiry, payloads) = hops.drop(1).reverse.foldLeft((finalPayload.amount, finalPayload.expiry, Seq[PaymentOnion.PerHopPayload](finalPayload))) {
+    // NB: the final payload will never reach the recipient, since the next-to-last node in the trampoline route will convert that to a legacy payment.
+    // We use the smallest final payload possible, otherwise we may overflow the trampoline onion size.
+    val dummyFinalPayload = PaymentOnion.createSinglePartPayload(finalPayload.amount, finalPayload.expiry, finalPayload.paymentSecret, None)
```
pm47, line 218: I'm lost here. Why did you have to do this in a seemingly completely unrelated commit ("Include payment metadata in all invoices")? Did you actually make the onion overflow during tests? Why can't the next-to-last node pass the `metadata` when it builds the legacy payload. What does `legacy` mean here: non-trampoline or non-tlv? 😵‍💫 

```suggestion
    // NB: the final payload will never reach the recipient, since the next-to-last node in the trampoline route will convert that to a legacy payment.
    // We use the smallest final payload possible, otherwise we may overflow the trampoline onion size.
    val dummyFinalPayload = PaymentOnion.createSinglePartPayload(finalPayload.amount, finalPayload.expiry, finalPayload.paymentSecret, None)
```

pm47 (commented):
LMostlyGTM, just a few nits or suggestions.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -38,6 +38,13 @@ trait Feature {
   def mandatory: Int
   def optional: Int = mandatory + 1
 
+  /** If true, feature should be sent in init messages. */
+  def init: Boolean
+  /** If true, feature should be sent in node announcements. */
+  def nodeAnnouncement: Boolean
+  /** If true, feature should be set in invoices. */
+  def invoice: Boolean
+
```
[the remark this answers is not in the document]
t-bast: > Interesting approach, but how will you deal with C/C-/C+ with a boolean? Is this why you have left those for later?

Yeah it's because it's unclear how these will ever be used in the spec (and they're currently unused).
So we'll deal with it if it ever becomes a thing.

> Btw, shouldn't the new option_payment_metadata be an I+?

That was discussed on the spec PR, at first it was 9+ but then it was changed because there's no reason not to allow it to be optional and it's a very good way to measure adoption before actually relying on it (and making it mandatory).

> why does the spec use 9 to identify invoice feature bits and not, say, V?

IIRC that's because features are encoded as `f` in invoices, and in bech32 `f` has the value `9`. I believe Rusty chose that a while ago when doing these categories.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -219,11 +285,17 @@ object Features {
...
     val mandatory = 50
+    val init = true
+    val nodeAnnouncement = true
+    val invoice = true
   }
 
   case object KeySend extends Feature {
     val rfcName = "keysend"
     val mandatory = 54
+    val init = true
```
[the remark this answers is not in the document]
t-bast: Good catch, thanks!

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -223,14 +229,25 @@ object MultiPartHandler {
...
+              val invoiceFeatures = {
+                val activatedInvoiceFeatures = nodeParams.features.invoiceFeatures().activated.map { case (f, s) => f.supportBit(s) }.toSet
+                val allInvoiceFeatures = if (nodeParams.enableTrampolinePayment) activatedInvoiceFeatures + Features.TrampolinePayment.optional else activatedInvoiceFeatures
+                allInvoiceFeatures.toSeq
```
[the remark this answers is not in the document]
t-bast: I'll try that.

eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
```diff
@@ -76,15 +76,16 @@ object Announcements {
       case address@(_: Tor2) => (3, address)
       case address@(_: Tor3) => (4, address)
     }.sortBy(_._1).map(_._2)
-    val witness = nodeAnnouncementWitnessEncode(timestamp, nodeSecret.publicKey, color, alias, features, sortedAddresses, TlvStream.empty)
+    val nodeAnnouncementFeatures = features.nodeAnnouncementFeatures()
```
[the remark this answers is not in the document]
t-bast: Yes agreed, we should have done this earlier! I think doing it now shouldn't harm anyone, and we should do it asap because in the future it may become more harmful.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: I hesitated and omitted it, but since you also thought about it I'll add a metric.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: It's a trade-off: when we'll want to actually use this, we will probably need 32 to 64 bytes to do DB-less invoices, so shouldn't we test that sender are able to send that much?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: > Did you actually make the onion overflow during tests?

Yes that's exactly it, existing tests start failing because of the addition of the `payment_metadata` that ended up duplicated in the trampoline onion :/

> Why can't the next-to-last node pass the metadata when it builds the legacy payload

That's what it does, so we don't need to have the `payment_metadata` in the _trampoline_ payload for the last node (since that payload is completely ignored because we instead convert the payment to a non-trampoline one).

Here legacy means non-trampoline, I'll clarify that.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -271,13 +271,13 @@ class NodeRelay private(nodeParams: NodeParams,
         val paymentSecret = payloadOut.paymentSecret.get // NB: we've verified that there was a payment secret in validateRelay
         if (Features(features).hasFeature(Features.BasicMultiPartPayment)) {
           context.log.debug("sending the payment to non-trampoline recipient using MPP")
-          val payment = SendMultiPartPayment(payFsmAdapters, paymentSecret, payloadOut.outgoingNodeId, payloadOut.amountToForward, payloadOut.outgoingCltv, nodeParams.maxPaymentAttempts, routingHints, routeParams)
+          val payment = SendMultiPartPayment(payFsmAdapters, paymentSecret, payloadOut.outgoingNodeId, payloadOut.amountToForward, payloadOut.outgoingCltv, nodeParams.maxPaymentAttempts, payloadOut.paymentMetadata, routingHints, routeParams)
```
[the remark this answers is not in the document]
t-bast: I don't think we should: this is used in only one place, and there are subtleties, you need to carefully choose what you take from `payloadOut` and what you take from elsewhere. For example, we don't want to use the `payment_secret` from `payloadOut`. Extracting this to a separate function creates the risk of a subtle bug here.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2063/commits/4c39fd36e1f78f7efdfbdcc533dab6069174043f

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -289,14 +298,21 @@ object PaymentOnion {
   def createNodeRelayToNonTrampolinePayload(amount: MilliSatoshi, totalAmount: MilliSatoshi, expiry: CltvExpiry, targetNodeId: PublicKey, invoice: PaymentRequest): NodeRelayPayload = {
     val tlvs = Seq[OnionPaymentPayloadTlv](AmountToForward(amount), OutgoingCltv(expiry), OutgoingNodeId(targetNodeId), InvoiceFeatures(invoice.features.toByteVector), InvoiceRoutingInfo(invoice.routingInfo.toList.map(_.toList)))
     val tlvs2 = invoice.paymentSecret.map(s => tlvs :+ PaymentData(s, totalAmount)).getOrElse(tlvs)
-    NodeRelayPayload(TlvStream(tlvs2))
+    val tlvs3 = invoice.paymentMetadata.map(m => tlvs2 :+ PaymentMetadata(m)).getOrElse(tlvs2)
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2063/commits/4c39fd36e1f78f7efdfbdcc533dab6069174043f

eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
```diff
@@ -76,15 +76,16 @@ object Announcements {
...
+    val witness = nodeAnnouncementWitnessEncode(timestamp, nodeSecret.publicKey, color, alias, nodeAnnouncementFeatures, sortedAddresses, TlvStream.empty)
     val sig = Crypto.sign(witness, nodeSecret)
     NodeAnnouncement(
       signature = sig,
       timestamp = timestamp,
       nodeId = nodeSecret.publicKey,
       rgbColor = color,
       alias = alias,
-      features = features,
+      features = nodeAnnouncementFeatures,
```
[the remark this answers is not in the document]
t-bast: Agreed, typing those becomes a bit messy because we can upcast `Map[Feature with NodeFeature, FeatureSupport]` to `Map[Feature, FeatureSupport]` so we'd need conversions...

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2063/commits/4c39fd36e1f78f7efdfbdcc533dab6069174043f
I typed `PaymentRequestFeatures.apply`, but I didn't use lenses.
Trampoline should be the only exception, it's mostly a hack because we don't want to advertise yet our trampoline support in `init` and `node_announcement` until the spec PR has more traction, but we'll remove that hack once trampoline is accepted.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2063/commits/4c39fd36e1f78f7efdfbdcc533dab6069174043f

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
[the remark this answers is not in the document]
t-bast: I switched to using traits in https://github.com/ACINQ/eclair/pull/2063/commits/4c39fd36e1f78f7efdfbdcc533dab6069174043f

I tried going further, and propagating the types to `Features`, but it becomes a mess quickly.
The main reason it's hard to properly handle is that we want to accept external data that doesn't comply with those type constraints (we don't want to be too strict and reject `init` or `node_announcement` that include a feature that we can simply ignore).

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: I updated the comment in https://github.com/ACINQ/eclair/pull/2063/commits/87aca520f7da877e5d58db728c1a7c64cb27c1c4

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
pm47: Ah! 🤔 Well I think that here we are just testing that the sender _supports_ the metadata, nothing more. Going further and trying to validate additional requirements (that we are not even sure of?) is a different concern imo. Debatable...

t-bast: That's true, maybe we should wait to have a scheme that we want to use with `payment_metadata` (most likely to get rid of invoice storage until they're paid), see how many bytes that uses, and at that point set a payment metadata of that length?

Then after X months if senders are able to send that many bytes, we can actually use the designed scheme.

t-bast: Updated in https://github.com/ACINQ/eclair/pull/2063/commits/c2daf43855f1effb147119f8451b9f239f2e7548, we now set a single byte metadata.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -137,91 +155,99 @@ object Features {
...
     val mandatory = 14
   }
 
-  case object BasicMultiPartPayment extends Feature {
+  case object BasicMultiPartPayment extends Feature with InitFeature with NodeFeature with InvoiceFeature {
     val rfcName = "basic_mpp"
     val mandatory = 16
+    val init = true
+    val nodeAnnouncement = true
+    val invoice = true
```
thomash-acinq, line 204: This is no longer needed.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -213,9 +213,12 @@ object OutgoingPaymentPacket {
...
-    val (firstAmount, firstExpiry, payloads) = hops.drop(1).reverse.foldLeft((finalPayload.amount, finalPayload.expiry, Seq[PaymentOnion.PerHopPayload](finalPayload))) {
+    // NB: the final payload will never reach the recipient, since the next-to-last node in the trampoline route will convert that to a non-trampoline payment.
+    // We use the smallest final payload possible, otherwise we may overflow the trampoline onion size.
+    val dummyFinalPayload = PaymentOnion.createSinglePartPayload(finalPayload.amount, finalPayload.expiry, finalPayload.paymentSecret, None)
+    val (firstAmount, firstExpiry, payloads) = hops.drop(1).reverse.foldLeft((finalPayload.amount, finalPayload.expiry, Seq[PaymentOnion.PerHopPayload](dummyFinalPayload))) {
       case ((amount, expiry, payloads), hop) =>
-        // The next-to-last trampoline hop must include invoice data to indicate the conversion to a legacy payment.
+        // The next-to-last node in the trampoline route must receive invoice data to indicate the conversion to a non-trampoline payment.
```
thomash-acinq, line 221: It would be good to have integration tests to check that the payment metadata works well with trampoline.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -71,6 +82,13 @@ case class Features(activated: Map[Feature, FeatureSupport], unknown: Set[Unknow
     unknownFeaturesOk && knownFeaturesOk
   }
 
+  def initFeatures(): Features = Features(activated.collect { case (f: InitFeature, s) => (f: Feature, s) }, unknown)
+
+  def nodeAnnouncementFeatures(): Features = Features(activated.collect { case (f: NodeFeature, s) => (f: Feature, s) }, unknown)
+
+  // NB: we don't include unknown features in invoices, which means plugins cannot inject invoice features.
```
thomash-acinq, line 89: Is there a good reason to not let plugins do that?

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: Good catch, thanks!

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: I'll add a check.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: I think we should allow them to include feature bits in invoices at some point.
But this was already the existing behavior: even before this PR, we never included plugin features in invoices (but it wasn't obvious).
So I added this comment to make it obvious, so that we can fix it in a later PR.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: Actually it's not that simple, this data isn't exposed end-to-end (since we don't want to add it to the payments DB).
Do you have a suggestion on how we could test that without changing too much the internals?

I beefed up a bit the unit test coverage, but I agree it would be good to be able to test it end-to-end (but without modifying the business logic too much).

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/2063/commits/bffbec2c81ac25357da22d1556057d5a7e24e3bb

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: The simplest option I see is to emit a new event, which contains the `paymentMetadata` we receive, but that seems wasteful to emit a new event for every payment received only for tests (but maybe this is ok?).

thomash-acinq: We don't do anything with the metadata but it will need to be exposed somehow to be useful. But for now I guess the existing tests are enough.

thomash-acinq (approved):

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: Yeah it would be nice to be able to test it...

I don't think we will expose it though: the way I'd like to use it is to "replace" what we store in the DB, so when we receive a `payment_metadata` we would decrypt it and extract the values inside, and if they are looking good, we would accept the payment, and otherwise reject it.

If we only do that it will stay "hidden" in our business logic internals. But we probably want applications to also be able to handle "db-less" payments, in that case we would probably have a mechanism for an external app to "register" and listen to a specific `payment_metadata`?

pm47: > The simplest option I see is to emit a new event, which contains the paymentMetadata we receive, but that seems wasteful to emit a new event for every payment received only for tests (but maybe this is ok? I don't actually know how expensive that is).

It's not expensive, and even if it were, you can do it only in the tests by setting your own `paymentHandler`, like we already do in integration tests:
https://github.com/ACINQ/eclair/blob/27579a5786db083700bb668a93eb86583bdf9d74/eclair-core/src/test/scala/fr/acinq/eclair/integration/ChannelIntegrationSpec.scala#L358

thomash-acinq: You're ignoring what happens outside of eclair, if the `payment_metadata` is correct, we should not just accept the payment and be done with it, we should also deliver the item that was bought to the buyer and for that we need the address that would be stored in the `payment_metadata`. Currently we expose the `payment_hash` so that the delivery system (or whatever comes after the payment) can look up the information it needs in its database, if we want to switch to not using a database, we should think about that delivery system too, not just the DB inside eclair.

t-bast: > You're ignoring what happens outside of eclair, if the payment_metadata is correct, we should not just accept the payment and be done with it, we should also deliver the item that was bought to the buyer and for that we need the address that would be stored in the payment_metadata. Currently we expose the payment_hash so that the delivery system (or whatever comes after the payment) can look up the information it needs in its database, if we want to switch to not using a database, we should think about that delivery system too, not just the DB inside eclair.

Yep, that's exactly what I'm saying, we will need to build such a system.
But since we don't have the protocol sketched out at all, we don't know yet how to expose this properly, so we have to wait (and we need to wait for senders to support this feature before we can hope to actually use it, so we have many months before that's close to ready).

t-bast: > you can do it only in the tests by setting your own paymentHandler, like we already do in integration tests:

Good idea, I'll try that!

t-bast: I went for the simplest version possible in https://github.com/ACINQ/eclair/pull/2063/commits/cb2aa019d62aed38c288786de59540b7df3fe580, where I just emit an event. We can rework this when we get closer to implementing features based on `payment_metadata`.

pm47 (approved):

thomash-acinq (approved):

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -114,6 +115,8 @@ object PaymentReceived {
 
 }
 
+case class PaymentMetadataReceived(paymentHash: ByteVector32, paymentMetadata: ByteVector)
```
thomash-acinq, line 118: You can also add it to the `PaymentReceived` defined above instead of creating a new event just for the metadata.

t-bast: I tried, but it is much more invasive...and I'm not sure in the long run we will want it in `PaymentReceived` (which happens after the payment has been accepted and fulfilled), but instead before we accept it.

PR #2066, opened 2021-11-17
Distinct default relay fees if we're funder or fundee

thomash-acinq:


t-bast:
From an economics perspective this makes a lot of sense at first glance: we want to extract more from channels we funded, because we'll pay more on-chain fees.

But that means people will prioritize channels where we're not the funder in their route, and non-strict forwarding will not help use because we won't be able to route these payments through the channels we funded if the ones where we're fundee are depleted and they didn't provide enough fees...so it may create more routing failures which isn't great.

Moreover this has quite an impact on privacy. If we use different values, this lets anyone know whether we're funder or fundee, which means they learn whether the on-chain utxo belonged to us or to our peer. While there are ways an attacker can discover that anyway once the channel is closed (in some cases where the node operator isn't cautious), this change makes it completely trivial.

I'm not sure the trade-off is worth it, what do others think?

t-bast:
What should we do about this PR? Any opinions on my previous comments?

PR #2071, opened 2021-11-19
Use balance estimates from past payments in path-finding

thomash-acinq:
Currently when a payment fails we ban the channel for some time (60 seconds by default). There are several problems with that:
- If we try to relay a huge payment and it fails, we may still be able to relay a smaller payment
- If the ban duration is too short, it still won't be able to relay payments when we unban it. Many channels are unbalanced and inactive and if they are considered attractive by our path-finding algorithm (because they're big and cheap), they'll keep coming back and make our payment attempts fail.
In this PR, I introduce a probabilistic estimation of the channel balance to replace this binary behavior.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
@@ -36,19 +36,79 @@ import scala.concurrent.duration.FiniteDuration
  */
 case class BalanceEstimate private(low: MilliSatoshi, lowTimestamp: TimestampSecond, high: MilliSatoshi, highTimestamp: TimestampSecond, totalCapacity: Satoshi, halfLife: FiniteDuration) {
 
-  def otherSide: BalanceEstimate = BalanceEstimate(totalCapacity - high, highTimestamp, totalCapacity - low, lowTimestamp, totalCapacity, halfLife)
+  /* The goal of this class is to estimate the probability that a given edge can relay the amount that we plan to send
+   * through it. We model this probability with 3 pieces of linear functions.
+   *
+   * Without any information we use the following baseline (x is the amount we're sending and y the probability it can be relayed):
+   *
+   *   1 |****
+   *     |    ****
+   *     |        ****
+   *     |            ****
+   *     |                ****
+   *     |                    ****
+   *     |                        ****
+   *     |                            ****
+   *     |                                ****
+   *     |                                    ****
+   *     |                                        ****
+   *     |                                            ****
+   *   0 +------------------------------------------------****
+   *     0                                                capacity
+   *
+   * If we get the information that the edge can (or can't) relay a given amount (because we tried), then we get a lower
+   * bound (or upper bound) that we can use and our model becomes:
+   *
+   *   1 |***************
+   *     |              |*
+   *     |              | *
+   *     |              | *
+   *     |              |  *
+   *     |              |   *
+   *     |              |   *
+   *     |              |    *
+   *     |              |     *
+   *     |              |     *
+   *     |              |      *
+   *     |              |       *
+   *   0 +--------------|-------|***************************
+   *     0             low     high                      capacity
+   *
+   * However this lower bound (or upper bound) is only valid at the moment we got that information. If we wait, the
+   * information decays and we slowly go back towards our baseline:
+   *
+   *   1 |*****
+   *     |     *****
+   *     |          *****
+   *     |              |*
+   *     |              | *
+   *     |              |  *
+   *     |              |   *
+   *     |              |    *
+   *     |              |     *
+   *     |              |      *
+   *     |              |       **********
+   *     |              |       |         **********
+   *   0 +--------------|-------|-------------------**********
+   *     0             low     high                       capacity
+   */
 
-  /* When we probe an edge, we get certain information on its balance, we know for sure that it can relay at least X or
-   * at most Y. However after some time has passed, other payments may have changed the balance and we can't be so sure
-   * anymore.
-   * We model this decay with a half-life H: every H units of time, our confidence decreases by half and our estimated
+  /**
+   * We model the decay with a half-life H: every H units of time, our confidence decreases by half and our estimated
    * probability distribution gets closer to the baseline uniform distribution of balances between 0 and totalCapacity.
+   *
+   * @param amount                the amount that we knew we could send or not send at time t
+   * @param successProbabilityAtT probability that we could relay amount at time t (usually 0 or 1)
+   * @param t                     time at which we knew if we could or couldn't send amount
+   * @return                      the probability that we can send amount now
    */
   private def decay(amount: MilliSatoshi, successProbabilityAtT: Double, t: TimestampSecond): Double = {
     val decayRatio = 1 / math.pow(2, (TimestampSecond.now() - t) / halfLife)
     val baseline = 1 - amount.toLong.toDouble / toMilliSatoshi(totalCapacity).toLong.toDouble
     baseline * (1 - decayRatio) + successProbabilityAtT * decayRatio
   }
+
+  def otherSide: BalanceEstimate = BalanceEstimate(totalCapacity - high, highTimestamp, totalCapacity - low, lowTimestamp, totalCapacity, halfLife)
 
   def couldNotSend(amount: MilliSatoshi, timestamp: TimestampSecond): BalanceEstimate =
     if (amount < high) {
@@ -79,12 +139,16 @@ case class BalanceEstimate private(low: MilliSatoshi, lowTimestamp: TimestampSec
   def didSend(amount: MilliSatoshi, timestamp: TimestampSecond): BalanceEstimate = {
     val newLow = (low - amount) max MilliSatoshi(0)
     val newHigh = (high - amount) max MilliSatoshi(0)
-    val pLow = decay(newLow, 1, lowTimestamp)
-    val pHigh = decay(newHigh, 0, highTimestamp)
-    if (???) {
-      copy(low = newLow, high = (totalCapacity - amount) max MilliSatoshi(0), highTimestamp = timestamp)
+    /* We could shift everything left by amount without changing the timestamps (a) but we may get more information by
+     * ignoring the old high (b) if if has decayed too much. We try both and choose the one that gives the lowest
+     * probability for high.
+     */
+    val a = copy(low = newLow, high = newHigh)
+    val b = copy(low = newLow, high = (totalCapacity - amount) max MilliSatoshi(0), highTimestamp = timestamp)
+    if (a.canSend(newHigh) < b.canSend(newHigh)) {
+      a
     } else {
-      copy(low = newLow, high = newHigh)
+      b
     }
   }
 
@@ -96,8 +160,8 @@ case class BalanceEstimate private(low: MilliSatoshi, lowTimestamp: TimestampSec
    *
    * We estimate this probability with a piecewise linear function:
    * - probability that it can relay a payment of 0 is 1
-   * - probability that it can relay a payment of low is close to 1 if lowTimestamp is recent
-   * - probability that it can relay a payment of high is close to 0 if highTimestamp is recent
+   * - probability that it can relay a payment of low is decay(low, 1, lowTimestamp) which is close to 1 if lowTimestamp is recent
+   * - probability that it can relay a payment of high is decay(high, 0, highTimestamp) which is close to 0 if highTimestamp is recent
    * - probability that it can relay a payment of totalCapacity is 0
    */
   def canSend(amount: MilliSatoshi): Double = {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
@@ -1,0 +1,170 @@
+/*
+ * Copyright 2021 ACINQ SAS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fr.acinq.eclair.router
+
+import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
+import fr.acinq.bitcoin.scalacompat.{LexicographicalOrdering, Satoshi}
+import fr.acinq.eclair.MilliSatoshi.toMilliSatoshi
+import fr.acinq.eclair.router.Graph.GraphStructure.{DirectedGraph, GraphEdge}
+import fr.acinq.eclair.router.Router.{ChannelHop, PublicChannel}
+import fr.acinq.eclair.{MilliSatoshi, TimestampSecond}
+
+import scala.concurrent.duration.FiniteDuration
+
+/** Estimates the balance between a pair of nodes
+ *
+ * @param low           lower bound on the balance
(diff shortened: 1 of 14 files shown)


t-bast (commented):
I think it's a really good direction to explore! I like the model, I put some comments mostly on what functions it should expose, but the actual logic sounds good to me.

I would scope this PR to only create the balance estimates though, without the path-finding changes. I expect that we will need to iterate on it after seeing how it behaves in practice, before we can start an A/B experiment to actually use it.

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -0,0 +1,170 @@
...
+      }
+    }
+
+  def couldSend(amount: MilliSatoshi, timestamp: TimestampSecond): BalanceEstimate =
+    otherSide.couldNotSend(totalCapacity - amount, timestamp).otherSide
+
+  def didSend(amount: MilliSatoshi, timestamp: TimestampSecond): BalanceEstimate =
+    copy(low = (low - amount) max MilliSatoshi(0), high = (high - amount) max MilliSatoshi(0))
+
+  def addChannel(capacity: Satoshi): BalanceEstimate = copy(high = high + toMilliSatoshi(capacity), totalCapacity = totalCapacity + capacity)
```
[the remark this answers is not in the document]
renepickhardt: For a very long time I was in favor of combining parallel channels for Uncertainty Estimates into one large virtual channel with the sum of the capacities. However this becomes really tricky if these channels have different fee policies and one may wish to include those as an additional feature to the cost function of the min cost flow problem. I finally arrived at the point where I maintain those parallel channels in my data structure. While I belief it messes up the uncertainty prediction in particular as node operators may chose to forward on one of the other channels I think it is a "them" problem and not a "me" problem

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -0,0 +1,182 @@
...
+   *
+   * We estimate this probability with a piecewise linear function:
+   * - probability that it can relay a payment of 0 is 1
+   * - probability that it can relay a payment of low is close to 1 if lowTimestamp is recent
+   * - probability that it can relay a payment of high is close to 0 if highTimestamp is recent
+   * - probability that it can relay a payment of totalCapacity is 0
+   */
+  def canSend(amount: MilliSatoshi): Double = {
+    val x = amount.toLong.toDouble
+    val a = low.toLong.toDouble
```
renepickhardt, line 105: the following formulas are extremly hard to read because of the naming that you introduce. if you want shortcuts (which makes it easyer) why not using the following? 

* a = amount
* l = low
* h = high
* c = capacity

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -0,0 +1,182 @@
...
+import fr.acinq.eclair.MilliSatoshi.toMilliSatoshi
+import fr.acinq.eclair.router.Graph.GraphStructure.{DirectedGraph, GraphEdge}
+import fr.acinq.eclair.router.Router.{ChannelHop, PublicChannel}
+import fr.acinq.eclair.{MilliSatoshi, TimestampSecond}
+
+import scala.concurrent.duration.FiniteDuration
+
+/** Estimates the balance between a pair of nodes
+ *
+ * @param low           lower bound on the balance
```
renepickhardt, line 30: Really like the naming `low` and `high` might overtake this. I used `minLiquidity` and `maxLiquidity` which is a bit more expressive but always a bit of a burden to carry around.

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  /* Estimate the probability that we can successfully send `amount` through the channel
+   *
+   * We estimate this probability with a piecewise linear function:
+   * - probability that it can relay a payment of 0 is 1
+   * - probability that it can relay a payment of low is close to 1 if lowTimestamp is recent
+   * - probability that it can relay a payment of high is close to 0 if highTimestamp is recent
+   * - probability that it can relay a payment of totalCapacity is 0
+   */
+  def canSend(amount: MilliSatoshi): Double = {
```
renepickhardt, line 103: a comment what this function does would have been nice. I initially thought you try to estimate a probability or figure out if you can send a certain amount on a channel. I am now certain you do neither (see comments below) but I still have no idea what is actually happening in this function.

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -0,0 +1,182 @@
...
+    val x = amount.toLong.toDouble
+    val a = low.toLong.toDouble
+    val b = high.toLong.toDouble
+    val c = toMilliSatoshi(totalCapacity).toLong.toDouble
+
+    // Success probability at the low and high points
+    val pLow = decay(low, 1, lowTimestamp)
+    val pHigh = decay(high, 0, highTimestamp)
+
+    if (amount < low) {
```
renepickhardt, line 113: In this case the probability of sending the amount should be 1 as you are certain that you can send the amount!

However taking your notation `a-x` is the amount that can be sent. However I do not understand the semantics of x*pLow and what is acchieved by dividing all by the lower bound of the channel `a ` It certainly does not give you the satoshis that may savely be sent and it most certaintly does not give you an accurate probability. 

I have similar issues understanding the next 5 lines.

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -0,0 +1,182 @@
...
+/** Estimates the balance between a pair of nodes
+ *
+ * @param low           lower bound on the balance
+ * @param lowTimestamp  time at which the lower bound was known to be correct
+ * @param high          upper bound on the balance
+ * @param highTimestamp time at which the upper bound was known to be correct
+ * @param totalCapacity total capacity of all the channels between the pair of nodes
+ * @param halfLife      time after which the certainty of the lower/upper bounds is halved
+ */
+case class BalanceEstimate private(low: MilliSatoshi, lowTimestamp: TimestampSecond, high: MilliSatoshi, highTimestamp: TimestampSecond, totalCapacity: Satoshi, halfLife: FiniteDuration) {
```
renepickhardt, line 37: I think  `LiquidityEstimate` would be a much better name because it is the uncertainty in Liqudity Information that yields a problem. I know in our early work we also called it `Balance` but this term turned out to be very confusing in many conversations which is why I arrived at the point to consistantly call it `Liquidity`. 
That being said if you look at the reference implementation that I did for the stuff that I think you try to achieve here I use the terms `UncertaintyChannel` that quantifies my `belief` about the channels `Liqudity` and estimates probabilities and costs that can be used in routing decisions (in particular for the min cost flow solver) and `UncertaintyNetwork` for the entire network that maintains such information: 

you can find a very will documented notebook in with a glossary at https://github.com/renepickhardt/mpp-splitter/blob/pickhardt-payments-simulation-dev/Pickhardt-Payments-Simulation.ipynb which is currently up for review at: https://github.com/renepickhardt/mpp-splitter/pull/11

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -0,0 +1,182 @@
...
+    }
+  }
+
+  def addChannel(capacity: Satoshi): BalanceEstimate = copy(high = high + toMilliSatoshi(capacity), totalCapacity = totalCapacity + capacity)
+
+  def removeChannel(capacity: Satoshi): BalanceEstimate = copy(low = (low - toMilliSatoshi(capacity)) max MilliSatoshi(0), high = high min toMilliSatoshi(totalCapacity - capacity), totalCapacity = totalCapacity - capacity)
+
+  /* Estimate the probability that we can successfully send `amount` through the channel
+   *
+   * We estimate this probability with a piecewise linear function:
```
renepickhardt, line 97: I agree with the gist / goal here but I don't see this being reflected in the code or understand why this is the comment for the function `removeChannel`

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+object BalanceEstimate {
+  def noChannels(halfLife: FiniteDuration): BalanceEstimate = BalanceEstimate(MilliSatoshi(0), TimestampSecond(0), MilliSatoshi(0), TimestampSecond(0), Satoshi(0), halfLife)
+
+  def baseline(capacity: Satoshi, halfLife: FiniteDuration): BalanceEstimate = noChannels(halfLife).addChannel(capacity)
+}
+
+case class BalancesEstimates(balances: Map[(PublicKey, PublicKey), BalanceEstimate], defaultHalfLife: FiniteDuration) {
```
renepickhardt, line 129: as mentioned above I call this the `UncertaintyNetwork`

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -298,7 +305,7 @@ object Graph {
    * @param weightRatios            ratios used to 'weight' edges when searching for the shortest path
    * @param includeLocalChannelCost if the path is for relaying and we need to include the cost of the local channel
    */
-  private def addEdgeWeight(sender: PublicKey, edge: GraphEdge, prev: RichWeight, currentBlockHeight: BlockHeight, weightRatios: Either[WeightRatios, HeuristicsConstants], includeLocalChannelCost: Boolean): RichWeight = {
+  private def addEdgeWeight(sender: PublicKey, edge: GraphEdge, balance: BalanceEstimate, prev: RichWeight, currentBlockHeight: BlockHeight, weightRatios: Either[WeightRatios, HeuristicsConstants], includeLocalChannelCost: Boolean): RichWeight = {
```
renepickhardt, line 308: assuming the `canSend` function returns the actual conditional probability given your knowledge. The entire thing here is a bit weired to me. IIRC Dijsktra works by minimizing the sum of weights. However an onion is successfull if it is successfull at every hop which means the probabilities along the path have to be multiplied. 
So just factoring in your probability weights from your estimates would again break the probabilistic model. 

Without deeper understanding I initially used the negative logarithm to transform the multiplication of probabilities to an addition that can be used in dijkstra or for min cost flow problems with seperable cost functions. However while creating my reference implementation and tinkering about feature engineering I realized that for a channel of capacity `c` we have `-log(P(X>=c) = log(c+1)` which means that the result is just the entropy of the channel. This is what made me realize that the negative logarithms are not just a neat mathematical trick but rather something that I now call the `uncertainty cost`. This can btw easily be linearized and transered to a unit cost that is comparable to the fee rate of routing. I strongly suggest to look at the glossary of https://github.com/renepickhardt/mpp-splitter/blob/pickhardt-payments-simulation-dev/Pickhardt-Payments-Simulation.ipynb where I explain this all in more detail. A lot of the choices in feature engineering become very natural an clean when thinking about the situation from the perspective of uncertainty costs (or `linearized integer uncertainty unit costs`.

renepickhardt (commented):
I aborted my review as I cannot make any sense of the semantics of the `canSend` method which I guess is the main point where you estimate the actual probability of a payment to be successful given your belief about the network. 

Overall I can state that a lot of the correct ideas and approaches are in here and some of my comments are very detailed and nuanced because of that. 

In between the comments I have linked to my reference implementation of handling the probabilities. I strongly encourage you to read through the glossary (Especially terms like `cost`, `uncertaintay cost`, `linearized integer uncertainty unit cost` ,`unit cost` and `features`) of that note book and the `UncertaintChannel` class itself. Maybe this can help resolve the current misunderstandings that we seem to have. 

What my code does on top of your is to also account for the inflight HTLCs that we currently put on the network due to an ongoing `PaymentSession`. My code currently does not decay the belief about the network and since it is mainly created for educational purpose it does not yet adopt the belief in both directions of the channel if we have updates (I should probably do this rather sooner than later)

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
thomash-acinq: This is unrelated to this PR, I've fixed it in #2257
However I'm not convinced the fix is actually useful in practice. By using 1/p we can give a cost to failed attempts which is something concrete, when using -log(p) we fix Dijkstra but the penalty factor does not correspond to anything we care about.

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -0,0 +1,182 @@
...
+  /* When we probe an edge, we get certain information on its balance, we know for sure that it can relay at least X or
+   * at most Y. However after some time has passed, other payments may have changed the balance and we can't be so sure
+   * anymore.
+   * We model this decay with a half-life H: every H units of time, our confidence decreases by half and our estimated
+   * probability distribution gets closer to the baseline uniform distribution of balances between 0 and totalCapacity.
+   */
+  private def decay(amount: MilliSatoshi, successProbabilityAtT: Double, t: TimestampSecond): Double = {
+    val decayRatio = 1 / math.pow(2, (TimestampSecond.now() - t) / halfLife)
+    val baseline = 1 - amount.toLong.toDouble / toMilliSatoshi(totalCapacity).toLong.toDouble
+    baseline * (1 - decayRatio) + successProbabilityAtT * decayRatio
```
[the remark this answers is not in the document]
thomash-acinq: Decaying `low` towards 0 doesn't make sense. At a time t we have a given probability distribution for the balance and decaying going back to the prior (which is our uniform baseline), that's why I'm averaging the distribution at time t and the baseline.

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
thomash-acinq: No.
`low` was the lower bound at some time t. If you call `canSend` at that time t, then yes the result will be 1 but if call it later it will be lower than 1.

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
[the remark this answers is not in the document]
t-bast: > However this becomes really tricky if these channels have different fee policies

Do note that it doesn't make any sense IMO to have difference fee policies on different channels with the same peer (because non-strict forwarding)

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -61,11 +61,12 @@ object Graph {
...
-   * @param lockedFundsRisk cost of having funds locked in htlc in msat per msat per block
-   * @param failureCost     fee for a failed attempt
-   * @param hopCost         virtual fee per hop (how much we're willing to pay to make the route one hop shorter)
+   * @param lockedFundsRisk   cost of having funds locked in htlc in msat per msat per block
+   * @param failureCost       fee for a failed attempt
+   * @param hopCost           virtual fee per hop (how much we're willing to pay to make the route one hop shorter)
+   * @param usePastRelaysData use data from past relays to estimate the balance of the channels
    */
-  case class HeuristicsConstants(lockedFundsRisk: Double, failureCost: RelayFees, hopCost: RelayFees)
+  case class HeuristicsConstants(lockedFundsRisk: Double, failureCost: RelayFees, hopCost: RelayFees, usePastRelaysData: Boolean)
```
t-bast, line 69: Why isn't this called `useBalanceEstimates`? It would be more coherent, wouldn't it?

t-bast:
Since we merged #2272, we can probably close this PR for now? We'll gather mainnet data from our node and will later submit a new PR to use the balance estimation once our model seems to be accurate.

t-bast:
That won't be lost, you just have to keep the branch...?

All of the comments made on the balance estimation part will be obsolete once you rebase, making it unnecessarily harder for people to read through the PR. It feels better to me to start from scratch and have a new PR focused on only _using_ the balance estimation to keep it focused?

PR #2073, opened 2021-11-23
Add channel type feature bit

t-bast:
We already support channel types, but we make it explicit with a feature bit as required by https://github.com/lightning/bolts/pull/906


pm47 (commented):
Do I understand correctly that we don't use the feature bit in our code because we will always set the tlv? But then:
- shouldn't we add a check at startup that forces the feature bit to be enabled? 
- shouldn't we check the remote feature bit and reject if the feature bit is set and there is no corresponding tlv?

t-bast:
> Do I understand correctly that we don't use the feature bit in our code because we will always set the tlv?

Yes, that is correct, because the spec allows it and it's an odd tlv.
It's less code and less branches to test to always include it rather than check for the feature bit.
LND insisted to add this feature bit to the spec to eventually be able to only connect to nodes that support the feature.

> shouldn't we add a check at startup that forces the feature bit to be enabled?

We could indeed, I'll add that.

> shouldn't we check the remote feature bit and reject if the feature bit is set and there is no corresponding tlv?

We could. I thought it would better to be lenient here because we automatically fallback to the most probable channel type based on feature bits, which will very likely work, but I can change that.

t-bast:
Done in https://github.com/ACINQ/eclair/pull/2073/commits/42e15edce3ffaed2b43782bef53dd32744d521ea and https://github.com/ACINQ/eclair/pull/2073/commits/0bd5017169f72f606c801a1aea6145f0178d316b

pm47 (approved):
LGTM

Issue #2070, opened 2021-11-19
Channel failure due to unknown issue with eclair


viaj3ro:
I had a channel failure recently which according to Anton from BLW seems to be caused by an unknown issue with eclair. Diskussion here: https://t.me/eclair_lightning

```
2021-11-09 07:18:23,908 INFO  f.a.e.b.b.ZmqWatcher - triggering WatchFundingSpent(Actor[akka://eclair-node/user/$h/switchboard/peer-03ee58475055820fbfa52e356a8920f62f8316129c39369dbdde3e5d0198a9e315/$a#-996100026],62c8129b56bc0622ffcd26cb058e49e8691003fffabf35977f1df0c44ec905b4,1,Set(54d6ea4e322496d7c222583af1012be2c656f95395683fd2fcee9a26f133026a, c3edafc8d79833437890ec68b3c2a72c367949c15ea7f455709e9d186797e758))
2021-11-09 07:18:23,947 WARN  f.a.e.channel.Channel n:03ee58475055820fbfa52e356a8920f62f8316129c39369dbdde3e5d0198a9e315 c:b405c94ec4f01d7f9735bffaff031069e8498e05cb26cdff2206bc569b12c863 - they published their current commit in txid=0d2eb24a2e60f1a6143db02aef7ec517351363be6f082585395a3872cf9f5b0b
2021-11-09 07:18:23,948 INFO  f.a.e.db.DbEventHandler n:03ee58475055820fbfa52e356a8920f62f8316129c39369dbdde3e5d0198a9e315 c:b405c94ec4f01d7f9735bffaff031069e8498e05cb26cdff2206bc569b12c863 - paying mining fee=0 sat for txid=0d2eb24a2e60f1a6143db02aef7ec517351363be6f082585395a3872cf9f5b0b desc=remote-commit
[...]

PR #2076, opened 2021-11-26
Remove misleading claim-htlc-success log

t-bast:
Whenever a remote commitment is confirmed in which we received htlcs, we generate a claim-htlc-success transaction even if we don't have the preimage yet (so we can't broadcast it).

This created a confusing log line:

```log
tx generation success: desc=claim-htlc-success
```

We want to create that log line only when we have the preimage and the node operator can expect the HTLC to be redeemed on-chain.

This confusion happened in the investigation of #2070


pm47 (approved):
LGTM 💯

PR #2079, opened 2021-11-29
Rename feeThresholdSat to maxFeeFlatSat

thomash-acinq:
The new name should be much easier to understand and matches the name used in eclair-core


t-bast (commented):
Good idea, this name was really confusing.

We are losing the unit with your proposal though, it's not obvious for most users whether we'll expect satoshis or millisatoshis here. It's not very elegant, but we've chosen for now to include the units in the API parameter names, I think it's a good idea to keep that, so what about one of these:

* `maxFeeSat` / `maxFeePct`
* `maxFeeSatoshis` / `maxFeePct`
* `maxFeeFlatSat` / `maxFeePct`
* `maxFeeFlatSatoshis` / `maxFeePct`

t-bast (approved):

thomash-acinq:
You're right, I just copied the name that we have in the code but there we have types to encode the unit.

PR #2080, opened 2021-12-01
Abort HTLC tx publisher if remote commit confirms

t-bast:
If the remote commit confirms before our local commit, there is no reason to try to publish our HTLC transactions, we will instead directly claim the htlc outputs from the remote commit.

We previously checked timelocks before checking preconditions, which in this case means we would be waiting for a confirmation on our local commit forever. We now check preconditions before timelocks, and added a precondition that verifies that the remote commit isn't confirmed before publishing our HTLC txs.

It wasn't a big issue, we just had a watch that would linger forever and a few idle actors (until the channel closes and the channel actors stops, which stops all its publisher children), but we can easily avoid it without making the code more complex.


pm47 (approved):
LGTM

PR #2082, opened 2021-12-02
Support private channels in SendToRoute

### LINKED ISSUE #2081: [Bug] SendToRoute doesn't allow to send via private channels
Trying to use the `SendToRoute` API to send via a private channel throws this error: 

```
router error: The sequence of channels provided cannot be used to build a route to the target node
```

t-bast:
For some reason, when sending to a pre-defined channel route, we only looked at the public channel graph. This was incorrect, we should also include private channels and routing hints.

Thanks @viaj3ro for reporting this, definitely a bug in my code.

Fixes #2081

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -34,7 +34,6 @@ import kamon.tag.TagSet
 
 import scala.annotation.tailrec
 import scala.collection.mutable
-import scala.concurrent.duration._
 import scala.util.{Failure, Random, Success, Try}
 
 object RouteCalculation {
@@ -69,11 +68,19 @@ object RouteCalculation {
         case PredefinedChannelRoute(targetNodeId, channels) =>
           val (end, hops) = channels.foldLeft((localNodeId, Seq.empty[ChannelHop])) {
             case ((start, current), shortChannelId) =>
-              d.channels.get(shortChannelId).flatMap(c => start match {
-                case c.ann.nodeId1 => g.getEdge(ChannelDesc(shortChannelId, c.ann.nodeId1, c.ann.nodeId2))
-                case c.ann.nodeId2 => g.getEdge(ChannelDesc(shortChannelId, c.ann.nodeId2, c.ann.nodeId1))
+              val channelDesc_opt = d.channels.get(shortChannelId).flatMap(c => start match {
+                case c.ann.nodeId1 => Some(ChannelDesc(shortChannelId, c.ann.nodeId1, c.ann.nodeId2))
+                case c.ann.nodeId2 => Some(ChannelDesc(shortChannelId, c.ann.nodeId2, c.ann.nodeId1))
                 case _ => None
-              }) match {
+              }).orElse(d.privateChannels.get(shortChannelId).flatMap(c => start match {
+                case c.nodeId1 => Some(ChannelDesc(shortChannelId, c.nodeId1, c.nodeId2))
+                case c.nodeId2 => Some(ChannelDesc(shortChannelId, c.nodeId2, c.nodeId1))
+                case _ => None
+              })).orElse(assistedChannels.get(shortChannelId).flatMap(c => start match {
+                case c.extraHop.nodeId => Some(ChannelDesc(shortChannelId, c.extraHop.nodeId, c.nextNodeId))
+                case _ => None
+              }))
+              channelDesc_opt.flatMap(c => g.getEdge(c)) match {
                 case Some(edge) => (edge.desc.b, current :+ ChannelHop(edge.desc.a, edge.desc.b, edge.update))
                 case None => (start, current)
               }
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -69,11 +68,19 @@ object RouteCalculation {
...
                 case _ => None
-              }) match {
+              }).orElse(d.privateChannels.get(shortChannelId).flatMap(c => start match {
+                case c.nodeId1 => Some(ChannelDesc(shortChannelId, c.nodeId1, c.nodeId2))
+                case c.nodeId2 => Some(ChannelDesc(shortChannelId, c.nodeId2, c.nodeId1))
+                case _ => None
+              })).orElse(assistedChannels.get(shortChannelId).flatMap(c => start match {
+                case c.extraHop.nodeId => Some(ChannelDesc(shortChannelId, c.extraHop.nodeId, c.nextNodeId))
+                case _ => None
+              }))
```
thomash-acinq, line 82:
```suggestion
              ).orElse(d.privateChannels.get(shortChannelId).flatMap(c =>
                if (start == c.nodeId1) Some(ChannelDesc(shortChannelId, c.nodeId1, c.nodeId2))
                else if (start == c.nodeId2) Some(ChannelDesc(shortChannelId, c.nodeId2, c.nodeId1))
                else None
              )).orElse(assistedChannels.get(shortChannelId).flatMap(c =>
                if(start == c.extraHop.nodeId) Some(ChannelDesc(shortChannelId, c.extraHop.nodeId, c.nextNodeId))
                else None
              ))
```

eclair-core/src/test/scala/fr/acinq/eclair/router/RouterSpec.scala
```diff
@@ -514,9 +514,72 @@ class RouterSpec extends BaseRouterSpec {
...
+      sender.send(router, FinalizeRoute(10000 msat, preComputedRoute))
+      val response = sender.expectMsgType[RouteResponse]
+      assert(response.routes.length === 1)
+      val route = response.routes.head
+      assert(route.hops.map(_.lastUpdate) === Seq(update_ag))
+      assert(route.hops.head.nodeId === a)
+      assert(route.hops.head.nextNodeId === g)
+    }
+    {
+      val preComputedRoute = PredefinedChannelRoute(h, Seq(channelId_ag, channelId_gh))
```
thomash-acinq, line 537: Are `channelId_ag` and `channelId_gh` private channels? There is no way to tell without digging into other files, maybe renaming `*_ag` to `*_ag_private` would make i more obvious.

thomash-acinq (approved):
Good fix, I have suggested what I think are additional improvements

eclair-core/src/test/scala/fr/acinq/eclair/router/RouterSpec.scala
t-bast: I agree, I had to go to inside the `BaseRouterSpec` to find a private channel, I'll rename that.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -69,11 +68,19 @@ object RouteCalculation {
         case PredefinedChannelRoute(targetNodeId, channels) =>
           val (end, hops) = channels.foldLeft((localNodeId, Seq.empty[ChannelHop])) {
             case ((start, current), shortChannelId) =>
-              d.channels.get(shortChannelId).flatMap(c => start match {
-                case c.ann.nodeId1 => g.getEdge(ChannelDesc(shortChannelId, c.ann.nodeId1, c.ann.nodeId2))
-                case c.ann.nodeId2 => g.getEdge(ChannelDesc(shortChannelId, c.ann.nodeId2, c.ann.nodeId1))
+              val channelDesc_opt = d.channels.get(shortChannelId).flatMap(c => start match {
+                case c.ann.nodeId1 => Some(ChannelDesc(shortChannelId, c.ann.nodeId1, c.ann.nodeId2))
+                case c.ann.nodeId2 => Some(ChannelDesc(shortChannelId, c.ann.nodeId2, c.ann.nodeId1))
                 case _ => None
```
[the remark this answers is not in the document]
t-bast: We tend to use pattern matching a lot in that kind of scenario, I personally find it slightly more readable than cascading ifs (but it's not a strong feeling).

When I first started working on eclair, I wasn't really used to it but just followed it for consistency since that was what the rest of the codebase tended to do, and now my brain got really used to reading those :nerd_face:

eclair-core/src/test/scala/fr/acinq/eclair/router/RouterSpec.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2082/commits/d930bf1777e845773447e1664f0059cc32f18bfc

thomash-acinq (approved):

PR #2086, opened 2021-12-06
No error when updating the relay fees not in normal state

### LINKED ISSUE #2085: Feature Request
I would like to update the fees on a channel before it has completely opened. Currently, this requires that I edit the config file before I issue the open command because using updateRelayFee returns `"RES_FAILURE(CMD_UPDATE_RELAY_FEE( ... ),fr.acinq.eclair.channel.CommandUnavailableInThisState: cannot execute command=updaterelayfee in state=WAIT_FOR_FUNDING_CONFIRMED)`

What problems need to be solved so users can revise the record of what the base and rate-per-million fees should be for the node with which a channel is being opened?

thomash-acinq:
When using the API to update the relay fees of a channel, the changes are saved in the DB and then the channel is updated. If the channel can't be updated, because it is not ready yet for instance, that's fine because it will use the values from the DB as soon as it is ready. So there is no need to return an error in that case.
Fixes #2085

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1888,7 +1888,10 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
         case _ => handleCommandError(CommandUnavailableInThisState(d.channelId, "forceclose", stateName), c)
       }
 
-    case Event(c: CMD_UPDATE_RELAY_FEE, d) => handleCommandError(CommandUnavailableInThisState(d.channelId, "updaterelayfee", stateName), c)
+    case Event(c: CMD_UPDATE_RELAY_FEE, d) =>
+      val replyTo = if (c.replyTo == ActorRef.noSender) sender() else c.replyTo
+      replyTo ! RES_SUCCESS(c, d.channelId)
+      stay()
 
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1888,7 +1888,10 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
         case _ => handleCommandError(CommandUnavailableInThisState(d.channelId, "forceclose", stateName), c)
       }
 
-    case Event(c: CMD_UPDATE_RELAY_FEE, d) => handleCommandError(CommandUnavailableInThisState(d.channelId, "updaterelayfee", stateName), c)
+    case Event(c: CMD_UPDATE_RELAY_FEE, d) =>
```
t-bast, line 1891: Can you add a comment here? Something like:

```suggestion
    // In states where we don't explicitly handle this command, we won't broadcast a new channel update immediately,
    // but we will once we get back to NORMAL, because the updated fees have been saved to our peers DB.
    case Event(c: CMD_UPDATE_RELAY_FEE, d) =>
```

thomash-acinq: Done

t-bast (approved):
Thanks, LGTM :+1:

PR #2089, opened 2021-12-08
Make relaying onion messages optional

thomash-acinq:


eclair-core/src/main/resources/reference.conf
```diff
@@ -347,6 +347,14 @@ eclair {
       "mempool.space"
     ]
   }
+
+  onion-messages {
+    # Valid values are
+    # - no-relay: Allow sending and receiving onion messages but never relay them
+    # - no-new-connections: Relay messages to peers that have channel with us or are connected to us but do not initiate connections to unknown peers
+    # - relay-all: Relay everything and create new connections if necessary
```
t-bast, line 355: I'm not sure we should offer `no-relay`, because if we advertise the feature, we should relay onion messages (we can throttle them, but we should try to at least relay some of them, especially when we have channels with the peers we're asked to relay to).

Can you explain why you proposed `no-new-connections` instead of something like `relay-channels` (only relay if we have a channel with the next node)? I feel like it's better to restrict to peers we have channels with instead of connected peers, because we can't prevent malicious peers from connecting to us and then having us relay onion messages to them, whereas if we impose having a channel, it has a cost for them.

IMHO the default configuration should probably be `relay-channels`, and it should be a conscious choice to instead switch to `relay-all` (mostly if you're an LSP and want to help your users).

t-bast (commented):
Let's agree on the default relay policies first, then I think the code should be simple (but it does need tests!).

eclair-core/src/main/resources/reference.conf
thomash-acinq: A node may want to be able to receive messages without wanting to relay them, `no-relay` is for that, I can people that are worried about spam using that.

For `no-new-connections` I don't understand your concern. Any node can connect to us and then we will relay to them which is exactly the point so that they can receive a reply to their message, if not we would never forward the reply. Why would a malicious node want us to relay onion message to them and would we want to prevent that?

We can add `relay-channels` if you want but I don't really see why you want to use it.

thomash-acinq: I guess another useful policy would be to consider the origin and only relay messages coming from a peer with which we have a channel.

t-bast: > A node may want to be able to receive messages without wanting to relay them, no-relay is for that, I can see people that are worried about spam using that.

Let's keep it then, but I think it really shouldn't be the default one.

> I guess another useful policy would be to consider the origin and only relay messages coming from a peer with which we have a channel.

Less is more: let's avoid introducing too many policies that no-one will use, it introduces unnecessary complexity. It's better to have only a handful of them, and if a new one is really necessary, users will ask for it.

> For no-new-connections I don't understand your concern. Any node can connect to us and then we will relay to them which is exactly the point so that they can receive a reply to their message, if not we would never forward the reply. Why would a malicious node want us to relay onion message to them and why would we want to prevent that?

The way I see it, from a node operator's perspective, I would like to be able to choose between 3 rough policies:

- don't relay anything (`no-relay`): I'm cheap and only want to receive
- relay everything (`relay-all`): I want to help the network as much as I can and have the resources to do so
- relay only what doesn't expose my node to DoS: the safe middle ground between the other two

Usually to achieve the third policy, we simply restrict it to peers with whom we have a channel: this guarantees that there is a cost to the connection, which should mitigate DoS risk (that's also a policy we will apply for gossip in the long run). That's why I think this third policy should be `relay-channels`, where you would only relay onion messages to peers you have a channel with if it was coming from a peer with whom you have a channel (note that I apply the rule for both incoming and outgoing peer, I think that's important).

Using `no-new-connections` instead feels like taking a risk for no good reason: it creates a bigger attack surface. Maybe it's hard to exploit, maybe it's not, but why take the risk when we know `relay-channels` is safer? For example, an attacker could have a big botnet available (lots of machine, even if they're not powerful). All these machines can run an empty lightning node (without any channel), connect to us, and ask us to relay onion messages, and we end up relaying a ton of them. With `relay-channels` this attack isn't possible. Maybe it's far-fetched, but a motivated attacker will probably find better attacks, so I'd err on the side of caution. Does that make sense?

thomash-acinq: Checking if we have a channel with the destination peer adds a lot of complexity.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -490,4 +499,12 @@ object Peer {
...
       initFeatures = initFeatures)
   }
+
+  sealed trait RelayPolicy
+
+  case object NoRelay extends RelayPolicy
+
+  case object RelayChannelsOnly extends RelayPolicy
+
+  case object RelayAll extends RelayPolicy
```
t-bast, line 509: I think this would be better inside `MessageRelay` (or rename it `OnionMessageRelayPolicy` if you'd rather keep it here)?
I guess you put it in `Peer` to potentially later reuse them for gossip policies, but I think it will be confusing as gossip won't have exactly the same behavior as onion messages, so it's probably better to introduce a slightly different set of relay policies for gossip later.

t-bast (commented):
> Checking if we have a channel with the destination peer adds a lot of complexity.

I think that this complexity only appears because you're trying to put this logic inside `MessageRelay`, but I think it becomes much simpler if we add this logic inside `Peer` and `Switchboard` instead, because they have access to the data needed.

I think the high-level logic should be:

- The peer actor receives an incoming onion message:
  - if `no-relay` -> just drop it
  - if `channels-only` and we don't have channels -> drop it
  - otherwise tell the switchboard that we should _probably_ relay this message
- The switchboard receives a request to relay an onion message:
  - if `no-relay` -> just drop it
  - if `channels-only` -> if we already have a peer actor for the remote node, forward it the request, otherwise just drop it (if we had channels, we would have a live peer actor)
  - if `relay-all` -> create or get peer and forward it the request
- The peer actor receives a request for relay an onion message:
  - if `no-relay` -> just drop it
  - if `channels-only` and we don't have channels -> drop it
  - otherwise send it

Do you think that makes sense?
I believe it wouldn't require a lot of code to implement, would it?
I'll double-check that the `Peer` has all the data required for that.

thomash-acinq:
>     * The `peer` actor receives a request for relay an onion message

I think it only makes it more complicated to involve the `Peer` in this.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala

thomash-acinq:
I've reorganized everything to go through the `Peer`.

PR #2090, opened 2021-12-08
Rate limit onion messages

thomash-acinq:
The rate limit is applied on the destination side as it is too easy to cheat on the origin side.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -356,6 +356,8 @@ eclair {
 
     # Transient connections opened to relay messages will be closed after this delay of inactivity
     kill-transient-connection-after = 30 seconds
+
+    max-per-peer-per-second = 10
   }
 }
 
--- a/eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
@@ -25,7 +25,7 @@ import fr.acinq.eclair.Features.{BasicMultiPartPayment, ChannelRangeQueries, Pay
 import fr.acinq.eclair.TestConstants._
 import fr.acinq.eclair._
 import fr.acinq.eclair.crypto.TransportHandler
-import fr.acinq.eclair.message.OnionMessages.{IntermediateNode, Recipient, buildMessage}
+import fr.acinq.eclair.message.OnionMessages.{Recipient, buildMessage}
 import fr.acinq.eclair.router.Router._
 import fr.acinq.eclair.router.RoutingSyncSpec
 import fr.acinq.eclair.wire.protocol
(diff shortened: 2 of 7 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -347,6 +347,10 @@ eclair {
       "mempool.space"
     ]
   }
+
+  onion-messages {
+    max-per-peer-per-second = 10
```
t-bast, line 352: I think it will be worth rebasing this PR on top of #2089 and create a new `OnionMessagesConfig` object in `NodeParams` that will reflect all the fields we'll put in here (more future-proof than adding individual fields to `NodeParams` every time).

t-bast (commented):
This needs tests, particularly since we depend on an external component here, we must thoroughly test it to ensure it behaves as we expect.

t-bast:
> The rate limit is applied on the destination side as it is too easy to cheat on the origin side.

Can you explain why you don't apply it in both directions? It feels to me that it would be useful.

Otherwise an attacker connects to you, and as long as they split the load between N outgoing peers, they are able to send 10 * N incoming onion messages without being rate-limited, right? I believe we'd like to protect against that.

t-bast:
It would also be useful to have metrics related to onion messages: recording how many onion messages we receive and send per second and how many onion messages we throttle per second would be interesting monitoring data.

thomash-acinq:
I've added rate limiting in both directions, metrics and tests.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
```diff
@@ -421,5 +421,35 @@ class PeerConnectionSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike wi
       case _ => fail()
     }
   }
+
+  test("incoming rate limiting") { f =>
+    import f._
+    connect(nodeParams, remoteNodeId, switchboard, router, connection, transport, peerConnection, peer, isPersistent = false)
+    val (_, message) = buildMessage(randomKey(), randomKey(), Nil, Left(Recipient(nodeParams.nodeId, None)), Nil)
+    for (_ <- 1 to 11) {
```
t-bast, line 429: Why isn't this `1 to 10` if we allow 10 per second?
I guess this is because the rate limiter's behavior isn't entirely predictable, is it?
This can make it hard to test if we want to be really precise, maybe we should keep the tests more vague.
For example, we could do the following:

- override the throttling to 1 per second
- send 5 messages in a loop: verify that at least one went through but not all of them
- sleep 1 second
- send a message and verify it goes through

This provides less strong guarantees but ensures that rate-limiting "roughly works", and I'm not sure we'll be able to get much more from black-box testing if we don't want our tests to randomly fail.

thomash-acinq: Done

t-bast (approved):
LGTM :100:

PR #2091, opened 2021-12-08
Write received onion messages to the websocket

thomash-acinq:


eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
```diff
@@ -402,6 +405,33 @@ object GlobalBalanceSerializer extends MinimalSerializer({
     JObject(JField("total", JDecimal(o.total.toDouble))) merge Extraction.decompose(o)(formats)
 })
 
+object OnionMessageReceivedSerializer extends MinimalSerializer({
```
t-bast, line 408: pm47 introduced a new `ConvertClassSerializer` which should be used for new JSON serializers (we should definitely do some clean-up on the old serializers and migrate them to this new mechanism).

The basic idea is to convert an `OnionMessages.ReceiveMessage` to a new `data class OnionMessageReceivedJson()` that would only contain fields which are themselves serializable.

Can you try doing this for onion messages?

thomash-acinq: Done

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -1168,6 +1170,18 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
...
         system.eventStream.publish(chcl)
         wsClient.expectMessage(expectedSerializedChcl)
+
+        val msgrcv = OnionMessages.ReceiveMessage(MessageOnion.FinalPayload(TlvStream[OnionMessagePayloadTlv](
+          Seq(
+            OnionMessagePayloadTlv.EncryptedData(ByteVector.empty),
+            OnionMessagePayloadTlv.ReplyPath(Sphinx.RouteBlinding.create(PrivateKey(hex"414141414141414141414141414141414141414141414141414141414141414101"), Seq(bobNodeId), Seq(hex"000000")))
+          ), Seq(
+            GenericTlv(UInt64(5), hex"1111")
+          ))), None)
```
t-bast, line 1180: Can you add a `path_id` so that we test its serialization as well?

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
```diff
@@ -402,6 +404,11 @@ object GlobalBalanceSerializer extends MinimalSerializer({
     JObject(JField("total", JDecimal(o.total.toDouble))) merge Extraction.decompose(o)(formats)
 })
 
+private[json] case class MessageReceivedJson(pathId: Option[ByteVector], replyPath: Option[BlindedRoute], unknownTlvs: Seq[GenericTlv])
```
t-bast, line 407: It's a nit, but I'm mentioning it now because serialization is difficult to change afterwards since it would break backwards compatibility: what about using a map for the unknown tlvs?

```suggestion
private[json] case class MessageReceivedJson(pathId: Option[ByteVector], replyPath: Option[BlindedRoute], unknownTlvs: Map[UInt64, ByteVector])
```

It feels slightly better for applications using it, they will usually be interested in reading the contents of a specific custom tlv tag.

t-bast (commented):
Looks good, a couple of nits.

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
thomash-acinq: Done

t-bast (approved):

PR #2092, opened 2021-12-08
Disable gossip for transient peers

thomash-acinq:
When relaying onion messages, we may connect to new peers for a short period but we don't want to listen to their gossip or gossip to them.
If a channel is later established with that peer, then we enable the gossip.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -78,7 +78,7 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
       }
       context.watch(transport)
       startSingleTimer(AUTH_TIMER, AuthTimeout, conf.authTimeout)
-      goto(AUTHENTICATING) using AuthenticatingData(p, transport)
+      goto(AUTHENTICATING) using AuthenticatingData(p, transport, p.enableGossip)
   }
 
   when(AUTHENTICATING) {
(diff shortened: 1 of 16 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -162,7 +164,13 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
...
       case Event(msg: LightningMessage, d: ConnectedData) if sender() != d.transport => // if the message doesn't originate from the transport, it is an outgoing message
         d.transport forward msg
-        stay()
+        msg match {
+          // If we send any channel management message to this peer, we enable the gossip.
+          case _ : ChannelMessage if !d.enableGossip =>
+            context.system.eventStream.subscribe(self, classOf[Rebroadcast])
+            stay() using d.copy(enableGossip = true)
+          case _ => stay()
+        }
```
t-bast, line 173: This is clever, I initially thought we would need to do it in the `Peer`, which isn't great because that would require sending messages from `PeerConnection` (potentially on front machines) to `Peer` (on backend machines). I'm glad you thought of this trick, it's better to do this here.

However matching on any `ChannelMessage` is too broad and easily gameable by our peer: they could send us a dummy `open_channel` and not follow through, and we'd enable the gossip. I think we should instead match on `channel_reestablish` and `funding_locked` to enable the gossip:

- if we send an outgoing `channel_reestablish`, it means we previously had a channel
- if we send an outgoing `funding_locked`, it means that we've seen a funding tx broadcast, so there really is a new channel being created

t-bast (commented):
This is a very sensitive and subtle part of the codebase, I'd like @pm47 to also review it.
This really needs some tests on the new behavior.

I'm also afraid that the mechanism to enable sync when a channel is created simply doesn't work...because the `DoSync` message may be handled before you switch `enableGossip` to `true`. In that case we will send out gossip queries, but we will drop the responses, which will completely break our gossip sync state machine. We probably need a more holistic approach where we treat both incoming and outgoing gossip similarly: maybe we should only enable gossip when we receive the `DoSync` request? And I'm noticing that we trigger this on `ChannelIdAssigned`, but we should actually do it once the channel is confirmed.

It makes me think that we may want to change our expectations, and we should first agree on exactly what we want to achieve before trying to implement it. Should we change the scope of this PR to introduce gossip relay policies, that would be configurable by the node operator? If so, what relay policies can we introduce that are useful for node operators and easy enough to implement? I think we'll need such a mechanism in the future, but it may be complex to implement and not a priority right now.

t-bast:
Shall we close this PR for now? We've discussed postponing this to work on a general gossip anti-DoS mechanism later.

PR #2094, opened 2021-12-09
Remove network stats computation

t-bast:
We introduced a while ago a task to regularly compute network statistics (mostly about channel parameters such as expiry and fees). The goal was to use this information in the MPP split algorithm to decide whether to split a payment or not, and if we were splitting, what size we should use for the parts.

But we have never used it, and I'm not sure anymore that it's useful at all...we should probably have introduced it only once we actually had a splitting algorithm that really needed it.

We used to expose these stats to node operators via the API, which I'm removing. If node operators are interested in network stats, an ad-hoc on-the-fly computation would make more sense. Let's see if people are complaining about this or not, AFAIK nobody depends on it.


thomash-acinq (dismissed):

t-bast:
Rebased to fix merge conflict.

thomash-acinq (approved):

PR #2096, opened 2021-12-09
Kill idle peers

thomash-acinq:
We may create connections to new peers for relaying onion messages but we don't want to keep them open for long.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -365,8 +365,12 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
         log.info(s"resuming processing of network announcements for peer")
         stay() using d.copy(behavior = d.behavior.copy(fundingTxAlreadySpentCount = 0, ignoreNetworkAnnouncement = false))
 
-      case Event(KillIdle, d: ConnectedData) if !d.isPersistent =>
-        stop(FSM.Normal)
+      case Event(KillIdle, d: ConnectedData) =>
+        if (!d.isPersistent) {
+          stop(FSM.Normal)
+        } else {
+          stay()
+        }
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
@@ -78,7 +78,7 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
       }
       context.watch(transport)
       startSingleTimer(AUTH_TIMER, AuthTimeout, conf.authTimeout)
-      goto(AUTHENTICATING) using AuthenticatingData(p, transport)
+      goto(AUTHENTICATING) using AuthenticatingData(p, transport, p.isPersistent)
   }
 
   when(AUTHENTICATING) {
@@ -88,7 +88,7 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
       log.info(s"connection authenticated with $remoteNodeId@${address.getHostString}:${address.getPort} direction=${if (d.pendingAuth.outgoing) "outgoing" else "incoming"}")
       Metrics.PeerConnectionsConnecting.withTag(Tags.ConnectionState, Tags.ConnectionStates.Authenticated).increment()
       switchboard ! Authenticated(self, remoteNodeId)
-      goto(BEFORE_INIT) using BeforeInitData(remoteNodeId, d.pendingAuth, d.transport)
+      goto(BEFORE_INIT) using BeforeInitData(remoteNodeId, d.pendingAuth, d.transport, d.isPersistent)
 
     case Event(AuthTimeout, d: AuthenticatingData) =>
       log.warning(s"authentication timed out after ${conf.authTimeout}")
@@ -104,7 +104,7 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
       val localInit = protocol.Init(localFeatures, TlvStream(InitTlv.Networks(chainHash :: Nil)))
       d.transport ! localInit
       startSingleTimer(INIT_TIMER, InitTimeout, conf.initTimeout)
-      goto(INITIALIZING) using InitializingData(chainHash, d.pendingAuth, d.remoteNodeId, d.transport, peer, localInit, doSync)
+      goto(INITIALIZING) using InitializingData(chainHash, d.pendingAuth, d.remoteNodeId, d.transport, peer, localInit, doSync, d.isPersistent)
   }
 
   when(INITIALIZING) {
@@ -148,7 +148,7 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
           log.info(s"rebroadcast will be delayed by $rebroadcastDelay")
           context.system.eventStream.subscribe(self, classOf[Rebroadcast])
 
-          goto(CONNECTED) using ConnectedData(d.chainHash, d.remoteNodeId, d.transport, d.peer, d.localInit, remoteInit, rebroadcastDelay)
+          goto(CONNECTED) using ConnectedData(d.chainHash, d.remoteNodeId, d.transport, d.peer, d.localInit, remoteInit, rebroadcastDelay, isPersistent = d.isPersistent)
         }
 
       case Event(InitTimeout, d: InitializingData) =>
@@ -162,7 +162,15 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
     heartbeat {
       case Event(msg: LightningMessage, d: ConnectedData) if sender() != d.transport => // if the message doesn't originate from the transport, it is an outgoing message
         d.transport forward msg
-        stay()
+        msg match {
+          case _: OnionMessage if !d.isPersistent =>
+            startSingleTimer(KILL_IDLE_TIMER, KillIdle, conf.killIdleDelay)
+            stay()
+          // If we send any channel management message to this peer, the connection should be persistent.
+          case _: ChannelMessage if !d.isPersistent =>
+            stay() using d.copy(isPersistent = true)
+          case _ => stay()
+        }
 
       case Event(SendPing, d: ConnectedData) =>
         if (d.expectedPong_opt.isEmpty) {
@@ -279,7 +287,6 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
             // this is actually for the channel
             d.transport ! TransportHandler.ReadAck(msg)
             d.peer ! msg
-            stay()
           case _: ChannelAnnouncement | _: ChannelUpdate | _: NodeAnnouncement if d.behavior.ignoreNetworkAnnouncement =>
             // this peer is currently under embargo!
             d.transport ! TransportHandler.ReadAck(msg)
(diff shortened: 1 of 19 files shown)


thomash-acinq:
It depends on  #2092.
I'm not very happy with the name `enableGossip` because now it's not just about the gossip. But I would like to keep a positive value for the good kind of connection. Do you have a name suggestion?

thomash-acinq:
> It depends on #2092. I'm not very happy with the name `enableGossip` because now it's not just about the gossip. But I would like to keep a positive value for the good kind of connection. Do you have a name suggestion?

I've renamed it to `isPersistent`

thomash-acinq:
Removed the gossip part

t-bast:
Have you tried putting all this logic inside the `MessageRelay` actor instead?
Here is what I had in mind, which I believe would be less invasive:

- the `MessageRelay` does its initial magic to create a peer if needed and relay the onion message
- then it schedules a message to itself after `kill-transient-connection-after` (with `timers.startSingleTimer`)
- when the timer expires, the `MessageRelay` checks whether the outgoing `Peer` has channels -> if not it sends it a `Peer.Disconnect`

I believe this doesn't need changes anywhere else than inside the `MessageRelay` actor, which is good. It also makes it very easy to unit test.

thomash-acinq:
Since we want to add a `isPersistent` or equivalent to the `peerConnection` anyway to ignore the gossip anyway, I thought I would be simpler to reuse it. Most of the changes are to add this field, the rest is not invasive at all.

Also there is a problem with what you suggest in the case when we often relay messages to the same node:  If we relay a message to A every 10 seconds and `kill-transient-connection-after = 60s` we never want to kill the connection and that's not possible if we put that logic in several `MessageRelay`s that don't talk to each other (and we shouldn't make them talk to each other).

t-bast:
> Since we want to add a isPersistent or equivalent to the peerConnection anyway to ignore the gossip anyway, I thought I would be simpler to reuse it.

That's true, but now that we've realized that ignoring the gossip was a much harder task than we thought, I wasn't sure this was the way to go :grinning: 

> Also there is a problem with what you suggest in the case when we often relay messages to the same node

That's a very good point, since we can have N `MessageRelay` actor relaying to only 1 `Peer`, we cannot put this logic inside `MessageRelay`, it makes more sense to have it in the `Peer` or `PeerConnection`. That means we need to refresh that timer every time we relay a new onion message, and maybe disable it on some channel messages (or switch to `isPersistent = true`). We also need to make sure that the `ReconnectionTask` will not try to reconnect automatically.

I'll spend more time on this review once we're done with #2089

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -162,7 +162,15 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
     heartbeat {
       case Event(msg: LightningMessage, d: ConnectedData) if sender() != d.transport => // if the message doesn't originate from the transport, it is an outgoing message
         d.transport forward msg
-        stay()
+        msg match {
+          case _: OnionMessage if !d.isPersistent =>
+            startSingleTimer(KILL_IDLE_TIMER, KillIdle, conf.killIdleDelay)
+            stay()
+          // If we send any channel management message to this peer, the connection should be persistent.
+          case _: ChannelMessage if !d.isPersistent =>
```
t-bast, line 170: As I commented in the other PR, treating all channel messages equally is too easy to game: https://github.com/ACINQ/eclair/pull/2092#discussion_r766664266

t-bast (commented):
I'd like to brainstorm this a bit further, and I think @pm47's feedback would be useful.

While at first glance it seems to address our requirement (whenever we initiate a new connection to a peer to relay an onion message, we avoid keeping that connection open too long), if we think just a bit further I don't believe this achieves much...we treat all incoming connections as persistent, so if the peer just reconnects back to us, they upgraded the connection to be persistent.

It looks like this leads to the same kind of discussions we had regarding #2092, where my personal conclusion is that we need a more thorough re-work of connection handling with DoS in mind, and that half-measures like this are probably not very useful...

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -357,6 +364,9 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
       case Event(ResumeAnnouncements, d: ConnectedData) =>
         log.info(s"resuming processing of network announcements for peer")
         stay() using d.copy(behavior = d.behavior.copy(fundingTxAlreadySpentCount = 0, ignoreNetworkAnnouncement = false))
+
+      case Event(KillIdle, d: ConnectedData) if !d.isPersistent =>
+        stop(FSM.Normal)
```
t-bast, line 369: nit: it would be cleaner to handle the message in all cases, to avoid a warning log if the connection became persistent (because a channel was opened):

```suggestion
      case Event(KillIdle, d: ConnectedData) =>
          if (!d.isPersistent) {
              stop(FSM.Normal)
          } else {
              stay()
          }
```

t-bast (commented):
I've thought about it more, and while this doesn't address malicious peers, it is an improvement in the nominal case, and future DoS improvements will very likely build on top of this new boolean, so concept ACK.

This needs a few tests though:

- non-persistent -> persistent switch
- automatic killing when non-persistent

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
t-bast: But as a first step for the non-malicious cases, this is a good improvement, so we can address that later.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -357,6 +364,13 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
       case Event(ResumeAnnouncements, d: ConnectedData) =>
         log.info(s"resuming processing of network announcements for peer")
         stay() using d.copy(behavior = d.behavior.copy(fundingTxAlreadySpentCount = 0, ignoreNetworkAnnouncement = false))
+
+      case Event(KillIdle, d: ConnectedData) =>
+        if (!d.isPersistent) {
+          stop(FSM.Normal)
```
t-bast, line 370: We should log something, every other place where we stop ourselves, we have a log line to explain why, it will help troubleshooting user issues.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
```diff
@@ -372,5 +373,49 @@ class PeerConnectionSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike wi
...
   }
 
+  test("establish transient connection") { f =>
+    import f._
+    connect(nodeParams, remoteNodeId, switchboard, router, connection, transport, peerConnection, peer, isPersistent = false)
+    val probe = TestProbe()
+    val (_, message) = buildMessage(randomKey(), randomKey(), Nil, Left(Recipient(remoteNodeId, None)), Nil)
+    probe.send(peerConnection, message)
+    probe watch peerConnection
+    Thread.sleep(900)
```
t-bast, line 383: We should generally avoid `Thread.sleep` unless absolutely required, because it blocks an entire thread which could otherwise be used to run concurrent tests if you used a non-blocking construct instead. This will be especially penalizing during CI builds.

In this specific test, you can replace it with:

```scala
awaitCond(peerConnection.stateName === PeerConnection.CONNECTED)
probe.expectTerminated(peerConnection, max = Duration.Zero)
```

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
```diff
@@ -372,5 +373,49 @@ class PeerConnectionSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike wi
...
+    assert(peerConnection.stateName === PeerConnection.CONNECTED)
+    Thread.sleep(200)
+    probe.expectTerminated(peerConnection, max = Duration.Zero)
+  }
+
+  test("keep using transient connection") { f =>
```
t-bast, line 389: This test is very hard to write without `Thread.sleep`, I simply don't think it's worth the effort.

Your current version risks being very flaky, because it depends on timing, and during CI where a lot of tests are run in parallel, the execution of this specific test may pause for a while which could make it fail. I would simply remove that test, unless you find a clever way of testing this in a non-blocking way?

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
```diff
@@ -372,5 +373,49 @@ class PeerConnectionSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike wi
...
+    probe.expectTerminated(peerConnection, max = Duration.Zero)
+  }
+
+  test("convert transient connection to persistent") { f =>
+    import f._
+    connect(nodeParams, remoteNodeId, switchboard, router, connection, transport, peerConnection, peer, isPersistent = false)
+    val probe = TestProbe()
+    val (_, message) = buildMessage(randomKey(), randomKey(), Nil, Left(Recipient(remoteNodeId, None)), Nil)
+    probe.send(peerConnection, message)
+    Thread.sleep(900)
```
t-bast, line 414: Same, this test doesn't really need the `Thread.sleep`, we only want to send a channel message and then check that the connection is now persistent: I think you can access the actor state to directly verify the boolean, can't you?

t-bast (commented):
Thanks! I'd really, really like to avoid relying on `Thread.sleep` and timing assumptions in tests, we've had a lot of flaky tests and slow runs in the past because of it (and got rid of it almost everywhere afterwards). I've detailed why in my comments, let me know if that makes sense.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerConnectionSpec.scala
```diff
@@ -372,5 +373,53 @@ class PeerConnectionSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike wi
...
+    val probe = TestProbe()
+    val (_, message) = buildMessage(randomKey(), randomKey(), Nil, Left(Recipient(remoteNodeId, None)), Nil)
+    probe.send(peerConnection, message)
+    probe watch peerConnection
+    probe.expectTerminated(peerConnection, max = Duration(1500, MILLISECONDS))
+  }
+
+  def sleep(millis: Int): Unit = {
+    val probe = TestProbe()
+    system.scheduler.scheduleOnce(Duration(millis, MILLISECONDS), probe.ref, ())(system.dispatcher)
```
t-bast, line 388: If you import `import scala.concurrent.duration.DurationInt` you will have access to syntaxic sugars for creating `Duration` values, for exemple `100 milliseconds`, I recommend using those and having `sleep` directly take a `FiniteDuration` as argument.

thomash-acinq: Done

t-bast (approved):
Nice trick, LGTM!

I'm usually wary of testing specific timing values, as this can be flaky, but if it works let's try it, we can always remove it later if it's flaky.

thomash-acinq:
I've made so changes to the tests to not use `Thread.sleep` and ran it through the CI 6 times without failure.

PR #2101, opened 2021-12-13
Add payment hash to `ClaimHtlcSuccessTx`

t-bast:
This can be useful to match a preimage to a given `claim-htlc-success` transaction without going through the process of re-creating the transaction from scratch.

We already include the payment hash in `htlc-success` transactions for that reason.

Fortunately our codecs system is now very flexible: this was prefixed with a type, so we can easily decode old data and encode with the new codec. Code that depends on this will need to explicitly handle the case where `paymentHash == ByteVector32.Zeroes`.

This shouldn't be a an issue in practice though: these transactions are re-generated when calling `handleRemoteSpentCurrent` which happens when nodes restart as we replay `WatchFundingSpent`, so we should always have the payment hash available.


eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
```diff
@@ -137,7 +136,7 @@ private[channel] object ChannelCodecs0 {
       .typecase(0x01, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec)).as[CommitTx])
       .typecase(0x02, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec) :: ("paymentHash" | bytes32) :: ("htlcId" | provide(0L))).as[HtlcSuccessTx])
       .typecase(0x03, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec) :: ("htlcId" | provide(0L))).as[HtlcTimeoutTx])
-      .typecase(0x04, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec) :: ("htlcId" | provide(0L))).as[ClaimHtlcSuccessTx])
+      .typecase(0x04, (("inputInfo" | inputInfoCodec) :: ("tx" | txCodec) :: ("paymentHash" | provide(ByteVector32.Zeroes)) :: ("htlcId" | provide(0L))).as[ClaimHtlcSuccessTx])
```
pm47, line 139: Setting `paymentHash` to all-zeroes is different from, say, setting `htlcId` to zero because, as the comment above explains, the latter is explicitly backwards compatible.

What makes this change backwards compatible is the fact that we won't use the value for older non-replaceable commitments right? In that case an additional comment would be helpful:

```
// NB: 
// - we can safely set htlcId = 0 for htlc txs. This information is only used to find upstream htlcs to fail when a
// downstream htlc times out, and `Helpers.Closing.timedOutHtlcs` explicitly handles the case where htlcId is missing.
// - we can safely set paymentHash = 00...0 for claim-htlc-success, since we will only use the value for more recent commitments than the ones we were using when these transactions were serialized.
```

Alternatively, how about defining a `LegacyClaimHtlcSuccessTx`? I see the following +/-:
+ no need to alter older codecs
+ instead of relying on conventions "it's all zero but we don't use it", we would enforce with types
- more legacy code (but can be dropped as some point)

t-bast: > Alternatively, how about defining a LegacyClaimHtlcSuccessTx?

That's a good idea, I'll try that. I wanted to avoid letting this bleed into `Transactions.scala`, but when I used it in #2102 I initially forgot to handle the `paymentHash == ByteVector32.Zeroes` case, so it's a good indication that this is dangerous.

t-bast: Done in https://github.com/ACINQ/eclair/pull/2101/commits/83029b0b2fb30df37c02c3d29c6885b5dc5fd734

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -125,7 +125,8 @@ object Transactions {
   case class HtlcTimeoutTx(input: InputInfo, tx: Transaction, htlcId: Long) extends HtlcTx { override def desc: String = "htlc-timeout" }
   case class HtlcDelayedTx(input: InputInfo, tx: Transaction) extends TransactionWithInputInfo { override def desc: String = "htlc-delayed" }
   sealed trait ClaimHtlcTx extends TransactionWithInputInfo { def htlcId: Long }
-  case class ClaimHtlcSuccessTx(input: InputInfo, tx: Transaction, htlcId: Long) extends ClaimHtlcTx { override def desc: String = "claim-htlc-success" }
+  case class LegacyClaimHtlcSuccessTx(input: InputInfo, tx: Transaction, htlcId: Long) extends ClaimHtlcTx { override def desc: String = "claim-htlc-success" }
+  case class ClaimHtlcSuccessTx(input: InputInfo, tx: Transaction, paymentHash: ByteVector32, htlcId: Long) extends ClaimHtlcTx { override def desc: String = "claim-htlc-success" }
```
pm47, line 129: I suppose it wasn't useful to define:

```scala
sealed trait ClaimHtlcSuccessTx extends ClaimHtlcTx
case class LegacyClaimHtlcSuccessTx extends ClaimHtlcSuccessTx 
case class ReplaceableClaimHtlcSuccessTx extends ClaimHtlcSuccessTx 
```

t-bast: No, it's actually not used anywhere for now. I'll see if it's necessary in the follow-up PR once I rebase it (#2102)

pm47 (approved):

PR #2102, opened 2021-12-14
Make `ClaimHtlcTx` replaceable

t-bast:
We add `claim-htlc` transactions to the set of transactions that we will RBF if they don't confirm quickly.

This is important because the feerate is evaluated [when we see the remote commitment in the mempool](https://github.com/ACINQ/eclair/blob/ac9e274f0d1083319f70e08d495f20e27dd4cccd/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala#L1048), so it may be outdated by the time we can actually publish the `claim-htlc` transaction (either because of the absolute timelock or because the remote commitment took a long time to confirm).

It's very simple to set the fees for these transactions: since we can unilaterally sign, we simply decrease the output amount and re-sign. Note that we do it for both standard commitments and anchor output commitments. For standard commitments, these transactions are marked as non-RBFable, so RBF attempts will simply fail if the previous transaction is in the mempool. But it's still worth attempting it, as it may have been dropped from the mempool in high-fee environments.

The `ReplaceableTxPublisher` deserves some refactoring: integrating these various types of transactions shows where the model works and where it doesn't. But I've got a few other changes I'd like to make which impact the architecture, so I'd like to keep it like this for now and refactor after a few more PRs.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
@@ -124,7 +124,7 @@ object Transactions {
   case class HtlcSuccessTx(input: InputInfo, tx: Transaction, paymentHash: ByteVector32, htlcId: Long) extends HtlcTx { override def desc: String = "htlc-success" }
   case class HtlcTimeoutTx(input: InputInfo, tx: Transaction, htlcId: Long) extends HtlcTx { override def desc: String = "htlc-timeout" }
   case class HtlcDelayedTx(input: InputInfo, tx: Transaction) extends TransactionWithInputInfo { override def desc: String = "htlc-delayed" }
-  sealed trait ClaimHtlcTx extends TransactionWithInputInfo { def htlcId: Long }
+  sealed trait ClaimHtlcTx extends ReplaceableTransactionWithInputInfo { def htlcId: Long }
   case class LegacyClaimHtlcSuccessTx(input: InputInfo, tx: Transaction, htlcId: Long) extends ClaimHtlcTx { override def desc: String = "claim-htlc-success" }
   case class ClaimHtlcSuccessTx(input: InputInfo, tx: Transaction, paymentHash: ByteVector32, htlcId: Long) extends ClaimHtlcTx { override def desc: String = "claim-htlc-success" }
   case class ClaimHtlcTimeoutTx(input: InputInfo, tx: Transaction, htlcId: Long) extends ClaimHtlcTx { override def desc: String = "claim-htlc-timeout" }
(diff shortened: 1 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -124,7 +124,7 @@ object Transactions {
   case class HtlcSuccessTx(input: InputInfo, tx: Transaction, paymentHash: ByteVector32, htlcId: Long) extends HtlcTx { override def desc: String = "htlc-success" }
   case class HtlcTimeoutTx(input: InputInfo, tx: Transaction, htlcId: Long) extends HtlcTx { override def desc: String = "htlc-timeout" }
   case class HtlcDelayedTx(input: InputInfo, tx: Transaction) extends TransactionWithInputInfo { override def desc: String = "htlc-delayed" }
-  sealed trait ClaimHtlcTx extends TransactionWithInputInfo { def htlcId: Long }
+  sealed trait ClaimHtlcTx extends ReplaceableTransactionWithInputInfo { def htlcId: Long }
```
pm47, line 127: Neat

pm47 (approved):
I like how easy this change is! Shows that our model is good.

> The ReplaceableTxPublisher deserves some refactoring

A few minor remarks:
- How about renaming `RawTxPublisher` to `FinalTxPublisher`?
- https://github.com/ACINQ/eclair/blob/0b807d257a1ee1d33830fb91a11aa47ba7d3dcb0/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/TxPublisher.scala#L187 Wonder if this precheck should be done when starting the `RawTxPublisher` actor, which would then stop right away (same for ``ReplaceableTxPublisher`).
- https://github.com/ACINQ/eclair/blob/0b807d257a1ee1d33830fb91a11aa47ba7d3dcb0/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/TxPublisher.scala#L207 This log should make it clear that the existing tx has a higher feerate and that's why we don't replace it (e.g. append `" with higher feerate={}"`).

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
t-bast: :sunglasses:

t-bast:
> How about renaming `RawTxPublisher` to `FinalTxPublisher`?

Good idea, it's a better distinction with `ReplaceableTxPublisher`.
Even though I have in the back of my mind ideas for a future where such "final" txs can be CPFP-ed through their change output (e.g. for funding transactions), but whether it's `Raw` or `Final` doesn't matter for that.

I'll do this in a follow-up PR.

> Wonder if this precheck should be done when starting the RawTxPublisher actor, which would then stop right away

We can't do that for the `ReplaceableTxPublisher`, because the decision is based on the map kept by the `TxPublisher`.
It's true that we could do it for the `RawTxPublisher`, but I don't like creating this asymmetry, I think it's better to do it at the same place for both types of publishers (so it has to be in the `TxPublisher`).

> This log should make it clear that the existing tx has a higher feerate and that's why we don't replace it

In the next PR (#2113, not yet completely ready for review) I change this to use a deadline instead of an explicit feerate, so I'll fix it there.

PR #2104, opened 2021-12-14
Notify node operator on low fee bumping reserve

t-bast:
With anchor outputs, we need to keep utxos available for fee bumping: having enough on-chain funds to claim HTLCs in case channels force-close is necessary to guarantee funds safety.

There is no perfect formula to estimate how much we should reserve: it depends on the feerate when the channels force-close, the number of impacted HTLCs and their amount. We use a very rough estimation, to avoid needlessly spamming node operators while still notifying them when the situation becomes too risky.

The choice of `sqrt` to estimate the portion of malicious channels is particularly arbitrary, but we do have to choose _something_, and it feels somewhat reasonable (or at least we can argue that in practice, it has a reasonable distribution).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
@@ -27,7 +27,7 @@ object BalanceActor {
   private final case object TickBalance extends Command
   final case class GetGlobalBalance(replyTo: ActorRef[Try[GlobalBalance]], channels: Map[ByteVector32, HasCommitments]) extends Command
   private final case class WrappedChannels(wrapped: ChannelsListener.GetChannelsResponse) extends Command
-  private final case class WrappedGlobalBalance(wrapped: Try[GlobalBalance]) extends Command
+  private final case class WrappedGlobalBalanceWithChannels(wrapped: Try[GlobalBalance], channelsCount: Int) extends Command
   private final case class WrappedUtxoInfo(wrapped: Try[UtxoInfo]) extends Command
   // @formatter:on
 
@@ -79,9 +79,10 @@ private class BalanceActor(context: ActorContext[Command],
       context.pipeToSelf(checkUtxos(bitcoinClient))(WrappedUtxoInfo)
       Behaviors.same
     case WrappedChannels(res) =>
-      context.pipeToSelf(CheckBalance.computeGlobalBalance(res.channels, db, bitcoinClient))(WrappedGlobalBalance)
+      val channelsCount = res.channels.size
+      context.pipeToSelf(CheckBalance.computeGlobalBalance(res.channels, db, bitcoinClient))(b => WrappedGlobalBalanceWithChannels(b, channelsCount))
       Behaviors.same
-    case WrappedGlobalBalance(res) =>
+    case WrappedGlobalBalanceWithChannels(res, channelsCount) =>
       res match {
         case Success(result) =>
           log.info("current balance: total={} onchain.confirmed={} onchain.unconfirmed={} offchain={}", result.total.toDouble, result.onChain.confirmed.toDouble, result.onChain.unconfirmed.toDouble, result.offChain.total.toDouble)
@@ -90,11 +91,11 @@ private class BalanceActor(context: ActorContext[Command],
           val perChannelFeeBumpingReserve = 50_000.sat
           // Instead of scaling this linearly with the number of channels we have, we use sqrt(channelsCount) to reflect
           // the fact that if you have channels with many peers, only a subset of these peers will likely be malicious.
-          val estimatedFeeBumpingReserve = perChannelFeeBumpingReserve * Math.sqrt(result.channelsCount)
+          val estimatedFeeBumpingReserve = perChannelFeeBumpingReserve * Math.sqrt(channelsCount)
           if (result.onChain.confirmed < estimatedFeeBumpingReserve) {
             val message =
               s"""On-chain confirmed balance is low (${result.onChain.confirmed.toMilliBtc}): eclair may not be able to guarantee funds safety in case channels force-close.
-                 |You have ${result.channelsCount} open channels, which could cost $estimatedFeeBumpingReserve in fees if some of these channels are malicious.
+                 |You have $channelsCount channels, which could cost $estimatedFeeBumpingReserve in fees if some of these channels are malicious.
                  |Please note that the value above is a very arbitrary estimation: the real cost depends on the feerate and the number of malicious channels.
                  |You should add more utxos to your bitcoin wallet to guarantee funds safety.
                  |""".stripMargin
--- a/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
@@ -85,9 +85,20 @@ private class BalanceActor(context: ActorContext[Command],
       res match {
         case Success(result) =>
           log.info("current balance: total={} onchain.confirmed={} onchain.unconfirmed={} offchain={}", result.total.toDouble, result.onChain.confirmed.toDouble, result.onChain.unconfirmed.toDouble, result.offChain.total.toDouble)
-          log.debug("current balance details : {}", result)
-          if (result.onChain.confirmed < 1.millibtc) {
-            context.system.eventStream ! EventStream.Publish(NotifyNodeOperator(NotificationsLogger.Warning, s"on-chain confirmed balance is low (${result.onChain.confirmed.toMilliBtc}), eclair may not be able to guarantee funds safety in case channels force-close: you should add some utxos to your bitcoin wallet"))
+          log.debug("current balance details: {}", result)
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
```diff
@@ -85,9 +85,20 @@ private class BalanceActor(context: ActorContext[Command],
       res match {
         case Success(result) =>
           log.info("current balance: total={} onchain.confirmed={} onchain.unconfirmed={} offchain={}", result.total.toDouble, result.onChain.confirmed.toDouble, result.onChain.unconfirmed.toDouble, result.offChain.total.toDouble)
-          log.debug("current balance details : {}", result)
-          if (result.onChain.confirmed < 1.millibtc) {
-            context.system.eventStream ! EventStream.Publish(NotifyNodeOperator(NotificationsLogger.Warning, s"on-chain confirmed balance is low (${result.onChain.confirmed.toMilliBtc}), eclair may not be able to guarantee funds safety in case channels force-close: you should add some utxos to your bitcoin wallet"))
+          log.debug("current balance details: {}", result)
+          // This is a very rough estimation of the fee we would need to pay for a force-close with 5 pending HTLCs at 100 sat/byte.
+          val perChannelFeeBumpingReserve = 0.5.millibtc
```
pm47, line 90: Maybe arbitrary, but I would have set this in `sat`, because it's the unit we use to pay the fees. The `millibtc` unit makes sense for aggregated amounts in the monitoring console. 🤔

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
```diff
@@ -85,9 +85,20 @@ private class BalanceActor(context: ActorContext[Command],
...
+          // Instead of scaling this linearly with the number of channels we have, we use sqrt(channelsCount) to reflect
+          // the fact that if you have channels with many peers, only a subset of these peers will likely be malicious.
+          val estimatedFeeBumpingReserve = perChannelFeeBumpingReserve * Math.sqrt(result.channelsCount)
+          if (result.onChain.confirmed < estimatedFeeBumpingReserve) {
```
pm47, line 94: We should consider the _safe_ amount instead, which includes unconfirmed change.

eclair-core/src/main/scala/fr/acinq/eclair/balance/CheckBalance.scala
```diff
@@ -289,7 +289,7 @@ object CheckBalance {
     }
   } yield CorrectedOnChainBalance(detailed.confirmed, detailed.unconfirmed)
 
-  case class GlobalBalance (onChain: CorrectedOnChainBalance, offChain: OffChainBalance) {
+  case class GlobalBalance (onChain: CorrectedOnChainBalance, offChain: OffChainBalance, channelsCount: Int) {
```
pm47, line 292: I see how useful the `channelsCount` may be, but the inconsistency is making me OCD a little bit 😵

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
```diff
@@ -399,22 +399,15 @@ object OriginSerializer extends MinimalSerializer({
...
 
-object GlobalBalanceSerializer extends MinimalSerializer({
-  case o: GlobalBalance =>
-    val formats = DefaultFormats + ByteVector32KeySerializer + BtcSerializer + SatoshiSerializer
-    JObject(JField("total", JDecimal(o.total.toDouble))) merge Extraction.decompose(o)(formats)
-})
-
 // @formatter:off
+private case class GlobalBalanceJson(total: Double, onChain: CorrectedOnChainBalance, offChain: OffChainBalance)
+object GlobalBalanceSerializer extends ConvertClassSerializer[GlobalBalance](b => GlobalBalanceJson(b.total.toDouble, b.onChain, b.offChain))
```
pm47, line 404:
```suggestion
private case class GlobalBalanceJson(total: Btc, onChain: CorrectedOnChainBalance, offChain: OffChainBalance)
object GlobalBalanceSerializer extends ConvertClassSerializer[GlobalBalance](b => GlobalBalanceJson(b.total, b.onChain, b.offChain))
```

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
t-bast: I'm not sure we can, because BIP125 has restrictions on unconfirmed inputs, even if they're safe, so there's a chance that we actually can't use these inputs for RBF...so I'd rather be very conservative here and consider only confirmed utxos.

eclair-core/src/main/scala/fr/acinq/eclair/balance/CheckBalance.scala
t-bast: Yeah, I tried wrapping this in a parent type, but it wasn't very pretty either...but if you prefer I can do that, at least it keeps `GlobalBalance` with only balance-related things.

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2104/commits/e28956e4dacc3e630b88c41660074e5e6361568d

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
pm47: 0.5 millibtc = 5000 sat ?

eclair-core/src/main/scala/fr/acinq/eclair/balance/BalanceActor.scala
```diff
@@ -79,15 +79,27 @@ private class BalanceActor(context: ActorContext[Command],
...
+    case WrappedGlobalBalanceWithChannels(res, channelsCount) =>
       res match {
         case Success(result) =>
           log.info("current balance: total={} onchain.confirmed={} onchain.unconfirmed={} offchain={}", result.total.toDouble, result.onChain.confirmed.toDouble, result.onChain.unconfirmed.toDouble, result.offChain.total.toDouble)
-          log.debug("current balance details : {}", result)
-          if (result.onChain.confirmed < 1.millibtc) {
-            context.system.eventStream ! EventStream.Publish(NotifyNodeOperator(NotificationsLogger.Warning, s"on-chain confirmed balance is low (${result.onChain.confirmed.toMilliBtc}), eclair may not be able to guarantee funds safety in case channels force-close: you should add some utxos to your bitcoin wallet"))
+          log.debug("current balance details: {}", result)
+          // This is a very rough estimation of the fee we would need to pay for a force-close with 5 pending HTLCs at 100 sat/byte.
+          val perChannelFeeBumpingReserve = 50_000.sat
```
pm47, line 91:
```suggestion
          val perChannelFeeBumpingReserve = 5_000.sat
```

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
```diff
@@ -399,22 +399,15 @@ object OriginSerializer extends MinimalSerializer({
...
 private case class PeerInfoJson(nodeId: PublicKey, state: String, address: Option[InetSocketAddress], channels: Int)
 object PeerInfoSerializer extends ConvertClassSerializer[Peer.PeerInfo](peerInfo => PeerInfoJson(peerInfo.nodeId, peerInfo.state.toString, peerInfo.address, peerInfo.channels))
-// @formatter:on
 
-// @formatter:off
 private[json] case class MessageReceivedJson(pathId: Option[ByteVector], replyPath: Option[BlindedRoute], unknownTlvs: Map[String, ByteVector])
-object OnionMessageReceivedSerializer extends ConvertClassSerializer[OnionMessages.ReceiveMessage]({ m: OnionMessages.ReceiveMessage =>
-  MessageReceivedJson(m.pathId, m.finalPayload.replyPath.map(_.blindedRoute), m.finalPayload.records.unknown.map(tlv => (tlv.tag.toString -> tlv.value)).toMap)
-})
+object OnionMessageReceivedSerializer extends ConvertClassSerializer[OnionMessages.ReceiveMessage](m => MessageReceivedJson(m.pathId, m.finalPayload.replyPath.map(_.blindedRoute), m.finalPayload.records.unknown.map(tlv => tlv.tag.toString -> tlv.value).toMap))
```
pm47, line 410: Side note : not sure why the internal case class one line above was defined with `private[json]`.

pm47 (approved):

eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
t-bast: It's because it's needed for the websocket type hints: https://github.com/ACINQ/eclair/blob/535daec0657acf2e05f5c8fca046ae9ab3f70434/eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala#L455

PR #2106, opened 2021-12-14
Increase timeout onion message integration tests

t-bast:
In channel tests, we set a 60 seconds timeout on the receiver end, otherwise we sometimes run into timeout on slow CI machines that run a lot of tests in parallel.

It makes sense to do the same for onion messages, to help us figure out whether there is a race condition or it's just the machines that are slow.


thomash-acinq (approved):

t-bast:
I've ran the build 8 times, all green, so I guess this should be working!

PR #2108, opened 2021-12-15
Avoid default arguments in test helpers

t-bast:
This is tiny refactoring PR of some test helpers: I extracted this from a bigger branch to avoid cluttering a future PR (with anchor outputs stuff).

A few of our test helper functions started using too many default arguments, which makes them a bit messy to use in tests, especially when we want to add new arguments without fighting to fix all existing tests that relied on positional arguments...

We change this to use method overloading instead, which makes it easier to add new overloads without changes to existing tests.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/ChannelStateTestsHelperMethods.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/ChannelStateTestsHelperMethods.scala
@@ -228,18 +228,38 @@ trait ChannelStateTestsHelperMethods extends TestKitBase {
     channelUpdateListener.expectMsgType[LocalChannelUpdate]
   }
 
-  def localOrigin(replyTo: ActorRef): Origin.LocalHot = Origin.LocalHot(replyTo, UUID.randomUUID)
+  def localOrigin(replyTo: ActorRef): Origin.LocalHot = Origin.LocalHot(replyTo, UUID.randomUUID())
 
-  def makeCmdAdd(amount: MilliSatoshi, destination: PublicKey, currentBlockHeight: Long, paymentPreimage: ByteVector32 = randomBytes32(), upstream: Upstream = Upstream.Local(UUID.randomUUID), replyTo: ActorRef = TestProbe().ref): (ByteVector32, CMD_ADD_HTLC) = {
+  def makeCmdAdd(amount: MilliSatoshi, destination: PublicKey, currentBlockHeight: Long): (ByteVector32, CMD_ADD_HTLC) = {
+    makeCmdAdd(amount, CltvExpiryDelta(144), destination, randomBytes32(), currentBlockHeight, Upstream.Local(UUID.randomUUID()))
+  }
+
+  def makeCmdAdd(amount: MilliSatoshi, destination: PublicKey, currentBlockHeight: Long, paymentPreimage: ByteVector32): (ByteVector32, CMD_ADD_HTLC) = {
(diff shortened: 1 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/channel/states/ChannelStateTestsHelperMethods.scala
```diff
@@ -228,18 +228,38 @@ trait ChannelStateTestsHelperMethods extends TestKitBase {
...
+  }
+
+  def makeCmdAdd(amount: MilliSatoshi, destination: PublicKey, currentBlockHeight: Long, paymentPreimage: ByteVector32): (ByteVector32, CMD_ADD_HTLC) = {
+    makeCmdAdd(amount, destination, currentBlockHeight, paymentPreimage, Upstream.Local(UUID.randomUUID()))
+  }
+
+  def makeCmdAdd(amount: MilliSatoshi, destination: PublicKey, currentBlockHeight: Long, paymentPreimage: ByteVector32, upstream: Upstream): (ByteVector32, CMD_ADD_HTLC) = {
+    makeCmdAdd(amount, CltvExpiryDelta(144), destination, paymentPreimage, currentBlockHeight, upstream)
+  }
+
```
thomash-acinq, line 244: Isn't this strictly equivalent to what you're replacing?

thomash-acinq (commented):
I don't understand the benefits of this change. You should add new optional arguments at the end and it won't break anything.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/ChannelStateTestsHelperMethods.scala
t-bast: Yes it's exactly equivalent, the only distinction is that it's nicer to use in tests: you don't have to name arguments and you can put the most relevant arguments at the beginning of the call, so it makes tests slightly easier to read.

> You should add new optional arguments at the end and it won't break anything.

I'd like to avoid that because I'd rather have the important arguments (the one we're actually testing) at the beginning of the arguments list. I agree it's a very minor change, but it's a minor help in test readability, and the tests I was updating in `NormalStateSpec` could use some help to make them easier to read!

thomash-acinq: Since both versions are equivalent, I think we should chose the shorter one. I don't see how expanding one line to nine lines makes it more readable.

```suggestion
  def makeCmdAdd(amount: MilliSatoshi, destination: PublicKey, currentBlockHeight: Long, paymentPreimage: ByteVector32 = randomBytes32(), upstream: Upstream = Upstream.Local(UUID.randomUUID())): (ByteVector32, CMD_ADD_HTLC) = {
    makeCmdAdd(amount, CltvExpiryDelta(144), destination, paymentPreimage, currentBlockHeight, upstream)
  }
```

thomash-acinq: I understand that it is not possible because of limitations from the scala compiler.

thomash-acinq (approved):

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/ChannelStateTestsHelperMethods.scala
t-bast: Yes exactly, unfortunately because we are inside a `trait`...

PR #2109, opened 2021-12-17
Add defenses against looping paths

thomash-acinq:


t-bast (approved):
:+1:

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -130,6 +130,10 @@ object RouteCalculation {
             Metrics.RouteResults.withTags(tags).record(routes.length)
             routes.foreach(route => Metrics.RouteLength.withTags(tags).record(route.length))
             ctx.sender() ! RouteResponse(routes)
+          case Failure(InfiniteLoop(loop)) =>
+            log.error(s"found infinite loop ${loop.map(edge => edge.desc).mkString(" -> ")}")
+            Metrics.FindRouteErrors.withTags(tags.withTag(Tags.Error, "InfiniteLoop")).increment()
+            ctx.sender() ! Status.Failure(InfiniteLoop(loop))
```
pm47, line 136: Small detail (no unnecessary degree of liberty by instantiating a new exception, and more consistent with the next handler):
```scala
          case Failure(failure: InfiniteLoop) =>
            log.error(s"found infinite loop ${failure.loop.map(edge => edge.desc).mkString(" -> ")}")
            Metrics.FindRouteErrors.withTags(tags.withTag(Tags.Error, "InfiniteLoop")).increment()
            ctx.sender() ! Status.Failure(failure)
```

PR #2111, opened 2021-12-20
Fix potential loop in path-finding

thomash-acinq:
Use the amount **without** the fees when computing the success probability of routing through an edge.
This fixes a problem where the computed probability could be negative, leading to a shorter path after adding an edge.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -325,7 +325,7 @@ object Graph {
         val riskCost = totalAmount.toLong * totalCltv.toInt * heuristicsConstants.lockedFundsRisk
         // If the edge was added by the invoice, it is assumed that it can route the payment.
         // If we know the balance of the channel, then we will check separately that it can relay the payment.
-        val successProbability = if (edge.update.chainHash == ByteVector32.Zeroes || edge.balance_opt.nonEmpty) 1.0 else 1.0 - totalAmount.toLong.toDouble / edge.capacity.toMilliSatoshi.toLong.toDouble
+        val successProbability = if (edge.update.chainHash == ByteVector32.Zeroes || edge.balance_opt.nonEmpty) 1.0 else 1.0 - prev.amount.toLong.toDouble / edge.capacity.toMilliSatoshi.toLong.toDouble
         val totalSuccessProbability = prev.successProbability * successProbability
         val failureCost = nodeFee(heuristicsConstants.failureCost.feeBase, heuristicsConstants.failureCost.feeProportionalMillionths, totalAmount)
         val weight = totalFees.toLong + totalHopsCost.toLong + riskCost + failureCost.toLong / totalSuccessProbability
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -90,6 +90,8 @@ object Graph {
 
   case class InfiniteLoop(path: Seq[GraphEdge]) extends Exception
 
+  case class NegativeProbability(edge: GraphEdge, weight: RichWeight, heuristicsConstants: HeuristicsConstants) extends Exception
+
   /**
    * Yen's algorithm to find the k-shortest (loop-less) paths in a graph, uses dijkstra as search algo. Is guaranteed to
    * terminate finding at most @pathsToFind paths sorted by cost (the cheapest is in position 0).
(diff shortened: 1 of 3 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -239,4 +238,19 @@ class GraphSpec extends AnyFunSuite {
 
   def edgeFromNodes(shortChannelId: Long, a: PublicKey, b: PublicKey): GraphEdge = makeEdge(shortChannelId, a, b, 0 msat, 0)
 
+  test("amount with fees larger than channel capacity") {
```
t-bast, line 241: Please add a diagram, it's much easier to understand this test and verify that the route with the problematic channel capacity is taken into account.

There are a few things I don't understand about this test.
First, why is the D -> C channel needed?
Second, the faulty route (C -> D -> E) costs more fees than the route C -> E, so it's really not obvious that it's considered by the algorithm. I think you should make that route cheaper than C -> E to help readers see that it's the best route, but it has a capacity issue.

Last, I think we should run  `yenKshortestPaths` with both `HeuristicsConstants` and `WeightRatios`, it's good to test both of them in case we introduce a similar issue in the `WeightRatios` case in the future (we would catch it instantly).

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -325,7 +325,7 @@ object Graph {
         val riskCost = totalAmount.toLong * totalCltv.toInt * heuristicsConstants.lockedFundsRisk
         // If the edge was added by the invoice, it is assumed that it can route the payment.
         // If we know the balance of the channel, then we will check separately that it can relay the payment.
-        val successProbability = if (edge.update.chainHash == ByteVector32.Zeroes || edge.balance_opt.nonEmpty) 1.0 else 1.0 - totalAmount.toLong.toDouble / edge.capacity.toMilliSatoshi.toLong.toDouble
+        val successProbability = if (edge.update.chainHash == ByteVector32.Zeroes || edge.balance_opt.nonEmpty) 1.0 else 1.0 - prev.amount.toLong.toDouble / edge.capacity.toMilliSatoshi.toLong.toDouble
```
t-bast, line 328: It feels to me that we should still take `totalAmount` into account since that's what will flow through this channel, isn't it?
But we should do a `.max(0)` to ensure it cannot be negative, WDYT? 

Otherwise a future reader of that code may think "this is a mistake, we should use `totalAmount`" and would "fix" this without realizing that they're re-introducing a bug.

t-bast (commented):
Good catch!

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
thomash-acinq: No, only `prev.amount` will flow through the channel.

t-bast: Don't you think it's safer to add a `.max(0)` anyway though?

thomash-acinq: If the probability is negative, it means we have a bug somewhere. Adding `.max(0)` will only make sure that we never discover that bug.

t-bast: Fair enough, let's hope we catch those bugs before production though

t-bast: Or better, we can add a `require`, this way if it happens it crashes the actor and tells us exactly what happened, instead of causing all kinds of unexpected behavior?

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
thomash-acinq: Should I just remove it then?

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
thomash-acinq: I could add an assert if you think there is a risk. That way we would get something logged, similar to what I did with `InfiniteLoop`.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
t-bast: No, clearly not, but it should contain comments to explain what the issue was and how it's tested. Just imagine someone reading this test in one year, and think about how you could help them understand what this is testing and why it's important.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
t-bast: It's your call, whatever you think is best here.

t-bast (approved):

PR #2115, opened 2021-12-24
Allow validation skip for non-standard channels

btcontract:
Currently non-standard channels can not get into base node graph because they fail validation checks (such as presence of tx on chain which is impossible by design for hosted channels for example). This makes them unusable for locally originated payments or trampoline routing, which hurts liquidity.

This new strategy adds an `AbstractLocalChannelUpdate` trait and `ValidationSkippingLocalChannelUpdate` invariant which would make router to just include a non-standard local channel into private channel map without doing any checks.

I'm interested in your opinion on this approach.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Logs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Logs.scala
@@ -24,7 +24,7 @@ import akka.io.Tcp
 import fr.acinq.bitcoin.ByteVector32
 import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.eclair.blockchain.bitcoind.ZmqWatcher.ValidateResult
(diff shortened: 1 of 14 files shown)


btcontract:
Just to clarify, this is a draft but I've stopped at this point to get some initial opinion whether the direction is fine. If it is I'll continue with tests and whatever remains to be done.

t-bast:
I think this is way to invasive, we must find a better solution. In the longer term, we will fully rework the gossip to be private (ie not expose the underlying on-chain utxo), so validation will have to be more flexible, but I'm not sure how exactly.

It would be nice if the mechanisms we put in place now can easily be extended in the future for such private gossip.
Sorry I don't have good guidance yet (I don't have time to investigate this much), but you should keep exploring other ways (less invasive on core eclair types) of achieving this.

PR #2117, opened 2021-12-31
Process replies to onion messages

thomash-acinq:
In preparation for offers, add a postman that listen to incoming replies and delivers them to the right actor.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -33,14 +33,13 @@ import fr.acinq.eclair.blockchain.bitcoind.rpc.BitcoinCoreClient.WalletTx
 import fr.acinq.eclair.blockchain.fee.{FeeratePerByte, FeeratePerKw}
 import fr.acinq.eclair.channel._
 import fr.acinq.eclair.crypto.Sphinx
-import fr.acinq.eclair.crypto.Sphinx.RouteBlinding.BlindedRoute
 import fr.acinq.eclair.db.AuditDb.{NetworkFee, Stats}
 import fr.acinq.eclair.db.{IncomingPayment, OutgoingPayment}
-import fr.acinq.eclair.io.MessageRelay.RelayAll
 import fr.acinq.eclair.io.Peer.{GetPeerInfo, PeerInfo}
-import fr.acinq.eclair.io.Switchboard.{RelayMessage, SendMessage}
+import fr.acinq.eclair.io.Switchboard.SendMessage
 import fr.acinq.eclair.io.{MessageRelay, NodeURI, Peer, PeerConnection, Switchboard}
-import fr.acinq.eclair.message.OnionMessages
+import fr.acinq.eclair.message.OnionMessages.OnionMessageResponse
+import fr.acinq.eclair.message.{OnionMessages, Postman}
 import fr.acinq.eclair.payment._
 import fr.acinq.eclair.payment.receive.MultiPartHandler.ReceivePayment
 import fr.acinq.eclair.payment.relay.Relayer.{GetOutgoingChannels, OutgoingChannels, RelayFees, UsableBalance}
@@ -68,8 +67,8 @@ case class SignedMessage(nodeId: PublicKey, message: String, signature: ByteVect
 
 case class VerifiedMessage(valid: Boolean, publicKey: PublicKey)
 
-case class ResponsePayload(replyPath: Option[String], unknownTlvs: Map[String, ByteVector])
-case class SendOnionMessageResponse(failureMessage: Option[String], response: Option[ResponsePayload])
+case class SendOnionMessageResponsePayload(encodedReplyPath: Option[String], replyPath: Option[Sphinx.RouteBlinding.BlindedRoute], unknownTlvs: Map[String, ByteVector])
+case class SendOnionMessageResponse(sent: Boolean, failureMessage: Option[String], response: Option[SendOnionMessageResponsePayload])
 
 object SignedMessage {
   def signedBytes(message: ByteVector): ByteVector32 =
@@ -158,7 +157,7 @@ trait Eclair {
 
   def verifyMessage(message: ByteVector, recoverableSignature: ByteVector): VerifiedMessage
 
-  def sendOnionMessage(intermediateNodes: Seq[OnionMessages.IntermediateNode], destination: Either[OnionMessages.Recipient, Sphinx.RouteBlinding.BlindedRoute], replyPath: Option[Seq[PublicKey]], userCustomContent: ByteVector)(implicit timeout: Timeout): Future[SendOnionMessageResponse]
+  def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: Either[PublicKey, Sphinx.RouteBlinding.BlindedRoute], replyPath: Option[Seq[PublicKey]], userCustomContent: ByteVector)(implicit timeout: Timeout): Future[SendOnionMessageResponse]
 }
 
 class EclairImpl(appKit: Kit) extends Eclair with Logging {
@@ -511,8 +510,8 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
     }
   }
 
-  override def sendOnionMessage(intermediateNodes: Seq[OnionMessages.IntermediateNode],
-                                destination: Either[OnionMessages.Recipient, Sphinx.RouteBlinding.BlindedRoute],
+  override def sendOnionMessage(intermediateNodes: Seq[PublicKey],
+                                destination: Either[PublicKey, Sphinx.RouteBlinding.BlindedRoute],
                                 replyPath: Option[Seq[PublicKey]],
                                 userCustomContent: ByteVector)(implicit timeout: Timeout): Future[SendOnionMessageResponse] = {
     codecs.list(TlvCodecs.genericTlv).decode(userCustomContent.bits) match {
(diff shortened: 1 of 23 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -154,7 +158,7 @@ trait Eclair {
 
   def verifyMessage(message: ByteVector, recoverableSignature: ByteVector): VerifiedMessage
 
-  def sendOnionMessage(route: Seq[PublicKey], userCustomContent: ByteVector, pathId: Option[ByteVector])(implicit timeout: Timeout): Future[SendOnionMessageResponse]
+  def sendOnionMessage(intermediateNodes: Seq[OnionMessages.IntermediateNode], destination: Either[OnionMessages.Recipient, Sphinx.RouteBlinding.BlindedRoute], replyPath: Option[Seq[PublicKey]], userCustomContent: ByteVector)(implicit timeout: Timeout): Future[SendOnionMessageResponse]
```
t-bast, line 161: This function should avoid using business logic types unless really necessary, it should stay as general as possible (and avoid needing to change when business logic types change): callers don't need to know about the `OnionMessages.*` types in most cases, a simple `PublicKey` is enough.

I think we can use simplify it to the following definition and create the business logic types inside of the function body:
```suggestion
  def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: Either[PublicKey, Sphinx.RouteBlinding.BlindedRoute], replyPath: Option[Seq[PublicKey]], userCustomContent: ByteVector)(implicit timeout: Timeout): Future[SendOnionMessageResponse]
```

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Message.scala
```diff
@@ -41,8 +43,16 @@ trait Message {
   }
 
   val sendOnionMessage: Route = postRequest("sendonionmessage") { implicit t =>
-    formFields("route".as[List[PublicKey]](pubkeyListUnmarshaller), "content".as[ByteVector](binaryDataUnmarshaller), "pathId".as[ByteVector](binaryDataUnmarshaller)?) { (route, userCustomContent, pathId) =>
-      complete(eclairApi.sendOnionMessage(route, userCustomContent, pathId))
+    formFields(
+      "destination".as[Either[OnionMessages.Recipient, Sphinx.RouteBlinding.BlindedRoute]](destinationUnmarshaller),
```
t-bast, line 47: From an API's point of view, I think it's confusing to use the same field for two very distinct cases (and having to match on its length to decide which case we're using). What about using two distinct arguments (e.g. `recipientNode` and `recipientBlindedRoute`) and requiring that exactly one is provided (like we do in some places where we want either an invoice to be provided or an explicit payment hash)?

For the caller apps, I believe it's cleaner as well as it maps to separate use-cases (e.g. messaging a given nodeId vs fetching an invoice through offers).

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -65,7 +68,8 @@ case class SignedMessage(nodeId: PublicKey, message: String, signature: ByteVect
 
 case class VerifiedMessage(valid: Boolean, publicKey: PublicKey)
 
-case class SendOnionMessageResponse(sent: Boolean, failureMessage: Option[String])
+case class ResponsePayload(replyPath: Option[String], unknownTlvs: Map[String, ByteVector])
+case class SendOnionMessageResponse(failureMessage: Option[String], response: Option[ResponsePayload])
```
t-bast, line 72: This new data contract makes it harder to use for fire-and-forget onion messages (which could be used for messaging for example), as they would get a completely empty response when the message is correctly sent. I think we should keep the `sent` boolean for those use-cases, don't you think?

Similarly to my comment on the `OnionMessageReceived`, I think it's useful to keep the reply path in json to allow easy inspection, we can have both the json and wire-encoded versions in the response, can't we?

Apart from that, `ResponsePayload` is too broad for a case class that's put on the top-level namespace, what about:
```suggestion
case class SendOnionMessageResponsePayload(encodedReplyPath: Option[String], replyPath: Option[BlindedRoute], unknownTlvs: Map[String, ByteVector])
case class SendOnionMessageResponse(sent: Boolean, response: Option[SendOnionMessageResponsePayload], failureMessage: Option[String])
```

t-bast (commented):
I didn't realize that it would be as tricky as this!

This is a first pass, mostly on the parts that we expose to the outside world (API and types), I haven't finished reviewing the `Switchboard` and `Postman` yet.

But I have a high-level design comment: why does `SendMessage` even go through the `Switchboard`? Why isn't the `Postman` the entry point for sending onion messages, which would I believe let us get rid of the weird intermediate anonymous actors?

I may be missing something, but I would reverse the dependency flow: the `Switchboard` wouldn't know anything about the `Postman`, the `Postman` would ask the `Switchboard` to relay messages and put itself in the `replyTo` field, and would do the conversion for the actual caller (in all cases, regardless of whether there is a `pathId` or not).

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -18,17 +18,23 @@ package fr.acinq.eclair.message
...
 import scodec.{Attempt, DecodeResult}
 
+import scala.concurrent.duration.FiniteDuration
 import scala.util.{Failure, Success}
 
 object OnionMessages {
 
+  case class OnionMessageConfig(relayPolicy: RelayPolicy, timeout: FiniteDuration)
+
+  trait OnionMessageResponse
```
t-bast, line 36: A non-sealed trait doesn't let the compiler check for exhaustiveness in pattern matching.
Whenever you'll pattern-match on a trait, you should try to make it `sealed`.

I think this shows an issue with the architecture of this feature.
This trait should be the response trait for `Switchboard.SendMessage`: there should be one `case class` extending this trait for each possible thing that can happen when sending an onion message:

- we sent the message, and it's not expecting a response (yay!)
- we received a response (yay!)
- we couldn't send the message
- we sent the message, expected a response, but didn't get one

This trait (and its child case classes) should be defined in the actor that handles `SendMessage`.
I don't think it's the job of the `Switchboard` to do this mapping, I believe the `Postman` would be a better candidate for this.
The responsibilities of the `Postman` would then be:

- receiving `SendMessage` requests
- storing information about who to reply to if a response is expected
- asking the switchboard to actually send the onion message
- receiving a relay error, a response, or timing out -> crafting the corresponding `OnionMessageResponse` and sending it to its intended recipient actor

This feels like a cleaner separation of responsibilities (not overloading the switchboard with new responsibilities and somewhat weird anonymous short-lived actors).

This also removes the `SubscribeOnce` / `Unsubscribe` paradigm in favor of a simpler approach: the `SendMessage` request is that "subscription", and when that message has been completely handled, we remove it from the internal state of the `Postman`: no need for an external actor to call `Unsubscribe` at the right time (which could go wrong).

WDYT?

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -119,7 +129,11 @@ object OnionMessages {
                 MessageOnionCodecs.blindedRelayPayloadCodec.decode(decrypted.bits) match {
                   case Attempt.Successful(DecodeResult(relayNext, _)) =>
                     val toRelay = OnionMessage(relayNext.nextBlindingOverride.getOrElse(nextBlindingKey), nextPacket)
-                    SendMessage(relayNext.nextNodeId, toRelay)
+                    if (relayNext.nextNodeId == privateKey.publicKey) { // we may add ourselves to the route several times to hide the real length of the route
+                      process(privateKey, toRelay)
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -154,7 +157,7 @@ trait Eclair {
 
   def verifyMessage(message: ByteVector, recoverableSignature: ByteVector): VerifiedMessage
 
-  def sendOnionMessage(route: Seq[PublicKey], userCustomContent: ByteVector, pathId: Option[ByteVector])(implicit timeout: Timeout): Future[SendOnionMessageResponse]
+  def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: Either[PublicKey, Sphinx.RouteBlinding.BlindedRoute], replyPath: Option[Seq[PublicKey]], userCustomContent: ByteVector)(implicit timeout: Timeout): Future[SendOnionMessageResponse]
```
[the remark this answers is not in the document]
thomash-acinq: No, the `pathId` is a secret from the receiver to itself, there is no case in which the sender knows it. In the case you mention, the `pathId` is inside the `replyPath` and is not visible to us.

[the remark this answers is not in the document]
t-bast: Right, good catch!

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/integration/MessageIntegrationSpec.scala
```diff
@@ -65,11 +65,44 @@ class MessageIntegrationSpec extends IntegrationSpec {
...
+    val eventListener = TestProbe()
+    nodes("B").system.eventStream.subscribe(eventListener.ref, classOf[OnionMessages.ReceiveMessage])
+    alice.sendOnionMessage(Nil, Left(nodes("B").nodeParams.nodeId), Some(Nil), hex"3f00").pipeTo(probe.ref)
+
+    val recv = eventListener.expectMsgType[OnionMessages.ReceiveMessage](max = 60 seconds)
+    assert(recv.finalPayload.replyPath.nonEmpty)
+    bob.sendOnionMessage(Nil, Right(recv.finalPayload.replyPath.get.blindedRoute), None, hex"1d01ab")
+
+    val res = probe.expectMsgType[SendOnionMessageResponse]
+    print(res)
```
t-bast, line 87: nit: remove `print` here and `println` in this file

eclair-core/src/test/scala/fr/acinq/eclair/io/SwitchboardSpec.scala
```diff
@@ -20,7 +20,7 @@ class SwitchboardSpec extends TestKitBaseClass with AnyFunSuiteLike {
 
   test("on initialization create peers") {
     val nodeParams = Alice.nodeParams
-    val (probe, peer) = (TestProbe(), TestProbe())
+    val (probe, peer, postman) = (TestProbe(), TestProbe(), TestProbe())
```
t-bast, line 23: I think you can remove all the changes made to this file now that the `Switchboard` doesn't depend on the `postman`.

eclair-core/src/test/scala/fr/acinq/eclair/message/PostmanSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+import fr.acinq.eclair.wire.protocol.OnionMessagePayloadTlv.{EncryptedData, ReplyPath}
+import fr.acinq.eclair.wire.protocol.{GenericTlv, TlvStream}
+import fr.acinq.eclair.{UInt64, randomBytes32, randomKey}
+import org.scalatest.Outcome
+import org.scalatest.funsuite.FixtureAnyFunSuiteLike
+import scodec.bits.HexStringSyntax
+
+import scala.concurrent.duration._
+
+class PostmanSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("application")) with FixtureAnyFunSuiteLike {
```
t-bast, line 37: This file deserves a few more tests IMO:

- send message without pathId that fails to be relayed -> sender should receive a failure
- send message without pathId that is relayed -> sender should receive a success
- send message with pathId that fails to be relayed -> sender should receive a failure
- send message with pathId that is relayed (postman receives the event from switchboard) but then times out -> sender should only receive the timeout error, not the switchboard "sent" event

t-bast (commented):
Good job, a few nit comments but nothing major.

eclair-core/src/test/scala/fr/acinq/eclair/message/PostmanSpec.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/message/Postman.scala
```diff
@@ -0,0 +1,109 @@
...
+      val subscribed = new mutable.HashMap[ByteVector32, ActorRef[OnionMessageResponse]]()
+      val sendStatusTo = new mutable.HashMap[ByteVector32, ActorRef[OnionMessageResponse]]()
+      val sendFailureTo = new mutable.HashMap[ByteVector32, ActorRef[OnionMessageResponse]]()
+
+      Behaviors.receiveMessagePartial {
+        case WrappedMessage(finalPayload, Some(pathId)) if pathId.length == 32 =>
+          subscribed.get(ByteVector32(pathId)) match {
+            case Some(ref) =>
+              subscribed -= ByteVector32(pathId)
+              ref ! Response(finalPayload)
```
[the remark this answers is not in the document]
thomash-acinq: I actually think that the current mechanism does the job perfectly. If someone uses our reply path to send us a message from a different context, we'll just forward it to `ref`, `ref` will ignore it or send an error saying it didn't understand the message and there is no leakage of information.

t-bast (commented):
LGTM, but I really think we need to add the `TODO` for the unsubscribe case, this is a subtle issue that we can easily forget to fix later.

t-bast (approved):

PR #2120, opened 2022-01-03
Use psbt to fund and sign transactions

sstone:
Based on https://github.com/ACINQ/eclair/pull/2038. We use bitcoin-kmp's Psbt classes and methods directly (there is no need for a wrapper here imho).


sstone:
Not needed anymore, see https://github.com/ACINQ/eclair/pull/2613

PR #2122, opened 2022-01-04
Update cluster documentation [ci skip]

sstone:


docs/Cluster.md
```diff
@@ -93,24 +93,38 @@ Generate a self-signed certificate (set a strong password):
...
 $ cp akka-cluster-tls.jks ~/.eclair
 ```
+Add this to `eclair.conf` on all your frontend nodes:
+```
+akka.remote.artery.transport = "tls-tcp"
+```
 
-Add this to your `eclair.conf`:
+Start all your frontend nodes with the following environment variables:
+* BACKEND_IP set the the IP address of your backend node
```
pm47, line 113:
```suggestion
* BACKEND_IP set to the IP address of your backend node
```

docs/Cluster.md
```diff
@@ -93,24 +93,38 @@ Generate a self-signed certificate (set a strong password):
...
 ```shell
 $ cp akka-cluster-tls.jks ~/.eclair
 ```
+Add this to `eclair.conf` on all your frontend nodes:
+```
+akka.remote.artery.transport = "tls-tcp"
+```
 
-Add this to your `eclair.conf`:
+Start all your frontend nodes with the following environment variables:
```
pm47, line 112: The part you have added is useful, but unrelated to encryption of akka communication. Needs to be put in a different section.

```suggestion
#### Run cluster nodes on separate servers

Start all your frontend nodes with the following environment variables:
```

pm47 (commented):
Sorry for the late review.

There is an ambiguity in the documentation between _akka_ nodes and _lightning_ nodes. Any idea how to sort it out?

sstone:
> There is an ambiguity in the documentation between akka nodes and lightning nodes. Any idea how to sort it out?

What abut using "cluster node" explicitly ?

pm47:
Okay, and also replace the node at the beginning (I'll let you improve with your perfect oxford english 🇬🇧 ):

> In the following, what we previously called `eclair-node` will be called `backend`. It is to be launched, configured and backed-up exactly like in a standalone setup.

by:

> In the following:
>- what we previously called `eclair-node` will be called `backend`. It is to be launched, configured and backed-up exactly like in a standalone setup.
>- the _node_ terminology refers to _cluster nodes_, as in nodes of the akka cluster. They are not to be confounded with _lightning nodes_. Together, all _cluster nodes_ form a single logical _lightning node_.

pm47 (approved):

PR #2123, opened 2022-01-05
Choose node address at random when reconnecting

pm47:
Instead of always using the last known peer address from the `PeersDb`,
we randomly pick the address from the `PeersDb` and the `NetworkDb`.

We use clearnet or tor addresses depending on the node configuration:
- tor not supported: always use clearnet addresses
- tor supported but not enabled for clearnet addresses: use clearnet
  addresses when available
- tor supported and enabled for clearnet addresses: use tor
  addresses when available


t-bast (approved):
LGTM!

The PR description is inaccurate though, your last two conditions are the same, but with a different result.

t-bast (approved):
LGTM, would be interesting to have @rorp chime in on the tor part.

PR #2124, opened 2022-01-05
(Minor) Use `sys` package instead of `System` when applicable

pm47:
We use scala's `sys.exit()` instead of java's `System.exit()`, etc.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -124,11 +124,10 @@ object NodeParams extends Logging {
    * Order of precedence for the configuration parameters:
    * 1) Java environment variables (-D...)
    * 2) Configuration file eclair.conf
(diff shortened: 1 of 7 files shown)


t-bast:
Good idea to harmonize, as we have a mix of `System.exit` and `sys.exit`.
But if we want to use different return values for each error, we should define them in a single place, shouldn't we?
For example in a new file where we define a variable for each error (e.g. `val BitcoinStillIndexingErr = 5`) otherwise we need to search for all existing `sys.exit` to see which return values are already used.

pm47:
> But if we want to use different return values for each error, we should define them in a single place, shouldn't we?

Or we could always return `1`, which is probably best since no-one uses those return values anyway?

t-bast:
That works too, we can start with that and if someone needs eclair to return distinct return values, they'll say it and we can update

t-bast (approved):

pm47:
I'm stopping with that commit I swear!

t-bast (approved):
:+1:

PR #2126, opened 2022-01-11
Improve `getsentinfo` accuracy

### LINKED ISSUE #2055: `getsentinfo` returns an indeterminate response `[]`
`getsentinfo` may return `[]` because it doesn't know anything about the given `--paymentHash` or `--id`, which means a payment wasn't performed or attempted, and is not pending.

But it may also return `[]` for a payment that is in process of being attempted, for some seconds or fractions of second before it is actually attempted, which has the opposite meaning.

For services that send payouts automatically on behalf of users and then check these payment statuses automatically this can lead to terrible loss of funds.

t-bast:
When using MPP, we previously inserted an entry in the payments DB only after path-finding completed. Meanwhile, `getsentinfo` would return an empty array, so users cannot know that the payment is being attempted. This could lead to users eagerly retrying and double-paying.

We now store an entry in the DB as early as possible, and remove it once the payment is complete (unless the payment failed without sending any HTLC, otherwise `getsentinfo` would return an empty array).

However, this doesn't entirely fix the risk of double-paying, which cannot be completely fixed from inside eclair. External applications that make payments on behalf of users must use the `externalId` and check for double payments themselves before calling eclair if they want to ensure that there's no risk of double payments.

One drawback of this approach that must be properly documented is that while the payment is pending, if the user calls `getsentinfo`, there will be a "global" entry (with `id == parentId`) which could make the user think that there are too many pending outgoing payments and eclair is overpaying. We should document that an entry with `id == paymentId` is a summary of the whole payment, but only entries with `id != parentId` correspond to actual HTLCs. I don't think this is a big issue, users shouldn't be interpreting pending payments on-the-fly and reacting to it, and that "global" entry is removed whenever the payment completes.

Fixes #2055

### DIFF
--- a/docs/release-notes/eclair-vnext.md
+++ b/docs/release-notes/eclair-vnext.md
@@ -90,6 +90,13 @@ The detailed balance json format has been slightly updated for channels in state
 Amounts corresponding to incoming htlcs for which we knew the preimage were previously included in `toLocal`, they are
 now grouped with outgoing htlcs amounts and the field has been renamed from `htlcOut` to `htlcs`.
 
+#### `getsentinfo`
+
+The `getsentinfo` API has been updated to avoid returning an empty array while a multi-part payment is starting (#2126).
+While a payment is still pending, `getsentinfo` will now return an additional "virtual" entry that summarizes the whole payment.
+This entry uses the same value for `id` and `parentId`, which makes it easy to detect it.
+Once the payment is complete, this entry will be removed (unless it's the only entry, which can happen when there are no routes available to send the payment).
+
 #### Miscellaneous
 
 This release contains many other API updates:
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
@@ -361,6 +361,11 @@ case class DualPaymentsDb(sqlite: SqlitePaymentsDb, postgres: PgPaymentsDb) exte
     sqlite.updateOutgoingPayment(paymentResult)
   }
 
+  override def completeOutgoingPayment(parentId: UUID): Unit = {
+    runAsync(postgres.completeOutgoingPayment(parentId))
+    sqlite.completeOutgoingPayment(parentId)
+  }
+
   override def getOutgoingPayment(id: UUID): Option[OutgoingPayment] = {
     runAsync(postgres.getOutgoingPayment(id))
     sqlite.getOutgoingPayment(id)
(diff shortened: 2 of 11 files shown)


pm47 (commented):
I can't say I love this hack, especially the post-hoc `DELETE` from database.

Have you considered introducing a new `OutgoingPaymentStatus` for that purpose? e.g. `OutgoingPaymentStatus.MPP_PENDING`?

t-bast:
> I can't say I love this hack, especially the post-hoc DELETE from database.

I'm not a big fan of it either, but I don't see how else to fix this with our model...
An alternative would be to introduce a separate table, but it won't fundamentally change the fact that we'll need to detect when a payment attempt is complete and update (or delete) a row in that new table.

> Have you considered introducing a new OutgoingPaymentStatus for that purpose? e.g. OutgoingPaymentStatus.MPP_PENDING?

What would that change?
Do you mean that we would use this status only on the "meta" row that describes the whole payment?
Even if we did, we would still need to update this status when the payment completes, right?
So it's quite equivalent to deleting the row?

The fundamental issue here is that in our model, we only store rows in the DB for HTLCs that are sent out.
If we want a way to know that a payment is pending before HTLCs are actually sent out, we need a different mechanism.
I think that whatever concrete mechanism we choose, it will have to do the following:

- store something to mark the payment pending at the beginning of the MultiPartPaymentLifecycle
- detect when the payment is complete and update the status stored in the first step

Whether the second step is an update (setting a new status) or a delete feels equivalent to me, but I could be missing something.

pm47:
> Do you mean that we would use this status only on the "meta" row that describes the whole payment?
Even if we did, we would still need to update this status when the payment completes, right?
So it's quite equivalent to deleting the row?

Nevermind, I thought we were storing the status and I wanted to replace the following query by a much simpler `DELETE payments.sent WHERE id = ? AND status='MPP_PENDING'`:
https://github.com/ACINQ/eclair/blob/cd6919114dcbc69704e18a31a5ef1f9477611e4a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgPaymentsDb.scala#L144

How about a different approach:
- we maintain a Map[UUID, PendingPayment] for all children of `PaymentInitiator`
- we add the following request/reply to `PaymentInitiator`:
```scala
  case class GetPayment(id: Either[UUID, ByteVector32])
  case class GetPaymentResponse(paymentId: UUID, paymentHash: ByteVector32, cfg: SendPaymentConfig)
```
  we keep only one map indexed on the `paymentId` and iterate over the values if queried with `paymentHash`
- in `Eclair.scala` we implement `def sentInfo()` as:
  1. `db.payments.listOutgoingPayments()`
  2. if db returns nothing then query `PaymentInitiator`
  3. if `PaymentInitiator` returns something then build a dummy response


Advantages:
- no db hack
- makes sense to track children of `PaymentInitiator` anyway, and consistent across all payments
- "perfect fix" (no possible race condition)
- we solve the api issue at the api level

t-bast:
Thanks, I'll try that approach in a separate PR, it will be interesting to compare the two

t-bast:
Done in #2142, I agree that it's better, it doesn't need any DB change which is a big win.

t-bast:
Replaced by #2142

PR #2132, opened 2022-01-13
Fix onion message tests by changing the rate limiter

thomash-acinq:
We want to be able to send several messages immediately after opening the connection which is not supported by Google's rate limiter that assumes that we had already reached the maximum rate when the rate limiter was created.
We create our own rate limiter that does what we need.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -358,12 +358,7 @@ eclair {
     # Transient connections opened to relay messages will be closed after this delay of inactivity
     kill-transient-connection-after = 30 seconds
 
-    rate-limit-per-peer {
-      number = 10
-      per = 1 second
-    }
-    max-per-peer = 10
-
+    max-per-peer-per-second = 10
 
     # Consider a message to be lost if we haven't received a reply after that amount of time
     reply-timeout = 20 seconds
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -358,7 +358,12 @@ eclair {
     # Transient connections opened to relay messages will be closed after this delay of inactivity
     kill-transient-connection-after = 30 seconds
 
-    max-per-peer-per-second = 10
+    rate-limit-per-peer {
+      number = 10
+      per = 1 second
+    }
+    max-per-peer = 10
+
 
     # Consider a message to be lost if we haven't received a reply after that amount of time
     reply-timeout = 20 seconds
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/RateLimiter.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/RateLimiter.scala
@@ -16,19 +16,22 @@
 
 package fr.acinq.eclair.io
 
-import scala.concurrent.duration.FiniteDuration
-
-class RateLimiter(n: Int, per: FiniteDuration) {
-  val perMs: Long = per.toMillis
-
+/**
+ * @param qps maximum number or messages allowed per second
+ * This class is not thread-safe, the rate limiter must not be shared between actors.
+ */
+class RateLimiter(qps: Int) {
   var i = 0
-  val accepted: Array[Long] = Array.fill(n)(0)
+  val accepted: Array[Long] = Array.fill(qps)(0)
 
+  /**
+  * [[tryAcquire()]] is guaranteed to return `true` at most `qps` times per second.
+  */
   def tryAcquire(): Boolean = {
     val now = System.currentTimeMillis()
-    if (now - accepted(i) > perMs) {
+    if (now - accepted(i) > 1000) {
       accepted(i) = now
-      i = (i + 1) % n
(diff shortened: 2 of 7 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/integration/MessageIntegrationSpec.scala
```diff
@@ -65,6 +71,7 @@ class MessageIntegrationSpec extends IntegrationSpec {
     val probe = TestProbe()
     val eventListener = TestProbe()
     nodes("B").system.eventStream.subscribe(eventListener.ref, classOf[OnionMessages.ReceiveMessage])
+    sleep(100 millis)
```
sstone, line 74: this would work, another approach would be to test that the subscription is active: create a dummy  `OnionMessages.ReceiveMessage` and publish it in a loop until `eventListener.reveiceOne() == dummy`

t-bast: I agree with sstone that we need something more robust.
Nothing guarantees you than 100 milliseconds is enough, this is an arbitrary value.
It would be much better to find a business logic condition to test for, that will guarantee that your nodes are in the right state.

The trick that sstone shared is one we've already use in a few places (but I don't remember where exactly).

eclair-core/src/test/scala/fr/acinq/eclair/integration/MessageIntegrationSpec.scala
```diff
@@ -94,6 +102,7 @@ class MessageIntegrationSpec extends IntegrationSpec {
     val probe = TestProbe()
     val eventListener = TestProbe()
     nodes("A").system.eventStream.subscribe(eventListener.ref, classOf[OnionMessages.ReceiveMessage])
+    sleep(100 millis)
     bob.sendOnionMessage(Nil, Left(nodes("A").nodeParams.nodeId), Some(Nil), hex"3f00").pipeTo(probe.ref)
```
sstone, line 106: There should be a check that `probe` did receive a  `SendOnionMessageResponse` here ?

eclair-core/src/test/scala/fr/acinq/eclair/integration/MessageIntegrationSpec.scala
```diff
@@ -77,6 +84,7 @@ class MessageIntegrationSpec extends IntegrationSpec {
     val probe = TestProbe()
     val eventListener = TestProbe()
     nodes("B").system.eventStream.subscribe(eventListener.ref, classOf[OnionMessages.ReceiveMessage])
+    sleep(100 millis)
```
sstone, line 87: There should be a check that probe did receive a SendOnionMessageResponse afer `alice.sendOnionMessage` just below ?

thomash-acinq:
After investigating, the problem was actually that the rate limiter wouldn't let us send two messages immediately after creating a connection.

t-bast:
Good catch, well done finding that one!

eclair-core/src/main/resources/reference.conf
```diff
@@ -358,7 +358,12 @@ eclair {
     # Transient connections opened to relay messages will be closed after this delay of inactivity
     kill-transient-connection-after = 30 seconds
 
-    max-per-peer-per-second = 10
+    rate-limit-per-peer {
+      number = 10
+      per = 1 second
+    }
+    max-per-peer = 10
```
t-bast, line 365: Why do you need to separate `number` and `per`?
You use it with `1 second` everywhere, so can't we just keep the previous `max-per-peer-per-second`?

eclair-core/src/main/scala/fr/acinq/eclair/io/RateLimiter.scala
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fr.acinq.eclair.io
+
+import scala.concurrent.duration.FiniteDuration
+
+class RateLimiter(n: Int, per: FiniteDuration) {
```
t-bast, line 21: It's important to comment here that this is **not thread-safe**.
It must only be used within an actor and must not be shared (which is totally ok, this is how we use it currently).

t-bast (commented):
I like the simple design for this rate-limiter, well done.
I just think we can simplify the configuration to a single value, as it was previously.

eclair-core/src/main/resources/reference.conf
thomash-acinq: I've switched back to `max-per-peer-per-second`. It gives us less flexibility but I guess it's fine.

eclair-core/src/main/scala/fr/acinq/eclair/io/RateLimiter.scala
thomash-acinq: Done

t-bast (approved):
LGTM :+1:

PR #2135, opened 2022-01-19
Update bitcoind to 22.0

t-bast:
There was no release 22.1 made this time, and it looks like bitcoin core is focusing on the upcoming 23.0 release, so we should support 22.0.

We also drop support for versions < 0.21.1, which lets us simplify some of our tx publishing code (that worked around a bitcoind bug).

TODO: directly jump to bitcoind 23, which contains important bug fixes for fee-bumping!

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ You will find detailed guides and frequently asked questions there.
 
 ### Prerequisite: Bitcoin Core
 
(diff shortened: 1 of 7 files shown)


sstone:
It seems that there is a 22.1 milestone (see https://github.com/bitcoin/bitcoin/milestone/55) but I could not find a release date

PR #2137, opened 2022-01-20
Add random speed up to replaceable tx publishing

t-bast:
Now that #2113 is out of the way, we can start tweaking and making small improvements to the fee-bumping mechanism.

A deterministic schedule for confirmation during force-close scenarios could in theory help malicious peers starve us of our utxos in a high-fee environment, making it hard to fee-bump HTLC transactions.

Adding some randomness into the mix to non-deterministically speed up confirmations (which gives us newly confirmed utxos to bump other txs) shouldn't hurt.


t-bast:
Replaced by #2141

PR #2140, opened 2022-01-21
Postgres: add safety checks at startup

pm47:
When using postgres, at startup we optionnally run a few basic safety checks, e.g. the number of local channels, how long since the last local channel update, etc. The goal is to make sure that we are connected to the correct database instance.

### DIFF
--- a/docs/release-notes/eclair-vnext.md
+++ b/docs/release-notes/eclair-vnext.md
@@ -75,6 +75,27 @@ This is particularly useful for payment hubs that generate a lot of invoices (e.
 Eclair includes a small `payment_metadata` field in all invoices it generates.
 This lets node operators verify that payers actually support that feature.
 
+### Optional safety checks when using Postgres
+
+When using postgres, at startup we optionnally run a few basic safety checks, e.g. the number of local channels, how long since the last local channel update, etc. The goal is to make sure that we are connected to the correct database instance.
+
+Those checks are configured as follows and are disabled by default:
+```
+eclair.db.postgres.safety-checks {
+  enabled = false
+  max-age {
+    local-channels = 3 minutes
+    network-nodes = 30 minutes
+    audit-relayed = 10 minutes
+  }
+  min-count {
+    local-channels = 10
+    network-nodes = 3000
+    network-channels = 20000
+  }
+}
+```
+
 ### API changes
 
 #### Timestamps
--- a/docs/release-notes/eclair-vnext.md
+++ b/docs/release-notes/eclair-vnext.md
@@ -77,12 +77,13 @@ This lets node operators verify that payers actually support that feature.
 
 ### Optional safety checks when using Postgres
 
-When using postgres, at startup we optionnally run a few basic safety checks, e.g. the number of local channels, how long since the last local channel update, etc. The goal is to make sure that we are connected to the correct database instance.
+When using postgres, at startup we optionally run a few basic safety checks, e.g. the number of local channels, how long since the last local channel update, etc. The goal is to make sure that we are connected to the correct database instance.
 
-Those checks are configured as follows and are disabled by default:
+Those checks are disabled by default because they wouldn't pass on a fresh new node with zero channels. You should enable them when you already have channels, so that there is something to compare to, and the values should be specific to your setup, particularly for local channels. Configuration is done by overriding `max-age` and `min-count` values in your `eclair.conf`:  
 ```
-eclair.db.postgres.safety-checks {
-  enabled = false
+eclair.db.postgres.safety-checks
+ {
+  enabled = true
   max-age {
     local-channels = 3 minutes
     network-nodes = 30 minutes
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -328,17 +328,22 @@ eclair {
         auto-release-at-shutdown = true // automatically release the lock when eclair is stopping
       }
       safety-checks {
-        // a set of basic checks on data to make sure we use the correct database
+        // A set of basic checks on data to make sure we use the correct database
+        // Those checks are disabled by default because they wouldn't pass on a fresh new node with
+        // zero channels. You should enable them when you already have channels, so that there is
+        // something to compare to, and the values should be specific to your setup, especially
+        // for local channels. If your operate a busy node, you can reduce max-age.local-channels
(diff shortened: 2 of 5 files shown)


docs/release-notes/eclair-vnext.md
```diff
@@ -75,6 +75,27 @@ This is particularly useful for payment hubs that generate a lot of invoices (e.
 Eclair includes a small `payment_metadata` field in all invoices it generates.
 This lets node operators verify that payers actually support that feature.
 
+### Optional safety checks when using Postgres
+
+When using postgres, at startup we optionnally run a few basic safety checks, e.g. the number of local channels, how long since the last local channel update, etc. The goal is to make sure that we are connected to the correct database instance.
```
t-bast, line 80: nit:

```suggestion
When using postgres, at startup we optionally run a few basic safety checks, e.g. the number of local channels, how long since the last local channel update, etc. The goal is to make sure that we are connected to the correct database instance.
```

eclair-core/src/main/resources/reference.conf
```diff
@@ -327,6 +327,20 @@ eclair {
         lock-timeout = 5 seconds // timeout for the lock statement on the lease table
         auto-release-at-shutdown = true // automatically release the lock when eclair is stopping
       }
+      safety-checks {
+        // a set of basic checks on data to make sure we use the correct database
+        enabled = false
+        max-age {
+          local-channels = 3 minutes
+          network-nodes = 30 minutes
+          audit-relayed = 10 minutes
```
t-bast, line 336: It's not very clear from this file alone what those refer to.
Can you add comments to each of them to help node operators properly configure it?

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -162,6 +174,71 @@ object Databases extends Logging {
...
+
+            def checkMaxAge(name: String, maxAge: FiniteDuration, sqlQuery: String): Unit = {
+              import ExtendedResultSet._
+              val smallestAge_opt = statement
+                .executeQuery(sqlQuery)
+                .headOption // sql max() will always return a result, with perhaps a null value if there was no records
+                .flatMap(_.getTimestampNullable("max"))
+                .map(ts => TimestampMilli.now() - TimestampMilli.fromSqlTimestamp(ts))
+              require(smallestAge_opt.isDefined, s"db check failed: no $name found")
+              require(smallestAge_opt.get <= maxAge, s"db check failed: most recent $name is too old (${smallestAge_opt.get.toMinutes} minutes > ${maxAge.toMinutes} minutes)")
```
t-bast, line 190: nit: I thought we chose to use `sys.exit` for that kind of scenarios (it's easier to search for them in the codebase than filter all `require`). Same for `checkMinCount`.

Also, in my understanding `require` should be used mostly to check for logic bugs, incorrect inputs or invalid assumptions (so the check that `smallestAge_opt.isDefined` does make sense as a `require`).

But it would make it much harder to unit test...so probably better to ignore this ¯\_(ツ)_/¯

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -162,6 +174,71 @@ object Databases extends Logging {
...
+              logger.info(s"db check ok: max age ${smallestAge_opt.get.toMinutes} minutes <= ${maxAge.toMinutes} minutes for $name")
+            }
+
+            checkMaxAge(name = "local channel",
```
t-bast, line 194: This check will always fail for a node that has no channels.
Would it make sense to skip it if we don't have any channels or if they are all closed?
Otherwise node operators will need to first disable the checks in their config, open channels, then re-enable it.

I think it's fine to let this check fail when eclair has no channel, but then we should probably mention it in `reference.conf` and tell node operators to only enable the checks once they expect their node to be a somewhat busy node.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -162,6 +174,71 @@ object Databases extends Logging {
...
+                """
+                  |SELECT MAX(timestamp)
+                  |FROM audit.relayed""".stripMargin)
+
+            def checkMinCount(name: String, minCount: Int, sqlQuery: String): Unit = {
+              import ExtendedResultSet._
+              val count = statement
+                .executeQuery(sqlQuery)
+                .map(_.getInt("count"))
+                .head
```
t-bast, line 221: nit: add a comment here saying that `COUNT(*)` will always return one row?

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
pm47: We're not very consistent in that file. I could `throw` instead.

docs/release-notes/eclair-vnext.md
pm47: Done with cd1905115.

eclair-core/src/main/resources/reference.conf
pm47: Done with cd1905115.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
pm47: > This check will always fail for a node that has no channels.

Yes, that's why it's disabled by default.

> I think it's fine to let this check fail when eclair has no channel, but then we should probably mention it in reference.conf and tell node operators to only enable the checks once they expect their node to be a somewhat busy node.

This check only makes sens for an already established node, because the values are entirely dependent on how big and how busy the node is. To be more precise, routing table-related values are independent, but they are much less important. What matters is local channels. This was missing from the doc, I've added a paragraph in cd1905115 to hopefully make it more clear.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
pm47: Done with 4ebd783e3.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
t-bast: Let's keep it like that, it's not very important.

t-bast (approved):

PR #2141, opened 2022-01-21
More aggressive confirmation target when low on utxos

t-bast:
Now that #2113 is out of the way, we can start tweaking and making small improvements to the fee-bumping mechanism.

An optimistic schedule for confirmation during force-close scenarios could in theory help malicious peers starve us of our utxos, making it hard to fee-bump HTLC transactions in high-fee environments.

If we detect that our safe utxos count is low, we try to get force-close transactions confirmed as quickly as possible, which will yield a new utxo that can be used to fund other transactions.

This PR is a deterministic alternative to #2137

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
@@ -75,8 +75,8 @@ object ReplaceableTxPublisher {
   }
 
   def getFeerate(feeEstimator: FeeEstimator, confirmBefore: BlockHeight, currentBlockHeight: BlockHeight, hasEnoughSafeUtxos: Boolean): FeeratePerKw = {
+    val remainingBlocks = confirmBefore - currentBlockHeight
     if (hasEnoughSafeUtxos) {
-      val remainingBlocks = confirmBefore - currentBlockHeight
       val blockTarget = remainingBlocks match {
         // If our target is still very far in the future, no need to rush
         case t if t >= 144 => 144
@@ -91,7 +91,11 @@ object ReplaceableTxPublisher {
       feeEstimator.getFeeratePerKw(blockTarget)
     } else {
       // We don't have many safe utxos so we want the transaction to confirm quickly.
-      feeEstimator.getFeeratePerKw(1)
+      if (remainingBlocks <= 1) {
+        feeEstimator.getFeeratePerKw(1)
+      } else {
+        feeEstimator.getFeeratePerKw(2)
+      }
     }
   }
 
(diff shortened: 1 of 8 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
```diff
@@ -73,20 +74,25 @@ object ReplaceableTxPublisher {
...
+        // However, if we get closer to the target, we start being more aggressive
+        case t if t >= 18 => 12
+        case t if t >= 12 => 6
+        case t if t >= 2 => 2
+        case _ => 1
+      }
+      feeEstimator.getFeeratePerKw(blockTarget)
+    } else {
+      // We don't have many safe utxos so we want the transaction to confirm quickly.
+      feeEstimator.getFeeratePerKw(1)
```
pm47, line 94: I know it makes sense to target one block here, but I'm afraid we overpay. How about two blocks?

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
```diff
@@ -347,5 +374,10 @@ private class ReplaceableTxPublisher(nodeParams: NodeParams,
     }
   }
 
+  /** If we don't have a lot of safe utxos left, we will use an aggressive feerate to ensure our utxos aren't locked for too long. */
+  private def hasEnoughSafeUtxos(threshold: Int): Future[Boolean] = {
+    bitcoinClient.listUnspent().map(_.count(utxo => utxo.safe && utxo.amount >= 10_000.sat) >= threshold)
```
pm47, line 379: We could have made the threshold value configurable, but this is short term.

pm47 (approved):
This is easier to understand and more foolproof than #2137, and also more optimal in a normal low fees scenario.

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
t-bast: Yes, I hesitated but I don't think it's worth it, unless we hit a roadblock on the future use of unsafe utxos.

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxPublisher.scala
t-bast: Let's do 2 blocks, unless the confirmation target is in less than 6 blocks, in which case we do 1 block, ok?

t-bast: Done in https://github.com/ACINQ/eclair/pull/2141/commits/901b565fd492b81ea47400a9a13ed675e6e72566

pm47 (approved):

PR #2142, opened 2022-01-21
Improve `getsentinfo` accuracy

### LINKED ISSUE #2055: `getsentinfo` returns an indeterminate response `[]`

t-bast:
When using MPP, we previously inserted an entry in the payments DB only after path-finding completed. Meanwhile, `getsentinfo` would return an empty array, so users cannot know that the payment is being attempted. This could lead to users eagerly retrying and double-paying.

We now keep track of pending payments in the `PaymentInitiator`. When the payment doesn't seem to be pending based on the DB alone, we also check the `PaymentInitiator`.

However, this doesn't entirely fix the risk of double-paying, which cannot be completely fixed from inside eclair. External applications that make payments on behalf of users must use the `externalId` and check for double payments themselves before calling eclair if they want to ensure that there's no risk of double payments.

Each commit is independent, as I took this opportunity to do a bit of clean up.

Replaces #2126 to fix #2055

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -34,7 +34,7 @@ import fr.acinq.eclair.blockchain.fee.{FeeratePerByte, FeeratePerKw}
 import fr.acinq.eclair.channel._
 import fr.acinq.eclair.crypto.Sphinx
 import fr.acinq.eclair.db.AuditDb.{NetworkFee, Stats}
-import fr.acinq.eclair.db.{IncomingPayment, OutgoingPayment}
+import fr.acinq.eclair.db.{IncomingPayment, OutgoingPayment, OutgoingPaymentStatus}
 import fr.acinq.eclair.io.Peer.{GetPeerInfo, PeerInfo}
 import fr.acinq.eclair.io._
 import fr.acinq.eclair.message.{OnionMessages, Postman}
@@ -42,7 +42,7 @@ import fr.acinq.eclair.payment._
 import fr.acinq.eclair.payment.receive.MultiPartHandler.ReceivePayment
 import fr.acinq.eclair.payment.relay.Relayer.{GetOutgoingChannels, OutgoingChannels, RelayFees, UsableBalance}
 import fr.acinq.eclair.payment.send.MultiPartPaymentLifecycle.PreimageReceived
-import fr.acinq.eclair.payment.send.PaymentInitiator.{SendPaymentToNode, SendPaymentToRoute, SendPaymentToRouteResponse, SendSpontaneousPayment}
+import fr.acinq.eclair.payment.send.PaymentInitiator._
 import fr.acinq.eclair.router.Router
 import fr.acinq.eclair.router.Router._
 import fr.acinq.eclair.wire.protocol.MessageOnionCodecs.blindedRouteCodec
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -60,11 +60,13 @@ class PaymentInitiator(nodeParams: NodeParams, outgoingPaymentFactory: PaymentIn
           sender() ! PaymentFailed(paymentId, r.paymentHash, LocalFailure(r.recipientAmount, Nil, PaymentSecretMissing) :: Nil)
         case Some(paymentSecret) if r.paymentRequest.features.allowMultiPart && nodeParams.features.hasFeature(BasicMultiPartPayment) =>
           val fsm = outgoingPaymentFactory.spawnOutgoingMultiPartPayment(context, paymentCfg)
-          fsm ! SendMultiPartPayment(sender(), paymentSecret, r.recipientNodeId, r.recipientAmount, finalExpiry, r.maxAttempts, r.paymentRequest.paymentMetadata, r.assistedRoutes, r.routeParams, userCustomTlvs = r.userCustomTlvs)
+          fsm ! SendMultiPartPayment(self, paymentSecret, r.recipientNodeId, r.recipientAmount, finalExpiry, r.maxAttempts, r.paymentRequest.paymentMetadata, r.assistedRoutes, r.routeParams, userCustomTlvs = r.userCustomTlvs)
```
pm47, line 63: An alternative to putting `replyTo = self` would have been to preserve the original sender and use a `watch`. In theory there would be a tiny performance gain, but I suppose you have favored consistency between trampoline and non-trampoline payments.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -375,11 +375,30 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
     }
   }
 
-  override def sentInfo(id: Either[UUID, ByteVector32])(implicit timeout: Timeout): Future[Seq[OutgoingPayment]] = Future {
-    id match {
+  override def sentInfo(id: Either[UUID, ByteVector32])(implicit timeout: Timeout): Future[Seq[OutgoingPayment]] = {
+    val outgoingDbPayments = id match {
```
pm47, line 379: The database call should be done within a `Future`, like before. It forces you to compose futures which is not very elegant but I think it's really important.

pm47 (commented):
I like this version better. Given how far you went in the implementation I think you too?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
t-bast: I've always wanted to get rid of setting `replyTo` to `sender`, it was the simple thing to do but it introduced the annoying side effect that senders received the `PreimageReceived` event when using MPP, whereas they were only interested in the `PaymentSent` event. This was a good opportunity to fix that (and make trampoline and non-trampoline more consistent as well)!

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: Good catch, done in https://github.com/ACINQ/eclair/pull/2142/commits/bf8a61a7937712e49b41b2c97494196843c4e60f

t-bast:
> I like this version better. Given how far you went in the implementation I think you too?

100% agree, this is much better, I don't have the reflex of putting more logic inside the API handlers themselves, but it makes a lot of sense in that case.

pm47 (approved):

fiatjaf:
Does this mean there will never again be an empty `[]` if a payment is being tried? What will it return if the payment is being tried but not written to the DB yet?

What is this risk of double-payments you mention? Paying the same invoice twice?

t-bast:
> Does this mean there will never again be an empty [] if a payment is being tried? 

Yes, there will always be something returned if you use a `pay` command then call `getsentinfo` on the id it returns.

> What will it return if the payment is being tried but not written to the DB yet?

It will return the same structure as it did if the payment was pending, but without route details (since no htlcs have been sent yet) and with `id == parentId`.

> What is this risk of double-payments you mention? Paying the same invoice twice?

Actually this completely solves the issue you had.
The only potential race condition that remains and cannot be fixed from within eclair would be:

- you call `payinvoice`
- you call `getsentinfo --paymentHash=<invoice_payment_hash>` before the previous call returned
- then it's not guaranteed that you will see a result

But as long as you wait for `payinvoice` (or other `pay` APIs) to return you the ID, eclair should now always return something in `getsentinfo`. I believe that completely fixes the issue you had.

PR #2146, opened 2022-01-25
Front: use IMDSv2 to retrieve instance ip

pm47:
This is used by the built-in AWS Beanstalk package.

See
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html.


t-bast (commented):
Can you add the documentation link in the script?

PR #2148, opened 2022-01-25
Type `ChannelFlags` instead of using a raw `Byte`

pm47:
This is an itch I have been meaning to scratch for a long time.

⚠️ The codec update is trivial but a bug would be very bad, please review that part extra carefully.

This has the same kind of impact as the typing of `Features`. In particular we can now explicitly set the values in `eclair.conf`:
`eclair.channel.channel-flags.announce-channel=true`. I took the opportunity to move this channel-related config key in its own config section, with the goal that we move the other fields in a follow-up PR.

It also has the nice side effect of providing a pretty json formatting out of the box:

```json
"channelFlags": {
  "announceChannel": true
}
```

instead of:

```json
"channelFlags": 1
```

The `open` method of the API has been updatedi: `channelFlags` has been replaced by a new `announceChannel` form param is available.


eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -139,7 +139,7 @@ object TestConstants {
       initialRandomReconnectDelay = 5 seconds,
       maxReconnectInterval = 1 hour,
       chainHash = Block.RegtestGenesisBlock.hash,
-      channelFlags = 1,
+      channelFlags = ChannelFlags(announceChannel = true),
```
t-bast, line 142: nit: why not having a `ChannelFlags.Public` instance the way you defined a `ChannelFlags.Private`?

eclair-core/src/test/scala/fr/acinq/eclair/json/JsonSerializersSpec.scala
```diff
@@ -151,6 +151,14 @@ class JsonSerializersSpec extends AnyFunSuite with Matchers {
     JsonSerializers.serialization.write(inputInfo)(JsonSerializers.formats) shouldBe """{"outPoint":"9f0b9c0ce92c175ca4e78acfd13a718099c73818b6d3140cfe6f04ec052b5b34:42","amountSatoshis":456651}"""
   }
 
+  test("channel flags serialization") {
+    //object ChannelFlags {
+    //  val AnnounceChannel = 0x01.toByte
+    //  val Empty = 0x00.toByte
+    //}
+    // val announceChannel: Boolean = (channelFlags & 0x01) != 0
+  }
```
t-bast, line 160: Looks like you're missing that test ;)

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
```diff
@@ -371,7 +371,8 @@ object PaymentPacketSpec {
     val params = LocalParams(null, null, null, null, null, null, null, 0, isFunder = true, null, None, null)
     val remoteParams = RemoteParams(randomKey().publicKey, null, null, null, null, null, maxAcceptedHtlcs = 0, null, null, null, null, null, null, None)
     val commitInput = InputInfo(OutPoint(randomBytes32(), 1), TxOut(testCapacity, Nil), Nil)
-    new Commitments(channelId, ChannelConfig.standard, ChannelFeatures(), params, remoteParams, 0.toByte, null, null, null, null, 0, 0, Map.empty, null, commitInput, null) {
+    val channelFlags = ChannelFlags(announceChannel = false)
```
t-bast, line 374: nit: `ChannelFlags.Private`?

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Channel.scala
```diff
@@ -33,8 +33,8 @@ trait Channel {
   import fr.acinq.eclair.api.serde.JsonSupport.{formats, marshaller, serialization}
 
   val open: Route = postRequest("open") { implicit t =>
-    formFields(nodeIdFormParam, "fundingSatoshis".as[Satoshi], "pushMsat".as[MilliSatoshi].?, "channelType".?, "fundingFeerateSatByte".as[FeeratePerByte].?, "channelFlags".as[Int].?, "openTimeoutSeconds".as[Timeout].?) {
-      (nodeId, fundingSatoshis, pushMsat, channelType, fundingFeerateSatByte, channelFlags, openTimeout_opt) =>
+    formFields(nodeIdFormParam, "fundingSatoshis".as[Satoshi], "pushMsat".as[MilliSatoshi].?, "channelType".?, "fundingFeerateSatByte".as[FeeratePerByte].?, "channelFlags".as[Int].?, "announceChannel".as[Boolean].?, "openTimeoutSeconds".as[Timeout].?) {
```
t-bast, line 36: I would just remove the previous `channelFlags` parameter.
We will mention it in the release notes and I'll update the API documentation once the release is out.

t-bast (commented):
LGTM, a few nits.
The codec change doesn't seem very risky and is well tested.

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
pm47: I flipped-flopped on this. Those shortcuts woudn't make sense if we had more than one bit set.

Put it back in 8d8ee2f2f.

eclair-core/src/test/scala/fr/acinq/eclair/json/JsonSerializersSpec.scala
pm47: Yeah, I realized later that I didn't need a custom serializer... Besides, the serialization is actually tested by the `backward compatibility older codecs (integrity)`.

Remove this draft in 8d8ee2f2f.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
pm47: Done in 8d8ee2f2f.

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Channel.scala
pm47: Done in 8d8ee2f2f and updated the release note.

t-bast (approved):

PR #2150, opened 2022-01-27
Add support for funding txs with unsafe inputs

t-bast:
This feature was added in bitcoin core 22.0 (see https://github.com/bitcoin/bitcoin/pull/21359).
We can use it to reduce the requirements on safe utxos for anchor outputs fee bumping, once we deprecate support for bitcoin core < 22.0.

This PR adds support at the `BitcoinCoreClient` level, but doesn't actually use it yet.
It builds on top of #2135


t-bast:

PR #2151, opened 2022-01-27
Add support for funding txs with non-wallet inputs

t-bast:
This feature was added in bitcoin core 23.0 (see https://github.com/bitcoin/bitcoin/pull/23201).
We can use it to greatly simplify our tx publishing logic, once we deprecate support for bitcoin core < 23.0.

This PR builds on top of #2150 and must wait for bitcoind 23.0.


t-bast:

PR #2152, opened 2022-01-27
Make safety checks independent from db backend

pm47:
Should we rename `safety-checks` to `init-checks`?


pm47:
Closed in favor of #2153.

PR #2153, opened 2022-01-27
Have sqlite also write to jdbc url file

pm47:
This is an alternative to #2152.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
@@ -73,15 +73,18 @@ object Databases extends Logging {
   }
 
   object SqliteDatabases {
-    def apply(auditJdbc: Connection, networkJdbc: Connection, eclairJdbc: Connection): SqliteDatabases = SqliteDatabases(
-      network = new SqliteNetworkDb(networkJdbc),
-      audit = new SqliteAuditDb(auditJdbc),
-      channels = new SqliteChannelsDb(eclairJdbc),
-      peers = new SqlitePeersDb(eclairJdbc),
-      payments = new SqlitePaymentsDb(eclairJdbc),
-      pendingCommands = new SqlitePendingCommandsDb(eclairJdbc),
-      backupConnection = eclairJdbc
-    )
+    def apply(auditJdbc: Connection, networkJdbc: Connection, eclairJdbc: Connection, jdbcUrlFile_opt: Option[File]): SqliteDatabases = {
+      jdbcUrlFile_opt.foreach(checkIfDatabaseUrlIsUnchanged("sqlite", _))
+      SqliteDatabases(
+        network = new SqliteNetworkDb(networkJdbc),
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -241,19 +244,19 @@ object Databases extends Logging {
...
-        val oldUrl = readString(urlFile.toPath)
-        if (oldUrl != url)
-          throw JdbcUrlChanged(oldUrl, url)
-      } else {
-        writeString(urlFile.toPath, url)
-      }
+    if (urlFile.exists()) {
+      val oldUrl = readString(urlFile.toPath)
+      if (oldUrl != url)
+        throw JdbcUrlChanged(oldUrl, url)
```
t-bast, line 257: Do we want to slightly update this to make the transition from `sqlite` to `postgres` more user friendly?
Right now the error they would get would be "The database URL has changed since the last start. It was `sqlite`, now it's `<some_postgres_url>`".

Maybe we could add a condition:

```
if (oldUrl != url) {
    if (oldUrl == "sqlite" || url == "sqlite") throw DbTypeChanged(<something>)
    else throw JdbcUrlChanged(oldUrl, url)
}
```

Where `DbTypeChanged` would have a message saying that switching from one type of DB to another is dangerous, make sure you have migrated the data manually first, etc?

pm47: I went with something in-between in 9cb7edbab. I kept the existing exception, otherwise it forced us to list all possible cases, which was tedious, error prone, and not really useful: even when moving from postgres to postgres, the migration warning applies.

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
```diff
@@ -241,19 +244,22 @@ object Databases extends Logging {
 
       databases
     }
+  }
 
-    private def checkIfDatabaseUrlIsUnchanged(url: String, urlFile: File): Unit = {
-      def readString(path: Path): String = Files.readAllLines(path).get(0)
+  /** We raise this exception when the jdbc url changes, to prevent using a different server involuntarily. */
+  case class JdbcUrlChanged(before: String, after: String) extends RuntimeException(s"The database URL has changed since the last start. It was `$before`, now it's `$after`. If this was intended, make sure you have migrated your data, otherwise your channels will be force-closed and you may lose funds.")
```
t-bast, line 250: Very nit: maybe we can rename this to `DbUrlChanged` which is more generic?

t-bast (approved):

eclair-core/src/main/scala/fr/acinq/eclair/db/Databases.scala
pm47: I don't think it's worth it since we still have references to "jdbc" or even "url" elsewhere.

t-bast: Fair enough

PR #2154, opened 2022-01-27
Relay blinded payments

thomash-acinq:
Allow relaying blinded payments (https://github.com/lightning/bolts/pull/765)


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
```diff
@@ -66,26 +121,64 @@ object RouteBlindingEncryptedDataCodecs {
...
+    case tlvs => Attempt.successful(MessageRelayData(tlvs))
+  },{
+    case MessageRelayData(tlvs) => tlvs
+  })
 
-  private val encryptedDataTlvCodec = discriminated[RouteBlindingEncryptedDataTlv].by(varint)
+  private val paymentRelayDataTlvCodec = discriminated[RouteBlindingEncryptedDataTlv].by(varint)
     .typecase(UInt64(1), padding)
     .typecase(UInt64(2), outgoingChannelId)
-    .typecase(UInt64(4), outgoingNodeId)
```
t-bast, line 73: Actually you cannot remove this line, otherwise blinded trampoline routes don't work!

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
```diff
@@ -66,26 +121,64 @@ object RouteBlindingEncryptedDataCodecs {
...
+    case tlvs if tlvs.get[PaymentRelay].isEmpty => Attempt.failure(MissingRequiredTlv(UInt64(10)))
+    case tlvs => Attempt.successful(PaymentRelayData(tlvs))
+  }, {
+    case PaymentRelayData(tlvs) => tlvs
+  })
+
+  private val finalRecipientDataTlvCodec = discriminated[RouteBlindingEncryptedDataTlv].by(varint)
+    .typecase(UInt64(1), padding)
+    .typecase(UInt64(6), pathId)
+    .typecase(UInt64(12), paymentConstraints)
```
t-bast, line 165: This has the same issue that the `BlindOnionPaymentPayloadTlv` trait in `PaymentOnion.scala` has: you're encoding requirements in the codec (by removing the tlvs that shouldn't be provided for the final payload), but you're not actually encoding them all, because a final payload for a message shouldn't have a `paymentConstraints` field...

Similarly to my comment in `PaymentOnion.scala`, I think we should make the codecs and type hierarchy simpler here and instead do the validation in higher layers. At first I wouldn't even do that validation, if other tlvs are provided we can simply ignore them. Moreover, for the final payload, we generated that payload, so we know it cannot contain any unwanted tlv!

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/PaymentOnionSpec.scala
```diff
@@ -82,17 +82,18 @@ class PaymentOnionSpec extends AnyFunSuite {
   test("encode/decode variable-length (tlv) relay per-hop payload") {
     val testCases = Map(
       TlvStream[OnionPaymentPayloadTlv](AmountToForward(561 msat), OutgoingCltv(CltvExpiry(42)), OutgoingChannelId(ShortChannelId(1105))) -> hex"11 02020231 04012a 06080000000000000451",
-      TlvStream[OnionPaymentPayloadTlv](AmountToForward(561 msat), OutgoingCltv(CltvExpiry(42)), OutgoingChannelId(ShortChannelId(1105)), EncryptedRecipientData(hex"0123456789abcdef"), BlindingPoint(PublicKey(hex"036d6caac248af96f6afa7f904f550253a0f3ef3f5aa2fe6838a95b216691468e2"))) -> hex"3e 02020231 04012a 06080000000000000451 0a080123456789abcdef 0c21036d6caac248af96f6afa7f904f550253a0f3ef3f5aa2fe6838a95b216691468e2",
```
t-bast, line 85: We shouldn't just delete this, we should replace it with a new test vector that contains valid `encrypted_data`, shouldn't we?

t-bast (commented):
Nice work! I'm in favor of simplifying the type hierarchy and codec hierarchy, I'm afraid it makes the code more complex whereas we can be more lax at the wire layer and do the validation logic at higher layers (which also makes it easier to unit test).

Apart from that, the logic looks good and matches the spec. I'd really like to have at least a concept ACK from other implementers on the `payment_relay` and `payment_constraints` mechanism before we merge this on `master` though...and if we do have concept ACKs we should create a good set of test vectors which will hopefully be final!

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
```diff
@@ -51,10 +55,61 @@ object RouteBlindingEncryptedDataTlv {
...
+          cltvExpiry <= constraints.maxCltvExpiry &&
+          Features.areCompatible(features, constraints.allowedFeatures)
+      )
+    }
+  }
+
+  case class FinalRecipientData(records: TlvStream[RouteBlindingEncryptedDataTlv]) extends MessageData with PaymentData {
+    val pathId_opt: Option[ByteVector] = records.get[PathId].map(_.data)
+
+    def isValidPayment(amount: MilliSatoshi, cltvExpiry: CltvExpiry, features: Features[Feature]): Boolean = {
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
thomash-acinq: That's correct, blinded trampoline won't work. We need a lot more spec work to do it. For now I'm ignoring the problem.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
thomash-acinq: > because a final payload for a message shouldn't have a `paymentConstraints` field...

Why not? Actually, having `paymentConstraints` in the intermediary hops is just a way to make the route fail early but we shouldn't trust it, the important `paymentConstraints` is the one in the final payload.

> Moreover, for the final payload, we generated that payload, so we know it cannot contain any unwanted tlv!

No, anyone that has our public key can generate such a payload, it's not necessarily us.

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/PaymentOnionSpec.scala
thomash-acinq: I can't because this test checks the values of `amountToForward`, `outgoingCltv` and `outgoingChannelId` which are not available in the case of a blinded route. I've added back this test case in its own test for blinded routes.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
```diff
@@ -66,26 +121,64 @@ object RouteBlindingEncryptedDataCodecs {
...
   private val nextBlinding: Codec[NextBlinding] = (("length" | constant(hex"21")) :: ("blinding" | publicKey)).as[NextBlinding]
+  private val paymentRelay: Codec[PaymentRelay] = (
+    ("length" | constant(hex"0a")) ::
+    ("fee_base_msat" | millisatoshi32) ::
+      ("fee_proportional_millionths" | uint32) ::
+      ("cltv_expiry_delta" | cltvExpiryDelta)).as[PaymentRelay]
+  private val paymentConstraints: Codec[PaymentConstraints] = variableSizeBytesLong(varintoverflow,
+    ("max_cltv_expiry" | cltvExpiry) ::
+      ("htlc_minimum_msat" | millisatoshi) ::
+      ("allowed_features" | featuresCodec)).as[PaymentConstraints]
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
t-bast: Fair enough, it's indeed not an issue to disallow it for now since we'll need to add some code to implement blinded trampoline routes anyway, we'll add it when we're ready.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
t-bast: > Why not? Actually, having paymentConstraints in the intermediary hops is just a way to make the route fail early but we shouldn't trust it, the important paymentConstraints is the one in the final payload.

You're completely right, I got confused between payment constraints and payment info!

> No, anyone that has our public key can generate such a payload, it's not necessarily us.

Good point as well! We will validate the `pathId` though at a higher layer, so in practice this isn't an issue, is it?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -203,16 +205,6 @@ object PaymentOnion {
    * type of onion packet they can be used with (PacketType).
    */
 
-  sealed trait PerHopPayloadFormat
```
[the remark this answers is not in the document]
thomash-acinq: https://github.com/ACINQ/eclair/pull/2235

t-bast (commented):
I dived deep into that rabbit hole to see if we could simplify the type hierarchy and codecs a bit more (they're really tedious to read), let me know what you think of #2238

PR #2156, opened 2022-01-27
Database migration Sqlite->Postgres

pm47:
Add a migration tool and a step-by-step guide for users who want to switch from Sqlite to Postgres.

That's a lot of code, but it's always the same things. I suggest starting with `ComparePeersDb`/`MigratePeersDb`.

### DIFF
--- a/docs/PostgreSQL.md
+++ b/docs/PostgreSQL.md
@@ -111,4 +111,34 @@ Eclair stores the latest database settings in the `${data-dir}/last_jdbcurl` fil
 
 The node operator can force Eclair to accept new database 
 connection settings by removing the `last_jdbcurl` file. 
-    
+
+### Migrating from Sqlite to Postgres
+
+Eclair supports migrating your existing node from Sqlite to Postgres. Note that the opposite (from Postgres to Sqlite) is not supported.
+
+:warning: Once you have migrated from Sqlite to Postgres there is no going back!
+
+To migrate from Sqlite to Postgres, follow these steps:
+1. Stop Eclair
+2. Edit `eclair.conf`
+   1. Set `eclair.db.postgres.*` as explained in the section [Connection Settings](#connection-settings).
+   2. Set `eclair.db.driver=dual-sqlite-primary`. This will make Eclair use both databases backends. All calls to sqlite will be replicated in postgres.
+   3. Set `eclair.db.dual.migrate-on-restart=true`. This will make Eclair migrate the data from Sqlite to Postgres at startup.
+   4. Set `eclair.db.dual.compare-on-restart=true`. This will make Eclair compare Sqlite and Postgres at startup. The result of the comparison is displayed in the logs.
+3. Delete the file `~/.eclair/last_jdbcurl`. The purpose of this file is to prevent accidental change in the database backend.
+4. Start Eclair. You should see in the logs:
+   1. `migrating all tables...`
+   2. `migration complete`
+   3. `comparing all tables...`
+   4. `comparison complete identical=true` (NB: if `identical=false`, contact support)
+5. Eclair should then finish startup and operate normally. Data has been migrated to Postgres, and Sqlite/Postgres will be maintained in sync going forward.
+6. Edit `eclair.conf` and set `eclair.db.dual.migrate-on-restart=false` but do not restart Eclair yet.
+7. We recommend that you leave Eclair in dual db mode for a while, to make sure that you don't have issues with your new Postgres database. This a good time to set up [Backups and replication](#backups-and-replication).
+8. After some time has passed, restart Eclair. You should see in the logs:
+   1. `comparing all tables...`
+   2. `comparison complete identical=true` (NB: if `identical=false`, contact support)
+9. At this point we have confidence that the Postgres backend works normally, and we are ready to drop Sqlite for good.
+10. Edit `eclair.conf`
+    1. Set `eclair.db.driver=postgres`
+    2. Set `eclair.db.dual.compare-on-restart=false`
+11. Restart Eclair. From this moment, you cannot go back to Sqlite! If you try to do so, Eclair will refuse to start.
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
@@ -6,8 +6,7 @@ import fr.acinq.eclair.channel._
 import fr.acinq.eclair.db.Databases.{FileBackup, PostgresDatabases, SqliteDatabases}
 import fr.acinq.eclair.db.DbEventHandler.ChannelEvent
 import fr.acinq.eclair.db.DualDatabases.runAsync
-import fr.acinq.eclair.db.pg._
-import fr.acinq.eclair.db.sqlite._
+import fr.acinq.eclair.io.Peer
 import fr.acinq.eclair.payment._
 import fr.acinq.eclair.payment.relay.Relayer.RelayFees
 import fr.acinq.eclair.router.Router
@@ -23,25 +22,30 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
 /**
- * An implementation of [[Databases]] where there are two separate underlying db, one sqlite and one postgres.
- * Sqlite is the main database, but we also replicate all calls to postgres.
- * Calls to postgres are made asynchronously in a dedicated thread pool, so that it doesn't have any performance impact.
+ * An implementation of [[Databases]] where there are two separate underlying db, one primary and one secondary.
+ * All calls to primary are replicated asynchronously to secondary.
+ * Calls to secondary are made asynchronously in a dedicated thread pool, so that it doesn't have any performance impact.
  */
-case class DualDatabases(sqlite: SqliteDatabases, postgres: PostgresDatabases) extends Databases with FileBackup {
+case class DualDatabases(primary: Databases, secondary: Databases) extends Databases with FileBackup {
 
-  override val network: NetworkDb = DualNetworkDb(sqlite.network, postgres.network)
+  override val network: NetworkDb = DualNetworkDb(primary.network, secondary.network)
 
-  override val audit: AuditDb = DualAuditDb(sqlite.audit, postgres.audit)
+  override val audit: AuditDb = DualAuditDb(primary.audit, secondary.audit)
 
-  override val channels: ChannelsDb = DualChannelsDb(sqlite.channels, postgres.channels)
+  override val channels: ChannelsDb = DualChannelsDb(primary.channels, secondary.channels)
 
-  override val peers: PeersDb = DualPeersDb(sqlite.peers, postgres.peers)
+  override val peers: PeersDb = DualPeersDb(primary.peers, secondary.peers)
 
-  override val payments: PaymentsDb = DualPaymentsDb(sqlite.payments, postgres.payments)
+  override val payments: PaymentsDb = DualPaymentsDb(primary.payments, secondary.payments)
 
-  override val pendingCommands: PendingCommandsDb = DualPendingCommandsDb(sqlite.pendingCommands, postgres.pendingCommands)
+  override val pendingCommands: PendingCommandsDb = DualPendingCommandsDb(primary.pendingCommands, secondary.pendingCommands)
 
-  override def backup(backupFile: File): Unit = sqlite.backup(backupFile)
+  /** if one of the database supports file backup, we use it */
+  override def backup(backupFile: File): Unit = (primary, secondary) match {
+    case (f: FileBackup, _) => f.backup(backupFile)
+    case (_, f: FileBackup) => f.backup(backupFile)
+    case _ => ()
+  }
 }
 
 object DualDatabases extends Logging {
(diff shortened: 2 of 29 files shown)


docs/PostgreSQL.md
```diff
@@ -111,4 +111,34 @@ Eclair stores the latest database settings in the `${data-dir}/last_jdbcurl` fil
 
 The node operator can force Eclair to accept new database 
 connection settings by removing the `last_jdbcurl` file. 
-    
\ No newline at end of file
+
+### Migrating from Sqlite to Postgres
```
pm47, line 115: @sstone @t-bast wanna try following this migration guide and report any issues? 🙏

eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
```diff
@@ -55,360 +59,369 @@ object DualDatabases extends Logging {
...
     }
   }
+
+  def getDatabases(dualDatabases: DualDatabases): (SqliteDatabases, PostgresDatabases) =
+    (dualDatabases.primary, dualDatabases.secondary) match {
+      case (sqliteDb: SqliteDatabases, postgresDb: PostgresDatabases) =>
+        (sqliteDb, postgresDb)
+      case (postgresDb: PostgresDatabases, sqliteDb: SqliteDatabases) =>
+        (sqliteDb, postgresDb)
+      case _ => ???
```
t-bast, line 69: Can you explicitly throw with an error message here instead?

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareDb.scala
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    var hashes2 = List.empty[ByteVector]
+    while (rs2.next()) {
+      hashes2 = hash2(rs2) +: hashes2
+    }
+
+    val res = hashes1.sorted == hashes2.sorted
+    val diff1 = hashes1 diff hashes2
+    val diff2 = hashes2 diff hashes1
```
t-bast, line 33: nit: you can put that in the `else` branch below, no need to compare a `diff` on these big lists if everything if ok?

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareDb.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    res
+  }
+
+  // @formatter:off
+  import fr.acinq.eclair.db.jdbc.JdbcUtils.ExtendedResultSet._
+  def bytes(rs: ResultSet, columnName: String): ByteVector = rs.getByteVector(columnName)
+  def bytesnull(rs: ResultSet, columnName: String): ByteVector = rs.getByteVectorNullable(columnName).getOrElse(ByteVector.fromValidHex("deadbeef"))
+  def hex(rs: ResultSet, columnName: String): ByteVector = rs.getByteVectorFromHex(columnName)
+  def hexnull(rs: ResultSet, columnName: String): ByteVector = rs.getByteVector32FromHexNullable(columnName).map(_.bytes).getOrElse(ByteVector.fromValidHex("deadbeef"))
```
t-bast, line 49: Why is this one using `ByteVector32` but the one above uses `ByteVector`? Also its fallback isn't a `ByteVector32`

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareDb.scala
```diff
@@ -0,0 +1,80 @@
...
+  def int(rs: ResultSet, columnName: String): ByteVector = ByteVector.fromInt(rs.getInt(columnName))
+  def ts(rs: ResultSet, columnName: String): ByteVector = ByteVector.fromLong((rs.getTimestamp(columnName).getTime.toDouble / 1_000_000).round)
+  def tsnull(rs: ResultSet, columnName: String): ByteVector = ByteVector.fromLong(rs.getTimestampNullable(columnName).map(t => (t.getTime.toDouble / 1_000_000).round).getOrElse(42))
+  def tssec(rs: ResultSet, columnName: String): ByteVector = ByteVector.fromLong((rs.getTimestamp(columnName).toInstant.getEpochSecond.toDouble / 1_000_000).round)
+  def tssecnull(rs: ResultSet, columnName: String): ByteVector = ByteVector.fromLong(rs.getTimestampNullable(columnName).map(t => (t.toInstant.getEpochSecond.toDouble / 1_000_000).round).getOrElse(42))
+  // @formatter:on
+
+  def compareAll(dualDatabases: DualDatabases): Unit = {
+    logger.info("comparing all tables...")
+    val (sqliteDb, postgresDb) = DualDatabases.getDatabases(dualDatabases)
```
t-bast, line 66: This assumes that `sqlite` is primary, all these comparisons won't work if postgres is primary (because inside the comparisons you assign the table names so that the first one is sqlite...)

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareDb.scala
```diff
@@ -0,0 +1,80 @@
...
+    PgUtils.inTransaction { postgres =>
+      val result = List(
+        CompareChannelsDb.compareAllTables(sqliteDb.channels.sqlite, postgres),
+        ComparePendingCommandsDb.compareAllTables(sqliteDb.channels.sqlite, postgres),
+        ComparePeersDb.compareAllTables(sqliteDb.channels.sqlite, postgres),
+        ComparePaymentsDb.compareAllTables(sqliteDb.channels.sqlite, postgres),
```
t-bast, line 72: I know these are all using the same connection under the hood, but it's weird to read them all using `sqlite.channels` when the two below use something different. I think it's better to make them explicitly correct:

```suggestion
        CompareChannelsDb.compareAllTables(sqliteDb.channels.sqlite, postgres),
        ComparePendingCommandsDb.compareAllTables(sqliteDb.pendingCommands.sqlite, postgres),
        ComparePeersDb.compareAllTables(sqliteDb.peers.sqlite, postgres),
        ComparePaymentsDb.compareAllTables(sqliteDb.payments.sqlite, postgres),
```

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/MigrateDb.scala
```diff
@@ -0,0 +1,58 @@
...
+      MigrateChannelsDb.migrateAllTables(eclair_sqlite, postgres)
+      MigratePendingCommandsDb.migrateAllTables(eclair_sqlite, postgres)
+      MigratePeersDb.migrateAllTables(eclair_sqlite, postgres)
+      MigratePaymentsDb.migrateAllTables(eclair_sqlite, postgres)
+      // network.sqlite
+      val network_sqlite = sqliteDb.network.sqlite
+      MigrateNetworkDb.migrateAllTables(network_sqlite, postgres)
+      // audit.sqlite
+      val audit_sqlite = sqliteDb.audit.sqlite
+      MigrateAuditDb.migrateAllTables(audit_sqlite, postgres)
```
t-bast, line 53: Same as the comment in `CompareDb`, I think the "implementation detail" that some DBs are in the same instance don't matter, it's more clear to simply have:

```suggestion
      MigrateChannelsDb.migrateAllTables(sqliteDb.channels.sqlite, postgres)
      MigratePendingCommandsDb.migrateAllTables(sqliteDb.pendingCommands.sqlite, postgres)
      MigratePeersDb.migrateAllTables(sqliteDb.peers.sqlite, postgres)
      MigratePaymentsDb.migrateAllTables(sqliteDb.payments.sqlite, postgres)
      MigrateNetworkDb.migrateAllTables(sqliteDb.network.sqlite, postgres)
      MigrateAuditDb.migrateAllTables(sqliteDb.audit.sqlite, postgres)
```

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/MigratePeersDb.scala
```diff
@@ -0,0 +1,41 @@
...
+      insertStatement.setString(1, rs.getByteVector("node_id").toHex)
+      insertStatement.setLong(2, rs.getLong("fee_base_msat"))
+      insertStatement.setLong(3, rs.getLong("fee_proportional_millionths"))
+    }
+
+    migrateTable(source, destination, sourceTable, insertSql, migrate)
+  }
+
+  def migrateAllTables(source: Connection, destination: Connection): Unit = {
+    checkVersions(source, destination, "peers", 2, 3)
```
t-bast, line 36: By hard-coding the values here instead of using `SqlitePeersDb.CURRENT_VERSION` and `PgPeersDb.CURRENT_VERSION`, you ensure that this migration will fail in future versions of eclair where the DB versions change.

That's expected, right? Because we would need to re-write the migration as they are version-specific?
Users can only do the Sqlite -> Postgres migration once, when we release this eclair version.
Then if we want to support migrating in next eclair versions, we will update the code here.

t-bast (commented):
A first pass without looking at each individual table (which I'll do next).

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/ComparePaymentsDb.scala
```diff
@@ -0,0 +1,87 @@
...
+        string(rs, "payment_type") ++
+        long(rs, "amount_msat") ++
+        longnull(rs, "fees_msat") ++
+        long(rs, "recipient_amount_msat") ++
+        hex(rs, "recipient_node_id") ++
+        stringnull(rs, "payment_request") ++
+        bytesnull(rs, "payment_route") ++
+        bytesnull(rs, "failures") ++
+        ts(rs, "created_at") ++
+        tsnull(rs, "completed_at")
```
t-bast, line 76:
```suggestion
        longts(rs, "created_at") ++
        longtsnull(rs, "completed_at")
```

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/MigratePeersDb.scala
pm47: Exactly. It's pretty convenient and allowed me to quickly upate the migration code for the audit and peers db when preparing this PR.

eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
pm47: Done with 5c3eac546.

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareDb.scala
pm47: Done with 5c3eac546.

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareDb.scala
pm47: Fixed with 5c3eac546.

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareDb.scala
pm47: Nice catch, fixed with 5c3eac546. I've updated `Databases.init` so we only migrate/compare when sqlite is primary, and I added explicit casts in `migrateAll`/`compareAll`.

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareDb.scala
pm47: Done with 5c3eac546.

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/MigrateDb.scala
pm47: Done with 5c3eac546.

eclair-core/src/test/scala/fr/acinq/eclair/db/DbMigrationSpec.scala
```diff
@@ -0,0 +1,112 @@
...
+import org.scalatest.funsuite.AnyFunSuite
+import org.sqlite.SQLiteConfig
+
+import java.io.File
+import java.sql.{Connection, DriverManager}
+import java.util.UUID
+import javax.sql.DataSource
+
+@Ignore
+class DbMigrationSpec extends AnyFunSuite {
```
t-bast, line 20: nit: can you add a comment to guide users on how to test migration with these tests?

t-bast (commented):
Reviewed each comparison and migration.

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareChannelsDb.scala
```diff
@@ -0,0 +1,80 @@
...
+    val table1 = "local_channels"
+    val table2 = "local.channels"
+
+    def hash1(rs: ResultSet): ByteVector = {
+      val data = ByteVector(rs.getBytes("data"))
+      val data_modified = stateDataCodec.decode(data.bits).require.value match {
+        case c: DATA_WAIT_FOR_FUNDING_CONFIRMED => stateDataCodec.encode(c.copy(waitingSince = BlockHeight(0))).require.toByteVector
+        case c: DATA_CLOSING => stateDataCodec.encode(c.copy(waitingSince = BlockHeight(0))).require.toByteVector
+        case _ => data
+      }
```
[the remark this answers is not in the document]
pm47: Because of this:
https://github.com/ACINQ/eclair/blob/a5bfde7b18a7db23025d2148cc2eee304c8f75e5/eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala#L358

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/CompareChannelsDb.scala
```diff
@@ -0,0 +1,80 @@
...
+    compareTable(conn1, conn2, table1, table2, hash1, hash2)
+  }
+
+  private def compareHtlcInfosTable(conn1: Connection, conn2: Connection): Boolean = {
+    val table1 = "htlc_infos"
+    val table2 = "local.htlc_infos"
+
+    def hash1(rs: ResultSet): ByteVector = {
+      bytes(rs, "channel_id") ++
+        string(rs, "commitment_number") ++
```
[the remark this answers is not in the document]
pm47: Because before it got fixed in e14c40d7c36d4fd19d84dfeddbd711a9c74b7516, the `commitment_number` had a `BLOB` typed. Not that it matters since sqlite uses [dynamic typing](https://www.sqlite.org/datatype3.html). As I write this I just noticed that:
> Update: As of version 3.37.0 (2021-11-27), SQLite provides STRICT tables that do rigid type enforcement, for developers who prefer that kind of thing.

Fixed with e69e624c9.

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/ComparePaymentsDb.scala
```diff
@@ -0,0 +1,87 @@
...
+
+    def hash2(rs: ResultSet): ByteVector = {
+      hex(rs, "payment_hash") ++
+        string(rs, "payment_type") ++
+        hex(rs, "payment_preimage") ++
+        string(rs, "payment_request") ++
+        longnull(rs, "received_msat") ++
+        ts(rs, "created_at") ++
+        ts(rs, "expire_at") ++
+        tsnull(rs, "received_at")
```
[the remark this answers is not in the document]
pm47: No because the types are not the same: they are stored as unix timestamps in sqlite, and actual timestamps in postgres.

eclair-core/src/main/scala/fr/acinq/eclair/db/migration/MigratePaymentsDb.scala
```diff
@@ -0,0 +1,60 @@
...
+      insertStatement.setString(4, rs.getByteVector("payment_hash").toHex)
+      insertStatement.setString(5, rs.getByteVector32Nullable("payment_preimage").map(_.toHex).orNull)
+      insertStatement.setString(6, rs.getString("payment_type"))
+      insertStatement.setLong(7, rs.getLong("amount_msat"))
+      insertStatement.setObject(8, rs.getLongNullable("fees_msat").orNull)
+      insertStatement.setLong(9, rs.getLong("recipient_amount_msat"))
+      insertStatement.setString(10, rs.getByteVector("recipient_node_id").toHex)
+      insertStatement.setString(11, rs.getStringNullable("payment_request").orNull)
+      insertStatement.setBytes(12, rs.getBytes("payment_route"))
+      insertStatement.setBytes(13, rs.getBytes("failures"))
```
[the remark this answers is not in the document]
pm47: There is no such method as `setBytesNullable`?  And `getBytes` handles `null` values seamlessly:

> /**
     * [...]
     * @return the column value; if the value is SQL <code>NULL</code>, the
     * value returned is <code>null</code>
     * [...]
     */
    byte[] getBytes(String columnLabel) throws SQLException;

eclair-core/src/test/scala/fr/acinq/eclair/db/DbMigrationSpec.scala
pm47: Done with 8e8292732. It's pretty basic!

t-bast (approved):
LGTM!

I tested the migration steps on regtest, and it seems to be working fine :+1:

PR #2157, opened 2022-01-28
Fee bumping with unsafe external inputs

t-bast:
We greatly simplify and improve the fee-bumping logic thanks to two new `bitcoind` options for funding transactions.

We first use unsafe inputs to fund replaceable transactions. This ensures that we can use all our available funds for funding force-close transactions.

We also specify the weight of external inputs when calling `bitcoind`. We were previously forced to remove our inputs before calling bitcoind and re-add them afterwards, which forced us to have a lot of code that was hard to read and inefficient.

The result is that we're able to create complete trees of unconfirmed transactions during force-close, such as:

```
+-----------------+
|                 |
| htlc-success tx |--------+
|                 |        |
+-----------------+        |
                           |
+-----------+              |         +-----------------+
|           |              +-------->|                 |
| anchor tx |----------------------->| htlc-success tx |------------+
|           |                        |                 |            |
+-----------+                        +-----------------+            |
                                                                    |
+-----------------+                                                 |           +-----------+                 +-----------------+
|                 |                                                 |           |           |                 |                 |
| htlc-timeout tx |--------+                                        +---------->| anchor tx |---------------->| htlc-success tx |
|                 |        |                                        +---------->|           |                 |                 |
+-----------------+        |                                        |           +-----------+                 +-----------------+
                           |                                        |
+-----------------+        |         +-----------------+            |
|                 |        +-------->|                 |            |
| htlc-success tx |----------------->| htlc-timeout tx |------------+
|                 |                  |                 |
+-----------------+                  +-----------------+
```

This is great for capital efficiency: we're able to use every utxo we can spend!

However, it's not all sunshine and roses, because we'll run into heavy limitations because of [BIP 125](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki) (mostly rules 2 and 3).

Rule 2 means that we won't be able to be as capital-efficient as we though: our unsafe inputs cannot be used for RBF, only for the first time we fund a transaction.

Rule 3 means that if we want to replace a transaction that has a long chain of low feerate descendants, we're going to have a hard time doing so: the fee we will need to pay will be absurdly high so in practice we simply won't be able to RBF that transaction...Maybe it would make sense to explicitly monitor chains of unconfirmed transactions, and heavily RBF the last one to make the whole chain confirm when one of the transactions in the tree is reaching its deadline?

These are serious limitations because we can't tell `bitcoind` to stop using inputs for funding once their chain of ancestors has reached a certain length. Maybe this can be improved by better RPC support on the `bitcoind` side, but for now the wallet doesn't have enough hooks into the mempool to get that kind of information. I'm frankly unsure whether we should deploy this update before bitcoin's RBF policies have been revamped...

NB: this PR builds upon #2151 which requires bitcoind 23.0.


t-bast:

PR #2163, opened 2022-02-02
Don't broadcast commit before funding locked

t-bast:
If a channel is force-closed before the funding transaction is deeply confirmed, broadcasting our local commit can be a problem if the funding tx is double spent. When that happens, the channel stays stuck in the closing state trying to publish a commit tx with an invalid input.

If we haven't even seen the funding tx in the mempool, we have no way of being sure that it was double spent, so we would need to keep trying forever, which pollutes the logs with publishing errors.

Whenever we have nothing at stake and a channel closes before reaching the normal state, we now directly go to closed.


pm47 (commented):
LGTM but the commit message needs to be updated, especially since the back-and-forth on this portion of the code will be confusing to our future selves.

pm47 (approved):

t-bast:
> LGTM but the commit message needs to be updated, especially since the back-and-forth on this portion of the code will be confusing to our future selves.

Done, I updated the commit message while squashing/rebase.

PR #2164, opened 2022-02-03
Typed features

thomash-acinq:
Give different types to init features / node announcement features / invoice features.

This also fixes the lack of unknown features in the invoice (users can now add their custom unknown features) and we now accept more invoices (we would previously reject invoices if they required a non invoice feature that we don't support).


t-bast:
> we now accept more invoices (we would previously reject invoices if they required a non invoice feature that we don't support).

But that's not what we want though, we want exactly the opposite behavior.
If the invoice feature bits contain a mandatory feature that we don't know, we must reject that invoice.

I really like having better types, but as I [pointed out](https://github.com/ACINQ/eclair/pull/2063#discussion_r779601182) the first time we explored it, this creates a tension with dynamic data we receive from the external world that might not pass these static checks (and that we still want to allow because we need to be a bit lenient here in case other implementations aren't strict enough with how they use feature bits).

Properly typing this is useful only if we can actually use these types for:

- invoices: but we can't because we don't want to automatically reject invoices that contain non-invoice feature bits
- node announcements: same reason, we actually don't want to enforce it
- init: here we can enforce it, but mostly because all feature bits except `option_payment_metadata` are init features, so it doesn't bring much

If we were the only implementation in the network, we'd know that external data would follow these rules and could enforce them, but in a decentralized network such as lightning I don't think we can be that strict.

thomash-acinq:
> > we now accept more invoices (we would previously reject invoices if they required a non invoice feature that we don't support).
> 
> But that's not what we want though, we want exactly the opposite behavior. If the invoice feature bits contain a mandatory feature that we don't know, we must reject that invoice.

I don't think you understand what it does. If the invoice contains a mandatory feature that we don't know, we will reject the invoice, the case when we would accept a previously rejected invoice is it contains a mandatory feature we don't support but that **we know is not for invoices**.

>     we don't want to automatically reject invoices that contain non-invoice feature bits

And we won't, we'll just ignore these non-invoice feature bits.

t-bast:
Ok, then your description was very misleading!

>  If the invoice contains a mandatory feature that we don't know, we will reject the invoice, the case when we would accept a previously rejected invoice is it contains a mandatory feature we don't support but that we know is not for invoices.

I'm still not sure how that sentence should be interpreted.
I'll go look at the code, it's probably easier than english to verify that it has the right behavior.

> And we won't, we'll just ignore these non-invoice feature bits.

But we don't want to ignore them completely, if they are mandatory, we do want to verify that we also support them.
But again, I may be misinterpreting your sentence.

thomash-acinq:
> > And we won't, we'll just ignore these non-invoice feature bits.
> 
> But we don't want to ignore them completely, if they are mandatory, we do want to verify that we also support them.

Why do we care about non-invoice features ? If we receive an invoice that says "option_anchor_outputs is mandatory" and we don't support anchor outputs, why should we reject the invoice, it means we probably don't have a direct channel with that node as our channel features are incompatible but that doesn't prevent us from paying this invoice.

t-bast:
> Why do we care about non-invoice features ?

Because if it's an unknown feature, it could be an invoice feature, we simply don't know, right?
If you see feature bit 64 for example, we don't know what it is, but it may be a new invoice feature that has been introduced and we don't know about.

I haven't read the code yet, so maybe this is properly handled already?

thomash-acinq:
Unknown features are processed separately, I'm talking here about features that we know are not invoice features, these are features that we know, or features above 5120 because they don't fit in invoices anyway.

t-bast:
Great, then that should work!
I'll make comments directly on the PR.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt11Invoice.scala
```diff
@@ -414,7 +414,7 @@ object Bolt11Invoice {
       .typecase(2, dataCodec(bits).as[UnknownTag2])
       .typecase(3, dataCodec(listOfN(extraHopsLengthCodec, extraHopCodec)).as[RoutingInfo])
       .typecase(4, dataCodec(bits).as[UnknownTag4])
-      .typecase(5, dataCodec(bits).xmap[Features](Features(_), features2bits).as[InvoiceFeatures])
+      .typecase(5, dataCodec(bits).xmap[Features[InvoiceFeature]](Features(_).invoiceFeatures(), features2bits).as[InvoiceFeatures])
```
t-bast, line 417: Unfortunately I think this breaks the signature check in `fromString` if the invoice contains a non-invoice feature bit.
We should add test cases:

- read an invoice containing non-invoice known optional feature bits
- read an invoice containing non-invoice known mandatory feature bits
- read an invoice containing unknown optional feature bits
- read an invoice containing unknown mandatory feature bits

We can probably rework the `fromString` method to keep all feature bits for the signature check while still keeping the benefits of using correctly typed features once the checks are done.

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -55,6 +55,11 @@ object TestConstants {
     val mandatory = 50000
   }
 
+  case object LowTestFeature extends Feature with InitFeature with NodeFeature {
```
t-bast, line 58: That name can be improved...If we start having multiple unknown test features, we should rename `TestFeature` to `UnknownExperimentalFeature` (IIRC we called that range "experimental range") and `LowTestFeature` to `UnknownSpecFeature` or something like this?

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
```diff
@@ -575,4 +576,12 @@ class Bolt11InvoiceSpec extends AnyFunSuite {
       assert(Bolt11Invoice.fromString(req).toString == req)
     }
   }
+
+  test("large unknown feature"){
```
t-bast, line 580: This test should also serialize and deserialize the invoice and verify that we get the same result.

Can you also add a separate test that showcases the exception we get if we try to serialize an invoice that contains a feature bit that's way too big?

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentInitiatorSpec.scala
```diff
@@ -128,7 +128,7 @@ class PaymentInitiatorSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike
       Bolt11Invoice.Description("Some invoice"),
       Bolt11Invoice.PaymentSecret(randomBytes32()),
       Bolt11Invoice.Expiry(3600),
-      Bolt11Invoice.InvoiceFeatures(Features(bin"000001000000000000000000000000000100000100000000")) // feature 42
+      Bolt11Invoice.InvoiceFeatures(Features(bin"000001000000000000000000000000000100000100000000").invoiceFeatures()) // feature 42
```
t-bast, line 131: Now that we don't really control what happens in `.invoiceFeatures`, we should make this more explicit and use:
```suggestion
      Bolt11Invoice.InvoiceFeatures(Features[InvoiceFeature](Map(VariableLengthOnion -> Mandatory, PaymentSecret -> Mandatory), Set(UnknownFeature(42))))
```

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecsSpec.scala
```diff
@@ -269,7 +269,7 @@ class LightningMessageCodecsSpec extends AnyFunSuite {
     val commit_sig = CommitSig(randomBytes32(), randomBytes64(), randomBytes64() :: randomBytes64() :: randomBytes64() :: Nil)
     val revoke_and_ack = RevokeAndAck(randomBytes32(), scalar(0), point(1))
     val channel_announcement = ChannelAnnouncement(randomBytes64(), randomBytes64(), randomBytes64(), randomBytes64(), Features(bin(7, 9)), Block.RegtestGenesisBlock.hash, ShortChannelId(1), randomKey().publicKey, randomKey().publicKey, randomKey().publicKey, randomKey().publicKey)
-    val node_announcement = NodeAnnouncement(randomBytes64(), Features(bin(1, 2)), 1 unixsec, randomKey().publicKey, Color(100.toByte, 200.toByte, 300.toByte), "node-alias", IPv4(InetAddress.getByAddress(Array[Byte](192.toByte, 168.toByte, 1.toByte, 42.toByte)).asInstanceOf[Inet4Address], 42000) :: Nil)
+    val node_announcement = NodeAnnouncement(randomBytes64(), Features(bin(1, 2)).nodeAnnouncementFeatures(), 1 unixsec, randomKey().publicKey, Color(100.toByte, 200.toByte, 300.toByte), "node-alias", IPv4(InetAddress.getByAddress(Array[Byte](192.toByte, 168.toByte, 1.toByte, 42.toByte)).asInstanceOf[Inet4Address], 42000) :: Nil)
```
t-bast, line 272: We should instead explicitly create a `Features[NodeFeature]` here, that would be clearer.

t-bast (commented):
I think we can make that work, and it would be a nice improvement.
We need to test and fix the subtle signature issues, but that shouldn't be impossible to handle.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -113,26 +115,26 @@ case class Features(activated: Map[Feature, FeatureSupport], unknown: Set[Unknow
 
 object Features {
 
-  def empty = Features(Map.empty[Feature, FeatureSupport])
+  def empty: Features[FeatureScope] = Features[FeatureScope](Map.empty[Feature with FeatureScope, FeatureSupport])
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -29,22 +29,22 @@ import shapeless._
...
 
-  val featuresCodec: Codec[Features] = varsizebinarydata.xmap[Features](
+  val featuresCodec: Codec[Features[FeatureScope]] = varsizebinarydata.xmap[Features[FeatureScope]](
     { bytes => Features(bytes) },
     { features => features.toByteVector }
   )
 
   /** For historical reasons, features are divided into two feature bitmasks. We only send from the second one, but we allow receiving in both. */
-  val combinedFeaturesCodec: Codec[Features] = (
+  val combinedFeaturesCodec: Codec[Features[FeatureScope]] = (
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecsSpec.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentInitiatorSpec.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
thomash-acinq: The distinction between "spec feature" and "experimental feature" is irrelevant here and would only add confusion, what matters is that one uses a low number and the other one uses a high number which is why I named it like that. I could rename `TestFeature` to `HighTestFeature` to be more consistent.

eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
```diff
@@ -146,7 +146,7 @@ object EclairInternalsSerializer {
   def initializeConnectionCodec(system: ExtendedActorSystem): Codec[PeerConnection.InitializeConnection] = (
     ("peer" | actorRefCodec(system)) ::
       ("chainHash" | bytes32) ::
-      ("features" | variableSizeBytes(uint16, featuresCodec)) ::
+      ("features" | variableSizeBytes(uint16, featuresCodec.xmap[Features[InitFeature]](_.initFeatures(), _.unscoped()))) ::
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
t-bast: I thought you meant to use that to test the difference between feature bits in the experimental range allowed for blips vs feature bits in the range reserved for bolts (which we currently don't distinguish). This misunderstanding does show that the name is quite bad though.

What's the threshold between low and high number? That's completely arbitrary and not explained, as it depends on _what_ you plan to test with this, right? Can you clarify your goal here? That would help find the right name for this.

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -82,12 +82,14 @@ case class Features(activated: Map[Feature, FeatureSupport], unknown: Set[Unknow
...
-  def initFeatures(): Features = Features(activated.collect { case (f: InitFeature, s) => (f: Feature, s) }, unknown)
+  def initFeatures(): Features[InitFeature] = Features[InitFeature](activated.collect { case (f: InitFeature, s) => (f, s) }, unknown)
 
-  def nodeAnnouncementFeatures(): Features = Features(activated.collect { case (f: NodeFeature, s) => (f: Feature, s) }, unknown)
+  def nodeAnnouncementFeatures(): Features[NodeFeature] = Features[NodeFeature](activated.collect { case (f: NodeFeature, s) => (f, s) }, unknown)
 
-  // NB: we don't include unknown features in invoices, which means plugins cannot inject invoice features.
-  def invoiceFeatures(): Features = Features(activated.collect { case (f: InvoiceFeature, s) => (f: Feature, s) })
+  // NB: features above 5 * 2^10 do not fit in bolt 11 invoices
+  def invoiceFeatures(): Features[InvoiceFeature] = Features[InvoiceFeature](activated.collect { case (f: InvoiceFeature, s) => (f, s) }, unknown.filter(_.bitIndex < 5 * 1024))
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
thomash-acinq: It was used to test that invoices wouldn't crash with features higher than 5120. I've removed it since we've decided that invoices can't have unknown features.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt11Invoice.scala
thomash-acinq: done

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -263,7 +263,7 @@ case class Tor3(tor3: String, port: Int) extends OnionAddress { override def soc
 // @formatter:on
 
 case class NodeAnnouncement(signature: ByteVector64,
-                            features: Features,
+                            features: Features[NodeFeature],
```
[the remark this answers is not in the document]
thomash-acinq: Fixed, I put untyped features inside `NodeAnnouncement` so it can deal with anything it receives but I restrict building `NodeAnnouncement` in our code to using node features.

thomash-acinq:
It should be ready now.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -58,8 +58,8 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
                       color: Color,
                       publicAddresses: List[NodeAddress],
                       torAddress_opt: Option[NodeAddress],
-                      features: Features,
-                      private val overrideFeatures: Map[PublicKey, Features],
+                      features: Features[FeatureScope],
+                      private val overrideFeatures: Map[PublicKey, Features[InitFeature]],
```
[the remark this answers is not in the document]
pm47: How about renaming `overrideFeatures` to `overrideInitFeatures` ?

eclair-core/src/test/scala/fr/acinq/eclair/router/AnnouncementsSpec.scala
```diff
@@ -128,4 +129,11 @@ class AnnouncementsSpec extends AnyFunSuite {
     assert(!channelUpdate2_disabled.channelFlags.isEnabled)
   }
 
+  test("announce irrelevant features") {
+    // This announcement has option_payment_metadata which is not a node feature
+    val encoded = hex"25d8bf19e2c6562a85b3109122118a50728d42c9054537b5a26c1e982407f0923a6f162cf0939bfaf58d7e7a7a7c86d7dfd02dda9b0d2036e8fd35a7afc78daa00070100000000000061fcfc1d039093816bb908c043341f3ee47ea86179627d26aa00704593503e684a0b9a38cb01020374657374206e6f646500000000000000000000000000000000000000000000000007018c5279042607"
+    val ann = nodeAnnouncementCodec.decode(encoded.bits).require.value
```
t-bast, line 135: Can you add an assert here that it does have the `option_payment_metadata` feature bit?

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
```diff
@@ -510,69 +510,84 @@ class Bolt11InvoiceSpec extends AnyFunSuite {
...
-    for (req <- requests) {
-      assert(Bolt11Invoice.fromString(req).toString == req)
+    for ((req, nodeId) <- requests) {
+      assert(Bolt11Invoice.fromString(req).nodeId === nodeId)
+      assert(Bolt11Invoice.fromString(req).toString === req)
     }
   }
+
+  test("no unknown feature in invoice"){
+    val invoice = Bolt11Invoice(Block.LivenetGenesisBlock.hash, Some(123 msat), ByteVector32.One, priv, Left("Some invoice"), CltvExpiryDelta(18), features = TestConstants.Alice.nodeParams.features.invoiceFeaturesNoUnknown())
```
t-bast, line 585: This isn't testing much since we're explicitly calling `invoiceFeaturesNoUnknown`...
Maybe it would make sense to have our constructor for `Bolt11Invoice` explicitly removing the `unknown` features (to protect against a caller mistakenly providing `unknown` features) and in this test we would call it with `invoiceFeaturesWithUnknown` and verify that the unknown ones are removed?

t-bast (commented):
Nice, looks good to me! A couple of test improvements and this should be good to go :+1:

eclair-core/src/test/scala/fr/acinq/eclair/router/AnnouncementsSpec.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
thomash-acinq: I now strip the unknown is the constructor, this make the distinction `invoiceFeaturesNoUnknown`/`invoiceFeaturesWithUnknown` irrelevant, I've reunified them.

eclair-core/src/main/resources/reference.conf
```diff
@@ -60,7 +60,7 @@ eclair {
     trampoline_payment = disabled
     keysend = disabled
   }
-  override-features = [ // optional per-node features
+  override-init-features = [ // optional per-node features
```
t-bast, line 63: Can you also update `Configure.md` with that new parameter name?

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -268,7 +269,7 @@ object NodeParams extends Logging {
     val nodeAlias = config.getString("node-alias")
     require(nodeAlias.getBytes("UTF-8").length <= 32, "invalid alias, too long (max allowed 32 bytes)")
 
-    def validateFeatures(features: Features): Unit = {
+    def validateFeatures(features: Features[InitFeature]): Unit = {
```
t-bast, line 272: Do you really need to make this `InitFeature`? There's no reason that it will be restricted to `init`, this function makes sense for all feature types, it should be `FeatureScope`

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -288,13 +289,13 @@ object NodeParams extends Logging {
...
+    val coreAndPluginFeatures: Features[FeatureScope] = features.copy(unknown = features.unknown ++ pluginMessageParams.map(_.pluginFeature))
 
-    val overrideFeatures: Map[PublicKey, Features] = config.getConfigList("override-features").asScala.map { e =>
+    val overrideInitFeatures: Map[PublicKey, Features[InitFeature]] = config.getConfigList("override-init-features").asScala.map { e =>
       val p = PublicKey(ByteVector.fromValidHex(e.getString("nodeid")))
-      val f = Features.fromConfiguration(e)
+      val f = Features.fromConfiguration[InitFeature](e.getConfig("features"), Features.knownFeatures.collect { case f: Feature with InitFeature => f })
       validateFeatures(f)
-      p -> f.copy(unknown = f.unknown ++ pluginMessageParams.map(_.pluginFeature))
+      p -> (f.initFeatures().copy(unknown = f.unknown ++ pluginMessageParams.map(_.pluginFeature)): Features[InitFeature])
```
t-bast, line 298: You already have a `Features[InitFeature]`:
```suggestion
      p -> (f.copy(unknown = f.unknown ++ pluginMessageParams.map(_.pluginFeature)): Features[InitFeature])
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt11Invoice.scala
```diff
@@ -167,7 +167,7 @@ object Bolt11Invoice {
         fallbackAddress.map(FallbackAddress(_)),
         expirySeconds.map(Expiry(_)),
         Some(MinFinalCltvExpiry(minFinalCltvExpiryDelta.toInt)),
-        Some(InvoiceFeatures(features))
+        Some(InvoiceFeatures(features.copy(unknown = Set.empty).unscoped()))
```
t-bast, line 170: Can you add a comment here?
```suggestion
        // We want to keep invoices as small as possible, so we explicitly remove unknown features.
        Some(InvoiceFeatures(features.copy(unknown = Set.empty).unscoped()))
```

eclair-core/src/test/scala/fr/acinq/eclair/StartupSpec.scala
```diff
@@ -211,15 +211,7 @@ class StartupSpec extends AnyFunSuite {
...
-    val nodeParams = makeNodeParamsWithDefaults(perNodeConf.withFallback(defaultConf))
-    val perNodeFeaturesA = nodeParams.initFeaturesFor(PublicKey(ByteVector.fromValidHex("02aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")))
-    val perNodeFeaturesB = nodeParams.initFeaturesFor(PublicKey(ByteVector.fromValidHex("02bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb")))
-    val defaultNodeFeatures = nodeParams.initFeaturesFor(PublicKey(ByteVector.fromValidHex("02cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc")))
-    // Some features should never be sent in init messages.
-    assert(nodeParams.features.hasFeature(PaymentMetadata))
-    assert(!perNodeFeaturesA.hasFeature(PaymentMetadata))
-    assert(!perNodeFeaturesB.hasFeature(PaymentMetadata))
-    assert(!defaultNodeFeatures.hasFeature(PaymentMetadata))
+    assertThrows[IllegalArgumentException](makeNodeParamsWithDefaults(perNodeConf.withFallback(defaultConf)))
```
t-bast, line 214: The test should be renamed ("reject non-init features in node override")

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
```diff
@@ -41,6 +41,46 @@ class Bolt11InvoiceSpec extends AnyFunSuite {
...
+            privateKey: PrivateKey,
+            description: Either[String, ByteVector32],
+            minFinalCltvExpiryDelta: CltvExpiryDelta,
+            fallbackAddress: Option[String] = None,
+            expirySeconds: Option[Long] = None,
+            extraHops: List[List[ExtraHop]] = Nil,
+            timestamp: TimestampSecond = TimestampSecond.now(),
+            paymentSecret: ByteVector32 = randomBytes32(),
+            paymentMetadata: Option[ByteVector] = None,
+            features: Features[InvoiceFeature] = defaultFeatures): Bolt11Invoice = {
```
t-bast, line 57: This could be even better and simply take `Features[FeatureScope]`, which lets us easily test known, non-invoice features in invoices?

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
```diff
@@ -445,8 +485,8 @@ class Bolt11InvoiceSpec extends AnyFunSuite {
     )
 
     for ((features, res) <- featureBits) {
-      val invoice = Bolt11Invoice(Block.LivenetGenesisBlock.hash, Some(123 msat), ByteVector32.One, priv, Left("Some invoice"), CltvExpiryDelta(18), features = features)
-      assert(Result(invoice.features.hasFeature(BasicMultiPartPayment), invoice.features.hasFeature(PaymentSecret, Some(Mandatory)), nodeParams.features.areSupported(invoice.features)) === res)
+      val invoice = createInvoiceUnsafe(Block.LivenetGenesisBlock.hash, Some(123 msat), ByteVector32.One, priv, Left("Some invoice"), CltvExpiryDelta(18), features = features.invoiceFeatures())
```
t-bast, line 488: It would be nice to keep testing non-invoice features as well here:
```suggestion
      val invoice = createInvoiceUnsafe(Block.LivenetGenesisBlock.hash, Some(123 msat), ByteVector32.One, priv, Left("Some invoice"), CltvExpiryDelta(18), features = features)
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
```diff
@@ -510,69 +550,84 @@ class Bolt11InvoiceSpec extends AnyFunSuite {
...
-    for (req <- requests) {
-      assert(Bolt11Invoice.fromString(req).toString == req)
+    for ((req, nodeId) <- requests) {
+      assert(Bolt11Invoice.fromString(req).nodeId === nodeId)
+      assert(Bolt11Invoice.fromString(req).toString === req)
     }
   }
+
+  test("no unknown feature in invoice"){
+    val invoice = Bolt11Invoice(Block.LivenetGenesisBlock.hash, Some(123 msat), ByteVector32.One, priv, Left("Some invoice"), CltvExpiryDelta(18), features = TestConstants.Alice.nodeParams.features.invoiceFeatures())
```
t-bast, line 625: Can you also verify that `TestConstants.Alice.nodeParams.features.invoiceFeatures()` does contain an unknown feature? Otherwise if we independently update `TestConstants.Alice` to remove that unknown feature, we won't realize that this test doesn't test anything anymore.

eclair-core/src/main/resources/reference.conf
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt11Invoice.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/StartupSpec.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
thomash-acinq: Done

t-bast (approved):
LGTM, thanks!
It's good to get that refactoring done!

PR #2166, opened 2022-02-04
Update a few dependencies

pm47:
This updates versions of a few major dependencies: scala, akka, postgres, etc.

Supersedes #2160.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
@@ -22,7 +22,6 @@ import fr.acinq.eclair.blockchain.Monitoring.{Metrics, Tags}
 import org.json4s.JsonAST.{JString, JValue}
 import org.json4s.jackson.Serialization
 import org.json4s.{CustomSerializer, DefaultFormats}
-import sttp.capabilities
 import sttp.client3._
 import sttp.client3.json4s._
 import sttp.model.StatusCode
@@ -33,7 +32,7 @@ import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-class BasicBitcoinJsonRPCClient(rpcAuthMethod: BitcoinJsonRPCAuthMethod, host: String = "127.0.0.1", port: Int = 8332, ssl: Boolean = false, wallet: Option[String] = None)(implicit sb: SttpBackend[Future, capabilities.WebSockets]) extends BitcoinJsonRPCClient {
+class BasicBitcoinJsonRPCClient(rpcAuthMethod: BitcoinJsonRPCAuthMethod, host: String = "127.0.0.1", port: Int = 8332, ssl: Boolean = false, wallet: Option[String] = None)(implicit sb: SttpBackend[Future, _]) extends BitcoinJsonRPCClient {
 
   // necessary to properly serialize ByteVector32 into String readable by bitcoind
   object ByteVector32Serializer extends CustomSerializer[ByteVector32](_ => ( {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
@@ -16,14 +16,16 @@
 
 package fr.acinq.eclair.blockchain.bitcoind.rpc
 
-import com.softwaremill.sttp._
-import com.softwaremill.sttp.json4s._
 import fr.acinq.bitcoin.ByteVector32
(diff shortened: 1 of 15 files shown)


t-bast (approved):
LGTM I guess, let's see if it works fine on testnet before we merge

pm47:
I'm investigating a conflict related to akka-http versions.

t-bast:
The usual, I feel like I already heard that during past updates :sweat_smile:

pm47:
🧘‍♂️

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/ExplorerApi.scala
```diff
@@ -130,60 +132,58 @@ object ExplorerApi {
...
-        .send()
-        .map(r => r.code match {
+        .send(sb)
+        .map(r => r.body match {
           // HTTP 404 is a "normal" error: we're trying to lookup future blocks that haven't been mined.
-          case StatusCodes.NotFound => Seq.empty
-          case _ => r.unsafeBody \ "error" match {
-            case JString(error) if error == s"Block $blockCount not found." => Seq.empty
-            case _ => Seq(r.unsafeBody)
-          }
```
pm47, line 157: That part was a bit weird.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
```diff
@@ -16,22 +16,24 @@
...
+import sttp.model.StatusCode
 
 import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Path}
 import java.util.concurrent.atomic.AtomicReference
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-class BasicBitcoinJsonRPCClient(rpcAuthMethod: BitcoinJsonRPCAuthMethod, host: String = "127.0.0.1", port: Int = 8332, ssl: Boolean = false, wallet: Option[String] = None)(implicit http: SttpBackend[Future, Nothing]) extends BitcoinJsonRPCClient {
+class BasicBitcoinJsonRPCClient(rpcAuthMethod: BitcoinJsonRPCAuthMethod, host: String = "127.0.0.1", port: Int = 8332, ssl: Boolean = false, wallet: Option[String] = None)(implicit sb: SttpBackend[Future, capabilities.WebSockets]) extends BitcoinJsonRPCClient {
```
t-bast, line 36: Can you explain why we choose the `WebSockets` capabilities? I don't see what we should choose, we're not actually using websockets but that `capabilites` object doesn't contain anything obvious...

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/ExplorerApi.scala
```diff
@@ -130,60 +132,58 @@ object ExplorerApi {
...
+        headers <- basicRequest.readTimeout(10 seconds).get(baseUri.addPath("blocks"))
           .response(asJson[JArray])
-          .send()
+          .send(sb)
           .map(r => r.code match {
             // HTTP 404 is a "normal" error: we're trying to lookup future blocks that haven't been mined.
-            case StatusCodes.NotFound => Seq.empty
-            case _ => r.unsafeBody.arr
+            case StatusCode.NotFound => Seq.empty
+            case _ => r.body.fold(exc => throw exc, jvalue => jvalue).arr
```
t-bast, line 186: This pattern comes regularly enough that we may want to create a utility function that does this fold, don't we?
This could even be called `getBodyUnsafe` or something like that and very explicitly say that it throws?

t-bast (commented):
To be honest I don't feel completely safe with these sttp changes because it's hard to know what happens under the hood, but let's test this in the wild for a while and see how it behaves.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/watchdogs/ExplorerApi.scala
pm47: Indeed, besides a package rename, the main change was the removal of `unsafeBody` which I replaced by this `fold`. The idea was to just "make it work" now, and later improve our code to be more in line with their functional `scodec` style.

I could use the pimp pattern to define an `unsafeBody`, but it wouldn't make the diff much smaller, due to the other changes. If you think it's worth it, I can invest some time to make our code more functional and remove those `fold` alltogether.

t-bast: No, let's keep it like that for this PR, it's good to have the dependencies updated.
That can be done in a separate PR :+1:

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BasicBitcoinJsonRPCClient.scala
pm47: Done with df2adc950.

t-bast (approved):

t-bast (approved):

PR #2172, opened 2022-02-09
Fix path-finding

thomash-acinq:
We used to compute the edge weight before checking that the edge could actually relay the payment.
Since #2111 we throw an exception in case `heuristicsConstants` are used and the edge can't relay the payment instead of just ignoring the edge.
We fix this by checking that the edge can relay the payment before trying to call `addEdgeWeight`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -246,14 +246,14 @@ object Graph {
         }
         neighborEdges.foreach { edge =>
           val neighbor = edge.desc.a
-          // NB: this contains the amount (including fees) that will need to be sent to `neighbor`, but the amount that
-          // will be relayed through that edge is the one in `currentWeight`.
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -248,21 +248,24 @@ object Graph {
           val neighbor = edge.desc.a
           // NB: this contains the amount (including fees) that will need to be sent to `neighbor`, but the amount that
           // will be relayed through that edge is the one in `currentWeight`.
```
t-bast, line 250: This comment is now misplaced, can you put it right before `val neighborWeight = ...`?

t-bast (commented):
LGTM, how did you find this bug?

t-bast (approved):

thomash-acinq:
> LGTM, how did you find this bug?

Found it from the logs because of the `NegativeProbability` exception, which is also what causes the bug in the first place :(

PR #2173, opened 2022-02-09
Refactor channel states

t-bast:
We decouple the channel data from the channel state machine: what states we use in our state machine should not be publicly exposed. It is an implementation detail that may change (e.g. when we introduce a different channel opening protocol, implemented by a separate actor).
    
What we now call `ChannelData` maps to the various states defined by the lightning specification once we actually need to store data about the channel (previously `HasCommitments`, but renamed to be more general as in dual funding we may need to store DB data about the funding protocol before we have a signed commitment). It covers the BOLTs state machine, which can be implemented using multiple smaller state machines in the future (e.g. a state machine for the opening part, various state machines for the "normal" case depending on what payment protocol is used if we ever support `option_simplified_commitment`, etc).

It may seem useless because we currently have almost a 1:1 mapping between those two, but it unblocks future refactorings. It also lets us keep transient FSM data that doesn't need to be stored in the DB along with channel data that should be stored in the DB, which is something we've struggled with previously.

While this change seems trivial (useless?) it is actually quite subtle and required a few dangerous changes. I'm listing here points that reviewers should pay attention to:

- the `onTransition` now explicitly lists more cases: it previously relied on quite subtle implementation details, hopefully it's more explicit now, but there's a risk of regression
- `handleLocalError` and `handleRemoteError` now directly access the implicit `stateData` instead of having it passed in
- the channel data was removed from `ChannelErrorOccured`: did we need it there (we didn't write it to the DB)? 
- the JSON hints for channel data have changed: this is a good opportunity to nitpick on the names if you want to (e.g. `normal` doesn't speak that much to users, maybe `channel-operational`, `relaying-payments` or something like that makes more sense?)


eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -499,4 +475,49 @@ object ChannelFlags {
   val Private: ChannelFlags = ChannelFlags(announceChannel = false)
   val Public: ChannelFlags = ChannelFlags(announceChannel = true)
 }
+
+// NB: channel data only matters (and should be stored in the DB) once we have committed funds.
+// If we disconnect or abort before that, we can completely forget the channel.
+sealed trait ChannelData {
```
pm47, line 481: How about:

```scala
sealed trait ChannelData
sealed trait TransientChannelData extends ChannelData
sealed trait PersistentChannelData extends ChannelData
```

This  would allow you to define `case class WaitForFunding* extends TransientChannelData` and keep the corresponding fsm states lean?

t-bast: I can indeed introduce another trait for the transient state. That doesn't change the fact that I want to decouple this from the actual FSM state though, otherwise I won't be able to refactor the channel opening phase to a separate FSM.

When that happens, `Channel.scala` will contain a state `OPENING` that is transient but isn't in the same hierarchy as each state of the channel open protocol. On top of this, the channel opening phase does contain a mix of transient and persistent states (it becomes persistent as soon as we broadcast the funding tx).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -722,97 +722,97 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
...
   when(NORMAL)(handleExceptions {
-    case Event(c: CMD_ADD_HTLC, d: DATA_NORMAL) if d.localShutdown.isDefined || d.remoteShutdown.isDefined =>
+    case Event(c: CMD_ADD_HTLC, DATA_NORMAL(d)) if d.localShutdown.isDefined || d.remoteShutdown.isDefined =>
       // note: spec would allow us to keep sending new htlcs after having received their shutdown (and not sent ours)
       // but we want to converge as fast as possible and they would probably not route them anyway
       val error = NoMoreHtlcsClosingInProgress(d.channelId)
       handleAddHtlcCommandError(c, error, Some(d.channelUpdate))
 
-    case Event(c: CMD_ADD_HTLC, d: DATA_NORMAL) =>
+    case Event(c: CMD_ADD_HTLC, DATA_NORMAL(d)) =>
```
pm47: I'd rather use:
```suggestion
    case Event(c: CMD_ADD_HTLC, normal: DATA_NORMAL) =>
```
otherwise the pattern matching breaks every time we change the case class args. We can always `import normal._` to avoid `normal.data` in the body (but I think `normal.data` looks great).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -722,97 +722,97 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
...
       Commitments.sendAdd(d.commitments, c, nodeParams.currentBlockHeight, nodeParams.onChainFeeConf) match {
         case Right((commitments1, add)) =>
           if (c.commit) self ! CMD_SIGN()
           context.system.eventStream.publish(AvailableBalanceChanged(self, d.channelId, d.shortChannelId, commitments1))
-          handleCommandSuccess(c, d.copy(commitments = commitments1)) sending add
+          handleCommandSuccess(c, DATA_NORMAL(d.copy(commitments = commitments1))) sending add
```
pm47: Let's use lenses? Otherwise it's gonna be tedious when the state class has more than one argument:
```suggestion
          handleCommandSuccess(c, normal.modify(_.data.commitments).setTo(commitments1)) sending add
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
pm47: Okay, it's probably too early to make those design choices anyway.

t-bast: Adding this new trait to the existing PR does make sense though, it lets us group many of the `case class` together, I'll add it.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2173/commits/68d479fa9b720c40aa29e93ee669f23b51005007

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2173/commits/68d479fa9b720c40aa29e93ee669f23b51005007, don't hesitate to nitpick more on those, since I'm touching all those lines already, it's worth making it as good as we can!

Be careful when reviewing though, I hope I didn't mess something up without noticing.

t-bast:
Closing in favor of #2188 where I also extract the channel opening parts to a dedicated actor, I think it makes sense to commit those together and will help motivate the changes (but it is not a trivial PR to review unfortunately)

Issue #2078, opened 2021-11-29
Automatically delete old expired invoices


fiatjaf:
Maybe I have too many invoices?

```
sqlite> select count(*) from received_payments;                                                                                                                                      
48508
sqlite> select count(*) from received_payments where received_at is null;
30772
```

Can I delete these? I'm not interested in keeping expired invoices in the database and would also be glad to get rid of successfully paid invoices from months ago.

t-bast:
You can use the new `deleteinvoice` API: https://github.com/ACINQ/eclair/pull/1984

We also discussed in the past automatically deleting unpaid invoices that have expired, it shouldn't be hard to do, we even tagged this with "good first issue" but nobody picked it up.

fiatjaf:
Can I delete it directly on the SQLite database?

t-bast:
Yes you can, if you check that the invoice is expired and unpaid you're safe.

PR #2174, opened 2022-02-09
purge expired invoice at given time interval

remyers:
Initial attempt at actor for purging invoices periodically for #2078  based roughly on the FileBackupHandler actor.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -79,7 +79,8 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
                       enableTrampolinePayment: Boolean,
                       balanceCheckInterval: FiniteDuration,
                       blockchainWatchdogSources: Seq[String],
-                      onionMessageConfig: OnionMessageConfig) {
+                      onionMessageConfig: OnionMessageConfig,
+                      purgeInvoicesInterval: Option[FiniteDuration]) {
   val privateKey: Crypto.PrivateKey = nodeKeyManager.nodeKey.privateKey
 
   val nodeId: PublicKey = nodeKeyManager.nodeId
@@ -388,6 +389,13 @@ object NodeParams extends Logging {
       case "relay-all" => RelayAll
     }
 
+    val purgeInvoicesInterval = if (config.getBoolean("purge-expired-invoices.enabled")) {
+      Some(FiniteDuration(config.getDuration("purge-expired-invoices.interval").toMinutes, TimeUnit.MINUTES))
+    }
+    else {
+      None
+    }
+
     NodeParams(
       nodeKeyManager = nodeKeyManager,
       channelKeyManager = channelKeyManager,
@@ -491,7 +499,8 @@ object NodeParams extends Logging {
       onionMessageConfig = OnionMessageConfig(
         relayPolicy = onionMessageRelayPolicy,
         timeout = FiniteDuration(config.getDuration("onion-messages.reply-timeout").getSeconds, TimeUnit.SECONDS),
-      )
+      ),
+      purgeInvoicesInterval = purgeInvoicesInterval
     )
   }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -391,8 +391,7 @@ object NodeParams extends Logging {
 
     val purgeInvoicesInterval = if (config.getBoolean("purge-expired-invoices.enabled")) {
       Some(FiniteDuration(config.getDuration("purge-expired-invoices.interval").toMinutes, TimeUnit.MINUTES))
-    }
-    else {
+    } else {
       None
     }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/PurgeInvoicesHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/PurgeInvoicesHandler.scala
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2022 ACINQ SAS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fr.acinq.eclair.db
+
+import akka.Done
+import akka.actor.typed.Behavior
+import akka.actor.typed.eventstream.EventStream
+import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
+import fr.acinq.eclair.{KamonExt, TimestampMilli, TimestampMilliLong}
+import fr.acinq.eclair.db.PurgeInvoicesHandler._
+import fr.acinq.eclair.db.Monitoring.Metrics
+import java.util.concurrent.Executors
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration.FiniteDuration
+import scala.util.{Failure, Success, Try}
+
+
+/**
+ * This actor will purge expired invoices from the database it was initialized with at a scheduled interval.
+ */
+object PurgeInvoicesHandler {
+
+  sealed trait Command
+  private case object TickPurge extends Command
+  private case class PurgeResult(result: Try[Done]) extends Command
+
+  sealed trait PurgeEvent
+  // this notification is sent when we have completed our invoice purge process
+  case object PurgeCompleted extends PurgeEvent
+
+  // the purge task will run in this thread pool
+  private val ec = ExecutionContext.fromExecutor(Executors.newSingleThreadExecutor())
+
+  def apply(paymentsDb: PaymentsDb, interval: FiniteDuration): Behavior[Command] =
+    Behaviors.setup { context =>
+      // wait for purge events sent at `interval`
+      Behaviors.withTimers { timers =>
+        timers.startTimerAtFixedRate(TickPurge, interval)
+        new PurgeInvoicesHandler(paymentsDb, context).waiting(willPurgeAtNextTick = true)
+      }
+    }
+}
+
+class PurgeInvoicesHandler private(paymentsDb: PaymentsDb,
+                                   context: ActorContext[Command]) {
+
+  // purge at each tick unless already currently purging
+  def waiting(willPurgeAtNextTick: Boolean): Behavior[Command] =
+    Behaviors.receiveMessagePartial {
+      case TickPurge => if (willPurgeAtNextTick) {
+        context.log.debug("purging expired invoices")
+        context.pipeToSelf(doPurge())(PurgeResult)
+        purging(willPurgeAtNextTick = false)
+      } else {
+        Behaviors.same
+      }
+    }
+
+  def purging(willPurgeAtNextTick: Boolean): Behavior[Command] =
+    Behaviors.receiveMessagePartial {
+      case PurgeResult(res) =>
+        res match {
+          case Success(Done) => context.log.debug("invoice purge succeeded")
+          case Failure(cause) => context.log.warn(s"invoice purge failed: $cause")
+        }
+        waiting(willPurgeAtNextTick)
+    }
+
+  private def doPurge(): Future[Done] = Future {
+    val now = TimestampMilli.now()
+    KamonExt.time(Metrics.InvoicePurgeDuration.withoutTags()) {
+      val expiredPayments = paymentsDb.listExpiredIncomingPayments(0 unixms, now)
+      expiredPayments.foreach(p => paymentsDb.removeIncomingPayment(p.invoice.paymentHash))
+
+      // publish a notification that we have purged expired invoices
+      context.system.eventStream ! EventStream.Publish(PurgeCompleted)
+      Metrics.InvoicePurgeCompleted.withoutTags().increment()
+    }
+
+    Done
+  }(ec)
+
+}
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
@@ -23,6 +23,7 @@ import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
 import fr.acinq.eclair.db.PaymentsDb
 import fr.acinq.eclair.payment.receive.InvoicePurger.{Command, PurgeCompleted, PurgeResult, TickPurge}
 import fr.acinq.eclair.{TimestampMilli, TimestampMilliLong}
+
 import scala.concurrent.duration.FiniteDuration
 import scala.util.{Failure, Success, Try}
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
@@ -20,7 +20,7 @@ import akka.Done
 import akka.actor.typed.Behavior
 import akka.actor.typed.eventstream.EventStream
 import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
-import fr.acinq.eclair.db.PaymentsDb
+import fr.acinq.eclair.db.IncomingPaymentsDb
 import fr.acinq.eclair.payment.receive.InvoicePurger.{Command, PurgeCompleted, PurgeResult, TickPurge}
 import fr.acinq.eclair.{TimestampMilli, TimestampMilliLong}
 
@@ -32,7 +32,7 @@ import scala.util.{Failure, Success, Try}
  */
 object InvoicePurger {
 
-  def apply(paymentsDb: PaymentsDb, interval: FiniteDuration): Behavior[Command] =
+  def apply(paymentsDb: IncomingPaymentsDb, interval: FiniteDuration): Behavior[Command] =
     Behaviors.setup { context =>
       // wait for purge events sent at `interval`
       Behaviors.withTimers { timers =>
@@ -53,7 +53,7 @@ object InvoicePurger {
   private case object TickPurge extends Command
 }
 
-class InvoicePurger private(paymentsDb: PaymentsDb, context: ActorContext[Command]) {
+class InvoicePurger private(paymentsDb: IncomingPaymentsDb, context: ActorContext[Command]) {
 
   // purge at each tick unless currently purging
   def waiting(): Behavior[Command] =
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
@@ -77,7 +77,12 @@ class InvoicePurger private(paymentsDb: PaymentsDb,
     // purge expired payments
     expiredPayments.foreach(p => paymentsDb.removeIncomingPayment(p.invoice.paymentHash))
 
-    // publish a notification that we have purged expired invoices
-    context.system.eventStream ! EventStream.Publish(PurgeCompleted)
+    // publish a notification when we have purged expired invoices
+    if (expiredPayments.nonEmpty) {
+      context.system.eventStream ! EventStream.Publish(PurgeCompleted)
+    }
+
+    // purge complete
+    context.self ! PurgeResult(Success(Done))
   }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
@@ -16,16 +16,14 @@
 
 package fr.acinq.eclair.payment.receive
 
-import akka.Done
 import akka.actor.typed.Behavior
 import akka.actor.typed.eventstream.EventStream
 import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
 import fr.acinq.eclair.db.IncomingPaymentsDb
-import fr.acinq.eclair.payment.receive.InvoicePurger.{Command, PurgeCompleted, PurgeResult, TickPurge}
+import fr.acinq.eclair.payment.receive.InvoicePurger.{Command, PurgeCompleted, TickPurge}
 import fr.acinq.eclair.{TimestampMilli, TimestampMilliLong}
 
 import scala.concurrent.duration.FiniteDuration
-import scala.util.{Failure, Success, Try}
 
 /**
  * This actor will purge expired invoices from the database it was initialized with at a scheduled interval.
@@ -45,8 +43,6 @@ object InvoicePurger {
 
   sealed trait PurgeEvent
 
-  private case class PurgeResult(result: Try[Done]) extends Command
-
   // this notification is sent when we have completed our invoice purge process
   case object PurgeCompleted extends PurgeEvent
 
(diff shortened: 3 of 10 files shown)


remyers:
I have a simple test written that seems to work, but I want to make sure I'm on the right track before refining/testing it more.
Some questions:
 - does the basic approach of creating a scheduled actor with configured interval launched during setup make sense?
 - is it ok to specify interval in the config file as minutes? seconds/milliseconds seems too short
 - I setup metrics because it seemed easy enough when I looked at the file backup actor, but probably not really necessary. Should I keep that in?
 - I assumed eclair/db was the correct place for this actor, but maybe it should be in a new folder like eclair/maintenance ?
 - I copied the idea of a two state FSM from file backup, even though this actor is unlikely to run longer than the trigger interval. Is that ok, or unnecessary because there's no harm even if two actors happen to run at the same time?

If this look basically correct, I'll try writing more tests and then try testing the node in regtest and then testnet if I can think of anything I can't test in regtest.

t-bast:
Here are my recommendations regarding the architecture:

- this is arguable, but I think this new actor should live in the `payment/receive` folder: that's where we generate invoices, so that's probably where we should clean them up as well
- it shouldn't be a top-level actor (created in `Setup.scala`) but rather a child of the `MultiPartHandler`: we try to avoid having an explosion of top-level actors, it's better to organize them as a tree of logically-grouped entities
- the configuration values would make sense in the `db` section of `reference.conf` and I think should then be added to the `IncomingPaymentsDb` trait: when instantiating the db by reading the configuration, you could inject these values
- you can specify any type of duration in `reference.conf` (seconds, minutes, hours, days), see the `balance-check-interval` field for example (the default value here should probably be 24 hours)
- having metrics for this is overkill, let's remove them, a single log line at the end of the job is enough
- the two state FSM with a timer ticking regularly is the right approach :+1:

eclair-core/src/main/scala/fr/acinq/eclair/db/PurgeInvoicesHandler.scala
```diff
@@ -0,0 +1,97 @@
...
+    Behaviors.receiveMessagePartial {
+      case PurgeResult(res) =>
+        res match {
+          case Success(Done) => context.log.debug("invoice purge succeeded")
+          case Failure(cause) => context.log.warn(s"invoice purge failed: $cause")
+        }
+        waiting(willPurgeAtNextTick)
+    }
+
+  private def doPurge(): Future[Done] = Future {
```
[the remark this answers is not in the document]
t-bast: This is very good advice, you shouldn't need a `Future` here

eclair-core/src/test/scala/fr/acinq/eclair/TestConstants.scala
```diff
@@ -194,7 +194,8 @@ object TestConstants {
       onionMessageConfig = OnionMessageConfig(
         relayPolicy = RelayAll,
         timeout = 1 minute
-      )
+      ),
+      purgeInvoicesInterval = Option(24 hours)
```
thomash-acinq, line 198:
```suggestion
      purgeInvoicesInterval = Some(24 hours)
```

remyers:
> * it shouldn't be a top-level actor (created in `Setup.scala`) but rather a child of the `MultiPartHandler`: we try to avoid having an explosion of top-level actors, it's better to organize them as a tree of logically-grouped entities

I'm not sure the best way to spawn `InvoicePurger` in `MultipartHandler`. Should it happen when a particular message is received? or is there some implicit way to spawn a singleton child actor?

remyers:
> You can simply spawn it when the `MultiPartHandler` is created, ...:

Thanks for the tips; since `MultiPartHandler` is a `ReceiveHandler`, but not an Actor, it only has the context passed into the `handle` function. Perhaps I could instead put the spawning of `InvoicePurger` in the PaymentHandler class? Perhaps after where `defaultHandler` is defined?

remyers:
> You can simply spawn it when the `MultiPartHandler` is created

Thanks for the tips! I think perhaps the purger should be spawned in a similar way but from `PaymentHandler` since `MultiPartHandler` is a function that only gets passed ctx from `PaymentHandler` in it's  `handle` function. Am I missing something?

t-bast:
> the purger should be spawned in a similar way but from PaymentHandler since MultiPartHandler is a function

Right, good catch! That sounds good, spawning it from the `PaymentHandler` is perfect.

remyers:
@t-bast I have a race condition in my test and I'm not sure if it's worth fixing. When I call `addIncomingPayment` in a foreach loop and add many (eg. 1000)  invoices then they may not all have been added to the db when the purger runs. This causes an assert when the test checks if all the expired invoices have been purged.

I didn't find any tests with similar race conditions - most only add two invoices at a time. I tried to find a way to wait for `addIncomingPayment` to finished but I couldn't find a good example of how to do that.

Any suggestions?

t-bast:
> When I call addIncomingPayment in a foreach loop and add many (eg. 1000) invoices then they may not all have been added to the db

Can't you do the following steps?

1. Add many invoices to the DB
2. Then create your invoice purger

To be honest you don't need to stress-test this, adding only a few invoices is enough in unit tests, I don't see any added value in testing hundreds (or more) of deletions?

remyers:
> To be honest you don't need to stress-test this, adding only a few invoices is enough in unit tests, I don't see any added value in testing hundreds (or more) of deletions?

I agree - checking 10 is fine to show the idea. I'll still take a look at awaitCond though to make sure I understand how that works.

t-bast:
`awaitCond` is very handy, it's a non-blocking way of spinning until a condition is met (like a spinlock)

remyers:
I have now tested (regtest) to confirm it purges on the configured interval. The code should be ready for review now.

I assume the documentation in reference.conf is sufficient for this simple feature.

remyers:
I also created a simple functional test script: [purge_invoice.sh](https://github.com/remyers/lightning-cfg/blob/yakshaver/scripts/purge_invoice.sh).

eclair-core/src/main/scala/fr/acinq/eclair/db/PurgeInvoicesHandler.scala
```diff
@@ -0,0 +1,97 @@
...
+        timers.startTimerAtFixedRate(TickPurge, interval)
+        new PurgeInvoicesHandler(paymentsDb, context).waiting(willPurgeAtNextTick = true)
+      }
+    }
+}
+
+class PurgeInvoicesHandler private(paymentsDb: PaymentsDb,
+                                   context: ActorContext[Command]) {
+
+  // purge at each tick unless already currently purging
```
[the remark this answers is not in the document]
remyers: I agree - was just dumb copy/paste error.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
```diff
@@ -0,0 +1,88 @@
...
+        res match {
+          case Success(Done) => context.log.debug("invoice purge succeeded")
+          case Failure(cause) => context.log.warn(s"invoice purge failed: $cause")
+        }
+        waiting()
+    }
+
+  private def doPurge(): Unit = {
+    val now = TimestampMilli.now()
+    val expiredPayments = paymentsDb.listExpiredIncomingPayments(0 unixms, now)
```
t-bast, line 76: It's a bit inefficient to always scan *all* invoices (`from = 0 unixms`). If your node has received a lot of payments (that have been correctly paid), that's a lot of DB rows that we will re-scan regularly for no reason. What we could do to improve this without over-complicating it would be to scan the whole DB on the first run (when eclair has been restarted) and then only scan recent rows on the following runs.

A simple way to do that would be:

- add a boolean to the `waiting` state: `def waiting(fullScan: Boolean): Behavior[Command]`
- start with `fullScan = true`, and then transition to `fullScan = false` after the first run

It would look something like:

```scala
  def waiting(fullScan: Boolean): Behavior[Command] =
    Behaviors.receiveMessage {
      case TickPurge =>
        val now = TimestampMilli.now()
        val start = if (fullScan) 0 unixms else now - 15.days
        // purge expired payments    
        val expiredPayments = paymentsDb.listExpiredIncomingPayments(start, now)
        expiredPayments.foreach(p => paymentsDb.removeIncomingPayment(p.invoice.paymentHash))

        // publish a notification when we have purged expired invoices
        if (expiredPayments.nonEmpty) {
          context.system.eventStream ! EventStream.Publish(PurgeCompleted)
        }
        
        waiting(fullScan = false)
    }
```

The trick is that the interval chosen (15 days) must be greater than the invoice expiry. 15 days feels like a good trade-off, invoices should very likely expire sooner than that, and even if some invoices use a longer expiry, they will be cleaned up at the next full scan (when eclair restarts).

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+import org.scalatest.funsuite.AnyFunSuiteLike
+
+import scala.concurrent.duration.DurationInt
+
+class InvoicePurgerSpec extends TestKitBaseClass with AnyFunSuiteLike {
+
+  import PaymentsDbSpec._
+
+  test("purge invoices") {
+    val dbs = TestPgDatabases()
```
t-bast, line 36: nit: I would use sqlite in tests, it's usually faster

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+    val db = dbs.db.payments
+    val count = 10
+
+    // create expired invoices
+    val expiredInvoices = Seq.fill(count)(Bolt11Invoice(Block.TestnetGenesisBlock.hash, Some(100 msat), randomBytes32(), alicePriv, Left("expired invoice"), CltvExpiryDelta(18),
+      timestamp = 1 unixsec))
+    val expiredPayments = expiredInvoices.map( invoice => IncomingPayment(invoice, randomBytes32(), PaymentType.Standard, invoice.createdAt.toTimestampMilli, IncomingPaymentStatus.Expired))
+    expiredInvoices.lazyZip(expiredPayments).foreach((invoice, payment) => db.addIncomingPayment(invoice, payment.paymentPreimage))
```
t-bast, line 44: You don't need to zip, the payment already contains the invoice (same for the other cases in this test)

```suggestion
    expiredPayments.foreach(payment => db.addIncomingPayment(payment.invoice, payment.paymentPreimage))
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  test("purge invoices") {
+    val dbs = TestPgDatabases()
+    val db = dbs.db.payments
+    val count = 10
+
+    // create expired invoices
+    val expiredInvoices = Seq.fill(count)(Bolt11Invoice(Block.TestnetGenesisBlock.hash, Some(100 msat), randomBytes32(), alicePriv, Left("expired invoice"), CltvExpiryDelta(18),
+      timestamp = 1 unixsec))
+    val expiredPayments = expiredInvoices.map( invoice => IncomingPayment(invoice, randomBytes32(), PaymentType.Standard, invoice.createdAt.toTimestampMilli, IncomingPaymentStatus.Expired))
```
t-bast, line 43: nit: scala formatting doesn't use a space after the parenthesis, if you run `ctrl + alt + l` (lower-case `L`, not `i`) it will automatically fix it for you

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+      timestamp = TimestampSecond.now() - 600))
+    val pendingPayments = pendingInvoices.map( invoice => IncomingPayment(invoice, randomBytes32(), PaymentType.Standard, invoice.createdAt.toTimestampMilli, IncomingPaymentStatus.Pending))
+    pendingInvoices.lazyZip(pendingPayments).foreach((invoice, payment) => db.addIncomingPayment(invoice, payment.paymentPreimage))
+
+    // create paid invoices
+    val receivedAt = TimestampMilli.now() + 1.milli
+    val paidInvoices = Seq.fill(count)(Bolt11Invoice(Block.TestnetGenesisBlock.hash, Some(100 msat), randomBytes32(), alicePriv, Left("paid invoice"), CltvExpiryDelta(18),
+      timestamp = TimestampSecond.now()))
+    val paidPayments = paidInvoices.map( invoice => IncomingPayment(invoice, randomBytes32(), PaymentType.Standard, invoice.createdAt.toTimestampMilli, IncomingPaymentStatus.Received(100 msat, receivedAt)))
+    paidInvoices.lazyZip(paidPayments).foreach((invoice, payment) => { db.addIncomingPayment(invoice, payment.paymentPreimage)
```
t-bast, line 57: nit: it's a bit weird to have `db.addIncomingPayment` on the same line since there are other statements in that block, can you put it on the line below?

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+    expiredInvoices.lazyZip(expiredPayments).foreach((invoice, payment) => db.addIncomingPayment(invoice, payment.paymentPreimage))
+
+    // create pending invoices
+    val pendingInvoices = Seq.fill(count)(Bolt11Invoice(Block.TestnetGenesisBlock.hash, Some(100 msat), randomBytes32(), alicePriv, Left("pending invoice"), CltvExpiryDelta(18),
+      timestamp = TimestampSecond.now() - 600))
+    val pendingPayments = pendingInvoices.map( invoice => IncomingPayment(invoice, randomBytes32(), PaymentType.Standard, invoice.createdAt.toTimestampMilli, IncomingPaymentStatus.Pending))
+    pendingInvoices.lazyZip(pendingPayments).foreach((invoice, payment) => db.addIncomingPayment(invoice, payment.paymentPreimage))
+
+    // create paid invoices
+    val receivedAt = TimestampMilli.now() + 1.milli
```
t-bast, line 53: I don't think you should rely on `TimestampMilli.now()` and `TimestampSecond.now()` (lines 48 and 55) to generate these invoices, you can use purely hard-coded values like you did for expired invoices, it will be less confusing.

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+    assert(db.listPendingIncomingPayments(0 unixms, now) === pendingPayments)
+    assert(db.listReceivedIncomingPayments(0 unixms, now) === paidPayments)
+    assert(db.listExpiredIncomingPayments(0 unixms, now) === expiredPayments)
+
+    val interval = 1 seconds
+
+    val _ = system.spawn(InvoicePurger(db, interval), name = "purge-expired-invoices")
+
+    val probe = TestProbe()
+    system.eventStream.subscribe(probe.ref, classOf[PurgeEvent])
```
t-bast, line 74: You should subscribe *before* spawning the actor, otherwise there is a race condition if the actor emits the event before you've subscribed to it (this kind of race conditions can happen on our slow CI machines since we aggressively parallelize our tests)

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+import fr.acinq.eclair.TestDatabases.TestPgDatabases
+import fr.acinq.eclair.db.{IncomingPayment, IncomingPaymentStatus, PaymentType, PaymentsDbSpec}
+import fr.acinq.eclair.payment.Bolt11Invoice
+import fr.acinq.eclair.payment.receive.InvoicePurger.{PurgeCompleted, PurgeEvent}
+import fr.acinq.eclair.{CltvExpiryDelta, MilliSatoshiLong, TestKitBaseClass, TimestampMilli, TimestampMilliLong, TimestampSecond, TimestampSecondLong, randomBytes32}
+import org.scalatest.funsuite.AnyFunSuiteLike
+
+import scala.concurrent.duration.DurationInt
+
+class InvoicePurgerSpec extends TestKitBaseClass with AnyFunSuiteLike {
```
t-bast, line 31: You should extend the akka typed test kit, this will provide test utilities:

```suggestion
class InvoicePurgerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("application")) with AnyFunSuiteLike {
```

You should then use `testKit.spawn` to spawn your invoice purger.

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+    system.eventStream.subscribe(probe.ref, classOf[PurgeEvent])
+
+    // check that purge completed
+    probe.expectMsg(3 seconds, PurgeCompleted)
+    probe.expectNoMessage()
+    assert(db.listExpiredIncomingPayments(0 unixms, now).isEmpty)
+    assert(db.listIncomingPayments(0 unixms, now) === pendingPayments ++ paidPayments)
+
+    // add more expired invoices
+    expiredInvoices.lazyZip(expiredPayments).foreach((invoice, payment) => db.addIncomingPayment(invoice, payment.paymentPreimage))
```
t-bast, line 83: I now understand why you were concerned with race conditions, there is one here.
Since your invoice purger is running in the background, if it "ticks" before you've finished adding all these new payments your `probe.expectNoMessage()` won't work.

I don't think you need to add multiple invoices here for the test to be relevant. If you add a single new invoice (I would avoid reusing the previous `expiredInvoices`, it's a bit confusing) then this race condition disappears.

t-bast (commented):
Good job, this is very nice for a first scala PR!

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
```diff
@@ -0,0 +1,88 @@
...
+
+import akka.Done
+import akka.actor.typed.Behavior
+import akka.actor.typed.eventstream.EventStream
+import akka.actor.typed.scaladsl.{ActorContext, Behaviors}
+import fr.acinq.eclair.db.PaymentsDb
+import fr.acinq.eclair.payment.receive.InvoicePurger.{Command, PurgeCompleted, PurgeResult, TickPurge}
+import fr.acinq.eclair.{TimestampMilli, TimestampMilliLong}
+import scala.concurrent.duration.FiniteDuration
+import scala.util.{Failure, Success, Try}
```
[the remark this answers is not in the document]
remyers: oh nice! As per CONTRIBUTING.md I also ran ctrl + alt + L  which added some newlines and put the `apply` code at the top of the object. I'll add those shortcuts to my workflow.

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
remyers: I ran ctrl + alt + L on this one too, which fixed that and some spacing issues.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
```diff
@@ -0,0 +1,88 @@
...
+  // ignore ticks and wait for for purge to complete
+  def purging(): Behavior[Command] =
+    Behaviors.receiveMessagePartial {
+      case PurgeResult(res) =>
+        res match {
+          case Success(Done) => context.log.debug("invoice purge succeeded")
+          case Failure(cause) => context.log.warn(s"invoice purge failed: $cause")
+        }
+        waiting()
+    }
```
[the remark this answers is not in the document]
remyers: Nice! the code is much simpler without that unnecessarily `PurgeResult` message and 2nd state.

It appears that each call to `removeIncomingPayment` is run asynchronously via `runAsync` . If I wanted to wait for all of the calls to `removeIncomingPayment` to complete before potentially handling a new  `TickPurge` is there something like `awaitCond` I would need to use? 

I know in this case the interval will be too long for another `TickPurge` to come in, but just curious how that would work.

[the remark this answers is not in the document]
t-bast: We're only using `runAsync` when running in a dual DB mode (where both postgres and sqlite are running in parallel).
That's a mode that we introduced only to thoroughly test postgres before shipping it, but we will likely remove it soon.
So in reality, right now DB calls are mostly synchronous.

If you had an asynchronous call to do inside an actor and wanted to wait for it to complete without blocking, you can use `context.pipeToSelf`, which will let you send a message to yourself once a `Future` (abstraction of an asynchronous call) completes. See https://doc.akka.io/docs/akka/current/typed/interaction-patterns.html#send-future-result-to-self (it's used in many of our akka typed actors).

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -388,6 +389,13 @@ object NodeParams extends Logging {
       case "relay-all" => RelayAll
     }
 
+    val purgeInvoicesInterval = if (config.getBoolean("purge-expired-invoices.enabled")) {
+      Some(FiniteDuration(config.getDuration("purge-expired-invoices.interval").toMinutes, TimeUnit.MINUTES))
+    }
+    else {
+      None
+    }
```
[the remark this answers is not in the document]
remyers: fixed in commit [c1d5c23](https://github.com/remyers/eclair/commit/c1d5c23a6fba678c5d4b09995c3a141e6533872e)

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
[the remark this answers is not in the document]
remyers: fixed in commit [78c24e](https://github.com/remyers/eclair/commit/78c24eecb86afb8d5cf571caa28db948fbe51e68)

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
remyers: fixed in commit [ed1e103](https://github.com/remyers/eclair/commit/ed1e1038014e530e8efd94d86a3cfef9f974a57b)

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
remyers: fixed in commit [4c2fd30](https://github.com/remyers/eclair/commit/4c2fd309cc130586e3a48439a31fde79ef1f7f92)

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
remyers: fixed in commit [a38262f](https://github.com/ACINQ/eclair/pull/2174/commits/a38262f3f71b64fbc59beae0729d3a046cff222d) to create new expired invoices.

I forgot to make it  just a single 2nd expired invoice, but shouldn't hurt to do 10 of them either.

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
remyers: fixed in commit [aedcdf](https://github.com/ACINQ/eclair/pull/2174/commits/aedcdfd38978741c0248575a8df3af4a5ba2f76b)

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
remyers: fixed in commit [853a46b](https://github.com/ACINQ/eclair/pull/2174/commits/853a46b760b54ccbbf4771565c3c2980f7cc74b9)

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
t-bast: Actually it does, it introduces a race condition because your purger actor is also running in parallel and ticking every second. The following scenario can happen:

- you start adding these 10 new invoices to the DB
- you've added 3 out of 10 but the `TickPurge` event happens inside the purger actor
- it will purge those 3 invoices and emit a `PurgeCompleted` event
- meanwhile you've added more invoices to the DB
- `TickPurge` will tick again and purge the other invoices, and emit another `PurgeCompleted` event

So your `probe` will receive two distinct purge events, which will make the `purge.expectNoMessage()` fail.
Since this is a race condition it will rarely happen, but will nonetheless sometimes make the CI build fail, which is a pain...

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
remyers: fixed in commit [ed1e10](https://github.com/remyers/eclair/commit/ed1e1038014e530e8efd94d86a3cfef9f974a57b)

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
remyers: fixed in commit [aedcdf](https://github.com/remyers/eclair/commit/aedcdfd38978741c0248575a8df3af4a5ba2f76b)

eclair-core/src/test/scala/fr/acinq/eclair/payment/receive/InvoicePurgerSpec.scala
remyers: Good point; fixed in commit [600dfa](https://github.com/remyers/eclair/commit/600dfaa0e4941cfe9d2d19f9500dbe3cc7f88b97)

I only needed to set an invoice timestamp with  `TimestampMilli.now()` on line [118](https://github.com/remyers/eclair/commit/600dfaa0e4941cfe9d2d19f9500dbe3cc7f88b97#diff-5a95d11c0f8505b68f4a85b5a6dd3a1f9264880f632080db81c4c16ae2905400R118) in my second test to create an invoice that was expired within the 15 day look back period.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/InvoicePurger.scala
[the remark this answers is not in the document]
remyers: fixed in commit [78c24e](https://github.com/remyers/eclair/commit/78c24eecb86afb8d5cf571caa28db948fbe51e68).

t-bast (approved):
LGTM, good job :+1:

Issue #2168, opened 2022-02-07
Channel is never considered closed if remote doesn't claim its HTLC outputs


pm47:
In case of a force-close with existing htlcs that we can't spend and will never be able to (because the corresponding outgoing htlcs to the next hop has been failed), then if our counterparty never claims its htlc outputs, we will never consider the channel closed and it will stay lingering forever.

The relevant code is `Helpers.Closing.isClosed`, and more precisely the `isDone` methods of `[Local,Remote]CommitPublished`. In the following snippets, `allHtlcsSpent` will be `false` because the map will be non-empty, albeit with values equal to `None`:
https://github.com/ACINQ/eclair/blob/daddee1912e9fc627f0cabe841ca7504e00fef21/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala#L308

https://github.com/ACINQ/eclair/blob/daddee1912e9fc627f0cabe841ca7504e00fef21/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala#L341

We cannot just filter out the `None` values, because we may be able to spend the output later if we get a preimage. Then our counterparty on the right side will pull funds, and the counterparty on our on the left side will let the htlc timeout and be refunded. In other words we lose money.

I believe we should instead handle `CMD_FAIL_HTLC` messages in the `CLOSING` state, just like we do for `CMD_FULFILL_HTLC`. Currently, the post-restart-htlc-cleaner does send `CMD_FAIL_HTLC`, but they are ignored. If we go this way, we probably need to replace the value type of `htlcTxs: Map[OutPoint, Option[HtlcTx]]` from an `Option` to a type with 3 values:
- _known_ (we know how to spend it)
- _unknown_ (we don't know how to spend it, but all hope is not lost)
- _unspendable_ (we know for sure we will never be able to spend it)

PR #2178, opened 2022-02-15
Ignore unspendable htlc-success outputs

pm47:
This is an alternative to #2170 that only focuses on fixing #2168. It is much simpler and does not include any model/codec change.

We just filter htlc-success outputs that we know we will never be able to spend, because the downstream htlc has failed.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -755,7 +755,11 @@ object Helpers {
       val feeratePerKwHtlc = feeEstimator.getFeeratePerKw(target = 2)
 
       // those are the preimages to existing received htlcs
-      val preimages = commitments.localChanges.all.collect { case u: UpdateFulfillHtlc => u.paymentPreimage }.map(r => Crypto.sha256(r) -> r).toMap
+      val preimages: Map[ByteVector32, ByteVector32] = commitments.localChanges.all.collect { case u: UpdateFulfillHtlc => u.paymentPreimage }.map(r => Crypto.sha256(r) -> r).toMap
+      val failedIncomingHtlcs: Set[Long] = commitments.localChanges.all.collect {
+        case u: UpdateFailHtlc => u.id
+        case u: UpdateFailMalformedHtlc => u.id
+      }.toSet
 
       // remember we are looking at the remote commitment so IN for them is really OUT for us and vice versa
       val htlcTxs: Map[OutPoint, Option[ClaimHtlcTx]] = remoteCommit.spec.htlcs.collect {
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -652,28 +652,35 @@ object Helpers {
...
             // incoming htlc for which we have the preimage: we can spend it immediately
-            txInfo.input.outPoint -> withTxGenerationLog("htlc-success") {
+            Some(txInfo.input.outPoint -> withTxGenerationLog("htlc-success") {
               val localSig = keyManager.sign(txInfo, keyManager.htlcPoint(channelKeyPath), localPerCommitmentPoint, TxOwner.Local, commitmentFormat)
               Right(Transactions.addSigs(txInfo, localSig, remoteSig, preimages(paymentHash), commitmentFormat))
-            }
+            })
+          } else if (failedIncomingHtlcs.contains(txInfo.htlcId)) {
+            // incoming htlc that we know for sure will never be fulfilled downstream: we can safely discard it
+            None
```
t-bast, line 671: Good idea, it's a simpler way of filtering out unspendable htlcs without changing anything else, it's a good first step.
I find the use of `Option` and `flatten` a bit hard to read, we could leverage the `collect` and pattern matching instead:

```scala
      val htlcTxs: Map[OutPoint, Option[HtlcTx]] = localCommit.htlcTxsAndRemoteSigs.collect {
        // incoming htlc for which we have the preimage: we can spend it immediately
        case HtlcTxAndRemoteSig(txInfo@HtlcSuccessTx(_, _, paymentHash, _, _), remoteSig) if preimages.contains(paymentHash) =>
          txInfo.input.outPoint -> withTxGenerationLog("htlc-success") {
            val localSig = keyManager.sign(txInfo, keyManager.htlcPoint(channelKeyPath), localPerCommitmentPoint, TxOwner.Local, commitmentFormat)
            Right(Transactions.addSigs(txInfo, localSig, remoteSig, preimages(paymentHash), commitmentFormat))
          }
        // incoming htlc for which we don't have the preimage: we can't spend it immediately, but we may learn the
        // preimage later, otherwise it will eventually timeout and they will get their funds back
        // we ignore htlcs that have been failed downstream: we will never learn their preimage and the funds we may
        // have forwarded have already been refunded 
        case HtlcTxAndRemoteSig(txInfo: HtlcSuccessTx, _) if !failedIncomingHtlcs.contains(txInfo.htlcId) => txInfo.input.outPoint -> None
        case HtlcTxAndRemoteSig(txInfo: HtlcTimeoutTx, remoteSig) =>
          // outgoing htlc: they may or may not have the preimage, the only thing to do is try to get back our funds after timeout
          txInfo.input.outPoint -> withTxGenerationLog("htlc-timeout") {
            val localSig = keyManager.sign(txInfo, keyManager.htlcPoint(channelKeyPath), localPerCommitmentPoint, TxOwner.Local, commitmentFormat)
            Right(Transactions.addSigs(txInfo, localSig, remoteSig, commitmentFormat))
          }
      }.toMap
```

Not entirely sure whether it's better, feel free to chose the version you prefer.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -775,12 +789,12 @@ object Helpers {
...
           }.map(claimHtlcTx => {
-            claimHtlcTx.input.outPoint -> withTxGenerationLog("claim-htlc-timeout") {
+            Some(claimHtlcTx.input.outPoint -> withTxGenerationLog("claim-htlc-timeout") {
               val sig = keyManager.sign(claimHtlcTx, keyManager.htlcPoint(channelKeyPath), remoteCommit.remotePerCommitmentPoint, TxOwner.Local, commitments.commitmentFormat)
               Right(Transactions.addSigs(claimHtlcTx, sig))
-            }
+            })
           })
-      }.toSeq.flatten.toMap
+      }.toSeq.flatten.flatten.toMap
```
t-bast, line 797: Especially here, the `.flatten.flatten` is a bit weird, leveraging `collect` and pattern matching may be more clear?

t-bast (commented):
Looks good to me, I have a small suggestion to improve readability, but otherwise ACK.

PR #2183, opened 2022-02-16
Add txId in logback.xml

t-bast:
We changed the logging MDC key for tx-publishing in #2131, but forgot to add a formatting rule in logback.xml for that new field, so it wasn't written to logs.

### DIFF
--- a/eclair-core/src/test/resources/logback-test.xml
+++ b/eclair-core/src/test/resources/logback-test.xml
@@ -20,7 +20,7 @@
     <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
         <target>System.out</target>
         <encoder>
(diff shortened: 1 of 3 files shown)


sstone:
LGTM but I find the PR description (and the choice of tag `t:` in `logback.xml`) a bit surprising as we're not using a bitcoin transaction id but rather a "publishing attempt id" which is a random id that is specific to eclair ?

t-bast:
What other letter would you use that would make more sense?
`i` (which could have been `id`) and `p` (which could have been `publish`) are already used...
We can change from `t` to something else, but I couldn't find a letter that made more sense (and we don't care that much to be honest, we're more interested in the value it carries).

sstone:
Fair enough.

sstone (approved):

Issue #2176, opened 2022-02-12
eclair still fails to pay via direct channel, and chooses costly multi-hop route instead


viaj3ro:
I've had that issue many times and it still persists. I tried to pay an invoice of a peer whom I have a direct channel with but instead of choosing that channel, eclair takes a random and very costly route:

```
2022-02-12 09:22:07,997 INFO  f.a.eclair.router.Router PAY h:f31f3a8b5d2568600c5a247e58a504c45cf5f9c043d97d3dd091f31936b3ffc9 p:342be1cc i:342be1cc - finding routes 02e9046555a9665145b0dbd7f135744598418df7d61d3660659641886ef1274844->033b63e4a9931dc151037acbce12f4f8968c86f5655cf102bbfa85a26bd4adc6d9 with assistedChannels= ignoreNodes= ignoreChannels= excludedChannels=
2022-02-12 09:22:07,997 INFO  f.a.eclair.router.Router PAY h:f31f3a8b5d2568600c5a247e58a504c45cf5f9c043d97d3dd091f31936b3ffc9 p:342be1cc i:342be1cc - finding routes with randomize=false params=RouteParams(false,SearchBoundaries(21000 msat,0.03,6,CltvExpiryDelta(1008)),Left(WeightRatios(0.0,0.05,0.4,0.55,RelayFees(500 msat,200))),MultiPartParams(15000000 msat,5),control,false)
2022-02-12 09:22:08,691 INFO  f.a.e.p.s.MultiPartPaymentLifecycle PAY n:033b63e4a9931dc151037acbce12f4f8968c86f5655cf102bbfa85a26bd4adc6d9 h:f31f3a8b5d2568600c5a247e58a504c45cf5f9c043d97d3dd091f31936b3ffc9 p:342be1cc i:342be1cc - 1 routes found (attempt=1/5)
[...]

PR #2184, opened 2022-02-17
Add logs about local channel graph inclusion

t-bast:
We've seen a few reports of local channels sometimes being ignored during path-finding (e.g. #2176). When that happens, it's currently hard to know whether the channel is in the graph or not: knowing that would let us know if there is a bug in the graph management code or the path-finding algorithm itself.

I'm adding a few logs to the relevant components:

- when a route request is received, we log the direct channels we have with that peer (if any): that will let us know whether our channels are considered or not
- when a local channel is added or removed from the graph, we log something: that will let us pinpoint whether some channels are missing from the graph when they should be in it


thomash-acinq:
#2176 should be solved now, do you have any other reason to suspect that channels are not added to the graph when they should be?

t-bast:
> do you have any other reason to suspect that channels are not added to the graph when they should be?

I think it would have saved us a lot of time as well when investigating previous issues that seemed related to path finding but were actually disconnections (e.g. at the beginning of https://github.com/ACINQ/eclair/discussions/2045 when the connection timeout was too low for tor peers).

I'm always a bit worried about this because the inclusion of a channel in the graph involves code in many different components (and we've had a few bugs in there in the path) and we're currently in the dark when we only get logs from a node operator, so I think it's useful to have.

thomash-acinq (approved):

Issue #2180, opened 2022-02-16
Do not crash when regtest isn't up yet


NicolasDorier:
Version 0.7.0

I am unable to test eclair properly on https://github.com/btcpayserver/BTCPayServer.Lightning because it crashes as soon as the regtest is up. Instead please consider to wait and retry.

We are disabling tests of Eclair because of that.

NicolasDorier:
```
2022-02-16 05:04:07,216 INFO  fr.acinq.eclair.Setup - using chain=regtest chainHash=06226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf188910f
fatal error: assertion failed: bitcoind should be synchronized (initialblockdownload=true)
2022-02-16 05:04:07,730 ERROR fr.acinq.eclair.Boot - fatal error: assertion failed: bitcoind should be synchronized (initialblockdownload=true)java.lang.AssertionError: assertion failed: bitcoind should be synchronized (initialblockdownload=true)
        at scala.Predef$.assert(Predef.scala:279)
        at fr.acinq.eclair.Setup.<init>(Setup.scala:193)
        at fr.acinq.eclair.Boot$.delayedEndpoint$fr$acinq$eclair$Boot$1(Boot.scala:44)
        at fr.acinq.eclair.Boot$delayedInit$body.apply(Boot.scala:34)
        at scala.Function0.apply$mcV$sp(Function0.scala:39)
        at scala.Function0.apply$mcV$sp$(Function0.scala:39)
        at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:17)
        at scala.App.$anonfun$main$1(App.scala:76)
        at scala.App.$anonfun$main$1$adapted(App.scala:76)
        at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:563)
        at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:561)
        at scala.collection.AbstractIterable.foreach(Iterable.scala:919)
        at scala.App.main(App.scala:76)
        at scala.App.main$(App.scala:74)
        at fr.acinq.eclair.Boot$.main(Boot.scala:34)
        at fr.acinq.eclair.Boot.main(Boot.scala)
```

t-bast:
[...]

PR #2185, opened 2022-02-17
Ignore IBD in regtest

t-bast:
When starting `bitcoind` on `regtest`, it sets the IBD flag until a few blocks have been mined.
However, this isn't a real IBD and it doesn't prevent eclair from working at all, so we can safely ignore that requirement.
People running tests on regtest would still need to mine a few blocks to ensure eclair has funds available (because we directly use `bitcoind`'s wallet), but that should be an unrelated step.

If bitcoind isn't up, eclair will still stop though: we could instead retry in a loop, over a 1 minute period before giving up, do you think it would be useful?

This was initially discussed in #2180

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -187,12 +187,12 @@ class Setup(val datadir: File,
     val (progress, initialBlockDownload, chainHash, bitcoinVersion, unspentAddresses, blocks, headers) = await(future, 30 seconds, "bicoind did not respond after 30 seconds")
     assert(bitcoinVersion >= 180000, "Eclair requires Bitcoin Core 0.18.0 or higher")
     assert(chainHash == nodeParams.chainHash, s"chainHash mismatch (conf=${nodeParams.chainHash} != bitcoind=$chainHash)")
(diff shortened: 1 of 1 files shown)


sstone:
The more I think about this the less comfortable I am with this change: what are the consequences of letting eclair start like this since we rely a lot on bitcoin core working properly ? maybe other tests will fail and it will be difficult to understand why ?

sstone:
What are the limitations of using a bitcoin node that thinks it is out of sync ? Is it limited to the QT GUI displaying a warning or will it also affect wallet RPC calls for example ?

t-bast:
I've tried it, and was able to open/close channels between two eclair nodes without any issue, so it looked ok.

Do note that in this case, `bitcoind` doesn't consider itself out of sync, it has `progress` set to `1` and the same number of blocks and headers, it's really only the `initialblockdownload` flag that is (arguably incorrectly?) set to `true`.

sstone:
`bitcoind` thinks that the current tip is too old, that's why the IBD flag is set, and that flag is checked by a few fee and mempool-related methods.
If will not affect us since we always generate blocks because we need coins (and people who use eclair in regtest mode should do that too), I'm just worried that it may create problems later but don't have a very strong opinion on this.

PR #2190, opened 2022-02-25
Remove support for legacy Sphinx payloads

t-bast:
We previously supported a 65-bytes fixed-size sphinx payload, which has been deprecated in favor of variable length payloads containing a tlv stream (see https://github.com/lightning/bolts/pull/619).

It looks like the whole network now supports the variable-length format, so we can stop accepting the old one. It is also being removed from the spec (see https://github.com/lightning/bolts/pull/962).


eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -90,19 +89,11 @@ object Sphinx extends Logging {
...
-        Metrics.OnionPayloadFormat.withTag(Tags.LegacyOnion, value = true).increment()
-        65
-      case _ =>
-        // The 1.1 BOLT spec changed the frame format to use variable-length per-hop payloads.
-        // The first bytes contain a varint encoding the length of the payload data (not including the trailing mac).
-        // Since messages are always smaller than 65535 bytes, this varint will either be 1 or 3 bytes long.
-        Metrics.OnionPayloadFormat.withTag(Tags.LegacyOnion, value = false).increment()
-        MacLength + PaymentOnionCodecs.payloadLengthDecoder.decode(payload.bits).require.value.toInt
-    }
+    require(payload.head != 0, "legacy onion format is not supported anymore")
```
pm47, line 92: It would probably be more fool proof to make the method return a `Try[Int]` instead of relying on the caller to be careful, but that should probably be done in its own PR.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -279,16 +253,16 @@ object PaymentOnion {
     val invoiceRoutingInfo = records.get[InvoiceRoutingInfo].map(_.extraHops)
   }
 
-  case class FinalTlvPayload(records: TlvStream[OnionPaymentPayloadTlv]) extends FinalPayload with TlvFormat {
-    override val amount = records.get[AmountToForward].get.amount
-    override val expiry = records.get[OutgoingCltv].get.cltv
-    override val paymentSecret = records.get[PaymentData].get.secret
-    override val totalAmount = records.get[PaymentData].map(_.totalAmount match {
+  case class FinalPayload(records: TlvStream[OnionPaymentPayloadTlv]) extends PerHopPayload with PaymentPacket with TrampolinePacket {
```
pm47, line 256: I'm confused, how come all `FinalPayload` are `TrampolinePacket`?

pm47 (commented):
That is some serious cleanup 🧹 💨

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
t-bast: We use these traits (`PaymentPacket` and `TrampolinePacket`) as "this could be this packet type", not "it has to be this packet type". I agree that this is a bit confusing (it was already the case before but it was hidden by a more complex trait hierarchy) but it's quite handy! I'll try to think of an alternative to clean that up.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
t-bast: I'll give this a try, it may not be that much more to add, in which case it makes sense to do now.

t-bast: I does change quite a lot the whole Sphinx file...I think for the best, but I'd rather do that in a separate PR.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
t-bast: This hierarchy is quite handy, because the way we use the `PacketType` is to restrict what kinds of payloads can be used in each context (trampoline or normal payment). I think it works well, but if you have a better alternative let me know!

pm47: Maybe something like that?

```scala
  sealed trait FinalPayload extends PerHopPayload {
    val records: TlvStream[OnionPaymentPayloadTlv]
    val amount = records.get[AmountToForward].get.amount
    ...
    val paymentMetadata = records.get[PaymentMetadata].map(_.data)
  }

  case class PaymentFinalPayload(records: TlvStream[OnionPaymentPayloadTlv]) extends FinalPayload with PaymentPacket

  case class TrampolineFinalPayload(records: TlvStream[OnionPaymentPayloadTlv]) extends FinalPayload with TrampolinePacket
```

Didn't dig much, and that would be for another PR anyway.

t-bast: Agreed, I'll dig into this more in a follow-up PR, along with the improvements in `Sphinx.scala`

pm47 (dismissed):

t-bast:
Thanks for the reviews! Actually we should probably hold off on merging this, when looking at the `crypto.sphinx.onion-payload-format` it looks like a non-negligible portion of the payments we relay are still using the old format...so there may be an implementation out there that supports the new format but still encodes with the legacy one (as eclair did earlier)?
I'll wait for feedback from other implementations before merging, we don't want to lose too much relay fees by deploying this.

pm47 (approved):

PR #2191, opened 2022-02-28
Refactor `Closing.claimRemoteCommitMainOutput()`

pm47:
Building an inconsistent intermediate `RemoteCommitPublished` object was error-prone and defeats the purpose of strong typing.

We also handle the `paysDirectlyToWallet` case directly in the method, which enables a nice factorization and better contains this now deprecated feature.

⚠️ This is a trivial change, but it touches sensitive parts of the code. Please review extra carefully.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -2526,17 +2526,15 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
   private def handleRemoteSpentFuture(commitTx: Transaction, d: DATA_WAIT_FOR_REMOTE_PUBLISH_FUTURE_COMMITMENT) = {
     log.warning(s"they published their future commit (because we asked them to) in txid=${commitTx.txid}")
     context.system.eventStream.publish(TransactionPublished(d.channelId, remoteNodeId, commitTx, Closing.commitTxFee(d.commitments.commitInput, commitTx, d.commitments.localParams.isFunder), "future-remote-commit"))
-    d.commitments.channelFeatures match {
-      case ct if ct.paysDirectlyToWallet =>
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2526,17 +2526,15 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
   private def handleRemoteSpentFuture(commitTx: Transaction, d: DATA_WAIT_FOR_REMOTE_PUBLISH_FUTURE_COMMITMENT) = {
     log.warning(s"they published their future commit (because we asked them to) in txid=${commitTx.txid}")
     context.system.eventStream.publish(TransactionPublished(d.channelId, remoteNodeId, commitTx, Closing.commitTxFee(d.commitments.commitInput, commitTx, d.commitments.localParams.isFunder), "future-remote-commit"))
-    d.commitments.channelFeatures match {
-      case ct if ct.paysDirectlyToWallet =>
-        val remoteCommitPublished = RemoteCommitPublished(commitTx, None, Map.empty, List.empty, Map.empty)
-        val nextData = DATA_CLOSING(d.commitments, fundingTx = None, waitingSince = nodeParams.currentBlockHeight, Nil, futureRemoteCommitPublished = Some(remoteCommitPublished))
-        goto(CLOSING) using nextData storing() // we don't need to claim our main output in the remote commit because it already spends to our wallet address
```
t-bast, line 2533: It looks like at first glance that we're changing behavior here because we will now go through `doPublish(remoteCommitPublished, d.commitments)` in the `paysDirectlyToWallet` case but actually since that `RemoteCommitPublished` only contains the commit tx, the only effect is that we will put a duplicate `WatchConfirmed` on this commit tx (which will be ignored by the watcher), so it looks good to me.

t-bast (approved):
Agreed, the previous code was a bit confusing for no good reason.

PR #2192, opened 2022-02-28
Use direct channel when available

thomash-acinq:
Because of the virtual cost per hop, a path with several "good" channels may be considered better than using a direct channel.
We remove the virtual cost for the first hop of the path to ensure that a direct channel is always selected when available.


thomash-acinq:
Should fix #2176

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -325,7 +325,7 @@ object Graph {
         val totalWeight = prev.weight + (fee + hopCost).toLong * factor
         RichWeight(totalAmount, prev.length + 1, totalCltv, 1.0, totalFees, 0 msat, totalWeight)
       case Right(heuristicsConstants) =>
-        val hopCost = nodeFee(heuristicsConstants.hopCost, prev.amount)
+        val hopCost = if (edge.desc.a == sender) 0 msat else nodeFee(heuristicsConstants.hopCost, prev.amount)
```
t-bast, line 328: Would it make sense to apply the same `if` to the `failureCost` below?

thomash-acinq: It's unnecessary. All paths will have the same `failureCost` for the first edge. Actually it's also unnecessary here for `hopCost`, I'll remove it.

t-bast (approved):

PR #2194, opened 2022-03-01
Separate htlc calculation from local/remote commit (no quicklens)

pm47:
When we receive a preimage for a htlc during a local/remote force close, we only update the htlcs and keep the rest of `LocalCommitPublished` untouched. In particular, we don't regenerate the claim-main tx (which can cause issues if fees change), and leave the `irrevocablySpent` untouched.

Last commit is a rename of several methods, it's better to review it separately. I'm not entirely satisfied with it as the naming logic isn't consistent across closing types and tx types.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1428,7 +1428,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
         case c: CMD_FAIL_MALFORMED_HTLC => Commitments.sendFailMalformed(d.commitments, c)
       }) match {
         case Right((commitments1, _)) =>
-          log.info("got valid fulfill/fail for htlc={}, recalculating htlc transactions", c.id)
+          log.info("got valid settlement for htlc={}, recalculating htlc transactions", c.id)
 
           val localCommitPublished1 = d.localCommitPublished.map(localCommitPublished => localCommitPublished.copy(htlcTxs = Closing.LocalClose.claimHtlcOutputs(keyManager, commitments1)))
           val remoteCommitPublished1 = d.remoteCommitPublished.map(remoteCommitPublished => remoteCommitPublished.copy(claimHtlcTxs = Closing.RemoteClose.claimHtlcOutputs(keyManager, commitments1, commitments1.remoteCommit, nodeParams.onChainFeeConf.feeEstimator)))
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -43,8 +43,8 @@ import fr.acinq.eclair.crypto.keymanager.ChannelKeyManager
 import fr.acinq.eclair.db.DbEventHandler.ChannelEvent.EventType
 import fr.acinq.eclair.db.PendingCommandsDb
 import fr.acinq.eclair.io.Peer
-import fr.acinq.eclair.payment.{Bolt11Invoice, PaymentSettlingOnChain}
 import fr.acinq.eclair.payment.relay.Relayer
+import fr.acinq.eclair.payment.{Bolt11Invoice, PaymentSettlingOnChain}
 import fr.acinq.eclair.router.Announcements
 import fr.acinq.eclair.transactions.Transactions.{ClosingTx, TxOwner}
 import fr.acinq.eclair.transactions._
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -1421,10 +1421,14 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
   })
 
   when(CLOSING)(handleExceptions {
-    case Event(c: CMD_FULFILL_HTLC, d: DATA_CLOSING) =>
-      Commitments.sendFulfill(d.commitments, c) match {
+    case Event(c: HtlcSettlementCommand, d: DATA_CLOSING) =>
+      (c match {
+        case c: CMD_FULFILL_HTLC => Commitments.sendFulfill(d.commitments, c)
+        case c: CMD_FAIL_HTLC => Commitments.sendFail(d.commitments, c, nodeParams.privateKey)
+        case c: CMD_FAIL_MALFORMED_HTLC => Commitments.sendFailMalformed(d.commitments, c)
+      }) match {
         case Right((commitments1, _)) =>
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -1426,13 +1426,10 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, remo
...
-            val remoteCommit = commitments1.remoteNextCommitInfo.swap.toOption.get.nextRemoteCommit
-            Helpers.Closing.claimRemoteCommitTxOutputs(keyManager, commitments1, remoteCommit, remoteCommitPublished.commitTx, nodeParams.currentBlockHeight, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)
-          })
+
+          val localCommitPublished1 = d.localCommitPublished.map(localCommitPublished => localCommitPublished.copy(htlcTxs = Closing.LocalClose.claimHtlcTxOutputs(keyManager, commitments1)))
+          val remoteCommitPublished1 = d.remoteCommitPublished.map(remoteCommitPublished => remoteCommitPublished.copy(claimHtlcTxs = Closing.RemoteClose.claimHtlcTxOutputs(keyManager, commitments1, commitments1.remoteCommit, nodeParams.onChainFeeConf.feeEstimator)))
+          val nextRemoteCommitPublished1 = d.nextRemoteCommitPublished.map(remoteCommitPublished => remoteCommitPublished.copy(claimHtlcTxs = Closing.RemoteClose.claimHtlcTxOutputs(keyManager, commitments1, commitments1.nextRemoteCommit_opt.get, nodeParams.onChainFeeConf.feeEstimator)))
```
pm47, line 1432: Note that here we need to modify an attribute two level deep within the case class. It's not too bad, even without lenses.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -627,366 +630,394 @@ object Helpers {
...
 
-          // then we punish them by stealing their main output
-          val mainPenaltyTx = withTxGenerationLog("main-penalty") {
-            Transactions.makeMainPenaltyTx(commitTx, localParams.dustLimit, remoteRevocationPubkey, localParams.defaultFinalScriptPubKey, localParams.toSelfDelay, remoteDelayedPaymentPubkey, feeratePerKwPenalty).map(txinfo => {
-              val sig = keyManager.sign(txinfo, keyManager.revocationPoint(channelKeyPath), remotePerCommitmentSecret, TxOwner.Local, commitmentFormat)
-              Transactions.addSigs(txinfo, sig)
+      /**
+       * Claim our htlc outputs only
+       */
+      def claimHtlcTxOutputs(keyManager: ChannelKeyManager, commitments: Commitments, remoteCommit: RemoteCommit, feeEstimator: FeeEstimator)(implicit log: LoggingAdapter): Map[OutPoint, Option[ClaimHtlcTx]] = {
```
t-bast, line 819: Same as local close, I think this should be `claimHtlcOutputs` or `generateClaimHtlcTxs`

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -627,366 +630,394 @@ object Helpers {
...
+       * as been spent; note that those transactions will compete with [[HtlcSuccessTx]] and [[HtlcTimeoutTx]] published by the counterparty.
+       * - by spending the delayed output of [[HtlcSuccessTx]] and [[HtlcTimeoutTx]] if those get confirmed; because the output of these txs is protected by
+       * an OP_CSV delay, we will have time to spend them with a revocation key. In that case, we generate the spending transactions "on demand",
+       * this is the purpose of this method.
+       *
+       * NB: when anchor outputs is used, htlc transactions can be aggregated in a single transaction if they share the same
+       * lockTime (thanks to the use of sighash_single | sighash_anyonecanpay), so we may need to claim multiple outputs.
+       */
+      def claimHtlcTxOutputs(keyManager: ChannelKeyManager, commitments: Commitments, revokedCommitPublished: RevokedCommitPublished, htlcTx: Transaction, feeEstimator: FeeEstimator)(implicit log: LoggingAdapter): (RevokedCommitPublished, Seq[ClaimHtlcDelayedOutputPenaltyTx]) = {
```
t-bast, line 980: This one makes sense to keep plural, because our peer may have batched multiple htlc txs into a single one, so the name is good.

t-bast (commented):
Concept ACK, just some bikeshedding on the method names

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: Maybe `claimHtlcTxOutputs`->`claimHtlcOutputs` though?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -627,366 +630,394 @@ object Helpers {
...
-            Transactions.makeClaimRemoteDelayedOutputTx(tx, commitments.localParams.dustLimit, localPaymentPoint, commitments.localParams.defaultFinalScriptPubKey, feeratePerKwMain).map(claimMain => {
-              val sig = keyManager.sign(claimMain, keyManager.paymentPoint(channelKeyPath), TxOwner.Local, commitments.commitmentFormat)
-              Transactions.addSigs(claimMain, sig)
+      /**
+       * Claim the output of a 2nd-stage HTLC transaction. If the provided transaction isn't an htlc, this will be a no-op.
+       *
+       * NB: with anchor outputs, it's possible to have transactions that spend *many* HTLC outputs at once, but we're not
+       * doing that because it introduces a lot of subtle edge cases.
+       */
+      def claimHtlcDelayedOutputs(localCommitPublished: LocalCommitPublished, keyManager: ChannelKeyManager, commitments: Commitments, tx: Transaction, feeEstimator: FeeEstimator, feeTargets: FeeTargets)(implicit log: LoggingAdapter): (LocalCommitPublished, Option[HtlcDelayedTx]) = {
```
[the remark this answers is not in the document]
pm47: Done with 584d49b59.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: Done with 584d49b59.

pm47:
104eb64+9c36977 fix #2168: when we know for sure that an incoming htlc will never be fulfilled, we can safely ignore it. I can put them in a separate PR but this is closely related.

t-bast (approved):
LGTM :+1:

pm47:
🤘

PR #2195, opened 2022-03-01
minFinalExpiryDelta is not optional

thomash-acinq:
In invoices, `minFinalCltvExpiryDelta` is not optional.

An invoice that doesn't explicitly set min_final_cltv_expiry has an default min_final_cltv_expiry of 18.
This default allows invoices to be slightly shorter but does not mean that the min_final_cltv_expiry can be unknown and needs to be provided from somewhere else.


eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt11Invoice.scala
```diff
@@ -89,10 +89,12 @@ case class Bolt11Invoice(prefix: String, amount_opt: Option[MilliSatoshi], creat
 
   lazy val relativeExpiry: FiniteDuration = FiniteDuration(relativeExpiry_opt.getOrElse(DEFAULT_EXPIRY_SECONDS), TimeUnit.SECONDS)
 
-  lazy val minFinalCltvExpiryDelta: Option[CltvExpiryDelta] = tags.collectFirst {
+  lazy val minFinalCltvExpiryDelta_opt: Option[CltvExpiryDelta] = tags.collectFirst {
```
pm47, line 92: My first reaction was to delete `minFinalCltvExpiryDelta_opt`, do we really need to keep the granularity of knowing whether or not the field was actually defined in the invoice? It makes the API less foolproof, and that field was made mandatory almost two years ago (https://github.com/lightning/bolts/pull/785). WDYT? Let's at least add a comment:

```suggestion
  /**
   * @return the content of the 'min_final_cltv_expiry' field if present.
   *         Use [[minFinalCltvExpiryDelta]] if you need the applicable spec-defined value.
   */
  lazy val minFinalCltvExpiryDelta_opt: Option[CltvExpiryDelta] = tags.collectFirst {
```

(same for `relativeExpiry_opt`)

pm47 (commented):
I got confused at first: we cannot put a hard requirement on the existence of that field in `Bolt11Invoice` along the existing ones, because it wasn't always mandatory (https://github.com/lightning/bolts/pull/785):
```scala
require(tags.collect { case _: Bolt11Invoice.MinFinalCltvExpiry=> }.size == 1, "there must be exactly one min_final_cltv_expiry tag")
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt11Invoice.scala
thomash-acinq: The only reason I've kept it is so that the json output matches the invoice exactly. But it doesn't seem important so I've removed it.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt11Invoice.scala
```diff
@@ -89,9 +89,9 @@ case class Bolt11Invoice(prefix: String, amount_opt: Option[MilliSatoshi], creat
 
   lazy val relativeExpiry: FiniteDuration = FiniteDuration(relativeExpiry_opt.getOrElse(DEFAULT_EXPIRY_SECONDS), TimeUnit.SECONDS)
 
-  lazy val minFinalCltvExpiryDelta: Option[CltvExpiryDelta] = tags.collectFirst {
+  lazy val minFinalCltvExpiryDelta: CltvExpiryDelta = tags.collectFirst {
```
pm47, line 92: Nit: Can you make `fallbackAddress`, `relativeExpiry_opt` and `minFinalCltvExpiryDelta` all on a single line? I'm OCDing on the inconsistency.

thomash-acinq: Done, I've actually also removed  `relativeExpiry_opt`

pm47 (approved):
Nice cleanup!

Issue #1741, opened 2021-03-24
`usablebalances` should list all channels


contrebande-labs:
Currently (0.5.1), `usablebalances` will not list a channel which is disabled as per the `channelFlags` in the outgoing/payable direction (and maybe under other circumstances that we have not been made aware of). Furthermore, we have been told by @t-bast that the `canReceive` and `canSend` values from the `usablebalances` result can be strictly positive (more than zero) even if the channel is disabled in the respective direction.

We think that unless the LN spec semantics suggest otherwise, all channels should be listed in the result for the `usablebalances` command, regardless of their incoming or outgoing balances and wether or not they are disabled in any direction for wathever reasons. Furthermore, if a channel is somehow disabled in a direction or in both directions, the respective `canReceive` and `canReceive` values should be set to zero (0), regardless of the actual balances, because that's what the "***can***" in `canReceive` and `canSend` means in natural language.

The first bug here is that If your node has a channel with only an incoming balance, that channel will not appear in the `usablebalances` API call result where it should. The second bug is that a disabled channel could appear with a non-null balance in its `canReceive` or `canSend` field.

Any node with a channel with only an incoming balances can reproduce the issue.

We are using eclair 0.5.1 pre-compiled, downloaded from Github.

The OS is Ubuntu 20.04, the JRE is AdoptOpenJDK 11.0.10.

We were asked to produce a bug report for this issue by @t-bast following a private Gitter conversation.

PR #2196, opened 2022-03-01
Add `channelbalances` API call

remyers:
Solves issue [#1741](https://github.com/ACINQ/eclair/issues/1741) by adding a new API call `channelbalances`. This API lists all channel balances, including balances from disabled channels. Channels with zero `canSend` balance that are not shown by `usablebalances` can instead be seen with `channelbalances`.


remyers:
@t-bast How should I document this new call in the API documentation? should documentation be part of this PR too?

remyers:
> @t-bast How should I document this new call in the API documentation? should documentation be part of this PR too?

I'll add to the release notes the new `channelbalances` documentation so @t-bast can update the API docs when 0.71 is released.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -491,6 +490,9 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
   override def usableBalances()(implicit timeout: Timeout): Future[Iterable[UsableBalance]] =
     (appKit.relayer ? GetOutgoingChannels()).mapTo[OutgoingChannels].map(_.channels.map(_.toUsableBalance))
 
+  override def channelBalances()(implicit timeout: Timeout): Future[Iterable[UsableBalance]] =
```
t-bast, line 493: I think this should return something slightly different than `UsableBalance`: it would be useful to include the `isEnabled` boolean from the channel update. When the channel is disabled, this conceptually wouldn't be a `UsableBalance` anymore...

My suggestion is to rename `UsableBalance` to `ChannelBalance` and enrich it with a new field `isEnabled: Boolean`. The existing `usablebalances` API can then also return an `Iterable[ChannelBalance]`, where `isEnabled` will always be `true` (it's a bit redundant, but explicit, which is good).

eclair-core/eclair-cli
```diff
@@ -44,6 +44,7 @@ and COMMAND is one of the available commands:
     - allchannels
     - allupdates
     - channelstats
+    - channelbalances
```
t-bast, line 47: I think we should be consistent, if you put it in the `Channel` section instead of the `Payment` section, then the code for this API should be in `handlers/Channel.scala` (you currently put it next to `usablebalances` in `handlers/Payment.scala`).

Or we keep the code in `handlers/Payment.scala` and this put it in the `Payment` section here, whatever you prefer.

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -220,6 +221,26 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
       }
   }
 
+  test("'channelbalances' asks relayer for current balances from all channels even if disabled") {
```
t-bast, line 224: Actually this test doesn't verify that we ask the relayer for current balances (and the previous test doesn't either).
This test and the previous one only verify the JSON formatting.
It's worth renaming them to clarify this and adding tests to `EclairImplSpec.scala` to check that we ask the relayer for the right information.

eclair-core/eclair-cli
remyers: I think it's better to move it to `handlers/Channel.scala` since it is not specifically intended to be used for payments like `usablebalances` is.

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
remyers: I got a little side tracked trying to verify the data returned by the `relayer` actor. But then I realized it's enough to just verify that eclair asks the `relayer` for channels with the right messages in `EclairImplSpec.scala`. It's up to the test in `ChannelRelayerSpec.scala` called "get outgoing channels" to exercise the `isEnabled` logic.

If I understand this correctly, fixed in [e8f746d](https://github.com/ACINQ/eclair/pull/2196/commits/e8f746d3c116e7a15914f583252fbc32785e5336).

eclair-core/eclair-cli
remyers: fixed in [a897eb0](https://github.com/ACINQ/eclair/pull/2196/commits/a897eb0f0a3b64e02434f17c28c91596b3a3969c)

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
remyers: > > @t-bast How should I document this new call in the API documentation? should documentation be part of this PR too?
> 
> I'll add to the release notes the new `channelbalances` documentation so @t-bast can [update the API docs](https://github.com/ACINQ/eclair/blob/master/BUILD.md#build-the-api-documentation) when 0.71 is released.

fixed in [cdd0e38](https://github.com/ACINQ/eclair/pull/2196/commits/cdd0e385054ef253c0f26ff1a972a6ab72924730)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -139,12 +139,13 @@ object Relayer extends Logging {
...
   case class OutgoingChannel(nextNodeId: PublicKey, channelUpdate: ChannelUpdate, commitments: AbstractCommitments) {
-    def toUsableBalance: UsableBalance = UsableBalance(
+    def toChannelBalance: ChannelBalance = ChannelBalance(
       remoteNodeId = nextNodeId,
       shortChannelId = channelUpdate.shortChannelId,
       canSend = commitments.availableBalanceForSend,
       canReceive = commitments.availableBalanceForReceive,
-      isPublic = commitments.announceChannel)
+      isPublic = commitments.announceChannel,
+      isEnabled = channelUpdate.channelFlags.isEnabled)
```
t-bast, line 148: Actually there is a subtlety here: `isEnabled` only applies to our side of the channel, there is a separate `channel_update` for the other direction (our peer -> us).

Ideally we would expose two bools: `sendEnabled` and `receiveEnabled` (our current `isEnabled` is actually `sendEnabled`) to have the most accurate data. However the relayer only keeps our own `channe_ update`, not our peer's `channel_update` (because we only want to know if we can send, we don't care about receiving).

I really don't think it's worth fixing, unless someone has a very compelling use-case for it, but I thought it was interesting to raise the point.

t-bast (approved):
LGTM, just needs a rebase!

Issue #2110, opened 2021-12-20
[Feature Request] Delay channel fee update until gossip has caught up


viaj3ro:
AFAIK c-lightning and LND already implemented a feature where raised channel relay fees taking effect is delayed by 10 minutes or so to avoid routing failures and give the update message enough time to propagate through the gossip network.

If possible, this should come with an option to optionally enforce the new fee right away, when routing failures are preferred over a quickly depleted channel, i.e. when a channel was opened with an erroneously low fee.

Thinking about it, even lowering fees should come with a delay, to have time to fix typos that caused a way to low fee.

t-bast:
I think this makes sense, we can simply be lenient and keep accepting our previous fees for up to X minutes.

To keep it simple, we can make this an in-memory-only feature: we keep that information locally but if eclair restarts, we don't guarantee that we will allow old fees.

remyers:
The equivalent issue in c-lightning is [#4761](https://github.com/ElementsProject/lightning/issues/4761). This issue was fixed by adding the `enforcedelay` grace period (default 10 minutes) before enforcing a change using `setchannelfee`.

I could not find the equivalent feature in LND, anyone have a link?

remyers:
Implemented with PR #2201

PR #2201, opened 2022-03-04
Relay if fees satisfy previous channel update for X minutes after current update

remyers:
Adds feature  "Delay channel fee update until gossip has caught up [#2110](https://github.com/ACINQ/eclair/issues/2110)"

During the `relay.fees.enforce-delay` period after a local channel update, the channel relayer will not fail payments that satisfy the base and proportional relay fee from either the current or previous channel update.

When a local channel update comes in, we store the previous channel update in the `OutgoingChannel` of the relayer along with the new current channel update. If the current channel update is less than `relay.fees.enforce-delay` minutes old we will not fail for insufficient relay fees if a relayed payment satisfies either the current or previous channel update.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -132,6 +132,8 @@ eclair {
         fee-base-msat = 1000
         fee-proportional-millionths = 100
       }
+      // Delay enforcement of channel fee updates
+      enforce-delay = 10 minutes
     }
   }
 
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -133,7 +133,7 @@ eclair {
         fee-proportional-millionths = 100
       }
       // Delay enforcement of channel fee updates
-      enforce-delay = 10 minutes
+      enforcement-delay = 10 minutes
     }
   }
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -21,8 +21,8 @@ import fr.acinq.bitcoin.Crypto.PublicKey
 import fr.acinq.bitcoin.{Block, ByteVector32, Crypto, Satoshi}
 import fr.acinq.eclair.Setup.Seeds
 import fr.acinq.eclair.blockchain.fee._
-import fr.acinq.eclair.channel.{Channel, ChannelFlags}
 import fr.acinq.eclair.channel.Channel.{ChannelConf, UnhandledExceptionStrategy}
+import fr.acinq.eclair.channel.{Channel, ChannelFlags}
 import fr.acinq.eclair.crypto.Noise.KeyPair
 import fr.acinq.eclair.crypto.keymanager.{ChannelKeyManager, NodeKeyManager}
 import fr.acinq.eclair.db._
@@ -80,7 +80,8 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
                       balanceCheckInterval: FiniteDuration,
                       blockchainWatchdogSources: Seq[String],
                       onionMessageConfig: OnionMessageConfig,
-                      purgeInvoicesInterval: Option[FiniteDuration]) {
+                      purgeInvoicesInterval: Option[FiniteDuration],
+                      enforceDelay: FiniteDuration) {
   val privateKey: Crypto.PrivateKey = nodeKeyManager.nodeKey.privateKey
 
   val nodeId: PublicKey = nodeKeyManager.nodeId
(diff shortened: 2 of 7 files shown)


remyers:
With this implementation if you want your fees enforced immediately then you can issue the same update twice so that the prevChannelUpdate == the current channelUpdate for the channel.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
```diff
@@ -261,7 +267,7 @@ class ChannelRelay private(nodeParams: NodeParams,
         RelayFailure(CMD_FAIL_HTLC(add.id, Right(AmountBelowMinimum(payload.amountToForward, channelUpdate)), commit = true))
       case Some(channelUpdate) if r.expiryDelta < channelUpdate.cltvExpiryDelta =>
         RelayFailure(CMD_FAIL_HTLC(add.id, Right(IncorrectCltvExpiry(payload.outgoingCltv, channelUpdate)), commit = true))
-      case Some(channelUpdate) if r.relayFeeMsat < nodeFee(channelUpdate, payload.amountToForward) =>
+      case Some(channelUpdate) if r.relayFeeMsat < nodeFee(channelUpdate, payload.amountToForward) && !usePrevFees =>
```
thomash-acinq, line 270: I don't think you should define `usePrevFees` far from where it's used, it fits in one line here.

```suggestion
      case Some(channelUpdate) if r.relayFeeMsat < nodeFee(channelUpdate, payload.amountToForward) &&
        outgoingChannel_opt.forall(c => TimestampSecond.now() - c.channelUpdate.timestamp > nodeParams.enforceDelay || r.relayFeeMsat < nodeFee(c.prevChannelUpdate, payload.amountToForward)) =>
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -130,6 +130,7 @@ object Relayer extends Logging {
   }
 
   case class RelayForward(add: UpdateAddHtlc)
+
```
thomash-acinq, line 133: Don't add an empty line here

thomash-acinq (commented):
I looks good, just a few things to fix.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
remyers: Oh nice. I like how the logic mirrors the current channel update also.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -499,7 +500,8 @@ object NodeParams extends Logging {
         relayPolicy = onionMessageRelayPolicy,
         timeout = FiniteDuration(config.getDuration("onion-messages.reply-timeout").getSeconds, TimeUnit.SECONDS),
       ),
-      purgeInvoicesInterval = purgeInvoicesInterval
+      purgeInvoicesInterval = purgeInvoicesInterval,
+      enforceDelay = FiniteDuration(config.getDuration("relay.fees.enforce-delay").toMinutes, TimeUnit.MINUTES)
```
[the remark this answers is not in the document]
remyers: Fixed in [79a4f6d](https://github.com/ACINQ/eclair/pull/2201/commits/79a4f6dc79bb48153feda21cbde43e1cef8361b6)

eclair-core/src/main/resources/reference.conf
```diff
@@ -132,6 +132,8 @@ eclair {
         fee-base-msat = 1000
         fee-proportional-millionths = 100
       }
+      // Delay enforcement of channel fee updates
+      enforce-delay = 10 minutes
```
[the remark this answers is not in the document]
remyers: Fixed in [79a4f6d](https://github.com/ACINQ/eclair/pull/2201/commits/79a4f6dc79bb48153feda21cbde43e1cef8361b6)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
remyers: I had to add a `flatmap` because `nodeFee` does not take an `Option`. It's a bit less elegant but I couldn't find a better way to get it working.  Fixed in [a302f32](https://github.com/ACINQ/eclair/pull/2201/commits/a302f32f5cd3a7c22c22b342e8a54fe09ad15872).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
remyers: fixed in [ce0b37e](https://github.com/ACINQ/eclair/pull/2201/commits/ce0b37e6d83dbd9e0320cc46ba312f33331bb4b0)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
thomash-acinq: You're right, I had messed up the different channel updates, this should be correct.

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
thomash-acinq: I meant to change the name everywhere, you still use `enforceDelay` in other places.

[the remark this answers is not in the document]
remyers: done! see [1f06b4e](https://github.com/ACINQ/eclair/pull/2201/commits/1f06b4e19e2fed1e08064a04fc853c9940d1e55a)

thomash-acinq (approved):

PR #2202, opened 2022-03-08
Use `NodeAddress` everywhere instead of `InetAddress`

pm47:
This makes us control more strictly when and where name resolution happens, which is important in a security-hardened setup. The `InetAddress` jdk class indeed does a lot of things behind the scenes, but now we restrict it to tcp-related classes like `Client` and `Server`.

Also, in _cluster mode_ all outgoing connections (including tor) are now made on the front.


t-bast (commented):
Nice change, this will also make it much easier to implement https://github.com/lightning/bolts/pull/911

> Also, in cluster mode all outgoing connections (including tor) are now made on the front.

Can you remind me why we previously couldn't resolve tor addresses on the front?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -248,18 +248,17 @@ object NodeAddress {
 }
 
 object IPAddress {
-  def apply(inetAddress: InetAddress, port: Int): Option[IPAddress] = inetAddress match {
-    case address: Inet4Address => Some(IPv4(address, port))
-    case address: Inet6Address => Some(IPv6(address, port))
-    case _ => None
+  def apply(inetAddress: InetAddress, port: Int): IPAddress = inetAddress match {
+    case address: Inet4Address => IPv4(address, port)
+    case address: Inet6Address => IPv6(address, port)
```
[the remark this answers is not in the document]
pm47: 🦅 👁️ 

We can't! It's a trade-off. I had an offline discussion with @sstone and we agreed it was worth it. At worst if a new ip type shows up there will be a pattern matching exception, the `Server` will die and be restarted.

pm47:
> Can you remind me why we previously couldn't resolve tor addresses on the front?

There is no real reason, except that a prerequisite was to move the tor gateway on a separate machine, which happened later.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -232,7 +233,7 @@ object NodeAddress {
     host match {
       case _ if host.endsWith(".onion") && host.length == 22 => Tor2(host.dropRight(6), port)
       case _ if host.endsWith(".onion") && host.length == 62 => Tor3(host.dropRight(6), port)
-      case _ => IPAddress(InetAddress.getByName(host), port).get
+      case _ => IPAddress(InetAddress.getByName(host), port)
```
pm47, line 236: Note: as explained in the method comment, this does a name resolution. We could decide to drop that feature and replace by `InetAddresses.forUriString()` (which would ne a nice symmetry to the newly added `InetAddresses.toUriString()`) and would guarantee us that there is no name resolution, but supporting dns-based node uris seems useful.

t-bast: We will indeed need name resolution (see https://github.com/lightning/bolts/pull/911), I've asked @remyers to take a stab at implementing it once this PR is merged.

t-bast (approved):
LGTM

Issue #2077, opened 2021-11-28
Private channels should have their own dedicated  'min-funding-satoshis' setting


viaj3ro:
IMHO there is no reason, why peers shouldn't be able to open small private channels. Minimum size should be configurable with a dedicated `min-funding-satoshis` setting for private channels.

**Edit:** 
The idea is not to enable super small private channels, but rather to be able to raise the limit for public channels without affecting private ones.

t-bast:
But why would there be a reason for it?
The reason you set `min-funding-satoshis` is to avoid being spammed by other people opening channels to you.
If you use a lower value for private channels, then people can "spam" you with private channels instead of public ones, so it defeats the purpose...

btcontract:
I'd say it's not exactly spamming but rather for public channels we are interested in larger capacity because it increases routing success chances and makes it more attractive for pathfinding heuristics in other nodes. None of these apply to private channels and spamming is not free for attacker, and quite inconsequential for our node.

btcontract:
I can imagine setting some reasonable values like min public=5m SAT, min private=1m SAT to not make it prohibitively high for mobile wallet end users who wish to open a private channel to an arbitrary node. Dangers of way too low limits can be put in a config comments, there are quite a few of such warning there already.

t-bast:
It's true, we can document this and let users decide whether they're willing to take that risk or not.

You're saying "prohibitively high for mobile wallet end users" but there's absolutely no guarantee that people opening private channels to you will be mobile wallet users, they can very easily be malicious full nodes.

In your example of `5m SAT` and `1m SAT`, it means you make it 5 times less capital intensive for a potential attacker wanting to exploit a flood and loot attack against your node.

btcontract:
Yes, just speaking from my perspective: I did encounter mobile users frustrated with capacity requirements.

viaj3ro:
> I can imagine setting some reasonable values like min public=5m SAT, min private=1m SAT to not make it prohibitively high for mobile wallet end users who wish to open a private channel to an arbitrary node. Dangers of way too low limits can be put in a config comments, there are quite a few of such warning there already.

this! My min channel size is 5m but I don't mind offering 1m or 500k min size for private channels (depending on BTC price) to help out mobile users.

There is no added risk here as long as you set the same standard values for both and inform users that they should be aware what they are doing if they go lower.

Realistically though, it's more about going higher than standard setting with public channels, not lower with private.

PR #2203, opened 2022-03-10
new min-funding config parameter for private channels 

remyers:
Add the ability to set a different min-funding limit for private and public channels. See issue "Private channels should have their own dedicated 'min-funding-satoshis' setting" [#2077](https://github.com/ACINQ/eclair/issues/2077).

If `channel.private.enable-alt` = `true` then the `channel.private.min-funding-satoshis` value will be used for private channels, otherwise the `channel.min-funding-satoshis` will be used for both public and private channels.

A few questions:

- previous `min-funding-satoshis` tests compared `assert(error.toAscii ===  Error(...).toAscii)` while other tests in `WaitForOpenChannelStateSpec` left off the `.toAscii`. Is there a reason to use the `.toAscii` in the assert here?
- [Issue 2077 mentioned we should document](https://github.com/ACINQ/eclair/issues/2077#issuecomment-981415320) that users should not set unsafely low min-funding limits for private channels. Only if users set their public channels to have higher limits then they might allow private channels with a lower, but still safe, `min-funding-satoshis`. Should I add documentation to that effect in [docs\Configure.md](https://github.com/ACINQ/eclair/blob/master/docs/Configure.md) ? or in the [release notes](https://github.com/ACINQ/eclair/blob/master/docs/release-notes/eclair-vnext.md) ?


remyers:
My rational for the `channel.private.enable-alt` is that when people upgrade their node it should default to have the same min-funding behavior as before. If we just added a new setting for private channels then it would use whatever default was in the reference.conf and that would potentially be very different than what the user expects their private channels to use.

I'm not completely happy with the `enable-alt` as the name for that parameters - open to suggestions.

pm47:
@thomash-acinq beat me to it, indeed it's the way to go:
https://github.com/ACINQ/eclair/blob/f3c6c7885de5b4c8c48302b06b25b1e7422c763c/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala#L173

remyers:
> I understand your rationale for `channel.private.enable-alt` but I really don't like the result. An alternative would be to rename `min-funding-satoshis` to `min-public-funding-satoshis` and add a deprecation message for `min-funding-satoshis`, that way everyone using this parameter will see the message and will be forced to manually change it.

I agree. As long as we mention both new options in the deprecation message that should ensure they know to set both the public and private min funding values.

eclair-core/src/main/resources/reference.conf
```diff
@@ -82,7 +82,8 @@ eclair {
 
     reserve-to-funding-ratio = 0.01 // recommended by BOLT #2
     max-reserve-to-funding-ratio = 0.05 // channel reserve can't be more than 5% of the funding amount (recommended: 1%)
-    min-funding-satoshis = 100000
+    min-public-funding-satoshis = 100000 // high public minimums are attractive to pathfinding heuristics
+    min-private-funding-satoshis = 100000 // safe private minimums increase the cost of spamming and flood-and-loot attacks
```
thomash-acinq, line 86: I don't think these comments are really helpful. Larger channels are more attractive but that doesn't have anything to do with the minimum we set here, a small channel is more attractive than no channel.

remyers: Removed them in [6534977](https://github.com/ACINQ/eclair/pull/2203/commits/6534977f9f6d81813c31190663da069d230c0343). Sounds like its too nuanced to explain in comments.

thomash-acinq (approved):

thomash-acinq:
I've just remembered about the release notes, can you update them too?

pm47:
Also, for PR names and squashed commit messages please use a short sentence starting by a verb with an uppercase.

PR #2204, opened 2022-03-10
Add delayed fees to release notes

thomash-acinq:
Add release notes for #2201

### DIFF
--- a/docs/release-notes/eclair-vnext.md
+++ b/docs/release-notes/eclair-vnext.md
@@ -15,7 +15,7 @@
 #### Delay enforcement of new channel fees
 
 When updating the relay fees for a channel, eclair can now continue accepting to relay payments using the old fee even if they would be rejected with the new fee.
-The amount of time during which eclair still accepts the old fee is configured with `eclair.relay.fees.enforcement-delay`.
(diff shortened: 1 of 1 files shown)


pm47:
> @pm47 Should we set eclair.relay.fees.enforcement-delay to 0 by default so that there is no change in the default behaviour?

No I think the new default of 10 min is okay, but I would state the value explicitly in the release note.

docs/release-notes/eclair-vnext.md
```diff
@@ -12,7 +12,12 @@
 
 ### Miscellaneous improvements and bug fixes
 
-<insert changes>
+#### Delay enforcement of new channel fees
+
+When updating the relay fees for a channel, eclair can now continue accepting to relay payments using the old fee even if they would be rejected with the new fee.
```
remyers, line 17:
```suggestion
When updating the relay fees for a channel, eclair can now continue accepting relay payments using the old fee even if they would be rejected with the new fee.
```

remyers: typo

remyers (approved):
Looks good. I think it's ok to default to 10 minutes for updated nodes - Rusty's research shows that on average that's how long it takes for updates to propagate so it should be safe for planned updates. 

Only in the case where someone makes an incorrect change and does not know how to invalidate it immediately is there a problem - so it's good that this is discussed in the release notes.

PR #2207, opened 2022-03-15
Replace `FeatureScope` by type hierarchy

pm47:
Instead of defining a separate type `FeatureScope` with its own hierarchy, that was then mixed in `Feature` using the cake pattern, we go with a simpler type hierarchy for `Feature`.

This significantly simplifies type declarations (no more `Feature with FeatureScope`) especially in the tests.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -34,9 +34,8 @@ object FeatureSupport {
   case object Optional extends FeatureSupport { override def toString: String = "optional" }
 }
 
+/** Not a sealed trait, so it can be extended by plugins. */
 trait Feature {
-
-  this: FeatureScope =>
 
   def rfcName: String
   def mandatory: Int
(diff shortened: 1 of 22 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -84,13 +81,13 @@ case class Features[T <: FeatureScope](activated: Map[Feature with T, FeatureSup
     unknownFeaturesOk && knownFeaturesOk
   }
 
-  def initFeatures(): Features[InitFeature] = Features[InitFeature](activated.collect { case (f: InitFeature, s) => (f, s) }, unknown)
+  def initFeatures(): Features[InitFeature] = Features(activated.collect { case (f: InitFeature, s) => (f, s) }, unknown)
```
pm47, line 84: Compiler doesn't need that hint anymore.

eclair-core/src/test/scala/fr/acinq/eclair/FeaturesSpec.scala
```diff
@@ -235,8 +235,8 @@ class FeaturesSpec extends AnyFunSuite {
       hex"" -> Features.empty,
       hex"0100" -> Features(VariableLengthOnion -> Mandatory),
       hex"028a8a" -> Features(DataLossProtect -> Optional, InitialRoutingSync -> Optional, ChannelRangeQueries -> Optional, VariableLengthOnion -> Optional, ChannelRangeQueriesExtended -> Optional, PaymentSecret -> Optional, BasicMultiPartPayment -> Optional),
-      hex"09004200" -> Features[FeatureScope](Map[Feature with FeatureScope, FeatureSupport](VariableLengthOnion -> Optional, PaymentSecret -> Mandatory, ShutdownAnySegwit -> Optional), Set(UnknownFeature(24))),
-      hex"52000000" -> Features[FeatureScope](Map.empty[Feature with FeatureScope, FeatureSupport], Set(UnknownFeature(25), UnknownFeature(28), UnknownFeature(30)))
+      hex"09004200" -> Features(Map(VariableLengthOnion -> Optional, PaymentSecret -> Mandatory, ShutdownAnySegwit -> Optional), Set(UnknownFeature(24))),
```
pm47, line 238: Much simpler right?

thomash-acinq (approved):
Much nicer indeed.

Issue #1868, opened 2021-07-11
Mainnet node doesn't start without wallet name


engenegr:
I have two setups using bitcoind v21.99.0.

Eclair testnet (SQLite backend) has started without `wallet` parameter in config while mainnet (Postgre) didn't and I could only make it working with provided `wallet`.

I had to restart mainnet bitcoind because I was thinking that it was in failed state but apparently wallet was loaded and I reloaded it again after restart. Everything returned back to normal only after explicit definition `eclair.bitcoind.wallet`.

t-bast:
What is bitcoind v21.99.0? Note that we may not support bitcoind versions outside of the ones we recommend (0.20.1 or 0.21.1).

> apparently wallet was loaded and I reloaded it again after restart. Everything returned back to normal only after explicit definition eclair.bitcoind.wallet.

I think this is your issue: you had a wallet loaded in `bitcoind` and eclair didn't know about it.
We want this configuration to be very explicit, because we don't want eclair to be using funds from the linked bitcoind that it's not supposed to use.

IMO this is expected behavior. You must verify the wallet you configure for eclair before starting up. What kind of different behavior would you expect? Maybe better error messages to guide you to configure the wallet? We have a section in our README for that (https://github.com/ACINQ/eclair#configure-bitcoin-core-wallet), it's only the term "default loaded bitcoin wallet" which may be misleading.

engenegr:
> IMO this is expected behavior. You must verify the wallet you configure for eclair before starting up.

Fine. But I haven't configured it for testnet, started node, it said my wallet is not loaded into bitcoind, then I loaded it and restarted node. It was running without config on the testnet. This is why I didn't expect any disruption on Mainnet.

Only after Eclair declined to start without wallet (when in fact it was loaded) I discovered that there is new option now.

t-bast:
I think the difference you saw between testnet and mainnet is because in one case you had a wallet loaded (different from the `""` default wallet).

I agree though that we could make this better, it's hard to understand what's wrong for users.
At startup, we could make a few calls to `bitcoind` to list wallets and check how it's configured, and return a more detailed error message to users if we think it's not properly configured. Feel free to try to implement that if you have time, otherwise we'll do it when we have some buffer.

engenegr:
> I think the difference you saw between testnet and mainnet is because in one case you had a wallet loaded (different from the "" default wallet)

Yes. I agree, it looks like that.
On both machines I have bitcoind wallets with same names.

[...]

PR #2209, opened 2022-03-17
Add messages for when startup fails due to wrong bitcoind wallet(s) loaded

remyers:
Add more specific startup failure messages for situations when the configured `eclair.bitcoind.wallet` does not match the wallet(s) loaded in bitcoind.

The default `eclair.bitcoind.wallet` is the empty string. This corresponds with the situation before multi-wallet support was added to bitcoind where only one wallet could be loaded at a time. When this default setting is used, Eclair will fail at startup if more than one wallet is loaded in bitcoind. 

If only one wallet is loaded in bitcoind, then it will be used regardless of its name if `eclair.bitcoind.wallet = ""`.

If `eclair.bitcoind.wallet` is configured with a wallet name other than the empty string, then multiple wallets can be loaded in bitcoind as long as one of them is the configured wallet.

If bitcoind is launched (or compiled) with `-disablewallet` then Eclair will fail at startup with a message to indicate this.

This PR addresses issue: "Mainnet node doesn't start without wallet name" [#1868](https://github.com/ACINQ/eclair/issues/1868)

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -165,18 +165,22 @@ class Setup(val datadir: File,
     val future = for {
       json <- bitcoinClient.invoke("getblockchaininfo").recover { case e => throw BitcoinRPCConnectionException(e) }
       // Make sure wallet support is enabled in bitcoind.
-      _ <- bitcoinClient.invoke("getbalance").recover { case e => throw BitcoinWalletDisabledException(e) }
+      wallets <- bitcoinClient.invoke("listwallets").recover { case e => throw BitcoinWalletDisabledException(e) }
+        .collect {
+          case JArray(values) => values.map(value => value.extract[String])
+        }
       progress = (json \ "verificationprogress").extract[Double]
       ibd = (json \ "initialblockdownload").extract[Boolean]
       blocks = (json \ "blocks").extract[Long]
       headers = (json \ "headers").extract[Long]
       chainHash <- bitcoinClient.invoke("getblockhash", 0).map(_.extract[String]).map(s => ByteVector32.fromValidHex(s)).map(_.reverse)
       bitcoinVersion <- bitcoinClient.invoke("getnetworkinfo").map(json => json \ "version").map(_.extract[Int])
-      unspentAddresses <- bitcoinClient.invoke("listunspent").collect { case JArray(values) =>
-        values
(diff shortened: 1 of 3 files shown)


remyers:
I wasn't entirely sure where the tests should go directory/name wise, but I believe they need to be their own test suite because BitcoinCoreClientSpec does not extend `IntegrationSpec` needed for these tests.

sstone:
LGTM. Just one nit: there is no test where bitcoind starts without a wallet (but it would require changes to some of our base test classes).

sstone:
Unrelated to this PR, but it seems that users cannot specify an empty wallet (`""`) as it would be cast to `None` in https://github.com/ACINQ/eclair/blob/master/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala#L151. So if a user has 2 wallets `"mywallet"` and `""` they cannot select `""` ?

remyers:
> Just one nit: there is no test where bitcoind starts without a wallet (but it would require changes to some of our base test classes).

I managed to add a test ([36d0b8b](https://github.com/ACINQ/eclair/pull/2209/commits/36d0b8b66c4561f0eb10d969fc141161aa56f8eb)) for this case that avoids creating a new `bitcoincli` object for the `""` wallet as long as it runs before any test that creates a `""` wallet.

The error when no wallets is loaded looks like: `fatal error: configured wallet "" not in the set of loaded bitcoind wallets: []`

I  wish I didn't have to set the `"eclair.server.port"` in the config for each test. It seems like it should be defaulting to unused ports for each node, but in practice I get errors if I don't set the port in the config for each test.

remyers:
Interestingly [9d8bee8](https://github.com/ACINQ/eclair/pull/2209/commits/9d8bee8272050b9a37a8e4aca14847f2559e8610) works with the v21.99 build of bitcoind, but not with v0.21.1 of bitcoind. This difference can be seen directly by running these RPC commands on both v21.99 and v0.21.1:
```
btcoin-cli createwallet "miner"
bitcoin-cli -rpcwallet=miner unloadwallet "miner"
```
An error is thrown for v0.21.1 but fixed for v21.99 via PR [#20448](https://github.com/bitcoin/bitcoin/pull/20448). Eclair currently always does the equivalent of adding `-rpcwallet=` unless `eclair.bitcoind.wallet=""` is configured.

Once we upgrade to supporting v22.x of bitcoind via PR #2135 we can reenable this test. I've manually tested that Eclair throws the expected error when no wallet is loaded using both v0.21.1 and v21.99.

sstone (approved):

PR #2212, opened 2022-03-21
Fix TLV tag number in exception

thomash-acinq:
When the TLV tag for encrypted data was changed from 10 to 4 (https://github.com/lightning/bolts/pull/759/commits/6e99c5feaf60cb797507d181fe583224309318e9#diff-ad65f0beaac5cef88f5fd7a8b9ca36cbc5a790f4815b4b50b9ea794a55aaf012L551) I forgot to change it in the failure case.


remyers (approved):
Looks good to me; and a good excuse for me to read the onion message spec to start to understand what it means.

PR #2217, opened 2022-03-28
Refactor channel into separate traits

t-bast:
This PR is an alternative to #2188 and implements what was suggested in [this comment](https://github.com/ACINQ/eclair/pull/2188#discussion_r834507069). I believe it's a superior alternative, which lets us keep a single actor while separating code into smaller state machines.

It doesn't contain any logic change: reviewers should verify that the code is moved without any changes apart from the visibility of the methods. We will introduce some logic changes to better take advantage of this refactoring in future PRs.

The only added code are the abstract field definitions at the beginning of each trait.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ErrorHandlers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ErrorHandlers.scala
@@ -1,0 +1,377 @@
+/*
+ * Copyright 2022 ACINQ SAS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fr.acinq.eclair.channel
+
+import akka.actor.typed.scaladsl.adapter.actorRefAdapter
+import akka.actor.{ActorRef, FSM}
+import fr.acinq.bitcoin.Crypto.PublicKey
+import fr.acinq.bitcoin.{ByteVector32, OutPoint, SatoshiLong, Transaction}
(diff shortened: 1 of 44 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/ErrorHandlers.scala
```diff
@@ -0,0 +1,377 @@
...
+  def keyManager: ChannelKeyManager
+
+  def txPublisher: akka.actor.typed.ActorRef[TxPublisher.Command]
+  def blockchain: akka.actor.typed.ActorRef[ZmqWatcher.Command]
+  def peer: akka.actor.ActorRef
+  def activeConnection: akka.actor.ActorRef
+  def self: akka.actor.ActorRef
+  // @formatter:on
+
+  implicit def implicitLog: akka.event.DiagnosticLoggingAdapter
```
pm47, line 59: We should probably move those either in a `UtilityHandlers` or in an other adhoc trait.

pm47 (commented):
Phew, `Channel.scala` feels a bit lighter now ^^

What do you think of a new package hierarchy:
- `fr.acinq.eclair.channel.fsm`
  -  `Channel.scala`
- `fr.acinq.eclair.channel.fsm.handlers`
  - `*Handlers.scala`

eclair-core/src/main/scala/fr/acinq/eclair/channel/ErrorHandlers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2217/commits/b18bd74e935dfed75c1692157784ee63aa3eebf4

pm47: I have an alternative commit c9fa7bd2d37e937b883207aea572692537211d17 using the [cake pattern](https://medium.com/rahasak/scala-cake-pattern-e0cd894dae4e). In general I'm not a fan of this pattern, but here the traits are really not meant to be extended by other than `Channel`, and it saves us from duplicating the args altogether.

The move `val` -> `protected val` is optional.

t-bast: Nice, I agree, this is the perfect place for the cake pattern. Done in https://github.com/ACINQ/eclair/pull/2217/commits/7f2c4af930c62e675e04a37f61efb3775dc65128

pm47 (approved):
LGTM!

PR #2219, opened 2022-03-30
Update experimental trampoline feature bit

t-bast:
The `option_scid_alias` feature (https://github.com/lightning/bolts/pull/910) is going to clash with the feature bit we use for trampoline payments. Fortunately we've never advertised it, so we can simply update it to a much higher value for now (until the final version of trampoline is specified).

We don't need to implement backwards-compatibility because no-one is generating trampoline invoices from eclair or paying trampoline invoices from eclair AFAIK. We will need to implement some backwards-compatibility in `lightning-kmp` however to ensure that Phoenix can smoothly transition to the new feature bit.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -57,7 +57,7 @@ eclair {
     option_onion_messages = optional
     option_channel_type = optional
     option_payment_metadata = optional
-    trampoline_payment = disabled
+    trampoline_payment_prototype = disabled
     keysend = disabled
   }
   override-init-features = [ // optional per-node features
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -226,17 +226,20 @@ object Features {
     val mandatory = 48
   }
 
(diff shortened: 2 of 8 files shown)


pm47:
> We don't need to implement backwards-compatibility because no-one is generating trampoline invoices from eclair or paying trampoline invoices from eclair AFAIK.

Phoenix Android does, because it is still based on the eclair-android branch.

t-bast:
> Phoenix Android does, because it is still based on the eclair-android branch.

But we don't plan on updating the `eclair-android` branch to include this commit, do we?
As long as we don't do that, we're safe, right?

eclair-core/src/main/resources/reference.conf
```diff
@@ -57,7 +57,7 @@ eclair {
     option_onion_messages = optional
     option_channel_type = optional
     option_payment_metadata = optional
-    trampoline_payment = disabled
+    trampoline_payment_legacy = disabled
```
pm47, line 60: I think this is clearer:
```suggestion
    trampoline_payment_prototype = disabled
```

t-bast: Agreed, done.

pm47 (approved):

PR #2221, opened 2022-03-31
Use long id instead of short id in `ChannelRelay`

pm47:
This is a preparation for https://github.com/lightning/bolts/pull/910.

The relaying logic is now based on the `channelId` instead of
`shortChannelId`.

We used to consider `shortChannelId` as a unique identifier for the
channel, but it's not true anymore: there will be multiple aliases per
channel.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
@@ -201,56 +201,52 @@ class ChannelRelay private(nodeParams: NodeParams,
   def selectPreferredChannel(alreadyTried: Seq[ByteVector32]): Option[OutgoingChannel] = {
     val requestedShortChannelId = r.payload.outgoingChannelId
     context.log.debug("selecting next channel with requestedShortChannelId={}", requestedShortChannelId)
-    nextNodeId_opt match {
-      case Some(_) =>
-        // we then filter out channels that we have already tried
-        val candidateChannels: Map[ByteVector32, OutgoingChannel] = channels -- alreadyTried
-        // and we filter again to keep the ones that are compatible with this payment (mainly fees, expiry delta)
-        candidateChannels
-          .values
-          .map { channel =>
-            val relayResult = relayOrFail(Some(channel))
-            context.log.debug(s"candidate channel: channelId=${channel.channelId} availableForSend={} capacity={} channelUpdate={} result={}",
-              channel.commitments.availableBalanceForSend,
-              channel.commitments.capacity,
-              channel.channelUpdate,
-              relayResult match {
-                case _: RelaySuccess => "success"
-                case RelayFailure(CMD_FAIL_HTLC(_, Right(failureReason), _, _)) => failureReason
-                case other => other
-              })
-            (channel, relayResult)
-          }
-          .collect {
-            // we only keep channels that have enough balance to handle this payment
-            case (channel, _: RelaySuccess) if channel.commitments.availableBalanceForSend > r.payload.amountToForward => channel
-          }
-          .toList // needed for ordering
-          // we want to use the channel with:
-          //  - the lowest available capacity to ensure we keep high-capacity channels for big payments
-          //  - the lowest available balance to increase our incoming liquidity
-          .sortBy { channel => (channel.commitments.capacity, channel.commitments.availableBalanceForSend) }
-          .headOption match {
-          case Some(channel) =>
-            if (requestedChannelId_opt.contains(channel.channelId)) {
-              context.log.debug("requested short channel id is our preferred channel")
-              Some(channel)
-            } else {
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -53,13 +54,11 @@ object ChannelRelayer {
...
 
-  private type ChannelUpdates = Map[ShortChannelId, Relayer.OutgoingChannel]
-  private type NodeChannels = mutable.MultiDict[PublicKey, ShortChannelId]
-
   def apply(nodeParams: NodeParams,
             register: ActorRef,
-            channelUpdates: ChannelUpdates = Map.empty,
-            node2channels: NodeChannels = mutable.MultiDict.empty[PublicKey, ShortChannelId]): Behavior[Command] =
+            channels: Map[ByteVector32, Relayer.OutgoingChannel] = Map.empty,
+            scid2channels: Map[ShortChannelId, ByteVector32] = Map.empty,
```
t-bast, line 60: nit: maybe call this `resolveScid` to be consistent with the router?

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
```diff
@@ -488,14 +488,12 @@ class ChannelRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("a
 
     // Simulate a chain re-org that changes the shortChannelId:
     channelRelayer ! WrappedShortChannelIdAssigned(ShortChannelIdAssigned(null, channelId_ab, ShortChannelId(42), Some(channelUpdate_ab.shortChannelId)))
-    val channels7 = getOutgoingChannels(true)
-    assert(channels7.isEmpty)
```
t-bast, line 492: Shouldn't we keep that and assert that it's now `nonEmpty` since we changed that behavior?

t-bast (commented):
Looks mostly good to me :+1:

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
pm47: I'd rather rename `resolveScid` to `scid2channels` 😆

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
pm47: Yeah, btw this was removed from #2224, since `ChannelRelayer` isn't listening to `ShortChannelIdAssigned` anymore. But I should have kept a reorg test using the `LocalChannelUpdate`. Will update #2224 directly if that's okay with you.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
t-bast: True!

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
t-bast: Good point, that's a nice benefit that we won't have to listen to `ShortChannelIdAssigned` anymore here!

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
```diff
@@ -190,58 +189,68 @@ class ChannelRelay private(nodeParams: NodeParams,
...
+              case Some(requestedChannel) if alreadyTried.contains(requestedChannel.channelId) =>
+                context.log.debug("no channel seems to work for this payment and we have already tried the requested channel id: giving up")
+                None
+              case _ =>
+                context.log.debug("no channel seems to work for this payment, we will try to use the one requested")
+                requestedChannel_opt
+            }
         }
-      case _ => Some(requestedShortChannelId) // we don't have a channel_update for this short_channel_id
+      case _ => requestedChannelId_opt.flatMap(channels.get) // we don't have a channel_update for this short_channel_id
```
t-bast, line 253: It's a bit confusing to make it look like we could return a result here whereas it will always be `None` (because we reach that case if `channels` is empty, maybe it's worth being more explicit?

```suggestion
      case _ => None // we don't have a channel_update for this short_channel_id
```

pm47: Actually we can remove the whole pattern matching I believe. Done in 848062ff9.

t-bast (approved):

PR #2224, opened 2022-04-04
Add support for zero-conf and scid-alias

pm47:
Builds on #2221 and #2269.

This implements https://github.com/lightning/bolts/pull/910.

Commit-by-commit review strongly advised. [7e28c1a](https://github.com/ACINQ/eclair/pull/2224/commits/7e28c1af17ca5eff976fb5594d1469356cde30f8) is a bit of a mouthful, but the other commits are mostly small.

Commit [e7f7ef0](https://github.com/ACINQ/eclair/pull/2224/commits/e7f7ef057cc7ec4cd91ed5e0bfb39ce7c18167d9) is optional, it skips blockchain validation for local channels, which would save a lot of bitcoind calls at startup for large nodes.


eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
```diff
@@ -350,30 +350,33 @@ private[channel] object ChannelCodecs0 {
...
 
-    val DATA_WAIT_FOR_FUNDING_LOCKED_Codec: Codec[DATA_WAIT_FOR_FUNDING_LOCKED] = (
+    val DATA_WAIT_FOR_CHANNEL_READY_Codec: Codec[DATA_WAIT_FOR_CHANNEL_READY] = (
       ("commitments" | commitmentsCodec) ::
-        ("shortChannelId" | shortchannelid) ::
-        ("lastSent" | fundingLockedCodec)).as[DATA_WAIT_FOR_FUNDING_LOCKED].decodeOnly
+        ("shortChannelId" | realshortchannelid) ::
+        ("lastSent" | channelReadyCodec)).map {
+      case commitments :: shortChannelId :: lastSent :: HNil =>
+        DATA_WAIT_FOR_CHANNEL_READY(commitments, realShortChannelId_opt = Some(shortChannelId), localAlias = shortChannelId.toAlias, lastSent = lastSent)
```
pm47, line 378: Note how for backward compatibility we set both the `realShortChannelId_opt` and `localAlias` to the same value. It may not actually be the real scid on feature branches.

eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
```diff
@@ -56,6 +65,8 @@ object ShortChannelId {
   def outputIndex(shortChannelId: ShortChannelId): Int = (shortChannelId.id & 0xFFFF).toInt
 
   def coordinates(shortChannelId: ShortChannelId): TxCoordinates = TxCoordinates(blockHeight(shortChannelId), txIndex(shortChannelId), outputIndex(shortChannelId))
+
+  def generateLocalAlias(): LocalAlias = new ShortChannelId(System.nanoTime()) with LocalAlias // TODO: fixme (duplicate, etc.)
```
pm47, line 69: ⚠️  needs fixing, this is a dummy impl that leaks information on the creation date of the channel.

Since generating an alias only happens at channel creation, we can afford doing something "costly". For example this could be an `AtomicLong` that we initialize at startup by checking all currently assigned `localAlias`, and then we do an `addAndGet` with a random increment when we need to generate a new alias.

pm47:
Rebased

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -320,13 +320,22 @@ private[channel] object ChannelCodecs3 {
...
-        ("lastSent" | lengthDelimited(fundingLockedCodec))).as[DATA_WAIT_FOR_FUNDING_LOCKED]
+        ("lastSent" | lengthDelimited(channelReadyCodec))).map {
+      case commitments :: shortChannelId :: lastSent :: HNil =>
+        DATA_WAIT_FOR_CHANNEL_READY(commitments, realShortChannelId_opt = Some(shortChannelId.toReal), localAlias = shortChannelId.toAlias, lastSent = lastSent)
+    }.decodeOnly
+
+    val DATA_WAIT_FOR_CHANNEL_READY_Codec: Codec[DATA_WAIT_FOR_CHANNEL_READY] = (
+      ("commitments" | commitmentsCodec) ::
+        ("realShortChannelId_opt" | optional(bool8, realshortchannelid)) ::
+        ("localAlias" | localalias) ::
```
t-bast, line 337: Have you tried putting `localAlias` and `remoteAlias_opt` in `LocalParams` and `RemoteParams` instead? It feels like it would belong there, and since the codecs are already parameterized on `channelFeatures` maybe backwards-compatibility is easy to handle?

I think it would minimize the changes to `ChannelEvents` as well by having these fields accessible in the `Commitments` object.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -387,14 +414,15 @@ private[channel] object ChannelCodecs3 {
...
     .typecase(0x08, Codecs.DATA_SHUTDOWN_Codec)
-    .typecase(0x07, Codecs.DATA_NORMAL_Codec)
+    .typecase(0x07, Codecs.DATA_NORMAL_COMPAT_07_Codec)
     .typecase(0x06, Codecs.DATA_WAIT_FOR_REMOTE_PUBLISH_FUTURE_COMMITMENT_Codec)
     .typecase(0x05, Codecs.DATA_CLOSING_Codec)
     .typecase(0x04, Codecs.DATA_NEGOTIATING_Codec)
     .typecase(0x03, Codecs.DATA_SHUTDOWN_COMPAT_03_Codec)
     .typecase(0x02, Codecs.DATA_NORMAL_COMPAT_02_Codec)
-    .typecase(0x01, Codecs.DATA_WAIT_FOR_FUNDING_LOCKED_Codec)
+    .typecase(0x01, Codecs.DATA_WAIT_FOR_CHANNEL_READY_Codec)
```
t-bast, line 425: You forgot to add DATA_WAIT_FOR_CHANNEL_READY_COMPAT_01_Codec, we probably need a migration test on this!

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/ChannelTlv.scala
```diff
@@ -107,10 +108,17 @@ object FundingSignedTlv {
...
-sealed trait FundingLockedTlv extends Tlv
+sealed trait ChannelReadyTlv extends Tlv
 
-object FundingLockedTlv {
-  val fundingLockedTlvCodec: Codec[TlvStream[FundingLockedTlv]] = tlvStream(discriminated[FundingLockedTlv].by(varint))
+object ChannelReadyTlv {
+
+  case class ShortChannelIdTlv(alias: ShortChannelId) extends ChannelReadyTlv
+
+  val channelTypeCodec: Codec[ShortChannelIdTlv] = variableSizeBytesLong(varintoverflow, "alias" | shortchannelid).as[ShortChannelIdTlv]
```
t-bast, line 117: The name should be changed:

```suggestion
  val channelAliasCodec: Codec[ShortChannelIdTlv] = variableSizeBytesLong(varintoverflow, "alias" | shortchannelid).as[ShortChannelIdTlv]
```

Also, it would be nice to have tests that verify that we reject the `channel_type`s that use `zero_conf` or `scid_alias` without `anchor_zero_fees`.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/ChannelTlv.scala
```diff
@@ -107,10 +108,17 @@ object FundingSignedTlv {
...
 }
 
-sealed trait FundingLockedTlv extends Tlv
+sealed trait ChannelReadyTlv extends Tlv
 
-object FundingLockedTlv {
-  val fundingLockedTlvCodec: Codec[TlvStream[FundingLockedTlv]] = tlvStream(discriminated[FundingLockedTlv].by(varint))
+object ChannelReadyTlv {
+
+  case class ShortChannelIdTlv(alias: ShortChannelId) extends ChannelReadyTlv
```
t-bast, line 115: Shouldn't this be a `LocalAlias` instead of a `ShortChannelId`? Actually this raises a more general question: why did you define a `LocalAlias` instead of a more general `ShortChannelIdAlias` used for both local and remote? What they send us in here should be an alias, right?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
```diff
@@ -96,10 +97,10 @@ class WaitForAcceptChannelStateSpec extends TestKitBaseClass with FixtureAnyFunS
   test("recv AcceptChannel (anchor outputs zero fee htlc txs)", Tag(ChannelStateTestsTags.AnchorOutputsZeroFeeHtlcTxs)) { f =>
     import f._
     val accept = bob2alice.expectMsgType[AcceptChannel]
-    assert(accept.channelType_opt === Some(ChannelTypes.AnchorOutputsZeroFeeHtlcTx))
+    assert(accept.channelType_opt === Some(ChannelTypes.AnchorOutputsZeroFeeHtlcTx(scidAlias = false, zeroConf = false)))
```
t-bast, line 100: It would be useful to also have a test that verifies that if both peers have activated `scid_alias`, it becomes automatically used (and one in `WaitForOpenChannelStateSpec` as well).

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecsSpec.scala
```diff
@@ -221,7 +223,7 @@ class LightningMessageCodecsSpec extends AnyFunSuite {
       defaultEncoded ++ hex"0000" ++ hex"01021000" -> defaultOpen.copy(tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScriptTlv(ByteVector.empty), ChannelTlv.ChannelTypeTlv(ChannelTypes.StaticRemoteKey))),
       // non-empty upfront_shutdown_script + channel type
       defaultEncoded ++ hex"0004 01abcdef" ++ hex"0103101000" -> defaultOpen.copy(tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScriptTlv(hex"01abcdef"), ChannelTlv.ChannelTypeTlv(ChannelTypes.AnchorOutputs))),
-      defaultEncoded ++ hex"0002 abcd" ++ hex"0103401000" -> defaultOpen.copy(tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScriptTlv(hex"abcd"), ChannelTlv.ChannelTypeTlv(ChannelTypes.AnchorOutputsZeroFeeHtlcTx))),
+      defaultEncoded ++ hex"0002 abcd" ++ hex"0103401000" -> defaultOpen.copy(tlvStream = TlvStream(ChannelTlv.UpfrontShutdownScriptTlv(hex"abcd"), ChannelTlv.ChannelTypeTlv(ChannelTypes.AnchorOutputsZeroFeeHtlcTx(scidAlias = false, zeroConf = false)))),
```
t-bast, line 226: It would be useful to have a test for the channel types that also set `scid_alias` and/or `zero_conf`

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecsSpec.scala
```diff
@@ -345,7 +347,7 @@ class LightningMessageCodecsSpec extends AnyFunSuite {
     val accept = AcceptChannel(randomBytes32(), 3 sat, UInt64(4), 5 sat, 6 msat, 7, CltvExpiryDelta(8), 9, publicKey(1), point(2), point(3), point(4), point(5), point(6))
     val funding_created = FundingCreated(randomBytes32(), bin32(0), 3, randomBytes64())
     val funding_signed = FundingSigned(randomBytes32(), randomBytes64())
-    val funding_locked = FundingLocked(randomBytes32(), point(2))
+    val channel_ready = ChannelReady(randomBytes32(), point(2))
```
t-bast, line 350: Can you add one where `channel_ready` contains an `alias`?

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/Channel.scala
```diff
@@ -33,13 +33,13 @@ trait Channel {
...
+    formFields(nodeIdFormParam, "fundingSatoshis".as[Satoshi], "pushMsat".as[MilliSatoshi].?, "channelType".?, "fundingFeerateSatByte".as[FeeratePerByte].?, "announceChannel".as[Boolean].?, "zeroConf".as[Boolean]?false, "openTimeoutSeconds".as[Timeout].?) {
+      (nodeId, fundingSatoshis, pushMsat, channelType, fundingFeerateSatByte, announceChannel_opt, zeroConf, openTimeout_opt) =>
         val (channelTypeOk, channelType_opt) = channelType match {
           case Some(str) if str == ChannelTypes.Standard.toString => (true, Some(ChannelTypes.Standard))
           case Some(str) if str == ChannelTypes.StaticRemoteKey.toString => (true, Some(ChannelTypes.StaticRemoteKey))
           case Some(str) if str == ChannelTypes.AnchorOutputs.toString => (true, Some(ChannelTypes.AnchorOutputs))
-          case Some(str) if str == ChannelTypes.AnchorOutputsZeroFeeHtlcTx.toString => (true, Some(ChannelTypes.AnchorOutputsZeroFeeHtlcTx))
+          case Some(str) if str == "anchor_outputs_zero_fee_htlc_tx" => (true, Some(ChannelTypes.AnchorOutputsZeroFeeHtlcTx(scidAlias = announceChannel_opt.contains(true), zeroConf = zeroConf))) // alias feature is incompatible with public channel
```
t-bast, line 42: The value of `scidAlias` is incorrect here, it should be `announceChannel_opt.contains(true)` (actually it isn't even that simple, it depends on the user's `eclair.conf` when `announceChannel_opt` is `None`...)

eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -372,9 +372,9 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
       Route.seal(mockService.route) ~>
       check {
         assert(handled)
-        assert(status == OK)
+        //assert(status == OK)
```
t-bast, line 375: nit: revert?

t-bast (commented):
This is a first pass on some details and high-level architecture, I'll dive into the actual usage of `scid_alias` in a second pass later, my brain wasn't able to handle it today :)

eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
pm47: We cannot afford a duplicate here though and the space isn't that big. We should expect a collision for 5 billion attempts on 64 bits and only 80k attempts for 32 bits right? https://en.wikipedia.org/wiki/Birthday_attack

t-bast: It's true, the space isn't that big, but I'm afraid there may be privacy issues with generating those incrementally...I'll check what other implementations do

t-bast: I did some maths to confirm this, let me know if my calculations look correct:

- scids are 64 bits long
- we want to reserve the range [0; 2^22] for real scids
- that leaves us with 2^42 available values for aliases 
- we'll consider that we want to have 250k channels using aliases (~2^18)
- the probability of a collision is then p(n) ~= 2^36 / 2^43 ~= 0.8%

This is indeed not great! But using a counter that we increment leaks information about the approximate creation time of the channel, which could let an attacker find the channel outpoint on-chain...so I believe that what we should do instead is:

- generate an alias randomly in the [2^22:2^64] range
- check our alias map to see if it's already assigned and re-roll if it is

It means we'll probably need to send a message to the router to obtain a new alias, the router is where we can guarantee uniqueness.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
pm47: Dammit! Nice catch. I would probably have caught it when deploying on endurance, the ultimate migration test.

eclair-core/src/main/resources/reference.conf
```diff
@@ -57,9 +57,11 @@ eclair {
     option_dual_fund = disabled
     option_onion_messages = optional
     option_channel_type = optional
+    option_scid_alias = optional
     option_payment_metadata = optional
-    trampoline_payment_prototype = disabled
+    option_zeroconf = disabled
```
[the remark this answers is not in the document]
pm47: Done in d0a2c97c8.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/ZmqWatcher.scala
```diff
@@ -406,13 +412,21 @@ private class ZmqWatcher(nodeParams: NodeParams, blockHeight: AtomicLong, client
...
         client.getTransaction(w.txId).flatMap { tx =>
-          client.getTransactionShortId(w.txId).map {
-            case (height, index) => w match {
-              case w: WatchFundingConfirmed => context.self ! TriggerEvent(w.replyTo, w, WatchFundingConfirmedTriggered(height, index, tx))
-              case w: WatchFundingDeeplyBuried => context.self ! TriggerEvent(w.replyTo, w, WatchFundingDeeplyBuriedTriggered(height, index, tx))
-              case w: WatchTxConfirmed => context.self ! TriggerEvent(w.replyTo, w, WatchTxConfirmedTriggered(height, index, tx))
-              case w: WatchParentTxConfirmed => context.self ! TriggerEvent(w.replyTo, w, WatchParentTxConfirmedTriggered(height, index, tx))
-            }
+          w match {
+            case w: WatchFundingConfirmed if confirmations == 0 && w.minDepth == 0 =>
```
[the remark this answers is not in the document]
pm47: Done in d0a2c97c8.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenSingleFunder.scala
```diff
@@ -151,6 +151,7 @@ trait ChannelOpenSingleFunder extends FundingHandlers with ErrorHandlers {
             initFeatures = remoteInit.features,
             shutdownScript = remoteShutdownScript)
           log.debug("remote params: {}", remoteParams)
+          log.info("remote indicates they are going to use fundingMinDepth={}", accept.minimumDepth)
```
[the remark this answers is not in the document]
pm47: Done in d0a2c97c8.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/ChannelTlv.scala
pm47: Renamed in d0a2c97c8, deferring adding test until I convince myself the channel type way is the best.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
pm47: Fixed in 897f5f1b9.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/a/WaitForAcceptChannelStateSpec.scala
pm47: Done in d05f61108.

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecsSpec.scala
pm47: Done in 007583cc7.

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecsSpec.scala
pm47: Done in fc7d4ee88, but this test needs serious refactoring:
- test cases should be defined in a collection, not one-by-one
- should do non-reg with expected binary data instead of just roundtrip

pm47:
Rebased.

t-bast (commented):
I've finished reviewing the code up to https://github.com/ACINQ/eclair/pull/2224/commits/957f014d5dd101b03618a55da3cd792d61c058e9, but I haven't reviewed the tests yet.

Submitting this PR round and waiting for the changes to the scid types and #2310 and I'll move on to review the tests.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -392,9 +399,15 @@ object Router {
     /** Create an invoice routing hint from that channel. Note that if the channel is private, the invoice will leak its existence. */
     def toIncomingExtraHop: Option[ExtraHop] = {
       // we want the incoming channel_update
-      val remoteUpdate_opt = if (localNodeId == nodeId1) update_2_opt else update_1_opt
-      remoteUpdate_opt.map { remoteUpdate =>
-          ExtraHop(remoteNodeId, remoteUpdate.shortChannelId, remoteUpdate.feeBaseMsat, remoteUpdate.feeProportionalMillionths, remoteUpdate.cltvExpiryDelta)
+      val (localUpdate_opt, remoteUpdate_opt) = if (localNodeId == nodeId1) (update_1_opt, update_2_opt) else (update_2_opt, update_1_opt)
+      (localUpdate_opt, remoteUpdate_opt) match {
+        case (Some(localUpdate), Some(remoteUpdate)) =>
+          // this is tricky: for incoming payments we need the *remote alias*, we can find it in the channel_update that we sent them
```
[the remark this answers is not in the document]
pm47: If you liked it, you are going to love [c20f672](https://github.com/ACINQ/eclair/pull/2224/commits/c20f672be053214637dbaf4fa3cea259b17d87ce)!

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -418,66 +457,92 @@ object Validation {
...
             // maybe the local channel was pruned (can happen if we were disconnected for more than 2 weeks)
-            db.removeFromPruned(c.shortChannelId)
-            handleChannelUpdate(d1, db, routerConf, Left(lcu))
-          case None if d.privateChannels.contains(lcu.channelId) =>
-            // channel isn't announced but we already know about it, we can process the channel_update
-            handleChannelUpdate(d, db, routerConf, Left(lcu))
+            db.removeFromPruned(ann.shortChannelId)
+            log.debug("processing channel_update")
+            val d2 = handleChannelUpdate(d1, db, nodeParams.routerConf, Left(lcu))
+            d2
```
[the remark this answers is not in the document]
pm47: Done in [c20f672](https://github.com/ACINQ/eclair/pull/2224/commits/c20f672be053214637dbaf4fa3cea259b17d87ce).

t-bast (commented):
LGTM, I have some test improvements / fixes in #2314 and we should be good to go. I'll do some end-to-end tests on regtest with other implementations.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -290,7 +327,7 @@ object Validation {
           log.debug("ignoring {} (stale)", u)
           sendDecision(origins, GossipDecision.Stale(u))
           d
-        } else if (pc.getChannelUpdateSameSideAs(u).exists(_.timestamp >= u.timestamp)) {
+        } else if (pc.getChannelUpdateSameSideAs(u).exists(previous => previous.timestamp >= u.timestamp && previous.shortChannelId == u.shortChannelId)) { // NB: we also check the id because there could be a switch alias->real scid
```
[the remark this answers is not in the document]
t-bast: Actually not really, this isn't where a sleep is useful, it's because in the `onTransition` in `Channel.scala` we don't send the channel update if the timestamp is the same, so we need some kind of sleep. I improved it to use an `eventually` instead of `Thread.sleep`.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/ZeroConfActivationSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+    assert(!bob.nodeParams.features.activated.contains(ZeroConf))
+
+    connect(alice, bob)
+    val channelId = openChannel(alice, bob, 100_000 sat).channelId
+
+    assert(!getChannelData(alice, channelId).asInstanceOf[PersistentChannelData].commitments.channelFeatures.hasFeature(ZeroConf))
+    assert(!getChannelData(bob, channelId).asInstanceOf[PersistentChannelData].commitments.channelFeatures.hasFeature(ZeroConf))
+  }
+
+  test("open a channel alice-bob (zero-conf disabled on both sides, requested via channel type by alice)") { f =>
```
t-bast, line 52: In a follow-up PR, we should introduce a whitelist of nodes from which we accept 0-conf channels. If nodes are in that whitelist, we should accept the `channel_type` even though they didn't set the feature in their `init`.

We should maybe even go further than that and use 0-conf with those peers even if the `channel_type` doesn't include it. Either they'll accept it and send their `channel_ready`, or they won't and it's ok, we'll just wait for their `channel_ready` automatically.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/ZeroConfActivationSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+    assert(bob.nodeParams.features.activated.contains(ZeroConf))
+
+    connect(alice, bob)
+    val channelType = AnchorOutputsZeroFeeHtlcTx(scidAlias = false, zeroConf = true)
+    val channelId = openChannel(alice, bob, 100_000 sat, channelType_opt = Some(channelType)).channelId
+
+    assert(getChannelData(alice, channelId).asInstanceOf[PersistentChannelData].commitments.channelFeatures.hasFeature(ZeroConf))
+    assert(getChannelData(bob, channelId).asInstanceOf[PersistentChannelData].commitments.channelFeatures.hasFeature(ZeroConf))
+  }
+
```
t-bast, line 79: nit: why the double empty line here and below?

docs/release-notes/eclair-vnext.md
```diff
@@ -4,7 +4,64 @@
...
+]
+```
+
+Note that, as funder, Eclair will happily use an unconfirmed channel if the peer sends an early `channel_ready`, even if the `option_zeroconf` feature isn't enabled, as long as the peer provides a channel alias.
+
+##### Enabling through channel type
+
+You can enable `option_scid_alias` and `option_zeroconf` features by requesting them in the channel type, even if those options aren't enabled in your features.
+
+Below is how you would request a zeroconf channel thhrough the command-line interface:
```
t-bast, line 57: nit:
```suggestion
Below is how you would request a public zeroconf channel through the command-line interface:
```

docs/release-notes/eclair-vnext.md
```diff
@@ -4,7 +4,64 @@
...
+```shell
+$ ./eclair-cli open --nodeId=03864e... --fundingSatoshis=100000 --channelType=anchor_outputs_zero_fee_htlc_tx+zeroconf
```
t-bast, line 59: Can we also add one example for a private `scid_alias + zeroconf` channel?

```sh
$ ./eclair-cli open --nodeId=03864e... --fundingSatoshis=100000 --channelType=anchor_outputs_zero_fee_htlc_tx+scid_alias+zeroconf --announceChannel=false
```

t-bast (commented):
LGTM, we're almost ready to go!

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
```diff
@@ -90,35 +91,33 @@ object ChannelRelayer {
...
             val node2channels1 = node2channels.addOne(remoteNodeId, channelId)
             apply(nodeParams, register, channels1, scid2channels1, node2channels1)
 
-          case WrappedLocalChannelDown(LocalChannelDown(_, channelId, shortChannelId, remoteNodeId)) =>
-            context.log.debug(s"removed local channel info for channelId=$channelId shortChannelId=$shortChannelId")
+          case WrappedLocalChannelDown(LocalChannelDown(_, channelId, shortIds, remoteNodeId)) =>
+            context.log.debug(s"removed local channel info for channelId=$channelId localAlias=${shortIds.localAlias}")
             val channels1 = channels - channelId
-            val scid2Channels1 = scid2channels - shortChannelId
+            val scid2Channels1 = scid2channels - shortIds.localAlias
```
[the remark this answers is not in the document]
pm47: Good catch. I've done a minimalist cleanup in a1ebd760e, but it's not possible to test as-is (because we do clean-up the `channels` map, which will cause relays to fail).

I started to add a `GetData` method to `ChannelRelayer`, that would return the whole state, but it's not trivial because one of the maps is mutable, making it unsafe to share with the outside. So I rolled back, LMKWYT.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/ZeroConfActivationSpec.scala
pm47: No good reason, fixed in ac60f87ee.

docs/release-notes/eclair-vnext.md
pm47: Fixed in ac60f87ee.

docs/release-notes/eclair-vnext.md
pm47: Done in ac60f87ee.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelayer.scala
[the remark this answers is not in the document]
t-bast: That's good enough :+1:

t-bast (approved):
LGTM :rocket: :zap:

t-bast (approved):

pm47:
:love_you_gesture:

pm47:
Follow-ups to this PR include:
- better way of generating aliases (see https://github.com/ACINQ/eclair/pull/2224#discussion_r875068884)
- whitelist of zeroconf peers:
   - prevent users from accidentally enabling zeroconf for all peers
   - set `min_depth=0` and send early `channel_ready` even if `option_zeroconf` bit isn't set in the channel type

PR #2226, opened 2022-04-06
Check that channel seed has not been modified

sstone:
On startup, we check that for each of our channels the public key derived from the channel seed matches the key that was
used when the channel was funded.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/DBChecker.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/DBChecker.scala
@@ -16,29 +16,36 @@
 
 package fr.acinq.eclair
 
+import fr.acinq.eclair.channel.Commitments
 import grizzled.slf4j.Logging
 
 import scala.util.{Failure, Success, Try}
 
-object DBCompatChecker extends Logging {
+object DBChecker extends Logging {
 
   /**
-    * Tests if the channels data in the DB are compatible with the current version of eclair; throws an exception if incompatible.
-    *
-    * @param nodeParams
-    */
-  def checkDBCompatibility(nodeParams: NodeParams): Unit =
+   * Tests if the channels data in the DB are compatible with the current version of eclair; throws an exception if incompatible.
+   *
+   * @param nodeParams
+   */
+  def checkChannelsDB(nodeParams: NodeParams): Unit = {
+    logger.info("listLocalChannels()")
     Try(nodeParams.db.channels.listLocalChannels()) match {
-      case Success(_) => {}
+      case Success(channels) =>
+        channels.foreach(hasCommitments => Try(Commitments.validateSeed(hasCommitments.commitments, nodeParams.channelKeyManager)) match {
+          case Success(_) => ()
+          case Failure(e) => throw InvalidChannelSeedException(e)
+        })
       case Failure(_) => throw IncompatibleDBException
     }
+  }
 
   /**
-    * Tests if the network database is readable.
-    *
-    * @param nodeParams
-    */
-  def checkNetworkDBCompatibility(nodeParams: NodeParams): Unit =
+   * Tests if the network database is readable.
+   *
+   * @param nodeParams
+   */
+  def checkNetworkDB(nodeParams: NodeParams): Unit =
     Try(nodeParams.db.network.listChannels(), nodeParams.db.network.listNodes()) match {
       case Success(_) => {}
       case Failure(_) => throw IncompatibleNetworkDBException
--- a/eclair-core/src/main/scala/fr/acinq/eclair/DBChecker.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/DBChecker.scala
@@ -24,18 +24,16 @@ import scala.util.{Failure, Success, Try}
 object DBChecker extends Logging {
 
   /**
-   * Tests if the channels data in the DB are compatible with the current version of eclair; throws an exception if incompatible.
+   * Tests if the channels data in the DB is valid (and throws an exception if not):
+   * - it is compatible with the current version of eclair
+   * - channel keys can be re-generated from the channel seed
    *
-   * @param nodeParams
+   * @param nodeParams node parameters
    */
   def checkChannelsDB(nodeParams: NodeParams): Unit = {
-    logger.info("listLocalChannels()")
     Try(nodeParams.db.channels.listLocalChannels()) match {
       case Success(channels) =>
-        channels.foreach(hasCommitments => Try(Commitments.validateSeed(hasCommitments.commitments, nodeParams.channelKeyManager)) match {
-          case Success(_) => ()
-          case Failure(e) => throw InvalidChannelSeedException(e)
-        })
+        channels.foreach(data => if (!Commitments.validateSeed(data.commitments, nodeParams.channelKeyManager)) throw InvalidChannelSeedException(data.channelId))
       case Failure(_) => throw IncompatibleDBException
     }
   }
@@ -47,7 +45,7 @@ object DBChecker extends Logging {
    */
   def checkNetworkDB(nodeParams: NodeParams): Unit =
     Try(nodeParams.db.network.listChannels(), nodeParams.db.network.listNodes()) match {
-      case Success(_) => {}
+      case Success(_) => ()
       case Failure(_) => throw IncompatibleNetworkDBException
     }
 }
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -134,6 +134,13 @@ class Setup(val datadir: File,
   }
 
   val nodeParams = NodeParams.makeNodeParams(config, instanceId, nodeKeyManager, channelKeyManager, initTor(), databases, blockHeight, feeEstimator, pluginParams)
+
+  // validate that channel seed is the same that was used when local channels were created
+  nodeParams.db.channels.listLocalChannels().foreach(hasCommitments => Try(Commitments.validateSeed(hasCommitments.commitments, channelKeyManager)) match {
+    case Success(_) => ()
+    case Failure(e) => throw InvalidChannelSeedException(e)
+  })
```
pm47, line 142: This should be done here imo: https://github.com/ACINQ/eclair/blob/36ed788d392e53a95f707c33dd0930871c72936a/eclair-core/src/main/scala/fr/acinq/eclair/io/Switchboard.scala#L46-L60

Main reasons:
- single place where we load and check channels
- saves one (expensive) call to `listLocalChannels()`
- referencing `Commitments` from `Setup` feels weird.

pm47 (commented):
Concept ack.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
sstone: I chose to add the check in `Setup` to make sure that it's the very first thing we do with channel data. If the current channel seed does not match a channel's public key it means that something is very wrong and we should exit immediately. 

Once setup is over I think that we use `listLocalChannels()` in different places, in the following order:

1. DBCompatChecker
2. PostRestartHtlcCleaner
3. Switchboard

For this to be safe we have to make sure that everything we do before `Switchboard` starts is safe even if the channel seed has been modified which feels hard to enforce ?

sstone: Or we could add the seed check to `DBCompatChecker` ?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -294,6 +294,13 @@ case class Commitments(channelId: ByteVector32,
 
 object Commitments {
 
+  def validateSeed(commitments: Commitments, keyManager: ChannelKeyManager): Unit = {
+    val localFundingKey = keyManager.fundingPublicKey(commitments.localParams.fundingKeyPath).publicKey
+    val remoteFundingKey = commitments.remoteParams.fundingPubKey
+    val fundingScript = Script.write(Scripts.multiSig2of2(localFundingKey, remoteFundingKey))
+    require(commitments.commitInput.redeemScript == fundingScript, s"public key does not match channel funding transaction for channel ${commitments.channelId}")
```
t-bast, line 301: I'm not a fan of throwing here, we have mostly pure functions in this file, can't you return a `Boolean` and let the caller throw / stop eclair?

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
t-bast: > Or we could add the seed check to DBCompatChecker ?

That would be a good choice, I agree with you that it's good to do this at the beginning of `Setup`, before we spawn actors.
We should probably rename `DbCompatChecker` then to test both DB + seed, something like `StartupCompatChecker`?

pm47: >> Or we could add the seed check to DBCompatChecker ?

> That would be a good choice

+1

> We should probably rename DbCompatChecker then to test both DB + seed, something like StartupCompatChecker?

Or `DbChecker`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
sstone: We could but I did not want to give the caller a choice and possibly ignore the return value i.e. it really is an "exception" not an "error" (I even considered calling `sys.exit(1)` here)

t-bast: It's true that it's a fatal error...then do you mind putting that code inside the checker closer to startup instead of inside the `Commitments` object? It shouldn't be used for normal commitments handling

eclair-core/src/main/scala/fr/acinq/eclair/DBChecker.scala
```diff
@@ -16,29 +16,36 @@
...
-    * @param nodeParams
-    */
-  def checkNetworkDBCompatibility(nodeParams: NodeParams): Unit =
+   * Tests if the network database is readable.
+   *
+   * @param nodeParams
+   */
+  def checkNetworkDB(nodeParams: NodeParams): Unit =
     Try(nodeParams.db.network.listChannels(), nodeParams.db.network.listNodes()) match {
       case Success(_) => {}
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/eclair/pull/2226/commits/3ad382a40ebe86dbecaa5edf19aab03eb81e96ee

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -417,4 +417,6 @@ case object EmptyAPIPasswordException extends RuntimeException("must set a passw
 
 case object IncompatibleDBException extends RuntimeException("database is not compatible with this version of eclair")
 
-case object IncompatibleNetworkDBException extends RuntimeException("network database is not compatible with this version of eclair")
\ No newline at end of file
+case object IncompatibleNetworkDBException extends RuntimeException("network database is not compatible with this version of eclair")
+
+case class InvalidChannelSeedException(e: Throwable) extends RuntimeException("Channel seed has been modified", e)
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/eclair/pull/2226/commits/3ad382a40ebe86dbecaa5edf19aab03eb81e96ee

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
sstone: since you both agree :) done in https://github.com/ACINQ/eclair/pull/2226/commits/3ad382a40ebe86dbecaa5edf19aab03eb81e96ee (actually I find just returning a boolean more acceptable now that the check has been moved to `DBChecker`

pm47 (approved):

eclair-core/src/main/scala/fr/acinq/eclair/DBChecker.scala
```diff
@@ -16,31 +16,35 @@
...
-    *
-    * @param nodeParams
-    */
-  def checkDBCompatibility(nodeParams: NodeParams): Unit =
+   * Tests if the channels data in the DB are compatible with the current version of eclair; throws an exception if incompatible.
+   *
+   * @param nodeParams
+   */
+  def checkChannelsDB(nodeParams: NodeParams): Unit = {
+    logger.info("listLocalChannels()")
```
t-bast, line 32: nit: why that log line?

eclair-core/src/main/scala/fr/acinq/eclair/DBChecker.scala
```diff
@@ -16,31 +16,35 @@
...
     Try(nodeParams.db.channels.listLocalChannels()) match {
-      case Success(_) => {}
+      case Success(channels) =>
+        channels.foreach(hasCommitments => if (!Commitments.validateSeed(hasCommitments.commitments, nodeParams.channelKeyManager)) throw InvalidChannelSeedException(hasCommitments.channelId))
```
t-bast, line 35: nit: we've gotten rid of the `HasCommitments` trait:
```suggestion
        channels.foreach(data => if (!Commitments.validateSeed(data.commitments, nodeParams.channelKeyManager)) throw InvalidChannelSeedException(data.channelId))
```

t-bast (commented):
LGTM, only a couple of nits

eclair-core/src/main/scala/fr/acinq/eclair/DBChecker.scala
sstone: removed in https://github.com/ACINQ/eclair/commit/9444d2bd3993cf46ab54b6b4e77861acc088f522

t-bast (approved):

PR #2228, opened 2022-04-07
Refactor channel transitions

t-bast:
This PR contains changes from #2225 without the channel data type hierarchy changes.
It doesn't introduce a lot of functional changes, but they are quite sensitive: reviewers beware!
This is better reviewed commit-by-commit.


eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1629,43 +1596,44 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
       // if channel is private, we send the channel_update directly to remote
       // they need it "to learn the other end's forwarding parameters" (BOLT 7)
       (state, nextState, stateData, nextStateData) match {
-        case (_, _, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && !d1.buried && d2.buried =>
+        case (NORMAL | SYNCING, NORMAL, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && !d1.buried && d2.buried =>
```
pm47, line 1599: 🔍 🕵️‍♂️ 

We only update the `buried` boolean in `NORMAL` state:

```suggestion
        case (NORMAL, NORMAL, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && !d1.buried && d2.buried =>
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1629,43 +1596,44 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
...
           // for a private channel, when the tx was just buried we need to send the channel_update to our peer (even if it didn't change)
           send(d2.channelUpdate)
         case (SYNCING, NORMAL, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && d2.buried =>
           // otherwise if we're coming back online, we rebroadcast the latest channel_update
           // this makes sure that if the channel_update was missed, we have a chance to re-send it
           send(d2.channelUpdate)
-        case (_, _, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && d1.channelUpdate != d2.channelUpdate && d2.buried =>
+        case (NORMAL | SYNCING, NORMAL, d1: DATA_NORMAL, d2: DATA_NORMAL) if !d1.commitments.announceChannel && d1.channelUpdate != d2.channelUpdate && d2.buried =>
```
pm47, line 1606: If you go to that level of granularity here and make the `onTransition` smarter, then I _think_ we can make the [handler in `NORMAL` state](https://github.com/ACINQ/eclair/blob/rename-has-commitments/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala#L702-L712) dumber so the state of the universe is preserved. We were currently _not_ generating a new `channelUpdate` on purpose so that the `d1.channelUpdate != d2.channelUpdate` condition would not be met in the `onTransition`.

⚠️  🐇 🕳️

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenSingleFunder.scala
```diff
@@ -49,6 +49,30 @@ trait ChannelOpenSingleFunder extends FundingHandlers with ErrorHandlers {
 
   this: Channel =>
 
+  /*
+                                              NEW
```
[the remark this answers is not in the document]
t-bast: :eagle: :eye:, done in https://github.com/ACINQ/eclair/pull/2228/commits/eaa488b59a308307080e71689782d19899c21567

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1525,30 +1495,26 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
 
   when(ERR_INFORMATION_LEAK)(errorStateHandler)
 
-  when(ERR_FUNDING_LOST)(errorStateHandler)
```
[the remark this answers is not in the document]
t-bast: You're right, I missed it! Done in https://github.com/ACINQ/eclair/pull/2228/commits/eaa488b59a308307080e71689782d19899c21567

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
pm47: Wait, something is fishy. Indeed we never update the `channelUpdate` in `SYNCING`, and we don't always update it to `disabled` when going from `NORMAL`->`OFFLINE` (see [below](https://github.com/ACINQ/eclair/pull/2228#discussion_r845118794)). How can we be sure that `d2.channelUpdate` is `enabled` in the [middle handler](https://github.com/ACINQ/eclair/blob/eaa488b59a308307080e71689782d19899c21567/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala#L1602)? The comment seems inaccurate.

t-bast: You're right, there is a potential issue here (unrelated to the changes in this PR). If the channel has been disabled without changing the `channel_update`, when we come back online we will send the update with the `disabled` bit set.

I don't think it's a huge issue though, because this is a private channel: our peer only uses the fees/cltv to create invoices, they most likely ignore the `disabled` bit. But we could also check it here and avoid sending anything if the update is still disabled (so we will only send a channel update in the next `NORMAL -> NORMAL` transition).

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
t-bast: I'm not sure there is anything to change/simplify here. What do you mean by "he state of the universe is preserved."? We want to keep our protection against gossip spam, and I really want to avoid putting complex business logic in the transitions, I prefer having it in the handlers like we do today.

pm47: > I'm not sure there is anything to change/simplify here. 

Okay, let's not waste too much time on this since I will revisit in #2224.

> What do you mean by "the state of the universe is preserved."?

Nothing, just a silly joke.

pm47 (approved):

PR #2233, opened 2022-04-12
Add a "stop" API method

sstone:
This API call was added for certain uses cases where killing the process was impractical but internally it just calls `sys.exit()`.
Eclair is designed to shutdown cleanly when its process is killed and this is still the recommended way of stopping it.


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -559,4 +561,12 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
       case Attempt.Failure(cause) => Future.successful(SendOnionMessageResponse(sent = false, failureMessage = Some(s"the `content` field is invalid, it must contain encoded tlvs: ${cause.message}"), response = None))
     }
   }
+
+  override def stop(exitCode: Int): Future[Unit] = {
```
pm47, line 565: I wouldn't have provided the `exitCode` as parameter. We should use `0` as this is a graceful stop.

sstone: done in https://github.com/ACINQ/eclair/pull/2233/commits/974a73ce24972f6caac7d9dc31e3d2a4b0da24af

t-bast (dismissed):

t-bast (approved):

PR #2234, opened 2022-04-12
Support DNS hostnames in node announcements

remyers:
Add support for DNS host names

Notable changes:

- you can now specify a DNS host name as one of your `server.public-ips` addresses (see PR [#911](https://github.com/lightning/bolts/pull/911)). Note: you can not specify more than one DNS host name.
- An IPv4, IPv6 or DNS hostname with port of 0 is invalid as per [Bolt #7](https://github.com/lightning/bolts/blob/master/07-routing-gossip.md).
- DNS host names will not be resolved until `Client` attempts to connect to a peer.
- will not rebroadcast a `NodeAnnouncement` that contains more than one DNS host name
- the codec does _not_ implement punycode parsing; we should create an issue if we want to support non-ASCII character sets for DNS host names
- DNS host name addresses must be resolved to determine if they are IPv4 or IPv6 when deciding whether or not to use a  proxy.

To do:

- [X] [interop testing with c-lightning](https://github.com/remyers/lightning-cfg/blob/dnsgossip/scripts/dns_gossip.sh) PR# [4829](https://github.com/ElementsProject/lightning/pull/4829)
- [x] ~~interop testing with LND Issue# [6337](https://github.com/lightningnetwork/lnd/issues/6337)~~
- [x] ~~interop testing with LDK PR# [1329](https://github.com/lightningdevkit/rust-lightning/pull/1329)~~

See also [PR #2202](https://github.com/ACINQ/eclair/pull/2202) - Use NodeAddress everywhere 
and [PR #2296](https://github.com/ACINQ/eclair/pull/2296) - drop Tor v2 support


eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -136,6 +136,20 @@ class PeerSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with Paralle
     mockServer.close()
   }
 
+  test("return connection failure for a peer with an invalid dns host name") { f =>
+    import f._
+
+    // this actor listens to connection requests and creates connections
+    system.actorOf(ClientSpawner.props(nodeParams.keyPair, nodeParams.socksProxy_opt, nodeParams.peerConnectionConf, TestProbe().ref, TestProbe().ref))
+
+    val invalidDnsHostname_opt = NodeAddress.fromParts("eclair.invalid", 9735).toOption
```
remyers, line 145: Note: the [.invalid](https://en.wikipedia.org/wiki/.invalid) top level domain will never resolve.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -312,16 +312,22 @@ object NodeAddress {
...
    *
    * We don't attempt to resolve onion addresses (it will be done by the tor proxy), so we just recognize them based on
    * the .onion TLD and rely on their length to separate v2/v3.
+   *
+   * We resolve host names comprised of only numbers and periods (IPv4) or that contain a colon (IPv6).
+   * Other host names are assumed to be a DNS name and are not immediately resolved.
+   *
    */
   def fromParts(host: String, port: Int): Try[NodeAddress] = Try {
+    val ipv4v6 = "^([0-9.]*)?$|(:)".r
```
t-bast, line 325: Isn't there a function in the networking utilities we have available on the JVM to do that?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -357,7 +364,20 @@ case class NodeAnnouncement(signature: ByteVector64,
...
+                            tlvStream: TlvStream[NodeAnnouncementTlv] = TlvStream.empty) extends RoutingMessage with AnnouncementMessage with HasTimestamp {
+
+  def validAddresses: List[NodeAddress] = {
+    // if port is equal to 0, SHOULD ignore ipv6_addr OR ipv4_addr OR hostname; SHOULD ignore Tor v2 onion services.
+    val validAddresses = addresses.filter(address => address.port != 0 || address.isInstanceOf[Tor3]).filterNot( address => address.isInstanceOf[Tor2])
+    // if more than one type 5 address is announced, SHOULD ignore the additional data.
+    validAddresses.filter(!_.isInstanceOf[DnsHostname]) ++ validAddresses.filter(_.isInstanceOf[DnsHostname]).take(1)
+  }
+
+  def shouldRebroadcast: Boolean = {
```
t-bast, line 376: nit: this should be a `val`, and `validAddresses` as well.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -357,7 +364,20 @@ case class NodeAnnouncement(signature: ByteVector64,
...
                             alias: String,
                             addresses: List[NodeAddress],
-                            tlvStream: TlvStream[NodeAnnouncementTlv] = TlvStream.empty) extends RoutingMessage with AnnouncementMessage with HasTimestamp
+                            tlvStream: TlvStream[NodeAnnouncementTlv] = TlvStream.empty) extends RoutingMessage with AnnouncementMessage with HasTimestamp {
+
+  def validAddresses: List[NodeAddress] = {
+    // if port is equal to 0, SHOULD ignore ipv6_addr OR ipv4_addr OR hostname; SHOULD ignore Tor v2 onion services.
+    val validAddresses = addresses.filter(address => address.port != 0 || address.isInstanceOf[Tor3]).filterNot( address => address.isInstanceOf[Tor2])
+    // if more than one type 5 address is announced, SHOULD ignore the additional data.
+    validAddresses.filter(!_.isInstanceOf[DnsHostname]) ++ validAddresses.filter(_.isInstanceOf[DnsHostname]).take(1)
```
t-bast, line 373: nit:
```suggestion
    validAddresses.filter(!_.isInstanceOf[DnsHostname]) ++ validAddresses.find(_.isInstanceOf[DnsHostname])
```

eclair-core/src/test/scala/fr/acinq/eclair/StartupSpec.scala
```diff
@@ -273,4 +273,26 @@ class StartupSpec extends AnyFunSuite {
     assert(nodeParamsAttempt2.isSuccess)
   }
 
+  test("NodeParams should fail when server.public-ips addresses or server.port are invalid") {
```
t-bast, line 276: nit: there are whitespace formatting issues, can you run `Ctrl + Alt + L`?

eclair-core/src/test/scala/fr/acinq/eclair/StartupSpec.scala
```diff
@@ -273,4 +273,26 @@ class StartupSpec extends AnyFunSuite {
...
+      val serverConf = ConfigFactory.parseMap(Map(
+        s"server.public-ips" -> test.publicIps.asJava,
+        s"server.port" -> test.port,
+      ).asJava).withFallback(defaultConf)
+      val attempt = Try(makeNodeParamsWithDefaults(serverConf))
+      if (test.error.isEmpty)
+        assert(attempt.isSuccess)
+      else
+        assert(attempt.isFailure && attempt.failed.get.getMessage.contains(test.error.get) &&
+          (test.errorIp.isEmpty || attempt.failed.get.getMessage.contains(test.errorIp.get)))
```
t-bast, line 294: It would be more readable if this was split across three distinct assertions (and add braces around if/else).

eclair-core/src/test/scala/fr/acinq/eclair/db/NetworkDbSpec.scala
```diff
@@ -58,7 +58,7 @@ class NetworkDbSpec extends AnyFunSuite {
       val node_1 = Announcements.makeNodeAnnouncement(randomKey(), "node-alice", Color(100.toByte, 200.toByte, 300.toByte), NodeAddress.fromParts("192.168.1.42", 42000).get :: Nil, Features.empty)
       val node_2 = Announcements.makeNodeAnnouncement(randomKey(), "node-bob", Color(100.toByte, 200.toByte, 300.toByte), NodeAddress.fromParts("192.168.1.42", 42000).get :: Nil, Features(VariableLengthOnion -> Optional))
       val node_3 = Announcements.makeNodeAnnouncement(randomKey(), "node-charlie", Color(100.toByte, 200.toByte, 300.toByte), NodeAddress.fromParts("192.168.1.42", 42000).get :: Nil, Features(VariableLengthOnion -> Optional))
-      val node_4 = Announcements.makeNodeAnnouncement(randomKey(), "node-charlie", Color(100.toByte, 200.toByte, 300.toByte), Tor2("aaaqeayeaudaocaj", 42000) :: Nil, Features.empty)
+      val node_4 = Announcements.makeNodeAnnouncement(randomKey(), "node-charlie", Color(100.toByte, 200.toByte, 300.toByte), Tor3("of7husrflx7sforh3fw6yqlpwstee3wg5imvvmkp4bz6rbjxtg5nljad", 42000) :: Nil, Features.empty)
```
t-bast, line 61: It would be a good idea to also add a node announcement containing a DNS host in the DB and verify that we correctly read it back.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
t-bast, line 145: You can also add an assertion to verify that this is correctly parsed:

```scala
assert(invalidDnsHostname_opt.nonEmpty)
```

t-bast (commented):
The logic looks good to me, I only have small cosmetic/architectural comments, good job!
This does need a rebase though.

docs/release-notes/eclair-vnext.md
```diff
@@ -38,6 +38,12 @@ Expired incoming invoices that are unpaid will be searched for and purged from t
 * `eclair.purge-expired-invoices.enabled = true
 * `eclair.purge-expired-invoices.interval = 24 hours`
 
+#### Public IP addresses can be DNS host names, but not Tor v2 addresses
+
+You can now specify a DNS host name as one of your `server.public-ips` addresses (see PR [#911](https://github.com/lightning/bolts/pull/911)). Note: you can not specify more than one DNS host name.
+
+Tor v2 addresses are no longer supported as a `server.public-ips` address and will be ignored in gossip messages (see PR [#940](https://github.com/lightning/bolts/pull/940]).
```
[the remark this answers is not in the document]
remyers: fixed in [9d91c5a](https://github.com/ACINQ/eclair/pull/2234/commits/9d91c5a4b7daf9a611e0b73a62e34f59c19a0d27)
I think the corresponding code changes will have to be resolved when I rebase.

eclair-core/src/main/scala/fr/acinq/eclair/router/Announcements.scala
```diff
@@ -89,6 +91,17 @@ object Announcements {
     )
   }
 
+  case class AddressException(message: String) extends IllegalArgumentException(message)
+
+  def validateAddresses(addresses: List[NodeAddress]): Option[AddressException] = {
```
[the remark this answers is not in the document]
remyers: Good point, explicit braces are easier to read. Fixed in [a9429ba](https://github.com/ACINQ/eclair/pull/2234/commits/a9429bad747bf61566a1618bcfc75b0de1767c4b).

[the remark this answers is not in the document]
remyers: I was originally thinking about it as something related to the Announce specification, but agree now since it's only used when parsing NodeParams it should be moved.

Fixed in [34e6bef](https://github.com/ACINQ/eclair/pull/2234/commits/34e6befa476f84efec66559567b7e6442bf23c65).

remyers:
As per Roasbeef at today's [meeting](https://github.com/lightning/bolts/issues/1015), [LND 0.15](https://github.com/lightningnetwork/lnd/blob/master/docs/release-notes/release-notes-0.15.0.md) will propagate DNS hostnames, so nodes running a version from about a month ago should not disappear if they add dns host names to their node announcements.

eclair-core/src/main/scala/fr/acinq/eclair/tor/Socks5Connection.scala
```diff
@@ -237,6 +237,11 @@ object Socks5ProxyParams {
       case _: IPv6 if proxyParams.useForIPv6 => Some(proxyParams.address)
       case _: Tor2 if proxyParams.useForTor => Some(proxyParams.address)
       case _: Tor3 if proxyParams.useForTor => Some(proxyParams.address)
+      case _: DnsHostname => InetAddress.getByName(address.host) match {
```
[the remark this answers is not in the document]
remyers: Good point!  `proxyAddress()` should only decide whether or not to use a proxy for a particular address type.

I also agree it's better to add  a parameter that makes it explicit if a socks5 proxy should be used for DNS hostnames. I prefer `proxyParams.useForDnsHostnames` instead of `useForDnsHosts` because it better matches the internal naming if someone searches the code. The corresponding config file parameter is then `socks5.use-for-dnshostnames` .

Note: DNS hostnames are resolved as `remoteAddress` in Client.scala. After connecting to a proxy a `Socks5Connection` uses `remoteAddress` and expects it to be a resolved ipv4 or ipv6 address. 

fixed in [ec79f2e](https://github.com/ACINQ/eclair/pull/2234/commits/ec79f2e581f34af9a41356c0aade3e5e7640b4ed)

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/CommonCodecs.scala
```diff
@@ -126,12 +126,15 @@ object CommonCodecs {
 
   def base32(size: Int): Codec[String] = bytes(size).xmap(b => new Base32().encodeAsString(b.toArray).toLowerCase, a => ByteVector(new Base32().decode(a.toUpperCase())))
 
+  val punycode: Codec[String] = variableSizeBytes(uint8, ascii)
```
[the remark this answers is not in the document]
remyers: Although punycode is just ascii, if we wanted to detect misformed punycode we would need to implement a more complex codec. I don't think there's much to gain here though. If the punycode is wrong the dns lookup will just fail.

Fixed in [160db37](https://github.com/ACINQ/eclair/pull/2234/commits/160db3729f5f44fb04200be27d05de2609bd17c0).

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
remyers: fixed in [fd32764](https://github.com/ACINQ/eclair/pull/2234/commits/fd32764cf3f01be1f6e35832f365250de252c2aa)

eclair-core/src/test/scala/fr/acinq/eclair/StartupSpec.scala
remyers: fixed in [ff2ac04](https://github.com/ACINQ/eclair/pull/2234/commits/ff2ac0419ff96c9dabc00e68add94c35e0ad8344)

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
remyers: fixed in [ff2ac04](https://github.com/ACINQ/eclair/pull/2234/commits/ff2ac0419ff96c9dabc00e68add94c35e0ad8344)

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
remyers: fixed in [ff2ac04](https://github.com/ACINQ/eclair/pull/2234/commits/ff2ac0419ff96c9dabc00e68add94c35e0ad8344)

eclair-core/src/test/scala/fr/acinq/eclair/StartupSpec.scala
remyers: fixed in [ff2ac04](https://github.com/ACINQ/eclair/pull/2234/commits/ff2ac0419ff96c9dabc00e68add94c35e0ad8344)

eclair-core/src/test/scala/fr/acinq/eclair/db/NetworkDbSpec.scala
remyers: fixed in [fbf5529](https://github.com/ACINQ/eclair/pull/2234/commits/fbf5529f2d2a9785c94d5f31abee07f004d8b510)

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
remyers: fixed in [505c7cb](https://github.com/ACINQ/eclair/pull/2234/commits/505c7cb0442093fb134d503dde7689dfee5f8692)

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
remyers: Updated the comment too in [76d24d2](https://github.com/ACINQ/eclair/pull/2234/commits/76d24d2513bddcb0fd7de5930bbb32b4198d0005)

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -357,7 +363,20 @@ case class NodeAnnouncement(signature: ByteVector64,
                             rgbColor: Color,
                             alias: String,
                             addresses: List[NodeAddress],
-                            tlvStream: TlvStream[NodeAnnouncementTlv] = TlvStream.empty) extends RoutingMessage with AnnouncementMessage with HasTimestamp
+                            tlvStream: TlvStream[NodeAnnouncementTlv] = TlvStream.empty) extends RoutingMessage with AnnouncementMessage with HasTimestamp {
+
+  val validAddresses: List[NodeAddress] = {
+    // if port is equal to 0, SHOULD ignore ipv6_addr OR ipv4_addr OR hostname; SHOULD ignore Tor v2 onion services.
+    val validAddresses = addresses.filter(address => address.port != 0 || address.isInstanceOf[Tor3]).filterNot( address => address.isInstanceOf[Tor2])
```
t-bast, line 370: nit:
```suggestion
    val validAddresses = addresses.filter(address => address.port != 0 || address.isInstanceOf[Tor3]).filterNot(address => address.isInstanceOf[Tor2])
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/CommonCodecs.scala
```diff
@@ -132,6 +132,7 @@ object CommonCodecs {
       .typecase(2, (ipv6address :: uint16).as[IPv6])
       .typecase(3, (base32(10) :: uint16).as[Tor2])
       .typecase(4, (base32(35) :: uint16).as[Tor3])
+      .typecase( 5, (variableSizeBytes(uint8, ascii) :: uint16).as[DnsHostname])
```
t-bast, line 135: nit:
```suggestion
      .typecase(5, (variableSizeBytes(uint8, ascii) :: uint16).as[DnsHostname])
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -24,6 +24,7 @@ import fr.acinq.eclair.blockchain.fee.FeeratePerKw
 import fr.acinq.eclair.channel.{ChannelFlags, ChannelType}
 import fr.acinq.eclair.{BlockHeight, CltvExpiry, CltvExpiryDelta, Feature, Features, InitFeature, MilliSatoshi, ShortChannelId, TimestampSecond, UInt64}
 import scodec.bits.ByteVector
+import sun.net.util.IPAddressUtil.{isIPv4LiteralAddress, isIPv6LiteralAddress}
```
t-bast, line 27: I don't know if that dependency will restrict what JRE we can run on...@pm47 is that ok to use? Otherwise do you know where we can find an equivalent utility function to check if a `String` is a litteral ipv4/ipv6 address? I'm surprised this cannot be found in the common libraries

t-bast (commented):
LGTM, a couple of nits and a question for @pm47 about the `sun.net` dependency.
This needs a rebase, which should be trivial.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
pm47: I would use guava's `InetAddresses.isInetAddress`, it's portable and consistent with how we use it:
```scala
case _ if InetAddresses.isInetAddress(host) => IPAddress(InetAddress.getByName(host), port)
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
remyers: fixed in [e248b7c](https://github.com/ACINQ/eclair/pull/2234/commits/e248b7c043ffa4e74f61e3827264a9740ad612c4)

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/CommonCodecs.scala
remyers: fixed in [e248b7c](https://github.com/ACINQ/eclair/pull/2234/commits/e248b7c043ffa4e74f61e3827264a9740ad612c4)

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
remyers: Good catch! looks like a much better function. fixed in [a8e58ee](https://github.com/ACINQ/eclair/pull/2234/commits/a8e58ee76529776b56c432575da0f63a540d872a)

remyers: The change to `InetAddreses.isInetAddress` required a few changes to correct test failures:
 - [c2df6d2](https://github.com/ACINQ/eclair/pull/2234/commits/c2df6d2fbe710738efae018815d6c3bf71c4ba04) was needed because brackets are not part of a standard IPv6 address *except* when it also includes the port (eg. in a URI). It seems safer to filter the brackets just in-case some node-announcements or API call includes the brackets. Without this change the "get proxy address" test fails [here](https://github.com/remyers/eclair/blob/c2df6d2fbe710738efae018815d6c3bf71c4ba04/eclair-core/src/test/scala/fr/acinq/eclair/tor/Socks5ConnectionSpec.scala#L45) because `useForIPv6 = false` and it would otherwise be interpreted as a DNS hostname.
 - [c2df6d2](https://github.com/ACINQ/eclair/pull/2234/commits/c2df6d2fbe710738efae018815d6c3bf71c4ba04) was needed because the mock api uses `localhost` but the test expects a resolved address `127.0.0.1`. Because `localhost` does not return true for `isInetAddress` it will not be resolved to the default IPv4 localhost address.

t-bast (approved):
LGTM, good job :100:

PR #2236, opened 2022-04-13
Rename channel funder to initiator

t-bast:
With dual funding, both peers may be funders, but only one of them is the initiator.
This is just a dumb renaming, there is not logic change.

I haven't changed types such as `INPUT_INIT_FUNDER`, because they impact less files and I will make them inherit a shared trait with their dual-funding counterpart, so it's better done with some of the dual funding logic in the same PR.


remyers:
LGTM!

remyers (dismissed):

t-bast:
I have also updated string comments in a few places in https://github.com/ACINQ/eclair/pull/2236/commits/0ebd22aa1ab98758cf878ef8fa00aa61d4297e1c, I missed them on my first pass!

pm47 (approved):
This changes the formatting of the json persisted in database, but we will certainly have to modify it more thoroughly anyway for dual funding.

t-bast:
Yes, I figured this shouldn't be too much of an issue.

PR #2237, opened 2022-04-14
Add `localChannelReserve` and `remoteChannelReserve`

t-bast:
This is easier to use than having to decide which params we should look into (local or remote), which forces us to think through one level of indirection.

But the main reason I'm adding this is that it will also be easier to integrate with dual funding. When using dual funding, we won't negotiate a channel reserve, it will automatically be 1% of the funding outpoint and won't have a corresponding field in `open_channel` and `accept_channel`. We could inject that value in `LocalParams.channelReserve` and `RemoteParams.channelReserve`, but once we had splicing we would need to correctly update this value in our DB at every splice (since the channel capacity will change). It will instead be easier to avoid using that field and automatically compute the correct value in `localChannelReserve` and `remoteChannelReserve` based on the current channel capacity. It will look like this:

```scala
  } else {
    remoteParams.requestedChannelReserve
  }
```

Otherwise we would need to update the channel codecs to put the existing channel reserve field elsewhere, and there isn't even a good candidate on where to put that (since we'll keep the `Commitments` type mostly unchanged for dual funding).

The second commit is an unrelated renaming of a confusing field (`initialFeeratePerKw` -> `commitTxFeerate`).


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -372,15 +378,15 @@ object Commitments {
...
+      return Left(InsufficientFunds(commitments.channelId, amount = cmd.amount, missing = -missingForSender.truncateToSatoshi, reserve = commitments1.localChannelReserve, fees = if (commitments1.localParams.isInitiator) fees else 0.sat))
     } else if (missingForReceiver < 0.msat) {
       if (commitments.localParams.isInitiator) {
         // receiver is not the channel initiator; it is ok if it can't maintain its channel_reserve for now, as long as its balance is increasing, which is the case if it is receiving a payment
       } else {
-        return Left(RemoteCannotAffordFeesForNewHtlc(commitments.channelId, amount = cmd.amount, missing = -missingForReceiver.truncateToSatoshi, reserve = commitments1.remoteParams.channelReserve, fees = fees))
+        return Left(RemoteCannotAffordFeesForNewHtlc(commitments.channelId, amount = cmd.amount, missing = -missingForReceiver.truncateToSatoshi, reserve = commitments1.remoteChannelReserve, fees = fees))
```
t-bast, line 389: It was worth doing, the value here was wrong! It doesn't matter much, it's just an informative error, but still it could be confusing for readers.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -466,7 +466,7 @@ case class LocalParams(nodeId: PublicKey,
                        fundingKeyPath: DeterministicWallet.KeyPath,
                        dustLimit: Satoshi,
                        maxHtlcValueInFlightMsat: UInt64, // this is not MilliSatoshi because it can exceed the total amount of MilliSatoshi
-                       channelReserve: Satoshi,
+                       requestedChannelReserve: Satoshi,
```
pm47, line 469: You can probably make that an `Option` and update the codec to have it take a `channelFeatures` in parameter.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -481,7 +481,7 @@ case class LocalParams(nodeId: PublicKey,
 case class RemoteParams(nodeId: PublicKey,
                         dustLimit: Satoshi,
                         maxHtlcValueInFlightMsat: UInt64, // this is not MilliSatoshi because it can exceed the total amount of MilliSatoshi
-                        channelReserve: Satoshi,
+                        requestedChannelReserve: Satoshi,
```
pm47, line 484: Same as above.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
t-bast: Good idea! Done in https://github.com/ACINQ/eclair/pull/2237/commits/0bcf7a089b35765d8425f0ad7987b6a237e241fc

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2237/commits/0bcf7a089b35765d8425f0ad7987b6a237e241fc

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: I'm not surprised, those parameters that apply to the other side are tricky.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -196,7 +196,7 @@ final case class CMD_GET_CHANNEL_INFO(replyTo: ActorRef)extends HasReplyToComman
 /** response to [[Command]] requests */
 sealed trait CommandResponse[+C <: Command]
 sealed trait CommandSuccess[+C <: Command] extends CommandResponse[C]
-sealed trait CommandFailure[+C <: Command, +T <: Throwable] extends CommandResponse[C] { def t: Throwable }
+sealed trait CommandFailure[+C <: Command, +T <: Throwable] extends CommandResponse[C] { def t: T }
```
pm47, line 199: 🦅 👁️

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -466,22 +466,24 @@ case class LocalParams(nodeId: PublicKey,
...
+                       requestedChannelReserve_opt: Option[Satoshi],
                        htlcMinimum: MilliSatoshi,
                        toSelfDelay: CltvExpiryDelta,
                        maxAcceptedHtlcs: Int,
                        isInitiator: Boolean,
                        defaultFinalScriptPubKey: ByteVector,
                        walletStaticPaymentBasepoint: Option[PublicKey],
-                       initFeatures: Features[InitFeature])
+                       initFeatures: Features[InitFeature]) {
+  val requestedChannelReserve: Satoshi = requestedChannelReserve_opt.getOrElse(0 sat)
```
pm47, line 477: This is fishy, I expected the field `requestedChannelReserve_opt` to be made `private` (or with a comment to discourage its direct use) and the following method in `Commitments`:

```
val requestedChannelReserve: Satoshi = requestedChannelReserve_opt.getOrElse(1 % * capacity)
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
```diff
@@ -69,7 +68,7 @@ private[channel] object ChannelCodecs0 {
         ("channelPath" | keyPathCodec) ::
         ("dustLimit" | satoshi) ::
         ("maxHtlcValueInFlightMsat" | uint64) ::
-        ("channelReserve" | satoshi) ::
+        ("channelReserve" | conditional(included = true, satoshi)) ::
```
pm47, line 71: I suppose you have used the `conditional` codec here just for consistency?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -211,6 +211,12 @@ case class Commitments(channelId: ByteVector32,
 
   val capacity: Satoshi = commitInput.txOut.amount
 
+  /** Channel reserve that applies to our funds. */
+  val localChannelReserve: Satoshi = remoteParams.requestedChannelReserve
+
+  /** Channel reserve that applies to our peer's funds. */
+  val remoteChannelReserve: Satoshi = localParams.requestedChannelReserve
```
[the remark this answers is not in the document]
t-bast: This is done in the future PR, once it becomes necessary to have a channel reserve for dual-funding.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
t-bast: Yes, but I also don't see how else you would do it?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
pm47: > But I can remove the requestedChannelReserve field entirely, and let callers use commitments.xxxChannelReserve or when we have no commitments do the getOrElse themselves, what do you think?

Yes that works, we always have a `Commitments`, except in `Helpers.validate*`, where it makes sense to be more manual anyway.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version0/ChannelCodecs0.scala
pm47: Oh that's an optional field, right.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2237/commits/47cb396bd8216c843038cc8e80d2244b7afa5424
I will need to make sure the dual funding PRs correctly compute the reserve in `makeFirstCommitTxs`.

pm47 (approved):

PR #2238, opened 2022-04-15
Refactor blinded relay

t-bast:
The main reason for this refactoring is to simplify the type hierarchy and codecs inside `PaymentOnion.scala`. They were very hard to understand and the inheritance tree didn't map easily to the actual scenarios.

The goal of our codecs is to make it easy for readers to match the spec to our code. This is why for example I chose to have private tlv codecs that support all declared tlv fields and do the business logic of restricting which fields are required or forbidden in custom, higher-level codecs. This approach also lets us mark odd tlv types as forbidden if necessary, which was impossible with the previous approach. I also chose to relax some codec constraints to be more forward-compatible.

I added some comments to the `IncomingPaymentPacket.decrypt` function since it's becoming quite challenging to review! I think the flow is now still a bit complex (there are many cases to handle) but they are quite cleanly separated so it should be easier to read.

I made sure the commit compiles, but some tests should be updated. As mentioned in #2154, I think we need to add tests to `PaymentPacketSpec.scala` before this can be merged. We should also enrich the codec tests to test all the restrictions we're adding. But maybe we should wait to get some feedback on the spec before adding those tests, it would be sad to spend time on them and have to throw them away if some spec feedback makes us change our codecs...


thomash-acinq:
I think having the checks done at the end is a lot more error prone. I actually started with some code a bit similar to this but decided to not try to use the same codecs for different scenarios because it was too messy to do all the checks properly. For instance your code has a serious flaw that would allow an attacker to deanonymize the route: you allow setting `blinding_point` inside the onion for nodes after the introduction node, you even use it instead of the one provided by the previous node. There are four scenarios depending on if I am the last node and if I have received a blinding point from the previous node to decrypt the onion, I think these four scenarios must distinct codecs to prevent such mistakes. And the overlap between these four codecs is not that big anyway.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
```diff
@@ -114,58 +120,67 @@ object RouteBlindingEncryptedDataCodecs {
...
-
-  private val messageRelayDataTlvCodec = discriminated[RouteBlindingEncryptedDataTlv].by(varint)
+  private val encryptedDataTlvCodec = discriminated[RouteBlindingEncryptedDataTlv].by(varint)
     .typecase(UInt64(1), padding)
+    .typecase(UInt64(2), outgoingChannelId)
     .typecase(UInt64(4), outgoingNodeId)
+    .typecase(UInt64(6), pathId)
     .typecase(UInt64(8), nextBlinding)
+    .typecase(UInt64(10), paymentRelay)
+    .typecase(UInt64(12), paymentConstraints)
```
thomash-acinq, line 138: I really don't see this as a readability improvement, these two codecs only had two TLVs in common (padding being there only for testing purposes) and now you need to explicitly disallow a bunch of TLVs in the narrowed codecs.

t-bast: It is a readability improvement: when you read the spec you see a list of tlvs that can be included in onion payloads. Then you look at the `eclair` code and you see the same list and can easily check that everything that's in the spec is supported by `eclair`. In the previous version it was split into several codecs, there were more types in the type hierarchy and a `fallback(fallback(X, Y), Z)` which is very hacky...

> you need to explicitly disallow a bunch of TLVs in the narrowed codecs.

I only kept that code because it was in your version, but I'm really in favor of removing it.
It's completely unnecessary: what do we care if the sender wasted space by putting some tlvs that we'll simply ignore?
This is extra code that doesn't protect us more than simply ignoring these fields.
The only benefit I see is to provide senders with a nice error message to figure out where they messed up, but it doesn't make sense in route blinding since we will never return a sender-readable error.

t-bast:
> I think having the checks done at the end is a lot more error prone.

I see this as separation of concerns, which in my opinion is critical for complex projects. The codecs should contain code to encode and decode, and apply only the minimal validation needed to build easy-to-use types (so we need to validate that a few tlvs are present, because we don't want to make them `Option[...]` when they should always be there, otherwise it's cumbersome to use). I kept a few unnecessary checks because they were in your version, but I'm really in favor of removing them.

Validation should be a different component than the component that encodes/decodes: that simply doesn't work at the same logical layer.

> For instance your code has a serious flaw that would allow an attacker to deanonymize the route: you allow setting blinding_point inside the onion for nodes after the introduction node, you even use it instead of the one provided by the previous node.

How is that even a flaw? The `blinding_point` is either valid or invalid. If invalid we won't be able to decrypt and will reject the message, which doesn't open any attack vector. The malicious node is just losing fees if it does that, I don't see any reason to do that. Nodes cannot know the valid `blinding_point` for anyone except the node right after them.

> There are four scenarios 

That is true, but in my opinion it's not the codecs job to do this. It's the job of a business logic component such as `IncomingPaymentPacket`. I currently kept the same structure as your PR, but I agree that we should decompose it better to more explicitly handle each of these cases and properly unit test them. In the end, the only thing that really protects your code from bugs is tests.

> And the overlap between these four codecs is not that big anyway.

I agree with that, this is not why I dislike it. I dislike the more complex type/codec hierarchy and the `fallback(fallback(X, Y), Z)` codec, which is a very hacky way of achieving this.

Anyway, we need to wait for some spec feedback before we merge this to `master` to ensure other implementations are on board with the `payment_details` / `payment_constraints` flow, so let's wait for them before finalizing this if you don't like my proposal.

PR #2239, opened 2022-04-22
Give best capacity score to local channels

thomash-acinq:
In path-finding, we give a better score to channels with higher capacity because they have a higher change to have enough balance to relay the payment. But for local channels we know the balance so we can be sure that the payment can be relayed.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
@@ -1889,7 +1889,7 @@ class RouteCalculationSpec extends AnyFunSuite with ParallelTestExecution {
     // A === B === C -- D
     //  \_________/
     val g = DirectedGraph(List(
-      makeEdge(1L, a, b, 100 msat, 100, minHtlc = 1000 msat, capacity = 100000000 sat),
+      makeEdge(1L, a, b, 100 msat, 100, minHtlc = 1000 msat, capacity = 100000000 sat, balance_opt = Some(10000000 msat)),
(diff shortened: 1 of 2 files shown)


thomash-acinq:
> Can you add a unit test that fails without that change?

Done

eclair-core/src/test/scala/fr/acinq/eclair/router/RouteCalculationSpec.scala
```diff
@@ -1884,6 +1884,28 @@ class RouteCalculationSpec extends AnyFunSuite with ParallelTestExecution {
       assert(findMultiPartRoute(g, a, b, amount, 101_000 msat, Set.empty, Set.empty, Set.empty, Nil, routeParams = routeParams, currentBlockHeight = BlockHeight(400000)).isSuccess)
     }
   }
+
+  test("small local edge with liquidity is better than big remote edge") {
+    // A === B === C -- D
+    //  \_________/
+    val g = DirectedGraph(List(
+      makeEdge(1L, a, b, 100 msat, 100, minHtlc = 1000 msat, capacity = 100000000 sat),
```
t-bast, line 1892: This edge should also have its balance known to be representative of a real scenario since it's a local channel:
```suggestion
      makeEdge(1L, a, b, 100 msat, 100, minHtlc = 1000 msat, capacity = 100000000 sat, balance_opt = Some(150000 msat)),
```

I made it have an even bigger balance than the A -> C one to showcase that it's still ignored because the A -> C -> D route is better regardless (we only need balance to forward the payment, any "extra" balance doesn't matter for local channels).

thomash-acinq: Done

t-bast (approved):
LGTM :+1:

PR #2240, opened 2022-04-22
Disable invoice purger in tests

pm47:
It was creating a race condition in the test `MultiPartHandlerSpec`.`PaymentHandler should reject incoming multi-part payment with an invalid expiry` due the 0-expiry invoice being immediately purged.


remyers (approved):
LGTM! The purger tests don't use Alice/Bob so are not effected.

PR #2242, opened 2022-04-25
Log routing hint when finding a route

thomash-acinq:
When finding a payment route (typically for routing a trampoline payment) we already log the recipient node id, however it can be a private node hidden behind a routing hint in which case it is more useful to log the public node from the routing hint.
This is added to the `audit.path_finding_metrics` and we also add the `payment_hash` so that it can be joined with the `audit.relayed` and `audit.relayed_trampoline` tables.


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -264,7 +264,7 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
           }
           paymentSent.feesPaid + localFees
       }
-      context.system.eventStream.publish(PathFindingExperimentMetrics(cfg.recipientAmount, fees, status, duration, now, isMultiPart = true, request.routeParams.experimentName, cfg.recipientNodeId))
+      context.system.eventStream.publish(PathFindingExperimentMetrics(cfg.paymentHash, cfg.recipientAmount, fees, status, duration, now, isMultiPart = true, request.routeParams.experimentName, cfg.recipientNodeId, cfg.invoice.flatMap(_.routingInfo.headOption.flatMap(_.headOption.map(_.nodeId)))))
```
t-bast, line 267: Shouldn't it be `cfg.invoice.flatMap(_.routingInfo.headOption.flatMap(_.lastOption.map(_.nodeId)))`? What exactly are you interested in? I guess it should be the node closest to the recipient, right (routing hints may contain more than one hop)?

If there are multiple routes provided in the invoice, aren't we interested in all nodes that are connected to the recipient? In which case we would want to store `routingHintNeighbors: Set[PublicKey]`, that we would compute with `cfg.invoice.flatMap(_.routingInfo.flatMap(_.lastOption.map(_.nodeId))).toSet` (to be tested, I probably wrote that too quickly and it may be buggy)?

Same in `PaymentLifecycle`.

thomash-acinq: I am interested in the last public node on the route to the recipient. The first node of the routing hint has to be public but the others are probably private.
In the case of mobile wallets hidden behind a hub, getting the id of the mobile wallet is not very useful as we can't contact it directly, I want to know the id of the hub instead.
I am assuming here that it is extremely rare to have a private node advertising more than one hub in its routing hints, do you think it's unreasonable?

t-bast: > The first node of the routing hint has to be public but the others are probably private.

Got it, I see the assumption, that's why you take the first one in the routing hint instead of the last one.
We can't know for sure whether the ones afterwards are private or not though, but I agree it's a reasonable assumption.

> In the case of mobile wallets hidden behind a hub, getting the id of the mobile wallet is not very useful as we can't contact it directly, I want to know the id of the hub instead.

The id of the mobile wallet wouldn't appear in the routing hints anyway? In a route `A -> B -> mobile-wallet` the routing hints would contain nodeId(A) and nodeId(B). But what you're saying is that B is most likely a private node (otherwise the wallet would have only included `B -> mobile-wallet` in its routing hint).

> I am assuming here that it is extremely rare to have a private node advertising more than one hub in its routing hints, do you think it's unreasonable?

I think we should expect multiple routing hints: Breez and LDK include more than one for example. It doesn't cost us much to store a `Set[PublicKey]` instead of just one, so probably worth doing to avoid another future migration?

So let's keep your assumption about private nodes and log the first `node_id` of every routing hint? Something like `cfg.invoice.flatMap(_.routingInfo.flatMap(_.headOption.map(_.nodeId))).toSet`?

thomash-acinq: I've done it, I expect it to be a rare case but it's definitely cleaner like that.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -138,9 +150,12 @@ class SqliteAuditDb(val sqlite: Connection) extends AuditDb with Logging {
         statement.executeUpdate("CREATE INDEX metrics_timestamp_idx ON path_finding_metrics(timestamp)")
         statement.executeUpdate("CREATE INDEX metrics_mpp_idx ON path_finding_metrics(is_mpp)")
         statement.executeUpdate("CREATE INDEX metrics_name_idx ON path_finding_metrics(experiment_name)")
+        statement.executeUpdate("CREATE INDEX metrics_recipient_idx ON path_finding_metrics(recipient_node_id)")
+        statement.executeUpdate("CREATE INDEX metrics_hash_idx ON path_finding_metrics(payment_hash)")
+        statement.executeUpdate("CREATE INDEX metrics_hint_idx ON path_finding_metrics(routing_hint_node_ids)")
```
t-bast, line 155: Does that really make sense? Since this is a `TEXT` field, does the index let you efficiently filter on a substring? I imagine that if we want to filter on routing hints, we would filter on only one routing hint `node_id` but want to find all payments for which it was one of the routing hints?

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
```diff
@@ -306,6 +324,8 @@ class SqliteAuditDb(val sqlite: Connection) extends AuditDb with Logging {
       statement.setBoolean(6, m.isMultiPart)
       statement.setString(7, m.experimentName)
       statement.setBytes(8, m.recipientNodeId.value.toArray)
+      statement.setBytes(9, m.paymentHash.toArray)
+      statement.setString(10, Serialization.write(m.routingHints.map(_.value.toHex)))
```
t-bast, line 328: We usually don't put string serialization inside sqlite (we prefer deterministic binary serialization with scodec and only do json with postgres)...but in this case it's a really simple one, so maybe it doesn't hurt? I'm not sure what is best practice here, will defer to @pm47

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
thomash-acinq: I'm assuming that this field will contain 0 or 1 node id

thomash-acinq (commented):
Added tests for the migration and changed the serialization in sqlite to not use JSON but just concatenate the node ids

t-bast (commented):
I'll let @pm47 chime in on that one, he may have better feedback than me on how to serialize the `routing_hint_node_ids` for each type of DB and whether to add the index or not.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
t-bast: It's probably fine, but I'm just worried about users with sqlite on the same machine as their node, I'd like to avoid impacting their node performance as much as possible. Pierre-Marie probably has more experience with that, let's see what he thinks.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -251,11 +251,13 @@ object PaymentFailure {
...
+                                        amount: MilliSatoshi,
                                         fees: MilliSatoshi,
                                         status: String,
                                         duration: FiniteDuration,
                                         timestamp: TimestampMilli,
                                         isMultiPart: Boolean,
                                         experimentName: String,
-                                        recipientNodeId: PublicKey)
+                                        recipientNodeId: PublicKey,
+                                        routingHints: Set[PublicKey])
```
pm47, line 263: It's not obvious at all what `routingHints` means here. Looking at impl I think we put the first node of every available route hint. 

But nothing prevents us to not use hints at all (destination could be a public node) or to only use an arbitrary tail of a provided route hint right?

Why don't we just log the full routing hint as a json? It would make everything simpler and you could decide at query time what you want to do.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -264,7 +264,7 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
           }
           paymentSent.feesPaid + localFees
       }
-      context.system.eventStream.publish(PathFindingExperimentMetrics(cfg.recipientAmount, fees, status, duration, now, isMultiPart = true, request.routeParams.experimentName, cfg.recipientNodeId))
+      context.system.eventStream.publish(PathFindingExperimentMetrics(cfg.paymentHash, cfg.recipientAmount, fees, status, duration, now, isMultiPart = true, request.routeParams.experimentName, cfg.recipientNodeId, cfg.invoice.map(_.routingInfo.flatMap(_.headOption.map(_.nodeId)).toSet).getOrElse(Set.empty)))
```
pm47, line 267: This snippet is obscure enough to have to be put on a separate line with a name that gives a hint (pun intended) to the reader as to what it is.

```suggestion
      val firstHopOfRouteHints = cfg.invoice
        .map(_.routingInfo).getOrElse(Seq.empty)
        .flatMap(_.headOption.map(_.nodeId)).toSet
      context.system.eventStream.publish(PathFindingExperimentMetrics(cfg.paymentHash, cfg.recipientAmount, fees, status, duration, now, isMultiPart = true, request.routeParams.experimentName, cfg.recipientNodeId, firstHopOfRouteHints))
```

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
pm47: I would not bother implementing on sqlite. It's advanced stuff and we need json support for proper querying anyway.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
pm47: I would simply not implement it on sqlite.

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
pm47: Re: index and performance, why not just add the index later if it turns it's needed?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
t-bast: We've previously discussed this here for context: https://github.com/ACINQ/eclair/pull/2242#discussion_r858490598
I also think that it makes sense to log all routing hints (this way we have all potential information, even if some turns out to be irrelevant).

pm47: > We've previously discussed this here for context: https://github.com/ACINQ/eclair/pull/2242#discussion_r858490598

I somehow missed it.

> I'm assuming that this field will contain 0 or 1 node id

@thomash-acinq This assumption might hold true now but maybe not in the future, so better make it at query time I think.

Here is what a query would look like:
```sql
SELECT jsonb_array_elements(hints) #> '{0,nodeId}'
"02413957815d05abb7fc6d885622d5cdc5b7714db1478cb05813a8474179b83c5c"
"03933884aaf1d6b108397e5efe5c86bcf2d8ca8d2f700eda99db9214fc2712b134"
```

Assuming that the data is:
```sql
WITH data AS (SELECT '[
    [
      {
        "nodeId": "02413957815d05abb7fc6d885622d5cdc5b7714db1478cb05813a8474179b83c5c",
        "shortChannelId": "1975837x88x0",
        "feeBase": 1000,
        "feeProportionalMillionths": 100,
        "cltvExpiryDelta": 144
      },
      {
        "nodeId": "0225ff2ae6a3d9722b625072503c2f64f6eddb78d739379d2ee55a16b3b0ed0a17",
        "shortChannelId": "123454x42x2",
        "feeBase": 1000,
        "feeProportionalMillionths": 100,
        "cltvExpiryDelta": 144
      }
    ],
    [
      {
        "nodeId": "03933884aaf1d6b108397e5efe5c86bcf2d8ca8d2f700eda99db9214fc2712b134",
        "shortChannelId": "1976152x25x0",
        "feeBase": 1000,
        "feeProportionalMillionths": 100,
        "cltvExpiryDelta": 144
      }
    ]
  ]'::jsonb as hints)
```

thomash-acinq: Done, I now put everything and will let Postgres do its JSON magic

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteAuditDb.scala
thomash-acinq: Reverted all the sqlite changes

eclair-core/src/test/scala/fr/acinq/eclair/db/AuditDbSpec.scala
```diff
@@ -704,6 +705,184 @@ class AuditDbSpec extends AnyFunSuite {
...
+              assert(result.getString(4) == "030c3f19d742ca294a55c00376b3b355c3c90d61c6b6b39554dbc7ac19b141c14f")
+              assert(!result.next())
+            }
+          }
+        )
+      case _: TestSqliteDatabases => ()
+    }
+  }
+
+  test("migrate sqlite audit database v8 -> current") {
```
t-bast, line 798: You can remove that test since we removed the sqlite support.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -112,7 +120,7 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
...
           statement.executeUpdate("CREATE TABLE audit.channel_updates (channel_id TEXT NOT NULL, node_id TEXT NOT NULL, fee_base_msat BIGINT NOT NULL, fee_proportional_millionths BIGINT NOT NULL, cltv_expiry_delta BIGINT NOT NULL, htlc_minimum_msat BIGINT NOT NULL, htlc_maximum_msat BIGINT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL)")
-          statement.executeUpdate("CREATE TABLE audit.path_finding_metrics (amount_msat BIGINT NOT NULL, fees_msat BIGINT NOT NULL, status TEXT NOT NULL, duration_ms BIGINT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL, is_mpp BOOLEAN NOT NULL, experiment_name TEXT NOT NULL, recipient_node_id TEXT NOT NULL)")
+          statement.executeUpdate("CREATE TABLE audit.path_finding_metrics (amount_msat BIGINT NOT NULL, fees_msat BIGINT NOT NULL, status TEXT NOT NULL, duration_ms BIGINT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL, is_mpp BOOLEAN NOT NULL, experiment_name TEXT NOT NULL, recipient_node_id TEXT NOT NULL, payment_hash TEXT NOT NULL, routing_hint_node_ids JSONB NOT NULL)")
```
t-bast, line 123:
```suggestion
          statement.executeUpdate("CREATE TABLE audit.path_finding_metrics (amount_msat BIGINT NOT NULL, fees_msat BIGINT NOT NULL, status TEXT NOT NULL, duration_ms BIGINT NOT NULL, timestamp TIMESTAMP WITH TIME ZONE NOT NULL, is_mpp BOOLEAN NOT NULL, experiment_name TEXT NOT NULL, recipient_node_id TEXT NOT NULL, payment_hash TEXT NOT NULL, routing_hints JSONB NOT NULL)")
```

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -102,6 +103,13 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
         statement.executeUpdate("DROP TABLE audit.network_fees")
       }
 
+      def migration1011(statement: Statement): Unit = {
+        statement.executeUpdate("ALTER TABLE audit.path_finding_metrics ADD COLUMN payment_hash TEXT")
+        statement.executeUpdate("ALTER TABLE audit.path_finding_metrics ADD COLUMN routing_hint_node_ids JSONB")
```
t-bast, line 108:
```suggestion
        statement.executeUpdate("ALTER TABLE audit.path_finding_metrics ADD COLUMN routing_hints JSONB")
```

eclair-core/src/test/scala/fr/acinq/eclair/db/AuditDbSpec.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -102,6 +103,13 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
         statement.executeUpdate("DROP TABLE audit.network_fees")
       }
 
+      def migration1011(statement: Statement): Unit = {
+        statement.executeUpdate("ALTER TABLE audit.path_finding_metrics ADD COLUMN payment_hash TEXT")
+        statement.executeUpdate("ALTER TABLE audit.path_finding_metrics ADD COLUMN routing_hints JSONB")
```
pm47, line 108: The schema after the migration must be exactly the same as without migration, that's the goal of those `migrationXY` methods. Otherwise we have to think of every possible combination in the queries it will be become quickly unmangeable over time. Here the columns can be nullable or mandatory depending on history.

For this particular case, I think making `routing_hints` nullable is an entirely valid choice.

pm47 (changes requested):

t-bast (dismissed):

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
thomash-acinq: Done, I've made the two new columns nullable

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -314,6 +327,8 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
         statement.setBoolean(6, m.isMultiPart)
         statement.setString(7, m.experimentName)
         statement.setString(8, m.recipientNodeId.value.toHex)
+        statement.setString(9, m.paymentHash.toHex)
+        statement.setString(10, serialization.write(m.routingHints))
```
pm47, line 331: Note to self: we are using `writePretty` in `PgChannelsDb`/`PgNetworkDb` which is useless since `jsonb` doesn't preserve formatting.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
```diff
@@ -251,11 +251,13 @@ object PaymentFailure {
...
+                                        amount: MilliSatoshi,
                                         fees: MilliSatoshi,
                                         status: String,
                                         duration: FiniteDuration,
                                         timestamp: TimestampMilli,
                                         isMultiPart: Boolean,
                                         experimentName: String,
-                                        recipientNodeId: PublicKey)
+                                        recipientNodeId: PublicKey,
+                                        routingHints: Seq[Seq[ExtraHop]])
```
pm47, line 263: This would remove some boilerplate code and make us store `null` in database instead of an empty json list. It would also make the post-migration data consistent with migrated data.

```suggestion
                                        routingHints_opt: Option[Seq[Seq[ExtraHop]]])
```

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
pm47: Just to rephrase my previous comment and stress why it's really important: if the migration methods don't produce the exact same schema as the init script, our linear db schema history becomes a tree.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
pm47: Done in #2279.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentEvents.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgAuditDb.scala
```diff
@@ -314,6 +327,8 @@ class PgAuditDb(implicit ds: DataSource) extends AuditDb with Logging {
         statement.setBoolean(6, m.isMultiPart)
         statement.setString(7, m.experimentName)
         statement.setString(8, m.recipientNodeId.value.toHex)
+        statement.setString(9, m.paymentHash.toHex)
+        statement.setString(10, serialization.write(m.routingHints_opt))
```
pm47, line 331: For consistency with migrated data what we really want is a `null` value when there are no routing hints:

```suggestion
        statement.setString(10, m.routingHints_opt.map(serialization.write(_)).orNull)
```

That's what we do for optional channel updates in the network db:
https://github.com/ACINQ/eclair/blob/44e71277a6f1f80d28c1c14cee2f974eb2e1a6db/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala#L111-L112

thomash-acinq: Done.

t-bast (approved):

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/MultiPartPaymentLifecycle.scala
```diff
@@ -264,7 +264,8 @@ class MultiPartPaymentLifecycle(nodeParams: NodeParams, cfg: SendPaymentConfig,
           }
           paymentSent.feesPaid + localFees
       }
-      context.system.eventStream.publish(PathFindingExperimentMetrics(cfg.paymentHash, cfg.recipientAmount, fees, status, duration, now, isMultiPart = true, request.routeParams.experimentName, cfg.recipientNodeId, cfg.invoice.map(_.routingInfo)))
+      val hints_opt = cfg.invoice.flatMap(invoice => if (invoice.routingInfo.isEmpty) None else Some(invoice.routingInfo))
+      context.system.eventStream.publish(PathFindingExperimentMetrics(cfg.paymentHash, cfg.recipientAmount, fees, status, duration, now, isMultiPart = true, request.routeParams.experimentName, cfg.recipientNodeId, hints_opt))
```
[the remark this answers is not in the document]
thomash-acinq: So that we put a `null` in the db in case there is no routing hint instead of an empty list.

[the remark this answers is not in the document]
pm47: Right, `Bolt11Invoice.routingInfo` is a `Seq[Seq[]]`, not an `Option[Seq[Seq]]`. Which is correct because `r` is a repeated field in the invoice.

So there is an inconsistency in the model between `Bolt11Invoice.routingInfo` and `PathFindingExperimentMetrics.routingHints_opt`, but it's not a big deal and it makes the migration seamless.

PR #2244, opened 2022-04-26
Stop compressing with zlib

t-bast:
While zlib provides good compression results for gossip, it's a dependency that had a few important CVEs in the past. Some implementations are reluctant to import it, so we decided to remove it from the specification in https://github.com/lightning/bolts/pull/981

We stop compressing with zlib and only send uncompressed results, while still supporting receiving compressed data. We will remove support for decompression once our monitoring indicates that we stopped receiving compressed data.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -230,7 +230,6 @@ eclair {
 
     sync {
       request-node-announcements = true // if true we will ask for node announcements when we receive channel ids that we don't know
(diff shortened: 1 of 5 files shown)


sstone:
LGTM, but how do we make sure that query messages still fit in 65 Kb ?

t-bast:
> LGTM, but how do we make sure that query messages still fit in 65 Kb ?

Thanks for spotting that, there was indeed an issue!
I don't know how we came up with the previous number of 3200, but it was clearly wrong and untested :/
This is fixed in https://github.com/ACINQ/eclair/pull/2244/commits/a7653f9d4c32fd701e77de408369c20bf1e1fb60

sstone (approved):

PR #2245, opened 2022-04-26
Update dual funding codecs

t-bast:
We discussed the proposed changes (https://github.com/niftynei/lightning-rfc/pull/3) during yesterday's spec meeting, and agreed on the following changes:

* Use `u16` instead of `varint` for lengths
* Revert channel flags extension: we will use tlvs for future use-cases instead of channel flags
* Remove support for non-native segwit inputs (which removes the need to transmit a `scriptSig` field in `tx_add_input`)

NB: this builds on top of #2237 (necessary to continue dual funding work on other branches)


pm47 (commented):
I have not much to say, this doesn't impact current code, so I guess we'll see during interop testing if things work as intended.

t-bast:
Yes, cln is still working on being up-to-date with the latest spec changes.
But this PR now matches the spec, so it should be correct!

pm47 (approved):

PR #2246, opened 2022-04-26
Allow disabling no-htlc commitment fee-bump

t-bast:
When a channel force-close without any pending htlcs, funds are not at risk. We want to eventually get our main output back, but if we are not in a rush we can save on fees by never spending the anchors.

This is disabled by default as there is a potential risk: if the commit tx doesn't confirm and the feerate rises, the commit tx may eventually be below the network's min-relay-fee and won't confirm (at least until package relay is available).

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -167,12 +167,18 @@ eclair {
 
     // number of blocks to target when computing fees for each transaction type
     target-blocks {
-      funding = 6                       // target for the funding transaction
-      commitment = 2                    // target for the commitment transaction (used in force-close scenario) *do not change this unless you know what you are doing*
-      commitment-without-htlcs = 12     // target for the commitment transaction when we have no htlcs to claim (used in force-close scenario) *do not change this unless you know what you are doing*
-      mutual-close = 12                 // target for the mutual close transaction
-      claim-main = 12                   // target for the claim main transaction (tx that spends main channel output back to wallet)
-      safe-utxos-threshold = 10         // when our utxos count is below this threshold, we will use more aggressive confirmation targets in force-close scenarios
+      // target for the funding transaction
+      funding = 6
+      // target for the commitment transaction (used in force-close scenario) *do not change this unless you know what you are doing*
+      commitment = 2
+      // target for the commitment transaction when we have no htlcs to claim (used in force-close scenario) *do not change this unless you know what you are doing*
+      commitment-without-htlcs = 12
+      // target for the mutual close transaction
+      mutual-close = 12
+      // target for the claim main transaction (tx that spends main channel output back to wallet)
+      claim-main = 12
+      // when our utxos count is below this threshold, we will use more aggressive confirmation targets in force-close scenarios
+      safe-utxos-threshold = 10
     }
 
     feerate-tolerance {
@@ -206,6 +212,10 @@ eclair {
       #    }
       #  }
     ]
+
+    // if false, the commitment transaction will not be fee-bumped when we have no htlcs to claim (used in force-close scenario)
+    // *do not change this unless you know what you are doing*
+    spend-anchor-without-htlcs = true
 
     close-on-offline-feerate-mismatch = true // do not change this unless you know what you are doing
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -439,6 +439,7 @@ object NodeParams extends Logging {
       onChainFeeConf = OnChainFeeConf(
         feeTargets = feeTargets,
         feeEstimator = feeEstimator,
+        spendAnchorWithoutHtlcs = config.getBoolean("on-chain-fees.spend-anchor-without-htlcs"),
         closeOnOfflineMismatch = config.getBoolean("on-chain-fees.close-on-offline-feerate-mismatch"),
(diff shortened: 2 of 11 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -221,6 +229,8 @@ eclair {
 
   autoprobe-count = 0 // number of parallel tasks that send test payments to detect invalid channels
 
+  // TODO: add disable anchor-spend on commit txs without htlcs (false by default)
```
pm47, line 232: Did you forget that?

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -332,6 +332,7 @@ object NodeParams extends Logging {
       fundingBlockTarget = config.getInt("on-chain-fees.target-blocks.funding"),
       commitmentBlockTarget = config.getInt("on-chain-fees.target-blocks.commitment"),
       commitmentWithoutHtlcsBlockTarget = config.getInt("on-chain-fees.target-blocks.commitment-without-htlcs"),
+      disableCommitmentWithoutHtlcsFeeBumping = config.getBoolean("on-chain-fees.target-blocks.disable-commitment-without-htlcs-fee-bumping"),
```
pm47, line 335: I would avoid double-negative, how about `on-chain-fees.target-blocks.spend-anchors-when-no-htlcs` or `on-chain-fees.target-blocks.always-spend-anchors`?

eclair-core/src/main/resources/reference.conf
t-bast: Indeed, thanks!

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
t-bast: Agreed, I really was in a bad variable-naming day :sweat_smile: 
Done in https://github.com/ACINQ/eclair/pull/2246/commits/e2ddd111b205347440b7cf9f7aeec5eb85d461a6

eclair-core/src/main/resources/reference.conf
```diff
@@ -163,12 +163,20 @@ eclair {
...
-      claim-main = 12                   // target for the claim main transaction (tx that spends main channel output back to wallet)
-      safe-utxos-threshold = 10         // when our utxos count is below this threshold, we will use more aggressive confirmation targets in force-close scenarios
+      // target for the funding transaction
+      funding = 6
+      // target for the commitment transaction (used in force-close scenario) *do not change this unless you know what you are doing*
+      commitment = 2
+      // target for the commitment transaction when we have no htlcs to claim (used in force-close scenario) *do not change this unless you know what you are doing*
+      commitment-without-htlcs = 12
+      // if false, the commitment transaction will not be fee-bumped when we have no htlcs to claim (used in force-close scenario) *do not change this unless you know what you are doing*
+      spend-anchor-without-htlcs = true
```
pm47, line 173: Nit: I would have put that in the upper section, so that `target-blocks` stays homogeneous.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -687,16 +687,21 @@ object Helpers {
...
+            withTxGenerationLog("local-anchor") {
+              Transactions.makeClaimLocalAnchorOutputTx(tx, localFundingPubKey, confirmCommitBefore)
+            },
+            withTxGenerationLog("remote-anchor") {
+              Transactions.makeClaimRemoteAnchorOutputTx(tx, commitments.remoteParams.fundingPubKey)
+            }
+          ).flatten
+        } else {
+          Nil
+        }
```
pm47, line 704: Nit:

```suggestion
        val spendAnchors = htlcTxs.nonEmpty || feeTargets.spendAnchorWithoutHtlcs
        val claimAnchorTxs: List[ClaimAnchorOutputTx] = if (spendAnchors) {
          // If we don't have pending HTLCs, we don't have funds at risk, so we can aim for a slower confirmation.
          val confirmCommitBefore = htlcTxs.values.flatten.map(htlcTx => htlcTx.confirmBefore).minOption.getOrElse(currentBlockHeight + feeTargets.commitmentWithoutHtlcsBlockTarget)
          List(
            withTxGenerationLog("local-anchor") {
              Transactions.makeClaimLocalAnchorOutputTx(tx, localFundingPubKey, confirmCommitBefore)
            },
            withTxGenerationLog("remote-anchor") {
              Transactions.makeClaimRemoteAnchorOutputTx(tx, commitments.remoteParams.fundingPubKey)
            }
          ).flatten
        } else {
          Nil
        }
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/2246/commits/82e1f5e01819353dd88114d16bfd32d6f6e03398

eclair-core/src/main/resources/reference.conf
t-bast: That makes more sense indeed, done in https://github.com/ACINQ/eclair/pull/2246/commits/82e1f5e01819353dd88114d16bfd32d6f6e03398

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: I'm a positive person 😆

pm47 (approved):

PR #2251, opened 2022-04-28
Send additional commitment sigs with various feerates

pm47:
We add a new TLV-compatible field to `funding_signed`/`commit_sig` and send additional commitment signatures for a set of feerates (1, 2, 5, 10 sat/B) (only when there are no htlcs).

We also gracefully handle the case where remote decides to use one of those additional signatures to unilaterally close the channel. This creates yet another remote-close scenario, named _custom_, on top of the existing with _current_/_next_/_future_. 

Note that:
- since we use static remote key, there isn't anything to do in case of a custom close, since there are no htlcs
- codecs haven't been updated, there is no need.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
@@ -234,6 +234,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
               closing.localCommitPublished.foreach(doPublish)
               closing.remoteCommitPublished.foreach(doPublish)
               closing.nextRemoteCommitPublished.foreach(doPublish)
+              closing.customRemoteCommitPublished.values.foreach(doPublish)
               closing.revokedCommitPublished.foreach(doPublish)
               closing.futureRemoteCommitPublished.foreach(doPublish)
 
@@ -1277,6 +1278,9 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
       } else if (d.nextRemoteCommitPublished.map(_.commitTx.txid).contains(tx.txid)) {
         // this is because WatchSpent watches never expire and we are notified multiple times
         stay
+      } else if (d.customRemoteCommitPublished.values.exists(_.commitTx.txid == tx.txid)) {
+        // this is because WatchSpent watches never expire and we are notified multiple times
+        stay()
       } else if (d.futureRemoteCommitPublished.map(_.commitTx.txid).contains(tx.txid)) {
         // this is because WatchSpent watches never expire and we are notified multiple times
         stay
@@ -1326,6 +1330,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
         localCommitPublished = d.localCommitPublished.map(Closing.updateLocalCommitPublished(_, tx)),
         remoteCommitPublished = d.remoteCommitPublished.map(Closing.updateRemoteCommitPublished(_, tx)),
         nextRemoteCommitPublished = d.nextRemoteCommitPublished.map(Closing.updateRemoteCommitPublished(_, tx)),
+        customRemoteCommitPublished = d.customRemoteCommitPublished.mapValues(Closing.updateRemoteCommitPublished(_, tx)),
         futureRemoteCommitPublished = d.futureRemoteCommitPublished.map(Closing.updateRemoteCommitPublished(_, tx)),
         revokedCommitPublished = d.revokedCommitPublished.map(Closing.updateRevokedCommitPublished(_, tx))
       )
@@ -2223,6 +2228,22 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     goto(CLOSING) using nextData storing() calling (doPublish(remoteCommitPublished))
   }
 
+  def handleRemoteSpentCustom(commitTx: Transaction, d: HasCommitments, customRemoteCommit: RemoteCommit) = {
+    log.warning(s"they published a custom commitment with feerate={} sat/kw ({} sat/B), uin txid=${commitTx.txid}", customRemoteCommit.spec.feeratePerKw, feerateKw2Byte(customRemoteCommit.spec.feeratePerKw))
+    require(commitTx.txid == customRemoteCommit.txid, "txid mismatch")
+
+    val remoteCommitPublished = Helpers.Closing.claimRemoteCommitTxOutputs(keyManager, d.commitments, customRemoteCommit, commitTx, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)
+
+    val nextData = d match {
+      case closing: DATA_CLOSING => closing.copy(customRemoteCommitPublished = closing.customRemoteCommitPublished + (customRemoteCommit -> remoteCommitPublished))
+      case negotiating: DATA_NEGOTIATING => DATA_CLOSING(d.commitments, fundingTx = None, waitingSince = now, negotiating.closingTxProposed.flatten.map(_.unsignedTx), customRemoteCommitPublished = Map(customRemoteCommit -> remoteCommitPublished))
+      case waitForFundingConfirmed: DATA_WAIT_FOR_FUNDING_CONFIRMED => DATA_CLOSING(d.commitments, fundingTx = waitForFundingConfirmed.fundingTx, waitingSince = now, mutualCloseProposed = Nil, customRemoteCommitPublished = Map(customRemoteCommit -> remoteCommitPublished))
+      case _ => DATA_CLOSING(d.commitments, fundingTx = None, waitingSince = now, mutualCloseProposed = Nil, customRemoteCommitPublished = Map(customRemoteCommit -> remoteCommitPublished))
+    }
+
+    goto(CLOSING) using nextData storing() calling (doPublish(remoteCommitPublished))
+  }
+
   def doPublish(remoteCommitPublished: RemoteCommitPublished): Unit = {
     import remoteCommitPublished._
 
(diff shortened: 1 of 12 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -65,6 +66,13 @@ object LightningMessageCodecs {
   // we have limited space for backup, largest message is commit_sig with 30 htlcs in each direction: 65535B - (32B + 64B + 2*30*64B) = 61599B ~= 60000B
   val channeldataoptional: Codec[Option[ByteVector]] = choice(optional(magic, limitedSizeBytes(60000, variableSizeBytesLong(varintoverflow, bytes))), provide(Option.empty[ByteVector]))
 
+  // TODO: same hack as above, we use a carefully chosen magic to ensure compatibility with TLV encoding
```
t-bast, line 69: I'm really going to enjoy it when we delete that branch :sweat_smile:

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
```diff
@@ -269,6 +269,7 @@ object ChannelCodecs extends Logging {
       ("localCommitPublished" | optional(bool8, localCommitPublishedCodec)) ::
       ("remoteCommitPublished" | optional(bool8, remoteCommitPublishedCodec)) ::
       ("nextRemoteCommitPublished" | optional(bool8, remoteCommitPublishedCodec)) ::
+      ("customRemoteCommitPublished" | provide(Map.empty[RemoteCommit, RemoteCommitPublished])) ::
```
t-bast, line 272: The trick is that when we restart, we forget everything about the custom remote commit, but that's not an issue because:

1. We don't have anything to claim on our side (thanks to `static_remotekey`)
2. We will re-create it automatically because we will watch the funding output being spent

Is that correct? That probably deserves a comment, it's quite subtle.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -424,6 +425,9 @@ object Helpers {
           Some(CurrentRemoteClose(closing.commitments.remoteCommit, closing.remoteCommitPublished.get))
         case _ if closing.nextRemoteCommitPublished.exists(isRemoteCommitConfirmed) =>
           Some(NextRemoteClose(closing.commitments.remoteNextCommitInfo.left.get.nextRemoteCommit, closing.nextRemoteCommitPublished.get))
+        case _ if closing.customRemoteCommitPublished.values.exists(isRemoteCommitConfirmed) =>
+          val (customRemoteCommit, customRemoteCommitPublished) = closing.customRemoteCommitPublished.find(kv => isRemoteCommitConfirmed(kv._2)).get
+          Some(CustomRemoteClose(customRemoteCommit, customRemoteCommitPublished))
```
t-bast, line 430: Just to make sure I understand: we will never run into that code, right?
Since the codecs return an empty map on `INPUT_RESTORED`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -464,6 +466,22 @@ object Commitments {
 
   def revocationHash(seed: ByteVector32, index: Long): ByteVector32 = Crypto.sha256(revocationPreimage(seed, index))
 
+  /** /!\ WARNING : never remove a feerate from the list, because we will only try thosecan only add more */
+  val customRemoteSigsFeerates: Seq[Long] = List(1 sat, 2 sat, 5 sat, 10 sat).map(satPerByte => feerateByte2Kw(satPerByte.toLong))
+
+  def computeCustomRemoteSigs(commitments: Commitments, keyManager: KeyManager, remoteCommitIndex: Long, spec: CommitmentSpec, remotePerCommitmentPoint: PublicKey): Seq[CustomRemoteSig] = {
+    import commitments._
+    if (spec.htlcs.isEmpty) { // TODO: we could use htlcTxs.isEmpty to ignore trimmed htlcs
```
t-bast, line 474: I think we should stick to `spec.htlcs.isEmpty` which is simpler, otherwise there are multiple places where we'd need to match the logic and it could become non-trivial (in `remoteSpentOther` for example).

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -234,6 +234,7 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
               closing.localCommitPublished.foreach(doPublish)
               closing.remoteCommitPublished.foreach(doPublish)
               closing.nextRemoteCommitPublished.foreach(doPublish)
+              closing.customRemoteCommitPublished.values.foreach(doPublish)
```
t-bast, line 237: We should have this line of code for consistency, it will never be used, right (since we're not persisting the custom remote commits)?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
```diff
@@ -2221,6 +2228,22 @@ class Channel(val nodeParams: NodeParams, val wallet: EclairWallet, remoteNodeId
     goto(CLOSING) using nextData storing() calling (doPublish(remoteCommitPublished))
   }
 
+  def handleRemoteSpentCustom(commitTx: Transaction, d: HasCommitments, customRemoteCommit: RemoteCommit) = {
+    log.warning(s"they published a custom commitment with feerate={} sat/kw ({} sat/B), uin txid=${commitTx.txid}", customRemoteCommit.spec.feeratePerKw, feerateKw2Byte(customRemoteCommit.spec.feeratePerKw))
```
t-bast, line 2232:
```suggestion
    log.warning(s"they published a custom commitment with feerate={} sat/kw ({} sat/B), in txid=${commitTx.txid}", customRemoteCommit.spec.feeratePerKw, feerateKw2Byte(customRemoteCommit.spec.feeratePerKw))
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
pm47: FWIW, we have TLV support on this branch, but going with magic was so much easier 🤷‍♂️

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: That's correct, will add a comment.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: We do call this method in the `CLOSING` state. However the part in the `INPUT_RESTORED` handler could indeed be removed but I put it there for consistency with the other closing types.

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
```diff
@@ -120,7 +128,8 @@ object LightningMessageCodecs {
   val fundingSignedCodec: Codec[FundingSigned] = (
     ("channelId" | bytes32) ::
       ("signature" | bytes64) ::
-      ("channelData" | channeldataoptional)).as[FundingSigned]
+      ("channelData" | channeldataoptional) ::
+      ("customRemoteSigs" | customremotesigs)).as[FundingSigned]
```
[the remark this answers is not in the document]
pm47: I did that on the branch that receives those messages.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: Especially if some trimmed htlcs become non-trimmed due to lower feerates 😓

eclair-core/src/main/scala/fr/acinq/eclair/channel/Channel.scala
pm47: Yep!

eclair-core/src/main/scala/fr/acinq/eclair/wire/ChannelCodecs.scala
pm47: Done with 794c75fb8.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -464,6 +466,22 @@ object Commitments {
 
   def revocationHash(seed: ByteVector32, index: Long): ByteVector32 = Crypto.sha256(revocationPreimage(seed, index))
 
+  /** /!\ WARNING : never remove a feerate from the list, because we will only try thosecan only add more */
```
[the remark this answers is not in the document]
t-bast: It's just a comment at this point, we can easily change it later: if we ever want to remove something from this list, then we'll implement some logic to make up for it?

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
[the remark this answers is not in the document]
t-bast: But shouldn't we:

- have a unit test that writes these messages here
- have a unit test that reads them on the LSP side, with the same hard-coded hex?

That would ensure we're completely consistent, because the part where we read it probably doesn't use the magic but rather a tlv?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: Definitely, even though what I think we would do is to consider htlcs trimmed at the *current* feerate, regardless of what *custom* feerate we sign (this way we're sure we're always signing a commit tx without htlcs at all feerates). But that is more subtle so it's more risky!

eclair-core/src/main/scala/fr/acinq/eclair/wire/LightningMessageCodecs.scala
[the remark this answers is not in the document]
pm47: That's what I did (serialize on _this_ branch, deserialize on _that_ branch), but it's not enough: a test is still required here for non-reg.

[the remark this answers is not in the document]
t-bast: I verified the test cases, the hex does look like what we expect (and the end should correctly parse as a tlv stream) :+1: 
Were you able to test them against the server code as well?

t-bast (approved):

PR #2253, opened 2022-05-02
Relay blinded payments

thomash-acinq:
Allow relaying blinded payments (https://github.com/lightning/bolts/pull/765), alternative to #2154 built on top of #2238


thomash-acinq:
Rebased, just waiting for the spec to be approved now.

thomash-acinq:
Rebased on master

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
```diff
@@ -66,26 +118,73 @@ object RouteBlindingEncryptedDataCodecs {
   private val outgoingNodeId: Codec[OutgoingNodeId] = (("length" | constant(hex"21")) :: ("node_id" | publicKey)).as[OutgoingNodeId]
   private val pathId: Codec[PathId] = variableSizeBytesLong(varintoverflow, "path_id" | bytes).as[PathId]
   private val nextBlinding: Codec[NextBlinding] = (("length" | constant(hex"21")) :: ("blinding" | publicKey)).as[NextBlinding]
+  private val paymentRelay: Codec[PaymentRelay] = (
+    ("length" | constant(hex"0a")) ::
+      ("fee_base_msat" | millisatoshi32) ::
+      ("fee_proportional_millionths" | uint32) ::
+      ("cltv_expiry_delta" | cltvExpiryDelta)).as[PaymentRelay]
```
t-bast, line 125: The codec for `payment_relay` has been changed based on feedback by Rusty, the spec says it should be:

```
    1. type: 10 (`payment_relay`)
    2. data:
        * [`u16`:`cltv_expiry_delta`]
        * [`u32`:`fee_proportional_millionths`]
        * [`tu32`:`fee_base_msat`]
```

The reason for that is to be more future-proof in case we want to remove the base fee in the future.
Can you also add a test to verify we correctly encode/decode a base fee of 0 msat (which shouldn't use any byte)?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -42,16 +42,16 @@ object IncomingPaymentPacket {
 
   // @formatter:off
   /** We are the final recipient. */
-  case class FinalPacket(add: UpdateAddHtlc, payload: PaymentOnion.FinalPayload) extends IncomingPaymentPacket
+  case class FinalPacket(add: UpdateAddHtlc, payload: PaymentOnion.FinalTlvPayload) extends IncomingPaymentPacket
```
t-bast, line 45: This is a temporary change, right? When we add support for receiving blinded payments, we'll need `FinalPacket` to work for both blinded and non-blinded payloads somehow.

thomash-acinq: Yes.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
thomash-acinq: Done, there was already a test with a base fee of 0 and it works as expected, taking 0 byte.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/RouteBlinding.scala
```diff
@@ -118,11 +119,10 @@ object RouteBlindingEncryptedDataCodecs {
...
   private val pathId: Codec[PathId] = variableSizeBytesLong(varintoverflow, "path_id" | bytes).as[PathId]
   private val nextBlinding: Codec[NextBlinding] = (("length" | constant(hex"21")) :: ("blinding" | publicKey)).as[NextBlinding]
-  private val paymentRelay: Codec[PaymentRelay] = (
-    ("length" | constant(hex"0a")) ::
-      ("fee_base_msat" | millisatoshi32) ::
+  private val paymentRelay: Codec[PaymentRelay] = variableSizeBytesLong(varintoverflow,
+    ("cltv_expiry_delta" | cltvExpiryDelta) ::
       ("fee_proportional_millionths" | uint32) ::
-      ("cltv_expiry_delta" | cltvExpiryDelta)).as[PaymentRelay]
+      ("fee_base_msat" | tu32.xmap[MilliSatoshi](l => MilliSatoshi(l), m => m.toLong))).as[PaymentRelay]
```
t-bast, line 125: nit: we have a `tmillisatoshi` codec in `TlvCodecs.scala`, it's probably worth adding a `tmillisatoshi32` codec there?

thomash-acinq: Done

t-bast (approved):

PR #2255, opened 2022-05-03
Implement interactive-tx construction

t-bast:
We build the funding transaction based on the interactive-tx protocol. We currently stop before exchanging signatures. Future commits will add the signature exchange and complete the dual funding flow.

NB: this builds on top of https://github.com/ACINQ/eclair/pull/2247


t-bast:
Closing in favor of a better alternative (coming soon).

PR #2258, opened 2022-05-05
Improve wallet double-spend detection

t-bast:
Our mechanism to detect double-spending wasn't correctly taking into account unconfirmed inputs. This was only used in single-funder scenarios so it could only be an issue in rare edge cases, where it would not lead to any loss of funds as we keep commit tx data in our DB even for closed channels.

The second commit is a refactoring that introduces an `isTransactionOutputSpent` function, let me know if you like it, otherwise we can just keep the first commit alone. In the future, once (if) we have https://github.com/bitcoin/bitcoin/pull/24539 we can update `isTransactionOutputSpent` and `doubleSpent` with a `minDepth` parameter to make them reorg-safe(r).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BitcoinCoreClient.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BitcoinCoreClient.scala
@@ -17,8 +17,8 @@
 package fr.acinq.eclair.blockchain.bitcoind.rpc
 
 import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
-import fr.acinq.bitcoin.{Bech32, Block}
 import fr.acinq.bitcoin.scalacompat._
+import fr.acinq.bitcoin.{Bech32, Block}
 import fr.acinq.eclair.ShortChannelId.coordinates
 import fr.acinq.eclair.blockchain.OnChainWallet
 import fr.acinq.eclair.blockchain.OnChainWallet.{MakeFundingTxResponse, OnChainBalance}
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/bitcoind/rpc/BitcoinCoreClient.scala
```diff
@@ -95,11 +95,41 @@ class BitcoinCoreClient(val rpcClient: BitcoinJsonRPCClient) extends OnChainWall
...
       json <- rpcClient.invoke("gettxout", txid, outputIndex, includeMempool)
     } yield json != JNull
 
+  /**
+   * Return true if this output has already been spent by a confirmed transaction.
+   * Note that a reorg may invalidate the result of this function and make a spent output spendable again.
+   */
+  def isTransactionOutputSpent(txid: ByteVector32, outputIndex: Int)(implicit ec: ExecutionContext): Future[Boolean] = {
+    getTxConfirmations(txid).flatMap {
+      case Some(confirmations) if confirmations > 0 =>
```
sstone, line 116: I'm not sure this adds anything since later we'll check if `txid:outputIndex` is a confirmed UTXO ?
And the test passes even with the old version of `doubleSpent()`

t-bast: > And the test passes even with the old version of doubleSpent()

`detect if tx has been double-spent` passes with the old version of `doubleSpent()` (because it has been refactored but not changed to highlight the bug) but `detect if tx has been double-spent (with unconfirmed inputs)` does *not* pass with the old version of `doubleSpent()`.

> I'm not sure this adds anything since later we'll check if txid:outputIndex is a confirmed UTXO ?

What do you mean? Where do we check that "txid:outputIndex is a confirmed UTXO"? It is absolutely necessary, this function by definition checks "if this output has already been spent by a confirmed transaction."

Without this check this function would return `true` for unconfirmed inputs, which is incorrect.

sstone: Ok so it seems that I was completely wrong, misunderstood the PR was and somehow did not run the tests correctly :), nice fix!

sstone (approved):

PR #2261, opened 2022-05-09
Exchange dual funding sigs and wait for channel confirmation

t-bast:
We exchange signatures for the funding tx and commit tx, and wait for the funding tx to confirm.

Note that there is one main deviation from the single-funder case: while we're waiting for a funding tx to confirm, we never force-close, because we still don't know which version of the funding tx will confirm and our closing code doesn't handle multiple local commitments (yet). It's ok, we can just wait for one of the funding txs to confirm or double-spend them all (currently outside of eclair, in the future we could add an RPC for that if necessary) to unblock the channel.

NB: this builds on top of #2255


t-bast:
Closing in favor of a better alternative (coming soon).

PR #2262, opened 2022-05-10
Fix flaky replaceable tx publisher tests

t-bast:
There was a race condition when testing fee-bumping: we could emit the new block event before the underlying actor (`MempoolTxMonitor`) registered to this event type.

This is fixed by emitting the `TransactionPublished` event after registering to the new block events, and waiting for that in tests before publishing a new block event.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
@@ -101,7 +101,6 @@ private class MempoolTxMonitor(nodeParams: NodeParams,
     Behaviors.receiveMessagePartial {
       case PublishOk =>
         log.debug("txid={} was successfully published, waiting for confirmation...", cmd.tx.txid)
-        context.system.eventStream ! EventStream.Publish(TransactionPublished(txPublishContext.channelId_opt.getOrElse(ByteVector32.Zeroes), txPublishContext.remoteNodeId, cmd.tx, cmd.fee, cmd.desc))
         waitForConfirmation()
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/MempoolTxMonitor.scala
```diff
@@ -134,6 +133,7 @@ private class MempoolTxMonitor(nodeParams: NodeParams,
   def waitForConfirmation(): Behavior[Command] = {
     val messageAdapter = context.messageAdapter[CurrentBlockHeight](cbc => WrappedCurrentBlockHeight(cbc.blockHeight))
     context.system.eventStream ! EventStream.Subscribe(messageAdapter)
+    context.system.eventStream ! EventStream.Publish(TransactionPublished(txPublishContext.channelId_opt.getOrElse(ByteVector32.Zeroes), txPublishContext.remoteNodeId, cmd.tx, cmd.fee, cmd.desc))
```
pm47, line 136: Doesn't this make us emit a `TransactionPublished` with each passing block?

t-bast: I don't think so, what makes you say that?
This code is only run when we explicitly call `waitForConfirmation()`, which we do in only one place.
Then we simply send ourselves a `CheckTxConfirmations` message at each block.

pm47: I read to fast, thought we were in the `case WrappedCurrentBlockHeight(currentBlockHeight)` handler.

PR #2264, opened 2022-05-11
Introduce `ChannelRelayParams` in the graph

pm47:
We have two sources for channel routing parameters:
- channel updates
- routing hints in invoices

Instead of generating fake `channel_update`s from routing hints, we
define a `ChannelRelayParams` that can built from announcement or
routing hint.

This is cleaner but also is a first step to decorrelate the identifier
of a channel in our graph, from whatever identifies the source.


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -393,17 +409,52 @@ object Router {
...
+   * @param source         source for the channel parameters.
    */
-  case class ChannelHop(nodeId: PublicKey, nextNodeId: PublicKey, lastUpdate: ChannelUpdate) extends Hop {
-    override lazy val cltvExpiryDelta: CltvExpiryDelta = lastUpdate.cltvExpiryDelta
-
-    override def fee(amount: MilliSatoshi): MilliSatoshi = nodeFee(lastUpdate, amount)
+  case class ChannelHop(shortChannelId: ShortChannelId, nodeId: PublicKey, nextNodeId: PublicKey, source: ChannelSource) extends Hop {
+    // @formatter:off
+    override def cltvExpiryDelta: CltvExpiryDelta = source.cltvExpiryDelta
+    override def fee(amount: MilliSatoshi): MilliSatoshi = source.fee(amount)
```
pm47, line 456: I went back and forth several times between this model and the following, where `ChannelSource` is a simple data container:

````scala
sealed trait ChannelSource
object {
  case class Announcement(channelUpdate: ChannelUpdate) extends ChannelSource
  case class Hint(extraHop: Bolt11Invoice.ExtraHop, htlcMaximum: MilliSatoshi) extends ChannelSource
}

case class ChannelHop(shortChannelId: ShortChannelId, nodeId: PublicKey, nextNodeId: PublicKey, source: ChannelSource) extends Hop {
    override def cltvExpiryDelta: CltvExpiryDelta = source match {
      case s: ChannelSource.Announcement => s.channelUpdate.cltvExpiryDelta
      case s: ChannelSource.Hint => s.extraHop.cltvExpiryDelta
    }
    // ...
````

The model I retained more cleanly defines what a `ChannelSource` is and better separate concerns, but it's more verbose and calling `channelHop.source.htlcMinimum`  is a bit ugly. :-/ I think it's because we have to conflicting abstractions: Channel/Node and Annoucement/Hint

File: eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala [hunk 2]
```diff
@@ -423,27 +423,47 @@ object Graph {
      * Representation of an edge of the graph
      *
      * @param desc        channel description
-     * @param update      channel info
+     * @param source      source of the channel parameters: can be a channel_update or hints from an invoice // TODO: should we do a class hierarchy instead: KnownEdge/AssistedEdge?
```
thomash-acinq, line 426: No, I think it's better if the source details don't leak. I don't see any benefit of also splitting GraphEdge in KnownEdge/AssistedEdge.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -331,7 +331,7 @@ object Graph {
         val totalHopsCost = prev.virtualFees + hopCost
         // If the edge was added by the invoice, it is assumed that it can route the payment.
         // If we know the balance of the channel, then we will check separately that it can relay the payment.
-        val successProbability = if (edge.update.chainHash == ByteVector32.Zeroes || edge.balance_opt.nonEmpty) 1.0 else 1.0 - prev.amount.toLong.toDouble / edge.capacity.toMilliSatoshi.toLong.toDouble
+        val successProbability = if (edge.source.isInstanceOf[ChannelSource.Hint] || edge.balance_opt.nonEmpty) 1.0 else 1.0 - prev.amount.toLong.toDouble / edge.capacity.toMilliSatoshi.toLong.toDouble
```
thomash-acinq, line 334: For hints, you already set balance_opt so you don't need this anymore.
```suggestion
        val successProbability = if (edge.balance_opt.nonEmpty) 1.0 else 1.0 - prev.amount.toLong.toDouble / edge.capacity.toMilliSatoshi.toLong.toDouble
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
pm47: Ok, btw my plan is to eventually get rid of `ChannelDesc` and replace it by a `case class EdgeId(a: PublicKey, b: PublicKey)` like you suggested.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
pm47: Why weren't we able to do so before this PR?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -393,17 +409,61 @@ object Router {
...
+      lazy val cltvExpiryDelta: CltvExpiryDelta = extraHop.cltvExpiryDelta
+      lazy val relayFees: Relayer.RelayFees = RelayFees(extraHop.feeBase, extraHop.feeProportionalMillionths)
+      lazy val htlcMinimum: MilliSatoshi = 0 msat
+      lazy val htlcMaximum_opt: Option[MilliSatoshi] = Some(htlcMaximum)
+    }
+
+    def areSame(source1: ChannelSource, source2: ChannelSource): Boolean =
+      source1.relayFees == source2.relayFees &&
+        source1.cltvExpiryDelta == source2.cltvExpiryDelta &&
+        source1.htlcMinimum == source2.htlcMinimum // TODO: not comparing htlcMax otherwise network/invoice sources may never match
```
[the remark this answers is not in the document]
pm47: Right, this will be used in a later PR. Removed in 56b8744f1.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
thomash-acinq: You're right, it could have already been done before.
Before calling `addEdgeWeight`, we've already checked that the amount is less than `edge.balance_opt` if it is defined. So if `edge.balance_opt` is defined here, it means that we know the balance and it is high enough. `edge.source.isInstanceOf[ChannelSource.Hint]` was a bit of a hack, the thing we care about is if the `balance_opt` is defined an larger than the amount.

t-bast (commented):
Just a quick pass on the high-level design for now.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: > I think the issue around naming is because you're trying to do two different things at the same time:
>
>- factorize channel relay parameters
>- label them by different sources

Exactly. Also, there is kind of a competition between `ChannelHop` and `GraphEdge`, which are very similar and both take a `source` as argument. For `GraphEdge` it makes more sense to focus on the source because we may need to update it, whereas for `ChannelHop` we mainly want the parameters to build the onion. Maybe we should revisit after we have experimented with the new edge model suggested by @thomash-acinq. The idea is to index edges by (node1, node2) instead of channel id. That means each edge would contain a set of channels, and we would somehow combine all those into one.

> What do you think about the following model, which explicitly separates the two purposes?

I would prefer keeping the original data, and I don't like the use of `object`s as enums .

I went in b673910da with a simple rename `ChannelSource`->`ChannelRelayParams` which has the advantage of not creating potential issues with further changes that I have in the pipe:
- it now focuses on the "params" aspect rather than the "source"
- the type name still makes the source obvious, : `ChannelRelayParams.FromAnnouncement`/`ChannelRelayParams.FromHint`
- `channelHop.params.htlcMinimum` looks better than `channelHop.source.htlcMinimum`

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -423,27 +422,47 @@ object Graph {
...
      * @param desc        channel description
-     * @param update      channel info
+     * @param params      source of the channel parameters: can be a channel_update or hints from an invoice
      * @param capacity    channel capacity
      * @param balance_opt (optional) available balance that can be sent through this edge
      */
-    case class GraphEdge(desc: ChannelDesc, update: ChannelUpdate, capacity: Satoshi, balance_opt: Option[MilliSatoshi]) {
+    case class GraphEdge private(desc: ChannelDesc, params: ChannelRelayParams, capacity: Satoshi, balance_opt: Option[MilliSatoshi]) {
+
+      def this(u: ChannelUpdate, pc: PublicChannel) = this(
```
t-bast, line 431: We usually define constructors in the companion object (`apply`), what's the scala best practice here nowadays? I like the companion object approach, this way we only keep "business logic" inside the case class.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -47,9 +47,7 @@ object RouteCalculation {
       implicit val sender: ActorRef = ctx.self // necessary to preserve origin when sending messages to other actors
 
       val assistedChannels: Map[ShortChannelId, AssistedChannel] = fr.assistedRoutes.flatMap(toAssistedChannels(_, fr.route.targetNodeId, fr.amount)).toMap
-      val extraEdges = assistedChannels.values.map(ac =>
-        GraphEdge(ChannelDesc(ac.extraHop.shortChannelId, ac.extraHop.nodeId, ac.nextNodeId), toFakeUpdate(ac.extraHop, ac.htlcMaximum), htlcMaxToCapacity(ac.htlcMaximum), Some(ac.htlcMaximum))
-      ).toSet
+      val extraEdges = assistedChannels.values.map(ac => new GraphEdge(ac)).toSet
```
t-bast, line 50: This does feel better!

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -334,7 +320,7 @@ object RouteCalculation {
       val directChannels = g.getEdgesBetween(localNodeId, targetNodeId).collect {
         // We should always have balance information available for local channels.
         // NB: htlcMinimumMsat is set by our peer and may be 0 msat (even though it's not recommended).
-        case GraphEdge(_, update, _, Some(balance)) => DirectChannel(balance, balance <= 0.msat || balance < update.htlcMinimumMsat)
+        case GraphEdge(_, source, _, Some(balance)) => DirectChannel(balance, balance <= 0.msat || balance < source.htlcMinimum)
```
t-bast, line 323: nit:
```suggestion
        case GraphEdge(_, params, _, Some(balance)) => DirectChannel(balance, balance <= 0.msat || balance < params.htlcMinimum)
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -374,7 +387,10 @@ object Router {
   }
   // @formatter:on
 
-  case class AssistedChannel(extraHop: ExtraHop, nextNodeId: PublicKey, htlcMaximum: MilliSatoshi)
+  case class AssistedChannel(nextNodeId: PublicKey, source: ChannelRelayParams.FromHint) {
```
t-bast, line 390: nit:
```suggestion
  case class AssistedChannel(nextNodeId: PublicKey, params: ChannelRelayParams.FromHint) {
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -393,17 +409,47 @@ object Router {
     def cltvExpiryDelta: CltvExpiryDelta
   }
 
+  // @formatter:off
+  /** source of the channel routing parameters */
```
t-bast, line 413: The comment is a bit misleading now, maybe just:
```suggestion
  /** Channel routing parameters */
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -393,17 +409,47 @@ object Router {
...
+      lazy val htlcMinimum: MilliSatoshi = channelUpdate.htlcMinimumMsat
+      lazy val htlcMaximum_opt: Option[MilliSatoshi] = channelUpdate.htlcMaximumMsat
+    }
+    /** we learnt about this channel from hints in an invoice */
+    case class FromHint(extraHop: Bolt11Invoice.ExtraHop, htlcMaximum: MilliSatoshi) extends ChannelRelayParams {
+      lazy val cltvExpiryDelta: CltvExpiryDelta = extraHop.cltvExpiryDelta
+      lazy val relayFees: Relayer.RelayFees = RelayFees(extraHop.feeBase, extraHop.feeProportionalMillionths)
+      lazy val htlcMinimum: MilliSatoshi = 0 msat
+      lazy val htlcMaximum_opt: Option[MilliSatoshi] = Some(htlcMaximum)
+    }
```
t-bast, line 436: Are the `lazy` really useful here?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -481,11 +527,11 @@ object Router {
     }
 
     /** This method retrieves the channel update that we used when we built the route. */
-    def getChannelUpdateForNode(nodeId: PublicKey): Option[ChannelUpdate] = hops.find(_.nodeId == nodeId).map(_.lastUpdate)
+    def getChannelUpdateForNode(nodeId: PublicKey): Option[ChannelUpdate] = hops.find(_.nodeId == nodeId).map(_.params).collect { case s: ChannelRelayParams.FromAnnouncement => s.channelUpdate }
```
t-bast, line 530: The only place where we use this function is here: https://github.com/ACINQ/eclair/blob/ee74f10a8d956196dcc62ad297e83769aa7a2a82/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentLifecycle.scala#L245

Your change shows that this was incorrect, we also need to handle the case where we receive a channel update for a channel that was in the invoice routing hints, don't we?

IMO we should remove this method, and instead more explicitly handle all cases in `PaymentLifecycle.scala`. WDYT?

t-bast (commented):
LGTM, mostly nits

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -393,17 +409,47 @@ object Router {
...
+    def fee(amount: MilliSatoshi): MilliSatoshi = nodeFee(relayFees, amount)
+    def htlcMinimum: MilliSatoshi
+    def htlcMaximum_opt: Option[MilliSatoshi]
+  }
+
+  object ChannelRelayParams {
+    /** we learnt about this channel from a channel_update */
+    case class FromAnnouncement(channelUpdate: ChannelUpdate) extends ChannelRelayParams {
+      lazy val cltvExpiryDelta: CltvExpiryDelta = channelUpdate.cltvExpiryDelta
+      lazy val relayFees: Relayer.RelayFees = RelayFees(channelUpdate.feeBaseMsat, channelUpdate.feeProportionalMillionths)
```
thomash-acinq, line 426: You've created `channelUpdate.relayFees`, you should use it here.

eclair-core/src/test/scala/fr/acinq/eclair/TestUtils.scala
```diff
@@ -19,9 +19,11 @@ package fr.acinq.eclair
 import akka.actor.ActorRef
 import akka.event.DiagnosticLoggingAdapter
 import akka.testkit.{TestActor, TestProbe}
+import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
 import fr.acinq.eclair.channel.fsm.Channel
 import fr.acinq.eclair.io.Peer
-import fr.acinq.eclair.wire.protocol.LightningMessage
+import fr.acinq.eclair.router.Router.{ChannelHop, ChannelRelayParams}
+import fr.acinq.eclair.wire.protocol.{ChannelUpdate, LightningMessage}
```
thomash-acinq, line 26: You don't change anything in this file, you shouldn't add imports.

eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycleSpec.scala
```diff
@@ -698,13 +699,13 @@ object MultiPartPaymentLifecycleSpec {
...
-  val hop_ad = ChannelHop(a, d, channelUpdate_ad)
-  val hop_de = ChannelHop(d, e, channelUpdate_de)
+  val hop_ab_1 = ChannelHop(channelUpdate_ab_1.shortChannelId, a, b, ChannelRelayParams.FromAnnouncement(channelUpdate_ab_1))
+  val hop_ab_2 = ChannelHop(channelUpdate_ab_2.shortChannelId, a, b, ChannelRelayParams.FromAnnouncement(channelUpdate_ab_2))
+  val hop_be = ChannelHop(channelUpdate_be.shortChannelId, b, e, ChannelRelayParams.FromAnnouncement(channelUpdate_be))
+  val hop_ac_1 = ChannelHop(channelUpdate_ac_1.shortChannelId, a, c, ChannelRelayParams.FromAnnouncement(channelUpdate_ac_1))
+  val hop_ac_2 = ChannelHop(channelUpdate_ac_2.shortChannelId, a, c, ChannelRelayParams.FromAnnouncement(channelUpdate_ac_2))
+  val hop_ce = ChannelHop(channelUpdate_ce.shortChannelId, c, e, ChannelRelayParams.FromAnnouncement(channelUpdate_ce))
+  val hop_ad = ChannelHop(channelUpdate_ad.shortChannelId, a, d, ChannelRelayParams.FromAnnouncement(channelUpdate_ad))
+  val hop_de = ChannelHop(channelUpdate_de.shortChannelId, d, e, ChannelRelayParams.FromAnnouncement(channelUpdate_de))
```
thomash-acinq, line 709: You could use `channelHopFromUpdate`

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
```diff
@@ -103,7 +103,7 @@ class PaymentLifecycleSpec extends BaseRouterSpec {
     import cfg._
 
     // pre-computed route going from A to D
-    val route = Route(defaultAmountMsat, ChannelHop(a, b, update_ab) :: ChannelHop(b, c, update_bc) :: ChannelHop(c, d, update_cd) :: Nil)
+    val route = Route(defaultAmountMsat, ChannelHop(update_ab.shortChannelId, a, b, ChannelRelayParams.FromAnnouncement(update_ab)) :: ChannelHop(update_bc.shortChannelId, b, c, ChannelRelayParams.FromAnnouncement(update_bc)) :: ChannelHop(update_cd.shortChannelId, c, d, ChannelRelayParams.FromAnnouncement(update_cd)) :: Nil)
```
thomash-acinq, line 106: You could use `channelHopFromUpdate`

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
```diff
@@ -736,21 +736,21 @@ class PaymentLifecycleSpec extends BaseRouterSpec {
...
+      LocalFailure(defaultAmountMsat, ChannelHop(update_ab.shortChannelId, a, b, ChannelRelayParams.FromAnnouncement(update_ab)) :: Nil, ChannelUnavailable(ByteVector32.Zeroes)),
       LocalFailure(defaultAmountMsat, Nil, RouteNotFound)
     )
     val filtered = PaymentFailure.transformForUser(failures)
     val expected = Seq(
       LocalFailure(defaultAmountMsat, Nil, RouteNotFound),
-      RemoteFailure(defaultAmountMsat, ChannelHop(a, b, update_ab) :: Nil, Sphinx.DecryptedFailurePacket(a, TemporaryNodeFailure)),
-      LocalFailure(defaultAmountMsat, ChannelHop(a, b, update_ab) :: Nil, ChannelUnavailable(ByteVector32.Zeroes))
+      RemoteFailure(defaultAmountMsat, ChannelHop(update_ab.shortChannelId, a, b, ChannelRelayParams.FromAnnouncement(update_ab)) :: Nil, Sphinx.DecryptedFailurePacket(a, TemporaryNodeFailure)),
+      LocalFailure(defaultAmountMsat, ChannelHop(update_ab.shortChannelId, a, b, ChannelRelayParams.FromAnnouncement(update_ab)) :: Nil, ChannelUnavailable(ByteVector32.Zeroes))
```
thomash-acinq, line 747: You could use `channelHopFromUpdate`

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: I agree and kind of let this for later. We should be able to rewrite `PaymentLifecycle.handleUpdate` much nicer now that we have streamlined channel sources, but I didn't want to dig an even deeper rabbit hole than the one I was in 😆 

Let's add a TODO and fix later?

t-bast: Right, that works for me :+1:

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
pm47: Done with 6fe269081.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Done with 6fe269081.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Done with 6fe269081.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Probably not, replaced by simple `def` in 6fe269081.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Done with 6fe269081.

eclair-core/src/test/scala/fr/acinq/eclair/payment/MultiPartPaymentLifecycleSpec.scala
pm47: Nice catch, done in 623a71182. I'm not proud of `channelHopFromUpdate`, I want to get rid of it eventually but for now just cleaned up `Router.scala`.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
pm47: Done in 623a71182.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentLifecycleSpec.scala
pm47: Done in 623a71182.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
pm47: Best practice is indeed to use `apply` methods in the companion object. Done in 8fcfebba3.

This gets rid of `new` keyword.

thomash-acinq (approved):

t-bast (approved):
LGTM

PR #2268, opened 2022-05-13
Add simple integration test between `Channel` and `Router` 

pm47:
In this PR we add a basic integration test between `Channel` and `Router` that checks the proper handling of local announcements.

This test found two bugs, fixed in two separate commits.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -19,7 +19,7 @@ package fr.acinq.eclair.router
 import akka.Done
 import akka.actor.typed.scaladsl.adapter.actorRefAdapter
 import akka.actor.{Actor, ActorLogging, ActorRef, Props, Terminated, typed}
-import akka.event.DiagnosticLoggingAdapter
+import akka.event.{DiagnosticLoggingAdapter, EventStream}
 import akka.event.Logging.MDC
 import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
 import fr.acinq.bitcoin.scalacompat.{ByteVector32, Satoshi}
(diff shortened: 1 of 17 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -59,9 +59,12 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
...
 
-  context.system.eventStream.subscribe(self, classOf[LocalChannelUpdate])
-  context.system.eventStream.subscribe(self, classOf[LocalChannelDown])
-  context.system.eventStream.subscribe(self, classOf[AvailableBalanceChanged])
+  // this allows overriding the default eventstream in tests
+  val eventStream = eventStream_opt.getOrElse(context.system.eventStream)
+
+  eventStream.subscribe(self, classOf[LocalChannelUpdate])
+  eventStream.subscribe(self, classOf[LocalChannelDown])
+  eventStream.subscribe(self, classOf[AvailableBalanceChanged])
```
pm47, line 67: This is a bit dirty: we only override the event stream for subscriptions. It's enough for testing, and it was not possible to update all references in `Validation`, `RouteCalculation`, `Syncing` etc. due to how the handlers are implemented. We could migrate them to the same technique as what we used for the channel fsm.

eclair-front/src/test/scala/fr/acinq/eclair/router/FrontRouterSpec.scala
```diff
@@ -149,9 +148,18 @@ class FrontRouterSpec extends TestKit(ActorSystem("test")) with AnyFunSuiteLike
...
+      TestFSMRef[FrontRouter.State, FrontRouter.Data, FrontRouter](new FrontRouter(nodeParams.routerConf, router))
+    }
+    val front2 = {
+      implicit val system: ActorSystem = system2
+      TestFSMRef[FrontRouter.State, FrontRouter.Data, FrontRouter](new FrontRouter(nodeParams.routerConf, router))
+    }
+    val front3 = {
+      implicit val system: ActorSystem = system3
+      TestFSMRef[FrontRouter.State, FrontRouter.Data, FrontRouter](new FrontRouter(nodeParams.routerConf, router))
+    }
```
pm47, line 162: I just realized that this trick (providing a custom `ActorSystem`) would also make it possible to segregate event streams in tests. Should I do that instead of [40cc09f](https://github.com/ACINQ/eclair/pull/2268/commits/40cc09f102a0d0bc4b485061c4ea5e000174e6cf)? I think I should 😅

pm47:
Superseded by #2270.

PR #2269, opened 2022-05-13
Index unannounced channels on `channelId` instead of `shortChannelId` in the router

pm47:
Builds on #2270.

This is the final preparatory step before we can introduce channel aliases, which are arbitrary values of type `ShortChannelId`. We had to move to a stable identifier for private channels. Public channels can keep using the `shortChannelId`, it is stable because it is a "real scid" (and only used after 6 confs) and is useful as an index for channel queries.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -66,18 +66,22 @@ object RouteCalculation {
         case PredefinedChannelRoute(targetNodeId, shortChannelIds) =>
           val (end, hops) = shortChannelIds.foldLeft((localNodeId, Seq.empty[ChannelHop])) {
             case ((currentNode, previousHops), shortChannelId) =>
-              val channelDesc_opt = d.channels.get(shortChannelId).flatMap(c => currentNode match {
-                case c.ann.nodeId1 => Some(ChannelDesc(shortChannelId, c.ann.nodeId1, c.ann.nodeId2))
-                case c.ann.nodeId2 => Some(ChannelDesc(shortChannelId, c.ann.nodeId2, c.ann.nodeId1))
-                case _ => None
-              }).orElse(d.privateChannels.get(shortChannelId).flatMap(c => currentNode match {
-                case c.nodeId1 => Some(ChannelDesc(shortChannelId, c.nodeId1, c.nodeId2))
-                case c.nodeId2 => Some(ChannelDesc(shortChannelId, c.nodeId2, c.nodeId1))
-                case _ => None
-              })).orElse(assistedChannels.get(shortChannelId).flatMap(c => currentNode match {
-                case c.nodeId => Some(ChannelDesc(shortChannelId, c.nodeId, c.nextNodeId))
-                case _ => None
-              }))
+              val channelDesc_opt = d.resolve(shortChannelId) match {
+                case Some(c: PublicChannel) => currentNode match {
+                  case c.nodeId1 => Some(ChannelDesc(shortChannelId, c.nodeId1, c.nodeId2))
+                  case c.nodeId2 => Some(ChannelDesc(shortChannelId, c.nodeId2, c.nodeId1))
+                  case _ => None
+                }
+                case Some(c: PrivateChannel) => currentNode match {
+                  case c.nodeId1 => Some(ChannelDesc(c.shortChannelId, c.nodeId1, c.nodeId2))
+                  case c.nodeId2 => Some(ChannelDesc(c.shortChannelId, c.nodeId2, c.nodeId1))
+                  case _ => None
+                }
+                case None => assistedChannels.get(shortChannelId).flatMap(c => currentNode match {
+                  case c.nodeId => Some(ChannelDesc(shortChannelId, c.nodeId, c.nextNodeId))
+                  case _ => None
+                })
+              }
               channelDesc_opt.flatMap(c => g.getEdge(c)) match {
                 case Some(edge) => (edge.desc.b, previousHops :+ ChannelHop(edge.desc.shortChannelId, edge.desc.a, edge.desc.b, edge.params))
                 case None => (currentNode, previousHops)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -24,6 +24,7 @@ import akka.event.Logging.MDC
 import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
 import fr.acinq.bitcoin.scalacompat.{ByteVector32, Satoshi}
 import fr.acinq.eclair.Logs.LogCategory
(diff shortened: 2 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -339,6 +342,10 @@ object Router {
     update_1_opt.foreach(u => assert(u.channelFlags.isNode1))
     update_2_opt.foreach(u => assert(!u.channelFlags.isNode1))
 
+    val nodeId1: PublicKey = ann.nodeId1
+    val nodeId2: PublicKey = ann.nodeId2
+    def shortChannelId: ShortChannelId = ann.shortChannelId
+    def channelId: ByteVector32 = toLongId(fundingTxid.reverse, outputIndex(ann.shortChannelId))
```
t-bast, line 348: That cannot work with dual-funding, where the `channelId` isn't derived from the funding tx anymore. I think we need to explicitly add it to the arguments of the `PublicChannel` class (it is derived from `open_channel2` and `accept_channel2`'s revocation basepoints).

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -614,11 +621,26 @@ object Router {
                   stash: Stash,
                   rebroadcast: Rebroadcast,
                   awaiting: Map[ChannelAnnouncement, Seq[GossipOrigin]], // note: this is a seq because we want to preserve order: first actor is the one who we need to send a tcp-ack when validation is done
-                  privateChannels: Map[ShortChannelId, PrivateChannel],
+                  privateChannels: Map[ByteVector32, PrivateChannel], // indexed by channel id
+                  resolveScid: Map[ShortChannelId, ByteVector32], // scid to channel_id
```
t-bast, line 625: Shouldn't we make it more explicit that this map is only used for private channels (currently) and unconfirmed channels (in the future)? There's no reason to store the mapping for public, announced channels, it's already in `channels`, right?

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Since it's not _yet_ an issue, I suggest we fix it in #2224, because I'm changing the private->public channel logic.

Don't worry I'm keeping track of all the things that were deferred to #2224 😅

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -614,11 +621,26 @@ object Router {
...
+    def resolve(scid: ShortChannelId): Option[KnownChannel] = {
+      // let's assume this is a real scid
+      channels.get(scid) match {
+        case Some(publicChannel) => Some(publicChannel)
+        case None =>
+          // maybe it's an alias or a real scid
+          resolveScid.get(scid).flatMap(privateChannels.get) match {
+            case Some(privateChannel) => Some(privateChannel)
+            case None => None
+          }
```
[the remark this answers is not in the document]
pm47: It was for symmetry with the line above : `Some(publicChannel)`/`Some(privateChannel)`. I'm not really happy with the `resolve` method.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Done with 1c0f45355, also renamed the field. Added the comment, but it's week, I wish we could enforce that somehow. Well we could by making fields private, but then updating the case class would be horrible. 🤷‍♂️ 

> There's no reason to store the mapping for public, announced channels, it's already in channels, right?

Yes

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -66,18 +66,22 @@ object RouteCalculation {
...
-              }))
+              val channelDesc_opt = d.resolve(shortChannelId) match {
+                case Some(c: PublicChannel) => currentNode match {
+                  case c.nodeId1 => Some(ChannelDesc(shortChannelId, c.nodeId1, c.nodeId2))
+                  case c.nodeId2 => Some(ChannelDesc(shortChannelId, c.nodeId2, c.nodeId1))
+                  case _ => None
+                }
+                case Some(c: PrivateChannel) => currentNode match {
+                  case c.nodeId1 => Some(ChannelDesc(c.shortChannelId, c.nodeId1, c.nodeId2))
+                  case c.nodeId2 => Some(ChannelDesc(c.shortChannelId, c.nodeId2, c.nodeId1))
```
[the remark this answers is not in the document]
pm47: It's because this will be changed to `c.localAlias` in #2224. In the meantime wanted to make it different from the `PublicChannel` case but managed to make it more confusing.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -122,7 +122,8 @@ object Validation {
               ctx.self ! nodeAnn
             }
             // maybe this previously was a local unannounced channel
-            val privateChannel_opt = d0.privateChannels.get(c.shortChannelId)
+            val channelId = toLongId(tx.txid.reverse, outputIndex)
```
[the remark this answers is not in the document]
pm47: We probably could, but https://github.com/ACINQ/eclair/pull/2269#discussion_r876818487.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -432,15 +432,18 @@ object Validation {
...
           case None =>
             // channel isn't announced and we never heard of it (maybe it is a private channel or maybe it is a public channel that doesn't yet have 6 confirmations)
             // let's create a corresponding private channel and process the channel_update
             log.debug("adding unannounced local channel to remote={} shortChannelId={}", lcu.remoteNodeId, lcu.shortChannelId)
-            val pc = PrivateChannel(localNodeId, lcu.remoteNodeId, None, None, ChannelMeta(0 msat, 0 msat)).updateBalances(lcu.commitments)
-            val d1 = d.copy(privateChannels = d.privateChannels + (lcu.shortChannelId -> pc))
+            val pc = PrivateChannel(lcu.shortChannelId, lcu.channelId, localNodeId, lcu.remoteNodeId, None, None, ChannelMeta(0 msat, 0 msat)).updateBalances(lcu.commitments)
+            val d1 = d.copy(
+              privateChannels = d.privateChannels + (lcu.channelId -> pc),
+              resolveScid = d.resolveScid + (lcu.shortChannelId -> lcu.channelId)
```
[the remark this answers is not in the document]
pm47: This is the only place where we _can_ update `resolveScid`, because nowhere else do we have the mapping information.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: Sounds good!

t-bast (approved):
LGTM :+1:

PR #2270, opened 2022-05-13
Add simple integration test between `Channel` and `Router`

pm47:
This is a simpler alternative to #2268.

In this PR we add a basic integration test between `Channel` and `Router` that checks the proper handling of local announcements.

This test found two bugs, fixed in two separate commits.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -19,7 +19,7 @@ package fr.acinq.eclair.router
 import akka.Done
 import akka.actor.typed.scaladsl.adapter.actorRefAdapter
 import akka.actor.{Actor, ActorLogging, ActorRef, Props, Terminated, typed}
-import akka.event.DiagnosticLoggingAdapter
+import akka.event.{DiagnosticLoggingAdapter, EventStream}
 import akka.event.Logging.MDC
 import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
 import fr.acinq.bitcoin.scalacompat.{ByteVector32, Satoshi}
@@ -50,7 +50,7 @@ import scala.util.{Random, Try}
 /**
  * Created by PM on 24/05/2016.
  */
-class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Command], initialized: Option[Promise[Done]] = None) extends FSMDiagnosticActorLogging[Router.State, Router.Data] {
+class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Command], initialized: Option[Promise[Done]] = None, eventStream_opt: Option[EventStream] = None) extends FSMDiagnosticActorLogging[Router.State, Router.Data] {
 
   import Router._
 
@@ -59,9 +59,12 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
   // we pass these to helpers classes so that they have the logging context
   implicit def implicitLog: DiagnosticLoggingAdapter = diagLog
 
-  context.system.eventStream.subscribe(self, classOf[LocalChannelUpdate])
-  context.system.eventStream.subscribe(self, classOf[LocalChannelDown])
-  context.system.eventStream.subscribe(self, classOf[AvailableBalanceChanged])
+  // this allows overriding the default eventstream in tests
+  val eventStream = eventStream_opt.getOrElse(context.system.eventStream)
+
+  eventStream.subscribe(self, classOf[LocalChannelUpdate])
+  eventStream.subscribe(self, classOf[LocalChannelDown])
+  eventStream.subscribe(self, classOf[AvailableBalanceChanged])
 
   startTimerWithFixedDelay(TickBroadcast.toString, TickBroadcast, nodeParams.routerConf.routerBroadcastInterval)
   startTimerWithFixedDelay(TickPruneStaleChannels.toString, TickPruneStaleChannels, 1 hour)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -187,14 +187,6 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
       sender() ! updates
       stay()
 
-    case Event(PrintChannelUpdates, d) =>
-      println("public:")
-      d.channels.foreach { case (scid, pc) => println(s"$scid updates=${(pc.update_1_opt.toSeq ++ pc.update_2_opt.toSeq).size}") }
-      println("private:")
-      d.privateChannels.foreach { case (scid, pc) => println(s"$scid updates=${(pc.update_1_opt.toSeq ++ pc.update_2_opt.toSeq).size}") }
-      println("---------------------------------------------------")
-      stay()
-
     case Event(GetRouterData, d) =>
       sender() ! d
       stay()
@@ -566,7 +558,6 @@ object Router {
   case object GetChannels
   case object GetChannelsMap
   case object GetChannelUpdates
-  case object PrintChannelUpdates
   // @formatter:on
 
   // @formatter:off
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -187,6 +187,14 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
       sender() ! updates
       stay()
 
+    case Event(PrintChannelUpdates, d) =>
+      println("public:")
+      d.channels.foreach { case (scid, pc) => println(s"$scid updates=${(pc.update_1_opt.toSeq ++ pc.update_2_opt.toSeq).size}") }
+      println("private:")
+      d.privateChannels.foreach { case (scid, pc) => println(s"$scid updates=${(pc.update_1_opt.toSeq ++ pc.update_2_opt.toSeq).size}") }
+      println("---------------------------------------------------")
+      stay()
+
     case Event(GetRouterData, d) =>
       sender() ! d
       stay()
@@ -558,6 +566,7 @@ object Router {
   case object GetChannels
   case object GetChannelsMap
   case object GetChannelUpdates
+  case object PrintChannelUpdates
   // @formatter:on
 
   // @formatter:off
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -19,7 +19,7 @@ package fr.acinq.eclair.router
 import akka.Done
 import akka.actor.typed.scaladsl.adapter.actorRefAdapter
 import akka.actor.{Actor, ActorLogging, ActorRef, Props, Terminated, typed}
-import akka.event.{DiagnosticLoggingAdapter, EventStream}
+import akka.event.DiagnosticLoggingAdapter
 import akka.event.Logging.MDC
 import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
 import fr.acinq.bitcoin.scalacompat.{ByteVector32, Satoshi}
@@ -50,7 +50,7 @@ import scala.util.{Random, Try}
 /**
  * Created by PM on 24/05/2016.
  */
-class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Command], initialized: Option[Promise[Done]] = None, eventStream_opt: Option[EventStream] = None) extends FSMDiagnosticActorLogging[Router.State, Router.Data] {
+class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Command], initialized: Option[Promise[Done]] = None) extends FSMDiagnosticActorLogging[Router.State, Router.Data] {
 
   import Router._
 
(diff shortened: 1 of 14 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -558,6 +566,7 @@ object Router {
   case object GetChannels
   case object GetChannelsMap
   case object GetChannelUpdates
+  case object PrintChannelUpdates
```
t-bast, line 569: This was probably for your internal tests, remove?

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -145,12 +153,13 @@ object Validation {
...
-          // right after the channel_announcement, channel_updates will be moved from private to public at that time
           val d1 = d0.copy(
             channels = d0.channels + (c.shortChannelId -> pc),
             privateChannels = d0.privateChannels - c.shortChannelId, // we remove fake announcements that we may have made before
-            rebroadcast = d0.rebroadcast.copy(channels = d0.rebroadcast.channels + (c -> d0.awaiting.getOrElse(c, Nil).toSet)), // we also add the newly validated channels to the rebroadcast queue
+            rebroadcast = d0.rebroadcast.copy(
+              channels = d0.rebroadcast.channels + (c -> d0.awaiting.getOrElse(c, Nil).toSet), // we rebroadcast the channel to our peers
+              updates = d0.rebroadcast.updates ++ (pc.update_1_opt.toSet ++ pc.update_2_opt.toSet).map(u => u -> (if (pc.getNodeIdSameSideAs(u) == nodeParams.nodeId) Set[GossipOrigin](LocalGossip) else Set.empty[GossipOrigin])).toMap // those updates are only defined if this was a previously an unannounced local channel, we broadcast them
+            ), // we also add the newly validated channels to the rebroadcast queue
```
t-bast, line 162: I think it's worth extracting the updates to a separate line for readability:

```scala
// those updates are only defined if this was a previously an unannounced local channel, we broadcast them
val previousUpdates = (pc.update_1_opt.toSet ++ pc.update_2_opt.toSet).map(u => u -> (if (pc.getNodeIdSameSideAs(u) == nodeParams.nodeId) Set[GossipOrigin](LocalGossip) else Set.empty[GossipOrigin])).toMap
```

```suggestion
            rebroadcast = d0.rebroadcast.copy(
              channels = d0.rebroadcast.channels + (c -> d0.awaiting.getOrElse(c, Nil).toSet), // we rebroadcast the channel to our peers
              updates = d0.rebroadcast.updates ++ localUpdates
            ), // we also add the newly validated channels to the rebroadcast queue
```

Looking ahead, isn't there a risk here that we'll broadcast an update that contains an alias `short_channel_id` (which doesn't make sense, others won't be able to validate it)? Shouldn't we wait to receive a new `channel_update` with the final `short_channel_id`?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/ChannelStateTestsHelperMethods.scala
```diff
@@ -119,8 +125,10 @@ trait ChannelStateTestsHelperMethods extends TestKitBase {
     val alice2relayer = TestProbe()
     val bob2relayer = TestProbe()
     val channelUpdateListener = TestProbe()
-    system.eventStream.subscribe(channelUpdateListener.ref, classOf[LocalChannelUpdate])
-    system.eventStream.subscribe(channelUpdateListener.ref, classOf[LocalChannelDown])
+    systemA.eventStream.subscribe(channelUpdateListener.ref, classOf[LocalChannelUpdate])
+    systemA.eventStream.subscribe(channelUpdateListener.ref, classOf[LocalChannelDown])
+    systemB.eventStream.subscribe(channelUpdateListener.ref, classOf[LocalChannelUpdate])
+    systemB.eventStream.subscribe(channelUpdateListener.ref, classOf[LocalChannelDown])
```
t-bast, line 131: Isn't it worth splitting the `channelUpdateListener` as well (one for each peer)?

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
```diff
@@ -89,15 +89,19 @@ class ClosingStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
         bob2blockchain.expectMsgType[WatchFundingConfirmed]
         awaitCond(alice.stateName == WAIT_FOR_FUNDING_CONFIRMED)
         awaitCond(bob.stateName == WAIT_FOR_FUNDING_CONFIRMED)
-        system.eventStream.subscribe(eventListener.ref, classOf[TransactionPublished])
-        system.eventStream.subscribe(eventListener.ref, classOf[TransactionConfirmed])
+        alice.underlying.system.eventStream.subscribe(eventListener.ref, classOf[TransactionPublished])
+        alice.underlying.system.eventStream.subscribe(eventListener.ref, classOf[TransactionConfirmed])
+        bob.underlying.system.eventStream.subscribe(eventListener.ref, classOf[TransactionPublished])
+        bob.underlying.system.eventStream.subscribe(eventListener.ref, classOf[TransactionConfirmed])
```
t-bast, line 95: It's probably worth going the extra mile here as well and have one `eventListener` per peer, our assertions will be more accurate.

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRouterIntegrationSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+    }
+    withFixture(test.toNoArgTest(FixtureParam(router, rebroadcastListener, channels, test.tags)))
+  }
+
+  test("private local channel") { f =>
+    import f._
+
+    reachNormal(channels, testTags)
+
+    awaitCond(router.stateData.privateChannels.size == 1)
```
t-bast, line 42: Don't you want to introduce `awaitAssert` in new tests?

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRouterIntegrationSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    {
+      // only the local channel_update is known
+      val pc = router.stateData.privateChannels.values.head
+      assert(pc.update_1_opt.isDefined ^ pc.update_2_opt.isDefined)
```
t-bast, line 47: Why isn't Bob sending us his `channel_update`? Is it because it's not deeply buried yet? Can you add a comment to explain that?

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRouterIntegrationSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    val peerConnection = TestProbe()
+    val bobChannelUpdate = channels.bob.stateData.asInstanceOf[DATA_NORMAL].channelUpdate
+    router ! PeerRoutingMessage(peerConnection.ref, channels.bob.underlyingActor.nodeParams.nodeId, bobChannelUpdate)
+
+    awaitCond {
+      // only the local channel_update is known
```
t-bast, line 55: Wrong copy-paste?

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRouterIntegrationSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+    val aliceChannelUpdate = channels.alice.stateData.asInstanceOf[DATA_NORMAL].channelUpdate
+    val bobChannelUpdate = channels.bob.stateData.asInstanceOf[DATA_NORMAL].channelUpdate
+    router ! PeerRoutingMessage(peerConnection.ref, channels.bob.underlyingActor.nodeParams.nodeId, bobChannelUpdate)
+
+    awaitCond {
+      val pc = router.stateData.privateChannels.values.head
+      // both channel_updates are known
+      pc.update_1_opt.isDefined && pc.update_2_opt.isDefined
+    }
+    val privateChannel = router.stateData.privateChannels.values.head
```
t-bast, line 89: nit: unused

t-bast (commented):
Cool stuff!

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: I thought it could come in handy in the future, but now that we have [8aee27a](https://github.com/ACINQ/eclair/pull/2224/commits/8aee27a724cfef6f2b269e7ab37544b5dff77651) we can probably get rid of it. Let me append that commit to this PR.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
pm47: > Looking ahead, isn't there a risk here that we'll broadcast an update that contains an alias short_channel_id (which doesn't make sense, others won't be able to validate it)? Shouldn't we wait to receive a new channel_update with the final short_channel_id?

Nice catch. Either we:
(a) only add `channel_update` to the `PublicChannel` if they use the real scid
(b) only broadcast if the `channel_update` uses the real `scid`

I'm tempted to go with (b), so we don't risk losing information, but it feels like (a) is more correct. Otherwise we will definitely send out `channel_update` with aliases in the context of channel queries.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/ChannelStateTestsHelperMethods.scala
pm47: I went with the minimal change. I think the whole handing of events in the state tests needs to be reworked.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
pm47: I did tinker with that, and backed off when I saw random tests failing and a new rabbit hole opening. We definitely need to rework that.

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRouterIntegrationSpec.scala
pm47: I didn't discover it at that time ^^

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/ChannelStateTestsHelperMethods.scala
t-bast: Fair enough, that can be a separate PR later.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/h/ClosingStateSpec.scala
t-bast: Let's keep that for a separate PR then!

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
t-bast: I think this is the kind of code you keep in a local branch to help debug stuff, but we shouldn't have it on `master`.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
t-bast: I'm also more tempted to go with (b), because it lets us keep creating invoices with routing hints even before we receive the "official" channel update with the real scid. Also, keeping channel updates with aliases around isn't an issue as long as it stays internal to eclair, it only becomes an issue if we broadcast them, so it make sense to have filtering done in the broadcast code, doesn't it?

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRouterIntegrationSpec.scala
pm47: Indeed we currently only send the `channel_update` to our direct peer if the channel is deeply buried:
https://github.com/ACINQ/eclair/blob/30412766fdd701b03b4d5fc8922f42e9a7f131d6/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala#L1596-L1610

But that looks like a bug to me... which has been fixed in #2224. That's why we now have this in the test setup: https://github.com/ACINQ/eclair/blob/8aee27a724cfef6f2b269e7ab37544b5dff77651/eclair-core/src/test/scala/fr/acinq/eclair/channel/states/ChannelStateTestsHelperMethods.scala#L244-L246

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
pm47: > I think it's worth extracting the updates to a separate line for readability:

Done in a7a35c3bb.

WDYT of the following?

```scala
val dx = d0
  .modify(_.channels).using(_ + (c.shortChannelId -> pc))
  .modify(_.privateChannels).using(_ - c.shortChannelId) // we remove fake announcements that we may have made before
  .modify(_.rebroadcast.channels).using(_ + (c -> d0.awaiting.getOrElse(c, Nil).toSet)) // we rebroadcast the channel to our peers)
  .modify(_.rebroadcast.updates).using(_ ++ updates1)
  .modify(_.stash).setTo(stash1)
  .modify(_.awaiting).setTo(awaiting1)
```

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRouterIntegrationSpec.scala
pm47: I added a lame comment in a7a35c3bb, will get back to it in #2224 after the behavior changes.

eclair-core/src/test/scala/fr/acinq/eclair/router/ChannelRouterIntegrationSpec.scala
pm47: Oops, fixed in a7a35c3bb.

t-bast (approved):
LGTM :+1:

PR #2272, opened 2022-05-16
Balance estimate per side

thomash-acinq:
Alternative to #2263 that uses the maximum capacity of all the channels between two nodes instead of the sum.
It make the balance information not symmetrical.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -227,6 +227,7 @@ eclair {
     channel-exclude-duration = 60 seconds // when a temporary channel failure is returned, we exclude the channel from our payment routes for this duration
     broadcast-interval = 60 seconds // see BOLT #7
     init-timeout = 5 minutes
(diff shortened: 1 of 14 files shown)


t-bast:
> It make the balance information not symmetrical.

I don't understand why, can you detail this statement?
I believe we can easily keep the information symmetrical, I really don't see why we would lose that property.

t-bast:
I understand what you mean, the missing information is that you may send `x` in one direction through one channel, the other side then cannot send `maxCapacity - x` through this same channel but could send it through another channel. You really need to detail that in comments in that file otherwise readers will waste understanding this.

I think this really is the right model. Even with `totalCapacity`, the information actually wasn't symmetrical either because there may be unannounced channels between nodes. I think this PR is going in the right direction, but the build needs to be fixed and [this comment](https://github.com/ACINQ/eclair/pull/2263#discussion_r871061659) should be addressed. Once that's done, I can review it (now that I'm out of my dual funding rabbit hole) and we should be able to integrate this quickly. How does that sound?

t-bast (approved):

PR #2274, opened 2022-05-17
Dual funding channel confirmation

t-bast:
Finalize the dual-funding flow: exchange signatures and wait for the funding transaction to confirm.

If an error occurs, we need to wait for the funding tx to be double-spent before we can consider the channel closed and safely forget it.

There is some logic duplication between single-funded and dual-funded handlers...I don't know how far we should go to try to remove duplication. If we extract too much, it may make it harder to just read an event handler's code and verify that every step is there and correctly done in the right order. Also, there are small subtleties where single-funded and dual-funded differ, and there will be more as we build new protocols on top of dual funding (e.g. liquidity ads). In the long term, we will eventually remove the code for the single-funded case, as it can be achieved as just a special case of dual-funding (but we'll need all peers on the network to support dual funding before we can really remove that). Let me know what you think and where you think we should refactor.


eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
```diff
@@ -132,13 +132,20 @@ trait ErrorHandlers extends CommonHandlers {
...
-        // funds back without delays if they can't double-spend the funding tx.
         val commitTx = d.commitments.fullySignedLocalCommitTx(keyManager)
         txPublisher ! PublishFinalTx(commitTx, 0 sat, None)
         goto(CLOSED)
+      case d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED if Closing.nothingAtStake(d) =>
+        val commitTx = d.commitments.fullySignedLocalCommitTx(keyManager)
+        txPublisher ! PublishFinalTx(commitTx, 0 sat, None)
+        goto(CLOSED)
+      case _: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED =>
+        log.info("cannot close channel while dual-funding txs are unconfirmed: waiting for a transaction to confirm or be double-spent")
```
pm47, line 147: I propose a different approach:
- we add `alternativeCommitments: Seq[Commitments]` to `DATA_CLOSING`, which will be non-empty if we transition from `WAIT_FOR_FUNDING_CONFIRMED`->`CLOSING`
- in `CLOSING`, there are two cases:
  - either the most recent funding tx confirms: nominal case, nothing to do
  - an older funding tx confirms: we build a fresh `CLOSING` for the commitments that confirmed (we can forget all other commitments, and all child transactions, they have all been double-spent).

t-bast: Good idea, done in https://github.com/ACINQ/eclair/pull/2274/commits/075a606665582c54862b63202d2c0fd68b3c4b1a

t-bast:
I changed the watching behavior in https://github.com/ACINQ/eclair/pull/2274/commits/bd479921eae2270b731cd1d03457bdae20acb58c to correctly detect channel force-close while we're offline. The commit looks simple, but it's actually very subtle, please review carefully! The main subtlety is that we when we detect an alternative commitment being used, we don't claim it, we instead wait for the corresponding funding tx to be confirmed before acting. If we claimed it instantly and swapped the main `commitments` with this alternative one, we would be screwed if it's actually the main funding tx that ends up confirming...

I cannot add yet test cases for those alternative commitments since RBF isn't implemented, but it's on my TODO-list to add to the RBF PR.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -479,16 +479,18 @@ final case class DATA_WAIT_FOR_ACCEPT_DUAL_FUNDED_CHANNEL(init: INPUT_INIT_CHANN
...
+                                                      fundingTx: SignedSharedTransaction,
+                                                      fundingParams: InteractiveTxParams,
+                                                      previousFundingTxs: Seq[DualFundingTx],
+                                                      waitingSince: BlockHeight, // how long have we been waiting for a funding tx to confirm
+                                                      lastChecked: BlockHeight, // last time we checked if the channel was double-spent
+                                                      rbfAttempt: Option[typed.ActorRef[InteractiveTxBuilder.Command]],
+                                                      deferred: Option[ChannelReady]) extends PersistentChannelData
+final case class DATA_WAIT_FOR_DUAL_FUNDING_READY(commitments: Commitments,
+                                                  shortIds: ShortIds,
+                                                  otherFundingTxs: Seq[DualFundingTx],
```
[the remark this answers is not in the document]
t-bast: Removed in https://github.com/ACINQ/eclair/pull/2274/commits/bd479921eae2270b731cd1d03457bdae20acb58c

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -306,21 +312,19 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
-              case fundingTx: FullySignedSharedTransaction =>
-                goto(WAIT_FOR_DUAL_FUNDING_PLACEHOLDER) using nextData sending fundingTx.localSigs calling publishFundingTx(nextData)
+              case fundingTx: PartiallySignedSharedTransaction => goto(WAIT_FOR_DUAL_FUNDING_CONFIRMED) using nextData storing() sending fundingTx.localSigs
+              case fundingTx: FullySignedSharedTransaction => goto(WAIT_FOR_DUAL_FUNDING_CONFIRMED) using nextData storing() sending fundingTx.localSigs calling publishFundingTx(fundingParams, fundingTx)
             }
           case None =>
-            val (_, channelReady) = acceptFundingTx(commitments, RealScidStatus.Unknown)
-            val nextData = DATA_WAIT_FOR_DUAL_FUNDING_PLACEHOLDER(commitments, fundingTx, fundingParams, Nil, nodeParams.currentBlockHeight, nodeParams.currentBlockHeight, None, None)
+            val commitTxs = Set(commitments.localCommit.commitTxAndRemoteSig.commitTx.tx.txid, commitments.remoteCommit.txid)
+            blockchain ! WatchFundingSpent(self, commitments.commitInput.outPoint.txid, commitments.commitInput.outPoint.index.toInt, commitTxs)
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2274/commits/bd479921eae2270b731cd1d03457bdae20acb58c

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1073,6 +1084,22 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
...
 
+    case Event(w: WatchFundingConfirmedTriggered, d: DATA_CLOSING) =>
+      d.alternativeCommitments.find(_.commitInput.outPoint.txid == w.tx.txid) match {
+        case Some(alternativeCommitments) =>
+          log.info("an alternative funding tx with txid={} got confirmed", w.tx.txid)
+          watchFundingTx(alternativeCommitments)
+          context.system.eventStream.publish(TransactionConfirmed(d.channelId, remoteNodeId, w.tx))
+          val commitTx = alternativeCommitments.fullySignedLocalCommitTx(keyManager).tx
+          val localCommitPublished = Closing.LocalClose.claimCommitTxOutputs(keyManager, alternativeCommitments, commitTx, nodeParams.currentBlockHeight, nodeParams.onChainFeeConf)
+          stay() using d.copy(commitments = alternativeCommitments, localCommitPublished = Some(localCommitPublished)) storing() calling doPublish(localCommitPublished, alternativeCommitments)
```
pm47, line 1095: That's not correct, we must forget about all `*Proposed`/`*Published`, they have been double spent. If we are logical we should also forget the alternate commitments. I think the best is to create a brand new `DATA_CLOSING`. It makes sense because we are switching from a completely separate commitment tree.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1073,6 +1084,22 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
 
     case Event(BITCOIN_FUNDING_TIMEOUT, d: DATA_CLOSING) => handleFundingTimeout(d)
 
+    case Event(w: WatchFundingConfirmedTriggered, d: DATA_CLOSING) =>
+      d.alternativeCommitments.find(_.commitInput.outPoint.txid == w.tx.txid) match {
+        case Some(alternativeCommitments) =>
+          log.info("an alternative funding tx with txid={} got confirmed", w.tx.txid)
+          watchFundingTx(alternativeCommitments)
+          context.system.eventStream.publish(TransactionConfirmed(d.channelId, remoteNodeId, w.tx))
+          val commitTx = alternativeCommitments.fullySignedLocalCommitTx(keyManager).tx
```
pm47, line 1093:
```suggestion
          // Force-closing is our only option here, if we are in this state the channel was closing and it is too late
          // to negotiate a mutual close
          val commitTx = alternativeCommitments.fullySignedLocalCommitTx(keyManager).tx
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1073,6 +1084,22 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
 
     case Event(BITCOIN_FUNDING_TIMEOUT, d: DATA_CLOSING) => handleFundingTimeout(d)
 
+    case Event(w: WatchFundingConfirmedTriggered, d: DATA_CLOSING) =>
+      d.alternativeCommitments.find(_.commitInput.outPoint.txid == w.tx.txid) match {
+        case Some(alternativeCommitments) =>
+          log.info("an alternative funding tx with txid={} got confirmed", w.tx.txid)
```
pm47, line 1090:
```suggestion
          log.info("an alternative funding tx with txid={} got confirmed", w.tx.txid)
          // This is a corner case where:
          // - we are using dual funding
          // - *and* the funding tx was RBF-ed
          // - *and* we went to CLOSING before any funding tx got confirmed (probably due to a local or remote error)
          // - *and* an older version of the funding tx confirmed and reached min depth (it won't be re-orged out).
          //
          // This means that:
          // - the whole current commitment tree has been double-spent and can safely be forgotten
          // - from now on, we only need to keep track of the commitment associated to the funding tx that got confirmed
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
```diff
@@ -268,13 +274,31 @@ trait ErrorHandlers extends CommonHandlers {
...
+    // We wait for this alternative funding tx to be confirmed before claiming the commit tx outputs, as it could still
+    // be replaced by another funding tx.
+    blockchain ! WatchFundingConfirmed(self, commitments.commitInput.outPoint.txid, nodeParams.channelConf.minDepthBlocks)
+    val nextData = d match {
+      case waitForFundingConfirmed: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED => DATA_CLOSING(d.commitments, waitForFundingConfirmed.signedFundingTx_opt, waitingSince = nodeParams.currentBlockHeight, alternativeCommitments, mutualCloseProposed = Nil)
+      case closing: DATA_CLOSING => closing
+      case _ => DATA_CLOSING(d.commitments, None, waitingSince = nodeParams.currentBlockHeight, alternativeCommitments, mutualCloseProposed = Nil)
+    }
+    goto(CLOSING) using nextData storing()
+  }
```
pm47, line 300: I don't think this method is necessary: we only need to track the funding tx, until one of them reaches min_depth, which eliminates all alternative commitments. From then, we are back to the nominal case. Our system already supports reacting asynchronously to transactions, this is what we do when the node was offline during a force close.

pm47 (commented):
Quick feedback to make sure we are on the same page with the new approach.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
t-bast: Agreed, that makes more sense.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
t-bast: > we only need to track the funding tx

You mean funding txs, right? Note that for (future) splicing, we will have something similar, but in the `NORMAL` state, where we'll have a set of future funding txs that spend the current txs.

> we only need to track the funding tx, until one of them reaches min_depth, which eliminates all alternative commitments.

I don't understand how you would achieve that if you remove this method...in OFFLINE and SYNCING you need to identify cases where `WatchFundingSpentTriggered` is for an alternative commitment, to go to the closing state and put a `WatchConfirmed` on that tx? Remember that while offline, we don't have a watch-confirmed on those transactions, we only set these watches when receiving the remote `channel_reestablish`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -306,21 +312,19 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+            val (shortIds, channelReady) = acceptFundingTx(commitments, RealScidStatus.Unknown)
+            val nextData = DATA_WAIT_FOR_DUAL_FUNDING_READY(commitments, shortIds, Nil, channelReady)
             fundingTx match {
-              case fundingTx: PartiallySignedSharedTransaction =>
-                goto(WAIT_FOR_DUAL_FUNDING_PLACEHOLDER) using nextData sending Seq(fundingTx.localSigs, channelReady)
-              case fundingTx: FullySignedSharedTransaction =>
-                goto(WAIT_FOR_DUAL_FUNDING_PLACEHOLDER) using nextData sending Seq(fundingTx.localSigs, channelReady) calling publishFundingTx(nextData)
+              case fundingTx: PartiallySignedSharedTransaction => goto(WAIT_FOR_DUAL_FUNDING_READY) using nextData storing() sending Seq(fundingTx.localSigs, channelReady)
```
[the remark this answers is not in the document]
t-bast: Added in https://github.com/ACINQ/eclair/pull/2274/commits/1d846d22cd1884d6ec1e929d06709f820a42309e

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -349,8 +353,144 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+
+    case Event(remoteChannelReady: ChannelReady, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
+      // We can skip waiting for confirmations if:
+      //  - there is a single version of the funding tx (otherwise we don't know which one to use)
+      //  - they didn't contribute to the funding output or we trust them to not double-spend
+      val canUseZeroConf = remoteChannelReady.alias_opt.isDefined &&
+        d.previousFundingTxs.isEmpty &&
+        (d.fundingParams.remoteAmount == 0.sat || d.commitments.localParams.initFeatures.hasFeature(Features.ZeroConf))
+      if (canUseZeroConf) {
+        log.info("this chanel isn't zero-conf, but they sent an early channel_ready with an alias: no need to wait for confirmations")
```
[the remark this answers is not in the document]
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/2274/commits/1d846d22cd1884d6ec1e929d06709f820a42309e, I fixed other places of the codebase where we have that typo as well

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -349,8 +353,144 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+              d.deferred.foreach(self ! _)
+              val otherFundingTxs = allFundingTxs.filter(_.commitments.commitInput.outPoint.txid != confirmedTx.txid)
+              goto(WAIT_FOR_DUAL_FUNDING_READY) using DATA_WAIT_FOR_DUAL_FUNDING_READY(commitments, shortIds, otherFundingTxs, channelReady) storing() sending channelReady
+            case Failure(t) =>
+              log.error(t, s"rejecting channel with invalid funding tx: ${confirmedTx.bin}")
+              allFundingTxs.foreach(f => wallet.rollback(f.fundingTx.tx.buildUnsignedTx()))
+              goto(CLOSED)
+          }
+        case None =>
+          log.error(s"rejecting channel with invalid funding tx that doesn't match any of our funding txs: ${confirmedTx.bin}")
```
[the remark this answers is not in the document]
t-bast: Updated in https://github.com/ACINQ/eclair/pull/2274/commits/1d846d22cd1884d6ec1e929d06709f820a42309e

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/CommonCodecs.scala
```diff
@@ -110,6 +110,8 @@ object CommonCodecs {
 
   val listofsignatures: Codec[List[ByteVector64]] = listOfN(uint16, bytes64)
 
+  def seqOfN[A](countCodec: Codec[Int], valueCodec: Codec[A]): Codec[Seq[A]] = listOfN(countCodec, valueCodec).xmap(_.toSeq, _.toList)
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2274/commits/1d846d22cd1884d6ec1e929d06709f820a42309e

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -349,8 +353,144 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+        context.system.eventStream.publish(ShortChannelIdAssigned(self, d.commitments.channelId, shortIds = shortIds1, remoteNodeId = remoteNodeId))
+      }
+      log.info("shortIds: real={} localAlias={} remoteAlias={}", shortIds1.real.toOption.getOrElse("none"), shortIds1.localAlias, shortIds1.remoteAlias_opt.getOrElse("none"))
+      // we create a channel_update early so that we can use it to send payments through this channel, but it won't be propagated to other nodes since the channel is not yet announced
+      val scidForChannelUpdate = Helpers.scidForChannelUpdate(channelAnnouncement_opt = None, shortIds1.localAlias)
+      log.info("using shortChannelId={} for initial channel_update", scidForChannelUpdate)
+      val fees = getRelayFees(nodeParams, remoteNodeId, d.commitments)
+      val initialChannelUpdate = Announcements.makeChannelUpdate(nodeParams.chainHash, nodeParams.privateKey, remoteNodeId, scidForChannelUpdate, nodeParams.channelConf.expiryDelta, d.commitments.remoteParams.htlcMinimum, fees.feeBase, fees.feeProportionalMillionths, d.commitments.capacity.toMilliSatoshi, enable = Helpers.aboveReserve(d.commitments))
```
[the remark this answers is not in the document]
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/2274/commits/33b60156078b406135955af9dbd130d0556941fe

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -349,8 +353,144 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+      // we need to periodically re-send channel updates, otherwise channel will be considered stale and get pruned by network
+      context.system.scheduler.scheduleWithFixedDelay(initialDelay = REFRESH_CHANNEL_UPDATE_INTERVAL, delay = REFRESH_CHANNEL_UPDATE_INTERVAL, receiver = self, message = BroadcastChannelUpdate(PeriodicRefresh))
+      // used to get the final shortChannelId, used in announcements (if minDepth >= ANNOUNCEMENTS_MINCONF this event will fire instantly)
+      blockchain ! WatchFundingDeeplyBuried(self, d.commitments.commitInput.outPoint.txid, ANNOUNCEMENTS_MINCONF)
+      goto(NORMAL) using DATA_NORMAL(d.commitments.copy(remoteNextCommitInfo = Right(channelReady.nextPerCommitmentPoint)), shortIds1, None, initialChannelUpdate, None, None, None) storing()
+
+    case Event(_: TxInitRbf, d: DATA_WAIT_FOR_DUAL_FUNDING_READY) =>
+      // Our peer may not have received the funding transaction confirmation.
+      stay() sending TxAbort(d.channelId, InvalidRbfTxConfirmed(d.channelId).getMessage)
```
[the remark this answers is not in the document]
t-bast: Agreed, those two handlers are really worth factoring, done in https://github.com/ACINQ/eclair/pull/2274/commits/33b60156078b406135955af9dbd130d0556941fe

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
```diff
@@ -32,22 +37,56 @@ trait DualFundingHandlers extends CommonHandlers {
...
+    } else if (d.lastChecked + 6 < c.blockHeight) {
+      log.debug("checking if funding transactions have been double-spent")
+      val fundingTxs = (d.fundingTx +: d.previousFundingTxs.map(_.fundingTx)).map(_.tx.buildUnsignedTx())
+      // We check whether *all* funding attempts have been double-spent.
+      // Since we only consider a transaction double-spent when the spending transaction is confirmed, this will not
+      // return false positives when one of our transactions is confirmed, because its individual result will be false.
+      Future.sequence(fundingTxs.map(tx => wallet.doubleSpent(tx))).map(_.forall(_ == true)).map {
+        case true => self ! BITCOIN_FUNDING_DOUBLE_SPENT(fundingTxs.map(_.txid).toSet)
+        case false => publishFundingTx(d.fundingParams, d.fundingTx) // we republish the highest feerate funding transaction
+      }
```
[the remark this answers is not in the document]
t-bast: > Maybe we could just decorrelate from the block, or trigger a random check every 6 blocks after a random delay (similarly to what happens in the watcher).

I'm hesitant to do that for the following reasons:

- it creates a new event that needs to be listened to, which slightly increases code complexity (not a lot though, that's not a good enough reason on its own)
- we already have a protection against herd effect since each channel randomly delays block processing to ensure we don't have all channels handling blocks at exactly the same time (see `max-block-processing-delay`)
- this check only happens while funding is unconfirmed, which in practice shouldn't last very long
- even if we have multiple channels in an unconfirmed state, they won't check for double-spend at the same block, unless they all reached the wait_for_unconfirmed state at the same moment (or at the same multiple of 6)
- I arbitrarily chose 6 blocks, but we can check that less often if you'd like, a few times a day is likely enough

Let me know if you think it's worth doing regardless of these points and I'll do it.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
pm47: We leave the handlers in `OFFLINE`/`SYNCING`, but make them call `spendLocalCurrent` (after all this is an error scenario). Then we rely on the `WAIT_FOR_FUNDING_CONFIRMED` handler in `CLOSING` to deal with the commitment switch?

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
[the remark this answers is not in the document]
pm47: > - we already have a protection against herd effect since each channel randomly delays block processing to ensure we don't have all channels handling blocks at exactly the same time (see max-block-processing-delay)

Fair points! Especially this one. Happy to leave it as-is.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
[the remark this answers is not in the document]
pm47: One nit tough. I'd rather use `d1` instead of `nextData`, because the latter is dangerously close to `nextStateData`, which is a method from the FSM class that returns the same type. It's also a convention we have pretty much everywhere.

https://github.com/ACINQ/eclair/blob/33b60156078b406135955af9dbd130d0556941fe/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala#L459

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
t-bast: I think you're right, this should work. I implemented this in https://github.com/ACINQ/eclair/pull/2274/commits/7b7bfac5bb44092cac0a7439723abff63ae4f356 let me know if I missed something. There is duplication in the handlers in OFFLINE and SYNCING, but I'm not really sure how to name the method we'd use to factor those together, do you have an idea?

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
t-bast: Added in https://github.com/ACINQ/eclair/pull/2274/commits/7b7bfac5bb44092cac0a7439723abff63ae4f356

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
t-bast: Added in https://github.com/ACINQ/eclair/pull/2274/commits/7b7bfac5bb44092cac0a7439723abff63ae4f356

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2274/commits/7b7bfac5bb44092cac0a7439723abff63ae4f356

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2274/commits/7b7bfac5bb44092cac0a7439723abff63ae4f356

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
pm47: So here are the possible transitions (in a RBF scenario):
- `DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED`->`CLOSING`
_channel was created then closed, no disconnections_
  In this case, we have put a `WatchFundingConfirmed` and a `WatchFundingSpent` on all alternate funding txs in state `WAIT_FOR_DUAL_FUNDING_CREATED`. There will be a harmless race in state `CLOSING` between `WatchFundingSpentTriggered` (which cause us to put a duplicate `WatchFundingConfirmed`, deduped by the watcher) and `WatchFundingConfirmedTriggered`. Btw I think we should also handle the nominal case were the latest funding tx confirms, and clean up the alternate commitments.
- `WAIT_FOR_FUNDING_CREATED`->`OFFLINE`[->`SYNCING`]->`CLOSING`
  _channel was created then disconnected, then closed, but eclair stayed up the whole time_
  Same as above, except that  the `WatchFundingConfirmed` and `WatchFundingSpentTriggered` that were put in `WAIT_FOR_DUAL_FUNDING_CREATED` may also race in `OFFLINE`/`SYNCING`. The result is the same, we may put duplicate `WatchFundingConfirmed` (in that case they will be ignored), but it's harmless.
- `DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED`-><FSM stopped>->`WAIT_FOR_INIT_INTERNAL`[->`OFFLINE`][->`SYNCING`]->`CLOSING``
  _channel was created then eclair went offline, then tge channel closed_
  In this scenario, we have not put any `WatchFundingConfirmed`, so there won't be a race: we will (1) receive `WatchFundingSpentTriggered`, then (2) put `WatchFundingConfirmed`, then (3) receive `WatchFundingConfirmedTriggered`. ⚠️ Note that we currently have a bug where we don't put a `WatchFundingSpent` on alternative funding txs in `WAIT_FOR_INIT_INTERNAL`.

I think we could simplify the flow and not rely on `WatchFundingSpent` to "discover" alternate funding txs. This approach was justified when there was a huge list of potential spending txs, and we wanted to limit the number of `WatchConfirmed`, which are costly. But in this case, where there is only a handful of competing funding txs, we could only rely on `WatchConfirmed`. We would then use a catch-all handler in `whenUnhandled` to detect when an alternate commitment has taken over, and then go to the regular local close flow (using that commitment). This also clears up a potential confusion where we use `WatchFundingSpent` for two very different purposes: spending the output of the funding tx (original case), and double spending the funding tx (new use case).

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1073,6 +1092,36 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
...
+          watchFundingTx(commitments1)
+          context.system.eventStream.publish(TransactionConfirmed(d.channelId, remoteNodeId, w.tx))
+          val commitTx = commitments1.fullySignedLocalCommitTx(keyManager).tx
+          val localCommitPublished = Closing.LocalClose.claimCommitTxOutputs(keyManager, commitments1, commitTx, nodeParams.currentBlockHeight, nodeParams.onChainFeeConf)
+          val d1 = DATA_CLOSING(commitments1, None, d.waitingSince, alternativeCommitments = Nil, mutualCloseProposed = Nil, localCommitPublished = Some(localCommitPublished))
+          stay() using d1 storing() calling doPublish(localCommitPublished, commitments1)
+        case None if d.commitments.commitInput.outPoint.txid == w.tx.txid =>
+          // The best funding tx candidate has been confirmed, we can forget alternative commitments.
+          stay() using d.copy(alternativeCommitments = Nil)
+        case None =>
```
pm47, line 1120: Nit: order-sensitive pattern matching are a bit footgunny, we could do:
```scala
case None => if (...) {} else {}
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2274/commits/d5d0e36ccdf018dff3d845aa9cec8beb84089b97, I think it's indeed better to use `WatchFundingConfirmed` instead of `WatchFundingSpent`, the "trick" is to update the underlying state (`DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED`) to "erase" obsolete previous txs when one confirms.

So now in each transition, we have:

- `WAIT_FOR_DUAL_FUNDING_CONFIRMED` -> `CLOSING`: we have put `WatchFundingConfirmed` on each transaction (and no `WatchFundingSpent`), so the handler for that event in `CLOSING` should correctly settle every case
- `WAIT_FOR_FUNDING_CREATED`->`OFFLINE`[->`SYNCING`]->`CLOSING`: we have put `WatchFundingConfirmed` on each transaction (and no `WatchFundingSpent`). Once one confirms, we update the inner `DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED` to have this commitments as the main one and put a `WatchFundingSpent`. When the remote spends the funding tx, we go through the normal flow of `handleRemoteSpentCurrent`. If an error occurs, we will spend our local commit from this confirmed funding tx.
- `WAIT_FOR_DUAL_FUNDING_CONFIRMED`->FSM stopped->`WAIT_FOR_INIT_INTERNAL`[->`OFFLINE`][->`SYNCING`]->`CLOSING`: we put back a `WatchFundingConfirmed` for each transaction, so we should be in the same state as the previous one

t-bast: We now also have the case `WAIT_FOR_FUNDING_CONFIRMED` -> `OFFLINE` -> `WatchFundingConfirmed` -> `WAIT_FOR_FUNDING_CONFIRMED`.

In that case when receiving `channel_reestablish` we put a `WatchFundingConfirmed` again, which will trigger immediately the corresponding handler in `WAIT_FOR_FUNDING_CONFIRMED` and move us to the next state.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
```diff
@@ -28,26 +34,79 @@ import scala.util.{Failure, Success}
...
             channelOpenReplyToUser(Left(LocalError(t)))
             log.warning("error while publishing funding tx: {}", t.getMessage) // tx may be published by our peer, we can't fail-fast
         }
     }
   }
 
+  def handleDualFundingConfirmedOffline(w: WatchFundingConfirmedTriggered, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) = {
+    if (w.tx.txid == d.commitments.commitInput.outPoint.txid) {
+      watchFundingTx(d.commitments)
+      context.system.eventStream.publish(TransactionConfirmed(d.channelId, remoteNodeId, w.tx))
```
pm47, line 62: Sending `TransactionConfirmed` here will lead to a duplicate (we will send them again in state `WAIT_FOR_DUAL_FUNDING_CONFIRMED` after we reconnect), and is different from the single-funding flow (where we won't emit that event if the funding confirms when disconnected and then we go directly to close).

It troubles me a little bit that this handler is very similar, but not quite the same as in `WAIT_FOR_DUAL_FUNDING_CONFIRMED`. For example, we don't call `Transaction.correctlySpends`.

pm47 (commented):
I'm pretty sure this works fine but I have still doubt about the `OFFLINE`/`SYNCING` case. Let's sleep on it 🤞.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
pm47: Now that I think about it: why do we call `Transaction.correctlySpends` at all in `Channel.scala`? In the dual funding case we already know the full funding tx and have already verified it in `InteractiveTxBuilder.scala`.

t-bast: > Sending TransactionConfirmed here will lead to a duplicate

This isn't an issue, this is deduplicated in the DB. I believe it's important to send it here, because if we close before reconnecting, we would otherwise never send it (which is probably a bug in the single-funding flow, that we could fix when updating the single-funding flow to match the dual-funded one and rely on watching confirmation first?).

> For example, we don't call Transaction.correctlySpends.

We could, calling this function is really just a belts-and-suspenders because of the first CVE lightning had, it should never yield a failure, but I agree that it would be more symmetrical.

t-bast: > Now that I think about it: why do we call Transaction.correctlySpends at all in Channel.scala? In the dual funding case we already know the full funding tx and have already verified it in InteractiveTxBuilder.scala.

Let's just remove it, it simplifies the code!

t-bast: Removed in https://github.com/ACINQ/eclair/pull/2274/commits/f49116ab7af3733bd2b1055c66e8eaa27236276b

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2274/commits/f49116ab7af3733bd2b1055c66e8eaa27236276b

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1073,6 +1092,36 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
...
+          log.info("an alternative funding tx with txid={} got confirmed", w.tx.txid)
+          watchFundingTx(commitments1)
+          context.system.eventStream.publish(TransactionConfirmed(d.channelId, remoteNodeId, w.tx))
+          val commitTx = commitments1.fullySignedLocalCommitTx(keyManager).tx
+          val localCommitPublished = Closing.LocalClose.claimCommitTxOutputs(keyManager, commitments1, commitTx, nodeParams.currentBlockHeight, nodeParams.onChainFeeConf)
+          val d1 = DATA_CLOSING(commitments1, None, d.waitingSince, alternativeCommitments = Nil, mutualCloseProposed = Nil, localCommitPublished = Some(localCommitPublished))
+          stay() using d1 storing() calling doPublish(localCommitPublished, commitments1)
+        case None if d.commitments.commitInput.outPoint.txid == w.tx.txid =>
+          // The best funding tx candidate has been confirmed, we can forget alternative commitments.
+          stay() using d.copy(alternativeCommitments = Nil)
```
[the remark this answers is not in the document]
t-bast: Good catch, done in https://github.com/ACINQ/eclair/pull/2274/commits/f49116ab7af3733bd2b1055c66e8eaa27236276b

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
pm47: I think we should rollback alternative funding txs before "forgetting" them. Actually I believe there is a bug in the `WAIT_FOR_DUAL_FUNDING_CONFIRMED` handler: if the latest funding tx confirms, it automatically rollbacks all alternative txs (since they are double spent), but the opposite isn't true if an alternative tx confirms.

I also think that we should we rollback other funding txs as soon as possible, i.e. in `OFFLINE`/`SYNCING`. We may stay disconnected for a long time.

Nit: There are minor inconsistencies in logs between the two handlers.

t-bast: I harmonized the logs in https://github.com/ACINQ/eclair/pull/2274/commits/ccf43f9202609d78d053d36f1e27949365461944

Note that in the offline/syncing and closing cases, we don't log anything when the expected funding transaction confirms, because this logs would otherwise repeat every time we restart eclair, which is unnecessarily noisy. The node operator will get a log line when the channel comes back online or if it's closed before that, I believe this is enough?

pm47 (approved):
Onto the next one! ⛵

PR #2275, opened 2022-05-17
Dual funding RBF support

t-bast:
Add support for bumping the fees of a dual funding transaction. We spawn a transient dedicated actor: if the RBF attempt fails, or if we are disconnected before completing the protocol, we should forget it.

With that PR, dual-funding is fully supported by eclair! There is no way for the non-initiator to contribute though, as we'd need a mechanism to let node operators preset some "strategy". The first use-case for contributing as the non-initiator will be liquidity ads, which we will support soon.

Note that I haven't been able to do cross-compatibility tests with cln yet, as their version of dual funding doesn't match the latest state of the spec PR. @niftynei is working on this and I'll do cross-compat tests as soon as they're ready, but I don't think this prevents us from merging this PR to `master` since dual funding is disabled by default and we've documented that we still may break backwards-compatibility.


eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -362,23 +363,156 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
-    case Event(_: TxInitRbf, _: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
-      log.info("rbf not supported yet")
-      stay()
+    case Event(msg: TxInitRbf, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
+      if (d.fundingParams.isInitiator) {
+        // Only the initiator is allowed to initiate RBF.
+        log.info("rejecting tx_init_rbf, we're the initiator, not them!")
+        stay() sending TxAbort(d.channelId, InvalidRbfAttempt(d.channelId).getMessage)
+      } else {
+        val minNextFeerate = d.fundingParams.targetFeerate * 25 / 24
```
pm47, line 383: This ratio should probably be factored.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -362,23 +363,156 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+            } else {
+              stay() using d.copy(rbfAttempt = Some(Left(cmd.copy(replyTo = replyTo)))) sending TxInitRbf(d.channelId, cmd.lockTime, cmd.targetFeerate, d.fundingParams.localAmount)
+            }
+        }
+      }
+
+    case Event(msg: TxAckRbf, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
+      d.rbfAttempt match {
+        case Some(Left(cmd)) =>
+          log.info("our peer accepted our rbf attempt and will contribute {} to the funding transaction", msg.fundingContribution_opt.getOrElse(0 sat))
```
pm47, line 439: Should `.getOrElse(0 sat)` be done in the attribute definition (which wouldn't return an option)?

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -362,23 +363,156 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
       }
 
-    case Event(_: TxInitRbf, _: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
-      log.info("rbf not supported yet")
-      stay()
+    case Event(msg: TxInitRbf, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
+      if (d.fundingParams.isInitiator) {
+        // Only the initiator is allowed to initiate RBF.
+        log.info("rejecting tx_init_rbf, we're the initiator, not them!")
+        stay() sending TxAbort(d.channelId, InvalidRbfAttempt(d.channelId).getMessage)
```
pm47, line 381: Shouldn't we return an `Error` instead? This is a protocol violation after all.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -391,7 +525,19 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
           val realScidStatus = RealScidStatus.Temporary(RealShortChannelId(blockHeight, txIndex, commitments.commitInput.outPoint.index.toInt))
           val (shortIds, channelReady) = acceptFundingTx(commitments, realScidStatus = realScidStatus)
           d.deferred.foreach(self ! _)
-          goto(WAIT_FOR_DUAL_FUNDING_READY) using DATA_WAIT_FOR_DUAL_FUNDING_READY(commitments, shortIds, channelReady) storing() sending channelReady
+          val otherFundingTxs = allFundingTxs.filter(_.commitments.commitInput.outPoint.txid != confirmedTx.txid).map(_.fundingTx)
+          rollbackDualFundingTxs(otherFundingTxs)
```
pm47, line 529: Did we miss this `rollbackDualFundingTxs` in #2390?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -495,7 +497,7 @@ final case class DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED(commitments: Commitments,
                                                       previousFundingTxs: List[DualFundingTx],
                                                       waitingSince: BlockHeight, // how long have we been waiting for a funding tx to confirm
                                                       lastChecked: BlockHeight, // last time we checked if the channel was double-spent
-                                                      rbfAttempt: Option[typed.ActorRef[InteractiveTxBuilder.Command]],
+                                                      rbfAttempt: Option[Either[CMD_BUMP_FUNDING_FEE, typed.ActorRef[InteractiveTxBuilder.Command]]],
```
pm47, line 500: Maybe:
```scala
sealed trait RbfStatus
object RbfStatus {
  case object NoRbf extends RbfStatus
  case class RbfRequested(cmd: CMD_BUMP_FUNDING_FEE) extends RbfStatus
  case class RbfInProgress(rbf: typed.ActorRef[InteractiveTxBuilder.Command]) extends RbfStatus
}
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -362,23 +363,156 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+            wallet))
+          txBuilder ! InteractiveTxBuilder.Start(self, d.fundingTx +: d.previousFundingTxs.map(_.fundingTx))
+          stay() using d.copy(rbfAttempt = Some(Right(txBuilder))) sending TxAckRbf(d.channelId, fundingParams.localAmount)
+        }
+      }
 
-    case Event(_: TxAckRbf, _: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
-      log.info("rbf not supported yet")
-      stay()
+    case Event(cmd: CMD_BUMP_FUNDING_FEE, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
```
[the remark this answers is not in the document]
t-bast: Moved in https://github.com/ACINQ/eclair/pull/2275/commits/56be83e50b7e618c333db6d993fb0346aad73d8a

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -362,23 +363,156 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+      val replyTo = if (cmd.replyTo == ActorRef.noSender) sender() else cmd.replyTo
+      if (!d.fundingParams.isInitiator) {
+        replyTo ! Status.Failure(new RuntimeException("cannot initiate rbf: we're not the initiator of this dual-funding attempt"))
```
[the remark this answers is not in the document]
t-bast: Agreed, done in https://github.com/ACINQ/eclair/pull/2275/commits/ba3baefe8114d5930c8049d530a2224e1b8b104b

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -362,23 +363,156 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+        stay()
+      } else {
+        d.rbfAttempt match {
+          case Some(_) =>
+            log.warning("cannot initiate rbf, another one is already in progress")
+            replyTo ! Status.Failure(InvalidRbfAlreadyInProgress(d.channelId))
```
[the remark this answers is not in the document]
t-bast: Updated in https://github.com/ACINQ/eclair/pull/2275/commits/ba3baefe8114d5930c8049d530a2224e1b8b104b

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -362,23 +363,156 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+            stay()
+          case None =>
+            val minNextFeerate = d.fundingParams.targetFeerate * 25 / 24
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2275/commits/ba3baefe8114d5930c8049d530a2224e1b8b104b

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2275/commits/ba3baefe8114d5930c8049d530a2224e1b8b104b

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
t-bast: Good point, done in https://github.com/ACINQ/eclair/pull/2275/commits/ba3baefe8114d5930c8049d530a2224e1b8b104b

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
t-bast: Right, but it really doesn't matter as we simply couldn't have multiple funding txs since RBF wasn't implemented.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
```diff
@@ -80,7 +82,12 @@ trait DualFundingHandlers extends CommonFundingHandlers {
   }
 
   def handleNewBlockDualFundingUnconfirmed(c: CurrentBlockHeight, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) = {
-    if (Channel.FUNDING_TIMEOUT_FUNDEE < c.blockHeight - d.waitingSince && Closing.nothingAtStake(d)) {
+    // We regularly notify the node operator that they may want to RBF this channel.
+    val blocksSinceOpen = c.blockHeight - d.waitingSince
+    if (d.fundingParams.isInitiator && (blocksSinceOpen % 288 == 0)) {
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2275/commits/ba3baefe8114d5930c8049d530a2224e1b8b104b

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
t-bast: I'll try that.

t-bast: This is a nice clean-up, done in https://github.com/ACINQ/eclair/pull/2275/commits/439c45c9f90fd404359dfe8182157b82dd6fa2c9

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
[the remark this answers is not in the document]
pm47: I see now that the logic of your naming was `InvalidRbf` + `reason`. Which is why you also have `InvalidRbfTxConfirmed`. I think we either need to revert the former to `InvalidRbfAlreadyInProgress`, or change the latter to `RbfTxConfirmed`

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
pm47: There are two other cases in that handler.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
pm47: I think we should add a `val fundingTxid = this.commitInput.outPoint.txid` to `Commitments`, it would be more readable and  it's used 36 times (!) in the source code.

t-bast: Definitely, this was long overdue! Done in https://github.com/ACINQ/eclair/pull/2275/commits/135870bfb2f1660d74ab3eac7a07475b121651bb

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
t-bast: Actually I reverted that to a `TxAbort`  in https://github.com/ACINQ/eclair/pull/2275/commits/135870bfb2f1660d74ab3eac7a07475b121651bb for the following reason: sending an `error` here means we'll close the channel, which may cost us on-chain fees. We have a valid funding tx waiting to be confirmed, so it's a bit of a waste to force-close because of an rbf issue that we can safely ignore?

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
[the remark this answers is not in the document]
t-bast: Yes, I agree it's a bit hard to read, it should read as `InvalidRbf: AlreadyInProgress`, I reverted that in https://github.com/ACINQ/eclair/pull/2275/commits/135870bfb2f1660d74ab3eac7a07475b121651bb

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
pm47: If we refer to the spec, `TxAbort` isn't related to errors:
> The `tx_abort` message allows for the cancellation of an in progress negotiation, and a return to the initial starting state.

So I would argue that we could instead send a `warning`, if we don't want the channel to close. But we're nitpicking, reverting to `TxAbort` is fine.

t-bast: The benefit of sending `TxAbort` is that it puts the channel back into a valid state: they did an invalid RBF operation, so we just abort the rbf attempt, and the channel is in a clean state again, does that make sense?

We could send a warning on top of that, but since `TxAbort` already includes a failure message, I believe it would be redundant?

pm47: Let's leave it that way and see how cln implements?

pm47 (approved):
That one was easy 😎

PR #2278, opened 2022-05-20
Unlock utxos at startup

t-bast:
We remove utxos locks that were previously set, otherwise we may have utxos that stay locked indefinitely. This may happen if we started funding a channel and restarted before completing the funding flow: the channel will be safely forgotten but the corresponding utxos stay locked.

The only drawback that this may have is if we have funded and signed a funding transaction but didn't publish it and we accidentally double-spend it after a restart. This shouldn't be an issue because:

- locks are automatically removed when the transaction is published anyway
- funding transactions are republished at startup if they aren't in the blockchain or in the mempool
- funding transactions detect when they are double-spent and abort the channel creation
- the only case where double-spending a funding transaction causes  a loss of funds is when we accept a 0-conf channel and our peer double-spends it, but we should never accept 0-conf channels from peers we don't trust

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
@@ -257,9 +257,8 @@ class Setup(val datadir: File,
 
       bitcoinClient = new BitcoinCoreClient(bitcoin)
 
-      // We remove utxos locks that were previously set, otherwise we may have utxos that stay locked indefinitely.
-      // This may happen if we started funding a channel and restarted before completing the funding flow: the channel
-      // will be safely forgotten but the corresponding utxos stay locked.
+      // If we started funding a transaction and restarted before signing it, we may have utxos that stay locked forever.
+      // We want to do something about it: we can unlock them automatically, or let the node operator decide what to do.
       //
       // The only drawback that this may have is if we have funded and signed a funding transaction but didn't publish
       // it and we accidentally double-spend it after a restart. This shouldn't be an issue because:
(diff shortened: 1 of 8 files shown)


t-bast:
@pm47 I added configuration hooks to let node operators decide what to do with locked utxos at startup in the second commit.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -257,6 +257,33 @@ class Setup(val datadir: File,
...
+          bitcoinClient.unlockOutpoints(lockedOutpoints.toSeq)
+        case lockedOutpoints if lockedUtxosBehavior == "stop" =>
+          logger.warn(s"cannot start eclair with locked utxos: ${lockedOutpoints.map(o => s"${o.txid}:${o.index}").mkString(", ")}")
+          logger.info("if you want eclair to automatically unlock utxos at startup, set 'eclair.bitcoind.startup-locked-utxos-behavior = \"unlock\"' in your eclair.conf and restart")
+          logger.info("if you want to start eclair without unlocking those utxos, set 'eclair.bitcoind.startup-locked-utxos-behavior = \"ignore\"' in your eclair.conf and restart")
+          logger.info("otherwise, run the following command to unlock them and restart eclair: 'bitcoin-cli lockunspent true'")
+          throw new RuntimeException("cannot start with locked utxos: see logs for detailed instructions to fix it")
+        case lockedOutpoints =>
+          logger.warn(s"ignoring locked utxos: ${lockedOutpoints.map(o => s"${o.txid}:${o.index}").mkString(", ")}")
+          Future.successful(true)
```
pm47, line 284: I think we should use the `NotificationsLogger` for this use case.

I also replaced pattern matching by `if/else` statements because relying on the order is a bit footgunny:

```suggestion
      lockedUtxosBehavior = config.getString("bitcoind.startup-locked-utxos-behavior").toLowerCase
      _ <- bitcoinClient.listLockedOutpoints().flatMap { lockedOutpoints =>
        if (lockedOutpoints.isEmpty) {
          Future.successful(true)
        } else if (lockedUtxosBehavior == "unlock") {
          logger.info(s"unlocking utxos: ${lockedOutpoints.map(o => s"${o.txid}:${o.index}").mkString(", ")}")
          bitcoinClient.unlockOutpoints(lockedOutpoints.toSeq)
        } else if (lockedUtxosBehavior == "stop") {
          logger.warn(s"cannot start eclair with locked utxos: ${lockedOutpoints.map(o => s"${o.txid}:${o.index}").mkString(", ")}")
          NotificationsLogger.logFatalError(
            """aborting startup as configured strategy for locked utxos
              |
              |If you want eclair to automatically unlock utxos at startup, set 'eclair.bitcoind.startup-locked-utxos-behavior = "unlock"' in your eclair.conf and restart.
              |If you want to start eclair without unlocking those utxos, set 'eclair.bitcoind.startup-locked-utxos-behavior = "ignore"' in your eclair.conf and restart;
              |
              |otherwise, run the following command to unlock them and restart eclair: 'bitcoin-cli lockunspent true'
              |""".stripMargin)
          throw new RuntimeException("cannot start with locked utxos: see logs for detailed instructions to fix it")
        } else {
          logger.warn(s"ignoring locked utxos: ${lockedOutpoints.map(o => s"${o.txid}:${o.index}").mkString(", ")}")
          Future.successful(true)
        }
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/2278/commits/3d42fcd7255e213c30f4542053d96d7a6c6f4525

t-bast:
Rebased to fix the build issue.

pm47 (approved):

PR #2284, opened 2022-05-25
Don't block when generating shutdown script

t-bast:
We were previously using a blocking call, which can create issues on failures and timeouts.


eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -144,23 +145,31 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
...
-          val temporaryChannelId = randomBytes32()
-          val channelFeeratePerKw = nodeParams.onChainFeeConf.getCommitmentFeerate(remoteNodeId, channelType, c.fundingSatoshis, None)
-          val fundingTxFeeratePerKw = c.fundingTxFeeratePerKw_opt.getOrElse(nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(target = nodeParams.onChainFeeConf.feeTargets.fundingBlockTarget))
-          log.info(s"requesting a new channel with type=$channelType fundingSatoshis=${c.fundingSatoshis}, pushMsat=${c.pushMsat} and fundingFeeratePerByte=${c.fundingTxFeeratePerKw_opt} temporaryChannelId=$temporaryChannelId localParams=$localParams")
-          channel ! INPUT_INIT_FUNDER(temporaryChannelId, c.fundingSatoshis, c.pushMsat, channelFeeratePerKw, fundingTxFeeratePerKw, localParams, d.peerConnection, d.remoteInit, c.channelFlags.getOrElse(nodeParams.channelConf.channelFlags), channelConfig, channelType)
-          stay() using d.copy(channels = d.channels + (TemporaryChannelId(temporaryChannelId) -> channel))
+          // NB: we need to capture parameters in a val to use them in andThen
+          val (selfRef, origin) = (self, sender())
```
pm47, line 151: Putting those on the same line seems a little footgunny given they have the same type.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -144,23 +145,31 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
...
+          implicit val ec: ExecutionContext = ExecutionContext.Implicits.global
```
pm47, line 152: You are not using the actorsystem's dispatcher, which is what we want here 👍

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -144,23 +145,31 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
...
+          createLocalParams(nodeParams, d.localFeatures, channelType, isInitiator = true, c.fundingSatoshis).andThen {
+            case Success(localParams) => selfRef ! SpawnChannelInitiator(c, ChannelConfig.standard, channelType, localParams, origin)
+            case Failure(t) => origin ! Status.Failure(new RuntimeException(s"channel creation failed: reason=${t.getMessage}"))
```
pm47, line 155: As suggested by @sstone 

```suggestion
            case Failure(t) => origin ! Status.Failure(new RuntimeException("channel creation failed", t))
```

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -191,6 +202,14 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
             stay()
         }
 
+      case Event(SpawnChannelNonInitiator(msg, channelConfig, channelType, localParams), d: ConnectedData) =>
```
pm47, line 205: Nit:

```suggestion
      case Event(SpawnChannelNonInitiator(open, channelConfig, channelType, localParams), d: ConnectedData) =>
```

pm47 (changes requested):
`SpawnChannel*` need to be handled in state `DISCONNECTED`, when the future returns successfully, but too late.

t-bast:
> SpawnChannel* need to be handled in state DISCONNECTED, when the future returns successfully, but too late.

Done and rebased.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2284/commits/5b956cb456de1ac54175013116985654f06622f9

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2284/commits/5b956cb456de1ac54175013116985654f06622f9

pm47 (approved):

PR #2285, opened 2022-05-25
Add option to disable sending remote address in init

### LINKED ISSUE #2256: Fill in `init.remote_addr` from front machines
When running in `cluster` mode, we shouldn't return the `remote_addr` of one of our front machines, this should be written by the front machines themselves and match the real sender's IP address.

See https://github.com/lightning/bolts/pull/917#issuecomment-1118473439

This needs some investigation, maybe there is something else that messes things up (e.g. a load balancer's IP address being reported).

t-bast:
When running behind a load balancer that doesn't preserve the client source IP address, it doesn't make sense to include the `remote-address` tlv in our `init` message, it will contain an IP address that is completely unrelated to our peer.

Fixes #2256

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
@@ -31,7 +31,7 @@ import fr.acinq.eclair.wire.protocol.CommonCodecs._
 import fr.acinq.eclair.wire.protocol.LightningMessageCodecs._
 import fr.acinq.eclair.wire.protocol.QueryChannelRangeTlv.queryFlagsCodec
 import fr.acinq.eclair.wire.protocol._
-import fr.acinq.eclair.{CltvExpiryDelta, Feature, Features, InitFeature}
+import fr.acinq.eclair.{CltvExpiryDelta, Feature, Features}
 import scodec._
 import scodec.codecs._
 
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
```diff
@@ -108,7 +108,8 @@ object EclairInternalsSerializer {
       ("pingDisconnect" | bool(8)) ::
       ("maxRebroadcastDelay" | finiteDurationCodec) ::
       ("killIdleDelay" | finiteDurationCodec) ::
-      ("maxOnionMessagesPerSecond" | int32)).as[PeerConnection.Conf]
+      ("maxOnionMessagesPerSecond" | int32) ::
+      ("sendRemoteAddressInit" | bool(8))).as[PeerConnection.Conf]
```
pm47, line 112: Should be `bool8` but apparently we have other "violations" in that file.

pm47 (approved):
I'm not sure this is really necessary (we could fix our deployment, it's as simple as advertising a different public ip), but it doesn't cost much to support _not_ sending the tlv, so I'll leave it to you to decide.

eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
t-bast: I saw that, that's why I chose to be consistent instead of using `bool8`. I'll do a follow-up tiny refactoring to move to `bool8` everywhere in another PR.

PR #2286, opened 2022-05-27
Define a proper base class for fixture tests

pm47:
This PR does two main things:
- introduce a new `FixtureSpec` base class for tests that involve a fixture. See the scaladoc for more info. ~~The existing test `WaitForOpenChannelStateSpec` has been migrated as an example. The diff is small but we separate clearly the fixtures from the test framework. There is duplication for now, but the goal is for `ChannelStateTestsHelperMethods` to disappear.~~
- add new simple integration tests in package `integration.basic`. They are based on `MinimalNodeFixture`, which is a full setup for a node with real actors, except the bitcoin part (watcher/wallet) which is mocked. They are much lighter than our previous integration tests, which allow us to keep each test individual, as opposed to having all tests of the same suite depend on each other. We can define more complex fixtures with any number of nodes.

Also updated scalatest version and fixed a few issues.


eclair-core/src/test/scala/fr/acinq/eclair/blockchain/DummyOnChainWallet.scala
```diff
@@ -40,8 +41,11 @@ class DummyOnChainWallet extends OnChainWallet {
 
   override def getReceivePubkey(receiveAddress: Option[String] = None)(implicit ec: ExecutionContext): Future[Crypto.PublicKey] = Future.successful(dummyReceivePubkey)
 
-  override def makeFundingTx(pubkeyScript: ByteVector, amount: Satoshi, feeRatePerKw: FeeratePerKw)(implicit ec: ExecutionContext): Future[MakeFundingTxResponse] =
-    Future.successful(DummyOnChainWallet.makeDummyFundingTx(pubkeyScript, amount))
+  override def makeFundingTx(pubkeyScript: ByteVector, amount: Satoshi, feeRatePerKw: FeeratePerKw)(implicit ec: ExecutionContext): Future[MakeFundingTxResponse] = {
+    val tx = DummyOnChainWallet.makeDummyFundingTx(pubkeyScript, amount)
+    funded += (tx.fundingTx.txid -> tx.fundingTx)
+    Future.successful(tx)
```
t-bast, line 47: I'm making many changes to these test helpers in #2275, if you plan on making more changes here in follow-up PRs we should probably do it after dual-funding to minimize conflicts.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/MinimalNodeFixture.scala
```diff
@@ -0,0 +1,272 @@
...
+      channelKeyManager = new LocalChannelKeyManager(seed, Block.RegtestGenesisBlock.hash),
+      torAddress_opt = None,
+      database = TestDatabases.inMemoryDb(),
+      blockHeight = new AtomicLong(400_000),
+      feeEstimator = new TestFeeEstimator
+    )
+      .modify(_.alias).setTo(alias)
+      .modify(_.chainHash).setTo(Block.RegtestGenesisBlock.hash)
+      .modify(_.routerConf.routerBroadcastInterval).setTo(1 second)
+      .modify(_.peerConnectionConf.maxRebroadcastDelay).setTo(1 second)
```
t-bast, line 67: This indenting is a bit weird...

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/MinimalNodeFixture.scala
```diff
@@ -0,0 +1,272 @@
...
+    sender.expectMsgType[Router.Data]
+  }
+
+  private def deterministicShortId(txId: ByteVector32): (BlockHeight, Int) = {
+    val blockHeight = txId.take(3).toInt(signed = false)
+    val txIndex = txId.takeRight(2).toInt(signed = false)
+    (BlockHeight(blockHeight), txIndex)
+  }
+
+  /** */
```
t-bast, line 229: nit: remove or enrich?

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/ThreeNodesFixture.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import akka.actor.ActorSystem
+import akka.testkit.TestKit
+import fr.acinq.eclair.NodeParams
+
+
+case class ThreeNodesFixture(system: ActorSystem,
+                             alice: MinimalNodeFixture,
+                             bob: MinimalNodeFixture,
+                             charlie: MinimalNodeFixture) {
```
t-bast, line 11: What a hostile takeover! Are we deprecating poor Carol in favor of Charlie?

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/TwoNodesIntegrationSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+ * This test checks the integration between Channel and Router (events, etc.)
+ */
+class TwoNodesIntegrationSpec extends FixtureSpec with IntegrationPatience {
+
+  type FixtureParam = TwoNodesFixture
+
+  import fr.acinq.eclair.integration.basic.fixtures.MinimalNodeFixture._
+
+  override def createFixture(testData: TestData): FixtureParam = {
+    // seeds have been chose so that node ids start with 02aaaa for alice, 02bbbb for bob, etc.
```
t-bast, line 23: nit:
```suggestion
    // seeds have been chosen so that node ids start with 02aaaa for alice, 02bbbb for bob, etc.
```

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/TwoNodesIntegrationSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+import fr.acinq.eclair.{BlockHeight, MilliSatoshiLong}
+import org.scalatest.TestData
+import org.scalatest.concurrent.IntegrationPatience
+import scodec.bits.HexStringSyntax
+
+
+/**
+ * This test checks the integration between Channel and Router (events, etc.)
+ */
+class TwoNodesIntegrationSpec extends FixtureSpec with IntegrationPatience {
```
t-bast, line 16: So these integration tests should in theory be much slower than our current integration tests if each tests re-does channel setup, right? How much of an issue is it in practice? It's really a pain that our current integration tests depend on the previous tests, but last time we tried to get rid of it the performance was too bad...

t-bast (commented):
I like the direction in which we're going with that change, and I'm happy to merge that PR and start using these new fixtures for new code, but before migrating old code or removing `ChannelStateTestHelperMethods` we should really integrate dual-funding which touches those a lot, otherwise we'll make a complex set of PRs even more complex to integrate!

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/DummyOnChainWallet.scala
pm47: Gotcha, I don't plan to make more changes so we should be fine. Dealing with the mempool/blockchain is one of the trickiest parts, because they need to be shared among various `MinimalNodeFixture`. I have explored several approaches and am still not 100% happy.

In particular, there are two ways of confirming channels: `MinimalNodeFixture.confirmChannel` and ``MinimalNodeFixture.autoConfirmLocalChannels`. I think the latter (based on autopilot) is nicer when you don't need fine-grained control, but the core issue of having access to the actual transaction remains. For example, note that we pass a reference to alice's wallet for bob's autopilot:
https://github.com/ACINQ/eclair/blob/2e6e566ccc0dbaa49ed9678634b33df0d97f216a/eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/TwoNodesIntegrationSpec.scala#L50

Maybe with dual-funding this will be a tad simpler since both sides know the full tx?

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/MinimalNodeFixture.scala
pm47: Enriched in 3d42e3af8.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/ThreeNodesFixture.scala
pm47: Sorry Carol! Not sure why I switched to the NATO phonetic alphabet. And it's not the first time, did the same in other tests apparently.

Fixed in 3d42e3af8.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/TwoNodesIntegrationSpec.scala
pm47: Fixed in 3d42e3af8.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/TwoNodesIntegrationSpec.scala
pm47: I don't think they will be slower, quite the opposite. Our standard integration tests use a real bitcoind, which takes a long time to start up and initialize. Those tests are really light, and since they offer foolproof way of cleaning up fixtures, we don't end up consuming huge amount of resources and can parallelize better.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -108,7 +113,13 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
       }
       d.transport ! localInit
       startSingleTimer(INIT_TIMER, InitTimeout, conf.initTimeout)
+      unstashAll() // unstash remote init if it already arrived
```
pm47, line 116: NB: I'm only `unstashing` right before we are ready to process the remote `init`.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/MinimalNodeFixture.scala
pm47: Did differently in 3d42e3af8, but now lines are not aligned 🙈 .

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/TwoNodesIntegrationSpec.scala
pm47: A nice side effect of this trick is that node ids are now ordered, which means that in channel alice-bob, update_1 will always be for alice, and update_2 for bob. That is *really* convenient for debugging the router.

pm47 (commented):
I did some cleanup which shouldn't impact your dual-funding tests.

t-bast (approved):

PR #2290, opened 2022-05-29
Prometheus Monitoring

GoutamVerma:
- Added dependencies for Prometheus and HTTP server
- Added metrics for blockchain monitoring
- All metrics will be expose on "localhost:1234"

### DIFF
--- a/eclair-core/pom.xml
+++ b/eclair-core/pom.xml
@@ -271,6 +271,26 @@
         </dependency>
         <!-- MONITORING -->
         <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient</artifactId>
+            <version>0.15.0</version>
+        </dependency>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient_hotspot</artifactId>
+            <version>0.15.0</version>
+        </dependency>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient_httpserver</artifactId>
+            <version>0.15.0</version>
+        </dependency>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient_pushgateway</artifactId>
+            <version>0.15.0</version>
+        </dependency>
+        <dependency>
             <groupId>io.kamon</groupId>
             <artifactId>kamon-core_${scala.version.short}</artifactId>
             <version>${kamon.version}</version>
--- a/eclair-core/pom.xml
+++ b/eclair-core/pom.xml
@@ -270,26 +270,11 @@
             <version>1.5.0</version>
         </dependency>
         <!-- MONITORING -->
-        <dependency>
-            <groupId>io.prometheus</groupId>
-            <artifactId>simpleclient</artifactId>
-            <version>0.15.0</version>
-        </dependency>
-        <dependency>
-            <groupId>io.prometheus</groupId>
-            <artifactId>simpleclient_hotspot</artifactId>
-            <version>0.15.0</version>
-        </dependency>
-        <dependency>
-            <groupId>io.prometheus</groupId>
-            <artifactId>simpleclient_httpserver</artifactId>
-            <version>0.15.0</version>
-        </dependency>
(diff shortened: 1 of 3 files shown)


eclair-core/pom.xml
```diff
@@ -270,6 +270,26 @@
...
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient_httpserver</artifactId>
+            <version>0.15.0</version>
+        </dependency>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient_pushgateway</artifactId>
+            <version>0.15.0</version>
+        </dependency>
```
t-bast, line 292: Why do you need that? According to [Kamon's doc](https://kamon.io/docs/latest/reporters/prometheus/) you should only need to import:

```
<dependency>
  <groupId>io.kamon</groupId>
  <artifactId>kamon-prometheus_2.13</artifactId>
  <version>2.5.3</version>
</dependency>
```

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/Monitoring.scala
```diff
@@ -16,12 +16,20 @@
...
 import kamon.Kamon
 import kamon.metric.Metric
 
 object Monitoring {
 
   object Metrics {
+    //Prometheus Metrics
+    val MempoolMinFeeratePerKwPrometheus: Gauge  = Gauge.build().name("bitcoin_mempool_min_feerate_per_kw").help( "Minimum feerate (sat/kw) for a tx to be accepted in our mempool").register()
+    val CannotRetrieveFeeratesCountPrometheus: Counter  = Counter.build().name("bitcoin_rpc_feerates_error").help( "Number of failures to retrieve on-chain feerates").register()
+    val server: HTTPServer = new HTTPServer.Builder().withPort(1234).build();
```
t-bast, line 30: That's not the right way to do this. We already have metrics defined, you shouldn't create new ones. What you need to do is to ensure that all the existing metrics are correctly sent to prometheus if you define a prometheus endpoint.

You're doing it backwards: you are currently creating an in-process prometheus endpoint. What I'd like you to do is the reverse: send metrics to a prometheus endpoint that is potentially running on a remote machine.

Ideally, a node operator would provide the IP address of a prometheus endpoint in `eclair.conf`, and we would just automatically send metrics to that endpoint with minimal code additions to the [boot process](https://github.com/ACINQ/eclair/blob/9c7ddcd8745bbea9c7113457e1a1f378435347c7/eclair-node/src/main/scala/fr/acinq/eclair/Boot.scala#L46)

eclair-core/pom.xml
GoutamVerma: I have included dependencies suggested by [Kamon doc](https://kamon.io/docs/latest/reporters/prometheus/) and as a result, the metrics exposed on port localhost:9095/metrics and we can  change the exposing port in eclair.conf file. 

After this all. Pulling over metrics over HTTP offers a number of advantages but it should not be considered a major point when considering a monitoring system. According to task and discussion I suppose to implement [remote write tunning](https://prometheus.io/docs/practices/remote_write/) right?

As suggested by this documentation [remote_write](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write) in yaml file

If I am correct then I will add short and crisp code in boot process

t-bast: I think you're still misunderstanding the goal of this feature. What I'd like you to do is setup eclair to be able to export the **existing** metrics (not add new ones!) to a **remote** prometheus endpoint. Nothing more! You should definitely *not* create a prometheus http server inside eclair.

GoutamVerma: Okay got it
But before exporting to remote system. I tried to export it on localhost
and I found some issues maybe due to existing metrics.

it shows 
down

I tried to figure out problem with the help [promtool](https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/)(which is used for testing rules).

and got the output
output

due to metric which contains
port 8080

It is better to discuss whole things in meet at Monday,
see you there!

Issue #2292, opened 2022-05-29
Enhance GraphSpec tests


barrybecker4:
I would like to add some additional tests to GraphSpec. At a minimum, I would like to add 
 - a test that finds the shortest paths when "amount with fees smaller than channel capacity" .
 - a test that finds the shortest paths given a graph like the example used in the[ wikipedia page for Yen's algorithm](https://en.wikipedia.org/wiki/Yen%27s_algorithm#:~:text=Yen's%20algorithm%20computes%20single%2Dsource,deviations%20of%20the%20best%20path.).

If this goes well, I may later try to do some refactoring of the Graph code. It seems like it could be split into a set of smaller classes that would be easier to understand.

barrybecker4:
Yes, closing now that the [PR](https://github.com/ACINQ/eclair/pull/2293) is merged.

PR #2293, opened 2022-05-30
Add more GraphSpec tests (#2292)

barrybecker4:
I added the 2 tests that I suggested but noticed something surprising.

The test for the [wikipedia example](https://en.wikipedia.org/wiki/Yen%27s_algorithm#:~:text=Yen's%20algorithm) is not returning the expected results. I suspect that there is at least one bug. I'm reluctant to make the change in the production code because of my current lack of understanding, but if I remove the `reverse` from [this line](https://github.com/ACINQ/eclair/blob/9c7ddcd8745bbea9c7113457e1a1f378435347c7/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala#L227), then at least the call to `dijkstraShortestPath` returns the shortest path that was expected. But even after making that fix, assuming that it is a correct fix, the next 2 shortest paths also do not match what is in the wikipedia article and I have not tried to debug that yet.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
@@ -265,7 +265,8 @@ class GraphSpec extends AnyFunSuite {
 
   test("fees less along C->D->E than C->E") {
     /*
-    The channel C -> D is large enough for the payment and associated fees to go through, but C -> E is not.
+    The channel going through C -> D -> E  (fee = 1001 + 1 = 1002) is considered shorter than going
+    through C -> E (fee = 1003) because the fees are less.
     A --> B --> C <-> D
                  \   /
                   \ /
@@ -285,7 +286,7 @@ class GraphSpec extends AnyFunSuite {
       BlockHeight(714930), _ => true, includeLocalChannelCost = true)
 
     assert(paths.length == 2)
-    assert(paths.head.path == Seq(edgeAB, edgeBC, edgeCD, edgeDE))
+    assert(paths(0).path == Seq(edgeAB, edgeBC, edgeCD, edgeDE))
     assert(paths(1).path == Seq(edgeAB, edgeBC, edgeCE))
   }
 
@@ -317,25 +318,25 @@ class GraphSpec extends AnyFunSuite {
 
   /**
    * Find all the shortest paths using the example described in
-   * https://en.wikipedia.org/wiki/Yen%27s_algorithm#:~:text=Yen's%20algorithm
+   * https://en.wikipedia.org/wiki/Yen's_algorithm#Example
    */
   test("all shortest paths are found") {
     // There will be 3 shortest paths.
     // Edge capacities are set to be the same so that only feeBase will affect the RichWeight.
-    //  C --> D --> F
-    //   \    ^    /|\
-    //    \   |  /  | \
-    //     \ | /    |  \
-    //       E----->G-->H
-    val edgeCD = makeEdge(1L, c, d, 301 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
-    val edgeDF = makeEdge(2L, d, f, 401 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
-    val edgeCE = makeEdge(3L, c, e, 201 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
-    val edgeED = makeEdge(4L, e, d, 101 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
-    val edgeEF = makeEdge(5L, e, f, 201 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
-    val edgeFG = makeEdge(6L, f, g, 201 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
-    val edgeFH = makeEdge(7L, f, h, 101 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
-    val edgeEG = makeEdge(8L, e, g, 301 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
-    val edgeGH = makeEdge(9L, g, h, 201 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    // C --> D --> F
+    //   \   ^   / | \
+    //    \  |  /  |  \
+    //     \ | /   |   \
+    //       E --> G --> H
+    val edgeCD = makeEdge(1L, c, d, 3 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeDF = makeEdge(2L, d, f, 4 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeCE = makeEdge(3L, c, e, 2 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeED = makeEdge(4L, e, d, 1 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeEF = makeEdge(5L, e, f, 2 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeFG = makeEdge(6L, f, g, 2 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeFH = makeEdge(7L, f, h, 1 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeEG = makeEdge(8L, e, g, 3 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeGH = makeEdge(9L, g, h, 2 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
     val graph = DirectedGraph(Seq(edgeCD, edgeDF, edgeCE, edgeED, edgeEF, edgeFG, edgeFH, edgeEG, edgeGH))
 
     val paths = yenKshortestPaths(graph, c, h, 10000000 msat,
@@ -343,7 +344,7 @@ class GraphSpec extends AnyFunSuite {
       Left(WeightRatios(1, 0, 0, 0, RelayFees(0 msat, 0))),
       BlockHeight(714930), _ => true, includeLocalChannelCost = true)
     assert(paths.length == 3)
-    assert(paths.head.path == Seq(edgeCE, edgeEF, edgeFH))
+    assert(paths(0).path == Seq(edgeCE, edgeEF, edgeFH))
     assert(paths(1).path == Seq(edgeCE, edgeEG, edgeGH))
     assert(paths(2).path == Seq(edgeCD, edgeDF, edgeFH))
   }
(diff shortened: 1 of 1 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -255,10 +256,70 @@ class GraphSpec extends AnyFunSuite {
...
+    //       E----->G-->H
+    val edgeCD = makeEdge(1L, c, d, 10 msat, 0, capacity = 301000 sat, minHtlc = 1000 msat)
+    val edgeDF = makeEdge(2L, d, f, 10 msat, 0, capacity = 401000 sat, minHtlc = 1000 msat)
+    val edgeCE = makeEdge(3L, c, e, 10 msat, 0, capacity = 201000 sat, minHtlc = 1000 msat)
+    val edgeED = makeEdge(4L, e, d, 10 msat, 0, capacity = 101000 sat, minHtlc = 1000 msat)
+    val edgeEF = makeEdge(5L, e, f, 10 msat, 0, capacity = 201000 sat, minHtlc = 1000 msat)
+    val edgeFG = makeEdge(6L, f, g, 10 msat, 0, capacity = 201000 sat, minHtlc = 1000 msat)
+    val edgeFH = makeEdge(7L, f, h, 10 msat, 0, capacity = 101000 sat, minHtlc = 1000 msat)
+    val edgeEG = makeEdge(8L, e, g, 10 msat, 0, capacity = 301000 sat, minHtlc = 1000 msat)
+    val edgeGH = makeEdge(9L, g, h, 10 msat, 0, capacity = 201000 sat, minHtlc = 1000 msat)
```
thomash-acinq, line 311: This is absolutely not the wikipedia example. You matched the costs from the wikipedia example to the capacities but the capacities are actually inversely correlated with the cost.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -255,10 +256,70 @@ class GraphSpec extends AnyFunSuite {
...
+    val graph = DirectedGraph(Seq(edgeCD, edgeDF, edgeCE, edgeED, edgeEF, edgeFG, edgeFH, edgeEG, edgeGH))
+
+    val paths = yenKshortestPaths(graph, c, h, 10000000 msat,
+      Set.empty, Set.empty, Set.empty, 3,
+      Right(HeuristicsConstants(1.0E-8, RelayFees(2000 msat, 500), RelayFees(50 msat, 20), useLogProbability = true)),
+      BlockHeight(714930), _ => true, includeLocalChannelCost = true)
+    assert(paths.length == 3)
+    // Expect the 3 shortest paths to be CEFH(5), CEGH(7), CDFH(8)
+    // Expect shortest to be CEFH, but getting CEGH
```
thomash-acinq, line 320: CEGH has a lower cost than CEFH:
```
pathWeight(c, Seq(edgeCE, edgeEG, edgeGH), 10000000 msat, ...) == RichWeight(..., 1730.9506767500889)
pathWeight(c, Seq(edgeCE, edgeEF, edgeFH), 10000000 msat, ...) == RichWeight(..., 2224.268891921596)
```

thomash-acinq (commented):
It seems that you don't understand how we compute the cost of a path. You should look at `Graph.addEdgeWeight`. If you add test, you should make sure that you use the correct scoring.

thomash-acinq:
To test `yenKshortestPaths` and `dijkstraShortestPath` without worrying about the complex scoring of paths, you can use fees only which makes it easier to compute by hand: set all proportional fees to 0, only set the base fee and for `wr`, use `Left(WeightRatios(1, 0, 0, 0, RelayFees(0 msat, 0)))`

barrybecker4:
@thomash-acinq , thanks for the tips. I still don't understand much, but I did manage to get my additional tests to pass based on your suggestions. I rebased/squashed.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -261,4 +262,89 @@ class GraphSpec extends AnyFunSuite {
...
+    A --> B --> C <-> D
+                 \   /
+                  \ /
+                   E
+    */
+    val edgeAB = makeEdge(1L, a, b, 10001 msat, 0, capacity = 200000 sat)
+    val edgeBC = makeEdge(2L, b, c, 10000 msat, 0, capacity = 200000 sat)
+    val edgeCD = makeEdge(3L, c, d, 10001 msat, 0, capacity = 200000 sat)
+    val edgeDC = makeEdge(4L, d, c, 10 msat, 0, capacity = 200000 sat)
+    val edgeCE = makeEdge(5L, c, e, 10003 msat, 0, capacity = 200000 sat)
```
thomash-acinq, line 278: You wrote in the comment that "C -> D is large enough but C -> E is not" but C -> D and C -> E have the same capacity.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -261,4 +262,89 @@ class GraphSpec extends AnyFunSuite {
...
+    val edgeDE = makeEdge(6L, d, e, 1 msat, 0, capacity = 200000 sat)
+    val graph = DirectedGraph(Seq(edgeAB, edgeBC, edgeCD, edgeDC, edgeCE, edgeDE))
+
+    val paths = yenKshortestPaths(graph, a, e, 90000000 msat,
+      Set.empty, Set.empty, Set.empty, 2,
+      Left(WeightRatios(1, 0, 0, 0, RelayFees(0 msat, 0))),
+      BlockHeight(714930), _ => true, includeLocalChannelCost = true)
+
+    assert(paths.length == 2)
+    assert(paths.head.path == Seq(edgeAB, edgeBC, edgeCD, edgeDE))
```
thomash-acinq, line 288:
```suggestion
    assert(paths.(0).path == Seq(edgeAB, edgeBC, edgeCD, edgeDE))
```
to keep it consistent with the next line.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -261,4 +262,89 @@ class GraphSpec extends AnyFunSuite {
...
+   * https://en.wikipedia.org/wiki/Yen%27s_algorithm#:~:text=Yen's%20algorithm
+   */
+  test("all shortest paths are found") {
+    // There will be 3 shortest paths.
+    // Edge capacities are set to be the same so that only feeBase will affect the RichWeight.
+    //  C --> D --> F
+    //   \    ^    /|\
+    //    \   |  /  | \
+    //     \ | /    |  \
+    //       E----->G-->H
```
thomash-acinq, line 329:
```suggestion
    // C --> D --> F
    //   \   ^   / | \
    //    \  |  /  |  \
    //     \ | /   |   \
    //       E --> G --> H
```

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -261,4 +262,89 @@ class GraphSpec extends AnyFunSuite {
...
+    val edgeCD = makeEdge(1L, c, d, 301 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeDF = makeEdge(2L, d, f, 401 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeCE = makeEdge(3L, c, e, 201 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
```
thomash-acinq, line 332: Why 301, 401, 201 and not just 3, 4, 2?

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -261,4 +262,89 @@ class GraphSpec extends AnyFunSuite {
...
+      BlockHeight(714930), _ => true, includeLocalChannelCost = true)
+
+    // Even though paths to find is 2, we only find 1 because that is all the valid paths that there are.
+    assert(paths.length == 1)
+    assert(paths.head.path == Seq(edgeAB, edgeBC, edgeCD, edgeDE))
+  }
+
+  /**
+   * Find all the shortest paths using the example described in
+   * https://en.wikipedia.org/wiki/Yen%27s_algorithm#:~:text=Yen's%20algorithm
```
thomash-acinq, line 320:
```suggestion
   * https://en.wikipedia.org/wiki/Yen's_algorithm#Example
```

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -261,4 +262,89 @@ class GraphSpec extends AnyFunSuite {
...
+    val edgeEG = makeEdge(8L, e, g, 301 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val edgeGH = makeEdge(9L, g, h, 201 msat, 0, capacity = 100000 sat, minHtlc = 1000 msat)
+    val graph = DirectedGraph(Seq(edgeCD, edgeDF, edgeCE, edgeED, edgeEF, edgeFG, edgeFH, edgeEG, edgeGH))
+
+    val paths = yenKshortestPaths(graph, c, h, 10000000 msat,
+      Set.empty, Set.empty, Set.empty, 3,
+      Left(WeightRatios(1, 0, 0, 0, RelayFees(0 msat, 0))),
+      BlockHeight(714930), _ => true, includeLocalChannelCost = true)
+    assert(paths.length == 3)
+    assert(paths.head.path == Seq(edgeCE, edgeEF, edgeFH))
```
thomash-acinq, line 346:
```suggestion
    assert(paths.(0).path == Seq(edgeCE, edgeEF, edgeFH))
```

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
barrybecker4: I changed the comment to 
```
The channel going through C -> D -> E  (fee = 1001 + 1 = 1002) is considered shorter than going 
    through C -> E (fee = 1003) because the fees are less.
```

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
barrybecker4: Updated it to use the index for consistency. I had changed it to head only because Intellij had colored it yellow.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
barrybecker4: No good reason apparently. I changed it to the smaller values and it still worked. For some reason, I was thinking that the behavior would be different if they were too small.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
barrybecker4: Updated the URL.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
barrybecker4: Done.

barrybecker4:
@thomash-acinq , Thanks for the review. I made the suggested changes. Sorry for the delay. My day job only allows me to look at this on weekends.

thomash-acinq (approved):
Thanks Barry

Issue #2260, opened 2022-05-05
Error "got exception for batch of"


engenegr:
Error, initially discovered on Testnet node

```
2022-05-05 16:39:36,157 ERROR f.a.e.b.b.r.BatchingClient - got exception for batch of 49 requestssttp.client3.SttpClientException$ConnectException: Exception when sending request: POST http://192.168.2.6:18332/wallet/%2Fhome%2Fu%2F.bitcoin%2Ftestnet3%2Feclair
	at sttp.client3.SttpClientException$.defaultExceptionToSttpClientException(SttpClientException.scala:40)
	at sttp.client3.okhttp.OkHttpBackend$.exceptionToSttpClientException(OkHttpBackend.scala:175)
	at sttp.client3.okhttp.OkHttpBackend.$anonfun$adjustExceptions$1(OkHttpBackend.scala:50)
	at sttp.client3.SttpClientException$$anonfun$adjustExceptions$1.applyOrElse(SttpClientException.scala:59)
	at sttp.client3.SttpClientException$$anonfun$adjustExceptions$1.applyOrElse(SttpClientException.scala:58)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:490)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinPool.scan(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(Unknown Source)
Caused by: java.net.ProtocolException: unexpected end of stream
	at okhttp3.internal.http1.Http1ExchangeCodec$FixedLengthSource.read(Http1ExchangeCodec.kt:371)
	at okhttp3.internal.connection.Exchange$ResponseBodySource.read(Exchange.kt:276)
	at okio.RealBufferedSource$inputStream$1.read(RealBufferedSource.kt:158)
	at sttp.client3.internal.package$.transfer$1(package.scala:23)
	at sttp.client3.internal.package$.transfer(package.scala:30)
	at sttp.client3.internal.package$.toByteArray(package.scala:35)
[...]

Issue #2259, opened 2022-05-05
Eclair cant keep up with Core


engenegr:
My node constantly sends `DangerousBlocksSkew` events while apparently late on tip.

The node shouldn't be late.

In particular situation I've seen `735008` in output of `getinfo` while Core API was returning `735015` from the same host where Eclair is installed.

I've seen such behavior several times and verified additionally by curling Core API. The Bitcoin client has been in sync always.

To reproduce the issue one supposedly should have remotely operated Core node in VPN network.

Compiled, `0.7.0-a804905`.

Server with Linux OS installed.

t-bast:
Eclair's behavior is to:

- receive blocks over zmq socket
- if no blocks have been received for 15 minutes, actively request them from `bitcoind`

The second point is here to compensate failures to receive on zmq (which can happen on unreliable networks).

Here is the median time when these blocks were founds (taken from `mempool.space`):

- 735008: 2022-05-05 12:56
- 735009: 2022-05-05 13:28
- 735010: 2022-05-05 13:30
- 735011: 2022-05-05 13:31
- 735012: 2022-05-05 13:41
- 735013: 2022-05-05 13:42
- 735014: 2022-05-05 13:47
- 735015: 2022-05-05 13:53

Notice that there is a long delay between block 735008 and 735009, then three blocks are found very close to each other.
In your case, we've seen that your network is quite unreliable, so you're probably not always receiving blocks over the zmq socket and instead relying on eclair actively fetching them from `bitcoind` every 15 minutes. You also probably have a small delay in receiving blocks compared to the best connected nodes in the network (such as explorers, which we use as watchdogs). That means there will be some time during which you will be behind the tip, which explains why you're seeing `DangerousBlocksSkew` events.

[...]

PR #2301, opened 2022-06-06
Configurable watchdog and batching

engenegr:
Addresses https://github.com/ACINQ/eclair/issues/2260 and https://github.com/ACINQ/eclair/issues/2259.
Not tested yet on mainnet node with remote Bitcoind.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -350,7 +350,7 @@ eclair {
       }
       safety-checks {
         // A set of basic checks on data to make sure we use the correct database
-        // Those checks are disabled by default because they wouldnt pass on a fresh new node with
+        // Those checks are disabled by default because they would not pass on a fresh new node with
         // zero channels. You should enable them when you already have channels, so that there is
         // something to compare to, and the values should be specific to your setup, especially
         // for local channels. If your operate a busy node, you can reduce max-age.local-channels
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -385,13 +385,14 @@ eclair {
   blockchain-watchdog {
     // all available blockchain watchdog sources
     // you can remove items from this list to disable failing or non-needed sources
-    tolerance = 7
     sources = [
       "bitcoinheaders.net",
       "blockcypher.com",
       "blockstream.info",
       "mempool.space"
     ]
+    // maximum lag of chain height observed by eclair to blockchain watchdog sources
+    missing-blocks-threshold = 7
   }
 
   onion-messages {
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/resources/reference.conf
```diff
@@ -347,7 +348,7 @@ eclair {
       }
       safety-checks {
         // A set of basic checks on data to make sure we use the correct database
-        // Those checks are disabled by default because they wouldn't pass on a fresh new node with
+        // Those checks are disabled by default because they wouldnt pass on a fresh new node with
```
t-bast, line 351: The previous sentence was correct, this one isn't, revert?

eclair-core/src/main/resources/reference.conf
```diff
@@ -29,6 +29,7 @@ eclair {
     wallet = ""
     zmqblock = "tcp://127.0.0.1:29000"
     zmqtx = "tcp://127.0.0.1:29000"
+    batching = true
```
t-bast, line 32: Maybe rename and document:

```suggestion
    // Batching requests saves bandwidth but may slightly degrade latency and reliability.
    // You may want to disable this when bitcoin is running on a remote machine with an unreliable network.
    batch-requests = true
```

eclair-core/src/main/resources/reference.conf
```diff
@@ -382,6 +383,7 @@ eclair {
   blockchain-watchdog {
     // all available blockchain watchdog sources
     // you can remove items from this list to disable failing or non-needed sources
+    tolerance = 7
```
t-bast, line 386: You should probably move that below the `sources` list, now we don't understand what the documentation above applies to. You should also add a comment to explain what that field does.

I would rename it to `missing-blocks-threshold` which is more explicit (and please rename the corresponding field in the code as well).

t-bast (commented):
Thanks, that looks simple enough! I'd like to rename the fields for clarity and document them better, but otherwise looks good to me, let me know if it improves the reliability of your setup.

eclair-core/src/main/resources/reference.conf
engenegr: Sorry, I found out that, if copied entirely, your `reference.conf` leads to errors. And the error happens due to this symbol.
I will revert if you strongly disagree, of course.

eclair-core/src/main/resources/reference.conf
engenegr: Corrected

eclair-core/src/main/resources/reference.conf
t-bast: That's very surprising, I don't reproduce that...at least if you change it change it to proper english: `wouldn't` -> `would not`

engenegr: Fixed

t-bast (approved):
LGTM, thanks!

PR #2302, opened 2022-06-06
Fix minimal node fixture flakyness

t-bast:
There was a race condition between the `wait_for_funding_locked` state and the `normal` state. By only waiting for the channel to have handled the last message it could either be still in `waiting_for_funding_locked` or already in `normal.


eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/MinimalNodeFixture.scala
```diff
@@ -185,8 +188,10 @@ object MinimalNodeFixture extends Assertions {
     watch1.replyTo ! WatchFundingDeeplyBuriedTriggered(blockHeight, txIndex, fundingTx)
     watch2.replyTo ! WatchFundingDeeplyBuriedTriggered(blockHeight, txIndex, fundingTx)
 
-    waitReady(node1, channelId)
-    waitReady(node2, channelId)
+    eventually {
+      assert(getChannelState(node1, channelId) == NORMAL)
+      assert(getChannelState(node2, channelId) == NORMAL)
+    }
```
pm47, line 194: That doesn't do much since we are already in `NORMAL` when entering this method.

t-bast: Right, I see symmetry between the two methods but actually it doesn't make sense here.

pm47 (approved):

Issue #2298, opened 2022-06-04
Dockerfile does not build for arm64


chrisguida:
`docker build .` inside the main repo dir fails with this error, on my M1 Mac. It also fails similarly in a cross-platform build using docker buildx:

```
% docker build .
[+] Building 1.4s (4/4) FINISHED                                                                                                                                                                            
 => [internal] load build definition from Dockerfile                                                                                                                                                   0.0s
 => => transferring dockerfile: 3.06kB                                                                                                                                                                 0.0s
 => [internal] load .dockerignore                                                                                                                                                                      0.0s
 => => transferring context: 99B                                                                                                                                                                       0.0s
 => ERROR [internal] load metadata for docker.io/adoptopenjdk/openjdk11:jdk-11.0.3_7-alpine                                                                                                            1.4s
 => CANCELED [internal] load metadata for docker.io/library/openjdk:11.0.4-jre-slim                                                                                                                    1.4s
------
 > [internal] load metadata for docker.io/adoptopenjdk/openjdk11:jdk-11.0.3_7-alpine:
------
failed to solve with frontend dockerfile.v0: failed to create LLB definition: no match for platform in manifest sha256:012ff3a1fd9c9237b14f7522307f8f64a19da0f3db37015b3456dafa69a29c0e: not found
```

It appears that the openjdk alpine base image does not compile for arm64. I tried cloning [the openjdk-docker repo](https://github.com/AdoptOpenJDK/openjdk-docker) and building with `./build_all.sh` but that fails as well, with this error:

```
 % ./build_all.sh 
Snyk authentication token not set, skipping snyk analysis
Total reclaimed space: 0B
Total reclaimed space: 0B
==========================================================================================
                                                                                          
  14:27:19 :    Building Docker Images for Version 8 hotspot jdk build  
                                                                                          
==========================================================================================
Snyk authentication token not set, skipping snyk analysis
ERROR: Unsupported arch:arm64, Exiting
###############################################################

ERROR: Docker Build for Version 8 hotspot jdk failed.

###############################################################

WARNING: Pushing to AdoptOpenJDK repo on hub.docker.com
WARNING: If you did not intend this, quit now. (Sleep 5)
```

Eclair's docker image should build for arm64. This includes platforms such as MacOS and Raspberry Pi. I am attempting to pacakge Eclair for [EmbassyOS](https://github.com/Start9Labs/embassy-os), but that will not be possible until Eclair builds for arm64.

See above.

See above.

[...]

PR #2304, opened 2022-06-07
add arm64v8 Dockerfile

chrisguida:
Fixes https://github.com/ACINQ/eclair/issues/2298.

Adapts the official Eclair Dockerfile to Debian. This results in an image that is actually smaller than the official image (430MB vs 421MB).

### DIFF
--- a/arm64v8.Dockerfile
+++ b/arm64v8.Dockerfile
@@ -1,0 +1,65 @@
+FROM arm64v8/eclipse-temurin:11-jdk-jammy as BUILD
+
+# Setup maven, we don't use https://hub.docker.com/_/maven/ as it declare .m2 as volume, we loose all mvn cache
+# We can alternatively do as proposed by https://github.com/carlossg/docker-maven#packaging-a-local-repository-with-the-image
+# this was meant to make the image smaller, but we use multi-stage build so we don't care
+RUN apt update && apt install -y curl tar bash
+
+ARG MAVEN_VERSION=3.6.3
+ARG USER_HOME_DIR="/root"
+ARG SHA=c35a1803a6e70a126e80b2b3ae33eed961f83ed74d18fcd16909b2d44d7dada3203f1ffe726c17ef8dcca2dcaa9fca676987befeadc9b9f759967a8cb77181c0
+ARG BASE_URL=https://apache.osuosl.org/maven/maven-3/${MAVEN_VERSION}/binaries
+
+RUN mkdir -p /usr/share/maven /usr/share/maven/ref \
+  && curl -fsSL -o /tmp/apache-maven.tar.gz ${BASE_URL}/apache-maven-${MAVEN_VERSION}-bin.tar.gz \
+  && echo "${SHA}  /tmp/apache-maven.tar.gz" | sha512sum -c - \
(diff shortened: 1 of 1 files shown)


sstone:
Thanks! LGTM and it seems to be the simplest option to support arm64 target (as you pointed out the eclipse-temurin repo does not provide arm64 alpine-based containesr). I've just done basic tests so far but if the image is not bigger than the alpine one then we should just update our docker file with your changes: can you please update your PR to modify the existing docker file instead of adding a new one ?

chrisguida:
@sstone I think that would break the x86 builds since the `arm64v8/eclipse-temurin:11-jdk-jammy` image is just designed for arm64.

chrisguida:
Confirmed, build took 3 hours on my x86 box, produced an image that is 1.47GB, and when I run it, it does this:

```
$ docker run 57df63754195
WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested
Jun 14, 2022 1:51:17 PM java.util.prefs.FileSystemPreferences$1 run
INFO: Created user preferences directory.
|  Welcome to JShell -- Version 11.0.15
|  For an introduction type: /help intro

jshell> 
```

I did just notice this image though, not sure if you were referencing it specifically? Building on both arches to see what happens: https://hub.docker.com/layers/eclipse-temurin/library/eclipse-temurin/11-jdk-jammy/images/sha256-2c22f55fda06c413474bc0b04b30917eb5bfcaa46d239dcd99d55ead5b28712d?context=explore

chrisguida:
Ok, so with `eclipse-temurin:11-jdk-jammy` instead of `arm64v8/eclipse-temurin:11-jdk-jammy`, the build only took 10 minutes on my x86 box (it's pretty slow), but it produced a very similar image to the above, 1.47GB and has this output when I `docker run` it:

```
$ docker run 218e02166757
Jun 14, 2022 2:19:48 PM java.util.prefs.FileSystemPreferences$1 run
INFO: Created user preferences directory.
|  Welcome to JShell -- Version 11.0.15
|  For an introduction type: /help intro

jshell> 
```

The build works great on my Mac M1 (arm64) machine though. Same as before, 421 MB and outputs: 
```
% docker run 12d42dc234b9
fatal error: could not connect to bitcoind using json-rpc
```

when I run it, which I assume is correct.

I wonder if there's some tweak that can be done to the `eclipse-temurin:11-jdk-jammy` build to make it work properly on x86... does the `jshell` output mean anything to you?

Anyway it seems the alpine build is best for x86, while both `eclipse-temurin:11-jdk-jammy` and `arm64v8/eclipse-temurin:11-jdk-jammy` work just fine for arm64.

I can change the arm64 Dockerfile to use the `eclipse-temurin:11-jdk-jammy`, that way in case you guys discover how to make it work properly with x86 you can merge the two Dockerfiles?

chrisguida:
Ok, updated the PR to use the multi-architecture `eclipse-temurin:11-jdk-jammy` base image instead of the arm64-only `arm64v8/eclipse-temurin:11-jdk-jammy` base image. This is in case you guys can figure out how to get x86 builds working on the same image, then the Dockerfiles can simply be combined into one.

sstone:
I think we can merge as is, I'll add a follow-up PR to update our README.

PR #2307, opened 2022-06-08
Store disabled channel update when offline

t-bast:
When a channel is offline and we want to relay an HTLC through it, we emit a new `channel_update` with the disable bit set (unless it was already disabled).

We previously didn't update our internal state with this new `channel_update`, which created the following issue: if eclair is restarted before the channel comes back online, eclair would only try to rebroadcast the previous `channel_update` which has a lower timestamp than the one disabling the channel. That `channel_update` is then rejected by the network, causing the channel to stay disabled until the next refresh, which only happens after 10 days.

I didn't find an easy way to unit test this, but I manually made an e2e test to reproduce the issue and then check that it was fixed.


pm47 (commented):
I think we should also call `store()` here for the same reason:
https://github.com/ACINQ/eclair/blob/472275573cd4b9576d54a6e06373223877653554/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala#L713

Nit: in your commit message it should be

> We previously didn't persist our internal state with this new `channel_update`

We do _update_ our state in memory, the problem is that we don't _persist_ it.

pm47 (approved):

PR #2308, opened 2022-06-08
Use balance estimates from past payments in path-finding

thomash-acinq:


eclair-core/src/main/resources/reference.conf
```diff
@@ -497,6 +497,8 @@ eclair {
         // probability of success, however is penalizes less the paths with a low probability of success.
         use-log-probability = false
 
+        use-past-relay-data = false
```
t-bast, line 510: Can you add a comment to tell node operators what to expect when this is set to `true`? Also, can you add a section in the release notes?

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
```diff
@@ -235,12 +235,6 @@ case class BalanceEstimate private(low: MilliSatoshi,
     } else {
       0
     }
-
-    if (estimate < 0 || estimate > 1) {
-      log.error("Could not estimate balance: this={}, amount={}, now={}", this, amount, now)
-    }
```
t-bast, line 241: So what was the issue in the end for cases where we ended up logging this?

eclair-core/src/main/resources/reference.conf
thomash-acinq: I've added more details here and in the release notes.

eclair-core/src/main/scala/fr/acinq/eclair/router/BalanceEstimate.scala
thomash-acinq: This never ended up in the logs. I think I've probably fixed the bug in the same commit where I introduced this log: there were multiple calls to `TimestampSecond.now()` which were replaced by a single one.

thomash-acinq: That was https://github.com/ACINQ/eclair/pull/2939

eclair-core/src/main/resources/reference.conf
```diff
@@ -497,10 +504,10 @@ eclair {
         // probability of success, however is penalizes less the paths with a low probability of success.
         use-log-probability = false
 
-        mpp {
-          min-amount-satoshis = 15000 // minimum amount sent via partial HTLCs
-          max-parts = 5 // maximum number of HTLCs sent per payment: increasing this value will impact performance
-        }
+        // When set to false, we assume that the balances of channels are uniformly distributed. That means the
+        // probability that we can an amount X through a channel of capacity C is 1 - X / C.
+        //When set to true, past payments attempts will be used to compute a better estimate.
```
t-bast:
```suggestion
        // probability that we can send an amount X through a channel of capacity C is 1 - X / C.
        // When set to true, past payments attempts will be used to compute a better estimate.
```

thomash-acinq: Done

PR #2313, opened 2022-06-13
Reuse postgres instance in tests

pm47:
Create new database for each unit test instead of spawning a new db instance.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
@@ -104,7 +104,7 @@ object TestDatabases {
 
   case class TestPgDatabases() extends TestDatabases {
 
-    val datasource: DataSource = getNewDatabase
+    val datasource: DataSource = getNewDatabase()
     val hikariConfig = new HikariConfig
     hikariConfig.setDataSource(datasource)
     val lock: PgLock.LeaseLock = PgLock.LeaseLock(UUID.randomUUID(), 10 minutes, 8 minute, LockFailureHandler.logAndThrow, autoReleaseAtShutdown = false)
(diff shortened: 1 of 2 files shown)


pm47:
I'm not too sure about this. The response time for db tests is significantly reduced, but overall build time is unchanged, and it adds complexity.

eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
```diff
@@ -112,13 +115,34 @@ object TestDatabases {
...
+      system.terminate()
+    }
     // @formatter:on
   }
 
+  object TestPgDatabases {
+    /** single instance */
+    private val pg = EmbeddedPostgres.start()
+
+    def getNewDatabase: DataSource = {
```
t-bast, line 133: nit: should look like a function call, not a getter on an internal field:
```suggestion
    def getNewDatabase(): DataSource = {
```

t-bast (commented):
This isn't too much complexity and I believe it makes sense, it's worth having this :+1:

eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
```diff
@@ -95,13 +98,13 @@ object TestDatabases {
...
-    override def close(): Unit = ()
+    override def close(): Unit = connection.close()
     // @formatter:on
   }
 
   case class TestPgDatabases() extends TestDatabases {
-    private val pg = EmbeddedPostgres.start()
-    val datasource: DataSource = pg.getPostgresDatabase
+
+    val datasource: DataSource = getNewDatabase
```
t-bast, line 107: nit:
```suggestion
    val datasource: DataSource = getNewDatabase()
```

If we want to get rid of the warning that this looks like a getter (which is a dumb warning that only looks at the fact that the function name starts with `get`) we can rename that `createNewDatabase`

t-bast (approved):

PR #2316, opened 2022-06-15
Generate local alias before spawning channel

pm47:
As a first step before addressing https://github.com/ACINQ/eclair/pull/2224#discussion_r875068884, we need to make the alias generation asynchronous.

Peer generates the alias along with the shutdown script, and before spawning the channel.

This approach has several drawbacks:
- changes all data classes in the early channel states
- forces us to have a local alias in state `WAIT_FOR_FUNDING_CONFIRMED`, which is persisted and requires a migration. We can't set it to the real scid because we don't have it yet
- not consistent with a future (hypothetical) multi-alias support


eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -420,6 +431,6 @@ private[channel] object ChannelCodecs3 {
     .typecase(0x03, Codecs.DATA_SHUTDOWN_COMPAT_03_Codec)
     .typecase(0x02, Codecs.DATA_NORMAL_COMPAT_02_Codec)
     .typecase(0x01, Codecs.DATA_WAIT_FOR_CHANNEL_READY_COMPAT_01_Codec)
-    .typecase(0x00, Codecs.DATA_WAIT_FOR_FUNDING_CONFIRMED_Codec)
+    .typecase(0x00, Codecs.DATA_WAIT_FOR_FUNDING_CONFIRMED_COMPAT_00_Codec)
```
pm47, line 434: This is really footgunny and I got bitten before: all `Codecs.DATA_xxx_Codec` should be numbered , even the current ones, so the codec names are immutable.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/MigrationMethods.scala
```diff
@@ -0,0 +1,16 @@
...
+import fr.acinq.eclair.wire.protocol.{ChannelReady, FundingCreated, FundingSigned}
+import fr.acinq.eclair.{Alias, BlockHeight}
+import shapeless.{::, HNil}
+
+object MigrationMethods {
+
+  def genLocalAlias: Commitments :: Option[Transaction] :: BlockHeight :: Option[ChannelReady] :: Either[FundingCreated, FundingSigned] :: HNil => DATA_WAIT_FOR_FUNDING_CONFIRMED = {
+    case commitments :: fundingTx :: waitingSince :: deferred :: lastSent :: HNil =>
+      DATA_WAIT_FOR_FUNDING_CONFIRMED(commitments, fundingTx, waitingSince, Alias(commitments.channelId.take(32).toLong()), deferred, lastSent)
+  }
```
pm47, line 14: Tried this as an experiment. It allows us to factor codec migrations across all versions, and is safe since the compiler enforces a strict type check.

It should maybe return a `HList` though, so we can stack migrations.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
pm47: Fixed by bfba9e41193da5fbe76a7685485e131293f32789.

t-bast:
I have an alternative design for alias generation: what about inserting a new temporary channel state to generate it? Whenever we currently call `acceptFundingTx`, we would go to a `WAIT_FOR_LOCAL_ALIAS` state where we store the next state and next state data, generate the local alias (probably by sending a message to the `Router`), and once we receive the alias we go to the next state with the next state data. In that state we store incoming lightning messages and replay them once we have our local alias and transition to the next state. It's very simple for transient states, the only case where it may be a bit painful is the unrequested 0-conf where we are in `WAIT_FOR_FUNDING_CONFIRMED` (which isn't transient), this would need some prototyping...

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/MigrationMethods.scala
pm47: Done in [5f3fa19](https://github.com/ACINQ/eclair/pull/2316/commits/5f3fa197317b862812a34141311f3bdee9e757b5).

pm47:
> I have an alternative design for alias generation: what about inserting a new temporary channel state to generate it?

I thought about that, but I'd don't like adding new states because we then need to handle all the other events (remote errors, funding spent, disconnection, etc.) and it quickly becomes bulky.

> Something else that comes to mind: for public channels that aren't using 0-conf (which is going to be the most common flow for most node operators), we probably shouldn't really generate a localAlias, we can probably use 0?

That's a good point, but I think we don't want to have duplicate aliases, it should be an `Option` in that case. It could complicate things though, let me think about it...

t-bast:
Yes, it's true, after thinking about it more I think we should keep the `localAlias` non-optional, it really makes the code much cleaner. In that case it makes sense to generate it right at the beginning of the channel opening flow. It can be in the `Peer` for now, I think that once #2247 is integrated which refactor the channel `init` state, I may move that in the channel's init state instead, we'll see.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -525,6 +530,8 @@ object Peer {
   case class ConnectionDown(peerConnection: ActorRef) extends RemoteTypes
 
   case class RelayOnionMessage(messageId: ByteVector32, msg: OnionMessage, replyTo_opt: Option[typed.ActorRef[Status]])
+
+  case class ChannelCreationData(localParams: LocalParams)
```
t-bast, line 534: This is currently unused?

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/Mutators.scala
```diff
@@ -0,0 +1,50 @@
...
+ *   .map(mutator2)
+ *   .map(mutator3)
+ *   .decodeOnly
+ *   .as[MyClass]
+ * }}}
+ */
+object Mutators {
+
+  object AddLocalAlias {
+    def genDeterministicAlias(channelId: ByteVector32): Alias = Alias(channelId.take(7).toLong(signed = false))
```
t-bast, line 29: As seen offline, we should take the last bytes to avoid collisions when batching opens in a single transaction.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/Mutators.scala
```diff
@@ -0,0 +1,50 @@
...
+ *   ...
+ *   )
+ *   .map(mutator1)
+ *   .map(mutator2)
+ *   .map(mutator3)
+ *   .decodeOnly
+ *   .as[MyClass]
+ * }}}
+ */
+object Mutators {
```
t-bast, line 26: Should we put that in a versioned namespace (this one would go in the `version3` package since we introduced the need for it in `version3` and won't need it for future versions)?

eclair-core/src/test/scala/fr/acinq/eclair/channel/FuzzySpec.scala
```diff
@@ -77,12 +77,14 @@ class FuzzySpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with Channe
     within(30 seconds) {
       val aliceInit = Init(Alice.channelParams.initFeatures)
       val bobInit = Init(Bob.channelParams.initFeatures)
+      val aliceAlias = ShortChannelId.generateLocalAlias()
+      val bobAlias = ShortChannelId.generateLocalAlias()
```
t-bast, line 81: Would it make sense in these tests (and all the other tests affected by this PR) to use hard-coded local aliases instead? It will make debugging easier.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/Mutators.scala
pm47: Done in [d900265](https://github.com/ACINQ/eclair/pull/2316/commits/d9002656051f59e41d676f9ae0e22077d56d4fb9).

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
pm47: Removed in [4ee1025](https://github.com/ACINQ/eclair/pull/2316/commits/4ee1025748333067f709cbc7c050dc36c5ef169c).

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/Mutators.scala
pm47: Maybe, although we sort of adopted the opposite approach for `ChannelTypes0`.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -383,12 +383,18 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
...
-  def createLocalParams(nodeParams: NodeParams, initFeatures: Features[InitFeature], channelType: SupportedChannelType, isInitiator: Boolean, fundingAmount: Satoshi)(implicit ec: ExecutionContext): Future[LocalParams] = {
-    if (channelType.paysDirectlyToWallet) {
+  def createLocalParams(nodeParams: NodeParams, initFeatures: Features[InitFeature], channelType: SupportedChannelType, isInitiator: Boolean, fundingAmount: Satoshi)(implicit ec: ExecutionContext): Future[(LocalParams, Alias)] = {
+    val localParams_f = if (channelType.paysDirectlyToWallet) {
       wallet.getReceivePubkey().map(walletKey => makeChannelParams(nodeParams, initFeatures, Script.write(Script.pay2wpkh(walletKey)), Some(walletKey), isInitiator, fundingAmount))
     } else {
       wallet.getReceiveAddress().map(address => makeChannelParams(nodeParams, initFeatures, Script.write(addressToPublicKeyScript(address, nodeParams.chainHash)), None, isInitiator, fundingAmount))
     }
+    val localAlias_p = Promise[Alias]()
+    context.system.eventStream.publish(Register.GenerateLocalAlias(localAlias_p))
```
pm47, line 393: This isn't very academic, but:
- using the `eventStream` saves us from having to pass a reference to `register` all the way to the `peer`
- using a `Promise` saves us from creating an ad-hoc actor (`pipeTo` doesn't work with the `eventStream` approach)

eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
```diff
@@ -61,7 +63,11 @@ object ShortChannelId {
 
   def toShortId(blockHeight: Int, txIndex: Int, outputIndex: Int): Long = ((blockHeight & 0xFFFFFFL) << 40) | ((txIndex & 0xFFFFFFL) << 16) | (outputIndex & 0xFFFFL)
 
-  def generateLocalAlias(): Alias = Alias(System.nanoTime()) // TODO: fixme (duplicate, etc.)
+  def generateLocalAlias(): Alias = {
+    // at block height 1000 LN didn't exist, so all real scids less than that will never be used
+    val upperBound = RealShortChannelId.apply(BlockHeight(1000),1,0).toLong
+    Alias(Random.nextLong(upperBound))
```
t-bast, line 69: We should use the `randomLong` from `package.scala` here.

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/Mutators.scala
t-bast: On the contrary, we put `ChannelTypes0` in the `version1` package? But it doesn't matter right now, we can move those mutators to a versioned package when we introduce new ones that are for a different codecs version.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
```diff
@@ -79,6 +82,13 @@ class Register extends Actor with ActorLogging {
         case Some(channel) => channel.tell(msg, compatReplyTo)
         case None => compatReplyTo ! ForwardShortIdFailure(fwd)
       }
+
+    case GenerateLocalAlias(promise) =>
```
t-bast, line 86: Since we re-populate the `shortIds` map asynchronously as channels are restored and send `ShortChannelIdAssigned`, we may receive that message before we've restored all channels and recovered a complete `shortIds` map.

I believe it won't be an issue in practice as long as the range we use for `ShortChannelId.generateLocalAlias()` is big enough to make collisions unlikely.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -60,6 +62,18 @@ class PeerSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with Paralle
     }
   }
 
+  {
+    val generateLocalAliasListener = TestProbe()
+    system.eventStream.subscribe(generateLocalAliasListener.ref, classOf[GenerateLocalAlias])
+    waitEventStreamSynced(system.eventStream)
+    generateLocalAliasListener.setAutoPilot {
+      case (_, gen:GenerateLocalAlias)  =>
```
t-bast, line 70: nit:
```suggestion
      case (_, gen: GenerateLocalAlias) =>
```

t-bast (commented):
Looks good, only a few comments (and a rebase is needed).

eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
```diff
@@ -61,7 +63,11 @@ object ShortChannelId {
 
   def toShortId(blockHeight: Int, txIndex: Int, outputIndex: Int): Long = ((blockHeight & 0xFFFFFFL) << 40) | ((txIndex & 0xFFFFFFL) << 16) | (outputIndex & 0xFFFFL)
 
-  def generateLocalAlias(): Alias = Alias(System.nanoTime()) // TODO: fixme (duplicate, etc.)
+  def generateLocalAlias(): Alias = {
+    // at block height 1000 LN didn't exist, so all real scids less than that will never be used
+    val upperBound = RealShortChannelId.apply(BlockHeight(1000),1,0).toLong
```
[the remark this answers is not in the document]
pm47: The space is already huge : 1_099_511_627_841_536 (*) and we are checking for duplicates. I figure leaving some values available could prove handy later?

(*) So huge in fact that I believe our previous computation was invalid. If we indeed go up to block height 350 000 it gives us a ceiling of 384_829_069_721_665_536 which is about 2^59. The probability of a collision for 250 000 channels would be 0.000011 %, not 0.8 % ?

pm47:
Superseded by #2337.

PR #2320, opened 2022-06-16
Clean up scid parsing from coordinates string

t-bast:
This is a conversion that can fail by throwing an exception, which is dangerous.
It's only used in tests, so we're moving this to a test file to ensure we don't accidentally use it in non-test code.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
@@ -43,6 +43,7 @@ import fr.acinq.eclair.router.{Announcements, Router}
 import fr.acinq.eclair.wire.internal.channel.ChannelCodecsSpec
 import fr.acinq.eclair.wire.protocol.{ChannelUpdate, Color, NodeAnnouncement}
 import org.mockito.scalatest.IdiomaticMockito
+import org.scalatest.TryValues.convertTryToSuccessOrFailure
 import org.scalatest.funsuite.FixtureAnyFunSuiteLike
 import org.scalatest.{Outcome, ParallelTestExecution}
 import scodec.bits._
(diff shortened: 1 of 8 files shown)


pm47:
It's used in the API too I'm afraid

t-bast:
Right, I missed that! Then I'll go back to my first attempt, where I made this explicitly return a `Try` (and renamed it `fromCoordinates`). I'll update the PR.

t-bast:
Updated in https://github.com/ACINQ/eclair/pull/2320/commits/962838582befe16adf74dd6ea07f715e201cdaa5

eclair-core/src/test/scala/fr/acinq/eclair/EclairImplSpec.scala
```diff
@@ -124,11 +124,11 @@ class EclairImplSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with I
 
     // with assisted routes
     val externalId1 = "030bb6a5e0c6b203c7e2180fb78c7ba4bdce46126761d8201b91ddac089cdecc87"
-    val hints = List(List(ExtraHop(Bob.nodeParams.nodeId, ShortChannelId("569178x2331x1"), feeBase = 10 msat, feeProportionalMillionths = 1, cltvExpiryDelta = CltvExpiryDelta(12))))
+    val hints = List(List(ExtraHop(Bob.nodeParams.nodeId, ShortChannelId.fromCoordinates("569178x2331x1").get, feeBase = 10 msat, feeProportionalMillionths = 1, cltvExpiryDelta = CltvExpiryDelta(12))))
```
pm47, line 127: In theory we should use Scalatest's [TryValues](https://www.scalatest.org/scaladoc/3.0.6/org/scalatest/TryValues.html) here.

t-bast: Done in https://github.com/ACINQ/eclair/pull/2320/commits/74e19803a34225e3b36948af2cc8cb09958f5783

pm47 (approved):

PR #2322, opened 2022-06-16
Analyze flaky test

pm47:


eclair-core/src/main/scala/fr/acinq/eclair/IAmReady.scala
```diff
@@ -0,0 +1,10 @@
+package fr.acinq.eclair
+
+/**
+ * Since actors are initialized asynchronously, and the initialization sometimes involves subscribing to an
+ * [[akka.event.EventStream]], we don't know when they are ready to process messages, especially in tests, which
+ * lead to race conditions.
```
t-bast, line 6: nit:
```suggestion
 * leads to race conditions.
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
```diff
@@ -69,15 +70,19 @@ class Register extends Actor with ActorLogging {
       val compatReplyTo = if (replyTo == ActorRef.noSender) sender() else replyTo
       channels.get(channelId) match {
         case Some(channel) => channel.tell(msg, compatReplyTo)
-        case None => compatReplyTo ! ForwardFailure(fwd)
+        case None =>
+          println(ForwardFailure(fwd))
+          compatReplyTo ! ForwardFailure(fwd)
```
t-bast, line 75: nit: revert

eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
```diff
@@ -69,15 +70,19 @@ class Register extends Actor with ActorLogging {
...
 
     case fwd@ForwardShortId(replyTo, shortChannelId, msg) =>
       // for backward compatibility with legacy ask, we use the replyTo as sender
       val compatReplyTo = if (replyTo == ActorRef.noSender) sender() else replyTo
       shortIds.get(shortChannelId).flatMap(channels.get) match {
         case Some(channel) => channel.tell(msg, compatReplyTo)
-        case None => compatReplyTo ! ForwardShortIdFailure(fwd)
+        case None =>
+          println(ForwardShortIdFailure(fwd))
+          compatReplyTo ! ForwardShortIdFailure(fwd)
```
t-bast, line 85: nit: revert

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/ChannelRelay.scala
```diff
@@ -260,6 +260,7 @@ class ChannelRelay private(nodeParams: NodeParams,
     import r._
     outgoingChannel_opt match {
       case None =>
+        println(s"relay failure ${CMD_FAIL_HTLC(add.id, Right(UnknownNextPeer), commit = true)}")
```
t-bast, line 263: nit: remove

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/ThreeNodesIntegrationSpec.scala
```diff
@@ -46,7 +46,7 @@ class ThreeNodesIntegrationSpec extends FixtureSpec with IntegrationPatience {
     // alice now knows about bob-carol
     eventually {
       val routerData = getRouterData(alice)
-      //prettyPrint(routerData, alice, bob, carol)
+      prettyPrint(routerData, alice, bob, carol)
```
t-bast, line 49: nit: revert

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/ZeroConfAliasIntegrationSpec.scala
```diff
@@ -98,44 +98,38 @@ class ZeroConfAliasIntegrationSpec extends FixtureSpec with IntegrationPatience
       }
     }
 
-    eventually {
```
t-bast, line 101: I believe we can now revert the changes to that file, right?

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/MinimalNodeFixture.scala
```diff
@@ -71,6 +72,9 @@ object MinimalNodeFixture extends Assertions {
 
   def apply(nodeParams: NodeParams): MinimalNodeFixture = {
     implicit val system: ActorSystem = ActorSystem(s"system-${nodeParams.alias}")
+    val readyListener = TestProbe("ready-listener")
+    system.eventStream.subscribe(readyListener.ref, classOf[IAmReady])
+    waitEventStreamSynced(system.eventStream)
```
t-bast, line 77: I don't think we need that anymore now, do we? The `readyListener` will be registered before the other actors (since they're created after), so the `readyListener.expectMsgAllOf` below should be enough.

eclair-core/src/main/scala/fr/acinq/eclair/IAmReady.scala
```diff
@@ -0,0 +1,10 @@
...
+ * By making actors publish [[IAmReady]] on the same [[akka.event.EventStream]] they are subscribing to, we guarantee
+ * that if we receive [[IAmReady]] the actor has been initialized and its subscriptions have been taken into account.
+ */
+case class IAmReady(clazz: Class[_])
```
t-bast, line 10: I'm not a big fan of using `Class[_]` here, but I don't see a better identifier...

eclair-core/src/main/scala/fr/acinq/eclair/channel/Register.scala
pm47: I added them back on purpose, because the race is still there.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/ZeroConfAliasIntegrationSpec.scala
pm47: I did, but we can't remove pre-existing `eventually`.

eclair-core/src/main/scala/fr/acinq/eclair/IAmReady.scala
pm47: Yes, the problem is for nested actors like `ChannelRelayer`, for which we don't have the `ActorRef`.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -62,6 +62,7 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
   context.system.eventStream.subscribe(self, classOf[LocalChannelUpdate])
   context.system.eventStream.subscribe(self, classOf[LocalChannelDown])
   context.system.eventStream.subscribe(self, classOf[AvailableBalanceChanged])
+  context.system.eventStream.publish(IAmReady(this.getClass))
```
t-bast, line 65: Shouldn't we emit `IAmReady` just before `startWith`, once we've finished initializing all data?

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/ZeroConfAliasIntegrationSpec.scala
```diff
@@ -99,8 +99,8 @@ class ZeroConfAliasIntegrationSpec extends FixtureSpec with IntegrationPatience
     }
 
     eventually {
-      if (bcPublic && deepConfirm) {
-        // if channel bob-carol is public, we wait for alice to learn about it
+    if (bcPublic && deepConfirm) {
+      // if channel bob-carol is public, we wait for alice to learn about it
```
t-bast, line 103: nit: indenting

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: All messages sent to the router will wait in the actor's mailboc. I chose to keep `IAmReady` close to the `eventStream` subscriptions because this is what actually matters. Maybe I should rename `IAmReady` to `SubscriptionsComplete` so the semantics are more clear?

t-bast: > Maybe I should rename IAmReady to SubscriptionsComplete so the semantics are more clear?

Yes, I think it would make more sense.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/ZeroConfAliasIntegrationSpec.scala
pm47: Done in [9ca8551](https://github.com/ACINQ/eclair/pull/2322/commits/9ca8551a1d16a2cec2acb1ea46373a3d9266946e).

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
pm47: Done in [9ca8551](https://github.com/ACINQ/eclair/pull/2322/commits/9ca8551a1d16a2cec2acb1ea46373a3d9266946e).

pm47:
Addressed all comments. Unfortunately the race condition in `a->b->c (b-c private)` is still there 😠

t-bast (approved):
> Unfortunately the race condition in a->b->c (b-c private) is still there :angry:

Damn, then there's something else here...but this PR is desirable anyway so let's merge it!

Issue #2295, opened 2022-05-30
Possible bug in RoutingHeuristics.normalize


barrybecker4:
While I was adding some tests for RoutingHeuristics.normalize I noticed something that doesn't seem right. 

If you normalize a value that == max, then you get `1.0` as the result.
If you normalize a value that is > max. then you get `0.99999` as the result.
If you normalize a value that is epsilon less than max, then you get a value between 0.99999 and 1.0 as the result.

I don't think that this is expected. According to the comment, the result should be in (0, 1) - meaning that 0 and 1 should be excluded. Yet, when value is max, you can get 1.

I think that maybe the intended implementation should be 

```
def normalize(value: Double, min: Double, max: Double): Double = {
  val extent = max - min
  val clampedValue = Math.min(Math.max(min, value), max)
  val normalized = (clampedValue - min) / extent
  Math.min(Math.max(0.00001D, normalized), 0.99999D)
}
```
or perhaps
```
def normalize(value: Double, min: Double, max: Double): Double = {
  val extent = max - min
  val clampedValue = Math.min(Math.max(min, value), max)
  0.00001D + 0.99998D * (clampedValue - min) / extent
}
```
The first way will give the same values as before in most cases, but clamp the result to [0.00001, 0.99999]
The second way scales the normalized value into [0.00001, 0.99999]. An advantage of this second approach is that values very close to the min and max will always be different.

Which is better?
Should I make a PR for this change?

thomash-acinq:
You're right, the current implementation does not do what it claims to do. That's not a problem in practice because we don't rely on the fact that 0 and 1 are not possible return values. It would still be nice to fix it thought.

barrybecker4:
I think the second approach is more correct because it avoids 0 and 1 while still returning different values near the extremes.

Issue #2294, opened 2022-05-30
Refactor Graph


barrybecker4:
This is a proposal to do a refactoring of the Graph class and not add any features.

My thought is that it can be moved to a `fr.acinq.eclair.router.graph` package and split into separate, more easily testable and comprehensible, class files. Furthermore, I would like to extract the yenKshortestPaths and dijkstraShortestPath to a separate ShortestPathFinder class. I think that there may be a bug or two in the current implementation (see [this PR](https://github.com/ACINQ/eclair/pull/2293)).

PR #2325, opened 2022-06-20
Graph refactor (#2294)

barrybecker4:
@thomash-acinq, this PR got larger than I planned, so I understand if you don't want to merge it as is. Other than the small fix for #2295, there should be no functional changes.

The goal was just to split up some of the graph related files so that they could be understood (at least by me) and tested more easily. I used packages instead of objects to group related classes. That way each file could stand on its own and be tested independently.

If there is general agreement that these sorts of changes would be good, then I could start a new branch and recreate the changes in a series of smaller PRs if you prefer.

Whether or not this PR is accepted, it was a learning experience for me, so I don't consider it a wasted effort.

Question: Should it be possible to create a BockHeight with a negative height? If not, we might want an assertion in the constructor.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -224,6 +224,8 @@ object RouteCalculation {
     }
   }
 
+  val shortestPathFinder = new ShortestPathFinder()
+
   @tailrec
   private def findRouteInternal(g: DirectedGraph,
                                 localNodeId: PublicKey,
@@ -249,7 +251,7 @@ object RouteCalculation {
     val boundaries: RichWeight => Boolean = { weight => feeOk(weight.amount - amount) && lengthOk(weight.length) && cltvOk(weight.cltv) }
 
     val foundRoutes: Seq[Path.WeightedPath] =
-      new ShortestPathFinder().yenKshortestPaths(g, localNodeId, targetNodeId, amount, ignoredEdges, ignoredVertices, extraEdges,
+      shortestPathFinder.yenKshortestPaths(g, localNodeId, targetNodeId, amount, ignoredEdges, ignoredVertices, extraEdges,
                                                  numRoutes, routeParams.heuristics, currentBlockHeight, boundaries, routeParams.includeLocalChannelCost)
     if (foundRoutes.nonEmpty) {
       val (directRoutes, indirectRoutes) = foundRoutes.partition(_.path.length == 1)


eclair-core/src/main/scala/fr/acinq/eclair/router/graph/path/WeightRatios.scala
```diff
@@ -0,0 +1,38 @@
...
+ * The sum of the four weights must be one.
+ */
+case class WeightRatios(baseWeight: Double, cltvDeltaWeight: Double, ageWeight: Double, capacityWeight: Double, hopCost: RelayFees) {
+  require(baseWeight + cltvDeltaWeight + ageWeight + capacityWeight == 1, "The sum of heuristics ratios must be 1")
+  require(baseWeight >= 0.0, "ratio-base must be nonnegative")
+  require(cltvDeltaWeight >= 0.0, "ratio-cltv must be nonnegative")
+  require(ageWeight >= 0.0, "ratio-channel-age must be nonnegative")
+  require(capacityWeight >= 0.0, "ratio-channel-capacity must be nonnegative")
+
+  def calculateWeightedFactor(cltvFactor: Double, ageFactor: Double, capFactor: Double): Double = {
```
thomash-acinq, line 34: It's an interesting idea to compute the weight here, but if you're going in that direction you may as well do everything: take a `RichWeight` and a `GraphEdge` and return the new `RichWeight`.

thomash-acinq (commented):
Thanks @barrybecker4 for all of the work, however we are not interested in such refactoring. We try to only do refactorings when they are needed because of some functional change we want to make. Also I think that `Graph.scala` does not need to be split into so many files.
I may use some of your ideas next time I make some big changes to the path-finding logic but for now we'll keep it as it is.

barrybecker4:
> Thanks @barrybecker4 for all of the work, however we are not interested in such refactoring. We try to only do refactorings when they are needed because of some functional change we want to make. Also I think that `Graph.scala` does not need to be split into so many files. I may use some of your ideas next time I make some big changes to the path-finding logic but for now we'll keep it as it is.

OK. I won't put any more effort into this PR at this time, but here are some reasons why I think that splitting Graph into smaller classes organized my package would be a good idea.

- Smaller more single purposed classes and objects would be easier to understand and improve.
- Test coverage could be increased
- Tests would be organized into separate files instead of one giant one - making it easier to run small sets of tests.

t-bast:
But there are also caveats to what you suggest: it exposes internal implementation details, which ties tests to low-level implementation details that we may want to change in the future, thus making it more expensive to update the implementation because we'd need to update all the tests as well (and most of the tests wouldn't make sense after an implementation change).

The way I see it "route calculation" is the abstraction we want to have and test for, this is the logical component we'll always be interested in regardless of how it's implemented. This gives us freedom to change parts of the path-finding algorithm without invalidating the unit tests, which is extremely valuable. I agree that _some_ of the low-level details are sometimes worth unit testing (mostly when we find a bug in that code), but an explosion of smaller components and unit tests for trivial code is usually not really helping readability and is a bigger maintenance burden.

Also, as a new contributor to a big open-source codebase that handles people's money, I don't think it's a good idea to embark on big refactorings without a specific improvement in mind. If there isn't a strong motivation for it, reviewing many changes that could negatively impact a critical component is something maintainers will be wary of. It's easier to contribute small, focused changes to specific issues that you find (like you did in #2331, thanks for that!), work on a feature that people need (some of them can be found by browsing our issues) or help review existing PRs. Does that make sense?

barrybecker4:
Thanks for giving your reasoning. This was still a good exercise for me to help my understanding, but I understand why you don't want to merge it. I will look at other things.

PR #2328, opened 2022-06-23
Additive per-node feature override

t-bast:
The `override-init-features` field in `eclair.conf` was not previously applied on top of the `features` field, so node operators were usually copy-pasting their `features` overrides in every `override-init-features`.

The overrides are now applied on top of the base `features` configuration, which makes it easier and more intuitive to configure per-node features.

### DIFF
--- a/docs/Configure.md
+++ b/docs/Configure.md
@@ -102,22 +102,50 @@ When experimenting with non-default features, we recommend using this to scope t
 This is done with the `override-init-features` configuration parameter in your `eclair.conf`:
 
 ```conf
+eclair.features {
+  var_onion_optin = mandatory
+  payment_secret = mandatory
+  option_support_large_channel = optional
+  option_static_remotekey = optional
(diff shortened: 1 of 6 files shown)


t-bast:
I don't think this is a good idea, we want our configuration to be as simple as possible. I really don't expect node operators to need two different ways of configuring feature overrides, this is very footgunny, don't you think?

I believe it makes more sense to always start from the normal features, and on a per-peer basis choose what features you want to change (enable, disable, switch from optional to mandatory or the opposite).

pm47 (approved):
I believe we should probably mention the new behavior in the release notes.

t-bast:
> I believe we should probably mention the new behavior in the release notes.

Agreed, done in https://github.com/ACINQ/eclair/pull/2328/commits/ab8062fee4ba7ea795d72a0f800e12723d38682a

pm47 (approved):

PR #2331, opened 2022-06-25
Fix RoutingHeuristics.normalize (2295)

barrybecker4:
Fixed https://github.com/ACINQ/eclair/issues/2295 using the second proposed approach so that values near the extremes will still be distinct.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -415,9 +415,10 @@ object Graph {
      * extremes but always bigger than zero so it's guaranteed to never return zero
      */
     def normalize(value: Double, min: Double, max: Double): Double = {
-      if (value <= min) 0.00001D
-      else if (value > max) 0.99999D
-      else (value - min) / (max - min)
+      assert(max > min)
+      val clampedValue = Math.min(Math.max(min, value), max)
+      val extent = max - min
+      0.00001D + 0.99998D * (clampedValue - min) / extent
     }
 
   }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -415,7 +415,7 @@ object Graph {
      * extremes but always bigger than zero so it's guaranteed to never return zero
      */
     def normalize(value: Double, min: Double, max: Double): Double = {
-      assert(max > min)
+      require(max > min, "The max must be larger than the min")
       val clampedValue = Math.min(Math.max(min, value), max)
       val extent = max - min
       0.00001D + 0.99998D * (clampedValue - min) / extent
--- a/eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
@@ -348,4 +348,52 @@ class GraphSpec extends AnyFunSuite {
     assert(paths(1).path == Seq(edgeCE, edgeEG, edgeGH))
     assert(paths(2).path == Seq(edgeCD, edgeDF, edgeFH))
   }
+
+  test("RoutingHeuristics.normalize when value in range") {
+    assert(Graph.RoutingHeuristics.normalize(value = 10, min = 0, max = 100) == 0.100008)
+    assert(Graph.RoutingHeuristics.normalize(value = 20, min = 10, max = 200) == 0.05264052631578948)
+    assert(Graph.RoutingHeuristics.normalize(value = -11, min = -100, max = -10) == 0.9888791111111112)
+  }
+
+  test("Graph.RoutingHeuristics.normalize when value is == min") {
+    assert(Graph.RoutingHeuristics.normalize(value = 0, min = 0, max = 100) == 1.0E-5)
+    assert(Graph.RoutingHeuristics.normalize(value = 10, min = 10, max = 200) == 1.0E-5)
+    assert(Graph.RoutingHeuristics.normalize(value = -100, min = -100, max = -10) == 1.0E-5)
+  }
+
+  test("Graph.RoutingHeuristics.normalize when value is < min") {
+    assert(Graph.RoutingHeuristics.normalize(value = -1, min = 0, max = 100) == 1.0E-5)
+    assert(Graph.RoutingHeuristics.normalize(value = 9.1, min = 10, max = 200) == 1.0E-5)
+    assert(Graph.RoutingHeuristics.normalize(value = -101.1, min = -100, max = -10) == 1.0E-5)
+  }
+
+  test("Graph.RoutingHeuristics.normalize when value is == max") {
+    assert(Graph.RoutingHeuristics.normalize(value = 100, min = 0, max = 100) == 0.99999)
+    assert(Graph.RoutingHeuristics.normalize(value = 200, min = 10, max = 200) == 0.9999899999999999)
+    assert(Graph.RoutingHeuristics.normalize(value = -10, min = -100, max = -10) == 0.9999899999999999)
+  }
+
+  test("Graph.RoutingHeuristics.normalize when value is > max") {
+    assert(Graph.RoutingHeuristics.normalize(value = 105.2, min = 0, max = 100) == 0.99999)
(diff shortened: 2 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -348,4 +348,52 @@ class GraphSpec extends AnyFunSuite {
     assert(paths(1).path == Seq(edgeCE, edgeEG, edgeGH))
     assert(paths(2).path == Seq(edgeCD, edgeDF, edgeFH))
   }
+
```
thomash-acinq, line 351: That's a lot of tests for a single short function. You should at least group them together in a single test.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -348,4 +348,52 @@ class GraphSpec extends AnyFunSuite {
...
+  test("RoutingHeuristics.normalize when value in range") {
+    assert(Graph.RoutingHeuristics.normalize(value = 10, min = 0, max = 100) == 0.100008)
+    assert(Graph.RoutingHeuristics.normalize(value = 20, min = 10, max = 200) == 0.05264052631578948)
+    assert(Graph.RoutingHeuristics.normalize(value = -11, min = -100, max = -10) == 0.9888791111111112)
+  }
```
thomash-acinq, line 356: This test only tells us that the code hasn't changed but it doesn't test any property we care about. We don't want `normalize(10, 0, 100) == 0.100008`, we want it to be `0.1` and we tolerate some small deviation.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -348,4 +348,52 @@ class GraphSpec extends AnyFunSuite {
...
+
+  test("Graph.RoutingHeuristics.normalize when value is < min") {
+    assert(Graph.RoutingHeuristics.normalize(value = -1, min = 0, max = 100) == 1.0E-5)
+    assert(Graph.RoutingHeuristics.normalize(value = 9.1, min = 10, max = 200) == 1.0E-5)
+    assert(Graph.RoutingHeuristics.normalize(value = -101.1, min = -100, max = -10) == 1.0E-5)
+  }
+
+  test("Graph.RoutingHeuristics.normalize when value is == max") {
+    assert(Graph.RoutingHeuristics.normalize(value = 100, min = 0, max = 100) == 0.99999)
+    assert(Graph.RoutingHeuristics.normalize(value = 200, min = 10, max = 200) == 0.9999899999999999)
```
thomash-acinq, line 372: How come you don't expect the same result in both of these cases? It looks like a bug, but it's actually not because we don't care about the exact value.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -414,9 +414,10 @@ object Graph {
      * extremes but always bigger than zero so it's guaranteed to never return zero
      */
     def normalize(value: Double, min: Double, max: Double): Double = {
-      if (value <= min) 0.00001D
-      else if (value > max) 0.99999D
-      else (value - min) / (max - min)
+      assert(max > min)
```
thomash-acinq, line 417: If you put something here it should be a `require`.

thomash-acinq (commented):
Thanks.
I think you got a bit carried away with the tests, it's important to test but this amount seems excessive.

eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
barrybecker4: Done.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
barrybecker4: I changed to the other proposed implementation so that now we get 0.1 for this case.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
barrybecker4: I removed some and grouped them together in one test as you suggested. 
These tests are lightweight and should not take any significant time to execute. Having them in separate groups with descriptions made them easier to update if there were any failures or if the implementation changes.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
barrybecker4: OK - If you prefer the original implementation, I can switch back to it and add a tolerance to the test.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
thomash-acinq: The core issue here is that `normalize` is just a simple helper function that's only called in one other function and whose result doesn't need to be very precise (the worst thing that could happen is that we would select a path that's slightly worse than the optimal one). So while it's nice to fix it so that it's behavior matches the comment (which by the way could also be achieved by rewriting the comment), we don't want to spend too much energy on it and we don't want to add 48 lines of tests for it.

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
```diff
@@ -348,4 +348,28 @@ class GraphSpec extends AnyFunSuite {
...
+    assert(Graph.RoutingHeuristics.normalize(value = 99.999999, min = 0, max = 100) == 0.99999)
+    assert(Graph.RoutingHeuristics.normalize(value = 199.999999934, min = 10, max = 200) == 0.99999)
+  }
+
+
+  test("Graph.RoutingHeuristics.normalize should throw exception if min > max") {
+    assertThrows[IllegalArgumentException](
+      Graph.RoutingHeuristics.normalize(value = 9, min = 10, max = 1)
+    )
+  }
```
thomash-acinq, line 374:
```suggestion
  test("RoutingHeuristics.normalize") {
    // value inside the range
    assert(Graph.RoutingHeuristics.normalize(value = 10, min = 0, max = 100) === (10.0 / 100.0) +- 0.001)
    assert(Graph.RoutingHeuristics.normalize(value = 20, min = 10, max = 200) === (10.0 / 190.0) +- 0.001)
    assert(Graph.RoutingHeuristics.normalize(value = -11, min = -100, max = -10) == (89.0 / 90.0) +- 0.001)
    
    // value on the bounds
    assert(Graph.RoutingHeuristics.normalize(value = 0, min = 0, max = 100) > 0)
    assert(Graph.RoutingHeuristics.normalize(value = 10, min = 10, max = 200) > 0)
    assert(Graph.RoutingHeuristics.normalize(value = -100, min = -100, max = -10) > 0)
    assert(Graph.RoutingHeuristics.normalize(value = 9.1, min = 10, max = 200) > 0)
    assert(Graph.RoutingHeuristics.normalize(value = 100, min = 0, max = 100) < 1)
    assert(Graph.RoutingHeuristics.normalize(value = 200, min = 10, max = 200) < 1)
    
    // value outside the range
    assert(Graph.RoutingHeuristics.normalize(value = 105.2, min = 0, max = 100) < 1)
    
    // Should throw exception if min > max
    assertThrows[IllegalArgumentException](
      Graph.RoutingHeuristics.normalize(value = 9, min = 10, max = 1)
    )
  }
```

Tests should not contain unexplainable constants, to get these constants you just ran the function so it just tests that the function returns what the function returned.
The fact that we clamp at 0.00001 is an implementation detail, not something we want to test. We only want to be in (0, 1).

eclair-core/src/test/scala/fr/acinq/eclair/router/GraphSpec.scala
barrybecker4: OK

barrybecker4:
There seems to be an unrelated test failure - even after rebasing on the latest
```
The code passed to eventually never returned normally. 
Attempted 61 times over 15.116898905 seconds. 
Last failure message: assertion failed: expected class fr.acinq.eclair.payment.PaymentSent, found class fr.acinq.eclair.payment.PaymentFailed 
```

thomash-acinq (approved):

thomash-acinq:
Thanks, merged.

PR #2332, opened 2022-06-26
Resume reading after processing unknown messages

pm47:
If the current chunk of data read from the TCP connection only contains unknown messages (in particular, could be only one isolated unknown message on an otherwise idle connection), we never resumed reading on the connection.

This means all subsequent messages, including pings/pongs, won't be read, which is why the most visible effect is disconnecting due to no response to ping.

Related to https://github.com/ElementsProject/lightning/pull/5347.

Reported by @wtogami.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
@@ -94,7 +94,8 @@ class TransportHandler[T: ClassTag](keyPair: KeyPair, rs: Option[ByteVector], co
     makeReader(keyPair)
   }
 
-  def sendToListener(listener: ActorRef, plaintextMessages: Seq[ByteVector]): Map[T, Int] = {
+  def decodeAndSendToListener(listener: ActorRef, plaintextMessages: Seq[ByteVector]): Map[T, Int] = {
+    log.debug("decoding {} plaintext messages", plaintextMessages.size)
     var m: Map[T, Int] = Map()
     plaintextMessages.foreach(plaintext => Try(codec.decode(plaintext.toBitVector)) match {
       case Success(Attempt.Successful(DecodeResult(message, _))) =>
@@ -106,6 +107,7 @@ class TransportHandler[T: ClassTag](keyPair: KeyPair, rs: Option[ByteVector], co
       case Failure(t) =>
         log.error(s"cannot deserialize $plaintext: ${t.getMessage}")
     })
+    log.debug("decoded {} messages", m.values.sum)
     m
   }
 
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
```diff
@@ -178,13 +178,20 @@ class TransportHandler[T: ClassTag](keyPair: KeyPair, rs: Option[ByteVector], co
...
+        log.debug("received chunk of size={}", data.size)
         val (dec1, plaintextMessages) = d.decryptor.copy(buffer = d.decryptor.buffer ++ data).decrypt()
         if (plaintextMessages.isEmpty) {
           connection ! Tcp.ResumeReading
           stay() using d.copy(decryptor = dec1)
         } else {
-          log.debug("read {} messages, waiting for readacks", plaintextMessages.size)
+          log.debug("decoding {} raw messages", plaintextMessages.size)
           val unackedReceived = sendToListener(d.listener, plaintextMessages)
+          if (unackedReceived.isEmpty) {
```
t-bast, line 189: Why don't you need the same trick in `WaitingForListener` when receiving `Event(Listener(listener), ...)`?

eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
```diff
@@ -178,13 +178,20 @@ class TransportHandler[T: ClassTag](keyPair: KeyPair, rs: Option[ByteVector], co
...
+            log.debug("no decoded messages in this chunk, resuming reading")
+            connection ! Tcp.ResumeReading
```
t-bast, line 191: It's probably worth a comment explaining that when there are known messages in this chunk, we let the `PeerConnection` decide when to do a `ReadAck` which will then resume reading, that's why in the case where there's nothing forwarded to the `PeerConnection`, we need to do it ourselves here.

t-bast (commented):
Good catch! What's surprising is that we use a `lightningMessageCodecWithFallback` so the codec should always succeed even for messages we don't understand, unless they're not using the lightning message format at all - which should be a bug on the sender side as well, shouldn't it?

pm47:
> What's surprising is that we use a lightningMessageCodecWithFallback so the codec should always succeed even for messages we don't understand, unless they're not using the lightning message format at all

Thanks, I get now why the error I got when reproducing locally (`unknown discriminator`) was different from the one on the node (`cannot acquire 252176 bits from a vector that contains 440 bits`): I was using `lightningMessageCodec`, not `lightningMessageCodecWithFallback`.

But there is no "standard lightning message format" though, the `UnknownMessage` has been introduced by Anton for his plugins. We are actually defining our own standard there, and will never get cleaned errors for all other unknown messages...

eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
pm47: Done with e998fe58a. Much cleaner, but more changes.

eclair-core/src/main/scala/fr/acinq/eclair/crypto/TransportHandler.scala
pm47: Done with ff30986ce, but I think e998fe5 makes the logic obvious: whenever `unackedReceived` is empty, we call `Tcp.ResumeReading`.

t-bast (approved):
It does make more sense like this, LGTM

PR #2334, opened 2022-06-27
Add host metrics grafana dashboard

GoutamVerma:
In this PR I have added following:
- host-metircs.json file to monitor eclair host metrics on Grafana.
- Update docs/Monitoring.md file.   

The JSON file contain fours rows with various panels. That helps users to monitor their `eclair host metrics`. 
rows

- Load Average (1 panel).
loadaverage

- Memory and File System Usage (3 panels).
memory

- Network Usage (2 panels).
network usage

- CPU Usage (1 panels).
cpu usage


docs/Monitoring.md
```diff
@@ -86,6 +86,8 @@ scrape_configs:
       - targets: ['<url of the eclair http embedded server>']
  ```
 
+After setting up with the prometheus, download [grafana](https://grafana.com/grafana/download) and visit grafana-web via `http://localhost:3000`. Finally, you can import monitoring dashboards available at monitoring directory(see the [grafana documentation](https://grafana.com/docs/grafana/latest/dashboards/export-import/#import-dashboard) for more details).
```
t-bast, line 89:
```suggestion
Eclair provides many [Grafana](https://grafana.com/) dashboards to help you monitor your lightning node, which you can find in the `monitoring` folder of this repository. Follow the [Grafana documentation](https://grafana.com/docs/grafana/latest/dashboards/export-import/#import-dashboard) to import these dashboards and create new ones if necessary.
```

monitoring/dashboard.json
```diff
@@ -0,0 +1,955 @@
...
+  "templating": {
+    "list": []
+  },
+  "time": {
+    "from": "2022-06-24T04:54:44.507Z",
+    "to": "2022-06-24T04:58:02.485Z"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "Ecalir Node Monitor",
```
t-bast, line 951: Seriously, a typo in "eclair"? Also, the name of this dashboard is too broad, it just contains system metrics so it should be named "Eclair Host Metrics" or something like that.

monitoring/dashboard.json
```diff
@@ -0,0 +1,955 @@
...
+  "schemaVersion": 36,
+  "style": "dark",
+  "tags": [],
+  "templating": {
+    "list": []
+  },
+  "time": {
+    "from": "2022-06-24T04:54:44.507Z",
+    "to": "2022-06-24T04:58:02.485Z"
+  },
```
t-bast, line 948: Does this have to be set to a specific time-range? If doesn't make sense to have it focused by default on June 24th, this should be undefined by default.

monitoring/dashboard.json
```diff
@@ -0,0 +1,955 @@
...
+            "overrides": [
+              {
+                "__systemRef": "hideSeriesFrom",
+                "matcher": {
+                  "id": "byNames",
+                  "options": {
+                    "mode": "exclude",
+                    "names": [
+                      "{__name__=\"host_storage_mount_space_total_bytes\", component=\"host\", instance=\"localhost:9095\", job=\"eclair\", mount=\"/\"}",
+                      "{__name__=\"host_storage_mount_space_free_bytes\", component=\"host\", instance=\"localhost:9095\", job=\"eclair\", mount=\"/\"}"
```
t-bast, line 491: What do these refer to? It's probably not a good idea to have a hard-coded url like `localhost:9095`, this looks specific to your own setup.

t-bast (commented):
Please be more thorough when you open a PR, it really looks like you haven't reviewed it yourself before submitting it.

Can you:

- fix the following comments
- include a screenshot of that dashboard in the PR description
- rename the `dashboard.json` file: we will obviously have other dashboards, so it should have a more specific name that describes what this specific dashboard contains, like `host-metrics.json`

monitoring/dashboard.json
GoutamVerma: I agreed with you, It doesn't make any sense and should be undefined by default!

monitoring/dashboard.json
GoutamVerma: Ops My mistake, This file was generated through the Grafana web, and I added an instance at the time of dashboard creation but removed this in the latest commits. The presence of an instance doesn't make any sense.

t-bast (commented):
This is missing steps, if someone tries this it won't work because they need to first add a data source in Grafana to connect it to Prometheus. You should explain how to do this in `Monitoring.md`.

I think we should put the `json` files in `monitoring/grafana` instead of just `monitoring`, since these are really grafana-specific files.

Also, how did you choose the host metrics to display in that graph? Did you spend time thinking about whether it makes sense? For example, the network data read/write is a rate over 1h: this really doesn't make any sense, we want a rate over a much small duration to be useful, for example 1 minute. Please spend time thinking about the metrics you're displaying and how they're best useful for node operators.

GoutamVerma:
> This is missing steps, if someone tries this it won't work because they need to first add a data source in Grafana to connect it to Prometheus. You should explain how to do this in `Monitoring.md`.

Yes, the steps for adding data source are missing in current doc, It is good to add some information regards of this along with official [doc](https://prometheus.io/docs/visualization/grafana/#creating-a-prometheus-data-source)?

> Also, how did you choose the host metrics to display in that graph?

I have taken a reference from current `Kamon dashboard` and I think that this are major host metrics generated through `eclair-node`. But I will think again more about this in deep.

> the network data read/write is a rate over 1h: this really doesn't make any sense, we want a rate over a much small duration to be useful, for example 1 minute. 

I didn't define a time interval for monitoring and by default it is 6 hours for all panels, But I think it is good to monitor last 5 minute because under that time interval user will be able to understand complete scenario. 

Can we define some default time intervals for monitoring? But on the other hand, Grafana allows users to change time intervals while monitoring. In below snapshot we can see that Grafana provide option to select time interval. 

network


> I'm also surprised by the very high memory usage: are you sure you are plotting the correct thing? 

I have checked my OS task manager, and the metrics are correct, although it shows a high rate because of low memory resources in virtual machine.

> As for CPU, why did you choose this type of heatmap display? It's not very readable, is that a best practice?

At initial point, I was taken a reference from our working `Kamon dashboard` and this idea is really comes from this dashboard. But I am find better metrics or display for monitoring. 

kamon

t-bast:
> Can we define some default time intervals for monitoring? But on the other hand, Grafana allows users to change time intervals while monitoring. In below snapshot we can see that Grafana provide option to select time interval.

As with all software, we want to provide good default values.
That doesn't prevent users from then changing it in the UI.
A good default value for most rates is between 1 minute and 5 minutes.
Note that you can directly modify the `json` file to get the results you want.
For example, you currently have:

```
rate(host_network_data_read_bytes_total[1h])
```

You simply need to change it to:

```
rate(host_network_data_read_bytes_total[5m])
```

Let me know when you've fixed all the comments and the PR is ready for review, we really need to integrate this before the end of the week.

GoutamVerma:
> Let me know when you've fixed all the comments and the PR is ready for review, we really need to integrate this before the end of the week.

I am confused, What to use for `host_cpu_usage` heatmap or time series display?
The output of heatmap looks like:
system heat

Output of timeseries looks like:
system time series

I done with the almost all stuffs expect `host_cpu_usage` and also going to finish this work by tomorrow. It will be great if get some suggestion over heatmap or time series(in context of `host_cpu_usage`).

t-bast:
What do you think makes more sense? You should think about what a node operator will find easier to look at. In my opinion, a heatmap doesn't make any sense at all here, every task manager that I know of displays a timeseries...

GoutamVerma:
Mentor, I have tried the possible ways with timeseries display and it really looks more clear then heatmap. It will also contains max, min, first and last received values in %.

Here are outcomes:
Overall
overall

CPU System Usage
system

I think we should move ahead with this panels!

t-bast:
> I think we should move ahead with this panels!

Then please update the PR accordingly so we can make progress.

monitoring/grafana-dashboard/host-metrics.json
```diff
@@ -942,6 +942,10 @@
   "templating": {
     "list": []
   },
+  "time": {
+    "from": "now-5m",
```
t-bast, line 946: Defaulting to watching only the last 5 minutes doesn't make any sense? A node operator is interested in at least the last few hours, this should be something like `now-12h`.

docs/Monitoring.md
```diff
@@ -86,6 +86,11 @@ scrape_configs:
       - targets: ['<url of the eclair http embedded server>']
  ```
 
+Eclair provides many [Grafana](https://grafana.com/) dashboards to help you monitor your lightning node, which you can find in the `monitoring` folder of this repository. Follow the [Grafana documentation](https://grafana.com/docs/grafana/latest/dashboards/export-import/#import-dashboard) to import these dashboards and create new ones if necessary.
+
+Note: Do not forget to add `Prometheus` as data source in grafana(see [grafana documentation](https://prometheus.io/docs/visualization/grafana/#creating-a-prometheus-data-source) for more details)
```
t-bast, line 91:
```suggestion
Note: do not forget to add `Prometheus` as a data source in grafana (see [grafana documentation](https://prometheus.io/docs/visualization/grafana/#creating-a-prometheus-data-source) for more details)
```

monitoring/grafana-dashboard/host-metrics.json
```diff
@@ -0,0 +1,1045 @@
...
+          "editorMode": "builder",
+          "expr": "rate(host_cpu_usage_sum{mode=\"system\"}[1m])",
+          "legendFormat": "__auto",
+          "range": true,
+          "refId": "A"
+        }
+      ],
+      "title": "System",
+      "type": "timeseries"
+    }
```
t-bast, line 1026: This graph is redundant with the previous one, it shows exactly the same data, why did you include it? We only need one CPU graph, not two containing the same thing?

monitoring/grafana-dashboard/host-metrics.json
```diff
@@ -0,0 +1,1045 @@
...
+                "uid": "${DS_PROMETHEUS}"
+              },
+              "editorMode": "builder",
+              "expr": "rate(host_network_data_write_bytes_total[1m])",
+              "hide": false,
+              "legendFormat": "__auto",
+              "range": true,
+              "refId": "B"
+            }
+          ],
```
t-bast, line 689: When we look at the graph, we don't know which one is `read` and which one is `write`, the legend only says `{component="host", instance="localhost:9095", interface="wlp2s0", job="eclair"}` for both curves. Can you make sure the legend mentions the name of the metric?

You have the same issue for every of the graphs, the legend doesn't tell what is displayed, please fix it everywhere.

monitoring/grafana-dashboard/host-metrics.json
```diff
@@ -0,0 +1,1045 @@
...
+          },
+          "pluginVersion": "9.0.1",
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "${DS_PROMETHEUS}"
+              },
+              "editorMode": "builder",
+              "expr": "host_memory_usage",
```
t-bast, line 453: I don't understand the usefulness of displaying that card, since you're not using any expression it's just stuck forever at the first value it reads...you already have a graph that displays this metric with a rate above (`rate(host_memory_usage[1m])` so this one is redundant and should be removed?

monitoring/grafana-dashboard/host-metrics.json
```diff
@@ -0,0 +1,1045 @@
...
+  "templating": {
+    "list": []
+  },
+  "time": {
+    "from": "now-12h",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "eclair Host Metrics",
```
t-bast, line 1041:
```suggestion
  "title": "Eclair Host Metrics",
```

monitoring/grafana-dashboard/host-metrics.json
GoutamVerma: But if we remove `rate(host_memory_usage[1m])`, rate function will not work definitely and I also tried to combine name of metrics in `expr` but it doesn't work.

I think we should remove rate function and provide simple naming convention with `expr("host_swap_usage")`

t-bast: > But if we remove rate(host_memory_usage[1m]), 

That's not what I'm saying! I'm just saying that this graph is buggy but it's redundant anyway so it should just be removed.

t-bast: > But if we remove rate(host_memory_usage[1m]), 

That's not what I'm saying! I'm just saying that this graph is buggy but it's redundant anyway so it should just be removed.

monitoring/grafana-dashboard/host-metrics.json
GoutamVerma: I got an answer of this problem and replaced the metrics name with readable name(updated snapshot in description of PR).

t-bast (approved):
The mounted data metrics are still not correctly named but let's merge this and fix it later to make progress.

PR #2337, opened 2022-06-28
Random values for `generateLocalAlias()`

pm47:
This is an alternative to #2316.

Turns out the value space is large enough to not have to worry about duplicates.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
@@ -58,7 +58,24 @@ object ShortChannelId {
 
   def toShortId(blockHeight: Int, txIndex: Int, outputIndex: Int): Long = ((blockHeight & 0xFFFFFFL) << 40) | ((txIndex & 0xFFFFFFL) << 16) | (outputIndex & 0xFFFFL)
 
-  def generateLocalAlias(): Alias = Alias(System.nanoTime()) // TODO: fixme (duplicate, etc.)
+  /**
+   * At block height 350 000 LN didn't exist, so all real scids less than that will never be used. This results in
+   * `384_829_069_721_665_536` possible values, greater than `2^58`.
+   *
+   * The expected number of values before we have a collision can be approximated by (*):
+   * `Q(H) ~= sqrt( Pi * H / 2) = sqrt(Pi * 2^57) = 673 000 000`
+   *
+   * We don't expect to have anywhere close to that many channels at any given time on our node, so we don't need to
+   * check for duplicate values.
+   *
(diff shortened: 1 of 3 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
```diff
@@ -58,7 +58,24 @@ object ShortChannelId {
...
+   *
+   * We don't expect to have anywhere close to that many channels at any given time on our node, so we don't need to
+   * check for duplicate values.
+   *
+   * (*) https://en.wikipedia.org/wiki/Birthday_attack
+   */
+  private val aliasUpperBound = RealShortChannelId.apply(BlockHeight(350_000),1,0).toLong
+
+  def generateLocalAlias(): Alias = {
+    val l = randomLong() % aliasUpperBound
```
t-bast, line 76: Taking the modulo biases the distribution, but we probably don't care, it should be large enough for our use-case.

t-bast (approved):
It would be nice to also detect when we have a collision and log an error (which we can do for example in the register). This way if we completely messed up our maths or have an issue with our random generator, we will detect it?

eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
t-bast: Imagine for example that `randomLong()` returns values uniformly between 0, 1, 2 and 3. Each of them has a probability of 25% to be chosen. If you took the result modulo 3, then 0 would have probability 50% and 1 and 2 would have probability 25%, so we don't have a uniform distribution anymore.

I'm not sure it matters a lot in our case, but a better choice would be to loop until we get a value that is in the range we want.

pm47: Done in [09d32d8](https://github.com/ACINQ/eclair/pull/2337/commits/09d32d8567d212a9290355fe80f57479ff3065c6). 64 bits is such a huge space that it can take a few hundreds tries before finding a correct value.

pm47:
> It would be nice to also detect when we have a collision and log an error (which we can do for example in the register). This way if we completely messed up our maths or have an issue with our random generator, we will detect it?

Done in [3ff4831](https://github.com/ACINQ/eclair/pull/2337/commits/3ff48316a7f8ea81160a129de810a6d98934de93).

eclair-core/src/main/scala/fr/acinq/eclair/ShortChannelId.scala
pm47: Back to modulo in 88cbc77b6.

t-bast (approved):
LGTM :100:

PR #2338, opened 2022-06-30
Remove compatibility code for waitingSince

### LINKED ISSUE #2125: Remove compatibility code for `waitingSince`
We used to store UNIX timestamps in the `waitingSince` field before moving to block count. In order to ensure backward compatibility, we converted from timestamps to blockheight based on the value.

See https://github.com/ACINQ/eclair/pull/2041#issuecomment-954654115, it's almost time!

t-bast:
We used to store UNIX timestamps in the waitingSince field before moving to block count. In order to ensure backward compatibility, we converted from timestamps to blockheight based on the value.

This code has shipped more than a year ago, so we can safely remove that compatibility code since it only applies during the channel open or close period, which cannot last long anyway.

Fixes #2125

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
@@ -284,12 +284,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val 
               }
           }
           // no need to go OFFLINE, we can directly switch to CLOSING
-          if (closing.waitingSince.toLong > 1_500_000_000) {
-            // we were using timestamps instead of block heights when the channel was created: we reset it *and* we use block heights
-            goto(CLOSING) using closing.copy(waitingSince = nodeParams.currentBlockHeight) storing()
-          } else {
-            goto(CLOSING) using closing
-          }
+          goto(CLOSING) using closing
 
         case normal: DATA_NORMAL =>
           watchFundingTx(data.commitments)
@@ -314,12 +309,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val 
           watchFundingTx(funding.commitments)
           // we make sure that the funding tx has been published
           blockchain ! GetTxWithMeta(self, funding.commitments.fundingTxId)
-          if (funding.waitingSince.toLong > 1_500_000_000) {
-            // we were using timestamps instead of block heights when the channel was created: we reset it *and* we use block heights
-            goto(OFFLINE) using funding.copy(waitingSince = nodeParams.currentBlockHeight) storing()
-          } else {
-            goto(OFFLINE) using funding
-          }
+          goto(OFFLINE) using funding
 
         case funding: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED =>
           // we make sure that the funding tx with the highest feerate has been published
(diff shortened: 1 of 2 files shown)


pm47 (commented):
I cannot easily verify on our own node because the json serialization for `BlockHeight` is `{}` 🙄 

Here is a fix:
```patch
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala b/eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
index a03abfb41..2bb7bc22a 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/json/JsonSerializers.scala
@@ -36,7 +36,7 @@ import fr.acinq.eclair.transactions.DirectedHtlc
 import fr.acinq.eclair.transactions.Transactions._
 import fr.acinq.eclair.wire.protocol.MessageOnionCodecs.blindedRouteCodec
 import fr.acinq.eclair.wire.protocol._
-import fr.acinq.eclair.{CltvExpiry, CltvExpiryDelta, Feature, FeatureSupport, Alias, MilliSatoshi, ShortChannelId, TimestampMilli, TimestampSecond, UInt64, UnknownFeature, channel}
+import fr.acinq.eclair.{Alias, BlockHeight, CltvExpiry, CltvExpiryDelta, Feature, FeatureSupport, MilliSatoshi, ShortChannelId, TimestampMilli, TimestampSecond, UInt64, UnknownFeature, channel}
 import org.json4s
 import org.json4s.JsonAST._
@@ -151,6 +151,10 @@ object CltvExpiryDeltaSerializer extends MinimalSerializer({
   case x: CltvExpiryDelta => JInt(x.toInt)
 })

+object BlockHeightSerializer extends MinimalSerializer({
+  case x: BlockHeight => JInt(x.toInt)
+})
+
 object FeeratePerKwSerializer extends MinimalSerializer({
   case x: FeeratePerKw => JLong(x.toLong)
 })
@@ -539,6 +543,7 @@ object JsonSerializers {
     MilliSatoshiSerializer +
     CltvExpirySerializer +
     CltvExpiryDeltaSerializer +
+    BlockHeightSerializer +
     FeeratePerKwSerializer +
     ShortChannelIdSerializer +
     ChannelIdentifierSerializer +
diff --git a/eclair-core/src/test/scala/fr/acinq/eclair/json/JsonSerializersSpec.scala b/eclair-core/src/test/scala/fr/acinq/eclair/json/JsonSerializersSpec.scala
index 0249b9a53..763d3c4b0 100644
--- a/eclair-core/src/test/scala/fr/acinq/eclair/json/JsonSerializersSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/json/JsonSerializersSpec.scala
@@ -320,6 +320,10 @@ class JsonSerializersSpec extends AnyFunSuite with Matchers {
     }
   }

+  test("serialize block height") {
+    JsonSerializers.serialization.write(BlockHeight(123456))(JsonSerializers.formats) shouldBe "123456"
+  }
+
   /** utility method that strips line breaks in the expected json */
   def assertJsonEquals(actual: String, expected: String) = {
     val cleanedExpected = expected

```

t-bast:
We've now added a serializer for block heights, so you should be able to test this.

pm47:
Note to self: need to check after #2340 is deployed on our node.

pm47 (approved):
> Note to self: need to check after https://github.com/ACINQ/eclair/pull/2340 is deployed on our node.

Verified with query:
```sql
select json->>'type', is_closed, count(*)
from local.channels
where 
	json->>'type' in ('DATA_WAIT_FOR_FUNDING_CONFIRMED', 'DATA_CLOSING') 
	and (json->'waitingSince')::bigint > 1500000000
group by 1, 2
order by 1, 2
```

Which yields:

state | is_closed |count
------|-----------|-------
DATA_CLOSING|true|xxx
DATA_WAIT_FOR_FUNDING_CONFIRMED|true|xxx

PR #2339, opened 2022-06-30
Call `resolve()` on configuration

### LINKED ISSUE #2336: Variables in config not resolved
fr/acinq/eclair/NodeParams.scala

Tested version:
Eclair v0.7.0

Description:
When trying to use variables in the config file (eg. ${eclair.router.path-finding.default}  in the default example config file)
This error happens:

2022-06-25 13:26:58,118 ERROR- could not start eclaircom.typesafe.config.ConfigException$NotResolved: need to Config#resolve() each config before using it, see the API docs for Config#resolve()
	at com.typesafe.config.impl.SimpleConfig.checkValid(SimpleConfig.java:1103)
	at akka.actor.ActorSystem$Settings.<init>(ActorSystem.scala:388)
	at akka.actor.ActorSystemImpl.<init>(ActorSystem.scala:818)
	at akka.actor.ActorSystem$.apply(ActorSystem.scala:272)
	at akka.actor.ActorSystem$.apply(ActorSystem.scala:316)
	at akka.actor.ActorSystem$.apply(ActorSystem.scala:290)
	at fr.acinq.eclair.Boot$.delayedEndpoint$fr$acinq$eclair$Boot$1(Boot.scala:41)
	at fr.acinq.eclair.Boot$delayedInit$body.apply(Boot.scala:34)
	at scala.Function0.apply$mcV$sp(Function0.scala:39)
	at scala.Function0.apply$mcV$sp$(Function0.scala:39)
	at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:17)
	at scala.App.$anonfun$main$1(App.scala:76)
	at scala.App.$anonfun$main$1$adapted(App.scala:76)
	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:563)
	at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:561)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:919)
	at scala.App.main(App.scala:76)
	at scala.App.main$(App.scala:74)
	at fr.acinq.eclair.Boot$.main(Boot.scala:34)
	at fr.acinq.eclair.Boot.main(Boot.scala)

Reproducing steps
Use the reference.conf as your eclair.conf

Possible Solution:
Call the .resolve() function when loading the config file
(line 112 in NodeParams.scala)

pm47:
Calling `ConfigFactory.load()` automatically resolves the config (that is, processes substitutions), but that's not the case for `ConfigFactory.parseFile`, which means that substitutions don't work in `eclair.conf`.

We could resolve each of config separately, but the library's doc recommends one global `resolve()`, and frankly it makes more sense:
> A given Config must be resolved before using it to retrieve config values, but ideally should be resolved one time for your entire stack of fallbacks (see withFallback)

https://github.com/lightbend/config/blob/4458ea947a7a2a668bb811a122455f1f05975172/config/src/main/java/com/typesafe/config/Config.java#L204-L206

Fixes #2336.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
@@ -114,6 +114,7 @@ object NodeParams extends Logging {
     ConfigFactory.systemProperties()
       .withFallback(ConfigFactory.parseFile(new File(datadir, "eclair.conf")))
       .withFallback(ConfigFactory.load())
+      .resolve()
 
   private def readSeedFromFile(seedPath: File): ByteVector = {
     logger.info(s"use seed file: ${seedPath.getCanonicalPath}")


DerEwige:
Ok I checked my local source and imoh it should be

```
  def loadConfiguration(datadir: File): Config =
    ConfigFactory.systemProperties()
      .withFallback(ConfigFactory.parseFile(new File(datadir, "eclair.conf")).resolve())
      .withFallback(ConfigFactory.load())
```

pm47:
> imoh it should be

Why should it be that way? Your fix allows using subsitutions within `eclair.conf`, but wouldn't work with substitutions from `reference.conf`. My fix works for both cases and happens to be the recommended way as per the library authors.

PR #2342, opened 2022-07-05
Add support for Peerswap atomic swap based local liquidity management protocol

remyers:
This PR implements the draft [PeerSwap protocol ](https://github.com/ElementsProject/peerswap/blob/master/docs/peer-protocol.md).

We create a `SwapRegister` actor at setup which will spawn swap sender actors in response to user requests. The `SwapRegister` will also spawn swap receiver actors in response to a swap requests received from peers. The `SwapRegister` forwards swap messages based on their `swapId` field to the corresponding local swap actor or remote peer. When a swap actor has completed it simply stops which triggers the `SwapRegister` to remove it.

Project TODOs:
- [X] SwapInSender actor and happy path tests
- [X] SwapInReceiver actor and happy path tests
- [x] SwapOutSender actor and happy path tests
- [x] SwapOutReceiver actor and happy path tests
- [X] SwapIn sender/receiver cross actor tests
- [x] SwapOut sender/receiver cross actor tests
- [x] CLI for SwapInSender
- [ ] CLI and/or configuration for SwapInReceiver
- [x] CLI for SwapOutSender
- [x] CLI and/or configuration for SwapOutReceiver
- [x] happy path signet testing with [CLN plugin](https://github.com/ElementsProject/peerswap/blob/master/docs/signetguide_cln.md) (bitcoin only)
- [x] persist swap info to a database before committing onchain txs or offchain payments
- [ ] adjust timeouts and timeout behavior to ensure safe operation of the protocol
- [x] convert implementation into a Eclair plug-in

Protocol implementation TODOs:
- [x] Only a single swap should be active on a given channel at one time.
- [ ] Channels should fail updates that would remove liquidity needed for an active swap.

PeerSwap [specification](https://github.com/ElementsProject/peerswap-spec) suggestions/questions to discuss:
 - [ ] custom byte encoded message rather than json [#1](https://github.com/ElementsProject/peerswap-spec/issues/1)
 - [ ] use 32-byte channel ids instead of short channel ids [#2](https://github.com/ElementsProject/peerswap-spec/issues/2)
 - [ ] nit: rename “OpeningTxBroadcasted” to grammatically correct “OpeningTxBroadcast” [#3](https://github.com/ElementsProject/peerswap-spec/issues/3)
 - [ ] send a signature instead of a private key for cooperative closes [#4](https://github.com/ElementsProject/peerswap-spec/issues/4)


eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapInReceiver.scala
```diff
@@ -0,0 +1,247 @@
...
+
+  private val keyManager: SwapKeyManager = nodeParams.swapKeyManager
+  private val feeRatePerKw: FeeratePerKw = nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(target = nodeParams.onChainFeeConf.feeTargets.fundingBlockTarget)
+  private val premium = (feeRatePerKw * claimByInvoiceTxWeight / 1000).toLong // TODO: how should swap receiver calculate an acceptable premium?
+  private val swapId: String = request.swapId
+  private val takerPrivkey: PrivateKey = keyManager.openingPrivateKey(SwapKeyManager.keyPath(swapId)).privateKey
+  private val takerPubkey: PublicKey = takerPrivkey.publicKey
+  private val makerPubkey: PublicKey = PublicKey(ByteVector.fromValidHex(request.pubkey))
+
+  private def CreateSwap(): Behavior[SwapCommand] = {
```
sstone, line 113: nit: methods should start with a lowercase letter (this would be consistent with how we define behaviours in the rest of the code)

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapInSender.scala
```diff
@@ -0,0 +1,289 @@
...
+  implicit val timeout: Timeout = 30 seconds
+  private val keyManager: SwapKeyManager = nodeParams.swapKeyManager
+  private implicit val feeRatePerKw: FeeratePerKw = nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(target = nodeParams.onChainFeeConf.feeTargets.fundingBlockTarget)
+  private val maxPremium = (feeRatePerKw * claimByInvoiceTxWeight / 1000).toLong // TODO: how should swap sender calculate an acceptable premium?
+
+  private def makerPrivkey(): PrivateKey = keyManager.openingPrivateKey(SwapKeyManager.keyPath(swapId)).privateKey
+  private def makerPubkey(): PublicKey = makerPrivkey().publicKey
+  private def takerPubkey(agreement: SwapInAgreement): PublicKey = PublicKey(ByteVector.fromValidHex(agreement.pubkey))
+
+  private def CreateSwap(): Behavior[SwapCommand] = {
```
sstone, line 120: nit: same as above, should start with a lowercase letter

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -435,6 +439,30 @@ case class GossipTimestampFilter(chainHash: ByteVector32, firstTimestamp: Timest
...
 
+sealed trait PeerSwapMessage extends LightningMessage
+
+sealed abstract class JSonBlobMessage() extends PeerSwapMessage {
+  def json: String = {
+    Serialization.write(this)(PeerSwapJsonSerializers.formats)
+  }
+}
+
+case class SwapInRequest(protocolVersion: Long, swapId: String, asset: String, network: String, scid: String, amount: Long, pubkey: String) extends JSonBlobMessage with HasSwapId
```
sstone, line 450: Should we use more restrictive types here, `ByteVector32` for swap ids for example ?

eclair-core/src/test/scala/fr/acinq/eclair/swap/SwapTransactionsSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  case class Fixture(bitcoinClient: BitcoinCoreClient, publisher: ActorRef[FinalTxPublisher.Command], watcher: TestProbe, probe: TestProbe)
+
+  test("check validity of PeerSwap claim transactions") {
+    val f = createFixture()
+    import f._
+
+    val swapTxOut = makeSwapOpeningTxOut(amount, makerRefundPriv.publicKey, takerPaymentPriv.publicKey, paymentHash)
+    val future = wallet.makeFundingTx(swapTxOut.publicKeyScript, amount, feeratePerKw)
```
sstone, line 86: can be simplified to 

```
wallet.makeFundingTx(swapTxOut.publicKeyScript, amount, feeratePerKw).pipeTo(probe.ref)
```

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapData.scala
```diff
@@ -0,0 +1,29 @@
...
+import fr.acinq.bitcoin.scalacompat.ByteVector32
+import fr.acinq.eclair.ShortChannelId
+import fr.acinq.eclair.payment.Bolt11Invoice
+import fr.acinq.eclair.wire.protocol.{OpeningTxBroadcasted, SwapInAgreement, SwapInRequest}
+
+object SwapData {
+
+  final case class SwapInSenderData(channelId: ByteVector32, request: SwapInRequest, agreement: SwapInAgreement, invoice: Bolt11Invoice, openingTxBroadcasted: OpeningTxBroadcasted)
+
+  final case class SwapInReceiverData(shortChannelId: ShortChannelId, request: SwapInRequest, agreement: SwapInAgreement, invoice: Bolt11Invoice, openingTxBroadcasted: OpeningTxBroadcasted)
```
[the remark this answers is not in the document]
remyers: Ideally the PeerSwap spec would only use the 32 byte channel id which would be in the `SwapInRequest` message so we could then drop `channelId` and `shortChannelId` from these data classes.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
remyers: Until the spec is changed from json to byte encodings our codec will have to work with swapId as a string. I couldn't figure out a way to parse the json into messages with fields of a different type.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapInSender.scala
remyers: fixed in [03323a3](https://github.com/ACINQ/eclair/pull/2342/commits/03323a3699644757ed366f21339c8ad86ad699fe)

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapInReceiver.scala
remyers: fixed in [eb3e7d9](https://github.com/ACINQ/eclair/pull/2342/commits/eb3e7d9ca8a3d00a8678cd48f5ef4185aae9aa01)

eclair-core/src/test/scala/fr/acinq/eclair/swap/SwapTransactionsSpec.scala
remyers: much nicer! fixed in [3bd8c52](https://github.com/ACINQ/eclair/pull/2342/commits/3bd8c52b6a9989031aa9145b6868af49a2a3ccc9)

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapHelpers.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  def paymentEventAdapter(context: ActorContext[SwapCommand]): ActorRef[PaymentEvent] = context.messageAdapter[PaymentEvent](PaymentEventReceived)
+
+  def send(message: HasSwapId)(implicit context: ActorContext[SwapCommand], register: actor.ActorRef, shortChannelId: ShortChannelId): Unit =
+    register ! Register.ForwardShortId(forwardShortIdAdapter(context).toClassic, shortChannelId, message)
+
+  def forwardShortIdAdapter(context: ActorContext[SwapCommand]): ActorRef[Register.ForwardShortId[HasSwapId]] =
+    context.messageAdapter[Register.ForwardShortId[HasSwapId]](ForwardFailure)
+
+  def fundOpening(request: SwapInRequest, agreement: SwapInAgreement, invoice: Bolt11Invoice)(implicit context: ActorContext[SwapCommand], wallet: OnChainWallet, feeRatePerKw: FeeratePerKw): Unit = {
```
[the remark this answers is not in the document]
remyers: I couldn't filed any explanation for how to use pilots to pass local fixed parameters, so I instead I just used currying to delineate them instead of using implicit. I still pass `context` as an implicit parameter though. see[4fcceb5](https://github.com/ACINQ/eclair/pull/2342/commits/4fcceb51723076fe756dfe310161c988bdf20630).

I could also define and use a local function like the following, but it seemed clearer to use the helper directly.
```
private def mySend(message: HasSwapId) = send(register, shortChannelId)(message)
```
Does this approach look ok?

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapInSender.scala
```diff
@@ -0,0 +1,333 @@
...
+  private def makerPubkey(): PublicKey = makerPrivkey().publicKey
+  private def takerPubkey(agreement: SwapInAgreement): PublicKey = PublicKey(ByteVector.fromValidHex(agreement.pubkey))
+
+  private def createSwap(): Behavior[SwapCommand] = {
+    // setup conditions satisfied, create the swap
+    val pubKey = keyManager.openingPublicKey(SwapKeyManager.keyPath(swapId)).publicKey.toString()
+    queryOnChainBalance(wallet)
+
+    receiveSwapMessage[CreateSwapMessages](context, "createSwap") {
+      case QueryOnChainBalance(Success(onChainBalance: OnChainBalance)) if onChainBalance.confirmed < amount + maxPremium.sat =>
```
t-bast, line 119: I don't know if this check and the `case ChannelInfoResponse(RES_GET_CHANNEL_INFO(_, _, _, d: DATA_NORMAL)) if d.commitments.availableBalanceForReceive.truncateToSatoshi < amount` are really necessary, especially for a first version.

I'd rather stay simple at first and just abort later when we realize we actually cannot pay (either when failing to fund an on-chain transaction or failing to send an HTLC).

We can add nice-to-have checks later, I think we should just focus on an MVP at first, which will help focus on the core of the protocol.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapResponses.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  case class InternalError(swapId: String, reason: String) extends Error {
+    override def toString: String = s"swap $swapId internal error: $reason."
+  }
+
+  sealed trait Status extends Response
+  case class SwapInStatus(swapId: String, actor: String, behavior: String, channelId: ByteVector32, request: SwapInRequest, agreement_opt: Option[SwapInAgreement] = None, invoice_opt: Option[Bolt11Invoice] = None, openingTxBroadcasted_opt: Option[OpeningTxBroadcasted] = None) extends Status {
+    override def toString: String = s"$actor[$behavior]: $swapId, $channelId, $request, $agreement_opt, $invoice_opt, $openingTxBroadcasted_opt"
+  }
```
t-bast, line 78: I don't think we should expose that much internal information, `actor` and `behavior` are really internal details IMO.
But I think it would be valuable to have something like:

```scala
sealed trait SwapInStatus extends Status
case class AwaitingSwapInAgreement(...) extends SwapInStatus
case class AwaitOpeningTxConfirmation(...) extends SwapInStatus
...
```

The list of statuses we should expose should map to the _protocol_ statuses where we're waiting for an external event (a response from our peer or a blockchain event like tx confirmation). We should _not_ include transient statuses (such as `CreatingOpeningTx` for example), because these statuses will automatically move to a "real" status once an internal call completes, they would only clutter the status trait for no good reason. When the actor is in one of these transient statuses and receives a `GetStatus` request, it should simply delay it and it will eventually be handled once we're in a "real" status.

This is quite subtle, does it make sense (and do you think it's a good idea)?

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapHelpers.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  def getInfoAdapter(context: ActorContext[SwapCommand]): ActorRef[CommandSuccess[CMD_GET_CHANNEL_INFO]] = context.messageAdapter[CommandSuccess[CMD_GET_CHANNEL_INFO]](ChannelInfoResponse)
+
+  def queryOnChainBalance(wallet: OnChainWallet)(implicit context: ActorContext[SwapCommand]): Unit =
+    context.pipeToSelf(wallet.onChainBalance())(QueryOnChainBalance)
+
+  def payInvoice(nodeParams: NodeParams)(paymentInitiator: actor.ActorRef, swapId: String, invoice: Bolt11Invoice): Unit =
+    paymentInitiator ! SendPaymentToNode(invoice.amount_opt.get, invoice, nodeParams.maxPaymentAttempts, Some(swapId), assistedRoutes = invoice.routingInfo, nodeParams.routerConf.pathFindingExperimentConf.getRandomConf().getDefaultRouteParams, blockUntilComplete = true)
+
+  def watchForPayment(watch: Boolean)(implicit context: ActorContext[SwapCommand]): Unit =
```
t-bast, line 77: This is a good idea! But I think we should be more specific instead of registering to `PaymentEvent`. We should explicitly register only to `PaymentReceived` when we generated an invoice and wait for it to be paid, and we should only register to `PaymentSent` and `PaymentFailed` when we're paying an invoice.

t-bast (commented):
Very nice work, I think you clearly understood how to architect such a system: the flow between actors looks good.

The gist of my comments for this first review is that we should really focus this PR to be an MVP, as simple as possible. I think we can remove some nice-to-haves that can come later which would simplify the number of events that need to be handled, and will let us focus on the core of the peer swap protocol. This is important to ensure we don't have a bug in the core protocol as it could lead to loss of funds. I know it's tempting to make things better whenever you see a potential improvement (such as making state checks before accepting swaps), but once we have an MVP we can add those later (just keep a list somewhere of all the improvements you want to add in follow-up PRs). Otherwise the PR will be cluttered with a lot of comments on non-critical parts, and we may miss more important protocol issues.

Once that's done and we only have the core protocol to focus on, I think it's important to add the DB support: how we interact with the DB can be critical to avoid loss of funds, so that's a really important part to review and get right.

Also a small comment on types: you can do more typing work on the codecs to make the `case class`es directly contain the most accurate type (e.g. `ByteVector32` for `swapId`, `Satoshi` for amounts, etc).

remyers:
My latest push is rebased on master (from c1daaf3a), cleans up the commit history and has been tested on [signet](https://mempool.space/signet/tx/c21cbbd47b5bbfd9cfe3653ca60fc68c1a1f7a964d1dcc0d55ddf6ac949f0992) between two eclair nodes.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapData.scala
```diff
@@ -0,0 +1,29 @@
...
+package fr.acinq.eclair.swap
+
+import fr.acinq.bitcoin.scalacompat.ByteVector32
+import fr.acinq.eclair.ShortChannelId
+import fr.acinq.eclair.payment.Bolt11Invoice
+import fr.acinq.eclair.wire.protocol.{OpeningTxBroadcasted, SwapInAgreement, SwapInRequest}
+
+object SwapData {
+
+  final case class SwapInSenderData(channelId: ByteVector32, request: SwapInRequest, agreement: SwapInAgreement, invoice: Bolt11Invoice, openingTxBroadcasted: OpeningTxBroadcasted)
```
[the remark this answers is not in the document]
remyers: It's on my list of proposed spec changes, but right now it matches their spec.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapRegister.scala
```diff
@@ -0,0 +1,136 @@
...
+import scala.reflect.ClassTag
+
+object SwapRegister {
+  // @formatter:off
+  sealed trait Command
+  sealed trait ReplyToMessages extends Command {
+    def replyTo: ActorRef[Response]
+  }
+  sealed trait InitializingMessages extends ReplyToMessages
+  case class RestoreData(replyTo: ActorRef[Response], swaps: Set[SwapInSenderData]) extends Command with InitializingMessages
```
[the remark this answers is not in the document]
remyers: I've generalized SwapData to be the same for both actors, see [c612e69](https://github.com/ACINQ/eclair/pull/2342/commits/c612e694a7d638b672dd877cdd53ff9960608f0b). I assume that until the opening transaction is broadcast there is nothing relevant to backup - it's just a failed negotiation until that point.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapRegister.scala
```diff
@@ -0,0 +1,136 @@
...
+  case class SwapTerminated(swapInSenderId: SwapInSenderId) extends RegisteringMessages
+  case class ListPendingSwaps(replyTo: ActorRef[Iterable[Status]]) extends RegisteringMessages
+  case class CancelSwapRequested(replyTo: ActorRef[Response], swapId: String) extends RegisteringMessages with ReplyToMessages
+
+  sealed trait SwapId {
+    def id: String
+  }
+  case class SwapInSenderId(id: String) extends SwapId {
+    def toByteVector32: ByteVector32 = ByteVector32(ByteVector.fromValidHex(id))
+  }
```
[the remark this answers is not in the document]
remyers: Currently the json based PeerSwap protocol spec uses a string to represent `swapId` in protocol messages. Once the spec evolves to use byte encodings then that would clean up this sort of awkward reinterpretation. Perhaps to simplify review until then, the register should just map from `String`, the native type of `swapId`,  to an actor, for example:

```
private def registering(swaps: Map[String, ActorRef[SwapCommands.SwapCommand]]): Behavior[Command] = {
```

remyers:
> Very nice work, I think you clearly understood how to architect such a system: the flow between actors looks good.
> 
> The gist of my comments for this first review is that we should really focus this PR to be an MVP, as simple as possible. I think we can remove some nice-to-haves that can come later which would simplify the number of events that need to be handled, and will let us focus on the core of the peer swap protocol. This is important to ensure we don't have a bug in the core protocol as it could lead to loss of funds. I know it's tempting to make things better whenever you see a potential improvement (such as making state checks before accepting swaps), but once we have an MVP we can add those later (just keep a list somewhere of all the improvements you want to add in follow-up PRs). Otherwise the PR will be cluttered with a lot of comments on non-critical parts, and we may miss more important protocol issues.

Thanks for the detailed review so far and helpful advice. I agree that it make sense to strip it down as much as possible to the critical protocol elements. Now that the basic architecture appears correct, I have gone ahead and finished off the protocol by adding the remaining states necessary to support the `swap-out` workflow. This workflow is the same as `swap-in` but begins with a Lightning fee payment from the swap initiator instead of a `premium' paid by the on-chain (`maker`).

I will now look at incorporating your `SwapRegister` suggestions as a prelude to persisting the swap states to a DB.

remyers:
I'd also like to rename my two main swap actors as follows:
 - `SwapInSender` -> `SwapMaker`
 - `SwapInReceiver` -> `SwapTaker`

The `maker` actor is the one that swaps on-chain UTXO value via the `opening tx` for off-chain Lightning liquidity. The `taker` actor always swaps off-chain Lightning liquidity for on-chain UTXO  value. This naming would be more consistent with the terminology used in the current PeerSwap specification and is a necessary change now that my SwapIn* actors now participate in both swap-in and swap-out workflows.

eclair-core/src/main/scala/fr/acinq/eclair/io/Peer.scala
```diff
@@ -402,6 +408,11 @@ class Peer(val nodeParams: NodeParams, remoteNodeId: PublicKey, wallet: OnChainA
     self ! Peer.OutgoingMessage(msg, peerConnection)
   }
 
+  def replyUnknownSwap(peerConnection: ActorRef, unknownSwapId: String): Unit = {
```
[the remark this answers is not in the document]
remyers: Fixed.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapRegister.scala
```diff
@@ -0,0 +1,136 @@
...
+            .onFailure(typed.SupervisorStrategy.restart), "SwapInSender")
+          context.watchWith(swap, SwapTerminated(SwapInSenderId(state.request.swapId)))
+          swap ! RestoreSwapInSender(replyTo, state)
+          SwapInSenderId(state.request.swapId) -> swap.unsafeUpcast
+        }.toMap
+        registering(swaps)
+    }
+
+  private def registering(swaps: Map[SwapInSenderId, ActorRef[Any]]): Behavior[Command] = {
+    // TODO: fail requests for swaps on a channel if one already exists for the channel; keep a list of channels with active swaps
```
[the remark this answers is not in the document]
remyers: I think this is fixed in the latest PeerSwap plugin branch: [SwapRegister.scala](https://github.com/remyers/eclair/blob/peerswap-plugin/plugins/peerswap/src/main/scala/fr/acinq/eclair/plugins/peerswap/SwapRegister.scala#L155)

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapRegister.scala
```diff
@@ -0,0 +1,136 @@
...
+          context.log.error(s"received unhandled message while in state $stateName of ${m.getClass.getSimpleName}")
+          Behaviors.same
+      }
+
+  private def initializing: Behavior[Command] =
+    myReceive[InitializingMessages]("initializing") {
+      case RestoreData(replyTo, data) =>
+        val swaps = data.map { state =>
+          val swap: typed.ActorRef[SwapCommands.SwapCommand] = context.spawn(Behaviors.supervise(SwapInSender(nodeParams, watcher, register, wallet))
+            .onFailure(typed.SupervisorStrategy.restart), "SwapInSender")
```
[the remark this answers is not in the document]
remyers: Moved to [#2496](https://github.com/ACINQ/eclair/pull/2496#issuecomment-1340714570)

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapRegister.scala
```diff
@@ -0,0 +1,136 @@
...
+            Behaviors.same
+        }
+
+      case SwapTerminated(swapInSenderId) => registering(swaps - SwapInSenderId(swapInSenderId.id))
+
+      case ListPendingSwaps(replyTo: ActorRef[Iterable[Status]]) =>
+        // TODO: is this the best way to do this?!
+        val statuses: Iterable[Future[Status]] = swaps.values.map(swap => swap.ask(ref => GetStatus(ref))(1 milliseconds, context.system.scheduler))
+        replyTo ! statuses.map(v => Await.result(v, 1 milliseconds))
+        Behaviors.same
```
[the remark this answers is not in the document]
remyers: Fixed in #2496

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapInSender.scala
```diff
@@ -0,0 +1,333 @@
...
+      }
+    }
+}
+
+private class SwapInSender(replyTo: ActorRef[SwapResponses.Response], amount: Satoshi, swapId: String, channelId: ByteVector32, nodeParams: NodeParams, watcher: ActorRef[ZmqWatcher.Command], register: actor.ActorRef, wallet: OnChainWallet, implicit val context: ActorContext[SwapCommands.SwapCommand]) {
+  val protocolVersion = 1
+  val noAsset = ""
+  implicit val timeout: Timeout = 30 seconds
+  private val keyManager: SwapKeyManager = nodeParams.swapKeyManager
+  private implicit val feeRatePerKw: FeeratePerKw = nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(target = nodeParams.onChainFeeConf.feeTargets.fundingBlockTarget)
```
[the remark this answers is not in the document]
remyers: Fixed in #2496

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapHelpers.scala
remyers: Fixed in [ad33a79](https://github.com/ACINQ/eclair/pull/2496/commits/ad33a79074fc1151abea0820d0cc7c67821fb15d). This also forced some reorganization of the coded that handles resume/restart scenarios because a Taker actor check points right before trying to pay the invoice. 

Now the `payInvoice` method first checks if the payment is in the database before paying the invoice. This ensures the `Taker` will not be pay the fee or claim invoices twice after a restart or resume.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapInSender.scala
remyers: Good point; this check is removed in recent versions.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapInSender.scala
```diff
@@ -0,0 +1,333 @@
...
+      case QueryOnChainBalance(Success(_)) => queryChannelInfo(register, channelId)
+        Behaviors.same
+      case QueryOnChainBalance(Failure(_)) => swapCanceled(InternalError(swapId, "QueryOnChainBalance returned Failure"))
+      case ChannelInfoResponse(RES_GET_CHANNEL_INFO(_, _, _, d: DATA_NORMAL)) if d.commitments.availableBalanceForReceive.truncateToSatoshi < amount =>
+        swapCanceled(SwapResponses.InsufficientBalanceForReceive(swapId, amount, d.commitments.availableBalanceForReceive.truncateToSatoshi))
+      case ChannelInfoResponse(RES_GET_CHANNEL_INFO(_, _, _, d: DATA_NORMAL)) =>
+        val scid = d.shortIds.real.asInstanceOf[RealScidStatus.Final].realScid
+        awaitAgreement(SwapInRequest(protocolVersion, swapId, noAsset, NodeParams.chainFromHash(nodeParams.chainHash), scid.toString, amount.toLong, pubKey))
+      case CancelReceived(c) if c.swapId == swapId => swapCanceled(PeerCanceled(swapId))
+      case CancelReceived(_) => Behaviors.same
```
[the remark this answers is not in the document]
remyers: Fixed in recent versions.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapInSender.scala
```diff
@@ -0,0 +1,333 @@
...
+    }
+}
+
+private class SwapInSender(replyTo: ActorRef[SwapResponses.Response], amount: Satoshi, swapId: String, channelId: ByteVector32, nodeParams: NodeParams, watcher: ActorRef[ZmqWatcher.Command], register: actor.ActorRef, wallet: OnChainWallet, implicit val context: ActorContext[SwapCommands.SwapCommand]) {
+  val protocolVersion = 1
+  val noAsset = ""
+  implicit val timeout: Timeout = 30 seconds
+  private val keyManager: SwapKeyManager = nodeParams.swapKeyManager
+  private implicit val feeRatePerKw: FeeratePerKw = nodeParams.onChainFeeConf.feeEstimator.getFeeratePerKw(target = nodeParams.onChainFeeConf.feeTargets.fundingBlockTarget)
+  private val maxPremium = (feeRatePerKw * claimByInvoiceTxWeight / 1000).toLong // TODO: how should swap sender calculate an acceptable premium?
```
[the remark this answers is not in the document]
remyers: Added explanatory comments in [ad28e4e](https://github.com/ACINQ/eclair/pull/2496/commits/ad28e4e89ac2f4f4c38049f447b01bdebe837673), but note that the expected fee/premium system I am assuming is not in the spec. 

I assume a Taker that accepts a swap-in expects a premium is added to the on-chain swap amount to compensate them for the cost of spending the claim-by-invoice amount on-chain. 

Likewise the Maker that accepts a swap-out expects an off-chain fee payment before they open the swap to compensate them for if they need to reclaim their opening tx using claim-by-csv. A swap maker may *also* want to be compensated for the cost of liquidity they are providing.

eclair-core/src/main/scala/fr/acinq/eclair/swap/SwapResponses.scala
remyers: Good point! I've updated GetStatus to emit the minimal set of user meaningful messages, and moved the verbose information to the debug log: [3403272](https://github.com/ACINQ/eclair/pull/2496/commits/3403272cb76c79b3aa00f5739c9b07692c760123)

PR #2344, opened 2022-07-10
Add `onchainunspent` RPC call

rorp:
This PR add an RPC call to retrieve the list of UTXOs from the connected Bitcoin Core instance.


sstone:
What is the rationale for adding this to eclair's API ? Users could simply call bitcoin core's API instead ?

docs/release-notes/eclair-vnext.md
```diff
@@ -94,6 +94,7 @@ upgrading to this release.
 - `channelbalances`: retrieves information about the balances of all local channels (#2196)
 - `stop`: stops eclair. Please note that the recommended way of stopping eclair is simply to kill its process (#2233)
 - `channelbalances` and `usablebalances` return a `shortIds` object instead of a single `shortChannelId` (#2323)
+- `onchainunspent`: retrieves the list of UTXOs (#2244)
```
t-bast, line 97:
```suggestion
- `onchainunspent`: retrieves the list of UTXOs (#2344)
```

t-bast (commented):
I don't mind adding this RPC, but are you sure this will provide what you want?
An application using this RPC will need to gracefully handle a few edge cases, among which:

- the utxo it selects is actually locked because `eclair` is in the process of using it for a transaction
- the utxo it selects is unlocked, but `eclair` also wants to use it in parallel and creates a conflicting transaction

The only way to work around this would be to expose `bitcoind`'s locking RPCs, which opens a real can of worms we don't want to get into.

t-bast:
@rorp can you reply to the previous comments? Otherwise I'll close this PR as inactive.

rorp:
I agree that it should differentiate between locked and unlocked UTXO.

As @sstone pointed out, all this can be done using Bitcoin Core RPC directly, so the can of worm is already wide open.
 
The lock/unlock UTXO API would be beneficial, because it could perform additional checks. For example, like LND's `leaseOutput`/`releaseOutput` require an app id, so the app cannot unlock outputs locked by Eclair or other apps.

t-bast:
> For example, like LND's leaseOutput/releaseOutput require an app id, so the app cannot unlock outputs locked by Eclair or other apps.

That can only be done when implementing your own bitcoin wallet. Since eclair uses bitcoind instead of re-implementing a wallet, this is impossible for us. That's why I think we simply should not go down that road, because we can't provide the necessary guarantees to the caller.

This should be done by querying bitcoind directly, I don't think we should add even more APIs to eclair for that.

t-bast:
Shall we close this PR? I'm not clear on what the status is here @rorp

PR #2349, opened 2022-07-22
Remove redundant routing hint

thomash-acinq:
In `SendPaymentToNode`, the routing hints were present twice: once in `invoice` and once in `extraEdges`. I'm only keeping `invoice`.
Also make the `ExtraEdge` trait use `def` as there may be problems with `val`.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/MinimalNodeFixture.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/MinimalNodeFixture.scala
@@ -282,9 +282,23 @@ object MinimalNodeFixture extends Assertions with EitherValues {
     val sender = TestProbe("sender")
     sender.send(node2.paymentHandler, MultiPartHandler.ReceivePayment(Some(amount), Left("test payment")))
     val invoice = sender.expectMsgType[Bolt11Invoice]
+    val invoiceWithHints = Bolt11Invoice(
+      chainHash = node2.nodeParams.chainHash,
+      amount = invoice.amount_opt,
+      paymentHash = invoice.paymentHash,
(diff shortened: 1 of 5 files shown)


pm47 (commented):
LGTM, can you just remove this import from `PaymentInitiator`:

```
import fr.acinq.eclair.payment.send.MultiPartPaymentLifecycle.{PreimageReceived, SendMultiPartPayment}
```

and use `MultiPartPaymentLifecycle.` as prefix in the file, for consistency with single part payments?

pm47 (approved):

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/fixtures/MinimalNodeFixture.scala
```diff
@@ -282,9 +282,23 @@ object MinimalNodeFixture extends Assertions with EitherValues {
...
+      privateKey = node2.nodeParams.privateKey,
+      description = invoice.description,
+      minFinalCltvExpiryDelta = invoice.minFinalCltvExpiryDelta,
+      fallbackAddress = invoice.fallbackAddress(),
+      expirySeconds = invoice.tags.collectFirst { case expiry: Bolt11Invoice.Expiry => expiry.toLong },
+      extraHops = hints.map(_.toList).toList,
+      timestamp = invoice.createdAt,
+      paymentSecret = invoice.paymentSecret.get,
+      paymentMetadata = invoice.paymentMetadata,
+      features = invoice.features)
```
pm47, line 298: I missed this, can you fix in #2350?

```suggestion
    assert(invoice.routingInfo.isEmpty)
    val invoiceWithHints = invoice
      .modify(_.tags).using(_ ++ hints.map(hops => Bolt11Invoice.RoutingInfo(hops.toList)))
```

PR #2351, opened 2022-07-22
Update offers types

thomash-acinq:
Most notable changes:
- the node id is now 33 bytes like everywhere else
- blinded paths are mandatory in invoices


eclair-core/src/main/scala/fr/acinq/eclair/crypto/Sphinx.scala
```diff
@@ -410,6 +411,19 @@ object Sphinx extends Logging {
...
 
+    /**
+     * Creates a direct blinded route with a single node
+     *
+     * @param sessionKey this node's session key.
+     * @param nodeId     public key of the target node, which is also the introduction point.
+     * @param selfId     payloads that should be encrypted for each node on the route.
+     * @return a blinded route.
+     */
+    def createDirect(sessionKey: PrivateKey, nodeId: PublicKey, selfId: ByteVector): BlindedRoute = {
```
t-bast, line 422: I don't think this function belongs here, it's more an application-level utility. Maybe it makes more sense in `Bolt12Invoice.scala` (or maybe it's not even need yet, it will be added once we add code to find suitable blinded paths)?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -145,27 +146,38 @@ object Bolt12Invoice {
    * @param preimage the preimage to use for the payment
    * @param nodeKey  the key that was used to generate the offer, may be different from our public nodeId if we're hiding behind a blinded route
    * @param features invoice features
+   * @param selfId   data to identify the payment when receiving it, can contain anything and is not readable by the payer
```
t-bast, line 149: Can you explain why this is named `selfId` instead of `pathId`?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/MessageOnion.scala
```diff
@@ -118,17 +119,19 @@ object MessageOnionCodecs {
     .typecase(UInt64(68), variableSizeBytesLong(varintoverflow, invoiceErrorCodec.as[InvoiceError]))
 
 
-  val perHopPayloadCodec: Codec[TlvStream[OnionMessagePayloadTlv]] = TlvCodecs.lengthPrefixedTlvStream[OnionMessagePayloadTlv](onionTlvCodec).complete
+  val perHopPayloadCodec: Codec[TlvStream[OnionMessagePayloadTlv]] = TlvCodecs.tlvStream[OnionMessagePayloadTlv](onionTlvCodec).complete
 
-  val relayPerHopPayloadCodec: Codec[RelayPayload] = perHopPayloadCodec.narrow({
+  val prefixedPerHopPayloadCodec: Codec[TlvStream[OnionMessagePayloadTlv]] = variableSizeBytesLong(CommonCodecs.varintoverflow, perHopPayloadCodec)
```
t-bast, line 124: This change seems unnecessary, you're not using the non-length-prefixed `perHopPayloadCodec` anywhere?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferCodecs.scala
```diff
@@ -130,16 +129,15 @@ object OfferCodecs {
   })
 
   private val paymentInfo: Codec[PaymentInfo] = (("fee_base_msat" | millisatoshi32) ::
-    ("fee_proportional_millionths" | tu32) ::
-    ("cltv_expiry_delta" | cltvExpiryDelta)).as[PaymentInfo]
+    ("fee_proportional_millionths" | uint32) ::
+    ("cltv_expiry_delta" | cltvExpiryDelta) ::
+    ("htlc_minimum_msat" | millisatoshi) ::
+    ("htlc_maximum_msat" | millisatoshi) ::
+    ("features" | variableSizeBytesLong(uint32, bytes).xmap[Features[Feature]](Features(_), _.toByteVector))).as[PaymentInfo]
```
t-bast, line 136: Why is this using a `uint32` for the length whereas everywhere else we use `uint16` for features?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -153,17 +149,17 @@ object Offers {
     val quantityMin: Option[Long] = records.get[QuantityMin].map(_.min)
     val quantityMax: Option[Long] = records.get[QuantityMax].map(_.max)
 
-    val nodeIdXOnly: ByteVector32 = records.get[NodeId].get.xonly
+    val nodeId: PublicKey =
+      records.get[NodeId].map(_.publicKey).getOrElse(records.get[Paths].get.paths.head.blindedNodes.last.blindedPublicKey)
```
t-bast, line 153: I'm not sure about using a `node_id` from the first blinded path here.
What about making `node_id` required regardless of whether there is a blinded path?
This gives more flexibility to the offer creator, doesn't it? The blinded paths don't have to match the offer's `node_id` this way, they are a completely separate concern.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -153,17 +149,17 @@ object Offers {
...
 
     val refundFor: Option[ByteVector32] = records.get[RefundFor].map(_.refundedPaymentHash)
 
     val signature: Option[ByteVector64] = records.get[Signature].map(_.signature)
 
-    val contact: Seq[OnionMessages.Destination] =
-      records.get[Paths].flatMap(_.paths.headOption).map(OnionMessages.BlindedPath).map(Seq(_))
-        .getOrElse(Seq(PublicKey(2.toByte +: nodeIdXOnly), PublicKey(3.toByte +: nodeIdXOnly)).map(nodeId => OnionMessages.Recipient(nodeId, None, None)))
+    val contact: Either[Seq[BlindedRoute], PublicKey] =
+      records.get[Paths].map(_.paths).map(Left(_)).getOrElse(Right(records.get[NodeId].get.publicKey))
```
t-bast, line 162: Shouldn't we prioritize the `NodeId` over a blinded path if it's available? Note that this will be redundant with `NodeId` if we make `NodeId` required in all offers.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -145,27 +146,38 @@ object Bolt12Invoice {
...
       Some(Chain(request.chain)),
-      Some(CreatedAt(TimestampSecond.now())),
-      Some(PaymentHash(Crypto.sha256(preimage))),
       Some(OfferId(offer.offerId)),
-      Some(NodeId(nodeKey.publicKey)),
-      Some(Amount(request.amount.orElse(offer.amount.map(_ * request.quantity)).get)),
+      Some(Amount(amount)),
       Some(Description(offer.description)),
+      if (!features.isEmpty) Some(FeaturesTlv(features.unscoped())) else None,
+      Some(Paths(Seq(Sphinx.RouteBlinding.createDirect(randomKey(), nodeKey.publicKey, selfId)))),
```
t-bast, line 166: Does it really make sense to include a 1-hop dummy blinded path (that's what we're doing here, right?)?
I think the blinded paths should be provided as arguments to this constructor instead, because they will be chosen and created by another component, won't they?

It feels to me that the flow will be:

- we receive an invoice request for an offer we created
- we extract an amount and use the graph to find suitable blinded paths
- we create the corresponding bolt 12 invoice
- we send it back to the requestor

Does that look like what you had in mind?

thomash-acinq: Made `path` an argument.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/MessageOnion.scala
thomash-acinq: Removed

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferCodecs.scala
thomash-acinq: It was a mistake. Fixed

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
thomash-acinq: In case the offer creator wants to use a blinded path, it doesn't make sense to require an additional `node_id` which would likely just be a random single-use key. It would just use QR code space, that's it.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
thomash-acinq: No, if there is a blinded path, we MUST use it. And if there is also a `node_id`, we should just ignore it.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/MessageOnion.scala
```diff
@@ -106,17 +106,17 @@ object MessageOnionCodecs {
     .typecase(UInt64(68), variableSizeBytesLong(varintoverflow, invoiceErrorCodec.as[InvoiceError]))
 
 
-  val perHopPayloadCodec: Codec[TlvStream[OnionMessagePayloadTlv]] = TlvCodecs.lengthPrefixedTlvStream[OnionMessagePayloadTlv](onionTlvCodec).complete
+  val prefixedPerHopPayloadCodec: Codec[TlvStream[OnionMessagePayloadTlv]] = TlvCodecs.lengthPrefixedTlvStream[OnionMessagePayloadTlv](onionTlvCodec).complete
```
t-bast, line 109: nit: either revert to `perHopPayloadCodec` or `lengthPrefixedPerHopPayloadCodec`?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferCodecs.scala
```diff
@@ -130,16 +129,15 @@ object OfferCodecs {
   })
 
   private val paymentInfo: Codec[PaymentInfo] = (("fee_base_msat" | millisatoshi32) ::
-    ("fee_proportional_millionths" | tu32) ::
-    ("cltv_expiry_delta" | cltvExpiryDelta)).as[PaymentInfo]
+    ("fee_proportional_millionths" | uint32) ::
+    ("cltv_expiry_delta" | cltvExpiryDelta) ::
+    ("htlc_minimum_msat" | millisatoshi) ::
+    ("htlc_maximum_msat" | millisatoshi) ::
+    ("features" | variableSizeBytes(uint16, bytes).xmap[Features[Feature]](Features(_), _.toByteVector))).as[PaymentInfo]
```
t-bast, line 136: Can you reuse the `lengthPrefixedFeaturesCodec`?

```suggestion
    ("features" | lengthPrefixedFeaturesCodec).as[PaymentInfo]
```

t-bast (commented):
Looks good to me, only a couple nits

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/MessageOnion.scala
thomash-acinq: Done, I prefer to keep the `lengthPrefixed` explicit, I was bitten by it before.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferCodecs.scala
thomash-acinq: Done

t-bast (approved):

PR #2352, opened 2022-07-22
Fix flaky zeroconf test

pm47:
The flaky test was `ZeroConfAliasIntegrationSpec`.`a->b->c (b-c private)`.

Due to a race between node watchers, we were sometimes entering into this scenario between bob and carol:
1) carol's watcher wins the race and sends her `channel_ready` to bob before bob has received `WatchFundingConfirmedTriggered` from his own watcher
2) the channel isn't officially zeroconf, but bob gladly accepts carol's early `channel_ready` and moves directly to `WAIT_FOR_CHANNEL_READY` and then `NORMAL`, *without a real scid*.
3) bob ignores the `WatchFundingConfirmedTriggered` that it receives in state `WAIT_FOR_CHANNEL_READY` or `NORMAL`
4) bob later receives the `WatchFundingDeeplyBuriedTriggered`, discovers his own real scid and emits a `ShortChannelIdAssigned` event, but the channel relayer only listens to `LocalChannelUpdate`.
5) subsequent a->b->c payment using a real scid for b-c as routing hint fails because b's relayer doesn't know it.

There are two ways of fixing this:
a) make the relayer listen to `ShortChannelIdAssigned` (like it did before)
b) emit a `LocalChannelUpdate` event containing the new real scid and the same `channel_update`.

I went with b) but it's really a tie. In general I'm really not satisfied with the `LocalChannelUpdate`/`ShortChannelIds` duality.


t-bast:
Note: `PaymentIntegrationSpec.wait for network announcements` also seems to be flaky, but that will be for another PR.

t-bast (approved):

PR #2353, opened 2022-07-22
Use Invoice instead of Bolt11Invoice

thomash-acinq:
Use the generic trait `Invoice` as much as possible instead of `Bolt11Invoice`


t-bast (commented):
This is actually quite impacted by #2253, I think it's worth rebasing on top of `master` now that it has been merged to see how it impacts this PR.

PR #2354, opened 2022-07-22
(Minor) Test improvements

pm47:
Minor improvements that can be reviewed commit by commit.

Note on log capture:
- akka version (`... with LogCapturing`): only works when tests are run sequentially (that is: useless on CI), otherwise outputs of all test are mixed together
- my version: separate between tests, but ugly formatting, because akka's `StdOutLogger` dosen't use logback settings.


eclair-core/src/test/scala/fr/acinq/eclair/testutils/FixtureSpec.scala
```diff
@@ -66,17 +71,52 @@ abstract class FixtureSpec extends FixtureAnyFunSuite
...
+    // we use reflection to find actorsystems to listen to
+    val testSystem = f.getClass.getDeclaredMethods
+      .collect {
+        case m if m.getReturnType == classOf[ActorSystem] => m.invoke(f).asInstanceOf[ActorSystem]
+      }.head
+    val systems = f.getClass.getDeclaredMethods
+      .collect {
+        case m if m.getReturnType == classOf[MinimalNodeFixture] => m.invoke(f).asInstanceOf[MinimalNodeFixture].system
+      }
+      .toSet
```
t-bast, line 109: That doesn't feel very robust...can't we avoid relying on hacky reflection and instead force fixtures to expose the actor systems they use? Since we'll never create actor systems on-the-fly during tests and only ever create them when initializing fixtures, I think we should be able to do better than this.

pm47: This is what I tried first, but it polluted the fixtures with details required for this hacky custom logger, and I wanted to keep the fixtures as lean as possible. How about:

```suggestion
    val (testSystem, systems) = f match {
      case f: TwoNodesFixture => (f.system, Set(f.alice.system, f.bob.system))
      case f: ThreeNodesFixture => (f.system, Set(f.alice.system, f.bob.system, f.carol.system))
    }
```

It's conceptually the same as what you propose (no reflection, and explicit declaration per-fixture), and doesn't touch the fixture classes. The downside is that we have to declare all new fixtures that we want to log, and it can be a bit tedious for fixtures with many nodes.

t-bast: Even with those limitations, it still feels better than the reflection approach, I think it's worth it (especially since log capturing isn't something we strongly need too often, so it's ok if we forget to declare some fixtures and just add them when we actually need to investigate).

pm47: I agree it's better, done in 927fa7bef.

t-bast (approved):
LGTM

PR #2355, opened 2022-07-23
Emit event when receiving `Pong`

DerEwige:
create pong event and publish pongs, so plugins can subscribe to the latency of the peers

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerEvents.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerEvents.scala
@@ -31,4 +31,6 @@ case class PeerDisconnected(peer: ActorRef, nodeId: PublicKey) extends PeerEvent
 
 case class LastChannelClosed(peer: ActorRef, nodeId: PublicKey) extends PeerEvent
 
+case class receivedPong(latency: FiniteDuration, nodeId: PublicKey) extends PeerEvent
+
 case class UnknownMessageReceived(peer: ActorRef, nodeId: PublicKey, message: UnknownMessage, connectionInfo: ConnectionInfo) extends PeerEvent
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/io/PeerEvents.scala
```diff
@@ -31,4 +31,6 @@ case class PeerDisconnected(peer: ActorRef, nodeId: PublicKey) extends PeerEvent
 
 case class LastChannelClosed(peer: ActorRef, nodeId: PublicKey) extends PeerEvent
 
+case class receivedPong(latency: FiniteDuration, nodeId: PublicKey) extends PeerEvent
```
t-bast, line 34: nit: to better match the rest of the codebase:

```suggestion
case class PongReceived(nodeId: PublicKey, latency: FiniteDuration) extends PeerEvent
```

t-bast (commented):
Concept ACK, just some cosmetic changes needed.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerEvents.scala
DerEwige: if you change the order in the event, you need to change the publush line too in PeerConnection.scala

t-bast: Yes, that's what I expect you to do when fixing this comment :)

t-bast: Also, you haven't properly imported `FiniteDuration`, the build is failing, that should be fixed.

t-bast (approved):

PR #2356, opened 2022-07-24
changing first finding routes to debug

DerEwige:
the first finding routes messages can be very big (if you have hundreds of excluded nodes/channels)
if you do thousands of routing finds per hour this generates hundreds of megabytes of log data each our.

This message should better only be logged on debug level


t-bast:
I disagree, this log message is generally useful. The case where you have many excluded nodes/channels is not the general case, it's rather the exception, so that's not what we should optimize for. We should optimize logging for most users, if we switch that to `debug` every current eclair user will lose this log line.

t-bast:
Sounds good! Closing this PR then.

PR #2359, opened 2022-08-01
Offline commands plugin

t-bast:
This PR introduces a new `plugins` directory to maintain official eclair plugins. It contains our first official plugin, the offline commands plugin. This plugin lets a node operator issue commands regardless of whether the target peer is online.
The plugin will store these commands and execute them once the target comes back online.

Its main use-case is to request closing a given channel with a timeout after which we should force-close. This feature has been requested several times by various node operators (most recently by @viaj3ro).

The PR is split in several independent commits, which can act as guidelines for future plugin writers (and should be converted to better documentation at some point).

The main point I'd like to discuss is whether those official plugins (which will contain `PeerSwap` as well) should directly be in the `eclair` repository or in a dedicated `eclair-plugins` repository. Here are the advantages and drawbacks that I see for each approach:

- Include plugins in the `eclair` repository:
  - Advantages: changes to eclair that break plugins are automatically detected and cannot be added to `master` without fixing the plugin first
  - Drawbacks: the build and tests are longer, which is painful for day-to-day development
- Create a new `eclair-plugins` repository:
  - Advantages: the `eclair` repository stays small with a fast build/test
  - Drawbacks: some changes to `eclair` may be merged that break plugins: we can add a nightly CI run on the `eclair-plugins` repository to build and test using the latest `eclair` master, which ensures we catch those quickly, but if they require completely re-working an `eclair` PR this isn't ideal. But maybe in practice that would happen often and we would just need to update the plugin code, which is ok?

I'm tempted to start with a separate repository for plugins, and later move them inside `eclair` if that turns out to be too painful in practice, what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,12 +245,15 @@ Here is how to run Eclair with plugins:
 eclair-node-<version>-<commit_id>/bin/eclair-node.sh <plugin1.jar> <plugin2.jar> <...>
 ```
 
-### Non-exhaustive plugins list
+We maintain some plugins in the [plugins](./plugins/) directory.
+The [offline commands plugin](./plugins/offline-commands/) is a good example of how to write a plugin that interacts with various features of eclair.
 
-Here are some plugins created by the Eclair community.
+There are also plugins provided by external contributors from the Eclair community.
+We provide a non-exhaustive list of these plugins below.
 If you need support for these plugins, head over to their respective github repository.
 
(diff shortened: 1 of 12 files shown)


t-bast:
This plugin has moved to the [eclair plugins repository](https://github.com/ACINQ/eclair-plugins)

PR #2361, opened 2022-08-03
Make `htlc_maximum_msat` mandatory in channel updates

t-bast:
The specification is removing support for old channel updates that didn't include an `htlc_maximum_msat` (https://github.com/lightning/bolts/pull/996).

Every implementation has been generating updates containing this field for years, so we can safely reject updates that don't contain it.


eclair-core/src/test/scala/fr/acinq/eclair/wire/internal/channel/ChannelCodecsSpec.scala
```diff
@@ -100,38 +100,6 @@ class ChannelCodecsSpec extends AnyFunSuite {
     assert(data_new == data_new2)
   }
 
-  test("backward compatibility DATA_NORMAL_COMPAT_03_Codec (roundtrip)") {
```
t-bast, line 103: I tried fixing this test, but it's a real pain because it uses one of our oldest codecs that has been updated years ago. I think we can safely drop it.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -333,36 +328,26 @@ object LightningMessageCodecs {
...
-          ("feeBaseMsat" | millisatoshi32) ::
-          ("feeProportionalMillionths" | uint32) ::
-          ("htlcMaximumMsat" | conditional(messageFlags.optionChannelHtlcMax, millisatoshi))
-      }).derive[MessageFlags].from {
-        // The purpose of this is to tell scodec how to derive the message flags from the data, so we can remove that field
-        // from the codec definition and the case class, making it purely a serialization detail.
-        // see: https://github.com/scodec/scodec/blob/series/1.11.x/unitTests/src/test/scala/scodec/examples/ProductsExample.scala#L108-L127
-        case _ :: _ :: _ :: _ :: _ :: htlcMaximumMsat_opt :: HNil => MessageFlags(optionChannelHtlcMax = htlcMaximumMsat_opt.isDefined)
-      }
+      ("messageFlags" | constant(hex"01")) ::
```
pm47, line 331: This drops the following `Unit` codec and removes the need for `:: () ::` when applying the codec

```suggestion
      ("messageFlags" | constant(hex"01")) :~>:
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -333,36 +328,26 @@ object LightningMessageCodecs {
...
-          ("htlcMinimumMsat" | millisatoshi) ::
-          ("feeBaseMsat" | millisatoshi32) ::
-          ("feeProportionalMillionths" | uint32) ::
-          ("htlcMaximumMsat" | conditional(messageFlags.optionChannelHtlcMax, millisatoshi)) ::
-          ("tlvStream" | ChannelUpdateTlv.channelUpdateTlvCodec)
-      })).derive[MessageFlags].from {
-      // same comment above
-      case _ :: _ :: _ :: _ :: _ :: _ :: _ :: _ :: htlcMaximumMsat_opt :: _ :: HNil => MessageFlags(optionChannelHtlcMax = htlcMaximumMsat_opt.isDefined)
-    }
+      ("messageFlags" | constant(hex"01")) ::
```
pm47, line 343:
```suggestion
      ("messageFlags" | constant(hex"01")) :~>:
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -321,10 +320,6 @@ object LightningMessageCodecs {
     ("signature" | bytes64) ::
       nodeAnnouncementWitnessCodec).as[NodeAnnouncement]
 
-  private case class MessageFlags(optionChannelHtlcMax: Boolean)
```
pm47, line 324: I would have kept this class, but I see you have reintroduced it in #2362.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -385,10 +385,9 @@ case class ChannelUpdate(signature: ByteVector64,
                          htlcMinimumMsat: MilliSatoshi,
                          feeBaseMsat: MilliSatoshi,
                          feeProportionalMillionths: Long,
-                         htlcMaximumMsat: Option[MilliSatoshi],
+                         htlcMaximumMsat: MilliSatoshi,
                          tlvStream: TlvStream[ChannelUpdateTlv] = TlvStream.empty) extends RoutingMessage with AnnouncementMessage with HasTimestamp with HasChainHash {
-
-  def messageFlags: Byte = if (htlcMaximumMsat.isDefined) 1 else 0
+  def messageFlags: Byte = 1
```
pm47, line 390: We need this to construct `ChannelDisabled` error messages, which should really only use `channelFlags`.

t-bast: Yes, it looks a bit weird here if it's just a constant, but if we reintroduce more message flags (such as in #2362) it makes a bit more sense. Let's see if #2362 is accepted at the spec level and we'll see.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
t-bast: Neat thanks, done in https://github.com/ACINQ/eclair/pull/2361/commits/06b4ad952e69524ffedeec2adaad516f5e6e75ae

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2361/commits/06b4ad952e69524ffedeec2adaad516f5e6e75ae

pm47 (approved):

PR #2362, opened 2022-08-03
Add private flag to channel updates

t-bast:
Add a message flag to channel update to specify that an update is private and should not be rebroadcast to other nodes.
We log an error if a private channel update gets into the rebroadcast list.

I also tried strongly typing channel updates, dividing them into `PrivateChannelUpdate` and `PublicChannelUpdate` to get compile-time guarantees that private channel updates wouldn't leak where they're not supposed to, but unfortunately I hit some roadblocks...the main issue with this approach is that we would need `Router.PublicChannel` and `Router.PrivateChannel` to map to these two types of channel updates, but that's not how we use them internally: when a public channel isn't confirmed yet, we store it as a private channel which messes things up. Ideally we would need to rework this to clearly separate channels that will be public once confirmed from channels that will stay private, but that's quite a lot of work for not much benefits, so I'm leaving it aside for now.

cln has already merged support for this on their `master` branch, and I have successfully tested interop with them.

See https://github.com/lightning/bolts/pull/999


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -317,6 +317,8 @@ object LightningMessageCodecs {
     ("signature" | bytes64) ::
       nodeAnnouncementWitnessCodec).as[NodeAnnouncement]
 
+  val messageFlagsCodec = ("messageFlags" | (ignore(6) :: bool :: constant(bin"1"))).as[ChannelUpdate.MessageFlags]
```
pm47, line 320:
```suggestion
  val messageFlagsCodec = ("messageFlags" | (ignore(6) :: bool :: constant(1))).as[ChannelUpdate.MessageFlags]
```

Using scodec's:
```scala
def constant[A: Integral](bits: A*): Codec[Unit] = constant(BitVector(bits: _*))
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/FailureMessage.scala
```diff
@@ -57,7 +57,7 @@ case object UnknownNextPeer extends Perm { def message = "processing node does n
 case class AmountBelowMinimum(amount: MilliSatoshi, update: ChannelUpdate) extends Update { def message = s"payment amount was below the minimum required by the channel" }
 case class FeeInsufficient(amount: MilliSatoshi, update: ChannelUpdate) extends Update { def message = s"payment fee was below the minimum required by the channel" }
 case object TrampolineFeeInsufficient extends Node { def message = "payment fee was below the minimum required by the trampoline node" }
-case class ChannelDisabled(messageFlags: Byte, channelFlags: ChannelUpdate.ChannelFlags, update: ChannelUpdate) extends Update { def message = "channel is currently disabled" }
+case class ChannelDisabled(messageFlags: ChannelUpdate.MessageFlags, channelFlags: ChannelUpdate.ChannelFlags, update: ChannelUpdate) extends Update { def message = "channel is currently disabled" }
```
pm47, line 60: Finally, consistency! ✌️

eclair-core/src/test/resources/nonreg/codecs/000003-DATA_NORMAL/fundee-announced/data.json
```diff
@@ -133,6 +133,9 @@
       "iso" : "2019-06-18T12:49:33Z",
       "unix" : 1560862173
     },
+    "messageFlags" : {
+      "isPrivate" : false
+    },
```
pm47, line 138: Love the pain-free nice serialization.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerConnection.scala
```diff
@@ -290,6 +289,9 @@ class PeerConnection(keyPair: KeyPair, conf: PeerConnection.Conf, switchboard: A
           case (count, (msg, origins)) if !timestampInRange(msg, origins, d.gossipTimestampFilter) =>
             // the peer has set up a filter on timestamp and this message is out of range
             count
+          case (count, (u: ChannelUpdate, _)) if u.isPrivate =>
+            log.error("private channel update for scid={} was incorrectly added to staggered broadcast", u.shortChannelId)
```
pm47, line 293: So this is the actual improvement brought by the spec change. Less nice that strong typing all the way down, but still very useful as runtime debugging.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -422,21 +422,24 @@ case class ChannelUpdate(signature: ByteVector64,
...
-  def messageFlags: Byte = 1
+  val isPrivate: Boolean = messageFlags.isPrivate
 
   def toStringShort: String = s"cltvExpiryDelta=$cltvExpiryDelta,feeBase=$feeBaseMsat,feeProportionalMillionths=$feeProportionalMillionths"
 
   def relayFees: Relayer.RelayFees = Relayer.RelayFees(feeBase = feeBaseMsat, feeProportionalMillionths = feeProportionalMillionths)
 }
 
 object ChannelUpdate {
+  case class MessageFlags(isPrivate: Boolean)
```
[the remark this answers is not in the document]
t-bast: Renamed in https://github.com/ACINQ/eclair/pull/2362/commits/1fd715c1c6aeb72bab4fd313501f5868a6dd2b1c

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
t-bast: Unfortunately not...if you change that all the codec tests break because it actually doesn't include just a bit (I don't dive into why it doesn't behave as expected).

pm47 (dismissed):

pm47 (approved):

PR #2367, opened 2022-08-08
Fix blinded path amount to forward calculation

t-bast:
Apply @rustyrussell's neat truncating integer arithmetic formula to calculate the amount that should be forwarded by blinded path nodes instead of our previous approximation.

This simply expands `ceil(a/b)` as `(a+b-1)/b` and provides better results in practice (I ran a simulation with many values for fees and amount).


thomash-acinq (approved):
Do you mean that you get the same results with better performance or that you get different numbers? If we get better performance by not going into float territory that's good but the number we get should be the same.

t-bast:
> Do you mean that you get the same results with better performance or that you get different numbers?

I haven't tested performance, only correctness. Going into float territory we had rounding issues that we don't have with this formula when running tests against random values for fees / recipient amount.

PR #2368, opened 2022-08-09
Add explicit duration to `ExcludeChannel`

### LINKED ISSUE #2364: Feature request: Create new function to ban channels for a defined time
Currently (if enabled) channels get banned for a fixed duration if they fail a payment.
If you use a plugin to monitor all payment you might find some bad channels that fail all the time and want them black listed for longer.
The whole process to black list channels and automatically reinstating them is already there.
It is just not possible to manually call this and define a custom duration for the black list.

Allowing this would give plugins much better control over payment routes.
Also it would remove the need to pass long list of exclude channels to rout find commands.

Currently I have 1500 channels on my black list and have to pass them to each and every find route command.

t-bast:
This lets callers override the default duration or ban channels for unlimited durations (until they send a `ListChannelExclusion`).

Fixes #2364

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -32,8 +32,8 @@ import fr.acinq.eclair.crypto.TransportHandler
 import fr.acinq.eclair.db.NetworkDb
 import fr.acinq.eclair.io.Peer.PeerRoutingMessage
 import fr.acinq.eclair.payment.Invoice.ExtraEdge
-import fr.acinq.eclair.payment.{Bolt11Invoice, Invoice}
 import fr.acinq.eclair.payment.relay.Relayer
+import fr.acinq.eclair.payment.{Bolt11Invoice, Invoice}
 import fr.acinq.eclair.remote.EclairInternalsSerializer.RemoteTypes
 import fr.acinq.eclair.router.Graph.GraphStructure.DirectedGraph
 import fr.acinq.eclair.router.Graph.{HeuristicsConstants, WeightRatios}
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -171,10 +171,9 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
     case Event(TickPruneStaleChannels, d) =>
       stay() using StaleChannels.handlePruneStaleChannels(d, nodeParams.db.network, nodeParams.currentBlockHeight)
 
-    case Event(ExcludeChannel(desc@ChannelDesc(shortChannelId, nodeId, _)), d) =>
-      val banDuration = nodeParams.routerConf.channelExcludeDuration
-      log.info("excluding shortChannelId={} from nodeId={} for duration={}", shortChannelId, nodeId, banDuration)
-      context.system.scheduler.scheduleOnce(banDuration, self, LiftChannelExclusion(desc))
+    case Event(ExcludeChannel(desc, duration_opt), d) =>
+      log.info("excluding shortChannelId={} from nodeId={} for duration={}", desc.shortChannelId, desc.a, duration_opt.getOrElse("n/a"))
+      duration_opt.foreach(banDuration => context.system.scheduler.scheduleOnce(banDuration, self, LiftChannelExclusion(desc)))
```
pm47, line 176: I would make `routerConf.channelExcludeDuration` a `Duration`, which can either be a `FiniteDuration` or `InfiniteDuration`. We can then define the default duration as `infinite` in `eclair.conf` and leave `ExcludeChannel` and `PaymentLifecycle` untouched.

```suggestion
    case Event(ExcludeChannel(desc), d) =>
      val banDuration = nodeParams.routerConf.channelExcludeDuration
      log.info("excluding shortChannelId={} from nodeId={} for duration={}", desc.shortChannelId, desc.a, banDuration)
      banDuration match {
        case duration: FiniteDuration => context.system.scheduler.scheduleOnce(duration, self, LiftChannelExclusion(desc))
        case _: Duration.Infinite => () // permanent ban, can still be manually lifted by sending a LiftChannelExclusion
      }
```

t-bast: > I would make routerConf.channelExcludeDuration a Duration, which can either be a FiniteDuration or InfiniteDuration. We can then define the default duration as infinite in eclair.conf and leave ExcludeChannel and PaymentLifecycle untouched.

But does it really make sense? I don't think the errors in `PaymentLifecycle` should ever trigger an infinite ban, so it would be a footgun to let node operators set this to `infinite` in `eclair.conf`, wouldn't it?

The goal of the issue I linked is to let plugins ban channels for the duration they'd like (infinite or not), independently of what `PaymentLifecycle` may do. It lets plugins experiment with graph pruning heuristics without requiring big changes in `eclair`.

pm47: > The goal of the issue I linked is to let plugins ban channels for the duration they'd like (infinite or not), independently of what PaymentLifecycle may do. It lets plugins experiment with graph pruning heuristics without requiring big changes in eclair.

Alright, then we can:
- keep `routeParams.channelExcludeDuration` as `FiniteDuration` (altough node operators can still set it to 1 year)
- add a `duration_opt: Option[Duration]` to `ExcludeChannel`. The semantics of the `Option` changes from "emulating an infite duration" to "override the default value"

t-bast: > add a duration_opt: Option[Duration] to ExcludeChannel. The semantics of the Option changes from "emulating an infite duration" to "override the default value"

I tried this, but the `InfiniteDuration` is quite awkward to use, we need an explicit check on `isFinite`...it feels more natural to provide `None` for a permanent exclusion than to use `Some(InfiniteDuration)` don't you think? If provided, the channel is banned for that duration, if not provided the channel is banned until lifted manually.

pm47 (approved):

PR #2376, opened 2022-08-10
Remove legacy force-close htlc matching

t-bast:
Before eclair v0.6.0, we didn't store a mapping between `htlc_id` and htlc txs, which made it tricky to correctly remove identical htlcs that used MPP during force-close, when an htlc tx was confirmed.

We have added that mapping since then and released it more than one year ago, so we can now safely remove that code.


pm47 (approved):
:broom: 💨

PR #2378, opened 2022-08-11
Delete `ChaCha20.decrypt`, rename `encrypt` to `xor`.

fiatjaf:
This changes nothing because the two functions do exactly the same thing, I just thought it would be clearer this way.

Protocol description: https://www.rfc-editor.org/rfc/rfc7539#section-2.4

>    Decryption is done in the same way.  The ChaCha20 block function is
>    used to expand the key into a keystream, which is XORed with the
>    ciphertext giving back the plaintext.

About the boolean parameter passed to `engine.init`: https://github.com/bcgit/bc-java/blob/36e75a9a26479420d44008cefd28de2f2db083c3/core/src/main/java/org/bouncycastle/crypto/engines/Salsa20Engine.java#L88-L96 (`ChaCha7539Engine` extends `Salsa20` Engine and also does not use `forEncryption`)

>        /* 
>        * Salsa20 encryption and decryption is completely
>        * symmetrical, so the 'forEncryption' is 
>        * irrelevant. (Like 90% of stream ciphers)
>        */

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/crypto/ChaCha20Poly1305.scala
@@ -50,10 +50,10 @@ object Poly1305 {
  * see https://tools.ietf.org/html/rfc7539#section-2.5
  */
 object ChaCha20 {
-  def encrypt(plaintext: ByteVector, key: ByteVector, nonce: ByteVector, counter: Int = 0): ByteVector = {
+  def xor(input: ByteVector, key: ByteVector, nonce: ByteVector, counter: Int = 0): ByteVector = {
     val engine = new ChaCha7539Engine()
     engine.init(true, new ParametersWithIV(new KeyParameter(key.toArray), nonce.toArray))
-    val ciphertext: Array[Byte] = new Array[Byte](plaintext.length.toInt)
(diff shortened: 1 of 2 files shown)


t-bast:
> I just thought it would be clearer this way.

Why do you think it's clearer?
On the contrary, the fact that encrypt and decrypt are the same thing are an implementation detail of stream ciphers.
From an abstraction's point of view, the functionality we want is really encrypt/decrypt, that's why we chose to expose it that way.

fiatjaf:
I was reading the code while reading about ChaCha20 and looking at other implementations from other languages and everywhere it was just one function, here it was two functions that appeared to be different.

Maybe the decrypt function should just call encrypt then?

Also bolt04 doesn't treat ChaCha20 as a means for encrypting anything in the traditional sense, instead it just "encrypts" a bunch of zeroes and uses that in a different way, so calling it "encrypt" was confusing there too.

Anyway, this is not important for he anymore now that I realized how it works, this PR was more to help future people that might be reading this code, but feel free to close.

t-bast:
It's really a matter of preferences, I always like exposing cipher operations as `encrypt` / `decrypt` methods, even for stream ciphers, as I think this is the abstraction callers are most used (and lets us swap a stream cipher with a block cipher without callers even having to care), but that's not very important, so I'm in favor of not changing anything to reduce churn since you don't mind.

PR #2379, opened 2022-08-11
Remove invalid channel updates from DB at startup

t-bast:
Following #2361, we reject channel updates that don't contain the `htlc_maximum_msat` field. However, the network DB may contain such channel updates, that we need to remove when starting up.

I've checked what we have in our network DB on the ACINQ node before opening that PR. We have ~100 channel updates that don't contain an `htlc_maximum_msat`, but all of them except one are completely outdated (from before 2020), which indicates that we have a pruning bug that I'll fix in a separate PR. The only non-outdated invalid channel update that we have is a 22,222 sat channel between a node called `dontconnect` and a node called `Sybil`: https://1ml.com/channel/588047405915373568

So we can pretty much ignore it.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
@@ -17,8 +17,7 @@
 package fr.acinq.eclair.db.pg
 
 import fr.acinq.bitcoin.scalacompat.{ByteVector32, Crypto, Satoshi}
-import fr.acinq.eclair.ShortChannelId
-import fr.acinq.eclair.RealShortChannelId
+import fr.acinq.eclair.{RealShortChannelId, ShortChannelId}
 import fr.acinq.eclair.db.Monitoring.Metrics.withMetrics
 import fr.acinq.eclair.db.Monitoring.Tags.DbBackends
 import fr.acinq.eclair.db.NetworkDb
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
```diff
@@ -80,7 +79,17 @@ class PgNetworkDb(implicit ds: DataSource) extends NetworkDb with Logging {
...
           }
-        case Some(CURRENT_VERSION) => () // table is up-to-date, nothing to do
+        case Some(CURRENT_VERSION) =>
+          // We clean up channels that contain an invalid channel update (e.g. missing htlc_maximum_msat).
+          val invalidChannels = statement.executeQuery("SELECT short_channel_id, channel_update_1, channel_update_2 FROM network.public_channels").map(rs => {
+            val shortChannelId = rs.getLong("short_channel_id")
+            val validChannelUpdate1 = rs.getBitVectorOpt("channel_update_1").forall(channelUpdateCodec.decode(_).isSuccessful)
+            val validChannelUpdate2 = rs.getBitVectorOpt("channel_update_2").forall(channelUpdateCodec.decode(_).isSuccessful)
+            (shortChannelId, validChannelUpdate1 && validChannelUpdate2)
+          }).filter { case (_, isValid) => !isValid }.map { case (scid, _) => scid }.toSeq
```
sstone, line 89:
```suggestion
          }).collect { case (scid, false) => scid }.toSeq
```

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteNetworkDb.scala
```diff
@@ -61,7 +60,17 @@ class SqliteNetworkDb(val sqlite: Connection) extends NetworkDb with Logging {
...
         migration12(statement)
-      case Some(CURRENT_VERSION) => () // table is up-to-date, nothing to do
+      case Some(CURRENT_VERSION) =>
+        // We clean up channels that contain an invalid channel update (e.g. missing htlc_maximum_msat).
+        val invalidChannels = statement.executeQuery("SELECT short_channel_id, channel_update_1, channel_update_2 FROM channels").map(rs => {
+          val shortChannelId = rs.getLong("short_channel_id")
+          val validChannelUpdate1 = rs.getBitVectorOpt("channel_update_1").forall(channelUpdateCodec.decode(_).isSuccessful)
+          val validChannelUpdate2 = rs.getBitVectorOpt("channel_update_2").forall(channelUpdateCodec.decode(_).isSuccessful)
+          (shortChannelId, validChannelUpdate1 && validChannelUpdate2)
+        }).filter { case (_, isValid) => !isValid }.map { case (scid, _) => scid }.toSeq
```
sstone, line 70:
```suggestion
        }).collect { case (scid, false) => scid }.toSeq
```

sstone (commented):
LGTM, just a few nits

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
t-bast: Good catch, I don't even need the intermediate variable, improved in https://github.com/ACINQ/eclair/pull/2379/commits/55b919f8ae079a559e7200c39b74f26bbc896304

eclair-core/src/main/scala/fr/acinq/eclair/db/sqlite/SqliteNetworkDb.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2379/commits/55b919f8ae079a559e7200c39b74f26bbc896304

sstone (approved):

PR #2384, opened 2022-08-17
Limit default `from` and `to` API parameters

### LINKED ISSUE #2383: An API call to “listinvoices” locks up the node if you have a few hundred thousand invoices
Tested version:
Eclair v0.7.0

Description:
if you call “listinvoices” without “from” or “to” parameter it will try to return all invoices.
This will lead to a timeout on the API if you have few hundred thousand invoices. But the calculation still goes on in the node and will use 100% CPU (even on multi core systems). If a multiple such requests are made the node becomes completely unresponsive and needs to be killed.

This problem happens in combination with RTL, as they all this function.

Possible solution: 
Add a limit parameter which defaults to 100, so the list returned is always manageable

t-bast:
When not provided, we previously used unrestricted `from` and `to` parameters when reading from the DB. This can create issues when accidentally reading too much data.

We now limit this to the last 24 hours, unless explicitly set by the caller. Note that when reading a lot of data, it's recommended to use postgres and read from a replica to ensure there's no impact on the running eclair instance.

Thanks @DerEwige for raising this!

Fixes #2383

### DIFF
--- a/eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
+++ b/eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
@@ -1096,8 +1096,10 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
       check {
         assert(handled)
         assert(status == OK)
-        val now = TimestampSecond.now()
-        eclair.audit(argThat[TimestampSecond](t => now - 25.hours <= t && t <= now), argThat[TimestampSecond](t => t <= now))(any[Timeout]).wasCalled(once)
+        // The default is to get data for the last day.
+        val from = TimestampSecond.now() - 1.day
+        val to = TimestampSecond.now()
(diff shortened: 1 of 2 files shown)


eclair-node/src/test/scala/fr/acinq/eclair/api/ApiServiceSpec.scala
```diff
@@ -1096,7 +1096,8 @@ class ApiServiceSpec extends AnyFunSuite with ScalatestRouteTest with IdiomaticM
       check {
         assert(handled)
         assert(status == OK)
-        eclair.audit(TimestampSecond.min, TimestampSecond.max)(any[Timeout]).wasCalled(once)
+        val now = TimestampSecond.now()
+        eclair.audit(argThat[TimestampSecond](t => now - 25.hours <= t && t <= now), argThat[TimestampSecond](t => t <= now))(any[Timeout]).wasCalled(once)
```
pm47, line 1100: Nit: why use `hours`?
```suggestion
        eclair.audit(argThat[TimestampSecond](t => now - 1.day <= t && t <= now), argThat[TimestampSecond](t => t <= now))(any[Timeout]).wasCalled(once)
```

pm47 (approved):

pm47 (approved):

PR #2386, opened 2022-08-18
Updated offers spec

thomash-acinq:
https://github.com/lightning/bolts/pull/798
- copy offer into invoice request and invoice request into invoice
- change signature calculation


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferCodecs.scala
```diff
@@ -19,135 +19,159 @@ package fr.acinq.eclair.wire.protocol
...
-  private val fallbackAddress: Codec[FallbackAddress] = variableSizeBytesLong(varintoverflow, ("version" | byte) :: ("address" | variableSizeBytes(uint16, bytes))).as[FallbackAddress]
+  private val fallbackAddress: Codec[FallbackAddress] = (("version" | byte) :: ("address" | variableSizeBytes(uint16, bytes))).as[FallbackAddress]
 
-  private val fallbacks: Codec[Fallbacks] = tlvField(list(fallbackAddress).xmap[Seq[FallbackAddress]](_.toSeq, _.toList).as[Fallbacks])
+  private val invoiceFallbacks: Codec[InvoiceFallbacks] = tlvField(list(fallbackAddress).xmap[Seq[FallbackAddress]](_.toSeq, _.toList))
 
-  private val refundSignature: Codec[RefundSignature] = tlvField(bytes64.as[RefundSignature])
+  private val invoiceFeatures: Codec[InvoiceFeatures] = tlvField(featuresCodec)
+
+  private val invoiceNodeId: Codec[InvoiceNodeId] = tlvField(publicKey)
```
t-bast, line 143: This should be replaced by `invoiceCode`

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -39,57 +39,34 @@ case class Bolt12Invoice(records: TlvStream[InvoiceTlv]) extends Invoice {
...
+  val fallbacks: Option[Seq[FallbackAddress]] = records.get[InvoiceFallbacks].map(_.addresses)
   val signature: ByteVector64 = records.get[Signature].get.signature
 
   // It is assumed that the request is valid for this offer.
-  def isValidFor(offer: Offer, request: InvoiceRequest): Boolean = {
-    nodeId == offer.nodeId &&
-      checkSignature() &&
-      offerId.contains(request.offerId) &&
-      request.chain == chain &&
+  def isValidFor(request: InvoiceRequest): Boolean = {
```
t-bast, line 59: Is this function still useful now that the invoice request is embedded?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -39,57 +39,34 @@ case class Bolt12Invoice(records: TlvStream[InvoiceTlv]) extends Invoice {
...
-  val payerInfo: Option[ByteVector] = records.get[PayerInfo].map(_.info)
-  val fallbacks: Option[Seq[FallbackAddress]] = records.get[Fallbacks].map(_.addresses)
-  val refundSignature: Option[ByteVector64] = records.get[RefundSignature].map(_.signature)
-  val replaceInvoice: Option[ByteVector32] = records.get[ReplaceInvoice].map(_.paymentHash)
+  override val nodeId: Crypto.PublicKey = records.get[InvoiceNodeId].get.nodeId
+  override val paymentHash: ByteVector32 = records.get[InvoicePaymentHash].get.hash
+  override val description: Either[String, ByteVector32] = Left(invoiceRequest.offer.description)
+  override val createdAt: TimestampSecond = records.get[InvoiceCreatedAt].get.timestamp
+  override val relativeExpiry: FiniteDuration = FiniteDuration(records.get[InvoiceRelativeExpiry].map(_.seconds).getOrElse(DEFAULT_EXPIRY_SECONDS), TimeUnit.SECONDS)
+  override val minFinalCltvExpiryDelta: CltvExpiryDelta = CltvExpiryDelta(0)
```
t-bast, line 51: That indicates that something should be changed, I don't think a value of `0` makes sense here?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -39,57 +39,34 @@ case class Bolt12Invoice(records: TlvStream[InvoiceTlv]) extends Invoice {
...
+  override val features: Features[InvoiceFeature] = records.get[InvoiceFeatures].map(_.features.invoiceFeatures()).getOrElse(Features.empty)
+  val blindedPaths: Seq[RouteBlinding.BlindedRoute] = records.get[InvoicePaths].get.paths
+  val blindedPathsInfo: Seq[PaymentInfo] = records.get[InvoiceBlindedPay].get.paymentInfo
```
t-bast, line 54: These two should probably be zipped together and only exposed as a `Seq[PaymentBlindedRoute]`, don't you think?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -103,65 +80,54 @@ case class PaymentBlindedRoute(route: Sphinx.RouteBlinding.BlindedRoute, payment
...
    * @param features invoice features
    * @param paths    the blinded paths to use to pay the invoice
    */
-  def apply(offer: Offer,
-            request: InvoiceRequest,
+  def apply(request: InvoiceRequest,
             preimage: ByteVector32,
             nodeKey: PrivateKey,
-            minFinalCltvExpiryDelta: CltvExpiryDelta,
             features: Features[InvoiceFeature],
```
t-bast, line 98:
```suggestion
            features: Features[Bolt12Feature],
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -103,65 +80,54 @@ case class PaymentBlindedRoute(route: Sphinx.RouteBlinding.BlindedRoute, payment
...
+      Some(InvoiceBlindedPay(paths.map(_.paymentInfo))),
+      Some(InvoiceCreatedAt(TimestampSecond.now())),
+      Some(InvoiceRelativeExpiry(300)),
+      Some(InvoicePaymentHash(Crypto.sha256(preimage))),
+      Some(InvoiceAmount(amount)),
+      if (!features.isEmpty) Some(InvoiceFeatures(features.unscoped())) else None,
+      Some(InvoiceNodeId(nodeKey.publicKey)),
     ).flatten
-    val signature = signSchnorr(signatureTag("signature"), rootHash(TlvStream(tlvs), OfferCodecs.invoiceTlvCodec), nodeKey)
+    val signature = signSchnorr(signatureTag, rootHash(TlvStream(tlvs), OfferCodecs.invoiceTlvCodec), nodeKey)
```
t-bast, line 112: Same issue here, we must also sign unknown fields of the invoice request (and add a unit test for it).

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -103,65 +80,54 @@ case class PaymentBlindedRoute(route: Sphinx.RouteBlinding.BlindedRoute, payment
...
-      Some(Cltv(minFinalCltvExpiryDelta)),
-      request.payerInfo.map(PayerInfo),
-      request.replaceInvoice.map(ReplaceInvoice),
+    require(request.amount.nonEmpty || request.offer.amount.nonEmpty)
+    val amount = request.amount.orElse(request.offer.amount.map(_ * request.quantity)).get
+    val tlvs: Seq[InvoiceTlv] = removeSignature(request.records).records.toSeq ++ Seq(
+      Some(InvoicePaths(paths.map(_.route))),
+      Some(InvoiceBlindedPay(paths.map(_.paymentInfo))),
+      Some(InvoiceCreatedAt(TimestampSecond.now())),
+      Some(InvoiceRelativeExpiry(300)),
```
t-bast, line 106: We shouldn't have a hard-coded value here, that should be provided in the constructor (we have a configurable value for that in `NodeParams`: `invoiceExpiry`).

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -103,65 +80,54 @@ case class PaymentBlindedRoute(route: Sphinx.RouteBlinding.BlindedRoute, payment
...
 
   def validate(records: TlvStream[InvoiceTlv]): Either[InvalidTlvPayload, Bolt12Invoice] = {
-    if (records.get[Amount].isEmpty) return Left(MissingRequiredTlv(UInt64(8)))
-    if (records.get[Description].isEmpty) return Left(MissingRequiredTlv(UInt64(10)))
-    if (records.get[Paths].isEmpty) return Left(MissingRequiredTlv(UInt64(16)))
-    if (records.get[PaymentPathsInfo].map(_.paymentInfo.length) != records.get[Paths].map(_.paths.length)) return Left(MissingRequiredTlv(UInt64(18)))
-    if (records.get[NodeId].isEmpty) return Left(MissingRequiredTlv(UInt64(30)))
-    if (records.get[CreatedAt].isEmpty) return Left(MissingRequiredTlv(UInt64(40)))
-    if (records.get[PaymentHash].isEmpty) return Left(MissingRequiredTlv(UInt64(42)))
+    InvoiceRequest.validate(TlvStream[InvoiceRequestTlv](records.records.collect { case tlv: InvoiceRequestTlv => tlv }, records.unknown.filter(_.tag < UInt64(160)))).fold(
```
t-bast, line 117: This filtering is repeated in a few places, maybe it would make sense to have helper functions in `OfferTypes.scala` to do that?

```scala
def filterOfferFields(tlvs: TlvStream[InvoiceTlv]): TlvStream[OfferTlv] = ...
def filterInvoiceRequestFields(tlvs: TlvStream[InvoiceTlv]): TlvStream[InvoiceRequestTlv] = ...
```

This way we'll also be able to add a comment there to explain the spec restriction on tlv fields (offers can only use < 80, invoice requests 80 <= ... < 160).

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -103,65 +80,54 @@ case class PaymentBlindedRoute(route: Sphinx.RouteBlinding.BlindedRoute, payment
...
+      invalidTlvPayload => return Left(invalidTlvPayload),
+      _ -> ()
+    )
+    if (records.get[InvoiceAmount].isEmpty) return Left(MissingRequiredTlv(UInt64(170)))
+    if (records.get[InvoicePaths].isEmpty) return Left(MissingRequiredTlv(UInt64(160)))
+    if (records.get[InvoiceBlindedPay].map(_.paymentInfo.length) != records.get[InvoicePaths].map(_.paths.length)) return Left(MissingRequiredTlv(UInt64(162)))
+    if (records.get[InvoiceNodeId].isEmpty) return Left(MissingRequiredTlv(UInt64(176)))
```
t-bast, line 124: This isn't required anymore.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
```diff
@@ -39,57 +39,34 @@ case class Bolt12Invoice(records: TlvStream[InvoiceTlv]) extends Invoice {
...
-    (refundSignature, refundFor, payerKey) match {
-      case (Some(sig), Some(hash), Some(key)) => verifySchnorr(signatureTag("payer_signature"), hash, sig, key)
-      case _ => false
-    }
+      request.amount.forall(_ == amount) &&
+      request.features.areSupported(features) &&
+      checkSignature()
   }
 
   def checkSignature(): Boolean = {
```
t-bast, line 68: Does it really make sense to have a `checkSignature` method (and same comment for `InvoiceRequest`)? Shouldn't we just verify it in the `validate` functions and then stop caring about it? Otherwise we just risk forgetting to verify it.

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/PaymentOnionSpec.scala
```diff
@@ -211,7 +211,6 @@ class PaymentOnionSpec extends AnyFunSuite {
       val Right(payload) = FinalPayload.Blinded.validate(decoded, blindedTlvs)
       assert(payload.amount == 561.msat)
       assert(payload.totalAmount == 1105.msat)
-      assert(payload.expiry == CltvExpiry(42))
```
t-bast, line 214: Shouldn't that be:

```scala
assert(payload.expiry == CltvExpiry(1234567))
```

Instead of removing the check?

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt12InvoiceSpec.scala
```diff
@@ -419,32 +312,41 @@ class Bolt12InvoiceSpec extends AnyFunSuite {
...
     assert(decodedInvoice.paymentHash == invoice.paymentHash)
     assert(decodedInvoice.description == invoice.description)
-    assert(decodedInvoice.payerKey == invoice.payerKey)
-    assert(decodedInvoice.chain == invoice.chain)
+    assert(decodedInvoice.invoiceRequest.unsigned == invoice.invoiceRequest.unsigned)
+  }
+
+  test("cln invoice"){
+    val encodedInvoice = "lni1qqgds4gweqxey37gexf5jus4kcrwuq3qqc3xu3s3rg94nj40zfsy866mhu5vxne6tcej5878k2mneuvgjy8s5predakx793pqfxv2rtqfajhp98c5tlsxxkkmzy0ntpzp2rtt9yum2495hqrq4wkj5pqqc3xu3s3rg94nj40zfsy866mhu5vxne6tcej5878k2mneuvgjy84yqucj6q9sggrnl24r93kfmdnatwpy72mxg7ygr9waxu0830kkpqx84pd5j65fhg2pxqzfnzs6cz0v4cff79zlup344kc3ru6cgs2s66ef8x64fd9cqc9t45s954fef6n3ql8urpc4r2vvunc0uv9yq37g485heph6lpuw34ywxadqypwq3hlcrpyk32zdvlrgfsdnx5jegumenll49v502862l9sq5erz3qqxte8tyk308ykd6fqy2lxkrsmeq77d8s5977pzmc68lgvs2xcn0kfvnlzud9fvkv900ggwe7yf9hf7lr6qz3pcqqqqqqqqqqqqqqq5qqqqqqqqqqqqqwjfvkl43fqqqqqqzjqgcuhrdv2sgq5spd8qp4ev2rw0v9r7cvvrntlzpvlwmd8vczycklu87336h55g24q8xykszczzqjvc5xkqnm9wz203ghlqvdddkyglxkzyz5xkk2fek42tfwqxp2ad8cypv26x5zxkyk675ep3v48grwydze6nvvg56cklgmvztuny58t5j0fl3hemx3lvd0ryx89jtf0h069z6r2qwqvjlyrewvzsfqmmfajs70q"
```
t-bast, line 344: It's probably a bit early to have tests against invoices generated by another implementation since there may still be some changes (for example the addition of `route_expiry` to `blinded_payinfo`).

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/OfferTypesSpec.scala
```diff
@@ -157,80 +115,60 @@ class OfferTypesSpec extends AnyFunSuite {
...
-    assert(request.chain == Block.LivenetGenesisBlock.hash)
-    assert(request.payerKey == ByteVector32(hex"c52f7240b14fd2baefda0d14ae9142de41891e5a5f7e34506bde9f4a60182750"))
-    assert(request.payerInfo == Some(hex"deadbeef"))
-    assert(request.payerNote == Some("I am Batman"))
-    assert(request.encode() == encoded)
-  }
-
   test("decode invalid invoice request") {
     val testCases = Seq(
       // Missing offer id.
```
t-bast, line 174: This test doesn't make sense anymore since offer id doesn't exist? Generally it looks like this test suite requires a careful rework to remove obsolete stuff and make sure we're correctly testing against the latest specification.

t-bast (commented):
The type hierarchy looks good, but you missed several changes in the latest version of the specification and the test suites need a bit of clean-up.

We also don't support yet the scenario where an `invoice_request` is generated without a matching `offer`: I think that's fine and should be added later, you're right to focus on the `offer` scenario first since it's the one we care the most about.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
thomash-acinq: It is checked in the `isValidFor`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
thomash-acinq: Actually it is in the latest latest version: https://github.com/rustyrussell/lightning-rfc/pull/13
I know it's confusing that rusty didn't put it in the main PR.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferCodecs.scala
thomash-acinq: No, it's been removed in https://github.com/rustyrussell/lightning-rfc/pull/13

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -38,86 +39,80 @@ object OfferTypes {
...
+  case class FallbackAddress(version: Byte, value: ByteVector)
 
-  case class Cltv(minFinalCltvExpiry: CltvExpiryDelta) extends InvoiceTlv
+  case class InvoiceFallbacks(addresses: Seq[FallbackAddress]) extends InvoiceTlv
 
-  case class FallbackAddress(version: Byte, value: ByteVector)
+  case class InvoiceFeatures(features: Features[Feature]) extends InvoiceTlv
 
-  case class Fallbacks(addresses: Seq[FallbackAddress]) extends InvoiceTlv
+  case class InvoiceNodeId(nodeId: PublicKey) extends InvoiceTlv
```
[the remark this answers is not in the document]
thomash-acinq: See https://github.com/rustyrussell/lightning-rfc/pull/13

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -209,37 +190,36 @@ object OfferTypes {
...
-    val payerNote: Option[String] = records.get[PayerNote].map(_.note)
-    val payerInfo: Option[ByteVector] = records.get[PayerInfo].map(_.info)
-    val replaceInvoice: Option[ByteVector32] = records.get[ReplaceInvoice].map(_.paymentHash)
-    val payerSignature: ByteVector64 = records.get[Signature].get.signature
+    val payerId: PublicKey = records.get[InvoiceRequestPayerId].get.publicKey
+    val payerNote: Option[String] = records.get[InvoiceRequestPayerNote].map(_.note)
+    private val signature: ByteVector64 = records.get[Signature].get.signature
 
-    def isValidFor(offer: Offer): Boolean = {
+    def isValidFor(otherOffer: Offer): Boolean = {
```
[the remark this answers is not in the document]
thomash-acinq: Same as for the similar function in `Bolt12Invoice`, when we receive an invoice request, we need to check that it is valid for our offer.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -266,23 +248,26 @@ object OfferTypes {
...
-        Some(OfferId(offer.offerId)),
-        Some(Amount(amount)),
-        if (offer.quantityMin.nonEmpty || offer.quantityMax.nonEmpty) Some(Quantity(quantity)) else None,
-        Some(PayerKey(payerKey.publicKey)),
-        if (!features.isEmpty) Some(FeaturesTlv(features.unscoped())) else None,
-      ).flatten
-      val signature = signSchnorr(InvoiceRequest.signatureTag, rootHash(TlvStream(tlvs), OfferCodecs.invoiceRequestTlvCodec), payerKey)
-      InvoiceRequest(TlvStream(tlvs :+ Signature(signature)))
+      require(quantity == 1 || offer.quantityMax.nonEmpty)
+      val tlvs: Seq[InvoiceRequestTlv] = InvoiceRequestMetadata(randomBytes32()) +: (offer.records.records.toSeq ++ Seq(
```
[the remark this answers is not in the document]
thomash-acinq: That's because I want to order the TLVs by tags and `InvoiceRequestMetadata` has tag 0 so it should go in front whereas the other invoice request TLVs are all higher than the offer TLVs.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -209,37 +190,36 @@ object OfferTypes {
...
 
   case class InvoiceRequest(records: TlvStream[InvoiceRequestTlv]) {
-    val chain: ByteVector32 = records.get[Chain].map(_.hash).getOrElse(Block.LivenetGenesisBlock.hash)
-    val offerId: ByteVector32 = records.get[OfferId].map(_.offerId).get
-    val amount: Option[MilliSatoshi] = records.get[Amount].map(_.amount)
-    val features: Features[InvoiceFeature] = records.get[FeaturesTlv].map(_.features.invoiceFeatures()).getOrElse(Features.empty)
-    val quantity_opt: Option[Long] = records.get[Quantity].map(_.quantity)
+    val offer: Offer = Offer.validate(TlvStream[OfferTlv](records.records.collect { case tlv: OfferTlv => tlv }, records.unknown.filter(_.tag < UInt64(80)))).toOption.get
+
+    val metadata: Option[ByteVector] = records.get[InvoiceRequestMetadata].map(_.data)
```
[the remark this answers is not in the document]
thomash-acinq: Yes it is in the new version of the spec, even though it's not used anywhere so we could still accept requests that don't have it.
Anyways, I now check for it being present.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -266,23 +248,26 @@ object OfferTypes {
...
-      InvoiceRequest(TlvStream(tlvs :+ Signature(signature)))
+      require(quantity == 1 || offer.quantityMax.nonEmpty)
+      val tlvs: Seq[InvoiceRequestTlv] = InvoiceRequestMetadata(randomBytes32()) +: (offer.records.records.toSeq ++ Seq(
+        Some(InvoiceRequestChain(chain)),
+        Some(InvoiceRequestAmount(amount)),
+        if (offer.quantityMax.nonEmpty) Some(InvoiceRequestQuantity(quantity)) else None,
+        if (!features.isEmpty) Some(InvoiceRequestFeatures(features.unscoped())) else None,
+        Some(InvoiceRequestPayerId(payerKey.publicKey)),
+      ).flatten)
+      val signature = signSchnorr(signatureTag, rootHash(TlvStream(tlvs), OfferCodecs.invoiceRequestTlvCodec), payerKey)
```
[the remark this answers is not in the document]
thomash-acinq: Fixed

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -266,23 +248,26 @@ object OfferTypes {
...
     }
 
     def validate(records: TlvStream[InvoiceRequestTlv]): Either[InvalidTlvPayload, InvoiceRequest] = {
-      if (records.get[OfferId].isEmpty) return Left(MissingRequiredTlv(UInt64(4)))
-      if (records.get[PayerKey].isEmpty) return Left(MissingRequiredTlv(UInt64(38)))
+      Offer.validate(TlvStream[OfferTlv](records.records.collect { case tlv: OfferTlv => tlv }, records.unknown.filter(_.tag < UInt64(80)))).fold(
+        invalidTlvPayload => return Left(invalidTlvPayload),
+        _ -> ()
+      )
+      if (records.get[InvoiceRequestPayerId].isEmpty) return Left(MissingRequiredTlv(UInt64(88)))
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -346,31 +333,22 @@ object OfferTypes {
...
     val h = hash(tag, msg)
     // NB: we don't add auxiliary random data to keep signatures deterministic.
     ByteVector64(ByteVector(Secp256k1JvmKt.getSecpk256k1.signSchnorr(h.toArray, key.value.toArray, null)))
   }
 
-  def verifySchnorr(tag: String, msg: ByteVector32, signature: ByteVector64, publicKey: ByteVector32): Boolean = {
+  def verifySchnorr(tag: ByteVector, msg: ByteVector32, signature: ByteVector64, publicKey: PublicKey): Boolean = {
     val h = hash(tag, msg)
-    Secp256k1JvmKt.getSecpk256k1.verifySchnorr(signature.toArray, h.toArray, publicKey.toArray)
+    Secp256k1JvmKt.getSecpk256k1.verifySchnorr(signature.toArray, h.toArray, publicKey.value.drop(1).toArray)
```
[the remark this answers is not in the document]
thomash-acinq: Done.
There is currently nothing Schnorr-related in bitcoin-lib.

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt12InvoiceSpec.scala
thomash-acinq: We can change it later if there are changes to the spec. But I think it's important to test against the only other implementation that can generate invoices.

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/PaymentOnionSpec.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/OfferTypesSpec.scala
thomash-acinq: Removed obsolete tests.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
t-bast: It is a bit misleading though, we're not checking that the invoice is valid for a given invoice request (which implies it could be valid for various invoice requests), we want it to exactly match the invoice request we sent.

Wouldn't it make more sense to take an `invoiceRequest: InvoiceRequest` argument in `Bolt12Invoice.validate` and verify that it matches? It's quite confusing to have `Bolt12Invoice.validate` and `isValidFor`, especially since both must be called before actually using the invoice?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
t-bast: What I mean is that if you really don't want to provide that value, it shouldn't be in the shared trait, this code shows that there's an issue in the abstraction here.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
t-bast: Linked to other comments, I believe all of those should be done in `validate`, it's confusing and error-prone to scatter validation across different methods that must be called separately, isn't it?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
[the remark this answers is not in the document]
t-bast: Then we really should add it to `bitcoin-lib`, can you work on such a PR?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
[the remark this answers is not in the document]
t-bast: Is it important to order them by tag? We don't really care, tags are a serialization thing, we're at a higher layer here aren't we? Is it to help reviewers verify that all fields are provided?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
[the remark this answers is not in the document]
t-bast: But it is backwards, isn't it? When we receive an invoice request, we will look at the enclosed offer and check the local DB to see if we have a matching offer that hasn't expired (we don't have "our offer" somehow pre-loaded)? So that check will be in the method that scans the DB, not a separate method we call afterwards?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
t-bast: Either we have a meaningful value to provide here, or we should remove it from the shared abstraction because it's actually not something that makes sense for both Bolt11 and Bolt12 invoices.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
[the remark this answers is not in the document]
thomash-acinq: Not necessarily, if we use blinded paths, we will receive an invoice request to a specific path_id and retrieve the offer matching the path_id.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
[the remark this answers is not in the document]
thomash-acinq: I think it makes it easier to follow and it allows us to to test for equality in tests.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
thomash-acinq: For me these two steps are quite distinct:
- `validate` checks that the TLVS is well-formed and returns the corresponding case class, it doesn't check any business logic, just that it parses correctly. Actually that's where I think the name is misleading, I would call it something like `create`, `parse` or `decode`.
- `isValidFor` checks that the other party is not trying to cheat.

For instance `validate` is currently used in `Invoice.fromString` and I don't think it would make sense to require an invoice request just to parse a string, especially since this method is shared with `Bolt11Invoice`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -185,7 +194,7 @@ object Features {
     val mandatory = 6
   }
 
-  case object VariableLengthOnion extends Feature with InitFeature with NodeFeature with InvoiceFeature {
+  case object VariableLengthOnion extends Feature with InitFeature with NodeFeature with Bolt11Feature with Bolt12Feature {
```
t-bast, line 197: Actually the specification only seems to allow `basic_mpp` in the Bolt 12 features, `var_onion_optin` and `option_route_blinding` seem to be mandatory but implicit.

I reworked this slightly in https://github.com/ACINQ/eclair/commit/9bf9941b9b3bb3d073baef5d65bd19d063010c4e, I think this way it makes more sense and is easier to work with, what do you think?

eclair-core/src/main/scala/fr/acinq/eclair/payment/Bolt12Invoice.scala
t-bast: That makes sense, thanks for the explanation!
I agree with you, we should rename the current `validate` functions of the Bolt 12 types.
I don't think `parse` and `decode` are a good match, the contents have already been parsed/decoded from binary.
These functions verify that the tlv stream contains all mandatory content, I guess `validateTlvs` is a better match?
This way it's a bit more obvious that we only validated tlvs, and further validation is needed afterwards to verify that the other party isn't cheating.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -361,9 +361,9 @@ object MultiPartHandler {
                     })
                   }
                 })).map(paths => {
-                  val invoiceFeatures = featuresTrampolineOpt.remove(Features.RouteBlinding).add(Features.RouteBlinding, FeatureSupport.Mandatory)
-                  val invoice = Bolt12Invoice(r.offer, r.invoiceRequest, paymentPreimage, r.nodeKey, nodeParams.channelConf.minFinalExpiryDelta, invoiceFeatures, paths.map { case (blindedRoute, paymentInfo, _) => PaymentBlindedRoute(blindedRoute.route, paymentInfo) })
-                  log.debug("generated invoice={} for offerId={}", invoice.toString, r.offer.offerId)
+                  val invoiceFeatures = nodeParams.features.bolt12Features().remove(Features.RouteBlinding).add(Features.RouteBlinding, FeatureSupport.Mandatory)
```
t-bast, line 364: If you apply https://github.com/ACINQ/eclair/commit/9bf9941b9b3bb3d073baef5d65bd19d063010c4e, you'll be able to remove that hack

```suggestion
                  val invoiceFeatures = nodeParams.features.bolt12Features()
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -261,8 +261,14 @@ object PaymentInitiator {
     def invoice: Invoice
     def recipientNodeId: PublicKey = invoice.nodeId
     def paymentHash: ByteVector32 = invoice.paymentHash
-    def finalExpiry(nodeParams: NodeParams): CltvExpiry = nodeParams.paymentFinalExpiry.computeFinalExpiry(nodeParams.currentBlockHeight, invoice.minFinalCltvExpiryDelta)
     // @formatter:on
+    def finalExpiry(nodeParams: NodeParams): CltvExpiry = {
+      val minFinalCltvExpiryDelta = invoice match {
+        case invoice: Bolt11Invoice => invoice.minFinalCltvExpiryDelta
+        case _: Bolt12Invoice => CltvExpiryDelta(0)
```
t-bast, line 268: I'm still concerned about this, something isn't right here. If we use an expiry delta of 0 here, the recipient will receive an HTLC that expires at the current block height. This creates a non-negligible risk that the next-to-last node will force-close because the HTLC is expired before the last node answers with fulfill or fail. We introduced a non-0 final cltv expiry delta for Bolt 11 explicitly to avoid that issue, we don't want to reintroduce it for Bolt 12?

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -38,27 +39,112 @@ object OfferTypes {
...
+   */
+  case class OfferNodeId(publicKey: PublicKey) extends OfferTlv
+
+  /**
+   * Random data to provide enough entropy so that some fields of the invoice request / invoice can be revealed without revealing the others.
+   */
+  case class InvoiceRequestMetadata(data: ByteVector) extends InvoiceRequestTlv
+
+  /**
+   * If `OfferChains` is present, this specifies which chain is going to be use to pay.
```
t-bast, line 115: nit:
```suggestion
   * If `OfferChains` is present, this specifies which chain is going to be used to pay.
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -38,27 +39,112 @@ object OfferTypes {
...
+   */
+  case class InvoiceRequestFeatures(features: Features[Feature]) extends InvoiceRequestTlv
+
+  /**
+   * Number of items to purchase. Only use if the offer supports purchasing multiple items at once.
+   */
+  case class InvoiceRequestQuantity(quantity: Long) extends InvoiceRequestTlv
+
+  /**
+   * A public key for which the sender know the corresponding private key.
```
t-bast, line 135: It would be useful to explain how this is used to complement the preimage proof of payment.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
thomash-acinq: No, the recipient won't receive an HTLC that expires at the current block height because there is a `minFinalCltvExpiryDelta` included in the blinded route, there is no problem here.
There was a problem in the invoice generation code where I didn't include the `minFinalCltvExpiryDelta` (it must have gotten lost in some rebase) but I've now fixed it.

t-bast: Great, that's a good solution, that works for me! It does need a unit test to verify that we correctly add this expiry delta when generating routes (which ensures that we don't regress here in a future change or rebase).

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
thomash-acinq: The  `MultiPartHandlerSpec` tests check that.

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt12InvoiceSpec.scala
```diff
@@ -58,219 +58,123 @@ class Bolt12InvoiceSpec extends AnyFunSuite {
...
+    val withModifiedAmount = Bolt12Invoice(TlvStream(invoice.records.records.map { case OfferAmount(amount) => OfferAmount(amount + 100.msat) case x => x }, invoice.records.unknown))
+    assert(!withModifiedAmount.isValidFor(request))
+  }
+
+  test("check invoice signature with unknown field from invoice request") {
+    val (nodeKey, payerKey, chain) = (randomKey(), randomKey(), randomBytes32())
+    val offer = Offer(Some(10000 msat), "test offer", nodeKey.publicKey, Features.empty, chain)
+    val basicRequest = InvoiceRequest(offer, 11000 msat, 1, Features.empty, payerKey, chain)
+    val requestWithUnknownTlv = basicRequest.copy(records = TlvStream(basicRequest.records.records, Seq(GenericTlv(UInt64(87), hex"0404"))))
+    val invoice = Bolt12Invoice(requestWithUnknownTlv, randomBytes32(), nodeKey, 300 seconds, Features.empty, Seq(createPaymentBlindedRoute(nodeKey.publicKey)))
```
t-bast, line 79: You should verify that the invoice contains the unknown tlv, it makes the test clearer.

eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt12InvoiceSpec.scala
```diff
@@ -58,219 +58,123 @@ class Bolt12InvoiceSpec extends AnyFunSuite {
     val (nodeKey, payerKey, chain) = (randomKey(), randomKey(), randomBytes32())
     val offer = Offer(Some(10000 msat), "test offer", nodeKey.publicKey, Features.empty, chain)
     val request = InvoiceRequest(offer, 11000 msat, 1, Features.empty, payerKey, chain)
-    val invoice = Bolt12Invoice(offer, request, randomBytes32(), nodeKey, CltvExpiryDelta(20), Features.empty, Seq(createPaymentBlindedRoute(nodeKey.publicKey)))
-    assert(invoice.isValidFor(offer, request))
-    assert(invoice.checkSignature())
-    assert(!invoice.checkRefundSignature())
+    val invoice = Bolt12Invoice(request, randomBytes32(), nodeKey, 300 seconds, Features.empty, Seq(createPaymentBlindedRoute(nodeKey.publicKey)))
+    assert(invoice.isValidFor(request))
```
t-bast, line 62: In this test and the next one, we should use `invoice.checkSignature()`, not `invoice.isValidFor()`, the only part that should fail validation is the signature part so we should ignore the rest (or we can `assert` on both, but it's useful to have one explicitly on `checkSignature()`). Otherwise if anything else in `isValidFor` makes that function return `false`, we end up not testing the signature code.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
```diff
@@ -168,7 +169,7 @@ class PaymentPacketSpec extends AnyFunSuite with BeforeAndAfterAll {
     assert(payment.cmd.nextBlindingKey_opt.isEmpty)
 
     val add_b = UpdateAddHtlc(randomBytes32(), 0, payment.cmd.amount, payment.cmd.paymentHash, payment.cmd.cltvExpiry, payment.cmd.onion, payment.cmd.nextBlindingKey_opt)
-    val Right(relay_b@ChannelRelayPacket(_, payload_b, packet_c)) = decrypt(add_b, priv_b.privateKey, Features.empty)
+    val Right(relay_b@ChannelRelayPacket(aaa, payload_b, packet_c)) = decrypt(add_b, priv_b.privateKey, Features.empty)
```
t-bast, line 172: Remove?

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
```diff
@@ -177,46 +178,46 @@ class PaymentPacketSpec extends AnyFunSuite with BeforeAndAfterAll {
...
     assert(payload_d.isInstanceOf[IntermediatePayload.ChannelRelay.Blinded])
     val blinding_e = payload_d.asInstanceOf[IntermediatePayload.ChannelRelay.Blinded].nextBlinding
 
-    val add_e = UpdateAddHtlc(randomBytes32(), 2, amount_de, paymentHash, expiry_de, packet_e, Some(blinding_e))
+    val add_e = UpdateAddHtlc(randomBytes32(), 2, relay_d.amountToForward, relay_d.add.paymentHash, relay_d.outgoingCltv, packet_e, Some(blinding_e))
     val Right(FinalPacket(_, payload_e)) = decrypt(add_e, priv_e.privateKey, Features(RouteBlinding -> Optional))
     assert(payload_e.amount == finalAmount)
     assert(payload_e.totalAmount == finalAmount)
-    assert(payload_e.expiry == finalExpiry)
+    assert(add_e.cltvExpiry == finalExpiry)
```
t-bast, line 207: There is an issue here...the onion payload will contain `finalExpiry + Bolt11Invoice.DEFAULT_MIN_CLTV_EXPIRY_DELTA`. This is incompatible with the checks made in `PaymentPacket.validateBlindedFinalPayload` where we reject the payment if `add.cltvExpiry < payload.expiry`, isn't it? This issue also appear in the next tests.

I'm not sure whether it's just an issue with how the test is constructed (not correctly taking into account the blinded path's expiry) or an issue with the code itself, it's worth looking into.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
```diff
@@ -528,26 +528,6 @@ class PaymentPacketSpec extends AnyFunSuite with BeforeAndAfterAll {
     assert(failure.isInstanceOf[InvalidOnionBlinding])
   }
 
-  test("fail to decrypt blinded payment at the final node when expiry is too low") {
```
t-bast, line 531: That test shouldn't be removed, this scenario should still be tested?

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/OfferTypesSpec.scala
```diff
@@ -50,72 +42,30 @@ class OfferTypesSpec extends AnyFunSuite {
...
-    assert(Offer.decode(encoded).get == signedOffer)
-    assert(signedOffer.checkSignature())
-    assert(signedOffer.amount.contains(50 msat))
-    assert(signedOffer.description == "offer with quantity")
-    assert(signedOffer.nodeId == nodeId)
-    assert(signedOffer.issuer.contains("alice@bigshop.com"))
-    assert(signedOffer.quantityMin.contains(1))
-  }
-
-  test("decode invalid offer") {
```
t-bast, line 111: We're now missing tests that decode invalid offers and invoice requests.

t-bast (commented):
The code looks good to me, the tests need a bit more work and we should be good to go.

eclair-core/src/test/scala/fr/acinq/eclair/wire/protocol/OfferTypesSpec.scala
thomash-acinq: I've removed them because I didn't think they bring much value. For instance in this PR, I've changed a lot of the code and all of these invalidity tests still pass without needing to be updated.
I've added some new tests to replace them: I create a minimal offer / invoice request and check that removing any TLV makes it invalid.

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
t-bast: Got it, on the recipient side for Bolt 12 we only want to check that we have a few blocks before the HTLC expires, and that the HTLC expiry is lower or equal to the onion's specified expiry (am I getting this one right?).

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentPacketSpec.scala
t-bast: I think it is still important to test: we don't want to let intermediate nodes lower the expiry too much, as that would let them probe their distance to the recipient. That's why we check `case payload if add.cltvExpiry < payload.expiry`.

What makes the test fail after your change is that the attacker can now lower the HTLC expiry by at most the `min_final_expiry_delta`: to make that test pass again, you need to tweak the expiry by 19 blocks.

t-bast: Thinking about it more, I think you're right, I'm not sure an intermediate node lowering the expiry can actually exploit it. But I can't guarantee that this isn't exploitable, and it still shows that a node deviated from the spec, so I'd rather keep the cltv check we're currently making (we don't have a good reason to remove it, do we?).

thomash-acinq: The problem is that in the standard case we expect `add.cltvExpiry == payload.expiry` so if an intermediary node lowers the expiry, we will notice it. But for the blinded case, we expect `add.cltvExpiry == payload.expiry + minFinalCltvExpiryDelta`, so to be able to detect an intermediary node lowering the expiry, we need to know `minFinalCltvExpiryDelta`. Unfortunately we don't have access to `minFinalCltvExpiryDelta` in this part of the code and I don't think it makes sense to add it.
If we really want to prevent intermediary nodes to lower the expiry, then we should add this check to `MultiPartHandler.validatePaymentCltv`. I've added it in 6afc6e680.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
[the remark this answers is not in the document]
thomash-acinq: https://github.com/ACINQ/bitcoin-lib/pull/73

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -444,10 +444,16 @@ object MultiPartHandler {
     val minExpiry = nodeParams.channelConf.minFinalExpiryDelta.toCltvExpiry(nodeParams.currentBlockHeight)
     if (add.cltvExpiry < minExpiry) {
       log.warning("received payment with expiry too small for amount={} totalAmount={}", add.amountMsat, payload.totalAmount)
-      false
-    } else {
-      true
+      return false
+    }
+    if (payload.isInstanceOf[FinalPayload.Blinded]) {
+      val expectedExpiry = payload.expiry + nodeParams.channelConf.minFinalExpiryDelta
```
t-bast, line 450: This is unreliable: if a node operator increases their `minFinalExpiryDelta` in their `eclair.conf`, they will reject all payments for invoices created before the change. It also wouldn't work if at some point we let the user use a different value than the one from node params in their blinded path creation. We would need to compare this to the final expiry delta used for this specific invoice, not the one from node params, but we don't store this information for Bolt 12 invoices.

Anyway I don't think this check is very important: since we check in `PaymentPacket.scala` that `add.cltvExpiry < payload.expiry`, intermediate nodes can decrease the cltv expiry by at most `min_final_expiry_delta`, which should be much smaller than any real hop's `cltv_expiry_delta`, so shouldn't let intermediate nodes probe their position.

I've removed this in [TODO](https://github.com/ACINQ/eclair/commit/26b427ad7f2bea6e365851eacf32a3e044b43c4a).

I'm noticing just now that this issue already exists a few lines above for Bolt 11 invoices: we should use the invoice's `minFinalExpiryDelta` instead of the one from `nodeParams`. This should be fixed in a separate PR.

thomash-acinq: Given that invoices expire relatively quickly (and Bolt 12 invoices can expire even quicker) and the `minFinalExpiryDelta` should be very rarely updated, I think it's fine.
Actually, if I ever increase my `minFinalExpiryDelta`, it means I don't feel safe with the old one. So I wouldn't want to continue using the old one, even for already created invoices.

t-bast: Right, that may be surprising behavior for node operators, but I agree with you that it makes sense.

We should still compare to the current block height though, I fixed that in https://github.com/ACINQ/eclair/commit/65624656e6390b03988d983f37e9529987562534 and re-added the payment packet test.

thomash-acinq: Added your commit.

> We should still compare to the current block height though

I was already doing it.

t-bast: True, but the code was more clumsy, it's cleaner now.

t-bast (approved):

PR #2390, opened 2022-08-19
Unlock utxos during dual funding failures

t-bast:
When an alternative funding transaction confirms, we need to unlock other candidates: we may not have published them yet if for example we didn't receive remote signatures.

This is a breaking change in codecs since #2274, but shouldn't impact anyone if we merge this before people start using dual funding.


eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -512,9 +517,9 @@ final case class DATA_NEGOTIATING(commitments: Commitments,
   require(!commitments.localParams.isInitiator || closingTxProposed.forall(_.nonEmpty), "initiator must have at least one closing signature for every negotiation attempt because it initiates the closing")
 }
 final case class DATA_CLOSING(commitments: Commitments,
-                              fundingTx: Option[Transaction], // this will be non-empty if we are the initiator and we got in closing while waiting for our own tx to be published
+                              fundingTx: Option[UnconfirmedFundingTx],
```
t-bast, line 520: The new trait is a bit verbose, but it highlights the fact that this is only used when the funding tx is unconfirmed (which wasn't obvious before). Alternatively, we could just use an `Option[Either[Transaction, SignedSharedTransaction]]`, let me know what you prefer.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1110,6 +1110,11 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
           log.info("channelId={} was confirmed at blockHeight={} txIndex={} with a previous funding txid={}", d.channelId, w.blockHeight, w.txIndex, w.tx.txid)
           watchFundingTx(commitments1)
           context.system.eventStream.publish(TransactionConfirmed(d.channelId, remoteNodeId, w.tx))
+          val otherFundingTxs = d.fundingTx match {
+            case Some(DualFundedUnconfirmedFundingTx(fundingTx)) => fundingTx +: d.alternativeCommitments.filter(_.commitments.commitInput.outPoint.txid != w.tx.txid).map(_.fundingTx)
+            case _ => d.alternativeCommitments.filter(_.commitments.commitInput.outPoint.txid != w.tx.txid).map(_.fundingTx)
+          }
```
pm47, line 1116:
```suggestion
          val otherFundingTxs =
            d.fundingTx.toSeq.collect { case DualFundedUnconfirmedFundingTx(fundingTx) => fundingTx } ++
              d.alternativeCommitments.filterNot(_.commitments.commitInput.outPoint.txid == w.tx.txid).map(_.fundingTx)
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
```diff
@@ -109,4 +113,16 @@ trait DualFundingHandlers extends CommonFundingHandlers {
     }
   }
 
+  /**
+   * In most cases we don't need to explicitly rollback funding transactions, as the locks are automatically removed by
+   * bitcoind when transactions are published. But if we didn't publish those transactions (e.g. because our peer never
+   * sent us their signatures), their inputs may still be locked.
+   */
+  def rollbackDualFundingTxs(txs: Seq[SignedSharedTransaction]): Unit = {
```
pm47, line 121: If our peer didn't send us their sig, shouldn't the signature be:
```suggestion
  def rollbackDualFundingTxs(txs: Seq[PartiallySignedSharedTransaction]): Unit = {
```

t-bast: We could, but it's more code to filter those out and it doesn't hurt to call `rollback` with all txs, so I opted for simplicity here.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2390/commits/5eda6bd4fff10fd8ee0d68f5f20addb34901d343

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
t-bast: Actually it's better to use them all, because there is also the case where we received their remote sigs but failed to publish (either because our mempool is full or anything else), in which case we still want to unlock even `FullySignedSharedTransaction`s. The comment mentions one reason why we didn't publish, but it's not meant to be exhaustive. I'll enrich it though to ensure we don't change this signature in the future.

pm47: How about keeping the signature as-is and filtering inside the method, or just adding a specific comment? Just to clear up the confusion, I thought I missed something when reading.

t-bast: Done in https://github.com/ACINQ/eclair/pull/2390/commits/764ec6d10e5da2faf10a20e9eacdb3f2b4940cae

pm47 (approved):

PR #2400, opened 2022-08-30
Add anchor outputs zero fee spec test vector

t-bast:
As specified in https://github.com/lightning/bolts/pull/1018

Our test suite for those test vectors isn't very flexible as it depends on the test name and expects all commitment formats to have exactly the same set of tests, which isn't true for zero fee htlc txs where the feerate doesn't impact which htlcs materialize in the commit tx (only the dust limit impacts that).

It's a bit hacky, but I chose to make minimal modifications to the test file to fit the new tests in, it can be a bit confusing but I don't think it's worth refactoring as we shouldn't change this often.


pm47 (approved):
It doesn't look _that_ hacky to me.

PR #2401, opened 2022-08-30
Allow receiving to blinded routes

thomash-acinq:
All the metadata that is present in classic payments (total amount, payment secret, etc) is put in the encrypted recipient data of the blinded route so that a payment to a blinded route can be processed the same as a classic payment.


thomash-acinq:
Replaced by the competing https://github.com/ACINQ/eclair/pull/2416

PR #2402, opened 2022-08-31
More lenient `interactive-tx` RBF validation

t-bast:
When fee-bumping an `interactive-tx`, we want to be more lenient and accept transactions that improve the feerate, even if peers didn't contribute equally to the feerate increase.

This is particularly useful for scenarios where the non-initiator dedicated a full utxo for the channel and doesn't want to expose new utxos when bumping the fees (or doesn't have more utxos to allocate).

Let's walk through a sample flow. Alice opens a dual-funded channel to Bob. Bob adds a whole utxo to the channel, without any change output. Then Alice wants to increase the feerate of the funding transaction because it doesn't confirm. Bob would have to add another input to the transaction to correctly contribute to the feerate increase (and potentially a change output). If Bob doesn't have funds left in his wallet, he will just use the same input again. Alice will add more fees (since she initiated the RBF attempt), so the feerate will not reached her desired feerate, but it will still increase compared to the previous transaction. From Alice's point of view, it's still worthwhile to accept this RBF attempt: it's better than being stuck with a transaction that doesn't confirm.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
@@ -621,8 +621,10 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
         // This is an RBF attempt: even if our peer does not contribute to the feerate increase, we'd like to broadcast
         // the new transaction if it has a better feerate than the previous one. This is better than being stuck with
         // a transaction that doesn't confirm.
-        if (previousTx.tx.remoteAmountIn == sharedTx.remoteAmountIn && previousTx.tx.remoteOutputs.map(_.amount).sum == sharedTx.remoteOutputs.map(_.amount).sum) {
-          log.info("peer did not contribute to the feerate increase to {}", fundingParams.targetFeerate)
+        val remoteInputsUnchanged = previousTx.tx.remoteInputs.map(_.outPoint).toSet == sharedTx.remoteInputs.map(_.outPoint).toSet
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
```diff
@@ -606,10 +616,25 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
...
-      return Left(InvalidCompleteInteractiveTx(fundingParams.channelId))
+    previousTransactions.headOption match {
+      case Some(previousTx) =>
+        // This is an RBF attempt: even if our peer does not contribute to the feerate increase, we'd like to broadcast
+        // the new transaction if it has a better feerate than the previous one. This is better than being stuck with
+        // a transaction that doesn't confirm.
+        val previousUnsignedTx = previousTx.tx.buildUnsignedTx()
+        val previousMinimumWeight = previousUnsignedTx.weight() + previousUnsignedTx.txIn.length * minimumWitnessWeight
+        val previousFeerate = Transactions.fee2rate(previousTx.tx.fees, previousMinimumWeight)
+        val nextFeerate = Transactions.fee2rate(sharedTx.fees, minimumWeight)
```
pm47, line 627: We should probably compare to `Transactions.weight2fee(fundingParams.targetFeerate, minimumWeight)` and log the fact that the non-initiator didn't contribute to the feerate increase.

t-bast: Done in https://github.com/ACINQ/eclair/pull/2402/commits/c9afb83fcdab18d91660b86a2d7307032db2ccc5, I added a log for the case where the remote peer contributed the exact same amount than the previous iteration.

pm47 (approved):

PR #2404, opened 2022-09-01
Support `scid_alias` and `zero_conf` for all commitment types

### LINKED ISSUE #2394: Channel Open from LDK to Acinq results in error
*Describe the issue*

Have not tested this with a local eclair yet but when opening a channel from an LDK node to the Acinq node, I got this error on the LDK side: 

```
EVENT: Channel feb7300a9210b8087834fc6a0f3f348b3561f56613f05ddb4ff917cd88bed418 closed due to: ProcessingError { err: "Channel Type in accept_channel didn't match the one sent in open_channel." } 
```

One thing I did try though was using the SCID_ALIAS flag so I don't know if that had anything to do with it. But chatting with Matt, apparently it's a protocol rule that channel type should always match when sending an `accept_channel`? Could it also be because the channel type was not anchor output? 

*What behaviour did you expect?*

Channel Opens

*What was the actual behaviour?*

Channel does not open

*How reliably can you reproduce the issue, what are the steps to do so?*

Use [Sensei](https://github.com/L2-Technology/sensei) to open a channel from it to an Eclair node. I could maybe do more testing locally to try that. 

*What version of Eclair are you using, where did you get it (website, self-compiled, etc)?*

The version that Acinq is currently using on Mainnet. 

*Any extra information that might be useful in the debugging process.*

I don't have much more logs on my side but happy to try to reproduce with a regtest elcair node at some point if it would be useful.

t-bast:
We previously only supported `scid_alias` and `zero_conf` when used in combination with `anchors_zero_fee_htlc_tx`, but we have no way of advertising that to our peers, which leads to confusing failures when opening channels.

Some nodes that don't have good access to a utxo pool may not migrate to anchor outputs, but may still want to use `scid_alias` (and potentially `zero_conf` as well).

This isn't very elegant, as we duplicate the `scidAlias` and `zeroConf` parameters, but at least it's very easy to read and reason about. This implies we may want to revisit the way we type `ChannelTypes`, but I'm deferring that to when we introduce another modifier that increases the cardinality (if we ever do).

Fixes #2394


eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
```diff
@@ -92,23 +94,35 @@ sealed trait SupportedChannelType extends ChannelType {
 object ChannelTypes {
 
   // @formatter:off
-  case object Standard extends SupportedChannelType {
-    override def features: Set[ChannelTypeFeature] = Set.empty
+  case class Standard(scidAlias: Boolean, zeroConf: Boolean) extends SupportedChannelType {
```
pm47, line 97: Default values would make sense here I think, and would make tests a bit less verbose.
```suggestion
  case class Standard(scidAlias: Boolean = false, zeroConf: Boolean = false) extends SupportedChannelType {
```

pm47 (approved):
LGTM, and getting back consistency between all channel types which is nice, but I think we are reaching the limit of our data model.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2404/commits/6979d3f8536adfe89979bf1049f3884665cfa69b

t-bast:
> but I think we are reaching the limit of our data model.

I agree! We need to make sure we revisit this when the next set of channel type modifiers is added.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelFeatures.scala
pm47: This could be done for the other classes as well don't you think? Sorry if that wasn't clear.

t-bast: I hesitated, since you specifically highlighted this one I think you only wanted it. In that case we should do it for all of them, including anchor-outputs-zero-fees, right?

pm47: Yes I guess so.

t-bast: Done in https://github.com/ACINQ/eclair/pull/2404/commits/96d7a478987a06c048a20040f4a4a2775a49fd76, this feels lighter!

pm47 (approved):

PR #2407, opened 2022-09-02
Use deterministic serial IDs in interactive-tx

t-bast:
We previously generated random values, but the randomness doesn't protect against anything and adds a risk of re-using the same serial ID twice. It's a better idea to just increment serial IDs (while respecting the parity spec requirement).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
@@ -30,8 +30,8 @@ import fr.acinq.eclair.crypto.keymanager.ChannelKeyManager
 import fr.acinq.eclair.transactions.Transactions
 import fr.acinq.eclair.transactions.Transactions.TxOwner
 import fr.acinq.eclair.wire.protocol._
-import fr.acinq.eclair.{Logs, MilliSatoshiLong, UInt64, randomBytes, randomKey}
-import scodec.bits.{ByteVector, HexStringSyntax}
+import fr.acinq.eclair.{Logs, MilliSatoshiLong, UInt64, randomKey}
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
```diff
@@ -387,7 +387,11 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
           log.info("added {} inputs and {} outputs to interactive tx", inputDetails.usableInputs.length, outputs.length)
           // We unlock the unusable inputs from previous iterations (if any) as they can be used outside of this session.
           unlock(unusableInputs.map(_.outpoint))
-          stash.unstashAll(buildTx(FundingContributions(inputDetails.usableInputs, outputs)))
+          // The initiator's serial IDs must use even values and the non-initiator odd values.
+          val serialIdParity = if (fundingParams.isInitiator) 0 else 1
+          val txAddInputs = inputDetails.usableInputs.zipWithIndex.map { case (input, i) => input.copy(serialId = UInt64(2 * i + serialIdParity)) }
+          val txAddOutputs = outputs.zipWithIndex.map { case (output, i) => output.copy(serialId = UInt64(2 * (i + txAddInputs.length) + serialIdParity)) }
```
pm47, line 393: For fun... yours is safer!

```suggestion
          val iterator = (serialIdParity to Int.MaxValue by 2).iterator
          val txAddInputs = inputDetails.usableInputs.zip(iterator).map { case (input, i) => input.copy(serialId = UInt64(i)) }
          val txAddOutputs = outputs.zip(iterator).map { case (output, i) => output.copy(serialId = UInt64(i)) }
```

pm47 (approved):

PR #2410, opened 2022-09-05
Assume sent htlcs will succeed in the balance computation

pm47:
While it makes sense to assume that relayed payments will fail in the context of balance computation (meaning that we don't earn a fee), the opposite is true for payments sent from the local node, which will cause the full htlc amount to be deducted from the balance.

This way we are consistent: the balance computation is pessismistic and always assume the lowest outcome.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/balance/CheckBalanceSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/balance/CheckBalanceSpec.scala
@@ -12,9 +12,10 @@ import fr.acinq.eclair.channel.states.ChannelStateTestsBase
 import fr.acinq.eclair.channel.{CLOSING, CMD_SIGN, DATA_CLOSING, DATA_NORMAL}
 import fr.acinq.eclair.db.jdbc.JdbcUtils.ExtendedResultSet._
 import fr.acinq.eclair.db.pg.PgUtils.using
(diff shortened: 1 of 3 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/balance/CheckBalanceSpec.scala
```diff
@@ -180,7 +183,7 @@ class CheckBalanceSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
     val aliceCommitTx = alice.stateData.asInstanceOf[DATA_NORMAL].commitments.localCommit.commitTxAndRemoteSig.commitTx.tx
     alice ! Error(ByteVector32.Zeroes, "oops")
     assert(alice2blockchain.expectMsgType[PublishFinalTx].tx.txid == aliceCommitTx.txid)
-    assert(aliceCommitTx.txOut.size == 6) // two main outputs and 4 pending htlcs
+    assert(aliceCommitTx.txOut.size == 7) // two main outputs and 4 pending htlcs (one is dust)
```
t-bast, line 186:
```suggestion
    assert(aliceCommitTx.txOut.size == 7) // two main outputs and 5 pending htlcs (one is dust)
```

t-bast (approved):

PR #2413, opened 2022-09-07
Nits

pm47:
Minor things that I caught during rebase.


eclair-core/src/main/scala/fr/acinq/eclair/remote/EclairInternalsSerializer.scala
```diff
@@ -99,7 +99,7 @@ object EclairInternalsSerializer {
       ("pathFindingExperimentConf" | pathFindingExperimentConfCodec) ::
       ("balanceEstimateHalfLife" | finiteDurationCodec)).as[RouterConf]
 
-  val overrideFeaturesListCodec: Codec[List[(PublicKey, Features[Feature])]] = listOfN(uint16, publicKey ~ variableSizeBytes(uint16, lengthPrefixedFeaturesCodec))
+  val overrideFeaturesListCodec: Codec[List[(PublicKey, Features[Feature])]] = listOfN(uint16, publicKey ~ lengthPrefixedFeaturesCodec)
```
pm47, line 102: This is a backward-incompatible change: front and back need to be updated simultaneously.

t-bast (approved):

PR #2414, opened 2022-09-07
Separate tlv decoding from content validation

t-bast:
When decoding a tlv stream, we previously also validated the stream's content at decoding time. This was a layer violation, as checking that specific tlvs are present in a stream is not an encoding concern.

This was somewhat fine when we only had very basic validation (presence or absence of specific tlvs), but blinded paths substantially changed that because one of the tlvs must be decrypted to yield another tlv stream that also needs to have its content validated. Our validation thus became incomplete, since it didn't decrypt the `encrypted_recipient_data` tlv to validate its own tlv contents.

This forced us to have an overly complex trait hierarchy in `PaymentOnion.scala` and expose a blinding key in classes that shouldn't care about whether blinding is used or not.

We now decouple that into two distinct steps:

* codecs simply return tlv streams and verify that tlvs are correctly encoded
* business logic case classes (such as `ChannelRelayPayload`) should be instantiated with a new `validate` method that takes tlv streams and verifies mandatory/forbidden tlvs

This lets us greatly simplify the trait hierarchy and deal with case class that only contain fully decrypted and valid data.

:information_source: note to reviewers: I recommend starting with the changes in `PaymentOnion.scala` and `PaymentPacket.scala`, as they contain the gist of the change. The other changes are simply replicating this same strategy to other codecs and fixing the build and tests.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
@@ -126,7 +126,7 @@ object IncomingPaymentPacket {
             }
           case None if add.blinding_opt.isDefined => Left(InvalidOnionPayload(UInt64(12), 0))
           case None => IntermediatePayload.ChannelRelay.Standard.validate(payload).left.map(_.failureMessage).map {
-            // NB: we don't validate the ChannelRelayPacket here because its fees and cltv depend on what channel we'll choose to use.
+            // NB: we will validate fees and cltv later, once we've found a suitable channel to relay this payment.
             payload => ChannelRelayPacket(add, payload, nextPacket)
           }
         }
@@ -155,7 +155,7 @@ object IncomingPaymentPacket {
     }
   }
 
-  private def validateBlindedChannelRelayPayload(add: UpdateAddHtlc, payload: TlvStream[OnionPaymentPayloadTlv], blindedPayload: TlvStream[RouteBlindingEncryptedDataTlv], nextBlinding: PublicKey, nextPacket: OnionRoutingPacket): Either[FailureMessage, IncomingPaymentPacket] = {
+  private def validateBlindedChannelRelayPayload(add: UpdateAddHtlc, payload: TlvStream[OnionPaymentPayloadTlv], blindedPayload: TlvStream[RouteBlindingEncryptedDataTlv], nextBlinding: PublicKey, nextPacket: OnionRoutingPacket): Either[FailureMessage, ChannelRelayPacket] = {
     IntermediatePayload.ChannelRelay.Blinded.validate(payload, blindedPayload, nextBlinding).left.map(_.failureMessage).flatMap {
       // TODO: return an unparseable error
       case payload if add.amountMsat < payload.paymentConstraints.minAmount => Left(InvalidOnionPayload(UInt64(12), 0))
@@ -165,7 +165,7 @@ object IncomingPaymentPacket {
     }
   }
 
-  private def validateFinalPayload(add: UpdateAddHtlc, payload: TlvStream[OnionPaymentPayloadTlv]): Either[FailureMessage, IncomingPaymentPacket] = {
+  private def validateFinalPayload(add: UpdateAddHtlc, payload: TlvStream[OnionPaymentPayloadTlv]): Either[FailureMessage, FinalPacket] = {
     FinalPayload.Standard.validate(payload).left.map(_.failureMessage).flatMap {
       case payload if add.amountMsat != payload.amount => Left(FinalIncorrectHtlcAmount(add.amountMsat))
       case payload if add.cltvExpiry != payload.expiry => Left(FinalIncorrectCltvExpiry(add.cltvExpiry))
@@ -173,7 +173,7 @@ object IncomingPaymentPacket {
     }
   }
 
-  private def validateFinalPayload(add: UpdateAddHtlc, outerPayload: TlvStream[OnionPaymentPayloadTlv], innerPayload: TlvStream[OnionPaymentPayloadTlv]): Either[FailureMessage, IncomingPaymentPacket] = {
+  private def validateFinalPayload(add: UpdateAddHtlc, outerPayload: TlvStream[OnionPaymentPayloadTlv], innerPayload: TlvStream[OnionPaymentPayloadTlv]): Either[FailureMessage, FinalPacket] = {
     // The outer payload cannot use route blinding, but the inner payload may (but it's not supported yet).
     FinalPayload.Standard.validate(outerPayload).left.map(_.failureMessage).flatMap { outerPayload =>
       FinalPayload.Standard.validate(innerPayload).left.map(_.failureMessage).flatMap {
@@ -189,7 +189,7 @@ object IncomingPaymentPacket {
     }
   }
 
-  private def validateNodeRelay(add: UpdateAddHtlc, outerPayload: TlvStream[OnionPaymentPayloadTlv], innerPayload: TlvStream[OnionPaymentPayloadTlv], next: OnionRoutingPacket): Either[FailureMessage, IncomingPaymentPacket] = {
+  private def validateNodeRelay(add: UpdateAddHtlc, outerPayload: TlvStream[OnionPaymentPayloadTlv], innerPayload: TlvStream[OnionPaymentPayloadTlv], next: OnionRoutingPacket): Either[FailureMessage, NodeRelayPacket] = {
     // The outer payload cannot use route blinding, but the inner payload may (but it's not supported yet).
     FinalPayload.Standard.validate(outerPayload).left.map(_.failureMessage).flatMap { outerPayload =>
       IntermediatePayload.NodeRelay.Standard.validate(innerPayload).left.map(_.failureMessage).flatMap {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
@@ -24,12 +24,12 @@ import fr.acinq.eclair.channel.{CMD_ADD_HTLC, CMD_FAIL_HTLC, CannotExtractShared
 import fr.acinq.eclair.crypto.Sphinx
 import fr.acinq.eclair.router.Router.{ChannelHop, Hop, NodeHop}
 import fr.acinq.eclair.wire.protocol._
-import fr.acinq.eclair.{CltvExpiry, CltvExpiryDelta, Feature, Features, MilliSatoshi, UInt64, randomKey}
+import fr.acinq.eclair.{CltvExpiry, CltvExpiryDelta, Features, MilliSatoshi, UInt64, randomBytes32, randomKey}
 import scodec.bits.ByteVector
-import scodec.{Attempt, Codec, DecodeResult}
+import scodec.{Attempt, DecodeResult}
 
 import java.util.UUID
-import scala.util.{Failure, Success, Try}
+import scala.util.Try
 
 /**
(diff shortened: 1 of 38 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -42,40 +42,68 @@ object IncomingPaymentPacket {
...
-          // It's hard to provide tag and offset information from scodec failures, so we currently don't do it.
-          case Attempt.Failure(_) => Left(InvalidOnionPayload(UInt64(0), 0))
+      case Right(p: Sphinx.DecryptedPacket) =>
+        PaymentOnionCodecs.perHopPayloadCodec.decode(p.payload.bits) match {
+          case Attempt.Successful(DecodeResult(perHopPayload, _)) if p.isLastPacket => Right(DecodedOnionPacket(perHopPayload, None))
+          case Attempt.Successful(DecodeResult(perHopPayload, _)) => Right(DecodedOnionPacket(perHopPayload, Some(p.nextPacket)))
+          case Attempt.Failure(_) =>
+            // Onion is correctly encrypted but the content of the per-hop payload couldn't be decoded.
+            // It's hard to provide tag and offset information from scodec failures, so we currently don't do it.
+            Left(InvalidOnionPayload(UInt64(0), 0))
```
pm47, line 70: Logging the decrypted bytes could be really helpful for debugging.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -42,40 +42,68 @@ object IncomingPaymentPacket {
...
         }
       case Left(badOnion) => Left(badOnion)
     }
 
+  case class DecodedEncryptedRecipientData(payload: TlvStream[RouteBlindingEncryptedDataTlv], nextBlinding: PublicKey)
+
+  private[payment] def decryptEncryptedRecipientData(add: UpdateAddHtlc, privateKey: PrivateKey, payload: TlvStream[OnionPaymentPayloadTlv], encryptedRecipientData: ByteVector): Either[FailureMessage, DecodedEncryptedRecipientData] = {
+    if (add.blinding_opt.isDefined && payload.get[OnionPaymentPayloadTlv.BlindingPoint].isDefined) {
+      // TODO: return an unparseable error
+      Left(InvalidOnionPayload(UInt64(12), 0))
```
pm47, line 80: Why 12?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -42,40 +42,68 @@ object IncomingPaymentPacket {
...
+        case Some(blinding) => RouteBlindingEncryptedDataCodecs.decode(privateKey, blinding, encryptedRecipientData) match {
+          case Left(_) =>
+            // There are two possibilities in this case:
+            //  - the blinding point is invalid: the sender or the previous node is buggy or malicious
+            //  - the encrypted data is invalid: the sender, the previous node or the recipient must be buggy or malicious
+            // TODO: return an unparseable error
+            Left(InvalidOnionPayload(UInt64(12), 0))
+          case Right(decoded) => Right(DecodedEncryptedRecipientData(decoded.tlvs, decoded.nextBlinding))
+        }
+        case None =>
```
pm47, line 92: I would have handled this in the `decrypt` method and simplify the signature of `decryptEncryptedRecipientData`

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -87,108 +115,87 @@ object IncomingPaymentPacket {
...
+            decryptEncryptedRecipientData(add, privateKey, payload, encryptedRecipientData).flatMap {
+              case DecodedEncryptedRecipientData(blindedPayload, nextBlinding) =>
+                validateBlindedChannelRelayPayload(add, payload, blindedPayload, nextBlinding, nextPacket)
             }
-          case _ if add.blinding_opt.isDefined => Left(InvalidOnionPayload(UInt64(12), 0))
-          // NB: we don't validate the ChannelRelayPacket here because its fees and cltv depend on what channel we'll choose to use.
-          case payload: PaymentOnion.ChannelRelayTlvPayload => Right(ChannelRelayPacket(add, payload, next, None))
+          case None if add.blinding_opt.isDefined => Left(InvalidOnionPayload(UInt64(12), 0))
+          case None => PaymentOnion.ChannelRelayPayload.validate(payload).left.map(_.failureMessage).map {
+            // NB: we don't validate the ChannelRelayPacket here because its fees and cltv depend on what channel we'll choose to use.
```
pm47, line 128: In other words there is nothing to validate right?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -87,108 +115,87 @@ object IncomingPaymentPacket {
...
         }
     }
   }
 
-  private def isValidBlindedPayment(data: BlindedRouteData.PaymentData, amount: MilliSatoshi, cltvExpiry: CltvExpiry, features: Features[Feature]): Boolean = {
-    val amountOk = amount >= data.paymentConstraints.minAmount
-    val expiryOk = cltvExpiry <= data.paymentConstraints.maxCltvExpiry
-    val featuresOk = Features.areCompatible(features, data.allowedFeatures)
-    amountOk && expiryOk && featuresOk
+  private def validateBlindedChannelRelayPayload(add: UpdateAddHtlc, payload: TlvStream[OnionPaymentPayloadTlv], blindedPayload: TlvStream[RouteBlindingEncryptedDataTlv], nextBlinding: PublicKey, nextPacket: OnionRoutingPacket): Either[FailureMessage, IncomingPaymentPacket] = {
```
pm47, line 157: Same on other methods:
```suggestion
  private def validateBlindedChannelRelayPayload(add: UpdateAddHtlc, payload: TlvStream[OnionPaymentPayloadTlv], blindedPayload: TlvStream[RouteBlindingEncryptedDataTlv], nextBlinding: PublicKey, nextPacket: OnionRoutingPacket): Either[FailureMessage, ChannelRelayPacket] = {
```

I wonder if, you could have used types instead of methods for the validation. E.g.: `TentativeChannelRelayHopPayload`/`ChannelRelayHopPayload`? Probably overkill and useless.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -185,91 +185,109 @@ object PaymentOnion {
...
+   * @param blindedRecords decrypted tlv stream from the encrypted_recipient_data tlv.
+   * @param nextBlinding   blinding point that must be forwarded to the next hop.
+   */
+  case class BlindedChannelRelayPayload(records: TlvStream[OnionPaymentPayloadTlv], blindedRecords: TlvStream[RouteBlindingEncryptedDataTlv], nextBlinding: PublicKey) extends ChannelRelayHopPayload {
+    // @formatter:off
+    override val outgoingChannelId = blindedRecords.get[RouteBlindingEncryptedDataTlv.OutgoingChannelId].get.shortChannelId
+    val paymentRelay = blindedRecords.get[RouteBlindingEncryptedDataTlv.PaymentRelay].get
+    val paymentConstraints = blindedRecords.get[RouteBlindingEncryptedDataTlv.PaymentConstraints].get
+    val allowedFeatures = blindedRecords.get[RouteBlindingEncryptedDataTlv.AllowedFeatures].map(_.features).getOrElse(Features.empty)
+    override def amountToForward(incomingAmount: MilliSatoshi): MilliSatoshi = ((incomingAmount - paymentRelay.feeBase).toLong * 1_000_000 + 1_000_000 + paymentRelay.feeProportionalMillionths - 1).msat / (1_000_000 + paymentRelay.feeProportionalMillionths)
```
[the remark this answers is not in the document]
t-bast: It's just standard formula to avoid truncation with integer arithmetic, see https://github.com/ACINQ/eclair/pull/2367
Rusty mentioned that this is how `ceil` operations are done inside the linux kernel.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: It's the tlv tag for the blinding point, but never mind that, it was there before with a TODO to fix it, and is fixed in #2408

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: I'm not convinced, this is really a validation step that is tied to the encrypted data, I think it makes sense to have it in this function...also if done in `decrypt`, we'd need to duplicate that check (since we do it for the intermediate and final case) or re-order them, which I don't like (I think the separation should be intermediate vs final *before* separating standard vs blinded, even though it's arguable).

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/2414/commits/318e59e425ebe19877db1800523aa31e788995f8

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: I clarified the comment in https://github.com/ACINQ/eclair/pull/2414/commits/318e59e425ebe19877db1800523aa31e788995f8, let me know if that's better

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: I wasn't a big fan of adding a logger to this function, and it's also somewhat of a DoS vector if we logged those. This error happens when the sender is buggy, I'd rather have them log their data to figure out what they did wrong.

We could also get spammed by such errors that could force us to log 1kB every time, which isn't great.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
pm47: I don't see your correction?

t-bast: That's strange, I do see it in the commit I'm linking...is this a github caching issue?

pm47: I think the validation for other messages are about the integrity and self-constency of the payload, there just isn't any validation to do for the `ChannelRelayPacket`. The fact that some other checks are happening later in the relayer is another matter, there may be other checks for other types of packets too?

t-bast: Yes that's true, it's a good argument for just removing this comment entirely, I'll do that.

t-bast: Removed in https://github.com/ACINQ/eclair/pull/2414/commits/2382b63691cbe853257f264398a03f51429fd56d

pm47 (approved):
LGTM! Feels much cleaner.

thomash-acinq (approved):
It's much better indeed. I'm just a bit sad that I will need to rebase my PR on top of this.

PR #2415, opened 2022-09-08
Fix collision bug in local alias generation

pm47:
The bug is due to mistakingly using the `^` as power operator, while it instead is a `xor`. As a result, the available space for local alias was tiny, resulting in collisions. This in turns causes the `relayer` to forward `UpdateAddHtlc` to the wrong node, which results in `UpdateFailMalformed` error due to the peers being unable to decrypt the onion that wasn't meant for them.

Nice catch @robbiehanson!


t-bast (approved):
Good catch!

PR #2416, opened 2022-09-09
Receive payments for Bolt 12 invoices

t-bast:
We add support for generating Bolt 12 invoices and storing them in our payments DB to then receive blinded payments. We implement the receiving part once a blinded payment has been decrypted (which means we also need #2401 to fully support receiving blinded payments). This uses the same payment flow as non-blinded payments, with slightly different validation steps.

Note that we put the payment preimage in the blinded paths' path_id field to verify that an incoming payment uses a valid blinded route generated by us. We store that as an arbitrary byte vector to allow future changes to this strategy.

Since this requires a DB schema update, please think carefully about this change and let me know if you think this is future-proof or if we can improve this more.


eclair-core/src/main/scala/fr/acinq/eclair/db/PaymentsDb.scala
```diff
@@ -88,27 +92,50 @@ trait OutgoingPaymentsDb {
...
  */
-case class IncomingPayment(invoice: Bolt11Invoice,
-                           paymentPreimage: ByteVector32,
-                           paymentType: String,
-                           createdAt: TimestampMilli,
-                           status: IncomingPaymentStatus)
+case class IncomingBlindedPayment(invoice: Bolt12Invoice,
+                                  paymentPreimage: ByteVector32,
+                                  paymentType: String,
+                                  pathIds: Map[PublicKey, ByteVector],
```
thomash-acinq, line 136: Why do we need a `Map[PublicKey, ByteVector]`? Isn't a single pathId enough?
And actually you've written in the commit description that "we put the payment preimage in the blinded paths' path_id field", so we don't even need to save it in a new field, it's already in `paymentPreimage`.

t-bast: This is my main open question, that's why I said in the PR description:

```
Note that we put the payment preimage in the blinded paths' path_id field to verify that an incoming payment uses a valid blinded route generated by us. We store that as an arbitrary byte vector to allow future changes to this strategy.

Since this requires a DB schema update, please think carefully about this change and let me know if you think this is future-proof or if we can improve this more.
```

I don't know how far ahead we should try to think, so I kept the most general design: we do need to keep the blinding point for each route (if that's not what we receive, something is wrong), and maybe it would make sense at some point to have different `path_ids` per route?

t-bast: I'm not sure for example if an attacker could exploit the fact that we use the same `path_id` on multiple routes...if that can be exploited we'll need to store a different one per route, for example preimage and a tweak?

t-bast: I really don't know what's best, but wanted to pick your brains on it, I think it's worth thinking about to avoid a future DB migration

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -226,32 +270,50 @@ object MultiPartHandler {
...
+                  )
+                  context.log.debug("generated invoice={} from amount={}", invoice.toString, r.amount_opt)
+                  nodeParams.db.payments.addIncomingPayment(invoice, paymentPreimage, r.paymentType)
+                  invoice
+                case r: ReceiveOfferPayment =>
+                  // TODO: get blinded paths from the router instead
+                  val pathId = RouteBlindingEncryptedDataTlv.PathId(randomBytes(32))
+                  val dummyConstraints = RouteBlindingEncryptedDataTlv.PaymentConstraints(CltvExpiry(nodeParams.currentBlockHeight + 144), 1 msat)
+                  val dummyRelay = RouteBlindingEncryptedDataTlv.PaymentRelay(CltvExpiryDelta(0), 0, 0 msat)
+                  val dummyScid = RouteBlindingEncryptedDataTlv.OutgoingChannelId(ShortChannelId(0))
```
thomash-acinq, line 305: There can be a collision with a local alias. In https://github.com/ACINQ/eclair/pull/2401, I've used `val toSelf: Alias = Alias(aliasUpperBound)`.
You also need to add some code in `IncomingPaymentPacket.decrypt` to peel one more layer when we get this alias.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
```diff
@@ -321,7 +321,7 @@ object PostRestartHtlcCleaner {
    */
   private def shouldFulfill(finalPacket: IncomingPaymentPacket.FinalPacket, paymentsDb: IncomingPaymentsDb): Option[ByteVector32] =
     paymentsDb.getIncomingPayment(finalPacket.add.paymentHash) match {
-      case Some(IncomingPayment(_, preimage, _, _, IncomingPaymentStatus.Received(_, _))) => Some(preimage)
+      case Some(p) if p.status.isInstanceOf[IncomingPaymentStatus.Received] => Some(p.paymentPreimage)
```
thomash-acinq, line 324: You don't like pattern matching?

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: My goal really wasn't to make workable blinded paths here, I was deferring that to a PR that would implement generating real blinded paths in the router. This PR in itself is incomplete, I was expecting us to integrate this one, your two PRs for sending (#2395) and receiving (#2401, which could add the decryption of dummy hops) and another PR to generate blinded paths before we can actually test this end-to-end, how does that sound?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala
t-bast: Right, I can make that more elegant, will do.

t-bast: Done in https://github.com/ACINQ/eclair/pull/2416/commits/35947e4c6e190168edda688b38a7f0ecfcc3ec49

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: Note that all of this is currently dead code, since `PaymentPacket.decrypt` just rejects blinded final payloads. My plan was that your two PRs (and the final one to create real blinded paths) would build on top of this to get us to feature complete.

thomash-acinq (approved):

PR #2418, opened 2022-09-13
Allow receiving to blinded routes

thomash-acinq:
Enable receiving blinded payments.
Unroll the dummy hops at the end of a blinded route.


t-bast (approved):
LGTM :100:

PR #2421, opened 2022-09-15
Add tests for invoice feature validation

t-bast:
Add non-reg tests to ensure that we only validate invoice features, and correctly refuse to pay invoices containing unsupported mandatory features.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
@@ -645,13 +645,13 @@ class Bolt11InvoiceSpec extends AnyFunSuite {
   test("filter non-invoice features when parsing invoices") {
     // The following invoice has feature bit 20 activated (option_anchor_outputs) without feature bit 12 (option_static_remotekey).
     // This doesn't satisfy the feature dependency graph, but since those aren't invoice features, we should ignore it.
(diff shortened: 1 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/payment/Bolt11InvoiceSpec.scala
```diff
@@ -642,7 +642,19 @@ class Bolt11InvoiceSpec extends AnyFunSuite {
     assert(Bolt11Invoice.fromString(invoice.toString).get == invoice)
   }
 
-  test("Invoices can't have high features") {
+  test("filter non-invoice features when parsing invoices") {
+    // The following invoice has feature bit 20 activated (option_anchor_outputs) without feature bit 12 (option_static_remotekey).
+    // This doesn't satisfy the feature dependency graph, but since those aren't invoice features, we should ignore it.
+    val bin = "lntb250u1p3jxdmwpp5mnjuf5zwkndq4e29lmvqen8cnwc57aqr0658jlfl3q42667q3lgsdpqdehkuttfdemx76trv5sxvetpw36hyetncqpxsp5a0epznfadhwjrtsrqykjpxu97800grzzjv3kuq3xnlsdgaytqn3s9pqzqqqqqqzqqqqqqqqqqqqqqqqqqqqqsgqk243r3nnm92d8qaam7hgwc5yhupq4wvmnufyyt6kk6hww98a4vtrzxy0pqrpj7dyffykdfx326k0ggcw4l2d08xkknes4unt6vu3ptgqk9zn8r"
```
thomash-acinq, line 648: Why `bin`? It sounds like "binary" but there is nothing binary here. I would name it `invoice`.

Also this is completely opaque and we just need to trust the comment that this test does what it claims to do. It would be nicer to build the invoice in a way that makes it visible that we set option_anchor_outputs.

t-bast: Agreed, I'll enrich the `createInvoiceUnsafe` to make this more explicit.

t-bast: Improved in https://github.com/ACINQ/eclair/pull/2421/commits/48eac3dd1141292c90de6b683fb1ad73963f4fc6

thomash-acinq (approved):

PR #2430, opened 2022-09-22
Don't set channel reserve for dual funded channels

pm47:
In dual funded channels the reserve is computed automatically. But our model allows setting a reserve even for dual funded channels.

Changing the model is too much work, instead this PR proposes to:
- add `require`s in the `Commitments` class to verify at runtime that we are consistent (it would cause the channel to fail before it is created)
- pass the `dualFunded` status to `Peer.makeChannelParams` so we only build valid parameters.

We could also alter the handlers for `SpawnChannelnitiator`/`SpawnChannelNonInitiator` and mutate the `LocalParams` depending on the value of `dualFunded`. It is less intrusive but arguably more hacky.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -19,6 +19,7 @@ package fr.acinq.eclair.channel
 import akka.event.LoggingAdapter
 import fr.acinq.bitcoin.scalacompat.Crypto.{PrivateKey, PublicKey, sha256}
 import fr.acinq.bitcoin.scalacompat.{ByteVector32, ByteVector64, Crypto, Satoshi, SatoshiLong, Script}
+import fr.acinq.eclair.Features.DualFunding
 import fr.acinq.eclair._
 import fr.acinq.eclair.blockchain.fee.{FeeratePerKw, OnChainFeeConf}
 import fr.acinq.eclair.channel.Helpers.Closing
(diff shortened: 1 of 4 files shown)


t-bast (approved):
LGTM :100:

pm47:
I think I should have done this at the same time:

```diff
diff --git a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
index cc0f65416..d426ff5aa 100644
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -222,14 +222,14 @@ case class Commitments(channelId: ByteVector32,
   val localChannelReserve: Satoshi = if (channelFeatures.hasFeature(Features.DualFunding)) {
     (capacity / 100).max(remoteParams.dustLimit)
   } else {
-    remoteParams.requestedChannelReserve_opt.getOrElse(0 sat)
+    remoteParams.requestedChannelReserve_opt.get
   }

   /** Channel reserve that applies to our peer's funds. */
   val remoteChannelReserve: Satoshi = if (channelFeatures.hasFeature(Features.DualFunding)) {
     (capacity / 100).max(localParams.dustLimit)
   } else {
-    localParams.requestedChannelReserve_opt.getOrElse(0 sat)
+    localParams.requestedChannelReserve_opt.get
   }

   // NB: when computing availableBalanceForSend and availableBalanceForReceive, the initiator keeps an extra buffer on

```

PR #2432, opened 2022-09-23
Remove `channelReserve.getOrElse(0 sat)`

pm47:
With the requirements added by #2430, we can get rid of the superfluous degrees of freedom around channel reserve, while still leaving the model untouched.

This revealed some inconsistencies in test data.


eclair-core/src/test/scala/fr/acinq/eclair/router/RouterSpec.scala
```diff
@@ -695,7 +695,7 @@ class RouterSpec extends BaseRouterSpec {
 
     {
       // Private channels should also update the graph when HTLCs are relayed through them.
-      val balances = Set(33000000 msat, 5000000 msat)
+      val balances = Set(4620000 msat, 32620000 msat)
```
pm47, line 698: 👀

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -222,14 +222,14 @@ case class Commitments(channelId: ByteVector32,
   val localChannelReserve: Satoshi = if (channelFeatures.hasFeature(Features.DualFunding)) {
     (capacity / 100).max(remoteParams.dustLimit)
   } else {
-    remoteParams.requestedChannelReserve_opt.getOrElse(0 sat)
+    remoteParams.requestedChannelReserve_opt.get // this is guarded by a require() in Commitments
```
t-bast, line 225: This isn't correct, the `require` in Commitments only checks that when using dual funding, `requestedChannelReserve_opt` is `None`, not that when dual funding isn't used it is `Some`. And it feels to me that it actually shouldn't, when using zero-reserve it feels more natural to set `requestedChannelReserve_opt` to `None` than to `Some(0 sat)`, isn't it?

pm47: Nice catch! Fixed in 6a4e972bd, and it looks better.

t-bast (approved):

PR #2434, opened 2022-09-23
Clean up channel commands generic types

t-bast:
Our channel commands had unused generic types, which were actually useless because of type erasure.
Removing them simplifies the types and removes a lot of redundant (and useless) test code.


pm47:
> unused generic types, which were actually useless because of type erasure

Useless at runtime, but still useful at compile time right? For example the compiler prevented you to put a `CmdResponse[CMD_FAIL_HTLC]` in a `WrappedAddResponse`.

t-bast:
You're right, that is indeed valuable! The warning that `C` isn't used in the `sealed trait CommandResponse[+C <: Command]` can then be ignored, we do want to keep it.

Issue #2424, opened 2022-09-20
Add support for async payments to offline nodes


remyers:
This issue is to explore and implement initial support and UI for the functionality envisioned by (in-progress) Bolt PR: "Add the ability to hold HTLCs before forwarding (FEAT 52/53) [#989](https://github.com/lightning/bolts/pull/989)" and to help inform the process of standardization for async payments.

Support will be added to hold trampoline payments until triggered using custom messages between Eclair and Phoenix instead of using onion messages. We will use temporary/experimental feature bits and TLV tags for this phase of implementation.

- [ ] eclair: support a new `async_payment` TLV in the onion payload of trampoline payments which causes the trampoline node to not relay a payment until triggered by a message that the receiver is online, canceled by the payment sender or timed out by the routing node itself. 
- [ ] lightning-kmp: relax the payment receiver's `cltv_expiry` check between outer and trampoline onion payloads in order to be able to accept an async payment that were held by a trampoline node.
- [ ] lightning-kmp: add the ability to create an async payment from a Bolt-11 invoice and a custom Lightning message to cancel a pending payment. 
- [ ] eclair: add support to trigger delivery of a pending trampoline payment when its destination Phoenix node re-connects to eclair
- [ ] phoenix:  add UI to support async payments

Once standardization work on [#989](https://github.com/lightning/bolts/pull/989) is complete we can implement the store and forward onion messages and feature bits envisioned for async payments. Integration testing could be made at this point with other nodes that implement async trampoline payments. Further work could then add PTLC and static invoice support.

PR #2435, opened 2022-09-23
Add initial support for async payment trampoline relay

remyers:
This is the first step in adding "async payments" as described in issue #2424 .

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -70,6 +70,7 @@ eclair {
     option_zeroconf = disabled
     keysend = disabled
     trampoline_payment_prototype = disabled
+    async_payment_prototype = disabled
   }
   // The following section lets you customize features for specific nodes.
   // The overrides will be applied on top of the default features settings.
@@ -150,10 +151,8 @@ eclair {
       enforcement-delay = 10 minutes
     }
 
-    async-payments {
-        enabled = false
-        timeout = 1 day // maximum time to hold an async payment while waiting for the receiver to come online
-    }
+    // Maximum time to hold an async payment while waiting for the receiver to come online
+    timeout = 1 day
   }
 
   on-chain-fees {
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -151,8 +151,12 @@ eclair {
       enforcement-delay = 10 minutes
     }
 
-    // Maximum time to hold an async payment while waiting for the receiver to come online
-    timeout = 1 day
+    async-payments {
+      // Maximum number of blocks to hold an async payment while waiting to receive a trigger from the receiver
+      hold-timeout-blocks = 1008
+      // Number of blocks before the incoming HTLC expires that an async payment must be triggered by the receiver
+      cancel-safety-before-timeout-blocks = 144
+    }
   }
 
   on-chain-fees {
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -149,6 +149,11 @@ eclair {
       // Delay enforcement of channel fee updates
       enforcement-delay = 10 minutes
     }
+
+    async-payments {
+        enabled = false
+        timeout = 1 day // maximum time to hold an async payment while waiting for the receiver to come online
+    }
   }
 
   on-chain-fees {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -323,7 +323,7 @@ object Features {
     RouteBlinding -> (VariableLengthOnion :: Nil),
     TrampolinePaymentPrototype -> (PaymentSecret :: Nil),
     KeySend -> (VariableLengthOnion :: Nil),
-    AsyncPaymentPrototype -> (VariableLengthOnion :: TrampolinePaymentPrototype :: Nil)
+    AsyncPaymentPrototype -> (TrampolinePaymentPrototype :: Nil)
   )
 
   case class FeatureException(message: String) extends IllegalArgumentException(message)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -282,7 +282,7 @@ object Features {
   }
 
   // TODO: @remyers update feature bits once spec-ed (currently reserved here: https://github.com/lightning/bolts/pull/989)
-  case object AsyncPaymentPrototype extends Feature with InvoiceFeature {
+  case object AsyncPaymentPrototype extends Feature with InitFeature with InvoiceFeature {
     val rfcName = "async_payment_prototype"
     val mandatory = 152
   }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -281,6 +281,12 @@ object Features {
     val mandatory = 148
   }
 
+  // TODO: @remyers update feature bits once spec-ed (currently reserved here: https://github.com/lightning/bolts/pull/989)
+  case object AsyncPaymentPrototype extends Feature with InvoiceFeature {
+    val rfcName = "async_payment_prototype"
+    val mandatory = 52
+  }
+
   val knownFeatures: Set[Feature] = Set(
     DataLossProtect,
     InitialRoutingSync,
@@ -303,7 +309,8 @@ object Features {
     PaymentMetadata,
     ZeroConf,
     KeySend,
-    TrampolinePaymentPrototype
+    TrampolinePaymentPrototype,
+    AsyncPaymentPrototype
   )
 
   // Features may depend on other features, as specified in Bolt 9.
@@ -315,7 +322,8 @@ object Features {
     AnchorOutputsZeroFeeHtlcTx -> (StaticRemoteKey :: Nil),
     RouteBlinding -> (VariableLengthOnion :: Nil),
     TrampolinePaymentPrototype -> (PaymentSecret :: Nil),
-    KeySend -> (VariableLengthOnion :: Nil)
+    KeySend -> (VariableLengthOnion :: Nil),
+    AsyncPaymentPrototype -> (VariableLengthOnion :: TrampolinePaymentPrototype :: Nil)
   )
 
   case class FeatureException(message: String) extends IllegalArgumentException(message)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
@@ -284,7 +284,7 @@ object Features {
   // TODO: @remyers update feature bits once spec-ed (currently reserved here: https://github.com/lightning/bolts/pull/989)
   case object AsyncPaymentPrototype extends Feature with InvoiceFeature {
     val rfcName = "async_payment_prototype"
-    val mandatory = 52
+    val mandatory = 152
   }
 
   val knownFeatures: Set[Feature] = Set(
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
@@ -209,8 +209,6 @@ class NodeRelay private(nodeParams: NodeParams,
               doSend(upstream, nextPayload, nextPacket)
             }
         }
-      case RelayAsyncPayment => context.log.debug("received async payment trigger while waiting to receive all incoming payment packets")
-        receiving(htlcs, nextPayload, nextPacket, handler)
     }
 
   private def waitForTrigger(upstream: Upstream.Trampoline, nextPayload: IntermediatePayload.NodeRelay.Standard, nextPacket: OnionRoutingPacket): Behavior[Command] = {
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
@@ -56,6 +56,7 @@ object NodeRelay {
   case class Relay(nodeRelayPacket: IncomingPaymentPacket.NodeRelayPacket) extends Command
   case object Stop extends Command
   case object RelayAsyncPayment extends Command
+  case object CancelAsyncPayment extends Command
   private case class WrappedMultiPartExtraPaymentReceived(mppExtraReceived: MultiPartPaymentFSM.ExtraPaymentReceived[HtlcPart]) extends Command
   private case class WrappedMultiPartPaymentFailed(mppFailed: MultiPartPaymentFSM.MultiPartPaymentFailed) extends Command
   private case class WrappedMultiPartPaymentSucceeded(mppSucceeded: MultiPartPaymentFSM.MultiPartPaymentSucceeded) extends Command
@@ -229,6 +230,10 @@ class NodeRelay private(nodeParams: NodeParams,
         stopping()
       case WrappedCurrentBlockHeight(blockHeight) if blockHeight >= timeoutBlock =>
         context.log.warn(s"rejecting async payment at block $blockHeight; was not triggered after waiting ${nodeParams.relayParams.asyncPaymentsParams.holdTimeoutBlocks} blocks")
+        rejectPayment(upstream, Some(TemporaryNodeFailure)) // TODO: replace failure type when async payment spec is finalized
+        stopping()
+      case CancelAsyncPayment =>
+        context.log.warn(s"payment sender canceled a waiting async payment")
         rejectPayment(upstream, Some(TemporaryNodeFailure)) // TODO: replace failure type when async payment spec is finalized
         stopping()
       case RelayAsyncPayment =>
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
@@ -54,7 +54,7 @@ object NodeRelay {
   sealed trait Command
   case class Relay(nodeRelayPacket: IncomingPaymentPacket.NodeRelayPacket) extends Command
   case object Stop extends Command
-  case object AsyncPaymentTrigger extends Command
+  case object RelayAsyncPayment extends Command
   private case class WrappedMultiPartExtraPaymentReceived(mppExtraReceived: MultiPartPaymentFSM.ExtraPaymentReceived[HtlcPart]) extends Command
   private case class WrappedMultiPartPaymentFailed(mppFailed: MultiPartPaymentFSM.MultiPartPaymentFailed) extends Command
   private case class WrappedMultiPartPaymentSucceeded(mppSucceeded: MultiPartPaymentFSM.MultiPartPaymentSucceeded) extends Command
@@ -208,7 +208,7 @@ class NodeRelay private(nodeParams: NodeParams,
               doSend(upstream, nextPayload, nextPacket)
             }
         }
-      case AsyncPaymentTrigger => context.log.debug("received async payment trigger while waiting to receive all incoming payment packets")
+      case RelayAsyncPayment => context.log.debug("received async payment trigger while waiting to receive all incoming payment packets")
         receiving(htlcs, nextPayload, nextPacket, handler)
     }
 
@@ -221,7 +221,7 @@ class NodeRelay private(nodeParams: NodeParams,
           context.log.warn(s"rejecting async payment that was not triggered before relay timeout: ${nodeParams.relayParams.timeout}")
           rejectPayment(upstream, Some(PaymentTimeout))
           stopping()
-        case AsyncPaymentTrigger =>
+        case RelayAsyncPayment =>
           // check cltv timeouts again before forwarding the payment
           validateRelay(nodeParams, upstream, nextPayload) match {
           case Some(failure) =>
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
@@ -43,7 +43,6 @@ import fr.acinq.eclair.{CltvExpiry, Features, Logs, MilliSatoshi, NodeParams, UI
 
 import java.util.UUID
 import scala.collection.immutable.Queue
-import scala.concurrent.duration.FiniteDuration
 
 /**
  * It [[NodeRelay]] aggregates incoming HTLCs (in case multi-part was used upstream) and then forwards the requested amount (using the
@@ -181,15 +180,14 @@ class NodeRelay private(nodeParams: NodeParams,
    * @param nextPayload relay instructions (should be identical across HTLCs in this set).
    * @param nextPacket  trampoline onion to relay to the next trampoline node.
    * @param handler     actor handling the aggregation of the incoming HTLC set.
-   * @param triggered   async payment trigger received
    */
-  private def receiving(htlcs: Queue[UpdateAddHtlc], nextPayload: IntermediatePayload.NodeRelay.Standard, nextPacket: OnionRoutingPacket, handler: ActorRef, triggered: Boolean = false): Behavior[Command] =
+  private def receiving(htlcs: Queue[UpdateAddHtlc], nextPayload: IntermediatePayload.NodeRelay.Standard, nextPacket: OnionRoutingPacket, handler: ActorRef): Behavior[Command] =
     Behaviors.receiveMessagePartial {
       case Relay(IncomingPaymentPacket.NodeRelayPacket(add, outer, _, _)) =>
         require(outer.paymentSecret == paymentSecret, "payment secret mismatch")
         context.log.debug("forwarding incoming htlc #{} from channel {} to the payment FSM", add.id, add.channelId)
         handler ! MultiPartPaymentFSM.HtlcPart(outer.totalAmount, add)
-        receiving(htlcs :+ add, nextPayload, nextPacket, handler, triggered)
+        receiving(htlcs :+ add, nextPayload, nextPacket, handler)
       case WrappedMultiPartPaymentFailed(MultiPartPaymentFSM.MultiPartPaymentFailed(_, failure, parts)) =>
         context.log.warn("could not complete incoming multi-part payment (parts={} paidAmount={} failure={})", parts.size, parts.map(_.amount).sum, failure)
         Metrics.recordPaymentRelayFailed(failure.getClass.getSimpleName, Tags.RelayType.Trampoline)
@@ -204,23 +202,23 @@ class NodeRelay private(nodeParams: NodeParams,
             rejectPayment(upstream, Some(failure))
             stopping()
           case None =>
-            if (!triggered && nextPayload.isAsyncPayment && nodeParams.asyncPaymentsTimeout.isDefined) {
-              waitForTrigger(upstream, nextPayload, nextPacket, nodeParams.asyncPaymentsTimeout.get)
+            if (nextPayload.isAsyncPayment && nodeParams.features.hasFeature(Features.AsyncPaymentPrototype)) {
+              waitForTrigger(upstream, nextPayload, nextPacket)
             } else {
               doSend(upstream, nextPayload, nextPacket)
             }
         }
       case AsyncPaymentTrigger => context.log.debug("received async payment trigger while waiting to receive all incoming payment packets")
-        receiving(htlcs, nextPayload, nextPacket, handler, triggered = true)
+        receiving(htlcs, nextPayload, nextPacket, handler)
     }
 
-  private def waitForTrigger(upstream: Upstream.Trampoline, nextPayload: IntermediatePayload.NodeRelay.Standard, nextPacket: OnionRoutingPacket, asyncPaymentsTimeout: FiniteDuration): Behavior[Command] = {
-    context.log.info(s"waiting for async payment to trigger before relaying trampoline payment (amountIn=${upstream.amountIn} expiryIn=${upstream.expiryIn} amountOut=${nextPayload.amountToForward} expiryOut=${nextPayload.outgoingCltv}, asyncPaymentsTimeout=$asyncPaymentsTimeout)")
+  private def waitForTrigger(upstream: Upstream.Trampoline, nextPayload: IntermediatePayload.NodeRelay.Standard, nextPacket: OnionRoutingPacket): Behavior[Command] = {
+    context.log.info(s"waiting for async payment to trigger before relaying trampoline payment (amountIn=${upstream.amountIn} expiryIn=${upstream.expiryIn} amountOut=${nextPayload.amountToForward} expiryOut=${nextPayload.outgoingCltv}, timeout=${nodeParams.relayParams.timeout})")
     Behaviors.withTimers { timers =>
-      timers.startSingleTimer(AsyncPaymentTimeout, asyncPaymentsTimeout)
+      timers.startSingleTimer(AsyncPaymentTimeout, nodeParams.relayParams.timeout)
       Behaviors.receiveMessagePartial {
         case AsyncPaymentTimeout =>
-          context.log.warn(s"rejecting async payment that was not triggered before timeout: $asyncPaymentsTimeout")
+          context.log.warn(s"rejecting async payment that was not triggered before relay timeout: ${nodeParams.relayParams.timeout}")
           rejectPayment(upstream, Some(PaymentTimeout))
           stopping()
         case AsyncPaymentTrigger =>
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
@@ -43,6 +43,7 @@ import fr.acinq.eclair.{CltvExpiry, Features, Logs, MilliSatoshi, NodeParams, UI
 
 import java.util.UUID
 import scala.collection.immutable.Queue
+import scala.concurrent.duration.FiniteDuration
 
 /**
  * It [[NodeRelay]] aggregates incoming HTLCs (in case multi-part was used upstream) and then forwards the requested amount (using the
@@ -54,12 +55,14 @@ object NodeRelay {
   sealed trait Command
   case class Relay(nodeRelayPacket: IncomingPaymentPacket.NodeRelayPacket) extends Command
   case object Stop extends Command
+  case object AsyncPaymentTrigger extends Command
   private case class WrappedMultiPartExtraPaymentReceived(mppExtraReceived: MultiPartPaymentFSM.ExtraPaymentReceived[HtlcPart]) extends Command
   private case class WrappedMultiPartPaymentFailed(mppFailed: MultiPartPaymentFSM.MultiPartPaymentFailed) extends Command
   private case class WrappedMultiPartPaymentSucceeded(mppSucceeded: MultiPartPaymentFSM.MultiPartPaymentSucceeded) extends Command
   private case class WrappedPreimageReceived(preimageReceived: PreimageReceived) extends Command
   private case class WrappedPaymentSent(paymentSent: PaymentSent) extends Command
   private case class WrappedPaymentFailed(paymentFailed: PaymentFailed) extends Command
+  private case object AsyncPaymentTimeout extends Command
   // @formatter:on
 
   trait OutgoingPaymentFactory {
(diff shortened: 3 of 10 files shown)


remyers:
Some Questions ffor my next set of changes to this PR:

 - What information should go in the `AsyncPaymentFeatures` onion tlv?
 
 I currently use invoice feature bits because an empty tlv seemed wrong. The sender will look for the `AsyncPaymentPrototype` feature bit in the invoice, but the trampoline relay does not have access to the invoice (for privacy). That's why the sender must add a tlv specifically to the onion of the trampoline relay.
 
 -  When I change the `AsyncPaymentTrigger` message to be a non-spec `LightningMessage` that is routed to the `NodeRelayer`, should it be keyed off of the `nodeRelayPacket.outerPayload.paymentSecret` for the node relayer ?

 - Would it be better to publish an event when an `AsyncPaymentTrigger` non-spec `LightningMessage` is received, or route it to the set of running `NodeRelayer` actors ?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -54,12 +55,14 @@ object NodeRelay {
   sealed trait Command
   case class Relay(nodeRelayPacket: IncomingPaymentPacket.NodeRelayPacket) extends Command
   case object Stop extends Command
+  case object AsyncPaymentTrigger extends Command
```
t-bast, line 58: nit: I slightly prefer renaming it to something closer to the only other public command (`Relay`):
```suggestion
  case object RelayAsyncPayment extends Command
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -178,14 +181,15 @@ class NodeRelay private(nodeParams: NodeParams,
    * @param nextPayload relay instructions (should be identical across HTLCs in this set).
    * @param nextPacket  trampoline onion to relay to the next trampoline node.
    * @param handler     actor handling the aggregation of the incoming HTLC set.
+   * @param triggered   async payment trigger received
    */
-  private def receiving(htlcs: Queue[UpdateAddHtlc], nextPayload: IntermediatePayload.NodeRelay.Standard, nextPacket: OnionRoutingPacket, handler: ActorRef): Behavior[Command] =
+  private def receiving(htlcs: Queue[UpdateAddHtlc], nextPayload: IntermediatePayload.NodeRelay.Standard, nextPacket: OnionRoutingPacket, handler: ActorRef, triggered: Boolean = false): Behavior[Command] =
```
t-bast, line 186: I don't think we should handle the trigger here, this won't always work.
In this phase, we are receiving incoming htlcs before relaying.
Imagine we have the following sequence of events:

- we receive the first incoming htlc, but not the second one (so we're not ready to relay)
- the recipient comes online, so we receive the trigger
- the recipient goes back offline
- we receive the second incoming htlc

We will try to relay instantly whereas the recipient has gone offline.

I think the only change to this behavior should be to go to `waitForTrigger` instead of `doSend` when the payload says it's an async payment.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -199,10 +203,43 @@ class NodeRelay private(nodeParams: NodeParams,
...
+              waitForTrigger(upstream, nextPayload, nextPacket, nodeParams.asyncPaymentsTimeout.get)
+            } else {
+              doSend(upstream, nextPayload, nextPacket)
+            }
+        }
+      case AsyncPaymentTrigger => context.log.debug("received async payment trigger while waiting to receive all incoming payment packets")
+        receiving(htlcs, nextPayload, nextPacket, handler, triggered = true)
+    }
+
+  private def waitForTrigger(upstream: Upstream.Trampoline, nextPayload: IntermediatePayload.NodeRelay.Standard, nextPacket: OnionRoutingPacket, asyncPaymentsTimeout: FiniteDuration): Behavior[Command] = {
```
t-bast, line 217: There is another timeout we need to watch out for: the incoming cltv timeout. We don't want to let the incoming channel force-close because of an HTLC-timeout while we're waiting for the receiver to come back online!

So we should also register to the `CurrentBlockHeight` event, and if we're getting too close to the expiry of the incoming HTLCs, we should fail the payment.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -181,6 +181,9 @@ object OnionPaymentPayloadTlv {
 
   /** Pre-image included by the sender of a payment in case of a donation */
   case class KeySend(paymentPreimage: ByteVector32) extends OnionPaymentPayloadTlv
+
+  /** Invoice feature bits. Only included for intermediate trampoline nodes that should wait before forwarding this payment */
+  case class AsyncPaymentFeatures(features: ByteVector) extends OnionPaymentPayloadTlv
```
t-bast, line 186: We don't need anything in here, the features are unused and will take space in the onion for no good reason. I don't think there's anything wrong with an "empty" tlv, it just acts as some kind of flag. We should probably just keep this empty until we figure out if there is useful data that we should add in there:

```suggestion
  case object AsyncPayment extends OnionPaymentPayloadTlv
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -331,6 +336,17 @@ object PaymentOnion {
...
+
+        /** Create a standard trampoline inner payload instructing the trampoline node to wait for a trigger before sending an async payment. */
+        def createNodeRelayForAsyncPayment(amount: MilliSatoshi, expiry: CltvExpiry, nextNodeId: PublicKey, invoiceFeatures: Features[InvoiceFeature]): Standard = {
+          val tlvs = Seq(
+            Some(AmountToForward(amount)),
+            Some(OutgoingCltv(expiry)),
+            Some(OutgoingNodeId(nextNodeId)),
+            Some(AsyncPaymentFeatures(invoiceFeatures.toByteVector))
+          ).flatten
+          Standard(TlvStream(tlvs))
```
t-bast, line 348: You don't really need to use optional fields and then flatten them since they are all filled:

```suggestion
          Standard(TlvStream(AmountToForward(amount), OutgoingCltv(expiry), OutgoingNodeId(nextNodeId), AsyncPaymentFeatures(invoiceFeatures.toByteVector)))
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -490,6 +508,7 @@ object PaymentOnionCodecs {
     .typecase(UInt64(66099), invoiceRoutingInfo)
     .typecase(UInt64(66100), trampolineOnion)
     .typecase(UInt64(5482373484L), keySend)
+    .typecase(UInt64(181324718L), asyncPayment)
```
t-bast, line 511: nit: since we're ordering that list by increasing values for readability (when it will start getting big), maybe move that one up above the `keySend` one?

t-bast (commented):
Nice, you found your way around the right parts of the codebase, most of this looks correct!

> When I change the AsyncPaymentTrigger message to be a non-spec LightningMessage that is routed to the NodeRelayer, should it be keyed off of the nodeRelayPacket.outerPayload.paymentSecret for the node relayer ?

I don't think we need to define a `LightningMessage` for the `AsyncPaymentTrigger`.
This won't be something we receive from the recipient peer when it comes back online.
It will just be an internal component that will send this internal message to the event stream.

Let me know what you think, I think the flow should be the following:

- we should add a singleton actor named something like `AsyncPaymentWatcher`
- when the `NodeRelay` component goes to the `waitForTrigger` behavior, it should send a `NotifyWhenPeerOnline(replyTo, remoteNodeId, paymentHash)` to the `AsyncPaymentWatcher`
- the job of the `AsyncPaymentWatcher` will be to check when the peer comes back online, and when the peer is back online, to send `AsyncPaymentTrigger/RelayAsyncPayment` to the corresponding `replyTo` actor

What I don't know yet is if this should be a new actor or if we should enrich an existing actor with that new responsibility.

In the private feature branches that contain the logic for Phoenix stuff, we have a component that detects when a Phoenix peer comes back online. It

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
remyers: The codec system seems to expect a case class here, it seems to work with the change here: [dc69ae5](https://github.com/ACINQ/eclair/pull/2435/commits/dc69ae5e372aba64a2795b7e25a20b620e82e60d)

I would have done this before,  but didn't find any examples of empty tlvs. I think the `provide` codec function solved it for me though.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
remyers: That makes sense. Instead of dealing with a trigger that comes before the `waitForTrigger` state in `NodeRelay` it makes more sense that when `waitForTrigger` registers the trigger callback it will immediately get a trigger if the destination node is already connected, or later when the destination node connects.

remyers: This is fixed in [96acc3a](https://github.com/ACINQ/eclair/pull/2435/commits/96acc3aaff05cac3c81714c865b6745e5b70df3a)

eclair-core/src/main/resources/reference.conf
```diff
@@ -149,6 +149,11 @@ eclair {
       // Delay enforcement of channel fee updates
       enforcement-delay = 10 minutes
     }
+
+    async-payments {
+        enabled = false
```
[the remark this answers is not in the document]
remyers: fixed in [96acc3a](https://github.com/ACINQ/eclair/pull/2435/commits/96acc3aaff05cac3c81714c865b6745e5b70df3a).

I went with `relay.timeout`, but was wondering if perhaps it should be more explicitly called `relay.async-payment-timeout`. Any preference?

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -281,6 +281,12 @@ object Features {
     val mandatory = 148
   }
 
+  // TODO: @remyers update feature bits once spec-ed (currently reserved here: https://github.com/lightning/bolts/pull/989)
+  case object AsyncPaymentPrototype extends Feature with InvoiceFeature {
+    val rfcName = "async_payment_prototype"
+    val mandatory = 52
```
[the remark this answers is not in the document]
remyers: I changed it to 152 which just adds 100 to feature bit that the BOLT issue seems to reserve. Fixed in [96acc3a](https://github.com/ACINQ/eclair/pull/2435/commits/96acc3aaff05cac3c81714c865b6745e5b70df3a).

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -315,7 +322,8 @@ object Features {
     AnchorOutputsZeroFeeHtlcTx -> (StaticRemoteKey :: Nil),
     RouteBlinding -> (VariableLengthOnion :: Nil),
     TrampolinePaymentPrototype -> (PaymentSecret :: Nil),
-    KeySend -> (VariableLengthOnion :: Nil)
+    KeySend -> (VariableLengthOnion :: Nil),
+    AsyncPaymentPrototype -> (VariableLengthOnion :: TrampolinePaymentPrototype :: Nil)
```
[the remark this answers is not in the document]
remyers: Fixed in [3a632cd](https://github.com/ACINQ/eclair/pull/2435/commits/3a632cd6c127100ec9cb3fe9a207d80f325ebfea).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
remyers: Good idea; fixed in [3a632cd](https://github.com/ACINQ/eclair/pull/2435/commits/3a632cd6c127100ec9cb3fe9a207d80f325ebfea).

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
remyers: fixed in [dc69ae5](https://github.com/ACINQ/eclair/pull/2435/commits/dc69ae5e372aba64a2795b7e25a20b620e82e60d)

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
t-bast: I would be more specific and name it `relay.async-payment-timeout`.
Otherwise it would conflict with types of "timeout".

For example, when we cannot relay a payment (because the next peer is offline or we don't have liquidity), we currently instantly fail the incoming htlc. But we could instead wait a bit and retry a few times before failing the incoming htlc. That feature would need its own timeout, which would be named differently.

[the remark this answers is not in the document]
t-bast: Since changes to the configuration file are a pain (we need to detect the old setting and warn users), better to get the naming right from the start ;)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -199,10 +201,43 @@ class NodeRelay private(nodeParams: NodeParams,
...
             stopping()
+          case None =>
+            if (nextPayload.isAsyncPayment && nodeParams.features.hasFeature(Features.AsyncPaymentPrototype)) {
+              waitForTrigger(upstream, nextPayload, nextPacket)
+            } else {
+              doSend(upstream, nextPayload, nextPacket)
+            }
+        }
+      case RelayAsyncPayment => context.log.debug("received async payment trigger while waiting to receive all incoming payment packets")
+        receiving(htlcs, nextPayload, nextPacket, handler)
```
t-bast, line 212: nit:
```suggestion
      case RelayAsyncPayment =>
        context.log.debug("received async payment trigger while waiting to receive all incoming payment packets")
        receiving(htlcs, nextPayload, nextPacket, handler)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -199,10 +201,43 @@ class NodeRelay private(nodeParams: NodeParams,
...
+    }
+
+  private def waitForTrigger(upstream: Upstream.Trampoline, nextPayload: IntermediatePayload.NodeRelay.Standard, nextPacket: OnionRoutingPacket): Behavior[Command] = {
+    context.log.info(s"waiting for async payment to trigger before relaying trampoline payment (amountIn=${upstream.amountIn} expiryIn=${upstream.expiryIn} amountOut=${nextPayload.amountToForward} expiryOut=${nextPayload.outgoingCltv}, timeout=${nodeParams.relayParams.timeout})")
+    Behaviors.withTimers { timers =>
+      timers.startSingleTimer(AsyncPaymentTimeout, nodeParams.relayParams.timeout)
+      Behaviors.receiveMessagePartial {
+        case AsyncPaymentTimeout =>
+          context.log.warn(s"rejecting async payment that was not triggered before relay timeout: ${nodeParams.relayParams.timeout}")
+          rejectPayment(upstream, Some(PaymentTimeout))
```
t-bast, line 222: We shouldn't use `PaymentTimeout`, this tell the sender that we didn't receive their incoming HTLCs, whereas the issue here is that the recipient is unavailable. Right now for Phoenix when this happens, we end up using `TemporaryNodeFailure`. This isn't ideal, but it's the best existing failure message that matches this. We should add a dedicated new failure code for this scenario, but I believe we should only do this for the final specification, for now `TemporaryNodeFailure` is a good enough choice.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -323,6 +327,131 @@ class NodeRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("appl
...
+    incomingAsyncPayment.foreach { p =>
+      val fwd = register.expectMessageType[Register.Forward[CMD_FAIL_HTLC]]
+      assert(fwd.channelId == p.add.channelId)
+      assert(fwd.message == CMD_FAIL_HTLC(p.add.id, Right(PaymentTimeout), commit = true))
+    }
+
+    register.expectNoMessage(100 millis)
+  }
+
+  test("fail async payment when only triggered before last incoming part received") { f =>
```
t-bast, line 346: You could use a test tag for all the async payment tests:
```suggestion
  test("fail async payment when only triggered before last incoming part received", Tag("async_payment")) { f =>
```

And when that test tag is present, in `withFixture` you would set the feature bit, so that you don't have to do it in every test.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -323,6 +327,131 @@ class NodeRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("appl
...
+    nodeRelayerAdapters ! createSuccessEvent()
+    val relayEvent = eventListener.expectMessageType[TrampolinePaymentRelayed]
+    validateRelayEvent(relayEvent)
+    assert(relayEvent.incoming.toSet == incomingAsyncPayment.map(i => PaymentRelayed.Part(i.add.amountMsat, i.add.channelId)).toSet)
+    assert(relayEvent.outgoing.nonEmpty)
+    parent.expectMessageType[NodeRelayer.RelayComplete]
+    register.expectNoMessage(100 millis)
+  }
+
+  test("fail to relay async payment when triggered if the outgoing expiry is not above chain height") { f =>
```
t-bast, line 425: We should add one more test that verifies that when the async payment feature is turned off, we instantly relay payments (this is what guarantees that it's safe to merge on `master` since it's off by default).

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -723,26 +852,32 @@ object NodeRelayerSpec {
...
     createValidIncomingPacket(1000000 msat, incomingAmount, CltvExpiry(499999), outgoingAmount, outgoingExpiry)
   )
   val incomingSinglePart = createValidIncomingPacket(incomingAmount, incomingAmount, CltvExpiry(500000), outgoingAmount, outgoingExpiry)
+  val incomingAsyncPayment: Seq[NodeRelayPacket] = incomingMultiPart.map(p => p.copy(innerPayload = IntermediatePayload.NodeRelay.Standard.createNodeRelayForAsyncPayment(p.innerPayload.amountToForward, p.innerPayload.outgoingCltv, outgoingNodeId)))
 
   def createSuccessEvent(): PaymentSent =
     PaymentSent(relayId, paymentHash, paymentPreimage, outgoingAmount, outgoingNodeId, Seq(PaymentSent.PartialPayment(UUID.randomUUID(), outgoingAmount, 10 msat, randomBytes32(), None)))
 
-  def createValidIncomingPacket(amountIn: MilliSatoshi, totalAmountIn: MilliSatoshi, expiryIn: CltvExpiry, amountOut: MilliSatoshi, expiryOut: CltvExpiry): IncomingPaymentPacket.NodeRelayPacket = {
+  def createValidIncomingPacket(amountIn: MilliSatoshi, totalAmountIn: MilliSatoshi, expiryIn: CltvExpiry, amountOut: MilliSatoshi, expiryOut: CltvExpiry, isAsyncPayment: Boolean = false): IncomingPaymentPacket.NodeRelayPacket = {
```
t-bast, line 866: It looks like you're not using `isAsyncPayment` anywhere...You should probably either revert the changes to that method or use it when creating your `incomingAsyncPayment` above (instead of modifying the `incomingMultiPart` onces), whatever you prefer!

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
t-bast: Yes, this is completely ok, inside tlv stream we really don't care, so this value is fine.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
t-bast: More details about how close to the expiry we should fail. We really want to avoid seeing force-closed channels with the sender caused by an offline recipient. So we'll probably need to take a very large buffer because once we decide we want to fail the payment, we need to wait for the sender to come back online to actually sign the `update_fail_htlc` (and that needs to happen before the incoming payment expires). We reasonably need to allow at least 144 blocks for the sender to come back online, what do you think?

We should probably add another configurable value in `reference.conf` to control that: `relay.async-payment-cancel-safety-before-timeout-blocks = 144`? Then in the `waitForTrigger` state, at each block we would check if `currentBlockHeight > upstream.expiryIn - asyncPaymentCancelSafetyBeforeTimeout`, and if so, we'd cancel the payment.

Now that we have two configuration values, it makes sense to group them in a dedicated object in `reference.conf`, and create a corresponding case class inside `RelayParams`:

```
relay {
  async-payments {
    timeout = 7 days
    async-payment-cancel-safety-before-timeout-blocks = 144
  }
}
```

When reading the configuration, we should also verify that the `timeout` is at least twice the cancel safety, otherwise the feature will be quite useless.

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
t-bast: Update: before renaming this, it's worth investigating this comment: https://github.com/ACINQ/eclair/pull/2435#discussion_r980290178

If we create a dedicated config section, the name `timeout` is good enough.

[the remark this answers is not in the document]
remyers: Good point - it seems like there is room to expand on this. A new `async-payments` section (AsyncPaymentsParams case class) within `relay` (and RelayParams) is justified.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
remyers: This makes sense to me. This value is similar to `channel.expiry-delta-blocks` for an immediately forwarded payment. If a relay forwards a payment, then it should have claimed or failed the downstream payment somehow (either on or off chain) `channel.expiry-delta-blocks` before the upstream payment expires. For waiting payments that have not been forwarded yet at that same point, then a relay should initiate a payment failure with the same (or more) lead time.

When reading the configuration, should we also fail if `async-payment-cancel-safety-before-timeout-blocks` < `channel.expiry-delta-blocks`?  Looking at NodeRelay.scala, it seems like it would be strictly unsafe to forward a payment after the `payloadOut.outgoingCltv` block which must be at least `nodeParams.channelConf.expiryDelta` sooner than `upstream.expiryIn` according to the check in `validateRelay`.

The `timeout` parameter prevents our node from holding onto many unclaimed payments with really long `upstream.expiryIn` values. Perhaps another way to define `timeout` would be to set a cap on the longest `upstream.expiryIn` we're willing to wait for? That way everything would be in block time. On the other hand, `timeout` is a nice human readable way of configuring how long to hold async-payments.

t-bast: > When reading the configuration, should we also fail if async-payment-cancel-safety-before-timeout-blocks < channel.expiry-delta-blocks? 

Yes, that's a good idea!

>  Perhaps another way to define timeout would be to set a cap on the longest upstream.expiryIn we're willing to wait for?

The issue with that is that we don't have a good mechanism to communicate this to the sender. I don't think it would be necessary to be honest, as it would be a poor UX to go back and forth with the user to tell them to set a smaller delay, that's something that should be hidden from them as much as possible.

I think we should let node operators choose that `timeout` regardless of what `expiryIn` they receive. I don't know if that `timeout` should be in blocks or time, I think both would work? Maybe using blocks is a bit more consistent?

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -57,11 +59,11 @@ class NodeRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("appl
 
   import NodeRelayerSpec._
 
-  case class FixtureParam(nodeParams: NodeParams, router: TestProbe[Any], register: TestProbe[Any], mockPayFSM: TestProbe[Any], eventListener: TestProbe[PaymentEvent]) {
+  case class FixtureParam(nodeParams: NodeParams, router: TestProbe[Any], register: TestProbe[Any], mockPayFSM: TestProbe[Any], eventListener: TestProbe[PaymentEvent], relayMonitor: TestProbe[NodeRelay.Command], blockHeight: AtomicLong) {
```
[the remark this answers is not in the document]
remyers: I would like a way to detect when the `MultiPartPaymentFSM.MultiPartPaymentSucceeded` message has been received by the `NodeRelay` actor so I can test the `waitForTrigger` state. I used `relayMonitor` to wait for each `Relay` message, and then wait for one more (private) message, which is `MultiPartPaymentFSM.MultiPartPaymentSucceeded`. After we're in that state, it's safe to do my tests of the timeouts, or send the payment trigger.

Can you suggest a better way to wait for that internal state change?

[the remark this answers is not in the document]
t-bast: I think there is a better way indeed. In our discussions, we identified that whenever we go to the `waitForTrigger` state, we will send a message to some actor somewhere to tell them to notify us when the peer is ready. We could _partially_ implement that today, by simply sending to the event stream a `WaitingToRelayPayment(remoteNodeId, paymentHash)` event when we enter the `waitForTrigger` state (and later we will change that to be sent to a specific actor).

Your tests can register to that event on the event stream and make sure they receive it (indicating that we are now in the `waitForTrigger` state) before testing timeout / trigger. How does that sound?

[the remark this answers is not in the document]
remyers: Great idea! and a good place for putting a TODO: note too.

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -323,6 +327,131 @@ class NodeRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("appl
     register.expectNoMessage(100 millis)
   }
 
+  test("fail async payment when relay is not triggered") { f =>
+    val g = f.copy(nodeParams = f.nodeParams.copy(features = f.nodeParams.features.add(AsyncPaymentPrototype, Optional), relayParams = f.nodeParams.relayParams.copy(timeout = 5 seconds)))
```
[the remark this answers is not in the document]
remyers: Change to using blockheight deltas instead of clocktime timeouts fixes this in [fcdc394](https://github.com/ACINQ/eclair/pull/2435/commits/fcdc394cb2751d7408318afb437cf49d18389d08)

eclair-core/src/main/resources/reference.conf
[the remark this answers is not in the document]
remyers: Replaced clock based timeout in the `relay` section with a new `async-payments` section and two block delta timeouts:
```
    async-payments {
      // Maximum number of blocks to hold an async payment while waiting to receive a trigger from the receiver
      hold-timeout-blocks = 1008
      // Number of blocks before the incoming HTLC expires that an async payment must be triggered by the receiver
      cancel-safety-before-timeout-blocks = 144
    }
```
 fixed in [fcdc394](https://github.com/ACINQ/eclair/pull/2435/commits/fcdc394cb2751d7408318afb437cf49d18389d08)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
remyers: I went with replacing `timeout` with `hold-timeout-blocks`. The `NodeRelay` actor will only wait for this many blocks (default: 1008 or ~ 7 days) before timing out the relay.

I also added the `cancel-safety-before-timeout-blocks` parameter and will timeout that many blocks (default: 144)  before an incoming payment is set to expire.

Fixed in [fcdc394](https://github.com/ACINQ/eclair/pull/2435/commits/fcdc394cb2751d7408318afb437cf49d18389d08)

eclair-core/src/main/scala/fr/acinq/eclair/Features.scala
```diff
@@ -281,6 +281,12 @@ object Features {
     val mandatory = 148
   }
 
+  // TODO: @remyers update feature bits once spec-ed (currently reserved here: https://github.com/lightning/bolts/pull/989)
+  case object AsyncPaymentPrototype extends Feature with InvoiceFeature {
```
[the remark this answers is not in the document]
remyers: Fixed in [a48a322](https://github.com/ACINQ/eclair/pull/2435/commits/a48a3226d73f275db1f7eda711640856b18d9fd9), and understand we'll not add it yet to `node_announcement`

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/Relayer.scala
```diff
@@ -129,7 +129,8 @@ object Relayer extends Logging {
   case class RelayParams(publicChannelFees: RelayFees,
                          privateChannelFees: RelayFees,
                          minTrampolineFees: RelayFees,
-                         enforcementDelay: FiniteDuration) {
+                         enforcementDelay: FiniteDuration,
+                         timeout: FiniteDuration) {
```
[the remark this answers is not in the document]
remyers: Obsoleted by change to block height deltas.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala
```diff
@@ -475,6 +485,8 @@ object PaymentOnionCodecs {
 
   private val keySend: Codec[KeySend] = variableSizeBytesLong(varintoverflow, bytes32).as[KeySend]
 
+  private val asyncPayment: Codec[AsyncPayment] = provide(AsyncPayment())
```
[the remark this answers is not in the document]
remyers: Fixed in [8acc728](https://github.com/ACINQ/eclair/pull/2435/commits/8acc728e8257b614be162d1f0640df270555ed3b).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
remyers: Actually, I think we can remove this case entirely. As you pointed out earlier, the trigger won't be returned until we register for it in the `waitForTrigger` state. Fixed in  [6400ce2](https://github.com/ACINQ/eclair/pull/2435/commits/6400ce2070e4aab21fc08801c1afe44aaa2d7747)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -199,10 +201,43 @@ class NodeRelay private(nodeParams: NodeParams,
...
+          stopping()
+        case RelayAsyncPayment =>
+          // check cltv timeouts again before forwarding the payment
+          validateRelay(nodeParams, upstream, nextPayload) match {
+          case Some(failure) =>
```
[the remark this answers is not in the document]
remyers: That's all reworked now as of [fcdc394](https://github.com/ACINQ/eclair/pull/2435/commits/fcdc394cb2751d7408318afb437cf49d18389d08) and I looked for other places I might have forgotten too.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
remyers: fixed in [fcdc394](https://github.com/ACINQ/eclair/pull/2435/commits/fcdc394cb2751d7408318afb437cf49d18389d08).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -199,10 +201,43 @@ class NodeRelay private(nodeParams: NodeParams,
...
+            context.log.warn(s"rejecting async payment, reason=$failure")
+            rejectPayment(upstream, Some(failure))
+            stopping()
           case None =>
             doSend(upstream, nextPayload, nextPacket)
         }
+        case Stop =>
+          context.log.warn(s"async payment stopped while waiting for trigger")
+          rejectPayment(upstream, Some(PaymentTimeout))
+          stopping()
```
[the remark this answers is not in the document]
remyers: I meant to handle the situation here where the sender cancels an async payment. They can't do it unilaterally and will need a way to signal the trampoline node.  I added the `CancelAsyncPayment` message and tests in [5fc3b0f](https://github.com/ACINQ/eclair/pull/2435/commits/5fc3b0f57dd3d02d34912a8fd541ecbc80a6d8a1)

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
remyers: Great tip! I updated the tests in [fcdc394](https://github.com/ACINQ/eclair/pull/2435/commits/fcdc394cb2751d7408318afb437cf49d18389d08).

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
remyers: Fixed in [36b0d42](https://github.com/ACINQ/eclair/pull/2435/commits/36b0d421cf820431a4ac9a8bbe54209cf9f5bdea)

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
remyers: Removed it in [fcdc394](https://github.com/ACINQ/eclair/pull/2435/commits/fcdc394cb2751d7408318afb437cf49d18389d08)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -200,10 +204,43 @@ class NodeRelay private(nodeParams: NodeParams,
...
+    context.system.eventStream ! EventStream.Publish(WaitingToRelayPayment(nextPayload.outgoingNodeId, paymentHash))
+    Behaviors.receiveMessagePartial {
+      case WrappedCurrentBlockHeight(blockHeight) if blockHeight >= safetyBlock =>
+        context.log.warn(s"rejecting async payment at block $blockHeight; was not triggered ${nodeParams.relayParams.asyncPaymentsParams.cancelSafetyBeforeTimeout} safety blocks before upstream cltv expiry at ${upstream.expiryIn}")
+        rejectPayment(upstream, Some(TemporaryNodeFailure)) // TODO: replace failure type when async payment spec is finalized
+        stopping()
+      case WrappedCurrentBlockHeight(blockHeight) if blockHeight >= timeoutBlock =>
+        context.log.warn(s"rejecting async payment at block $blockHeight; was not triggered after waiting ${nodeParams.relayParams.asyncPaymentsParams.holdTimeoutBlocks} blocks")
+        rejectPayment(upstream, Some(TemporaryNodeFailure)) // TODO: replace failure type when async payment spec is finalized
+        stopping()
```
t-bast, line 234: We should handle all `WrappedCurrentBlockHeight` cases, otherwise we'll have a warning in the logs at every block saying we didn't handle `WrappedCurrentBlockHeight`, which will be annoying. Just add the following:

```scala
      case WrappedCurrentBlockHeight(blockHeight) => 
        Behaviors.same
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
```diff
@@ -323,6 +329,146 @@ class NodeRelayerSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("appl
...
+  }
+
+  test("fail to relay when not triggered before the incoming expiry safety timeout", Tag("async_payments"), Tag("long_hold_timeout")) { f =>
+    import f._
+
+    val (nodeRelayer, _) = createNodeRelay(incomingAsyncPayment.head)
+    incomingAsyncPayment.foreach(p => nodeRelayer ! NodeRelay.Relay(p))
+    mockPayFSM.expectNoMessage(100 millis) // we should NOT trigger a downstream payment before we received a complete upstream payment
+
+    // publish block height one block before the cancel-safety-before-timeout-block interval before the current incoming payment expiry
```
t-bast, line 404: nit: comment is incorrect?

t-bast (commented):
LGTM, a last couple of nits and we should be good to go :rocket:

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/NodeRelayerSpec.scala
remyers: Copy/paste error fixed in [f670676](https://github.com/ACINQ/eclair/pull/2435/commits/f670676221f8cfacc5e5c489f2f69fcc203110a3)

t-bast (approved):
LGTM! Don't forget to use github's "squash and merge" option, this is good to go :+1:

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
remyers: Fixed in [4c0b4d2](https://github.com/ACINQ/eclair/pull/2435/commits/4c0b4d281e1a5386cb516cd2cf59b9a56bbdab99).

Am I correct in assuming that once we pass out of the `waitForTrigger` state, that the next state won't get any more `WrappedCurrentBlockHeight` messages because `messageAdapter` that they were being sent to will have gone out of scope?

t-bast: That's a good comment, I'm not sure...this is worth testing and potentially fixing before merging this PR, good catch! Otherwise I think there is a way to tell the event stream that we "unregister".

t-bast: There is an `EventStream.Unsubscribe` message available, it's probably a good idea to send that when moving out of the waiting state (no need to test it though).

PR #2440, opened 2022-09-29
Fix `availableForSend` fuzz tests

t-bast:
These two fuzz tests setup a random set of HTLCs and then try to send or receive the maximum available amount. The initial HTLC setup may fail, if the initial balances are too low.

It is hard to set those initial balances to ensure this will always work, but since this will only rarely randomly happen, we should simply ignore it (instead of failing the test).


remyers (approved):
LGTM! since the focus of the test is around the `availableBalanceForSend` functionality, it makes sense to ignore a failure during initialization. 

Did something change recently that would cause the initialization step to fail more often? It seems odd I'd be seeing it more now. Just unlucky?

t-bast:
> Did something change recently that would cause the initialization step to fail more often? It seems odd I'd be seeing it more now. Just unlucky?

No, nothing changed that would impact that, you've just been unlucky :smiley:

PR #2441, opened 2022-09-29
Add support for arbitrary length onion errors

### LINKED ISSUE #2438: Add support for onion errors larger than 256 bytes
When receiving an onion error, the spec says that we should verify it's no larger than 256 bytes.
But this is just a SHOULD, not a MUST, and it could be useful in the future to make this larger.
We currently don't support lengths greater than 256 byte, but this should be easy to change now on the reader side without creating any backwards-compatibility issue, and we will allow us to introduce bigger errors in the future once everyone supports reading them.

See https://github.com/lightning/bolts/pull/1021 for more context.

t-bast:
The specification recommends using a length of 256 for onion errors, but it doesn't say that we should reject errors that use a different length.

We may want to start creating errors with a bigger length than 256 if we need to transmit more data to the sender. In order to prepare for this, we keep creating 256-bytes onion errors, but allow receiving errors of arbitrary length.

See the specification here: https://github.com/lightning/bolts/pull/1021

Fixes #2438


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/FailureMessage.scala
```diff
@@ -139,19 +140,27 @@ object FailureMessageCodecs {
     }, (_: FailureMessage).code)
   )
 
+  private def failureOnionPayload(payloadAndPadLength: Int): Codec[FailureMessage] = Codec(
+    // We create payloads of the requested size: by always using the same size we ensure error messages are indistinguishable.
```
pm47, line 144: That comment should be on method `failureOnionCodec`, it's the one which has the hardcoded default length.

eclair-core/src/test/scala/fr/acinq/eclair/channel/states/e/NormalStateSpec.scala
```diff
@@ -1925,16 +1925,16 @@ class NormalStateSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with
...
     val (_, htlc) = addHtlc(50000000 msat, alice, bob, alice2bob, bob2alice)
     crossSign(alice, bob, alice2bob, bob2alice)
     // Bob receives a failure with a completely invalid onion error (missing mac)
-    bob ! CMD_FAIL_HTLC(htlc.id, Left(ByteVector.fill(260)(42)))
+    bob ! CMD_FAIL_HTLC(htlc.id, Left(ByteVector.fill(561)(42)))
     val fail = bob2alice.expectMsgType[UpdateFailHtlc]
     assert(fail.id == htlc.id)
-    // We should rectify the packet length before forwarding upstream.
-    assert(fail.reason.length == Sphinx.FailurePacket.PacketLength)
+    // We propagate failure upstream (hopefully the sender knows how to unwrap them).
```
pm47, line 1936: Is there a risk of compatibility issues on mainnet if we deploy?

t-bast: It shouldn't since cln and lnd said they already do this.
On the contrary, we create compatibility issues today by not following the spec and truncating those errors.
However nobody actually creates errors that aren't 256 bytes long yet, so it doesn't matter much.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/FailureMessage.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2441/commits/5dd46b5847e3d980190da2e9715e23825cf107c6

pm47 (approved):

PR #2452, opened 2022-10-07
Complete codec for TLV fields

thomash-acinq:
We currently accept some malformed TLVs with additional data that we ignore. This means that decoding and reencoding may give a different result. With this change, we now reject such TLVs.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/TlvCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/TlvCodecs.scala
@@ -107,10 +107,10 @@ object TlvCodecs {
   }
 
   /** Codec for a tlv field that contains the field length and its value, without its tag. */
-  def tlvField[T <: Tlv](valueCodec: Codec[T]): Codec[T] = variableSizeBytesLong(varintoverflow, valueCodec.complete)
+  def tlvField[T <: Tlv, A](valueCodec: Codec[A])(implicit as: scodec.Transformer[A, T]): Codec[T] = variableSizeBytesLong(varintoverflow, valueCodec.as[T].complete)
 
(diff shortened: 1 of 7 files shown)


t-bast:
Can you also update this codec to use `tlvField`: https://github.com/ACINQ/eclair/blob/06ead3cf28c299825e87bc579f0b0d3146891462/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/PaymentOnion.scala#L487

It was added in parallel with the `tlvField` but didn't create a merge conflict, so I missed it.

t-bast (approved):
LGTM :+1:

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/TlvCodecs.scala
```diff
@@ -107,10 +107,10 @@ object TlvCodecs {
   }
 
   /** Codec for a tlv field that contains the field length and its value, without its tag. */
-  def tlvField[T <: Tlv](valueCodec: Codec[T]): Codec[T] = variableSizeBytesLong(varintoverflow, valueCodec)
+  def tlvField[T <: Tlv, A](valueCodec: Codec[A])(implicit as: scodec.Transformer[A, T]): Codec[T] = variableSizeBytesLong(varintoverflow, valueCodec.as[T].complete)
```
thomash-acinq, line 110: with the `.as[T]` here, we can remove all the useless `.as[]` everywhere else

thomash-acinq:
Since we're touching `tlvField`, I've put the `.as[T]` in there too so that we no longer need all the redundant `.as[T]` in the codec definition.
I've updated all the codecs I could find, except for the offer codecs since they need to be completely changed to match the latest offer spec and touching the file now would just create more conflicts.

t-bast (approved):
Nice!

Issue #2453, opened 2022-10-08
Add some blockchain.fee tests and minor refactoring


barrybecker4:
As part of my learning and getting familiar with the codebase I would like to add some additional feerate tests and perhaps do some minor refactoring. I asked some preliminary questions in [this discussion](https://github.com/ACINQ/eclair/discussions/2446).

To keep the scope very small, here is what I propose in my initial change:
- Use case classes for the feerate classes because they are immutable.
- Add a test to BitcoinCoreFeeProviderSpec to verify that an error with the correct message occurs when we try to get fee rates when no fee rate is found.
- Add a test spec for ConstantFeeProvider where none currently exists.

barrybecker4:
One thing that I noticed while testing is that `feePerBlock(0)` for [FeeratesPerKB](https://github.com/ACINQ/eclair/blob/06ead3cf28c299825e87bc579f0b0d3146891462/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala#L112) returns an incorrect value. It returns `blocks_6` instead of either an error or `mempoolMinFee`. Would it be better to add `require(target > 0)` to feePerBlock or return membpooMinFee if target <= 0? I suspect that having a target <= 0 doesn't make sense, so probably the require is better.

t-bast:
Fixed by #2454

PR #2454, opened 2022-10-08
add some blockchain.fee tests (#2453)

barrybecker4:
- Use case classes for the feerate classes because they are immutable.
 - Add a test to BitcoinCoreFeeProviderSpec to verify that an error with the correct message occurs when we try to get fee rates when no fee rate is found.
 - Add a test spec for ConstantFeeProvider where none currently exists.


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
```diff
@@ -52,6 +53,7 @@ case class FeeratePerKB(feerate: Satoshi) extends Ordered[FeeratePerKB] {
 object FeeratePerKB {
   // @formatter:off
   def apply(feeratePerByte: FeeratePerByte): FeeratePerKB = FeeratePerKB(feeratePerByte.feerate * 1000)
+  // feerate-per-kw should be feerate-per-kvb / 4
```
t-bast, line 56: When the code is trivial like here, this comment is really unnecessary, can you remove it?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
```diff
@@ -21,13 +21,14 @@ import fr.acinq.bitcoin.scalacompat.{Satoshi, SatoshiLong}
...
 /**
+ * Used to fetch feerates.
  * Created by PM on 09/07/2017.
  */
 trait FeeProvider {
   def getFeerates: Future[FeeratesPerKB]
 }
 
-case object CannotRetrieveFeerates extends RuntimeException("cannot retrieve feerates, channels may be at risk: ensure bitcoind estimatesmartfee correctly returns feerates and restart eclair")
+case object CannotRetrieveFeerates extends RuntimeException("Cannot retrieve feerates. Channels may be at risk. Ensure bitcoind estimatesmartfee correctly returns feerates and restart eclair")
```
t-bast, line 31: I don't see a compelling reason to change that?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
```diff
@@ -95,6 +97,9 @@ object FeeratePerKw {
    * hence feerate-per-kw >= 253
    *
    * See also https://github.com/ElementsProject/lightning/pull/1251
+   * Notes:
+   *  - The size of a block is limited to 4,000,000 weight units which corresponds to 1,000,000 virtual bytes.
+   *  - The transaction fee doesn't depend on the amount being transferred. It depends on the amount of data being transferred.
```
t-bast, line 102: This is basic bitcoin trivia, and this repository isn't meant to be a learning resource for bitcoin, please remove it.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
```diff
@@ -105,9 +110,14 @@ object FeeratePerKw {
...
 /**
- * Fee rates in satoshi-per-kilo-bytes (1 kb = 1000 bytes).
+ * Fee rates in satoshis-per-kilo-bytes (1 kb = 1000 bytes).
  * The mempoolMinFee is the minimal fee required for a tx to enter the mempool (and then be relayed to other nodes and eventually get confirmed).
  * If our fee provider doesn't expose this data, using its biggest block target should be a good enough estimation.
+ *
+ * The mempool contains pending transactions. Fewer pending transactions means lower fees.
+ * The block targets of 1, 2, 6, 12, 36, 72, 144, 1008 roughly correspond to
+ * time intervals of 10 minutes, 20 minutes, 1 hour, 2 hours, 6 hours, 12 hours, 1 day, 1 week respectively.
+ * These are the time intervals eclair expects the transaction to take to be included into a block.
```
t-bast, line 120: Same, please remove this comment, it's really unnecessary. It's basic bitcoin trivia that the target block time is 10 minutes, so people can compute those themselves if they're interested.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/fee/BitcoinCoreFeeProviderSpec.scala
```diff
@@ -90,7 +85,33 @@ class BitcoinCoreFeeProviderSpec extends TestKitBaseClass with BitcoindService w
...
 
-    val mockBitcoinClient = new BasicBitcoinJsonRPCClient(rpcAuthMethod = UserPassword("", ""), host = "localhost", port = 0) {
+    val mockBitcoinClient = createMockBitcoinClient(fees, ref.mempoolMinFee)
+
+    val mockProvider = BitcoinCoreFeeProvider(mockBitcoinClient, FeeratesPerKB(FeeratePerKB(1 sat), FeeratePerKB(1 sat), FeeratePerKB(2 sat), FeeratePerKB(3 sat), FeeratePerKB(4 sat), FeeratePerKB(5 sat), FeeratePerKB(6 sat), FeeratePerKB(7 sat), FeeratePerKB(8 sat)))
+    mockProvider.getFeerates.pipeTo(sender.ref)
+    assert(sender.expectMsgType[FeeratesPerKB] == ref)
+  }
+
+  test("get fee rates when no fee rate is found") {
```
t-bast, line 95: I don't see the point of this additional test to be honest. It doesn't test a behavior we actually rely on and will break whenever `bitcoind` changes their error message...

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/fee/ConstantFeeProviderSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+package fr.acinq.eclair.blockchain.fee
+
+import fr.acinq.bitcoin.scalacompat.SatoshiLong
+import org.scalatest.funsuite.AnyFunSuite
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.concurrent.Await
+
+class ConstantFeeProviderSpec extends AnyFunSuite {
```
t-bast, line 26: Adding unit tests for such trivial code is wasteful IMO, we don't need more lines of code if they don't add meaningful value...

t-bast (commented):
I guess switching to `case class` makes sense, but I'm really not convinced by the additional tests nor the bitcoin trivia comments added.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
barrybecker4: OK. Removed it.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
barrybecker4: I reverted it, but I don't understand why you wouldn't want the error message to have correct punctuation and capitalization.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
barrybecker4: OK. Removed.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
barrybecker4: OK. Removed. The correspondence between blocks and time wasn't obvious to me, but I am a newbie.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/fee/BitcoinCoreFeeProviderSpec.scala
barrybecker4: OK. I removed it.

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/fee/ConstantFeeProviderSpec.scala
barrybecker4: OK. I removed the ConstantFeeProviderSpec.

barrybecker4:
Thanks for the review. I made the suggested changes.

One thing that I noticed while testing is that feePerBlock(0) for [FeeratesPerKB](https://github.com/ACINQ/eclair/blob/06ead3cf28c299825e87bc579f0b0d3146891462/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala#L112) returns an incorrect value. It returns blocks_6 instead of either an error or mempoolMinFee. Would it be better to add require(target > 0) to feePerBlock or return mempooMinFee if target <= 0? I suspect that having a target <= 0 doesn't make sense, so probably the require is better.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
t-bast: Because those exception messages usually get included in logs in the middle of sentences so shouldn't be capitalized otherwise the punctuation is then incorrect, and the current punctuation isn't blatantly incorrect?

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
t-bast: The comment hasn't been removed in the last version of the PR?

t-bast (commented):
> One thing that I noticed while testing is that feePerBlock(0) for FeeratesPerKB returns an incorrect value. It returns blocks_6 instead of either an error or mempoolMinFee. Would it be better to add require(target > 0) to feePerBlock or return mempooMinFee if target <= 0?

Right, using a `target <= 0` doesn't make sense, you can add a `require(target > 0)`. You should then add the same `require` on the configuration values that are used when calling this function, so that a misconfigured `eclair.conf` fails at start-up: in `FeeTargets`, all `Int` fields must be strictly positive.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/FeeProvider.scala
barrybecker4: Sorry - removed it this time.

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/fee/BlockThresholds.scala
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fr.acinq.eclair.blockchain.fee
+
+object BlockThresholds {
```
t-bast, line 19: This is unused and unnecessary, remove?

barrybecker4: Removed it.

t-bast (approved):

PR #2455, opened 2022-10-11
Add tlv stream to onion failures

t-bast:
Extend every onion failure with an optional tlv stream, as specified here: https://github.com/lightning/bolts/pull/1021

I have verified cross-compatibility with lnd (https://github.com/lightningnetwork/lnd/pull/6967) in both directions (reading and writing a tlv stream) and the spec PR has been merged, so this is ready to integrate.


eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/CommandCodecs.scala
```diff
@@ -19,14 +19,53 @@ package fr.acinq.eclair.wire.internal
...
+import fr.acinq.eclair.wire.protocol._
+import fr.acinq.eclair.{BlockHeight, MilliSatoshiLong}
 import scodec.Codec
 import scodec.codecs._
 
 import scala.concurrent.duration.FiniteDuration
 
 object CommandCodecs {
 
+  // A trailing tlv stream was added in https://github.com/lightning/bolts/pull/1021 which wasn't handled properly by
```
pm47, line 33: Damn! How about we just use a generic hardcoded error message? We should be able to work around the absence of length delimitation, because trailing fields are constant-size.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/PaymentInitiator.scala
```diff
@@ -136,7 +136,7 @@ class PaymentInitiator(nodeParams: NodeParams, outgoingPaymentFactory: PaymentIn
       case pp: PendingTrampolinePayment =>
         val trampolineHop = NodeHop(pp.r.trampolineNodeId, pp.r.recipientNodeId, pp.r.trampolineAttempts.last._2, pp.r.trampolineAttempts.last._1)
         val decryptedFailures = pf.failures.collect { case RemoteFailure(_, _, Sphinx.DecryptedFailurePacket(_, f)) => f }
-        val shouldRetry = decryptedFailures.contains(TrampolineFeeInsufficient) || decryptedFailures.contains(TrampolineExpiryTooSoon)
+        val shouldRetry = decryptedFailures.contains(TrampolineFeeInsufficient()) || decryptedFailures.contains(TrampolineExpiryTooSoon())
```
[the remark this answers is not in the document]
t-bast: Good catch! Fixed in https://github.com/ACINQ/eclair/pull/2455/commits/2d46a9f9c1afa8408a58db0581cba64526583a01

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/CommandCodecs.scala
t-bast: I'm not sure what you mean...we really need to introduce a new codec anyway that adds a length-prefix to failure messages, otherwise we won't be able to properly encode/decode failure messages that contain tlvs. I don't see how we could introduce that while keeping compatibility with the existing codec?

Another option could be to update the codec and do a DB migration. But that would also require having the legacy codec around (in the DB migration code), it doesn't remove it...

t-bast: Right, I didn't understand your earlier comment, that's a good idea to simplify the code!

t-bast: Done in https://github.com/ACINQ/eclair/pull/2455/commits/5caaadc4bd5a4efac8c18b63fd61bf1d60a4f54a

PR #2456, opened 2022-10-13
Fix pruned channels coming back online

### LINKED ISSUE #2388: Channels and nodes keep getting pruned then downloaded again
Running the latest master branch (after https://github.com/ACINQ/eclair/pull/2380) with a channel to horizon (which does not yet have https://github.com/ACINQ/eclair/pull/2380 afaik)  I keep seeing in the logs nodes and channels getting pruned. 
For example (happens with about 12000 channels):

```
2022-08-18 00:53:36,519 INFO  f.a.eclair.router.Router - pruning shortChannelId=528181x1103x0 (stale)
2022-08-18 00:54:59,142 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - channel shortChannelId=528181x1103x0 is back from the dead! requesting announcements about this channel
2022-08-18 01:02:08,070 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - validating shortChannelId=528181x1103x0
2022-08-18 01:53:33,874 INFO  f.a.eclair.router.Router - pruning shortChannelId=528181x1103x0 (stale)
2022-08-18 01:55:39,803 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - channel shortChannelId=528181x1103x0 is back from the dead! requesting announcements about this channel
2022-08-18 02:02:58,126 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - validating shortChannelId=528181x1103x0
2022-08-18 02:53:33,716 INFO  f.a.eclair.router.Router - pruning shortChannelId=528181x1103x0 (stale)
2022-08-18 02:55:15,422 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - channel shortChannelId=528181x1103x0 is back from the dead! requesting announcements about this channel
2022-08-18 03:02:16,040 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - validating shortChannelId=528181x1103x0
2022-08-18 03:53:30,260 INFO  f.a.eclair.router.Router - pruning shortChannelId=528181x1103x0 (stale)
2022-08-18 03:54:54,782 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - channel shortChannelId=528181x1103x0 is back from the dead! requesting announcements about this channel
2022-08-18 04:01:50,285 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - validating shortChannelId=528181x1103x0
2022-08-18 04:53:32,269 INFO  f.a.eclair.router.Router - pruning shortChannelId=528181x1103x0 (stale)
2022-08-18 04:55:46,839 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - channel shortChannelId=528181x1103x0 is back from the dead! requesting announcements about this channel
2022-08-18 05:02:56,789 INFO  f.a.eclair.router.Router SYN n:03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f - validating shortChannelId=528181x1103x0
```

and:

```
2022-08-18 00:54:49,762 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 01:55:28,783 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 02:55:05,713 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 03:54:43,609 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 04:55:33,217 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 05:55:04,920 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 06:54:50,142 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 07:55:53,136 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 08:55:09,046 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 09:54:52,352 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 10:55:42,531 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 11:55:13,588 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 12:54:51,964 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 13:56:04,738 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 14:55:10,676 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 15:55:01,797 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
2022-08-18 16:57:33,840 INFO  f.a.eclair.router.Router - pruning nodeId=02009947c197575f5a948e1e4343c41dc2e6122a9bd644629afb919f30e1115ff8 (stale)
```

t-bast:
When a channel was pruned and we receive a valid channel update for it, we need to wait until we're sure both sides of the channel are back online: a channel that has only one side available will most likely not be able to relay payments.

When both sides have created fresh, valid channel updates, we restore the channel to our channels map and to the graph.

We remove the custom `pruned` table and instead directly use the data from the channels table.

Fixes #2388

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
@@ -27,7 +27,7 @@ import fr.acinq.eclair.{RealShortChannelId, ShortChannelId}
 import grizzled.slf4j.Logging
 import scodec.bits.BitVector
 
-import java.sql.{Connection, Statement}
+import java.sql.{Connection, ResultSet, Statement}
 import javax.sql.DataSource
 import scala.collection.immutable.SortedMap
 
@@ -70,7 +70,6 @@ class PgNetworkDb(implicit ds: DataSource) extends NetworkDb with Logging {
           statement.executeUpdate("CREATE SCHEMA network")
           statement.executeUpdate("CREATE TABLE network.nodes (node_id TEXT NOT NULL PRIMARY KEY, data BYTEA NOT NULL, json JSONB NOT NULL)")
           statement.executeUpdate("CREATE TABLE network.public_channels (short_channel_id BIGINT NOT NULL PRIMARY KEY, txid TEXT NOT NULL, channel_announcement BYTEA NOT NULL, capacity_sat BIGINT NOT NULL, channel_update_1 BYTEA NULL, channel_update_2 BYTEA NULL, channel_announcement_json JSONB NOT NULL, channel_update_1_json JSONB NULL, channel_update_2_json JSONB NULL)")
-          statement.executeUpdate("CREATE TABLE network.pruned_channels (short_channel_id BIGINT NOT NULL PRIMARY KEY)")
         case Some(v@(2 | 3)) =>
           logger.warn(s"migrating db $DB_NAME, found version=$v current=$CURRENT_VERSION")
           if (v < 3) {
@@ -87,7 +86,9 @@ class PgNetworkDb(implicit ds: DataSource) extends NetworkDb with Logging {
             val validChannelUpdate2 = rs.getBitVectorOpt("channel_update_2").forall(channelUpdateCodec.decode(_).isSuccessful)
             (shortChannelId, validChannelUpdate1 && validChannelUpdate2)
           }).collect {
-            case (scid, false) => statement.executeUpdate(s"DELETE FROM network.public_channels WHERE short_channel_id=$scid")
+            case (scid, false) =>
+              logger.warn(s"removing channel update with scid=$scid from the network DB (update cannot be decoded)")
+              statement.executeUpdate(s"DELETE FROM network.public_channels WHERE short_channel_id=$scid")
           }
         case Some(unknownVersion) => throw new RuntimeException(s"Unknown version of DB $DB_NAME found, version=$unknownVersion")
       }
@@ -208,17 +209,31 @@ class PgNetworkDb(implicit ds: DataSource) extends NetworkDb with Logging {
     }
   }
 
+  private def parseChannel(rs: ResultSet): PublicChannel = {
+    val ann = channelAnnouncementCodec.decode(rs.getBitVectorOpt("channel_announcement").get).require.value
+    val txId = ByteVector32.fromValidHex(rs.getString("txid"))
+    val capacity = rs.getLong("capacity_sat")
+    val channel_update_1_opt = rs.getBitVectorOpt("channel_update_1").map(channelUpdateCodec.decode(_).require.value)
+    val channel_update_2_opt = rs.getBitVectorOpt("channel_update_2").map(channelUpdateCodec.decode(_).require.value)
+    PublicChannel(ann, txId, Satoshi(capacity), channel_update_1_opt, channel_update_2_opt, None)
+  }
+
+  override def getChannel(shortChannelId: RealShortChannelId): Option[PublicChannel] = withMetrics("network/get-channel", DbBackends.Postgres) {
+    inTransaction { pg =>
+      using(pg.prepareStatement("SELECT channel_announcement, txid, capacity_sat, channel_update_1, channel_update_2 FROM network.public_channels WHERE short_channel_id=?")) { statement =>
+        statement.setLong(1, shortChannelId.toLong)
+        statement.executeQuery().map(parseChannel).headOption
+      }
+    }
+  }
+
   override def listChannels(): SortedMap[RealShortChannelId, PublicChannel] = withMetrics("network/list-channels", DbBackends.Postgres) {
     inTransaction { pg =>
       using(pg.createStatement()) { statement =>
         statement.executeQuery("SELECT channel_announcement, txid, capacity_sat, channel_update_1, channel_update_2 FROM network.public_channels")
           .foldLeft(SortedMap.empty[RealShortChannelId, PublicChannel]) { (m, rs) =>
-            val ann = channelAnnouncementCodec.decode(rs.getBitVectorOpt("channel_announcement").get).require.value
-            val txId = ByteVector32.fromValidHex(rs.getString("txid"))
-            val capacity = rs.getLong("capacity_sat")
-            val channel_update_1_opt = rs.getBitVectorOpt("channel_update_1").map(channelUpdateCodec.decode(_).require.value)
-            val channel_update_2_opt = rs.getBitVectorOpt("channel_update_2").map(channelUpdateCodec.decode(_).require.value)
-            m + (ann.shortChannelId -> PublicChannel(ann, txId, Satoshi(capacity), channel_update_1_opt, channel_update_2_opt, None))
+            val channel = parseChannel(rs)
+            m + (channel.shortChannelId -> channel)
           }
       }
     }
@@ -246,35 +261,4 @@ class PgNetworkDb(implicit ds: DataSource) extends NetworkDb with Logging {
     }
   }
 
-  override def addToPruned(shortChannelIds: Iterable[RealShortChannelId]): Unit = withMetrics("network/add-to-pruned", DbBackends.Postgres) {
-    inTransaction { pg =>
-      using(pg.prepareStatement("INSERT INTO network.pruned_channels VALUES (?) ON CONFLICT DO NOTHING")) {
-        statement =>
-          shortChannelIds.foreach(shortChannelId => {
-            statement.setLong(1, shortChannelId.toLong)
-            statement.addBatch()
-          })
-          statement.executeBatch()
-      }
-    }
-  }
-
-  override def removeFromPruned(shortChannelId: RealShortChannelId): Unit = withMetrics("network/remove-from-pruned", DbBackends.Postgres) {
-    inTransaction { pg =>
-      using(pg.prepareStatement(s"DELETE FROM network.pruned_channels WHERE short_channel_id=?")) {
-        statement =>
-          statement.setLong(1, shortChannelId.toLong)
-          statement.executeUpdate()
-      }
-    }
-  }
-
-  override def isPruned(shortChannelId: ShortChannelId): Boolean = withMetrics("network/is-pruned", DbBackends.Postgres) {
-    inTransaction { pg =>
-      using(pg.prepareStatement("SELECT short_channel_id from network.pruned_channels WHERE short_channel_id=?")) { statement =>
-        statement.setLong(1, shortChannelId.toLong)
-        statement.executeQuery().nonEmpty
-      }
-    }
-  }
 }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -71,25 +71,24 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
 
   {
     log.info("loading network announcements from db...")
-    val channels = db.listChannels()
-    val nodes = db.listNodes()
+    val (pruned, channels) = db.listChannels().partition { case (_, pc) => pc.isStale(nodeParams.currentBlockHeight) }
+    val nodes = db.listNodes().map(n => n.nodeId -> n).toMap
     Metrics.Nodes.withoutTags().update(nodes.size)
     Metrics.Channels.withoutTags().update(channels.size)
     log.info("loaded from db: channels={} nodes={}", channels.size, nodes.size)
-    val initChannels = channels
+    log.info("{} pruned channels at blockHeight={}", pruned.size, nodeParams.currentBlockHeight)
     // this will be used to calculate routes
-    val graph = DirectedGraph.makeGraph(initChannels)
-    val initNodes = nodes.map(n => n.nodeId -> n).toMap
+    val graph = DirectedGraph.makeGraph(channels)
     // send events for remaining channels/nodes
-    context.system.eventStream.publish(ChannelsDiscovered(initChannels.values.map(pc => SingleChannelDiscovered(pc.ann, pc.capacity, pc.update_1_opt, pc.update_2_opt))))
-    context.system.eventStream.publish(ChannelUpdatesReceived(initChannels.values.flatMap(pc => pc.update_1_opt ++ pc.update_2_opt ++ Nil)))
-    context.system.eventStream.publish(NodesDiscovered(initNodes.values))
+    context.system.eventStream.publish(ChannelsDiscovered(channels.values.map(pc => SingleChannelDiscovered(pc.ann, pc.capacity, pc.update_1_opt, pc.update_2_opt))))
+    context.system.eventStream.publish(ChannelUpdatesReceived(channels.values.flatMap(pc => pc.update_1_opt ++ pc.update_2_opt ++ Nil)))
+    context.system.eventStream.publish(NodesDiscovered(nodes.values))
 
     // watch the funding tx of all these channels
     // note: some of them may already have been spent, in that case we will receive the watch event immediately
-    initChannels.values.foreach { pc =>
+    (channels.values ++ pruned.values).foreach { pc =>
       val txid = pc.fundingTxid
-      val TxCoordinates(_, _, outputIndex) = ShortChannelId.coordinates(pc.ann.shortChannelId)
+      val outputIndex = ShortChannelId.coordinates(pc.ann.shortChannelId).outputIndex
       // avoid herd effect at startup because watch-spent are intensive in terms of rpc calls to bitcoind
       context.system.scheduler.scheduleOnce(Random.nextLong(nodeParams.routerConf.watchSpentWindow.toSeconds).seconds) {
         watcher ! WatchExternalChannelSpent(self, txid, outputIndex, pc.ann.shortChannelId)
@@ -104,7 +103,7 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
     log.info(s"initialization completed, ready to process messages")
     Try(initialized.map(_.success(Done)))
     val data = Data(
-      initNodes, initChannels,
+      nodes, channels, pruned,
       Stash(Map.empty, Map.empty),
       rebroadcast = Rebroadcast(channels = Map.empty, updates = Map.empty, nodes = Map.empty),
       awaiting = Map.empty,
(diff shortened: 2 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -414,51 +414,55 @@ object Validation {
...
+                log.debug("channel shortChannelId={} is still pruned", u.shortChannelId)
+                sendDecision(origins, GossipDecision.RelatedChannelPruned(u))
+                db.updateChannel(u)
+                d
+              } else {
+                log.info("channel shortChannelId={} is back from the dead", u.shortChannelId)
+                sendDecision(origins, GossipDecision.Accepted(u))
+                val latestUpdates = pc1.update_1_opt ++ pc1.update_2_opt
+                ctx.system.eventStream.publish(ChannelUpdatesReceived(latestUpdates))
+                db.updateChannel(u)
```
pm47, line 451: Should you have instead called `addPublicChannel` (like we do for local channels that graduate from private to public) and `handleChannelUpdate` (like we do in `handleChannelValidationResponse` for stashed updates)? It guarantees that we do the exact same processing for announcements whatever the context (first seen or pruned). Btw, the behavior is changing compared to before: we may send a `ChannelUpdatesReceived` without a `ChannelDiscovered` depending on the pruning time and the node restart time.

Looking deeper into it, at first I thought that adding back the channel manually as opposed to calling `addPublicChannel` made us miss putting back the `WatchExternalChannelSpent`, but... it's already there because we watch _all_ (live and stale) channels on restart. _However_, the `handleChannelSpent` only checks the live channels map so there is indeed a bug (and probably a missing test).

There is another issue: not emitting `ChannelDiscovered` will not work in a clustered setup, because front routers rely on this event to find out about new channels and they will never be made aware of unpruned channels coming back to life.

pm47 (changes requested):
It's looking good, but there remaining issues.

eclair-core/src/main/scala/fr/acinq/eclair/router/StaleChannels.scala
```diff
@@ -19,46 +19,35 @@ package fr.acinq.eclair.router
...
+import fr.acinq.eclair.router.Router.{ChannelDesc, Data, hasChannels}
 import fr.acinq.eclair.wire.protocol.{ChannelAnnouncement, ChannelUpdate}
 import fr.acinq.eclair.{BlockHeight, ShortChannelId, TimestampSecond, TxCoordinates}
 
-import scala.collection.mutable
 import scala.concurrent.duration._
 
 object StaleChannels {
 
   def handlePruneStaleChannels(d: Data, db: NetworkDb, currentBlockHeight: BlockHeight)(implicit ctx: ActorContext, log: LoggingAdapter): Data = {
```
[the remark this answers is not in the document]
t-bast: I think it would be overkill to prune at every block, pruning at a regular interval is probably better.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -414,51 +414,55 @@ object Validation {
...
-                // otherwise we send the query right away
-                peerConnection ! query
-                d.copy(sync = d.sync + (remoteNodeId -> Syncing(remainingQueries = Nil, totalQueries = 1)))
-            }
-          case _ =>
-            // we don't know which node this update came from (maybe it was stashed and the channel got pruned in the meantime or some other corner case).
-            // or we don't have a peerConnection to send our query to.
-            // anyway, that's not really a big deal because we have removed the channel from the pruned db so next time it shows up we will revalidate it
-            d
+          db.getChannel(RealShortChannelId(u.shortChannelId.toLong)) match {
```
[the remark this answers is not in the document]
t-bast: We could indeed keep it in memory, especially since we initially load it, we just need to make sure it stays in sync with the DB. I'll try it.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -414,51 +414,55 @@ object Validation {
...
+                log.warning("bad signature for announcement shortChannelId={} {}", u.shortChannelId, u)
+                sendDecision(origins, GossipDecision.InvalidSignature(u))
+                d
+              } else if (pc1.isStale(currentBlockHeight)) {
+                log.debug("channel shortChannelId={} is still pruned", u.shortChannelId)
+                sendDecision(origins, GossipDecision.RelatedChannelPruned(u))
+                db.updateChannel(u)
+                d
+              } else {
+                log.info("channel shortChannelId={} is back from the dead", u.shortChannelId)
```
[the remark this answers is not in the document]
t-bast: We could, but do we really care? It's extra code and I'm not sure it brings a lot of value...

eclair-core/src/main/scala/fr/acinq/eclair/router/StaleChannels.scala
```diff
@@ -98,6 +87,4 @@ object StaleChannels {
     channelIsOldEnough && channelUpdateIsStale
```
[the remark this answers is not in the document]
t-bast: This is incorrect, this would return `false` when we don't have any channel update, which should count as a stale update. Fortunately, we have a unit test that catches this!

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -71,25 +71,24 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
 
   {
     log.info("loading network announcements from db...")
-    val channels = db.listChannels()
+    val (pruned, channels) = db.listChannels().partition { case (_, pc) => pc.isStale(nodeParams.currentBlockHeight) }
```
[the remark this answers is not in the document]
t-bast: Added a log in https://github.com/ACINQ/eclair/pull/2456/commits/e048636104d35d8e12b5b01a94e20886d24dbcad

I still kept them partitioned, as it's clearer now that we also keep pruned data in the router state and less foot-gunny, let me know what you think.

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -71,25 +71,24 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
...
     val nodes = db.listNodes()
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2456/commits/e048636104d35d8e12b5b01a94e20886d24dbcad, I don't know why we had those `init` variables, probably a left-over from previous code.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -60,7 +60,7 @@ object Validation {
       // adding the sender to the list of origins so that we don't send back the same announcement to this peer later
       val origins = d.awaiting(c) :+ origin
       d.copy(awaiting = d.awaiting + (c -> origins))
-    } else if (db.isPruned(c.shortChannelId)) {
+    } else if (db.getChannel(c.shortChannelId).nonEmpty) {
```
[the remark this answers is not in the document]
t-bast: Fixed in https://github.com/ACINQ/eclair/pull/2456/commits/e048636104d35d8e12b5b01a94e20886d24dbcad where we look into our `prunedChannels` map instead.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2456/commits/e048636104d35d8e12b5b01a94e20886d24dbcad, this is more robust.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -414,51 +414,55 @@ object Validation {
...
+          db.getChannel(RealShortChannelId(u.shortChannelId.toLong)) match {
+            case Some(pc) =>
+              // The channel was pruned (one of the two channel updates was older than 2 weeks).
+              val pc1 = pc.applyChannelUpdate(update)
+              if (!Announcements.checkSig(u, pc.getNodeIdSameSideAs(u))) {
+                log.warning("bad signature for announcement shortChannelId={} {}", u.shortChannelId, u)
+                sendDecision(origins, GossipDecision.InvalidSignature(u))
+                d
+              } else if (pc1.isStale(currentBlockHeight)) {
+                log.debug("channel shortChannelId={} is still pruned", u.shortChannelId)
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/ACINQ/eclair/pull/2456/commits/e048636104d35d8e12b5b01a94e20886d24dbcad

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
t-bast: Partially done in https://github.com/ACINQ/eclair/pull/2456/commits/e048636104d35d8e12b5b01a94e20886d24dbcad for the channel updates.

I didn't call `addPublicChannel` though, as we don't want to revalidate the on-chain outpoint, that would be another DoS vector wouldn't it? Also we can't be in a private -> public transition if a channel has been pruned, it must have been public before. So it feels like the only part from `addPublicChannel` that we actually need is the `ChannelsDiscovered` event, so I imported only that, but let me know if you think we should refactor `addPublicChannel` to be able to call it here.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -414,51 +415,58 @@ object Validation {
...
+              } else {
+                log.info("channel shortChannelId={} is back from the dead", u.shortChannelId)
+                // We notify front nodes that the channel is back.
+                ctx.system.eventStream.publish(ChannelsDiscovered(SingleChannelDiscovered(pc1.ann, pc1.capacity, pc1.update_1_opt, pc1.update_2_opt) :: Nil))
+                // We move the channel out of the pruned map and re-validate updates as if we were receiving them for the first time.
+                val d1 = d.copy(channels = d.channels + (pc.shortChannelId -> pc.copy(update_1_opt = None, update_2_opt = None)), prunedChannels = d.prunedChannels - pc.shortChannelId)
+                (pc1.update_1_opt ++ pc1.update_2_opt).foldLeft(d1) {
+                  case (current, chanUpdate) =>
+                    if (chanUpdate.channelFlags.isNode1 == u.channelFlags.isNode1) {
+                      handleChannelUpdate(current, db, currentBlockHeight, update)
```
sstone, line 458: It feels strange to see `handleChannelUpdate` call itself here ? Wouldn't be safer (and easier to parse) to extract code that needs to be used here and call it explicitly ?

sstone: I don't feel too strongly about this but `handleChannelUpdate` has many side-effects and it's a bit difficult to reason about what happens when it calls itself again, but I also see why it's the simplest way to not forget to apply one of these side-effects.

sstone: But it should be possible to extract what should be done in each case move it out of the match/case block ?

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
t-bast: You're right! But it does need a lot of tweaks to actually be used in our context:

- we don't want to waste a call to `db.addChannel`
- we don't want to rebroadcast the channel announcement to our peers
- there's no need to put the watch, it was already put at start-up
- all the private -> public code is unnecessary

Overall it feels like we only want a very small subset of what `addPublicChannel` does, that's why I'm a bit reluctant, it feels harder to reason about than copy-pasting the useful code...I tried refactoring it to extract the useful part, but it makes it actually even harder to read and check for correctness :/

pm47: Fair enough :cowboy_hat_face:

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
t-bast: I removed usage of the existing `addPublicChannel` and `handleChannelUpdate` methods that had side-effects in https://github.com/ACINQ/eclair/pull/2456/commits/47f8f97e9558bc947733f7aa2aa71d805c0ce45f

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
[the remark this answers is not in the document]
pm47: It could help to get a quick assessment of whether our pruning logic works. But you are right it's not trivial, nevermind.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -414,51 +415,63 @@ object Validation {
...
+                db.updateChannel(u)
+                val prunedChannels1 = d.prunedChannels + (pc1.shortChannelId -> pc1)
+                d.copy(prunedChannels = prunedChannels1)
+              } else {
+                log.info("channel shortChannelId={} is back from the dead", u.shortChannelId)
+                // We notify front nodes that the channel is back.
+                ctx.system.eventStream.publish(ChannelsDiscovered(SingleChannelDiscovered(pc1.ann, pc1.capacity, pc1.update_1_opt, pc1.update_2_opt) :: Nil))
+                sendDecision(origins, GossipDecision.Accepted(u))
+                db.updateChannel(u)
+                // We rebroadcast both updates: both of them are valid now that the channel isn't stale anymore.
```
pm47, line 455: Rebroadcasting both updates is a bit dubious, and order matters in any case: you want to broadcast the most recent first.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
```diff
@@ -27,7 +27,7 @@ import fr.acinq.eclair.{RealShortChannelId, ShortChannelId}
 import grizzled.slf4j.Logging
 import scodec.bits.BitVector
 
-import java.sql.{Connection, Statement}
+import java.sql.{Connection, ResultSet, Statement}
```
pm47, line 30: In theory we should update the db version and delete the old `network.pruned_channels` update I'm afraid

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -358,7 +357,9 @@ object Router {
 
     val nodeId1: PublicKey = ann.nodeId1
     val nodeId2: PublicKey = ann.nodeId2
-    def shortChannelId: RealShortChannelId = ann.shortChannelId
+    val shortChannelId: RealShortChannelId = ann.shortChannelId
```
pm47, line 360: 👀

eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
```diff
@@ -214,12 +213,12 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
...
 
     case Event(PeerRoutingMessage(peerConnection, remoteNodeId, c: ChannelAnnouncement), d) =>
-      stay() using Validation.handleChannelAnnouncement(d, nodeParams.db.network, watcher, RemoteGossip(peerConnection, remoteNodeId), c)
+      stay() using Validation.handleChannelAnnouncement(d, watcher, RemoteGossip(peerConnection, remoteNodeId), c)
 
     case Event(r: ValidateResult, d) =>
       stay() using Validation.handleChannelValidationResponse(d, nodeParams, watcher, r)
 
-    case Event(WatchExternalChannelSpentTriggered(shortChannelId), d) if d.channels.contains(shortChannelId) =>
+    case Event(WatchExternalChannelSpentTriggered(shortChannelId), d) if d.channels.contains(shortChannelId) || d.prunedChannels.contains(shortChannelId) =>
```
pm47, line 221: 👀

pm47 (approved):
Actually, cleaning up the network db table can wait.

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
t-bast: I think it's necessary if we want to be maximally useful to our peers.
If for example update B -> A is stale.
Whenever we receive an update for the other side (A -> B), we won't rebroadcast them to our peers, so at some point the update they have for A -> B may also become stale.

Once an update for B -> A comes and the channel isn't stale anymore, if we only forward that one to our peers, they may have a stale A -> B update, so they wouldn't restore the channel. We really need to send both to avoid this case.

I don't understand why order would matter? Both channel updates are not stale and for a different channel side, that's all that matters.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgNetworkDb.scala
t-bast: I hesitated, it doesn't hurt to keep the unused table around though and remove it once we have a real DB update to make, to minimize the number of version changes?

sstone:
utACK https://github.com/ACINQ/eclair/pull/2456/commits/47f8f97e9558bc947733f7aa2aa71d805c0ce45f

sstone (approved):
tACK https://github.com/ACINQ/eclair/commit/47f8f97e9558bc947733f7aa2aa71d805c0ce45f

PR #2458, opened 2022-10-18
Fix flaky CommitmentsSpec test

thomash-acinq:
In case feeRatePerKw is high and liquidity is low on the initiator side, the initiator can't send anything and the test would fail because we try to create a HTLC for an amount of 0.


t-bast (approved):
Good catch, thanks!

PR #2459, opened 2022-10-18
Remove onion message size limit

thomash-acinq:
We used to drop onion messages above a certain size, but the onion message packet is already limited to 65536 bytes so we only keep this larger limit instead.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
@@ -28,6 +28,7 @@ import scodec.{Attempt, DecodeResult}
 
 import scala.annotation.tailrec
 import scala.concurrent.duration.FiniteDuration
+import scala.util.Try
 
 object OnionMessages {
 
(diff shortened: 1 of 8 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/integration/MessageIntegrationSpec.scala
```diff
@@ -142,6 +143,39 @@ class MessageIntegrationSpec extends IntegrationSpec {
...
+    val bytes = randomBytes(65290)
+    val encodedBytes = genericTlv.encode(GenericTlv(UInt64(135), bytes)).require.bytes
+
+    val alice = new EclairImpl(nodes("A"))
+
+    val probe = TestProbe()
+    val eventListener = TestProbe()
+    nodes("C").system.eventStream.subscribe(eventListener.ref, classOf[OnionMessages.ReceiveMessage])
+    alice.sendOnionMessage(nodes("B").nodeParams.nodeId :: Nil, Left(nodes("C").nodeParams.nodeId), None, encodedBytes).pipeTo(probe.ref)
+    assert(probe.expectMsgType[SendOnionMessageResponse].sent)
```
t-bast, line 174: It's surprising that we actually send that message, shouldn't it overflow at creation time?
It feels like `alice.sendOnionMessage` should fail here without sending an outgoing message, shouldn't it?

eclair-core/src/test/scala/fr/acinq/eclair/message/OnionMessagesSpec.scala
```diff
@@ -216,4 +216,32 @@ class OnionMessagesSpec extends AnyFunSuite {
       case x => fail(x.toString)
     }
   }
+
+  test("very large multi-hop onion message") {
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/integration/MessageIntegrationSpec.scala
thomash-acinq: Changed it so that it fails loudly

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -94,6 +95,9 @@ object OnionMessages {
       1300
     } else if (payloadSize <= 32768) {
       32768
+    } else if (payloadSize > 65432) {
+      // A payload of size 65432 corresponds to a total lightning message size of 65535.
+      throw new Exception(s"Message is too large: payloadSize=$payloadSize")
```
t-bast, line 100: Note that this value will become incorrect once we start adding tlvs to the onion message's tlv stream. When we add tlvs to onion messages, we'll need to make sure we don't forget to update this code (we'll need to subtract the encoded length of the tlv stream).

t-bast (approved):

PR #2460, opened 2022-10-19
Use 0-conf based on local features only

t-bast:
If we have activated 0-conf support for a given peer, we send our `channel_ready` early regardless of whether our peer has activated support for 0-conf. If they also immediately send their `channel_ready` it's great, if they don't it's ok, we'll just wait for confirmations, but it was worth trying.


pm47 (approved):
LGTM, I don't see much disadvantages compared to the whitelist approach, except that the configuration is a bit more involved.

PR #2461, opened 2022-10-19
Add tlv to require confirmed inputs for dual funding

t-bast:
We add a tlv to `open_channel2` and `accept_channel2` to require our peer to only use confirmed inputs, as proposed [here](https://github.com/lightning/bolts/pull/851#discussion_r994954281). Using only confirmed inputs ensures that we don't end up paying the fees of one of our peer's unconfirmed ancestors.

This can be configured at two levels:

- globally using `eclair.conf`
- on a per-channel basis by overriding the default value in `Peer.OpenChannel`

This PR contains a breaking change in the codecs: since dual funding isn't released yet, it's ok to make such a breaking change now: this is why I'd like to integrate this PR now, even though the final value of the TLV isn't specified yet. Updating the TLV can be done later without breaking compatibility.

This PR is probably best reviewed commit-by-commit.

### DIFF
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -99,7 +99,7 @@ eclair {
     min-public-funding-satoshis = 100000
     min-private-funding-satoshis = 100000
     max-funding-satoshis = 16777215 // to open channels larger than 16777215 you must enable the large_channel_support feature in 'eclair.features'
-    require-confirmed-inputs-for-funding = false // if true, only confirmed inputs will be accepted for channel funding
+    require-confirmed-inputs-for-dual-funding = false // if true, only confirmed inputs will be accepted for channel dual funding
 
     to-remote-delay-blocks = 720 // number of blocks that the other node's to-self outputs must be delayed (720 ~ 5 days)
     max-to-local-delay-blocks = 2016 // maximum number of blocks that we are ready to accept for our own delayed outputs (2016 ~ 2 weeks)
--- a/eclair-core/src/main/resources/reference.conf
+++ b/eclair-core/src/main/resources/reference.conf
@@ -99,6 +99,7 @@ eclair {
     min-public-funding-satoshis = 100000
     min-private-funding-satoshis = 100000
     max-funding-satoshis = 16777215 // to open channels larger than 16777215 you must enable the large_channel_support feature in 'eclair.features'
+    require-confirmed-inputs-for-funding = false // if true, only confirmed inputs will be accepted for channel funding
 
     to-remote-delay-blocks = 720 // number of blocks that the other node's to-self outputs must be delayed (720 ~ 5 days)
     max-to-local-delay-blocks = 2016 // maximum number of blocks that we are ready to accept for our own delayed outputs (2016 ~ 2 weeks)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
@@ -442,17 +442,25 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
   private def getInputDetails(txIn: TxIn, currentInputs: Seq[TxAddInput]): Future[Either[UnusableInput, TxAddInput]] = {
     currentInputs.find(i => txIn.outPoint == toOutPoint(i)) match {
       case Some(previousInput) => Future.successful(Right(previousInput))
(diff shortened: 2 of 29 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
```diff
@@ -440,17 +442,25 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
...
-          Left(UnusableInput(txIn.outPoint))
-        } else if (!Script.isNativeWitnessScript(previousTx.txOut(txIn.outPoint.index.toInt).publicKeyScript)) {
-          // Wallet input must be a native segwit input.
-          Left(UnusableInput(txIn.outPoint))
-        } else {
-          Right(TxAddInput(fundingParams.channelId, UInt64(0), previousTx, txIn.outPoint.index, txIn.sequence))
+      case None =>
+        val inputTxDetails = for {
+          tx <- wallet.getTransaction(txIn.outPoint.txid)
+          confirmations <- if (fundingParams.requireConfirmedInputs.local) wallet.getTxConfirmations(txIn.outPoint.txid) else Future.successful(None)
```
pm47, line 448: `getrawtransaction` with `debug=true` returns transaction details + confirmations, but it's not currently supported in `BitcoinCoreClient` and probably not very important to optimize.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/ChannelTlv.scala
```diff
@@ -84,6 +88,7 @@ object OpenDualFundedChannelTlv {
   val openTlvCodec: Codec[TlvStream[OpenDualFundedChannelTlv]] = tlvStream(discriminated[OpenDualFundedChannelTlv].by(varint)
     .typecase(UInt64(0), upfrontShutdownScriptCodec)
     .typecase(UInt64(1), channelTypeCodec)
+    .typecase(UInt64(0x40000001), requireConfirmedInputsCodec)
```
pm47, line 91: What's the rationale behind this choice of tag?

eclair-core/src/main/resources/reference.conf
```diff
@@ -99,6 +99,7 @@ eclair {
     min-public-funding-satoshis = 100000
     min-private-funding-satoshis = 100000
     max-funding-satoshis = 16777215 // to open channels larger than 16777215 you must enable the large_channel_support feature in 'eclair.features'
+    require-confirmed-inputs-for-funding = false // if true, only confirmed inputs will be accepted for channel funding
```
pm47, line 102: Nits:
- we should at some point group under a `funding` section
- 
```suggestion
    require-confirmed-inputs-for-dual-funding = false // if true, only confirmed inputs will be accepted for dual-funding
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
```diff
@@ -98,6 +98,8 @@ object InteractiveTxBuilder {
   case class RemoteFailure(cause: ChannelException) extends Failed
   // @formatter:on
 
+  case class RequireConfirmedInputs(local: Boolean, remote: Boolean)
```
pm47, line 101: This is a bit ambiguous, in the same way that for `toSelfDelay`. Is `local` _requiring_ or _requested_ to use confirmed inputs? It's _requested_.

Maybe:
```suggestion
  case class RequireConfirmedInputs(forLocal: Boolean, forRemote: Boolean)
```
or:
```suggestion
  case class RequireConfirmedInputsFor(local: Boolean, remote: Boolean)
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/ChannelTlv.scala
t-bast: Completely random, hopefully we agree with CLN on a standard value for that TLV in the next couple of weeks and can fix this before the next eclair release!

eclair-core/src/main/resources/reference.conf
t-bast: Good idea to create a `funding` section. Since we already have other fields to move to that new function section, I'll do that in a follow-up PR.

t-bast: Renamed in https://github.com/ACINQ/eclair/pull/2461/commits/68bfcae48a269d6ded0c5966a06167e4984cdc31

pm47 (approved):

PR #2462, opened 2022-10-20
dual-fund: the txid coming in over the wire is the 'hash'

niftynei:
The txid is sent as the "hash"; same as the serialized rawtx version in a bitcoin tx.

Eclair was anticipating the txid (non-reversed).

Note that this behavior mirrors how FundingCreated works currently (txid on the wire is the hash'd/rawtx version)

Raw message sent by CLN:
type: 0047
channel_id: 808d9608e894de3fddea3c77afe65b0f73c9b70e725ba826250c46e653e8d8a7 txid: e3f11b4d2581b32fb902ef3c1940f43b5eaa24458d83a5956ddf55b87a833979 tlv: 0000

Parsed into "human readable" fields
WIRE_TX_SIGNATURES:
channel_id=808d9608e894de3fddea3c77afe65b0f73c9b70e725ba826250c46e653e8d8a7 txid=7939837ab855df6d95a5838d4524aa5e3bf440193cef02b92fb381254d1bf1e3 witness_stack=[]

error:
invalid funding signature: tx=0200000000010178c993c1113e9b46f18deb9895f3223046b19fa6762531f9fab1083510aa6efb0000000000fdffffff02a08601000000000022002051fdf3a4a4041fe59d585013f936d75825d3d5733ca178b9a03357e78fec646fba00f4050000000016001455b3e205a3a5bf7d68cfffb726cf83965c85892202473044022001ec7e1f38f0592309100d22ad364384d1b7d089cb216c7eb78b39989be6f4fa0220597bd7b5fec9fb99834bcaec9256f0001c7bfdbbe41e6d800e745eb16eeaf08101210334887313ee43c10e0b5973579544656658659cd233523a251b14531f87d9d325b1000000

values at err:
remoteSig.txid=e3f11b4d2581b32fb902ef3c1940f43b5eaa24458d83a5956ddf55b87a833979, signedTxid=7939837ab855df6d95a5838d4524aa5e3bf440193cef02b92fb381254d1bf1e3

PR #2463, opened 2022-10-21
Use `tx_hash` in dual funding and verify `nSequence`

t-bast:
Similarly to what we do in `funding_created`, we should use the `tx_hash` in the serialized `tx_signatures` message, not the `txid`.

We should also verify that the `nSequence` used in `tx_add_input` allows replacement according to BIP 125. In theory we only need one of the inputs to use an `nSequence` below `0xfffffffd` but it's clearer to require it for every input to be able to fail early. The proposal to update the spec to make that requirement explicit is [here](https://github.com/niftynei/lightning-rfc/pull/6).

Thanks @niftynei for spotting this!

Supersedes #2462


eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -106,9 +106,15 @@ case class TxComplete(channelId: ByteVector32,
...
+                        txHash: ByteVector32,
                         witnesses: Seq[ScriptWitness],
-                        tlvStream: TlvStream[TxSignaturesTlv] = TlvStream.empty) extends InteractiveTxMessage with HasChannelId
+                        tlvStream: TlvStream[TxSignaturesTlv] = TlvStream.empty) extends InteractiveTxMessage with HasChannelId {
+  val txId: ByteVector32 = txHash.reverse
+}
+
+object TxSignatures {
+  def apply(channelId: ByteVector32, tx: Transaction, witnesses: Seq[ScriptWitness]): TxSignatures = TxSignatures(channelId, tx.hash, witnesses)
+}
```
pm47, line 117: Nit: I understand the purpose of this, but it is still possible to footgun ourselves by calling the main constructor, and it makes the content of the case class less obvious from the caller perspective. I would instead instantiate by explicitly specifying the arg name `txHash = tx.txid`.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -205,7 +205,7 @@ object LightningMessageCodecs {
 
   val txSignaturesCodec: Codec[TxSignatures] = (
     ("channelId" | bytes32) ::
-      ("txId" | sha256) ::
+      ("txHash" | sha256) ::
```
pm47, line 208: This `sha256` codec is weird. Codecs are supposed to deal with types, otherwise we will declare one codec per possible usage of integers.

pm47 (approved):
LGTM

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
t-bast: I agree, it's a bit confusing since it's just a short-hand for `bytes32`, but was only added to make it a bit more clear that a field is actually the `sha256` hash of _something_ (but we don't know what at the type level). Whether we keep this or not is probably out of scope of this PR though...

pm47: It would only make sense (from a consistent scodec usage pov) if we had a `Sha256` type (backed by a `ByteVector32`). But we don't. It's definitely outside of this PR 👍

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
t-bast: > You can't make the constructor private or you will have problem at deserialization.

Right, indeed!

Anyway it's also really painful to make a `case class`'s `apply` method private. We can make the constructor private by just defining it as `case class TxSignatures private (...)` but that only makes `val txSigs = new TxSignatures()` private, not `val txSigs = TxSignatures()`.

All the ways I've found to work-around this are a bit ugly...so let's keep it that way and rely on tests to ensure we don't mess this up.

PR #2464, opened 2022-10-24
Add `PeerReadyNotifier` actor

t-bast:
We add an actor that waits for a given peer to be connected and ready to process payments.
This is useful in the context of async payments for the receiver's LSP.

@remyers you can start building the mechanism to trigger a pending async payment on top of that branch.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerReadyNotifier.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/io/PeerReadyNotifier.scala
@@ -1,0 +1,196 @@
+/*
+ * Copyright 2022 ACINQ SAS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package fr.acinq.eclair.io
+
+import akka.actor.typed.eventstream.EventStream
+import akka.actor.typed.scaladsl.adapter.TypedActorRefOps
+import akka.actor.typed.scaladsl.{ActorContext, Behaviors, TimerScheduler}
+import akka.actor.typed.{ActorRef, Behavior}
+import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
+import fr.acinq.eclair.blockchain.CurrentBlockHeight
+import fr.acinq.eclair.{BlockHeight, Logs, channel}
+
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
+
+/**
+ * This actor waits for a given peer to be online and ready to process payments.
+ * It automatically stops after the timeout provided.
+ */
+object PeerReadyNotifier {
+
+  // @formatter:off
+  sealed trait Command
+  case class NotifyWhenPeerReady(replyTo: ActorRef[Result]) extends Command
+  private case object PeerNotConnected extends Command
+  private case class SomePeerConnected(nodeId: PublicKey) extends Command
+  private case class PeerChannels(channels: Set[akka.actor.ActorRef]) extends Command
+  private case class NewBlockNotTimedOut(currentBlockHeight: BlockHeight) extends Command
+  private case object CheckChannelsReady extends Command
+  private case class ChannelStates(states: Seq[channel.ChannelState]) extends Command
+  private case object Timeout extends Command
+
+  sealed trait Result
+  case class PeerReady(remoteNodeId: PublicKey, channelsCount: Int) extends Result
+  case class PeerUnavailable(remoteNodeId: PublicKey) extends Result
+  // @formatter:on
+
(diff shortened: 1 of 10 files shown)


t-bast:
@remyers I updated the PR to allow ommitting the timeout if you want to rebase #2491

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerReadyNotifier.scala
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  // We use an exhaustive pattern matching here to ensure we explicitly handle future new channel states.
+  private def isChannelReady(state: channel.ChannelState): Boolean = state match {
+    case channel.WAIT_FOR_INIT_INTERNAL => false
+    case channel.WAIT_FOR_INIT_SINGLE_FUNDED_CHANNEL => false
+    case channel.WAIT_FOR_INIT_DUAL_FUNDED_CHANNEL => false
+    case channel.OFFLINE => false
+    case channel.SYNCING => false
+    case channel.WAIT_FOR_OPEN_CHANNEL => true
```
remyers, line 146: For async-payments, what happens if `isChannelReady`  is true, but the channel is in a state before  `NORMAL` ? Would the channel logic queue the payment until the channel reaches `NORMAL`?  or would it fail the payment?

For `SHUTDOWN` or later I assume it would be correct to trigger and then fail an async-payment.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerReadyNotifierSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+
+    // The target peer connects.
+    system.eventStream ! EventStream.Publish(PeerConnected(TestProbe().ref.toClassic, remoteNodeId, null))
+    val request2 = switchboard.expectMessageType[Switchboard.GetPeerInfo]
+    request2.replyTo ! Peer.PeerInfo(TestProbe().ref.toClassic, remoteNodeId, Peer.CONNECTED, None, channels)
+    channelProbes.foreach(_.expectMessageType[CMD_GET_CHANNEL_STATE].replyTo ! RES_GET_CHANNEL_STATE(NEGOTIATING))
+    probe.expectMessage(PeerReady(remoteNodeId, 2))
+  }
+
+  test("peer connects then disconnects") { f =>
```
remyers, line 126: Related to my other comment about `ChannelStatesCollector` - could this situation lead to many actors being spawned from the  `waitForChannelsReady` state?

remyers (commented):
Looks good! Straight forward and easy to understand.

I had a questions about the `waitForChannelsReady` state possibly spawning actors and one about `async-payments`.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerReadyNotifier.scala
```diff
@@ -0,0 +1,196 @@
...
+    def apply(replyTo: ActorRef[ChannelStates], channels: Set[akka.actor.ActorRef]): Behavior[Command] = {
+      Behaviors.setup { context =>
+        val channelStateAdapter = context.messageAdapter[channel.RES_GET_CHANNEL_STATE](WrappedChannelState)
+        channels.foreach(c => c ! channel.CMD_GET_CHANNEL_STATE(channelStateAdapter.toClassic))
+        collect(replyTo, Nil, channels.size)
+      }
+    }
+
+    private def collect(replyTo: ActorRef[ChannelStates], received: Seq[channel.ChannelState], remaining: Int): Behavior[Command] = {
+      Behaviors.receiveMessage {
```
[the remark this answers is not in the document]
t-bast: Great feedback, thanks! I've added the peer disconnection events in https://github.com/ACINQ/eclair/pull/2464/commits/91f13eddd578bb68f3ac9614d8da2c66d1b9a51b which should handle that correctly.

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerReadyNotifier.scala
t-bast: We only care about verifying that the channels are fully initialized here, that's what guarantees that the peer is "ready". The actual state of those channels is unrelated to whether the peer is ready or not, and should be handled by other components (for payments, path-finding will deal with whether we can relay the payment or not, I don't think we should hold the payment waiting for a channel to confirm, at least not in those actors).

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerReadyNotifierSpec.scala
t-bast: This is a good point, fixed in https://github.com/ACINQ/eclair/pull/2464/commits/91f13eddd578bb68f3ac9614d8da2c66d1b9a51b

eclair-core/src/main/scala/fr/acinq/eclair/io/PeerReadyNotifier.scala
remyers: Your comment now explains it well; it only waits until channels have finished figuring out what state they're actually in, but no need to wait for any particular state.

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerReadyNotifierSpec.scala
remyers: Change/test solved any concerns.

remyers (approved):
LGTM! looking forward to rebasing async-payments onto it.

pm47 (approved):

PR #2465, opened 2022-10-24
Register can forward messages to specific nodes

t-bast:
The register contains the mapping from channel IDs to the corresponding actors and acts as a routing table. It can be useful to have the same feature to send messages to given nodes, outside of the context of any given channel.

But I'm wondering if that change is really necessary, or if we should instead use the `Switchboard` and the `GetPeerInfo` command when we want to send a message to a given peer? That also lets the sender get connection information for that peer, which probably makes more sense?

@remyers this should be useful for the peer-swap plugin.


t-bast:
This PR is in fact unnecessary, clients can use the `Switchboard` and its `GetPeerInfo` command to achieve a similar (probably better) result.

remyers:
Updated Peerswap plugin to use `GetPeerInfo` in [e0c7471](https://github.com/ACINQ/eclair-plugins/pull/5/commits/e0c7471006cf6a54eff97a85ff232e4f2e32adbb)

PR #2466, opened 2022-10-25
Update to bitcoind 23.0

t-bast:
This release of bitcoind contains several bug fixes that let us simplify our fee bumping logic:

- fixed a bug where bitcoind dropped non-wallet signatures
- added an option to fund transactions containing non-wallet inputs

It also has support for Taproot, which we obviously love.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxFunder.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxFunder.scala
@@ -22,7 +22,7 @@ import akka.actor.typed.{ActorRef, Behavior}
 import fr.acinq.bitcoin.scalacompat.{ByteVector32, OutPoint, Satoshi, Script, Transaction, TxOut}
 import fr.acinq.eclair.NotificationsLogger.NotifyNodeOperator
 import fr.acinq.eclair.blockchain.bitcoind.rpc.BitcoinCoreClient
-import fr.acinq.eclair.blockchain.bitcoind.rpc.BitcoinCoreClient.FundTransactionOptions
+import fr.acinq.eclair.blockchain.bitcoind.rpc.BitcoinCoreClient.{FundTransactionOptions, InputWeight}
 import fr.acinq.eclair.blockchain.fee.FeeratePerKw
 import fr.acinq.eclair.channel.Commitments
 import fr.acinq.eclair.channel.publish.ReplaceableTxPrePublisher._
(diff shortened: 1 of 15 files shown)


pom.xml
```diff
@@ -72,7 +72,7 @@
         <akka.version>2.6.18</akka.version>
         <akka.http.version>10.2.7</akka.http.version>
         <sttp.version>3.4.1</sttp.version>
-        <bitcoinlib.version>0.24</bitcoinlib.version>
+        <bitcoinlib.version>0.25-SNAPSHOT</bitcoinlib.version>
```
t-bast, line 75: TODO: use the released version once available

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxFunder.scala
```diff
@@ -405,75 +353,46 @@ private class ReplaceableTxFunder(nodeParams: NodeParams,
...
+    // The anchor transaction is paying for the weight of the commitment transaction.
+    val anchorWeight = Seq(InputWeight(anchorTx.txInfo.input.outPoint, anchorInputWeight + commitTx.weight()))
+    bitcoinClient.fundTransaction(txNotFunded, FundTransactionOptions(targetFeerate, lockUtxos = true, inputWeights = anchorWeight)).flatMap(fundTxResponse => {
       // We merge the outputs if there's more than one.
       fundTxResponse.changePosition match {
         case Some(changePos) =>
           val changeOutput = fundTxResponse.tx.txOut(changePos)
-          val txSingleOutput = fundTxResponse.tx.copy(txOut = Seq(changeOutput.copy(amount = changeOutput.amount + dummyChangeAmount)))
-          Future.successful(fundTxResponse.copy(tx = txSingleOutput))
+          val txSingleOutput = fundTxResponse.tx.copy(txOut = Seq(changeOutput))
```
sstone, line 369: Do we still need to make this a single output tx ?

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxFunder.scala
```diff
@@ -405,75 +353,46 @@ private class ReplaceableTxFunder(nodeParams: NodeParams,
...
+          // We ask bitcoind to sign the wallet inputs to learn their final weight and adjust the change amount.
+          bitcoinClient.signTransaction(txSingleOutput, allowIncomplete = true).map(signTxResponse => {
+            val dummySignedTx = addSigs(anchorTx.updateTx(signTxResponse.tx).txInfo, PlaceHolderSig)
+            val packageWeight = commitTx.weight() + dummySignedTx.tx.weight()
+            val anchorTxFee = weight2fee(targetFeerate, packageWeight) - weight2fee(commitments.localCommit.spec.commitTxFeerate, commitTx.weight())
```
sstone, line 374: Could `anchorTxFee` be < 0 (because `targetFeerate` is much lower than `commitTxFeerate` for example) ?

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxFunder.scala
t-bast: Yes, it wouldn't make sense to have two change outputs, right (and it would cost more in terms of weight)? Since this tx is only used to do CPFP, its only output will be change. Ideally `bitcoind` would let us call `fundrawtransaction` without any output so we wouldn't need this "hack", but they don't because transactions without outputs aren't consensus valid...

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxFunder.scala
t-bast: I don't think it can, because we always have `targetFeerate > commitTxFeerate`, otherwise we don't even claim the anchor output. This is done at the very beginning of the `fund` method.

t-bast: I can add a `require` here to make this explicit?

t-bast:
Apparently Bitcoin Core will ship a 23.1 soon: https://github.com/bitcoin/bitcoin/issues/24987

I'll test that release candidate, depending on timing we can either ship eclair v0.8.0 with 23.1 if it's ready, or with 23.0 and then update master to use 23.1 by default, as I believe users will be able to use either version without any issues.

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -56,6 +58,8 @@ object Transactions {
     override val htlcOutputWeight = 172
     override val htlcTimeoutWeight = 663
     override val htlcSuccessWeight = 703
+    override val htlcTimeoutInputWeight = 449
+    override val htlcSuccessInputWeight = 488
```
sstone, line 62: I checked these weights and found the same values

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -67,6 +71,8 @@ object Transactions {
     override val htlcOutputWeight = 172
     override val htlcTimeoutWeight = 666
     override val htlcSuccessWeight = 706
+    override val htlcTimeoutInputWeight = 452
+    override val htlcSuccessInputWeight = 491
```
sstone, line 75: I checked these weights and found the same values

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -196,16 +202,12 @@ object Transactions {
   /**
    * these values are specific to us (not defined in the specification) and used to estimate fees
    */
-  val claimP2WPKHOutputWitnessWeight = 109
   val claimP2WPKHOutputWeight = 438
+  val anchorInputWeight = 279
```
sstone, line 206: I checked this weight and found the same value

sstone (approved):
Nice, It makes fee bumping code much simpler.

eclair-core/src/main/scala/fr/acinq/eclair/channel/publish/ReplaceableTxFunder.scala
sstone: You're right it's not needed, this method is private and would not be called if the new target fee rate was lower the the current one.

t-bast:
Thanks for checking the weights!

fanquake:
Just FYI, Bitcoin Core 23.1 is now available: https://github.com/bitcoin/bitcoin/releases/tag/v23.1.

t-bast:
Thanks for the heads-up, I'll update eclair now!

PR #2467, opened 2022-10-25
Allow paying to blinded routes

thomash-acinq:
Replaces #2429
Adds `BlindedHop` in addition to the existing `ChannelHop` and `NodeHop`.
Bolt12 invoices encode the blinded routes as extra edges.


t-bast:
Replaced by #2482

PR #2468, opened 2022-10-25
Allow nodes to overshoot final htlc amount and expiry

t-bast:
When nodes receive HTLCs, they verify that the contents of those HTLCs match the intructions that the sender provided in the onion. It is important to ensure that intermediate nodes and final nodes have similar requirements, otherwise a malicious intermediate node could easily probe whether the next node is the final recipient or not.

Unfortunately, the requirements for intermediate nodes were more lenient than the requirements for final nodes. Intermediate nodes allowed overpaying and increasing the CLTV expiry, whereas final nodes required a perfect equality between the HTLC values and the onion values.

This provided a trivial way of probing: when relaying an HTLC, nodes could relay 1 msat more than what the onion instructed (or increase the outgoing expiry by 1). If the next node was an intermediate node, they would accept this HTLC, but if the next node was the recipient, they would reject it.

We update those requirements to fix this probing attack vector.

See https://github.com/lightning/bolts/pull/1032

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
@@ -170,8 +170,8 @@ object IncomingPaymentPacket {
 
   private def validateFinalPayload(add: UpdateAddHtlc, payload: TlvStream[OnionPaymentPayloadTlv]): Either[FailureMessage, FinalPacket] = {
     FinalPayload.Standard.validate(payload).left.map(_.failureMessage).flatMap {
-      case payload if add.amountMsat != payload.amount => Left(FinalIncorrectHtlcAmount(add.amountMsat))
-      case payload if add.cltvExpiry != payload.expiry => Left(FinalIncorrectCltvExpiry(add.cltvExpiry))
+      case payload if add.amountMsat < payload.amount => Left(FinalIncorrectHtlcAmount(add.amountMsat))
+      case payload if add.cltvExpiry < payload.expiry => Left(FinalIncorrectCltvExpiry(add.cltvExpiry))
       case payload => Right(FinalPacket(add, payload))
     }
   }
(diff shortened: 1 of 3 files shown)


pm47:
> What this PR does is about the relationship between the HTLC amount (from update_add_htlc) and the onion amount (not the total_amount_msat).

Okay, this change applies to _htlcs_, not _payments_, it's a different layer. The original sender can overpay up to 2x, but intermediate nodes may overpay as much as they want. Receiver won't even see it, as it should only look at the final payload. But in practice it is a bit weird in our current code though, since in the following validation we consider the onion amount but we print the htlc amount :
https://github.com/ACINQ/eclair/blob/a8389a042a149b3b126c617e577a917ffb48448e/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala#L382-L388

t-bast:
> Okay, this change applies to htlcs, not payments, it's a different layer. The original sender can overpay up to 2x, but intermediate nodes may overpay as much as they want.

Yes, that is a good way of looking at it, this change is for the HTLC layer.

> But in practice it is a bit weird in our current code though, since in the following validation we consider the onion amount but we print the htlc amount

This code snippet only verify the spec quote from before: the onion `total_amount_msat` must be between the invoice amount and 2x that amount. It's unrelated to the current change. There is no reason to change anything here?

The `MultiPartHandler` then always uses the HTLC amount in the accounting we do, so since we validate that this amount is always greater than the `amt_to_fwd` provided in the onion, I believe we're doing everything correctly and storing exactly how much we actually received, don't you think?

pm47:
> The MultiPartHandler then always uses the HTLC amount in the accounting we do, so since we validate that this amount is always greater than the amt_to_fwd provided in the onion, I believe we're doing everything correctly and storing exactly how much we actually received, don't you think?

I guess before this PR we could (should for consistency?) have used `payload.amount` instead of `add.amountMsat`, but now we have to use `add.amountMsat`, because the last hop may have sent a higher amount. Wait, what if they sent a amount so high that it makes the payment valid in terms of total payment amount even if `payload.totalAmount` was too low. Fat fetch example, but doesn't it show the ambiguity here?

pm47:
Fair enough, I get it now.

pm47 (approved):

PR #2469, opened 2022-10-25
Randomize final cltv expiry

t-bast:
When sending outgoing payments, using an expiry that is very close to the current block height makes it obvious to the next-to-last node who the recipient is.

We add a random expiry to the amount we use to make it plausible that the route contains more hops.

Core Lightning and LDK already do this, LND doesn't.


eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentInitiatorSpec.scala
```diff
@@ -104,7 +104,7 @@ class PaymentInitiatorSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike
     payFsm.expectMsgType[SendPaymentConfig]
     val tlvs = payFsm.expectMsgType[PaymentLifecycle.SendPayment].finalPayload.records
     assert(tlvs.get[AmountToForward].get.amount == finalAmount)
-    assert(tlvs.get[OutgoingCltv].get.cltv == req.invoice.minFinalCltvExpiryDelta.toCltvExpiry(nodeParams.currentBlockHeight + 1))
+    assert(tlvs.get[OutgoingCltv].get.cltv > req.invoice.minFinalCltvExpiryDelta.toCltvExpiry(nodeParams.currentBlockHeight + 50))
```
thomash-acinq, line 107: Where does the 50 come from? Shouldn't it be 100?
```suggestion
    assert(tlvs.get[OutgoingCltv].get.cltv >= req.invoice.minFinalCltvExpiryDelta.toCltvExpiry(nodeParams.currentBlockHeight + 100))
    assert(tlvs.get[OutgoingCltv].get.cltv < req.invoice.minFinalCltvExpiryDelta.toCltvExpiry(nodeParams.currentBlockHeight + 350))
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentInitiatorSpec.scala
```diff
@@ -176,11 +179,14 @@ class PaymentInitiatorSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike
     val finalExpiryDelta = CltvExpiryDelta(24)
     val invoice = Bolt11Invoice(Block.LivenetGenesisBlock.hash, Some(finalAmount), paymentHash, priv_c.privateKey, Left("Some MPP invoice"), finalExpiryDelta, features = featuresWithMpp)
     val req = SendPaymentToNode(finalAmount, invoice, 1, routeParams = nodeParams.routerConf.pathFindingExperimentConf.getRandomConf().getDefaultRouteParams)
-    assert(req.finalExpiry(nodeParams.currentBlockHeight) == (finalExpiryDelta + 1).toCltvExpiry(nodeParams.currentBlockHeight))
+    assert(req.finalExpiry(nodeParams.currentBlockHeight) > (finalExpiryDelta + 1).toCltvExpiry(nodeParams.currentBlockHeight))
```
thomash-acinq, line 182:
```suggestion
    assert(req.finalExpiry(nodeParams.currentBlockHeight) >= (finalExpiryDelta + 100).toCltvExpiry(nodeParams.currentBlockHeight))
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentInitiatorSpec.scala
```diff
@@ -268,7 +278,7 @@ class PaymentInitiatorSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike
     val msg = multiPartPayFsm.expectMsgType[SendMultiPartPayment]
     assert(msg.paymentSecret !== invoice.paymentSecret) // we should not leak the invoice secret to the trampoline node
     assert(msg.targetNodeId == b)
-    assert(msg.targetExpiry.toLong == currentBlockCount + 9 + 12 + 1)
+    assert(msg.targetExpiry.toLong > currentBlockCount + 50)
```
thomash-acinq, line 281: Again, why 50?

eclair-core/src/test/scala/fr/acinq/eclair/payment/PaymentInitiatorSpec.scala
t-bast: Right, I wrote the test first and initially used 50, then changed it to 100 in the implementation but not in the tests...
Do you think we should make that configurable by the node operator?
This way they could disable it if they wish to?

thomash-acinq: Making it configurable is a good idea. If everyone uses a delay of at least 100, getting a cltv of 100 is just like getting a cltv of 0 today, so there needs to be the possibility that at least some people are using no delay.

t-bast: I've reworked this in https://github.com/ACINQ/eclair/pull/2469/commits/068619850075f8b36660290188543e39126492ba to make it configurable, which also lets node operators disable it by providing a "delta" containing a single value.

I've created a new `payment` section in `eclair.conf`, we can later migrate some parameters inside that new section (for example, `max-payment-attempts`).

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -98,6 +99,25 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
...
+  require(min <= max, "maximum cltv-expiry-delta cannot be smaller than minimum-cltv-expiry-delta")
+
+  /**
+   * When sending a payment, if the cltv expiry used for the final node is very close to the current block height, it
+   * lets intermediate nodes figure out their position in the route. To protect against this, a random delta is added
+   * to the current block height, which makes it look like there are more hops after the final node.
+   */
+  def computeFinalExpiry(currentBlockHeight: BlockHeight, minFinalExpiryDelta: CltvExpiryDelta): CltvExpiry = {
+    val additionalDelta = if (min < max) {
+      min + (randomLong() % (max - min).toInt).toInt.abs
```
thomash-acinq, line 113:
```suggestion
      min + (randomLong() % (max - min + 1).toInt).toInt.abs
```
You should add 1 so that the max is reachable.

thomash-acinq (dismissed):

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
t-bast: Good idea, that makes more sense. Done in https://github.com/ACINQ/eclair/pull/2469/commits/292a069e96fe65cf0bcefdc8495a0b418c069354

thomash-acinq (dismissed):

eclair-core/src/main/resources/reference.conf
```diff
@@ -160,6 +160,16 @@ eclair {
     }
   }
 
+  payment {
```
pm47, line 163: Should we name it after the package, like `relay`?
```suggestion
  send {
```

Let's dig another 🐰 🕳️ !
```
  relay {
    ...
  }

  send {  
    max-payment-attempts = 5
    recipient-final-expiry {
      min-delta = 150 // minimum value to add to the current block height
      max-delta = 350 // maximum value to add to the current block height
    }
  }

  receive {
    multi-part-payment-expiry = 60 seconds // default expiry for receiving all parts of a multi-part payment
    invoice-expiry = 1 hour // default expiry for invoices generated by this node
    purge-expired-invoices {
      enabled = true // enable automatic purges of expired invoices from the database
      interval = 24 hours // interval between expired invoice purges
    }
  }
```

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
```diff
@@ -98,6 +99,25 @@ case class NodeParams(nodeKeyManager: NodeKeyManager,
...
+   */
+  def computeFinalExpiry(currentBlockHeight: BlockHeight, minFinalExpiryDelta: CltvExpiryDelta): CltvExpiry = {
+    val additionalDelta = if (min < max) {
+      min + (randomLong() % (max - min + 1).toInt).toInt.abs
+    } else {
+      max
+    }
+    (minFinalExpiryDelta + additionalDelta).toCltvExpiry(currentBlockHeight)
+  }
+}
```
pm47, line 119: I don't think this method belongs in `NodeParams`. If we clean up the sections in `eclair.conf`, then we will probably end up with a `PaymentSendConf` class somewhere in `payments.send` where it would make more sense?

pm47 (commented):
Just nits, I'm fine with leaving as-is for now.

eclair-core/src/main/scala/fr/acinq/eclair/NodeParams.scala
t-bast: Agreed, we should move that when we clean up the `eclair.conf` sections!

eclair-core/src/main/resources/reference.conf
t-bast: Definitely a good idea, I didn't want to move existing values in the new section in this PR but we should clearly do a follow-up PR for that! I renamed the section to `send` in https://github.com/ACINQ/eclair/pull/2469/commits/8c1b462073f90d4e68a6cc15e006a7a085419fc8

pm47 (approved):

PR #2473, opened 2022-11-02
Explain how and why we use bitcoin core

sstone:
Explain why we chose to delegate most onchain tasks to bitcoin core (including onchain wallet management), the additional requirements that it creates and also the benefits in terms of security.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,14 +53,14 @@ You will also find detailed [guides](./docs/Guides.md) and [frequently asked que
 
 ### Prerequisite: Bitcoin Core
 
-Eclair relies on Bitcoin Core to interface with and monitor the blockchain and to manage on-chain funds (Eclair does not include an on-chain wallet, channel opening transactions are funded by your Bitcoin Core node, and channel closing transactions return funds to your Bitcoin Core node).
+Eclair relies on Bitcoin Core to interface with and monitor the blockchain and to manage on-chain funds: Eclair does not include an on-chain wallet, channel opening transactions are funded by your Bitcoin Core node, and channel closing transactions return funds to your Bitcoin Core node.
 
 This means that instead of re-implementing them, Eclair benefits from the verifications and optimisations (including fee management with RBF/CPFP, ...) that are implemented by Bitcoin Core. Eclair uses our own [bitcoin library](https://github.com/ACINQ/bitcoin-kmp) to verify data provided by Bitcoin Core.
 
-:warning: This also means that Eclair has strong requirements on how your Bitcoin Core node is configured (see below), and that you must back up your Bitcoin Core wallet as well as your eclair node (see [here](#configure-bitcoin-core-wallet)): 
+:warning: This also means that Eclair has strong requirements on how your Bitcoin Core node is configured (see below), and that you must back up your Bitcoin Core wallet as well as your Eclair node (see [here](#configure-bitcoin-core-wallet)): 
 - Eclair needs a _synchronized_, _segwit-ready_, 
 **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node. 
-- You must configure your Bitcoin node to use `bech32` (segwit) addresses. If your wallet has "non-segwit UTXOs" (outputs that are neither `p2sh-segwit` or `bech32`), you must send them to a `bech32` address before running eclair.
+- You must configure your Bitcoin node to use `bech32` (segwit) addresses. If your wallet has "non-segwit UTXOs" (outputs that are neither `p2sh-segwit` or `bech32`), you must send them to a `bech32` address before running Eclair.
 - Eclair requires Bitcoin Core 0.20.1 or 0.21.1. (other versions of Bitcoin Core are *not* actively tested - use at your own risk).  If you are upgrading an existing wallet, you may need to create a new address and send all your funds to that address.
 
 Run bitcoind with the following minimal `bitcoin.conf`:
--- a/README.md
+++ b/README.md
@@ -53,18 +53,15 @@ You will also find detailed [guides](./docs/Guides.md) and [frequently asked que
 
 ### Prerequisite: Bitcoin Core
 
-We chose to rely as much as possible on bitcoin core, to interface with and monitor the blockchain and also to manage onchain funds (eclair does not include an onchain wallet, channel opening transactions are funded by your bitcoin core node, and channel closing transactions return funds to your bitcoin core node).
+Eclair relies on Bitcoin Core to interface with and monitor the blockchain and to manage on-chain funds (Eclair does not include an on-chain wallet, channel opening transactions are funded by your Bitcoin Core node, and channel closing transactions return funds to your Bitcoin Core node).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,12 @@ You will also find detailed [guides](./docs/Guides.md) and [frequently asked que
 
 ### Prerequisite: Bitcoin Core
 
+We chose to rely as much as possible on bitcoin core, to interface with and monitor the blockchain and also to manage onchain funds (eclair does not include an onchain wallet, channel opening transactions are funded by your bitcoin core node, and channel closing transactions return funds to your bitcoin core node).
+
+This means that we have strong requirements on how your bitcoin node is configured (see below), and that you must backup your bitcoin core wallet as well as your eclair node (see https://github.com/ACINQ/eclair/edit/master/README.md#configure-bitcoin-core-wallet)
+
+But is also means that instead of re-implemeting them, we benefit from all the verifications and optimisations (including fee management with RBF/CPFP, ...) that are provided by bitcoin core, which is much better from a security point of view. We did have to implement some of the bitcoin protocol, but in the context of verifying data provided by bitcoin core.
```
joostjager, line 60: Out of curiosity, is there any part in Eclair that you'd still consider risky to some degree because it re-implements bitcoin logic? Perhaps any sort of re-parsing of blocks, or did you manage to avoid all of that?

joostjager: Definitely a strong point to rely on battle tested code as much as possible.

t-bast: We still have a utility function that verifies that an unconfirmed transaction is valid (have a look at occurrences of `Transaction.correctlySpends`), where we could potentially deviate from `bitcoind` which would create issues. However, this function is never used to validate transactions that are included in blocks, we only use it when collaboratively building a transaction (e.g. when validating remote signatures of a commit tx or a dual-funded tx). AFAICT, this should never make an `eclair` node halt. The only risks that I see are that we'd either:

1. Accept a transaction that would never confirm because it's actually invalid: that means we would have a channel stuck in the `unconfirmed` state, which doesn't create any risks as long as you don't use 0-conf (if you do, you have a trust assumption on your peer anyway)
2. Reject a transaction that is actually valid: that could lead to eclair rejecting channel funding attempts that should work, which doesn't incur any loss of funds either, but could be pretty annoying 

We also obviously also have custom code to deserialize txs and blocks into scala objects: if we were unable to deserialize some txs or blocks, it could lead to partial denial of service. But we don't run any validation on those objects, which makes the attack surface pretty low, but there is still a potential risk there.

README.md
```diff
@@ -53,6 +53,12 @@ You will also find detailed [guides](./docs/Guides.md) and [frequently asked que
 
 ### Prerequisite: Bitcoin Core
 
+We chose to rely as much as possible on bitcoin core, to interface with and monitor the blockchain and also to manage onchain funds (eclair does not include an onchain wallet, channel opening transactions are funded by your bitcoin core node, and channel closing transactions return funds to your bitcoin core node).
```
t-bast, line 56: nit: we're capitalizing `Bitcoin Core` everywhere else, and use `on-chain` instead of `onchain` in documentation

README.md
t-bast, line 60: nit:

```suggestion
But is also means that instead of re-implementing them, we benefit from all the verifications and optimisations (including fee management with RBF/CPFP, ...) that are provided by bitcoin core, which is much better from a security point of view. We did have to implement some of the bitcoin protocol, but in the context of verifying data provided by bitcoin core.
```

joostjager: Sounds good. Especially the part where you are not forcing every block seen through your own validation code 👍

t-bast (commented):
A few nits, otherwise LGTM

README.md
```diff
@@ -53,12 +53,15 @@ You will also find detailed [guides](./docs/Guides.md) and [frequently asked que
 
 ### Prerequisite: Bitcoin Core
 
-:warning: Eclair requires Bitcoin Core 0.20.1 or 0.21.1. (other versions of Bitcoin Core are *not* actively tested - use at your own risk).  If you are upgrading an existing wallet, you may need to create a new address and send all your funds to that address.
+Eclair relies on Bitcoin Core to interface with and monitor the blockchain and to manage on-chain funds (Eclair does not include an on-chain wallet, channel opening transactions are funded by your Bitcoin Core node, and channel closing transactions return funds to your Bitcoin Core node).
```
remyers, line 56: This deserves to be it's own un-parenthesized sentence:

> ...on-chain funds. Eclair does not include an on-chain wallet, channel opening transactions are funded by your Bitcoin Core node, and channel closing transactions return funds to your Bitcoin Core node.

README.md
```diff
@@ -53,12 +53,15 @@ You will also find detailed [guides](./docs/Guides.md) and [frequently asked que
...
 
-Eclair needs a _synchronized_, _segwit-ready_, **_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node.
+This means that instead of re-implementing them, Eclair benefits from the verifications and optimisations (including fee management with RBF/CPFP, ...) that are implemented by Bitcoin Core. Eclair uses our own [bitcoin library](https://github.com/ACINQ/bitcoin-kmp) to verify data provided by Bitcoin Core.
 
-You must configure your Bitcoin node to use `bech32` (segwit) addresses.
-If your wallet has "non-segwit UTXOs" (outputs that are neither `p2sh-segwit` or `bech32`), you must send them to a `bech32` address before running eclair.
+:warning: This also means that Eclair has strong requirements on how your Bitcoin Core node is configured (see below), and that you must back up your Bitcoin Core wallet as well as your eclair node (see [here](#configure-bitcoin-core-wallet)):
```
remyers, line 60: nit: s/eclair/Eclair

README.md
```diff
@@ -53,12 +53,15 @@ You will also find detailed [guides](./docs/Guides.md) and [frequently asked que
...
+This means that instead of re-implementing them, Eclair benefits from the verifications and optimisations (including fee management with RBF/CPFP, ...) that are implemented by Bitcoin Core. Eclair uses our own [bitcoin library](https://github.com/ACINQ/bitcoin-kmp) to verify data provided by Bitcoin Core.
 
-You must configure your Bitcoin node to use `bech32` (segwit) addresses.
-If your wallet has "non-segwit UTXOs" (outputs that are neither `p2sh-segwit` or `bech32`), you must send them to a `bech32` address before running eclair.
+:warning: This also means that Eclair has strong requirements on how your Bitcoin Core node is configured (see below), and that you must back up your Bitcoin Core wallet as well as your eclair node (see [here](#configure-bitcoin-core-wallet)): 
+- Eclair needs a _synchronized_, _segwit-ready_, 
+**_zeromq-enabled_**, _wallet-enabled_, _non-pruning_, _tx-indexing_ [Bitcoin Core](https://github.com/bitcoin/bitcoin) node. 
+- You must configure your Bitcoin node to use `bech32` (segwit) addresses. If your wallet has "non-segwit UTXOs" (outputs that are neither `p2sh-segwit` or `bech32`), you must send them to a `bech32` address before running eclair.
```
remyers, line 63: nit: s/eclair/Eclair

t-bast (dismissed):

remyers (changes requested):
LGTM! 

It might be useful to collaborate on a best-practices document to explore the interface between Lightning and on-chain where we could get into more technical details about the trade-offs.

remyers (approved):
Looks good!

t-bast (approved):

pm47 (approved):

Issue #2450, opened 2022-10-06
Add pagination on listInvoices


ShahanaFarooqui:
listinvoices has an option to fetch the list by duration but it does not provide total number of invoices to show on the UI. 

Pagination with offset and total records will be a better option to consume the API efficiently and effectively.

PR #2474, opened 2022-11-02
Add pagination on listInvoices

rorp:
Fixes https://github.com/ACINQ/eclair/issues/2450

cc: @ShahanaFarooqui

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
@@ -214,6 +214,13 @@ trait JdbcUtils {
     implicit def conv(rs: ResultSet): ExtendedResultSet = ExtendedResultSet(rs)
   }
 
+  def limited(sql: String, count_opt: Option[Int], skip_opt: Option[Int], countRequired: Boolean): String = {
+    if (countRequired && skip_opt.isDefined && count_opt.isEmpty) throw new IllegalArgumentException("Item count is required")
+    val sqlWithLimit = count_opt.map(count => s"$sql LIMIT $count").getOrElse(sql)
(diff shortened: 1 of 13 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
```diff
@@ -214,6 +214,13 @@ trait JdbcUtils {
     implicit def conv(rs: ResultSet): ExtendedResultSet = ExtendedResultSet(rs)
   }
 
+  def limited(sql: String, count_opt: Option[Int], skip_opt: Option[Int], countRequired: Boolean): String = {
```
t-bast, line 217: The `countRequired` argument is a bit hacky and from an API's perspective it would be weird to have different requirements if you use sqlite or postgres.

What about defining `case class Paginated(count: Int, skip: Int)` in `Eclair.scala` and take a `pagination_opt: Option[Paginated]` in the `allInvoices` and everything downstream?

Then the API handler would just verify that:

- if `skip` is provided, `count` has to be provided as well, otherwise reject the request
- if `count` is provided and `skip` isn't provided, use a default value of `0` for skip in the `Paginated` argument

What do you think? This `Paginated` structure can then easily be used in other APIs in the future if necessary.

t-bast (commented):
Thanks for doing that @rorp!

eclair-core/src/main/scala/fr/acinq/eclair/db/jdbc/JdbcUtils.scala
rorp: For some reason Sqlite doesn't accept `OFFSET` without `LIMIT`, but I can make `count` required.

As for `Paginated`: that's what I was thinking after mixing up `count` and `skip` parameters couple of times.

t-bast (approved):
LGTM, thanks!

PR #2476, opened 2022-11-04
Reduce diff with feature branch

pm47:
Best reviewed commit by commit. Those are minor changes that mostly touch tests.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
@@ -1699,7 +1699,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val 
       }
       emitEvent_opt.foreach {
         case EmitLocalChannelUpdate(reason, d, sendToPeer) =>
-          log.info(s"emitting channel update event: reason=$reason enabled=${d.channelUpdate.channelFlags.isEnabled} sendToPeer=$sendToPeer realScid=${d.shortIds.real} channel_update={} channel_announcement={}", d.channelUpdate, d.channelAnnouncement.map(_ => "yes").getOrElse("no"))
+          log.debug(s"emitting channel update event: reason=$reason enabled=${d.channelUpdate.channelFlags.isEnabled} sendToPeer=$sendToPeer realScid=${d.shortIds.real} channel_update={} channel_announcement={}", d.channelUpdate, d.channelAnnouncement.map(_ => "yes").getOrElse("no"))
           val lcu = LocalChannelUpdate(self, d.channelId, d.shortIds, d.commitments.remoteParams.nodeId, d.channelAnnouncement, d.channelUpdate, d.commitments)
           context.system.eventStream.publish(lcu)
           if (sendToPeer) {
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1699,7 +1699,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
       }
       emitEvent_opt.foreach {
         case EmitLocalChannelUpdate(reason, d, sendToPeer) =>
-          log.info(s"emitting channel update event: reason=$reason enabled=${d.channelUpdate.channelFlags.isEnabled} sendToPeer=$sendToPeer realScid=${d.shortIds.real} channel_update={} channel_announcement={}", d.channelUpdate, d.channelAnnouncement.map(_ => "yes").getOrElse("no"))
+          log.debug(s"emitting channel update event: reason=$reason enabled=${d.channelUpdate.channelFlags.isEnabled} sendToPeer=$sendToPeer realScid=${d.shortIds.real} channel_update={} channel_announcement={}", d.channelUpdate, d.channelAnnouncement.map(_ => "yes").getOrElse("no"))
```
pm47, line 1702: @t-bast Is this okay or do we need that level of granularity for debug?

Alternatively I can filter out the "restore" reason, which is the spammy one

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -134,15 +153,14 @@ class PeerSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with Paralle
     probe.send(peer, Peer.Connect(remoteNodeId, mockAddress_opt, probe.ref, isPersistent = true))
 
     // assert our mock server got an incoming connection (the client was spawned with the address from node_announcement)
-    awaitCond(mockServer.accept() != null, max = 30 seconds, interval = 1 second)
+    eventually(mockServer.accept() != null)
```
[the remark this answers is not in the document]
pm47: 🤦‍♂️

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
```diff
@@ -218,7 +239,7 @@ class PeerSpec extends TestKitBaseClass with FixtureAnyFunSuiteLike with Paralle
     val probe = TestProbe()
     switchboard.send(peer, Peer.Init(Set.empty))
 
-    awaitCond {
+    eventually {
       probe.send(peer, Peer.GetPeerInfo(None))
       probe.expectMsgType[Peer.PeerInfo].state == Peer.DISCONNECTED
```
t-bast, line 244:
```suggestion
      assert(probe.expectMsgType[Peer.PeerInfo].state == Peer.DISCONNECTED)
```

t-bast (commented):
Last one and otherwise LGTM

eclair-core/src/test/scala/fr/acinq/eclair/io/PeerSpec.scala
pm47: Right! Went too fast on that one and assumed the `probe.expectMsgType[Peer.PeerInfo]` would throw

PR #2479, opened 2022-11-09
Add support for paying offers

thomash-acinq:
- Add a new actor paying an offer: sends an invoice request, wait for the invoice, retries if needed, pay the invoice
- Add a new database for paying offers: keeps tracks of payment attempts and their statuses


t-bast:
Is this PR really ready to review? At first glance, it does way too much whereas it should be very easy to split it into several independent and more focused PRs, for example:

- a first PR that lets users generate offers that are stored in the DB and provides APIs to view/list/delete/etc offers
- another PR that responds to `invoice_request`s
- another PR that pays offers
- (this listing is just one way to split this work, but if you see a better way to split it don't hesitate!)

This PR is also clearly lacking test coverage, especially given the amount of new code that it's adding...

thomash-acinq:
> Is this PR really ready to review?

I think it is ready but I'm waiting until I have everything so I can write tests.

> At first glance, it does way too much whereas it should be very easy to split it into several independent and more focused PRs, for example:
> 
>     * a first PR that lets users generate offers that are stored in the DB and provides APIs to view/list/delete/etc offers
> 
>     * another PR that responds to `invoice_request`s
> 
>     * another PR that pays offers
> 
>     * (this listing is just one way to split this work, but if you see a better way to split it don't hesitate!)

1 and 2 are out of scope for this PR, as the title says, it's only about paying offers, not being paid for offers. I'm not sure what made you think that it does more than that.

> This PR is also clearly lacking test coverage, especially given the amount of new code that it's adding...

That's because it's only one side of the offer. I'm waiting for the other side to test everything together.

t-bast:
> 1 and 2 are out of scope for this PR, as the title says, it's only about paying offers, not being paid for offers. I'm not sure what made you think that it does more than that.

It's the DB part that threw me off, looking at it briefly I thought that was the recipient side! I guess I don't understand why it needs a new DB, intuitively it should just use the payments DB, but I'll need to dive more into it then.

I was surprised that we couldn't split the new DB code and the new actor into separate PRs, but maybe that doesn't make sense? However anything you can do to split big PRs into smaller units will greatly speed up the pace of getting code merged, so it's worth spending time thinking about it.

> That's because it's only one side of the offer. I'm waiting for the other side to test everything together.

That's ok for e2e tests, but there is a lot of code in here that can be unit tested and shouldn't need the other side at all. But if you want to wait for have the whole feature implemented before writing the tests, that works for me, I'll wait for that before reviewing the PR.

t-bast:
I've looked more into this PR, and given what it does, I still don't think we should introduce a new DB. Is there a good reason to add it for future work that builds on top of this and that I'm not seeing yet?

The reasons I think we should avoid it are:

- DB access is the main performance bottleneck (and not a small one), so we should restrict it to things we really need to store permanently (across restarts)
- using a new DB instead of the existing `PaymentsDb` means you're not correctly [updating payment status after a restart](https://github.com/ACINQ/eclair/blob/927e1c8ad4137e3539567330e4d6af0a33a218a5/eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/PostRestartHtlcCleaner.scala#L159)
- the only reason you seem to be using that DB is to be able to report progress to the user before we've received the invoice, but I'm not sure it's necessary, and even if it were, I think that should only be transient state, not persistent state

I believe you introduced this DB mostly for the last point, so we should probably discuss that one more in depth. The first thing we should decide is whether we actually want to report progress before we've received the invoice or not. If we work under the assumption that fetching the invoice should quickly succeed or quickly fail (which we actually control by setting timeouts), we can use this simpler flow:

- we receive a command to pay an offer
- we send the invoice request
- if we receive the invoice, we delegate the rest of the flow to the `PaymentInitiator` which will return a `paymentId` to the API: this can then be used exactly like normal payments to track progress (using `getsentinfo`)
- if something fails or we timeout before getting the invoice, we send an error back to the `sender`/`replyTo`

I think this flow is completely acceptable from a UX's point of view, what do you think? At least it's a leaner first version?

I think the only reason we'd want to avoid blocking the API request until we've received the invoice is if we want to support a large volume of concurrent outgoing payments, for example when running a custodial hub. But that's not a use-case we really have right now, so we can address it later depending on the needs of that hub: if for example they use a plugin instead of the API, they won't be blocked so they don't need additional work.

If it later turns out that we want the API to support a large volume of concurrent payments, we don't really need a DB for that either. We can simply have a singleton `OfferPaymentInitiator` actor that is exposed in the `kit` and will spawn the actual offer payments, and we can keep the internal state prior to receiving the invoice inside the offer payment actor state. We can then simply change `getsentinfo` to first try querying the `OfferPaymentInitiator` (which will ask the corresponding child actor for its current state, if there is a matching child), and if it returns "not found" then fallback to querying the `PaymentInitiator`?

Apart from the DB, I like how simple and concise the code is to pay an offer! :tada:

t-bast:
I haven't looked at the new changes yet, but high-level comments on your feedback.

> I've removed OfferDb, but now I don't know where to put the payerKey, we should store it somewhere.

Why don't we generate `payerKey` deterministically from our node private key? We could use HKDF with our private key and the payment ID, this way we don't have to store anything.

Otherwise if we think some scenarios require overriding that default behavior, we can simply update the `sent_payments` SQL table to add a new column for the payer key, can't we?

> Also we no longer have an easy way to list payments to a given offer.

That's a good point, that would be useful. But can't we simply add an optional `offer` column to the `sent_payments` SQL table?

eclair-core/src/main/scala/fr/acinq/eclair/db/DualDatabases.scala
```diff
@@ -10,7 +10,8 @@ import fr.acinq.eclair.db.DualDatabases.runAsync
 import fr.acinq.eclair.payment._
 import fr.acinq.eclair.payment.relay.Relayer.RelayFees
 import fr.acinq.eclair.router.Router
-import fr.acinq.eclair.wire.protocol.{ChannelAnnouncement, ChannelUpdate, NodeAddress, NodeAnnouncement}
+import fr.acinq.eclair.wire.protocol.OfferTypes.InvoiceRequest
+import fr.acinq.eclair.wire.protocol.{ChannelAnnouncement, ChannelUpdate, NodeAddress, NodeAnnouncement, OfferTypes}
```
t-bast, line 14: nit: remove changes to that file

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/OfferPayment.scala
```diff
@@ -0,0 +1,185 @@
...
+import fr.acinq.eclair.wire.protocol.{OfferTypes, OnionMessagePayloadTlv, TlvStream}
+import fr.acinq.eclair.{Features, InvoiceFeature, MilliSatoshi, NodeParams, TimestampSecond, randomKey}
+
+import java.util.UUID
+import scala.util.Random
+
+object OfferPayment {
+  sealed trait Result
+
+  sealed trait Failure extends Exception with Result
```
t-bast, line 41: We should avoid using `Exception` here, and only translate to an exception in `Eclair.scala` (at the boundaries with the HTTP API, where we don't have a choice because we need to use `Future.failed`).

Note that in the longer term, we don't really need to use `Future.failed` for the API, we could always return a `Future.successful` with an error object, which would be cleaner and better for performance. But that's for a very separate refactoring!

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -164,6 +165,8 @@ trait Eclair {
 
   def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: Either[PublicKey, Sphinx.RouteBlinding.BlindedRoute], replyPath: Option[Seq[PublicKey]], userCustomContent: ByteVector)(implicit timeout: Timeout): Future[SendOnionMessageResponse]
 
+  def payOffer(offer: Offer, amount: MilliSatoshi, quantity: Long, externalId_opt: Option[String] = None, maxAttempts_opt: Option[Int] = None, maxFeeFlat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None, pathFindingExperimentName_opt: Option[String] = None)(implicit timeout: Timeout): Future[UUID]
```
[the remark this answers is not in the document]
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/OfferPayment.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/OfferPayment.scala
```diff
@@ -0,0 +1,157 @@
...
+            Behaviors.stopped
+          } else {
+            val payerKey = randomKey()
+            val request = InvoiceRequest(offer, amount, quantity, nodeParams.features.bolt12Features(), payerKey, nodeParams.chainHash)
+            sendInvoiceRequest(nodeParams, postman, paymentInitiator, context, request, payerKey, replyTo, nodeParams.onionMessageConfig.maxAttempts, sendPaymentConfig)
+          }
+      })
+  }
+
+  val rand = new Random
```
t-bast, line 101: Instead of selecting a route at random, wouldn't it make more sense to iterate through them as we get failures, to ensure we try a different one? A simple way of doing that would be for example to replace `remainingAttempts` by `attemptNumber` (starting at 0) and using `blindedRoutes(attemptNumber % blindedRoutes.length)`.

eclair-core/src/main/resources/reference.conf
```diff
@@ -461,7 +461,10 @@ eclair {
     max-per-peer-per-second = 10
 
     # Consider a message to be lost if we haven't received a reply after that amount of time
-    reply-timeout = 20 seconds
+    reply-timeout = 5 seconds
+
+    # If we expect a reply but do not get one, retry until we reach this number of attempts
+    max-attempts = 1
```
t-bast, line 467: Wouldn't it make more sense to use a default that actually does a retry, for example 3? Especially when offers provide multiple blinded paths, it would be a waste to only try one of them...

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/OfferPayment.scala
```diff
@@ -0,0 +1,157 @@
...
+                         sendPaymentConfig: SendPaymentConfig): Behavior[Command] = {
+    val destination = request.offer.contactInfo match {
+      case Left(blindedRoutes) =>
+        val blindedRoute = blindedRoutes(rand.nextInt(blindedRoutes.length))
+        OnionMessages.BlindedPath(blindedRoute)
+      case Right(nodeId) =>
+        OnionMessages.Recipient(nodeId, None, None)
+    }
+    // TODO: random choice of intermediate nodes
+    val intermediateNodes = Nil
```
t-bast, line 120: It would be nice to implement this in a follow-up PR, as directly connecting to the recipient (or the introduction node) leaks the payer's identity if it's the default behavior...what's your plan here? Making a path-finding call to the router?

eclair-core/src/test/scala/fr/acinq/eclair/payment/send/OfferPaymentSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+import org.scalatest.Outcome
+import org.scalatest.funsuite.FixtureAnyFunSuiteLike
+import scodec.bits.HexStringSyntax
+
+import java.util.UUID
+import scala.concurrent.duration.DurationInt
+
+class OfferPaymentSpec extends ScalaTestWithActorTestKit(ConfigFactory.load("application")) with FixtureAnyFunSuiteLike {
+
+  case class FixtureParam(offerPayment: ActorRef[Command], nodeParams: NodeParams, postman: TypedProbe[Postman.Command], paymentInitiator: akka.testkit.TestProbe, routeParams: RouteParams)
```
t-bast, line 45: nit:
```suggestion
  case class FixtureParam(offerPayment: ActorRef[Command], nodeParams: NodeParams, postman: TypedProbe[Postman.Command], paymentInitiator: TestProbe, routeParams: RouteParams)
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/send/OfferPaymentSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  override def withFixture(test: OneArgTest): Outcome = {
+    val nodeParams = TestConstants.Alice.nodeParams
+    val postman = TypedProbe[Postman.Command]("postman")
+    val paymentInitiator = akka.testkit.TestProbe("paymentInitiator")(system.toClassic)
```
t-bast, line 50: nit:
```suggestion
    val paymentInitiator = TestProbe("paymentInitiator")(system.toClassic)
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/send/OfferPaymentSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+    val paymentRoute = PaymentBlindedRoute(RouteBlinding.create(randomKey(), Seq(merchantKey.publicKey), Seq(hex"7777")).route, PaymentInfo(0 msat, 0, CltvExpiryDelta(0), 0 msat, 1_000_000_000 msat, Features.empty))
+    val invoice = Bolt12Invoice(invoiceRequest, preimage, merchantKey, 1 minute, Features.empty, Seq(paymentRoute))
+    replyTo ! Postman.Response(FinalPayload(TlvStream(OnionMessagePayloadTlv.Invoice(invoice.records)), TlvStream.empty))
+    val send = paymentInitiator.expectMsgType[SendPaymentToNode]
+    assert(send.invoice == invoice)
+
+    val paymentId = UUID.randomUUID()
+    send.replyTo ! paymentId
+    val uuid = probe.expectMsgType[UUID]
+    assert(uuid == paymentId)
```
t-bast, line 85: This isn't testing anything is it? You're both sending the message and verifying you receive it inside the test...I believe you can just remove these 4 lines.

eclair-core/src/test/scala/fr/acinq/eclair/payment/send/OfferPaymentSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+    for (i <- 1 to nodeParams.onionMessageConfig.maxAttempts) {
+      val Postman.SendMessage(_, Recipient(recipientId, _, _), _, message, replyTo, _) = postman.expectMessageType[Postman.SendMessage]
+      assert(recipientId == merchantKey.publicKey)
+      assert(message.get[OnionMessagePayloadTlv.InvoiceRequest].nonEmpty)
+      val Right(invoiceRequest) = InvoiceRequest.validate(message.get[OnionMessagePayloadTlv.InvoiceRequest].get.tlvs)
+      assert(invoiceRequest.isValidFor(offer))
+      replyTo ! Postman.NoReply
+    }
+    probe.expectMsg(NoInvoice)
+    paymentInitiator.expectNoMessage()
```
t-bast, line 107: Always specify a duration in `expectNoMessage`, otherwise the default seems to be 500ms which can quickly aggregate and slow down the test suite for no good reason since we have thousands of tests...

```suggestion
    paymentInitiator.expectNoMessage(50 millis)
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/send/OfferPaymentSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+    assert(message.get[OnionMessagePayloadTlv.InvoiceRequest].nonEmpty)
+    val Right(invoiceRequest) = InvoiceRequest.validate(message.get[OnionMessagePayloadTlv.InvoiceRequest].get.tlvs)
+
+    val preimage = randomBytes32()
+    val paymentRoute = PaymentBlindedRoute(RouteBlinding.create(randomKey(), Seq(merchantKey.publicKey), Seq(hex"7777")).route, PaymentInfo(0 msat, 0, CltvExpiryDelta(0), 0 msat, 1_000_000_000 msat, Features.empty))
+    val invoice = Bolt12Invoice(invoiceRequest, preimage, randomKey(), 1 minute, Features.empty, Seq(paymentRoute))
+    replyTo ! Postman.Response(FinalPayload(TlvStream(OnionMessagePayloadTlv.Invoice(invoice.records)), TlvStream.empty))
+
+    probe.expectMsgType[InvalidInvoice]
+    paymentInitiator.expectNoMessage()
```
t-bast, line 130:
```suggestion
    paymentInitiator.expectNoMessage(50 millis)
```

t-bast (commented):
Looks mostly good to me, some minor comments (please don't rebase, just add new commits to fix those comments).

What's the status of e2e tests with cln? It would be good to know where we stand on cross-compatibility before merging this.

eclair-core/src/main/scala/fr/acinq/eclair/db/pg/PgPaymentsDb.scala
```diff
@@ -71,18 +71,23 @@ class PgPaymentsDb(implicit ds: DataSource, lock: PgLock) extends PaymentsDb wit
...
-          statement.executeUpdate("CREATE TABLE payments.sent (id TEXT NOT NULL PRIMARY KEY, parent_id TEXT NOT NULL, external_id TEXT, payment_hash TEXT NOT NULL, payment_preimage TEXT, payment_type TEXT NOT NULL, amount_msat BIGINT NOT NULL, fees_msat BIGINT, recipient_amount_msat BIGINT NOT NULL, recipient_node_id TEXT NOT NULL, payment_request TEXT, payment_route BYTEA, failures BYTEA, created_at TIMESTAMP WITH TIME ZONE NOT NULL, completed_at TIMESTAMP WITH TIME ZONE)")
+          statement.executeUpdate("CREATE TABLE payments.sent (id TEXT NOT NULL PRIMARY KEY, parent_id TEXT NOT NULL, external_id TEXT, payment_hash TEXT NOT NULL, payment_preimage TEXT, payment_type TEXT NOT NULL, amount_msat BIGINT NOT NULL, fees_msat BIGINT, recipient_amount_msat BIGINT NOT NULL, recipient_node_id TEXT NOT NULL, payment_request TEXT, offer_id TEXT, payer_key TEXT, payment_route BYTEA, failures BYTEA, created_at TIMESTAMP WITH TIME ZONE NOT NULL, completed_at TIMESTAMP WITH TIME ZONE)")
```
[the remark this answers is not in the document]
t-bast: They are, we should add them only for operations that will require frequent reads that filter on those fields, but that's really the case here, isn't it? Since we index on `payment_hash`, we must also index on `offer_id`, otherwise it wouldn't be consistent?

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/OfferPayment.scala
t-bast: Right, that's something we'll need to add to the graph: we're missing feature bits associated to edges and/or vertices. That's something we'll also need when we do PTLCs, we'll need to know which edges can relay PTLC payments, so at some point we'll need to update our `Graph` data to contain those.

t-bast (dismissed):

t-bast:
What's the status of e2e tests with cln? It would be good to know where we stand on cross-compatibility before merging this.

thomash-acinq:
Tests with CLN are successful, I was able to pay a CLN offer from eclair and an eclair offer from CLN. There were two small incompatibilities that are fixed in e8ff99c:
- the format for the reply path had changed in the spec
- we used to require an encrypted_recipient_data for onion messages but it does not really make sense when you send a message to a public node

t-bast (approved):
LGTM!

PR #2481, opened 2022-11-10
Introduce blinded edges and hops

t-bast:
Since blinded routes have to be used from start to end and are somewhat similar to Bolt 11 routing hints, we model them as an abstract single hop during path-finding. This makes it trivial to reuse existing algorithms for blinded payments without any modifications.

There is one subtlety: we previously used scids to identify hops (which is useful to ignore failing ones during retries and track the amount sent to a given edge when using MPP). To minimize code changes, we add a dummy scid to blinded edges to use as short-lived identifiers. They don't represent "real" channels, so we make that explicit even though it requires slightly more code.

We can do something similar in the future for trampoline payments, by adding a virtual edge that represents a full trampoline route: that will let us handle trampoline exactly the same way we handle other payments. This will also remove the hacky need for a `targetNodeId` in payment state machines that is different from the recipient nodeId, which will be a nice clean-up.

There is one thing I'm not satisfied with though in this model: it makes it look like a route could contain blinded hops following by channel hops followed by blinded hops. This isn't possible: routes will always start with a set of channel hops, optionally followed by a single blinded hop (or in the future trampoline hop). There is some opportunity to improve the model to cleanly express that, but I think it makes more sense to attempt it when we refactor trampoline payments.

NB: this builds on top of #2480


eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -264,7 +264,10 @@ object OutgoingPaymentPacket {
 
   /** Build the command to add an HTLC for the given recipient using the provided route. */
   def buildOutgoingPayment(replyTo: ActorRef, upstream: Upstream, paymentHash: ByteVector32, route: Route, recipient: Recipient): Try[OutgoingPaymentPacket] = {
-    val outgoingChannel = route.hops.head.shortChannelId
+    val outgoingChannel = route.hops.head match {
+      case hop: ChannelHop => hop.shortChannelId
+      case _: BlindedHop => ???
```
thomash-acinq, line 269: Implementation missing.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -353,9 +353,9 @@ object MultiPartHandler {
                   } else {
                     implicit val timeout: Timeout = 10.seconds
                     r.router.ask(Router.FinalizeRoute(r.amount, Router.PredefinedNodeRoute(route.nodes))).mapTo[Router.RouteResponse].map(routeResponse => {
-                      val clearRoute = routeResponse.routes.head
-                      val blindedRoute = createBlindedRouteFromHops(clearRoute.hops ++ dummyHops, pathId, nodeParams.channelConf.htlcMinimum, route.maxFinalExpiryDelta.toCltvExpiry(nodeParams.currentBlockHeight))
-                      val paymentInfo = aggregatePaymentInfo(r.amount, clearRoute.hops ++ dummyHops)
+                      val clearHops = routeResponse.routes.head.hops.collect { case h: ChannelHop => h }
```
thomash-acinq, line 356: If somehow we get hops that are not channel hops here, we should log it and fail, not just silently return an invalid invoice.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -65,14 +66,17 @@ object RouteCalculation {
             case edges if edges.nonEmpty && edges.forall(_.nonEmpty) =>
               // select the largest edge (using balance when available, otherwise capacity).
               val selectedEdges = edges.map(es => es.maxBy(e => e.balance_opt.getOrElse(e.capacity.toMilliSatoshi)))
-              val hops = selectedEdges.map(e => ChannelHop(getEdgeRelayScid(d, localNodeId, e), e.desc.a, e.desc.b, e.params))
+              val hops = selectedEdges.map(e => e.params match {
+                case params: ChannelRelayParams => ChannelHop(getEdgeRelayScid(d, localNodeId, e), e.desc.a, e.desc.b, params)
+                case params: HopRelayParams.FromBlindedRoute => BlindedHop(params.dummyId, params.route, params.paymentInfo)
```
thomash-acinq, line 71: Is it actually possible to get a blinded edge here?

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -88,7 +92,12 @@ object RouteCalculation {
                 case None => fr.extraEdges.map(GraphEdge(_)).find(e => e.desc.shortChannelId == shortChannelId && e.desc.a == currentNode).map(_.desc)
               }
               channelDesc_opt.flatMap(c => g.getEdge(c)) match {
-                case Some(edge) => (edge.desc.b, previousHops :+ ChannelHop(getEdgeRelayScid(d, localNodeId, edge), edge.desc.a, edge.desc.b, edge.params))
+                case Some(edge) =>
+                  val hop = edge.params match {
+                    case params: ChannelRelayParams => ChannelHop(getEdgeRelayScid(d, localNodeId, edge), edge.desc.a, edge.desc.b, params)
+                    case params: HopRelayParams.FromBlindedRoute => BlindedHop(params.dummyId, params.route, params.paymentInfo)
```
thomash-acinq, line 98: Same here, blinded route should only come from BOLT12 invoices, they should not come from the router.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
```diff
@@ -382,7 +394,11 @@ object RouteCalculation {
   /** Update used capacity by taking into account an HTLC sent to the given route. */
   private def updateUsedCapacity(route: Route, usedCapacity: mutable.Map[ShortChannelId, MilliSatoshi]): Unit = {
     route.hops.reverse.foldLeft(route.amount) { case (amount, hop) =>
-      usedCapacity.updateWith(hop.shortChannelId)(previous => Some(amount + previous.getOrElse(0 msat)))
+      val id = hop match {
+        case hop: ChannelHop => hop.shortChannelId
+        case hop: BlindedHop => hop.dummyId
```
thomash-acinq, line 399: Why not just put `shortChannelId` in the `ConnectedHop` trait?

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
t-bast: That's on purpose, it's not necessary to make that version of the code work since this is dead code (the code path cannot be triggered). It's done this way to split the implementation work between this PR and the next one, where this is actually implemented.

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
t-bast: At first I thought no, but actually we can if we add them to `extraEdges`, which could potentially make sense?

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
t-bast: But then they are injected via `extraEdges`, which can make sense?

eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
t-bast: Because that leads to bugs! This isn't a real `localAlias`, and by putting it in `ConnectedHop` and not thinking about it, we accidentally pass it on to components that expect an scid that matches one of our channels. I found one such case in one of your earlier PR, so I decided that we should make sure it's very explicit when we use that `dummyId`, to help us catch these issues.

PR #2482, opened 2022-11-10
Send payments to blinded routes

t-bast:
Since blinded routes have to be used from start to end and are somewhat similar to Bolt 11 routing hints, we model them as an abstract single hop during path-finding. This makes it trivial to reuse existing algorithms without any modifications.

We then add support for paying blinded routes. We introduce a new type of recipient for those payments, that uses blinded hops and creates onion payloads accordingly. There is a subtlety in the case where we're the introduction of the blinded route: when that happens we need to decrypt the first payload to figure out where to send the payment.

When we receive a failure from a blinded route, we simply ignore it in retries: we don't know what caused the issue so we assume it's permanent, which makes sense in most cases since we cannot change the relaying parameters (fees and expiry delta are chosen by the recipient).

By building on top of #2480, the change is quite easy to review and doesn't impact the payment state machines much. I've added a lot of e2e tests #2490, where the error handling has (finally) been correctly implemented.

This is an alternative to #2467: what I didn't like in that PR were:

- the new `Partial` payment onion class, which is a clear layering violation and requires reviewers to look at the whole flow of code to verify that this object contains the right data (it's quite foot-gunny because a tlv stream isn't strongly typed, it may records you don't expect...)
- the complexity of `PaymentPacket.scala`, which became very hard to read
- the use of a `ShortChannelId` in blinded hops, which was hiding subtle issues

I tried to build on top of #2467 at first, but when I realized that it was important to add the new abstraction proposed in #2480, it was impossible to do. I did pick a lot of the code from that PR though, most of the actual business logic is the same, it's just the abstractions that are different.


eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
```diff
@@ -0,0 +1,231 @@
...
+object BlindedRecipient {
+  def apply(invoice: Bolt12Invoice, totalAmount: MilliSatoshi, expiry: CltvExpiry, customTlvs: Seq[GenericTlv]): BlindedRecipient = {
+    val extraEdges = invoice.blindedPaths.zip(invoice.blindedPathsInfo).map {
+      case (route, info) =>
+        // We don't know the scids of channels inside the blinded route, but it's useful to have an ID to refer to a given
+        // edge in the graph, so we create a dummy one for the duration of the payment attempt.
+        val dummyId = ShortChannelId.generateLocalAlias()
+        Invoice.BlindedEdge(dummyId, invoice.nodeId, route, info)
+    }
+    BlindedRecipient(invoice.nodeId, invoice.features, totalAmount, expiry, extraEdges, customTlvs)
```
thomash-acinq, line 160: The `nodeId` should be generated randomly to ensure there is no public node with that id.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
```diff
@@ -0,0 +1,231 @@
...
+      val introductionExpiry = expiry + blindedHop.paymentInfo.cltvExpiryDelta
+      val blindedPayloads = PaymentPayloads(introductionAmount, introductionExpiry, introductionPayload +: intermediateBlindedPayloads :+ finalBlindedPayload)
+      if (channelHops.isEmpty) {
+        // We are the introduction node of the blinded route.
+        blindedPayloads
+      } else {
+        channelHops.tail.reverse.foldLeft(blindedPayloads) {
+          case (current, channelHop) =>
+            val payload = NodePayload(channelHop.nodeId, IntermediatePayload.ChannelRelay.Standard(channelHop.shortChannelId, current.amount, current.expiry))
+            PaymentPayloads(current.amount + channelHop.fee(current.amount), current.expiry + channelHop.cltvExpiryDelta, payload +: current.payloads)
```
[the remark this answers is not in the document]
t-bast: Not exactly, because we start with a sequence of payloads (for all blinded hops) instead of just a final payload. But you're right that we can easily change `OutgoingPaymentPacket.buildPayloads` to take a `finalPayloads: Seq[NodePayload]` to reuse it here, I'll do that.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
```diff
@@ -317,18 +239,61 @@ object OutgoingPaymentPacket {
...
+  def buildPayloads(finalAmount: MilliSatoshi, finalExpiry: CltvExpiry, finalPayload: NodePayload, hops: Seq[ChannelHop]): PaymentPayloads = {
+    // We ignore the first hop since the route starts at our node.
+    hops.tail.reverse.foldLeft(PaymentPayloads(finalAmount, finalExpiry, Seq(finalPayload))) {
+      case (current, hop) =>
+        val payload = NodePayload(hop.nodeId, IntermediatePayload.ChannelRelay.Standard(hop.shortChannelId, current.amount, current.expiry))
+        PaymentPayloads(current.amount + hop.fee(current.amount), current.expiry + hop.cltvExpiryDelta, payload +: current.payloads)
     }
   }
 
+  private def buildPayloads(privateKey: PrivateKey, paymentHash: ByteVector32, route: Route, recipient: Recipient): Try[(ShortChannelId, Option[PublicKey], PaymentPayloads)] = {
```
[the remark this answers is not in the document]
t-bast: I agree! I should rework that, I'll find a better name.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
t-bast: That's a good question, I'm not sure yet. If we get an invoice with a `nodeId` that matches a node in the public graph, why couldn't we just pay it without using route blinding if we're able to find a normal route? If the recipient doesn't want that to happen, shouldn't it be their responsibility to generate a random `nodeId`?

It would be nice to keep the real `nodeId` we were given in the `Recipient` to get better accounting of who we paid. Also, the code currently doesn't support making the payment if we found a normal route instead of using the blinded hops, but that should be very simple to add (if we think it's a good idea). What do you think?

t-bast:
> Can you add some tests to check that it actually works?

I added more tests in https://github.com/ACINQ/eclair/pull/2482/commits/55f69600effe5833ba2921ed881a7ccc1b142c37, let me know if there are specific test cases that would be useful that I missed. I didn't add yet an e2e test because it would require the code from #2471 

I also had to add utility test functions to generate blinded routes / bolt12 invoices, those should be cleaned up when we have #2471

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
thomash-acinq: We must use the blinded route, that was the whole point of making it mandatory. If the recipient doesn't want privacy, it can always use a blinded route with no hop.
And blinded route doesn't just provide privacy, it also has encrypted data that the recipient may need to finalize the payment.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferCodecs.scala
```diff
@@ -40,9 +40,11 @@ object OfferCodecs {
   private val absoluteExpiry: Codec[AbsoluteExpiry] = tlvField(tu64overflow.as[TimestampSecond].as[AbsoluteExpiry])
 
   private val blindedNodeCodec: Codec[BlindedNode] = (("nodeId" | publicKey) :: ("encryptedData" | variableSizeBytes(uint16, bytes))).as[BlindedNode]
-
-  private val pathCodec: Codec[BlindedRoute] = (("firstNodeId" | publicKey) :: ("blinding" | publicKey) :: ("path" | listOfN(uint8, blindedNodeCodec).xmap[Seq[BlindedNode]](_.toSeq, _.toList))).as[BlindedRoute]
-
+  private val blindedNodesCodec: Codec[Seq[BlindedNode]] = listOfN(uint8, blindedNodeCodec).exmap(
+    nodes => if (nodes.length <= 1) Attempt.Failure(Err("blinded route must contain at least two nodes")) else Attempt.Successful(nodes.toSeq),
+    nodes => if (nodes.length <= 1) Attempt.Failure(Err("blinded route must contain at least two nodes")) else Attempt.Successful(nodes.toList),
```
thomash-acinq, line 45: Why are you adding this artificial restriction?  Blinded routes with zero hop work just fine and are more efficient. Most big merchants are public and won't benefit from the increased privacy of a multi-hop blinded route and would prefer to use a minimal blinded route.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
```diff
@@ -0,0 +1,231 @@
...
+/** A payment recipient that hides its real identity using route blinding. */
+case class BlindedRecipient(nodeId: PublicKey,
+                            features: Features[InvoiceFeature],
+                            totalAmount: MilliSatoshi,
+                            expiry: CltvExpiry,
+                            extraEdges: Seq[Invoice.BlindedEdge],
+                            customTlvs: Seq[GenericTlv] = Nil) extends Recipient {
+  override def buildPayloads(paymentHash: ByteVector32, route: Route): Try[PaymentPayloads] = {
+    extractRoute(route).map { case (channelHops, blindedHop) =>
+      val introductionPayload = NodePayload(blindedHop.route.introductionNodeId, OutgoingBlindedPerHopPayload.createIntroductionPayload(blindedHop.route.introductionNode.encryptedPayload, blindedHop.route.introductionNode.blindingEphemeralKey))
```
thomash-acinq, line 121: I've added some tests and it's not working.
One of the problems is that we need to encrypt the onion with the blinded node id:
```suggestion
      val introductionPayload = NodePayload(blindedHop.route.blindedNodeIds.head, OutgoingBlindedPerHopPayload.createIntroductionPayload(blindedHop.route.introductionNode.encryptedPayload, blindedHop.route.introductionNode.blindingEphemeralKey))
```
But now I get an incorect HMAC error.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferCodecs.scala
t-bast: That's a very good point. I was under the impression that we needed to use a dummy hop in such cases, but we actually don't. Rusty's recent comments made it look like cln doesn't handle a 0-hop blinded path, but we should ask them and update this depending on their answer.

t-bast: That was discussed during yesterday's spec meeting, and it was concluded that we cannot (and most likely should not) prevent recipients from revealing their identity if they want to, so we should allow 0-hop blinded routes. I'll update the PR to take that into account.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
```diff
@@ -0,0 +1,227 @@
...
+      val intermediateBlindedPayloads = blindedHop.route.subsequentNodes.dropRight(1).map(n => NodePayload(n.blindedPublicKey, OutgoingBlindedPerHopPayload.createIntermediatePayload(n.encryptedPayload)))
+      val finalBlindedPayload = NodePayload(blindedHop.route.blindedNodes.last.blindedPublicKey, OutgoingBlindedPerHopPayload.createFinalPayload(route.amount, totalAmount, expiry, blindedHop.route.blindedNodes.last.encryptedPayload, customTlvs))
```
thomash-acinq, line 123: This does not work for a zero-hop route.

thomash-acinq (commented):
You should add the tests too now that #2499 has been merged, especially the zero-hop test which would not pass currently.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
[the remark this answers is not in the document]
t-bast: Fixed in the latest version.

eclair-core/src/main/scala/fr/acinq/eclair/payment/PaymentPacket.scala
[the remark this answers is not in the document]
t-bast: Fixed in the latest version.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
t-bast: Fixed in the latest version.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferCodecs.scala
t-bast: I removed this restriction in the latest version.

eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
t-bast: Fixed in the latest version. I added the e2e on the following PR (#2490) because failure handling was really incomplete, so I figured it made sense to add all the e2e tests once we have everything working.

thomash-acinq (approved):
Looks all good.

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -566,7 +566,10 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
           randomKey(),
           randomKey(),
           intermediateNodes.map(OnionMessages.IntermediateNode(_)),
-          destination match { case Left(key) => OnionMessages.Recipient(key, None) case Right(route) => OnionMessages.BlindedPath(route) },
```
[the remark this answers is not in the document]
t-bast: I initially modified this part a long time ago in the first versions of payment recipients, but that code has disappeared since then, so I'll revert that.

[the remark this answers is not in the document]
t-bast: Reverted in https://github.com/ACINQ/eclair/pull/2482/commits/855ce3b5ab0e54a70549a8bc62d1c941386ab2b1

thomash-acinq (approved):

PR #2483, opened 2022-11-12
`findroute*`: relax ignored channels validation for local channels

rorp:
The current implementation of the API validation fails to correctly validate temporarily disabled or offline local channels, because they are not present in the network graph.

This PR allows to exclude any open local channel from route finding.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -556,11 +556,11 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
     } else {
       for {
         (localChannelDescs, nonLocalChannelIds) <- getLocalChannelDescs
-        channelsMap <- (appKit.router ? GetChannelsMap).mapTo[SortedMap[ShortChannelId, PublicChannel]]
+        routerData <- (appKit.router ? GetRouterData).mapTo[Router.Data]
       } yield {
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -531,17 +531,36 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
...
+        }
+      }
+    }
+
     if (shortChannelIds.isEmpty) {
       Future.successful(Set.empty)
     } else {
       for {
+        (localChannelDescs, nonLocalChannelIds) <- getLocalChannelDescs
         channelsMap <- (appKit.router ? GetChannelsMap).mapTo[SortedMap[ShortChannelId, PublicChannel]]
```
t-bast, line 559: I believe a much simpler fix would be to do `(appKit.router ? GetRouterData)` and then use `d.resolve` which will look at both private and public channels and correctly handle scid aliases. This would require only a few lines of code and should be more future-proof.

t-bast: I don't understand what you did in https://github.com/ACINQ/eclair/pull/2483/commits/92c416986e08aa59f0b49e9474527cb6cabc547f

Once you cal `GetRouterData`, you don't need `getLocalChannelDescs` at all anymore, you have all the data necessary and can remove most of your added code?

t-bast:
Superseded by #2523

Issue #2475, opened 2022-11-03
`listinvoices` returns an incomplete list


rorp:
Steps to reproduce:

1. start Eclair node
2. create an invoice
3. run 
```bash 
eclair-cli listinvoices 
```
4. verify that the invoice created at step 2 is in the list
5. wait a second
6. create another invoice
7. run 
```bash 
eclair-cli listinvoices 
```
Expected result: both invoices should be in the list 
Actual result: only the first invoice is in the list

Apparently, the default values of `--from` and `--to` parameters get cached and get reset only after restart.
Reproduceable on current `master`.

t-bast:
> Apparently, the default values of --from and --to parameters get cached and get reset only after restart.

Huh, that's a very good catch, this is a bug indeed! Thanks for raising this, we should fix that and include it in the next release.

PR #2484, opened 2022-11-14
Fix static default `from` and `to` pagination

t-bast:
The default values for the `from` and `to` parameters of some API requests were generated once when the eclair node starts instead of being recomputed for every request.

We also add `count` and `skip` argument to `listPendingInvoices` since it was added to `listInvoices`.

Fixes https://github.com/ACINQ/eclair/issues/2475

### DIFF
--- a/eclair-node/src/main/scala/fr/acinq/eclair/api/directives/ExtraDirectives.scala
+++ b/eclair-node/src/main/scala/fr/acinq/eclair/api/directives/ExtraDirectives.scala
@@ -16,7 +16,7 @@
 
 package fr.acinq.eclair.api.directives
 
-import akka.http.scaladsl.common.{NameDefaultUnmarshallerReceptacle, NameReceptacle, NameUnmarshallerReceptacle}
+import akka.http.scaladsl.common.{NameReceptacle, NameUnmarshallerReceptacle}
 import akka.http.scaladsl.marshalling.ToResponseMarshaller
 import akka.http.scaladsl.model.StatusCodes.NotFound
 import akka.http.scaladsl.model.{ContentTypes, HttpResponse}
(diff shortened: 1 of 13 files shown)


eclair-node/src/main/scala/fr/acinq/eclair/api/directives/ExtraDirectives.scala
```diff
@@ -43,17 +43,16 @@ trait ExtraDirectives extends Directives {
   val nodeIdFormParam: NameReceptacle[PublicKey] = "nodeId".as[PublicKey]
   val nodeIdsFormParam: NameUnmarshallerReceptacle[List[PublicKey]] = "nodeIds".as[List[PublicKey]](pubkeyListUnmarshaller)
   val paymentHashFormParam: NameUnmarshallerReceptacle[ByteVector32] = "paymentHash".as[ByteVector32](sha256HashUnmarshaller)
-  // we limit default values to avoid accidentally reading too much data from the DB
-  val fromFormParam: NameDefaultUnmarshallerReceptacle[TimestampSecond] = "from".as[TimestampSecond](timestampSecondUnmarshaller).?(TimestampSecond.now() - 1.day)
-  val toFormParam: NameDefaultUnmarshallerReceptacle[TimestampSecond] = "to".as[TimestampSecond](timestampSecondUnmarshaller).?(TimestampSecond.now())
+  val fromFormParam: NameUnmarshallerReceptacle[TimestampSecond] = "from".as[TimestampSecond](timestampSecondUnmarshaller)
+  val toFormParam: NameUnmarshallerReceptacle[TimestampSecond] = "to".as[TimestampSecond](timestampSecondUnmarshaller)
```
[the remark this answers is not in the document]
t-bast: That's kind of what I'm doing, do you mean that I should just remove these two `vals` and directly do `"from".as[TimestampSecond]` inside the body of `withFromToTimestamps`?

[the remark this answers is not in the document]
pm47: No need to even define `withFromToTimestamps`, just `s/val/def` and the bug should be fixed I think.

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/ExtraDirectives.scala
```diff
@@ -55,6 +52,13 @@ trait ExtraDirectives extends Directives {
   val countFormParam: NameReceptacle[Int] = "count".as[Int]
   val skipFormParam: NameReceptacle[Int] = "skip".as[Int]
 
+  // @formatter:off
+  // We limit default values to avoid accidentally reading too much data from the DB.
+  // We cannot use a val, otherwise the default values would not be recomputed for every request.
+  def fromFormParam(): NameDefaultUnmarshallerReceptacle[TimestampSecond] = "from".as[TimestampSecond](timestampSecondUnmarshaller).?(TimestampSecond.now() - 1.day)
+  def toFormParam(): NameDefaultUnmarshallerReceptacle[TimestampSecond] = "to".as[TimestampSecond](timestampSecondUnmarshaller).?(TimestampSecond.now())
```
pm47, line 59: Nit: did you put trailing parentheses because those methods depend on `TimestampSecond.now()` which has parenthese, therefore . FWIW in theory neither should have parentheses because they are side-effect-free.

pm47 (approved):

eclair-node/src/main/scala/fr/acinq/eclair/api/directives/ExtraDirectives.scala
t-bast: My rule of thumb is to put parentheses whenever a function isn't a pure function. This isn't a pure function because invoking it repeatedly yields different results. Whenever there's no parenthesis, from the caller's point of view it looks like a `val` that should return the same result every time, don't you think?

pm47: Makes sense, it's a stricter definition of side effect, probably more correct than mine.

PR #2485, opened 2022-11-14
Fix circular dependency

thomash-acinq:
Fixes a circular dependency between `OfferCodecs.scala` and `LightningMessageCodecs.scala` by putting feature codecs in `CommonCodecs.scala` instead.

This bug did not prevent compilation of the whole project but did prevent running tests for Bolt12Invoice independently of the rest.


t-bast (approved):
It does fix the issue for me :+1: 

I *think* this fix means the issue was with the `Feature` trait which isn't sealed (to allow plugins to add new features that don't need to be supported by eclair-core).

I'm fine with fixing it this way.

PR #2486, opened 2022-11-14
Explicitly request bech32 static remotekey

t-bast:
When using `static_remotekey`, we ask `bitcoind` to generate a receiving address and send the corresponding public key to our peer. We previously assumed that `bitcoind` used `bech32`, which may change since #2466.

We need to explicitly tell `bitcoind` that we want to use a p2wpkh address in that case, otherwise it may generate a `p2tr` address, we will use that public key to get our funds back, but `bitcoind` will not detect that it can spend this output (because it only expects this public key to be used for a taproot address).

Thanks @sstone for hinting at this problem.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/OnChainWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/OnChainWallet.scala
@@ -84,11 +84,8 @@ trait OnChainAddressGenerator {
    */
   def getReceiveAddress(label: String = "")(implicit ec: ExecutionContext): Future[String]
 
-  /**
-   * @param receiveAddress if provided, will extract the public key from this address, otherwise will generate a new
-   *                       address and return the underlying public key.
-   */
-  def getReceivePubkey(receiveAddress: Option[String] = None)(implicit ec: ExecutionContext): Future[PublicKey]
+  /** Generate a p2wpkh wallet address and return the corresponding public key. */
+  def getReceivePubkey()(implicit ec: ExecutionContext): Future[PublicKey]
 
 }
 
--- a/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
@@ -21,7 +21,7 @@ import akka.pattern.pipe
 import akka.testkit.TestProbe
 import fr.acinq.bitcoin
 import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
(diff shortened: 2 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/blockchain/OnChainWallet.scala
```diff
@@ -84,11 +84,8 @@ trait OnChainAddressGenerator {
    */
   def getReceiveAddress(label: String = "")(implicit ec: ExecutionContext): Future[String]
 
-  /**
-   * @param receiveAddress if provided, will extract the public key from this address, otherwise will generate a new
-   *                       address and return the underlying public key.
-   */
-  def getReceivePubkey(receiveAddress: Option[String] = None)(implicit ec: ExecutionContext): Future[PublicKey]
+  /** Generate a p2wpkh wallet address and return the corresponding public key. */
+  def getReceivePubkey()(implicit ec: ExecutionContext): Future[PublicKey]
```
t-bast, line 88: Should we go further and rename the function to make it even more explicit? What name do you think would be best?

sstone: Yes I think the name should be explicit, since `static_remotekey` explicitly sets the `to-remote` output to a P2WPKH address and this method is only used with `static_remotekey`. Maybe `getP2wpkhPubkey()` ?
(We would also add a parameter to specify which type of key we want bitcoind to generate but we can always do that if/when we need it).

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
```diff
@@ -938,16 +938,20 @@ class BitcoinCoreClientSpec extends TestKitBaseClass with BitcoindService with A
     sender.expectMsg(tx1)
   }
 
-  test("compute pubkey from a receive address") {
+  test("get pubkey for p2wpkh receive address") {
```
sstone, line 941: we could run this test with bitcoin core set to use bech32m for both receive and change addresses ?

sstone (commented):
We probably also want to update our `addressToPublicKeyScript()` method which should call the method provided by `bitcoin-kmp`

eclair-core/src/main/scala/fr/acinq/eclair/blockchain/OnChainWallet.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2486/commits/b950b79a51c9a9e15ca0418838ceabb5dd50d79a

eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
t-bast: Done in https://github.com/ACINQ/eclair/pull/2486/commits/6bc5a1b49605eb307627af86489e61c4741a9c97

sstone (approved):

sstone (approved):

PR #2488, opened 2022-11-15
Make `eclair-cli -s` display short channel ids correctly

sstone:
in "short/compact mode" (-s), eclair-cli was not getting short channel ids from the right place and they were always set to null.


t-bast (approved):
Good catch!

PR #2490, opened 2022-11-16
Fix blinded route failure messages

t-bast:
This PR correctly implements the last three commits of the route blinding PR:

- https://github.com/lightning/bolts/pull/765/commits/eed2ab0c30ad7f93e3b2641ca9d7ade32f3d121d
- https://github.com/lightning/bolts/pull/765/commits/274c0c0e08c65e9e0964663e2061d73ff80997fe
- https://github.com/lightning/bolts/pull/765/commits/ce57116cc71ef8660ef651f018a42cbbb4662f46

In particular, failure messages were incorrectly implemented in #2408: we need to translate *all* downstream failures to a unique failure. This wasn't caught in the PR and was lacking tests (mostly because it required more route blinding glue to be implemented before it could be properly tested).

We also fix an issue we had with `update_fail_malformed_htlc`: when we received that kind of error from the downstream node, we also returned an `update_fail_malformed_htlc` error upstream, whereas the specification says we must convert it to an `update_fail_htlc`.

We add many e2e tests for the overall blinded payments feature :tada: :tada:


t-bast:
Note to self: the `buildHtlcFailure` function in `PaymentPacket` must use the blinded private key when generating errors, this currently isn't handled.

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/payment/BlindedPaymentSpec.scala
```diff
@@ -0,0 +1,319 @@
...
+      assert(getChannelData(bob, channelId_bc_2).asInstanceOf[DATA_NORMAL].shortIds.real.isInstanceOf[RealScidStatus.Final])
+    }
+  }
+
+  def createInvoice(recipient: MinimalNodeFixture, amount: MilliSatoshi, routes: Seq[ReceivingRoute], sender: TestProbe): Bolt12Invoice = {
+    val offer = Offer(None, "test", recipient.nodeParams.nodeId, Features.empty, recipient.nodeParams.chainHash)
+    val invoiceReq = InvoiceRequest(offer, amount, 1, Features.empty, randomKey(), recipient.nodeParams.chainHash)
+    sender.send(recipient.paymentHandler, MultiPartHandler.ReceiveOfferPayment(randomKey(), offer, invoiceReq, routes, recipient.router))
+    val invoice = sender.expectMsgType[Bolt12Invoice]
+    assert(invoice.nodeId != recipient.nodeParams.nodeId)
```
thomash-acinq, line 104: Why this check? You create an offer with `recipient.nodeParams.nodeId`, it would actually be a bug if the corresponding invoice didn't have the same node id.

thomash-acinq: Yes, the key provided must be the private corresponding to the node id in the offer (if not, the signature will be invalid and the payer will cancel the payment). I don't think we would gain anything with `OfferDetails(nodeKey: PrivateKey, offer: Offer)`, we just need to make it clearer that the key must match the offer, and maybe add a `require` in the `Bolt12Invoice` constructor.

t-bast: Thanks for the details, that makes sense. This isn't related to the current PR though, is it? I think it makes sense to make that change in the PR where we will actually generate offers and invoices in response to invoice requests, doesn't it? Because that's where we'll start generating these random receiver keys and will have to store them in the DB.

thomash-acinq (approved):

eclair-core/src/test/scala/fr/acinq/eclair/integration/basic/payment/BlindedPaymentSpec.scala
thomash-acinq: Fine

PR #2491, opened 2022-11-16
Add singleton actor `AsyncPaymentTriggerer` to monitor when the receiver of an async payment reconnects 

remyers:
The new `AsyncPaymentTriggerer` singleton actor spawns only a single `PeerReadyNotifier` actor for each peer that is watched. This prevents multiple actors from redundantly polling for when the same peer reconnects. 

Each `Watcher` for a given peer triggers a specific async payment (managed by a `NodeRelay` actor) and has a unique timeout (in block height). The `NodeRelay` actor will receive an `AsyncPaymentTriggered` when the target peer reconnects, or `AsyncPaymentTimeout` if the timeout block height is reached.

This PR requires a change to allow `PeerReadyNotifier` to have an infinite timeout so that `AsyncPaymentTriggerer` can instead manage multiple  timeouts for a given peer.

The `AsyncPaymentTriggerer` singleton must be created before the `Relayer` actor is created, but requires a reference to the singleton `Switchboard` actor to spawn `PeerReadyNotifier` actors. Because the swtichboard is indirectly instantiated with the `Relayer` reference (via `PeerFactory` and `ChannelFactory`), the `AsyncPaymentTriggerer` actor must be further initialized by sending a  `Start` message with the `Switchboard` actor reference before being used.

In the future the `AsyncPaymentTriggerer` actor can be extended to watch for onion messages from async payment receivers, instead of only watching for local peer re-connections. In the onion message case, the `paymentHash` will be used to identify which `Watcher` to trigger.


eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,122 @@
...
+    val peerReadyResultAdapter = context.messageAdapter[PeerReadyNotifier.Result](WrappedPeerReadyResult)
+    context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[CurrentBlockHeight](WrappedCurrentBlockHeight))
+
+    Behaviors.receiveMessage[Command] {
+      case Watch(replyTo, remoteNodeId, paymentHash, timeout) =>
+        triggers.get(remoteNodeId) match {
+          case None =>
+            // add a new trigger
+            val notifier = context.spawn(Behaviors.supervise(PeerReadyNotifier(remoteNodeId, switchboard, Left(Duration.Inf)))
+              .onFailure(SupervisorStrategy.restart), s"peer-ready-notifier-$remoteNodeId-$timeout")
```
t-bast, line 99: I don't think the supervisor strategy should be restart, it is useless because the `PeerReadyNotifier` will be waiting for `NotifyWhenPeerReady` and won't actually be useful. I think we should instead:

- set the supervisor strategy to `stop` with an error log (this really shouldn't happen and indicates a bug)
- use `context.watchWith(notifier, CancelPeerPayments(remoteNodeId))` where `CancelPeerPayments` is a new `Command` that cancels the pending payments for that peer and removes it from the map

We could also do nothing and just wait for the payments to timeout, but it's a bit wasteful, we know they won't be triggered at that point so it's better if we cancel it to give some feedback to the sender.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,122 @@
...
+  }
+}
+
+private class AsyncPaymentTriggerer(context: ActorContext[Command]) {
+  import AsyncPaymentTriggerer._
+
+  case class Watcher(replyTo: ActorRef[Result], timeout: BlockHeight, paymentHash: ByteVector32) {
+    def expired(currentBlockHeight: BlockHeight): Boolean = timeout <= currentBlockHeight
+  }
+  case class AsyncPaymentTrigger(notifier: ActorRef[PeerReadyNotifier.Command], watchers: Set[Watcher]) {
```
t-bast, line 63: I struggled at first because the names of these two classes are quite misleading. The `Watcher` is actually just a payment, and the `AsyncPaymentTrigger` is just a remote peer for which we have pending payments. I renamed these two classes in https://github.com/ACINQ/eclair/commit/ac5118b0b5dc1c1f543bca076a43aba5f4ea9ef3, let me know what you think.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,122 @@
...
+    def update(currentBlockHeight: BlockHeight): Option[AsyncPaymentTrigger] = {
+      // notify watchers that timeout occurred before offline peer reconnected
```
t-bast, line 65: When the code is simple and self-explanatory, we tend to avoid comments, they aren't useful at all and risk becoming obsolete if the code below gets updated but not the comment. We do add comments for non-trivial things though, such as the fact that the `PeerReadyNotifier` stops itself. I removed useless comments in https://github.com/ACINQ/eclair/commit/ac5118b0b5dc1c1f543bca076a43aba5f4ea9ef3

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggererSpec.scala
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  test("duplicate watches should emit only one trigger") { f =>
+    import f._
+
+    // create two identical watches
+    triggerer ! Watch(probe.ref, remoteNodeId, paymentHash = ByteVector32.Zeroes, timeout = BlockHeight(100))
+    assert(switchboard.expectMessageType[GetPeerInfo].remoteNodeId == remoteNodeId)
+    triggerer ! Watch(probe.ref, remoteNodeId, paymentHash = ByteVector32.Zeroes, timeout = BlockHeight(100))
+    switchboard.expectNoMessage()
```
t-bast, line 63: It isn't really useful to test the behavior of the underlying `PeerReadyNotifier`, it is already covered by the `PeerReadyNotifierSpec`. Calls to `expectNoMessage()` can be expensive when we make too many of them, because they slow down the test suite (which we fight hard to keep below 20 minutes on CI). I removed some of these redundants checks in https://github.com/ACINQ/eclair/commit/ac5118b0b5dc1c1f543bca076a43aba5f4ea9ef3

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,122 @@
...
+    Behaviors.receiveMessage[Command] {
+      case Start(switchboard) => watching(switchboard, Map())
+      case m => context.log.error(s"received unhandled message ${m.getClass.getSimpleName} before Start received.")
+        Behaviors.same
+    }
+  }
+
+  private def watching(switchboard: ActorRef[Switchboard.GetPeerInfo], triggers: Map[PublicKey, AsyncPaymentTrigger]): Behavior[Command] = {
+    val peerReadyResultAdapter = context.messageAdapter[PeerReadyNotifier.Result](WrappedPeerReadyResult)
+    context.system.eventStream ! EventStream.Subscribe(context.messageAdapter[CurrentBlockHeight](WrappedCurrentBlockHeight))
```
t-bast, line 91: You are re-subscribing to the event stream every time you update the internal state and call `watching`, which isn't what we want. I extract that to be done in the initialization in https://github.com/ACINQ/eclair/commit/ac5118b0b5dc1c1f543bca076a43aba5f4ea9ef3

t-bast (commented):
Looks good to me, the design is clean and test coverage looks good.
I did some refactoring and renaming in https://github.com/ACINQ/eclair/commit/ac5118b0b5dc1c1f543bca076a43aba5f4ea9ef3, let me know what you think.

I'll update #2464 to allow disabling the timeout, I think it's useful to have. I'll do it slightly differently than what you did, I think using an `Option` is more explicit than an infinite duration.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,122 @@
...
+    Behaviors.receiveMessage[Command] {
+      case Watch(replyTo, remoteNodeId, paymentHash, timeout) =>
+        triggers.get(remoteNodeId) match {
+          case None =>
+            // add a new trigger
+            val notifier = context.spawn(Behaviors.supervise(PeerReadyNotifier(remoteNodeId, switchboard, Left(Duration.Inf)))
+              .onFailure(SupervisorStrategy.restart), s"peer-ready-notifier-$remoteNodeId-$timeout")
+            notifier ! NotifyWhenPeerReady(peerReadyResultAdapter)
+            val newTrigger = AsyncPaymentTrigger(notifier, Set(Watcher(replyTo, timeout, paymentHash)))
+            watching(switchboard, triggers + (remoteNodeId -> newTrigger))
```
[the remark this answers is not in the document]
remyers: Oh great! I knew something felt wrong about the way I did it.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,122 @@
...
+import fr.acinq.eclair.io.PeerReadyNotifier.NotifyWhenPeerReady
+import fr.acinq.eclair.io.{PeerReadyNotifier, Switchboard}
+import fr.acinq.eclair.payment.relay.AsyncPaymentTriggerer.Command
+
+import scala.concurrent.duration.Duration
+
+/**
+ * This actor waits for an async payment receiver to become ready to receive a payment or for a block timeout to expire.
+ * If the receiver of the payment is a connected peer, spawn a PeerReadyNotifier actor.
+ * TODO: If the receiver is not a connected peer, wait for a `ReceiverReady` onion message containing the specified paymentHash.
```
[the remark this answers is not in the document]
remyers: Gotcha - it's a habit I need to break. :-)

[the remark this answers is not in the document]
remyers: My only defense is that it explains why we include `paymentHash` in `Watch` and `Payment` case classes, even though that field is not used (yet).

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
remyers: Yes - I agree this is a clearer naming scheme.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
remyers: That makes sense; implemented in [839f406](https://github.com/ACINQ/eclair/pull/2491/commits/839f406a4ccf527c19b1419bddba9e6a808b2454). Instead of `CancelPeerPayments` I went with a private command called `NotifierStopped`. Do we also want a public `CancelPeerPayments` command to do the same thing, but without the error in the log?

remyers: I'm still trying to understand the different supervision strategies.. if we used `resume` instead of `restart` it would resume in its previous state and resend/resubscribe to events to watch for the peer to be ready, but since this would be unexpected and not guaranteed to be cleanly handled by `PeerReadyNotifier` it's better to use `stop`?

remyers: I'd like to add a simple test that verifies the probe receives a timeout when `AsyncPaymentTriggerer` receives the `NotifierStopped` message, but since the message is private,  I can't send it from the test framework. I also can't get a ref to the internally spawned `PeerReadyNotifier`. Any tips on how create a test for that?

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
remyers: :+1:

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,127 @@
...
+      case WrappedPeerReadyResult(PeerReadyNotifier.PeerReady(remoteNodeId, _)) =>
+        // notify watcher that destination peer is ready to receive async payments; PeerReadyNotifier will stop itself
+        peers.get(remoteNodeId).foreach(_.trigger())
+        watching(peers - remoteNodeId)
+      case NotifierStopped(remoteNodeId) =>
+        peers.get(remoteNodeId) match {
+          case None => Behaviors.same
+          case Some(peer) => peer.timeout()
+            context.log.error(s"PeerReadyNotifier stopped unexpectedly while watching node $remoteNodeId.")
+            watching(peers - remoteNodeId)
```
t-bast, line 122: nit: when using a multi-line block of code, put all statements in the same block:

```suggestion
          case Some(peer) =>
            context.log.error(s"PeerReadyNotifier stopped unexpectedly while watching node $remoteNodeId.")
            peer.timeout()
            watching(peers - remoteNodeId)
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -101,8 +101,8 @@ object NodeRelay {
           context.messageAdapter[MultiPartPaymentFSM.MultiPartPaymentFailed](WrappedMultiPartPaymentFailed)
           context.messageAdapter[MultiPartPaymentFSM.MultiPartPaymentSucceeded](WrappedMultiPartPaymentSucceeded)
         }.toClassic
-        val incomingPaymentHandler = context.actorOf(MultiPartPaymentFSM.props(nodeParams, paymentHash, totalAmountIn, mppFsmAdapters))
-        new NodeRelay(nodeParams, parent, register, relayId, paymentHash, nodeRelayPacket.outerPayload.paymentSecret, context, outgoingPaymentFactory)
+        val incomingPaymentHandler: ActorRef = context.actorOf(MultiPartPaymentFSM.props(nodeParams, paymentHash, totalAmountIn, mppFsmAdapters))
```
t-bast, line 104: nit: we don't need the type annotation here, it's just noise since `context.actorOf` is quire self-explanatory, remove? 

```suggestion
        val incomingPaymentHandler = context.actorOf(MultiPartPaymentFSM.props(nodeParams, paymentHash, totalAmountIn, mppFsmAdapters))
```

t-bast (commented):
LGTM, we should be able to integrate this once we integrate #2464, good job :+1:

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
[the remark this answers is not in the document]
t-bast: Right, in that case a normal comment (not a TODO) around the case class that contains the `paymentHash` could make sense, to explain that even though it's not strictly necessary today it is more future-proof.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
t-bast: > Do we also want a public CancelPeerPayments command to do the same thing, but without the error in the log?

I think that's something we may add later when we implement the ability for the sender to cancel a pending async payments. It's a bit early to add it right now.

> I'd like to add a simple test that verifies the probe receives a timeout when AsyncPaymentTriggerer receives the NotifierStopped message, but since the message is private, I can't send it from the test framework.

Instead of making the message private, you can make it `private[relay]`, which will make it accessible from tests.
It is a bit leaky, but it makes testing much easier, so we've already used that trick several time across the codebase.
Since it's still marked `private[some-scope]` it is obvious to readers that this shouldn't normally be used, so it's somewhat ok.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
remyers: fixed in [3ed6ee8](https://github.com/ACINQ/eclair/pull/2491/commits/3ed6ee86f501be5a7484e35ecbe6ba1f6c0a41c3)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
remyers: Thanks! I added the test in [026ea4e](https://github.com/ACINQ/eclair/pull/2491/commits/026ea4ed25c00b5c36da6f2b773a2004b69905d9)

remyers:
Do we need to add `cancelasyncpayment` to the API? There's currently no way for a trampoline node to query for the list of pending payments. Trampoline nodes should not have to manage this list. Perhaps I should roll back [0c7b049](https://github.com/ACINQ/eclair/pull/2491/commits/0c7b049ad7a8913546838ab3f9040082735197a9).

Otoh, we will need a way for a node that creates an async payment to cancel a pending payment queued by the trampoline node they designated to hold it. This will likely be accomplished with a custom onion message, but could also be a custom Lightning Message.

t-bast:
> Do we need to add cancelasyncpayment to the API? There's currently no way for a trampoline node to query for the list of pending payments. Trampoline nodes should not have to manage this list. Perhaps I should roll back [0c7b049](https://github.com/ACINQ/eclair/pull/2491/commits/0c7b049ad7a8913546838ab3f9040082735197a9).

I think you should indeed rollback https://github.com/ACINQ/eclair/pull/2491/commits/0c7b049ad7a8913546838ab3f9040082735197a9

I think there is a confusion between the _sender node_ and the _relaying node_ here. Cancelling an async payment is done by the _sender node_ telling the _relaying node_ to abort the payment via a custom lightning message or an onion message. The API is for the _relaying node_ operator, but it's not their decision to cancel async payments.

I don't think you should add anything else to this PR, it's just making it harder to review. Follow-up work (e.g. adding the custom `cancel_async_payment` lightning message) should be done in follow-up PRs.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,142 @@
...
+    def update(currentBlockHeight: BlockHeight): Option[PeerPayments] = {
+      val expiredPayments = pendingPayments.filter(_.expired(currentBlockHeight))
+      expiredPayments.foreach(e => e.replyTo ! AsyncPaymentTimeout)
+      updatePaymentsOrStop(pendingPayments.removedAll(expiredPayments))
+    }
+
+    def cancel(paymentHash: ByteVector32): Option[PeerPayments] = {
+      val canceledPayment = pendingPayments.find(_.paymentHash == paymentHash)
+      if (canceledPayment.isDefined) canceledPayment.get.replyTo ! AsyncPaymentCanceled
+      updatePaymentsOrStop(pendingPayments.removedAll(canceledPayment))
```
t-bast, line 79: There may be multiple payments for the same payment hash, you should use `filter` instead of `find`:

```suggestion
      val canceledPayments = pendingPayments.filter(_.paymentHash == paymentHash)
      canceledPayments.foreach(_.replyTo ! AsyncPaymentCanceled)
      updatePaymentsOrStop(pendingPayments.removedAll(canceledPayments))
```

It's worth adding a unit test for this scenario.

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  private def watching(peers: Map[PublicKey, PeerPayments]): Behavior[Command] = {
+    Behaviors.receiveMessagePartial {
+      case Watch(replyTo, remoteNodeId, paymentHash, timeout) =>
+        peers.get(remoteNodeId) match {
+          case None =>
+            val notifier = context.spawn(
+              Behaviors.supervise(PeerReadyNotifier(remoteNodeId, switchboard, None)).onFailure(SupervisorStrategy.stop),
+              s"peer-ready-notifier-$remoteNodeId",
+            )
```
t-bast, line 108: I think it would be safer to spawn anonymous actors here:
```suggestion
            val notifier = context.spawnAnonymous(Behaviors.supervise(PeerReadyNotifier(remoteNodeId, switchboard, None)).onFailure(SupervisorStrategy.stop))
```

Otherwise we can run into the following race condition, which would be very rare, but would crash the actor because of a child actor name conflict:

- we successfully wake up a peer: in the handler for `PeerReadyNotifier.PeerReady` we remove it from our `peers` map, but we do so before the child actor has stopped itself
- another request comes in for the same peer: it's not in our map, but the child actor is still alive so we fail to create a new one because the name is already taken

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,142 @@
...
+        val peers1 = peers.flatMap {
+          case (remoteNodeId, peer) => peer.cancel(paymentHash).map(peer1 => remoteNodeId -> peer1)
+        }
+        watching(peers1)
+      case WrappedCurrentBlockHeight(CurrentBlockHeight(currentBlockHeight)) =>
+        val peers1 = peers.flatMap {
+          case (remoteNodeId, peer) => peer.update(currentBlockHeight).map(peer1 => remoteNodeId -> peer1)
+        }
+        watching(peers1)
+      case WrappedPeerReadyResult(PeerReadyNotifier.PeerReady(remoteNodeId, _)) =>
```
t-bast, line 127: You're not handling `PeerReadyNotifier.PeerUnavailable`! We should have a unit test for that, this is a bug.
It is safer to do:

```scala
case WrappedPeerReadyResult(result) => result match {
  ...
}
```

This will ensure that the compiler checks that all cases have been handled.

remyers: Because we create `PeerReadyNotifier` with `timeout_opt=None`, it should never emit  `PeerReadyNotifier.PeerUnavailable`, and if it did we should ignore it because the `AsyncPaymentsTriggerer` watches for block height events that is uses to time out all watched peers. 

Do you think I should add a comment to the current code about why I do not handle `PeerReadyNotifier` ? or should I instead match the `PeerReadyNotifier` result and ignore it with a comment about why?

t-bast: > Do you think I should add a comment to the current code about why I do not handle PeerReadyNotifier ? or should I instead match the PeerReadyNotifier result and ignore it with a comment about why?

Yes, that would be better, because if we add a new `PeerReadyNotifier.Result` in the future, the compiler will let us know that we may need to handle it here, otherwise with the current code we won't know that we may be missing a case.

eclair-core/src/main/scala/fr/acinq/eclair/Setup.scala
```diff
@@ -296,7 +296,8 @@ class Setup(val datadir: File,
       dbEventHandler = system.actorOf(SimpleSupervisor.props(DbEventHandler.props(nodeParams), "db-event-handler", SupervisorStrategy.Resume))
       register = system.actorOf(SimpleSupervisor.props(Register.props(), "register", SupervisorStrategy.Resume))
       paymentHandler = system.actorOf(SimpleSupervisor.props(PaymentHandler.props(nodeParams, register), "payment-handler", SupervisorStrategy.Resume))
-      relayer = system.actorOf(SimpleSupervisor.props(Relayer.props(nodeParams, router, register, paymentHandler, Some(postRestartCleanUpInitialized)), "relayer", SupervisorStrategy.Resume))
+      triggerer = system.spawn(Behaviors.supervise(AsyncPaymentTriggerer()).onFailure(typed.SupervisorStrategy.resume), name = "peer-watcher")
```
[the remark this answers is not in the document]
remyers: Fixed in [9619b6e](https://github.com/ACINQ/eclair/pull/2491/commits/9619b6e3ee7c59cdfd93d3630dcce7a55d54c567)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
remyers: Good point, no reason to give it a specific name; fixed it in [9619b6e](https://github.com/ACINQ/eclair/pull/2491/commits/9619b6e3ee7c59cdfd93d3630dcce7a55d54c567)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
remyers: Oh, I wrongly assumed there could be only one. Fixed and added a test here: [9619b6e](https://github.com/ACINQ/eclair/pull/2491/commits/9619b6e3ee7c59cdfd93d3630dcce7a55d54c567)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
remyers: Fixed and added a comment here: [9619b6e](https://github.com/ACINQ/eclair/pull/2491/commits/9619b6e3ee7c59cdfd93d3630dcce7a55d54c567)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/AsyncPaymentTriggerer.scala
```diff
@@ -0,0 +1,142 @@
...
+        // notify watcher that destination peer is ready to receive async payments; PeerReadyNotifier will stop itself
+        peers.get(remoteNodeId).foreach(_.trigger())
+        watching(peers - remoteNodeId)
+      case NotifierStopped(remoteNodeId) =>
+        peers.get(remoteNodeId) match {
+          case None => Behaviors.same
+          case Some(peer) =>
+            context.log.error(s"PeerReadyNotifier stopped unexpectedly while watching node $remoteNodeId.")
+            peer.timeout()
```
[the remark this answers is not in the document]
remyers: Fixed in [9619b6e](https://github.com/ACINQ/eclair/pull/2491/commits/9619b6e3ee7c59cdfd93d3630dcce7a55d54c567)

eclair-core/src/main/scala/fr/acinq/eclair/payment/relay/NodeRelay.scala
```diff
@@ -214,31 +215,24 @@ class NodeRelay private(nodeParams: NodeParams,
...
-        context.log.warn(s"rejecting async payment at block $blockHeight; was not triggered after waiting ${nodeParams.relayParams.asyncPaymentsParams.holdTimeoutBlocks} blocks")
-        rejectPayment(upstream, Some(TemporaryNodeFailure)) // TODO: replace failure type when async payment spec is finalized
-        stopping()
-      case WrappedCurrentBlockHeight(_) =>
-        Behaviors.same
       case CancelAsyncPayment =>
         context.log.warn(s"payment sender canceled a waiting async payment")
         rejectPayment(upstream, Some(TemporaryNodeFailure)) // TODO: replace failure type when async payment spec is finalized
         stopping()
-      case RelayAsyncPayment =>
```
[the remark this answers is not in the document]
remyers: Good catch - though we'll likely need to add a similar command to `AsyncPaymentTriggerer` to enable triggering via onion messages returned from receivers that aren't peers. Fixed in [9619b6e](https://github.com/ACINQ/eclair/pull/2491/commits/9619b6e3ee7c59cdfd93d3630dcce7a55d54c567)

t-bast (approved):
LGTM! This seems pretty robust and well tested now :+1:

PR #2495, opened 2022-11-18
Fix ability for plugins to exchange custom Lightning Messages with other LN nodes [PeerSwap]

remyers:
These changes are a prerequisite for the PeerSwap plugin.

This PR introduces a **potentially** breaking change to existing plugins by changing how the codec parses unknown messages. 

Custom Lightning Messages were previously assumed to be prepended with two-bytes for the size of the unknown message blob. Now unknown messages are assumed to be sent as just the custom message type followed by a blob of data, with no assumption that the first two bytes give the size of the following blob.

Unknown Lightning Messages sent to a channel will be forwarded to the channel peer if they are listed as a custom message used by a loaded plugin.   

The change to `unknownMessageCodec` caused `channelUpdateCodecWithType` to incorrectly return success with an `UnknownMessage`, instead of failure. This was solved by using just the `LightningMessageCodec` instead of `meteredLightningMessageCodec` which fell back to `UnknownMessage` for invalid messages.

I also unsealed `TransactionWithInputInfo` so that plugins can use `checkSpendable` with custom on-chain transactions.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
@@ -106,7 +106,7 @@ object Transactions {
     case object Remote extends TxOwner
   }
 
-  sealed trait TransactionWithInputInfo {
+  trait TransactionWithInputInfo {
     def input: InputInfo
     def desc: String
     def tx: Transaction
--- a/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
@@ -106,7 +106,7 @@ object Transactions {
     case object Remote extends TxOwner
   }
 
-  trait TransactionWithInputInfo {
+  sealed trait TransactionWithInputInfo {
     def input: InputInfo
     def desc: String
     def tx: Transaction
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/FailureMessage.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/FailureMessage.scala
@@ -20,11 +20,11 @@ import fr.acinq.bitcoin.scalacompat.ByteVector32
 import fr.acinq.eclair.crypto.Mac32
 import fr.acinq.eclair.wire.protocol.CommonCodecs._
 import fr.acinq.eclair.wire.protocol.FailureMessageCodecs.failureMessageCodec
-import fr.acinq.eclair.wire.protocol.LightningMessageCodecs.{channelFlagsCodec, channelUpdateCodec, messageFlagsCodec, lightningMessageCodec}
+import fr.acinq.eclair.wire.protocol.LightningMessageCodecs.{channelFlagsCodec, channelUpdateCodec, messageFlagsCodec, meteredLightningMessageCodec}
 import fr.acinq.eclair.{BlockHeight, CltvExpiry, MilliSatoshi, MilliSatoshiLong, UInt64}
 import scodec.bits.ByteVector
 import scodec.codecs._
-import scodec.{Attempt, Codec}
+import scodec.{Attempt, Codec, Err}
 
 /**
  * see https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md
@@ -94,7 +94,10 @@ object FailureMessageCodecs {
   val NODE = 0x2000
   val UPDATE = 0x1000
 
-  val channelUpdateCodecWithType = lightningMessageCodec.narrow[ChannelUpdate](f => Attempt.successful(f.asInstanceOf[ChannelUpdate]), g => g)
+  val channelUpdateCodecWithType = meteredLightningMessageCodec.narrow[ChannelUpdate]({
+    case f: ChannelUpdate => Attempt.successful(f)
+    case _ => Attempt.failure(Err("not a ChanelUpdate message"))
+  }, g => g)
 
   // NB: for historical reasons some implementations were including/omitting the message type (258 for ChannelUpdate)
   // this codec supports both versions for decoding, and will encode with the message type
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/FailureMessage.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/FailureMessage.scala
@@ -20,7 +20,7 @@ import fr.acinq.bitcoin.scalacompat.ByteVector32
 import fr.acinq.eclair.crypto.Mac32
 import fr.acinq.eclair.wire.protocol.CommonCodecs._
 import fr.acinq.eclair.wire.protocol.FailureMessageCodecs.failureMessageCodec
-import fr.acinq.eclair.wire.protocol.LightningMessageCodecs.{channelFlagsCodec, channelUpdateCodec, messageFlagsCodec, meteredLightningMessageCodec}
+import fr.acinq.eclair.wire.protocol.LightningMessageCodecs.{channelFlagsCodec, channelUpdateCodec, messageFlagsCodec, lightningMessageCodec}
 import fr.acinq.eclair.{BlockHeight, CltvExpiry, MilliSatoshi, MilliSatoshiLong, UInt64}
 import scodec.bits.ByteVector
 import scodec.codecs._
@@ -94,7 +94,7 @@ object FailureMessageCodecs {
   val NODE = 0x2000
   val UPDATE = 0x1000
 
-  val channelUpdateCodecWithType = meteredLightningMessageCodec.narrow[ChannelUpdate](f => Attempt.successful(f.asInstanceOf[ChannelUpdate]), g => g)
+  val channelUpdateCodecWithType = lightningMessageCodec.narrow[ChannelUpdate](f => Attempt.successful(f.asInstanceOf[ChannelUpdate]), g => g)
 
   // NB: for historical reasons some implementations were including/omitting the message type (258 for ChannelUpdate)
   // this codec supports both versions for decoding, and will encode with the message type
--- a/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
@@ -408,7 +408,7 @@ object LightningMessageCodecs {
 
   val unknownMessageCodec: Codec[UnknownMessage] = (
     ("tag" | uint16) ::
-      ("message" | varsizebinarydata)
+      ("message" | bytes)
     ).as[UnknownMessage]
 
   val lightningMessageCodec = discriminated[LightningMessage].by(uint16)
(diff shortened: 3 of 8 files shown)


thomash-acinq:
> Custom Lightning Messages were previously assumed to be prepended with two-bytes for the size of the unknown message blob. Now unknown messages are assumed to be sent as just the custom message type followed by a blob of data, with no assumption that the first two bytes give the size of the following blob.

This is an assumption of TLVs, does it mean that you're using messages that are not TLVs? Why?

remyers:
> This is an assumption of TLVs, does it mean that you're using messages that are not TLVs? Why?

The length is still implicitly encoded, none of the other Lightning Message codecs explicitly encode the length of the payload like `codecUnknownMessage` did before this change.

thomash-acinq:
> The length is still implicitly encoded, none of the other Lightning Message codecs explicitly encode the length of the payload like `codecUnknownMessage` did before this change.

It looked like a TLV at first glance but you're right, it's not, so the tag doesn't have to be followed by a size.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/FailureMessage.scala
```diff
@@ -89,7 +89,7 @@ object FailureMessageCodecs {
   val NODE = 0x2000
   val UPDATE = 0x1000
 
-  val channelUpdateCodecWithType = meteredLightningMessageCodec.narrow[ChannelUpdate](f => Attempt.successful(f.asInstanceOf[ChannelUpdate]), g => g)
+  val channelUpdateCodecWithType = lightningMessageCodec.narrow[ChannelUpdate](f => Attempt.successful(f.asInstanceOf[ChannelUpdate]), g => g)
```
thomash-acinq, line 92: That's not a good fix, the metered version adds metrics and we want to keep that. The problem is that we don't check that `f` is actually a `ChannelUpdate`.

```suggestion
  val channelUpdateCodecWithType = meteredLightningMessageCodec.narrow[ChannelUpdate]({
    case f: ChannelUpdate => Attempt.successful(f)
    case _ => Attempt.failure()
  }, g => g)
```

remyers: That looks like a good solution too. 

Is metering useful here? it appears this codec parses a `ChannelUpdate` messages that is returned as part of a `FailureMessage`, but a newly received `ChannelUpdate` will still go through `meteredLightningMessageCodec`.

thomash-acinq: I don't really know if metering is useful but I'm assuming it was put there for a reason, so we shouldn't remove it without a good reason. @t-bast what metrics are we interested in? Should this codec be metered?
Anyway I think it's worth fixing the `asInstanceOf[ChannelUpdate]`. There might be some edge case where a tagless `ChannelUpdate` happens to parse as a valid lightning message.
You also need to rebase on master to solve some conflicts.

remyers: > You also need to rebase on master to solve some conflicts.

Oh thanks; forgot to do that. I just pushed the rebase.

remyers: > Anyway I think it's worth fixing the `asInstanceOf[ChannelUpdate]`. There might be some edge case where a tagless `ChannelUpdate` happens to parse as a valid lightning message.
> You also need to rebase on master to solve some conflicts.

Your suggested fix is in [364f81c](https://github.com/ACINQ/eclair/pull/2495/commits/364f81ca4bf00e622ccd21973302dd7303c55678). I wanted to do something like that initially, but couldn't figure out the syntax.

thomash-acinq (approved):

remyers:
Unfortunately the issue of whether or not to include the type prefix remains unresolved in BOLT 4 [PR# 545](https://github.com/lightning/bolts/pull/545).

remyers:
Is there a way to find out how many failure messages we get that are missing the type prefix for the embedded `channel_update`? It would be nice to remove entirely the ambiguity in parsing failure messages now that the spec is updated.

tnull:
> Is there a way to find out how many failure messages we get that are missing the type prefix for the embedded `channel_update`? It would be nice to remove entirely the ambiguity in parsing failure messages now that the spec is updated.

I would also be interested in some numbers, as I'd like to be able to remove said parsing ambiguity from LDK/rust-lightning. However, I assume it's not ready yet, as LND still doesn't seem to write the type prefix?: https://github.com/lightningnetwork/lnd/issues/6461

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
```diff
@@ -100,7 +100,7 @@ object Transactions {
     case object Remote extends TxOwner
   }
 
-  sealed trait TransactionWithInputInfo {
+  trait TransactionWithInputInfo {
```
t-bast, line 103: This change is unrelated to the `UnkownMessage`, I think it should be removed from this PR.

I don't think it's useful at all anyway, `checkSpendable` is a one-line function that calls `bitcoin-lib`, you should just replicate it in the `PeerSwap` plugin instead of depending on `TransactionWithInputInfo`.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageCodecs.scala
```diff
@@ -417,7 +417,7 @@ object LightningMessageCodecs {
 
   val unknownMessageCodec: Codec[UnknownMessage] = (
     ("tag" | uint16) ::
-      ("message" | varsizebinarydata)
+      ("message" | bytes)
```
t-bast, line 420: Yep, this is what this codec should have done from the start, there is no good reason to require unknown messages to start with their length (especially a 2-bytes length which is incompatible with a TLV stream). You're not breaking compatibility here, plugins will simply have to separately read the first two bytes if they expect a length prefix :+1:

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
```diff
@@ -1648,6 +1648,11 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val
     case Event(WatchFundingSpentTriggered(tx), d: PersistentChannelData) if tx.txid == d.commitments.localCommit.commitTxAndRemoteSig.commitTx.tx.txid =>
       log.warning(s"processing local commit spent in catch-all handler")
       spendLocalCurrent(d)
+
+    // forward unknown messages that originate from loaded plugins
+    case Event(unknownMsg: UnknownMessage, _) if nodeParams.pluginMessageTags.contains(unknownMsg.tag) =>
```
t-bast, line 1653: This is probably worth a unit test to make sure we don't inadvertently break that forwarding code in a future refactoring.

t-bast (commented):
LGTM, only a couple of nits

eclair-core/src/main/scala/fr/acinq/eclair/transactions/Transactions.scala
remyers: Fixed in [8606def](https://github.com/ACINQ/eclair/commit/8606defdd703896982cd024185fa4ffa9bf6b021).

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
t-bast: I don't think it's worth creating a new test spec file, but I don't think `ChannelDataSpec` is the right place for it either. This is something that happens during channel's lifetime, so I guess `NormalStateSpec` is the most ok place to put it, above the `recv CurrentBlockCount` tests.

But it makes me wonder, do you really need that handler inside the `Channel` actor? This is really something that should directly go to the `Peer` actor, can't the `PeerSwap` plugin obtain a reference to the peer instead of going through the channel? It would be much cleaner.

remyers: > But it makes me wonder, do you really need that handler inside the `Channel` actor? This is really something that should directly go to the `Peer` actor, can't the `PeerSwap` plugin obtain a reference to the peer instead of going through the channel? It would be much cleaner.

One option would be to initialize a swap actor with the remote peer's `NodeId` and a reference to the `switchboard`. Then we could do something like:

```scala 
switchboard ! ForwardUnknownMessage(peerNodeId, unknownMessage)
```

and add to `Switchboard`:
```scala
case class Switchboard.ForwardUnknownMessage(nodeId: PublicKey, msg: UnknownMessage)

    case ForwardUnknownMessage(remoteNodeId, msg) =>
      getPeer(remoteNodeId) match {
        case Some(peer) => peer ! Peer.OutgoingMessage(msg, activeConnection)
      }
```
But only the channel keeps track of the current `activeConnection` so we could perhaps change `OutgoingMessage` in  `Peer` to make the `peerConnection` parameter optional. 

```scala
  case class OutgoingMessage(msg: LightningMessage, peerConnection: Option[ActorRef])
```

 ¯\\_(ツ)_/¯

t-bast: I see, the issue is that we currently don't have any places in the codebase where we send outgoing unknown messages yet, I thought we had that in the `Peer` but we actually only have code to _receive_ unknown messages.

The question is then what guarantees do we want to provide for outgoing unknown messages? The reason the channel keeps track of the active connection is because for channel messages, we want to avoid sending them when a reconnection happened because we have an explicit reconnection protocol to reconcile messages that weren't acked. I don't think we want that feature for unknown messages, do we? We simply always want them to be sent on the active connection, right?

If that's correct, we just need the `Peer` to define a new command `RelayUnknownMessage`, similar to `RelayOnionMessage`. We can then also have the `switchboard` provide a way to send it to a `remoteNodeId` with the `ForwardUnknownMessage` you suggest.

remyers: > I don't think we want that feature for unknown messages, do we? We simply always want them to be sent on the active connection, right?

I agree that there's no assumed re-connection protocol  for sending `UnknownMessages`. I'll make these changes to my Peerswap plugin branch to double check that it is sufficient to send a `RelayUnknownMessage` messages to the`Peer` via the `switchboard` instead of using a `Register.ForwardShortId` message.

remyers: Changed to forward unknown messages via the switchboard in [0532835](https://github.com/ACINQ/eclair/pull/2495/commits/0532835094c1eeb277b54b4657d754d7db3075e2).

I also refactored the `peerswap-plugin` branch to test that this change is sufficient for that use case.

remyers: PR #2465 adds `Register.ForwardNodeId` which does something similar to the the new `Switchboard.ForwardUnknownMessage` proposed by this PR. Using `Register.ForwardNodeId` does not add redundant code to `Switchboard` and includes `replyTo` with a failure message if the peer does not exist.

Once PR #2465 is merged I'll update this PR to use it.

remyers: [Rolled back (510b695)](https://github.com/ACINQ/eclair/pull/2495/commits/510b695f5a6aea2c90255d5bc54d2607a601b25c) addition of `Switchboard.ForwardUnknownMessage`. Instead of using the `switchboard` to forward to unknown messages to a remote peer,  plugins should query the `switchboard` using `GetPeerInfo` to retrieve the actor ref of the remote peer and use the new `Peer.RelayUnknownMessage` event to forward unknown messages.

PR #151, opened 2017-09-05
Fix byte order of chain hash

sstone:
see See https://github.com/lightningnetwork/lightning-rfc/issues/237

PR #2496, opened 2022-11-21
Add plugin for Peerswap atomic swap based local liquidity management protocol

remyers:
This PR implements the draft [PeerSwap protocol ](https://github.com/ElementsProject/peerswap/blob/master/docs/peer-protocol.md) as an Eclair Plugin and builds off of PR #2342 with a cleaner commit history after reorganization into a plug-in. This PR requires PR #2495 which makes some prerequisite changes to Eclair.

The plugin creates a `SwapRegister` actor at startup which will spawn swap sender actors in response to user CLI commands. The `SwapRegister` will also spawn swap receiver actors in response to swap requests received from peers. The `SwapRegister` forwards swap messages received from peers based on their `swapId` field to corresponding local swap actors. When a swap actor has completed, it simply stops and this triggers the `SwapRegister` to remove it from its map.

The Swap database check-points information about in-progress swaps after a swap maker has committed an on-chain tx or a swap taker has initiated an off-chain payment. Details about the result of the swap are added to the swap record when it completes.

Protocol implementation TODOs:
- [ ] add CLI and/or configurable "allow-list" for swap peers and other limits for accepting swap requests
- [ ] channels should fail proposed htlc updates that would remove liquidity needed for an active swap
- [ ] add or adjust timeouts and timeout behavior to ensure safe operation of the protocol
- [ ] re-enable the logic to check for acceptable premium values (see [PR #151](https://github.com/ElementsProject/peerswap/issues/151))

PeerSwap [specification](https://github.com/ElementsProject/peerswap-spec) suggestions/questions to discuss:
 - [ ] custom byte encoded message rather than json [#1](https://github.com/ElementsProject/peerswap-spec/issues/1)
 - [ ] use 32-byte channel ids instead of short channel ids [#2](https://github.com/ElementsProject/peerswap-spec/issues/2)
 - [ ] nit: rename “OpeningTxBroadcasted” to grammatically correct “OpeningTxBroadcast” [#3](https://github.com/ElementsProject/peerswap-spec/issues/3)
 - [ ] send a signature instead of a private key for cooperative closes [#4](https://github.com/ElementsProject/peerswap-spec/issues/4)

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
@@ -1648,6 +1648,11 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder, val 
     case Event(WatchFundingSpentTriggered(tx), d: PersistentChannelData) if tx.txid == d.commitments.localCommit.commitTxAndRemoteSig.commitTx.tx.txid =>
       log.warning(s"processing local commit spent in catch-all handler")
       spendLocalCurrent(d)
+
+    // forward unknown messages that originate from loaded plugins
+    case Event(unknownMsg: UnknownMessage, _) if nodeParams.pluginMessageTags.contains(unknownMsg.tag) =>
+      send(unknownMsg)
+      stay()
   }
 
   onTransition {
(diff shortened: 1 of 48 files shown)


remyers:
The [comment](https://github.com/ACINQ/eclair/pull/2342#discussion_r931985359) from @t-bast about making Maker/Taker actors use `SupervisorStrategy.resume` is addressed by [d272226](https://github.com/ACINQ/eclair/pull/2496/commits/d272226df23e6e7db0c40f4f92838cdac7c735f9). This commit also adds comments to explain the choice.

The change to `resume` has been made safe by checking that no swap has been saved as a checkpoint to the swap db before the `Maker` funds the  opening tx or the `Taker` pays the swap invoice.

Should I keep the logic from [036e7f6](https://github.com/ACINQ/eclair/pull/2496/commits/036e7f6d1e60163af2ae1930ff0776e107be29f7) that I added to handle the case where a `Maker` or `Taker` stops prematurely before recording a result?

t-bast:
I have created the `eclair-plugins` repository, it would make sense to move this work over there: https://github.com/ACINQ/eclair-plugins

remyers:
This PR is continued in the `eclair-plugins` repository as [PR #5](https://github.com/ACINQ/eclair-plugins/pull/5)

PR #2499, opened 2022-11-21
Build Bolt12 invoices with provided intermediary nodes

thomash-acinq:
Allows creating arbitrary blinded routes for BOLT12 invoices.


t-bast (commented):
I don't mind the use of `pipeTo`, it's a bit ugly but it removes the need for boilerplate actor management.

My other comments from #2471 still apply though: it would be useful to refactor most of what is happening inside the handler to dedicated functions that can be called in tests.

Also, test coverage is really low, you must test the cases where one of the `FinalizeRoute` fails, when one of them times out, when edge cases are hit, etc. Please spend more time on this and think about what makes sense to test.

thomash-acinq:
I've put the blinded route creation and the aggregation of fees in separate functions.
For the test where a `FinalizeRoute` times out, I'm not sure it would bring much value and just waste testing time.

t-bast (commented):
The logic looks good and the test coverage is nice, I have made two follow-up commits in #2504
One of them contains a change to provide more flexibility in building the dummy hops (to make them credible) and the other just moves the code around.

t-bast (approved):

PR #2500, opened 2022-11-23
Accept payments to zero-hop blinded route

thomash-acinq:
In case of a zero-hop blinded route, the final blinding point is not in the `UpdateAddHtlc` but in the payload itself.


t-bast (approved):
Good catch, LGTM :+1:

PR #2501, opened 2022-11-23
Implement correct ordering of `tx_signatures`

t-bast:
According to the dual funding specification, the peer that has [contributed the smaller amount](https://github.com/niftynei/lightning-rfc/blob/nifty/dual-fund/02-peer-protocol.md#requirements-4) of `tx_add_input` must sign first. We incorrectly used the contributions to the funding output instead of the sum of the inputs values. This requirement guarantees that there can be no deadlocks when nodes are batching multiple `interactive-tx` sessions.

When building the first version of a dual-funded transaction, we should use `error` instead of `tx_abort` if we encounter a failure. It makes it more explicit to our peer that the channel should be closed at that point, whereas `tx_abort` can be used in RBF attempts to abort a secondary `interactive-tx` sessions without closing the channel. However, our peer may need to wait for the funding transaction to be double-spent before actually forgetting the channel if they have already sent us their `tx_signatures`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
@@ -763,14 +763,12 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
   }
 
   def signFundingTx(completeTx: SharedTransaction, commitments: Commitments): Behavior[Command] = {
-    val shouldSignFirst = if (completeTx.localAmountIn < completeTx.remoteAmountIn) {
-      // The peer with the lowest total of input amount must transmit its `tx_signatures` first.
-      true
-    } else if (completeTx.localAmountIn == completeTx.remoteAmountIn) {
+    val shouldSignFirst = if (completeTx.localAmountIn == completeTx.remoteAmountIn) {
       // When both peers contribute the same amount, the peer with the lowest pubkey must transmit its `tx_signatures` first.
       LexicographicalOrdering.isLessThan(commitments.localParams.nodeId.value, commitments.remoteNodeId.value)
     } else {
-      false
+      // The peer with the lowest total of input amount must transmit its `tx_signatures` first.
+      completeTx.localAmountIn < completeTx.remoteAmountIn
     }
     if (shouldSignFirst) {
       signTx(completeTx, None)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
@@ -763,10 +763,10 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
   }
 
   def signFundingTx(completeTx: SharedTransaction, commitments: Commitments): Behavior[Command] = {
-    val shouldSignFirst = if (fundingParams.localAmount < fundingParams.remoteAmount) {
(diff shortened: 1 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
```diff
@@ -763,10 +763,10 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
...
   def signFundingTx(completeTx: SharedTransaction, commitments: Commitments): Behavior[Command] = {
-    val shouldSignFirst = if (fundingParams.localAmount < fundingParams.remoteAmount) {
+    val shouldSignFirst = if (completeTx.localAmountIn < completeTx.remoteAmountIn) {
       // The peer with the lowest total of input amount must transmit its `tx_signatures` first.
       true
-    } else if (fundingParams.localAmount == fundingParams.remoteAmount) {
+    } else if (completeTx.localAmountIn == completeTx.remoteAmountIn) {
       // When both peers contribute the same amount, the peer with the lowest pubkey must transmit its `tx_signatures` first.
       LexicographicalOrdering.isLessThan(commitments.localParams.nodeId.value, commitments.remoteNodeId.value)
     } else {
```
pm47, line 772: Nit:

```scala
    val shouldSignFirst = if (completeTx.localAmountIn == completeTx.remoteAmountIn) {
      // When both peers contribute the same amount, the peer with the lowest pubkey must transmit its `tx_signatures` first.
      LexicographicalOrdering.isLessThan(commitments.localParams.nodeId.value, commitments.remoteNodeId.value)
    } else {
      // The peer with the lowest total of input amount must transmit its `tx_signatures` first.
      completeTx.localAmountIn < completeTx.remoteAmountIn
    }
```

pm47 (commented):
> We incorrectly used the contributions to the funding output instead of the sum of the inputs values.

That's a subtle distinction! LGTM.

> When building the first version of a dual-funded transaction, we should use error instead of tx_abort if we encounter a failure. It makes it more explicit to our peer that the channel should be closed at that point

I think sending a `tx_abort` was plain wrong, because [according to the spec](https://github.com/niftynei/lightning-rfc/blob/nifty/dual-fund/02-peer-protocol.md#the-tx_abort-message) we should start over the negotiation, which we don't support for the first commitment:
> The tx_abort message allows for the cancellation of an in progress negotiation, and a return to the initial starting state. It is distinct from the error message, which triggers a channel close.

Technically I think the `WAIT_FOR_DUAL_FUNDING_CREATED`->`CLOSED` transition on `TxAbort` is wrong?

t-bast:
> Technically I think the WAIT_FOR_DUAL_FUNDING_CREATED->CLOSED transition on TxAbort is wrong?

Right, in theory we could abort and renegotiate without going to `CLOSED`, I'm curious to see what other implementations do here. We may change this depending on how other implementations work, but it's simpler for now to close (if the peer sends us `interactive-tx` messages after that `tx_abort`, we'll just reply with an `error` at some point because we don't know the channel and we'll restart negotiation "from scratch" with `open_channel2` / `accept_channel2`).

pm47 (dismissed):
> We may change this depending on how other implementations work, but it's simpler for now to close

That's seems very reasonable 👍

eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
t-bast: Thanks, done in https://github.com/ACINQ/eclair/pull/2501/commits/2fa206ca0b9b0185081d98e57d95f085eb03f31a

pm47 (approved):

PR #2503, opened 2022-11-24
Lift channel exclusion only once per exclusion

### LINKED ISSUE #2502: Multiple LiftChannelExclusion events possible in schedule, reduce actuall exclusion time
Sending an ExcludeChannel event to the router, adds the channel to the excluded list and schedules an LiftChannelExclusion according to the exclusion time.

If you send another ExcludeChannel while the channel is already on the exclusion list, this adds another LiftChannelExclusion  wihtout removeing the older ones.

This leads to exclusion duration beeing cut short, because the older LiftChannelExclusion will trigger first and unbann the channel to soon.

This is very easy to reproduce with a custom plugin but also happens a lot without.
Just wait for a channel to be excluded because of channel failure, then send a custom exclusion event 1 minute later to exclude the channel for 30 minutes.
The channel will be unbanned when the LiftChannelExclusion  from the channel failure triggers.

This can also happen if multiple payments are using the same channels in parallel.
If both payments fail at the same channel they will generate 2 LiftChannelExclusion .

I have channels that have literally hundres of LiftChannelExclusion because they fail all the time, but due to this behaviour they never really get exluded for more than a few seconds even if my config says to exclude them for 10 minutes on failure.

Version used: version=0.7.0 commit=a804905

thomash-acinq:
If a channel is already excluded, it means a `LiftChannelExclusion` is already scheduled, so we shouldn't schedule a new one.

Fixes #2502

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -172,7 +172,9 @@ class Router(val nodeParams: NodeParams, watcher: typed.ActorRef[ZmqWatcher.Comm
 
     case Event(ExcludeChannel(desc, duration_opt), d) =>
       log.info("excluding shortChannelId={} from nodeId={} for duration={}", desc.shortChannelId, desc.a, duration_opt.getOrElse("n/a"))
-      duration_opt.foreach(banDuration => context.system.scheduler.scheduleOnce(banDuration, self, LiftChannelExclusion(desc)))
+      if (!d.excludedChannels.contains(desc)) {
+        duration_opt.foreach(banDuration => context.system.scheduler.scheduleOnce(banDuration, self, LiftChannelExclusion(desc)))
+      }
       stay() using d.copy(excludedChannels = d.excludedChannels + desc)
 
     case Event(LiftChannelExclusion(desc@ChannelDesc(shortChannelId, nodeId, _)), d) =>


t-bast:
I don't think that the change you're making is what we actually want.

If I send a first `ExcludeChannel(1 minute)` and then `ExcludeChannel(None)`, I want the channel to be permanently excluded.
With your code change, the exclusion will be lifted after 1 minute. It's also what happens with the code on `master`.

But if I send a first `ExcludeChannel(None)` and then `ExcludeChannel(1 minute)`, I also want the channel to be permanently excluded, right?

Another scenario: I send a first `ExcludeChannel(1 hour)`, and 50 minutes later I send `ExcludeChannel(30 minutes)`. When that happens, I want to extend the exclusion duration.

I think we currently don't store enough information to correctly handle this. Maybe we should change `excludedChannels: Set[ChannelDesc]` to become `excludedChannels: Map[ChannelDesc, Option[TimestampSecond]]` where the second argument is the timestamp at which the exclusion will be automatically lifted. When we receive an `ExcludeChannel`, we can then look that information up and decide whether we should extend the exclusion duration, cancel the timer that will lift the exclusion, or ignore that request. WDYT @thomash-acinq @DerEwige ?

DerEwige:
I agree, that that simple solution does not entirely solve the problem.
But at least it prevents the accumulation of multiple scheduled `LiftChannelExclusion `events.
This already mitigates the problem a lot.

Your proposed solution is much better, but requires a larger rewrite.
Currently you sent the `LiftChannelExclusion `event to the scheduler, when you process the `ExcludeChannel `event. Once the event is scheduled, I’m not sure it could be canceled in the current approach.

So not only would you need store the information, you would also need a process to release the events.
(this is currently done by the akka event scheduler afaik)

t-bast:
> Once the event is scheduled, I’m not sure it could be canceled in the current approach.

`scheduleOnce` returns a `Cancellable`, we simply have to store that in the `Map` as well. Maybe worth introducing a bit more types to make it clean:

```scala
sealed trait ExcludedChannelStatus
case object ExcludedForever extends ExcludedChannelStatus
case class ExcludedUntil(liftExclusionAt: TimestampSecond, timer: Cancellable) extends ExcludedChannelStatus

excludeChannels: Map[ChannelDesc, ExcludedChannelStatus]
```

thomash-acinq:
My first idea to solve this was actually https://github.com/ACINQ/eclair/pull/2506
However I though the other solution was even simpler and that we don't need the added complexity.

> Another scenario: I send a first `ExcludeChannel(1 hour)`, and 50 minutes later I send `ExcludeChannel(30 minutes)`. When that happens, I want to extend the exclusion duration.

I don't see how that would happen, if the channel was excluded, we don't try to use it so we don't have any reason to exclude it again. In @DerEwige's case, it only happens because several payments try the same channel simultaneously and exclude the channel just seconds apart, and in that case I believe that excluding only once is was we want. But if you think the other scenario is important, then let's fix it.

t-bast:
> I don't see how that would happen, if the channel was excluded, we don't try to use it so we don't have any reason to exclude it again.

Not in our default payment flow, but plugins can run their own path-finding algorithm and use `FinalizeRoute`, which doesn't look at excluded channels. When doing that, they may use an excluded channel a while after its exclusion, which can lead to this issue. I don't think it's very hard to robustly fix it, so we should probably do it?

PR #2504, opened 2022-11-24
Improve Bolt 12 invoice building from intermediate nodes

t-bast:
Note that this is a PR on #2499, that contains fixes and refactoring for my comments on that PR.
Let me know if something looks weird or doesn't make sense.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
@@ -35,11 +35,11 @@ import fr.acinq.eclair.payment.Bolt11Invoice.ExtraHop
 import fr.acinq.eclair.payment.Monitoring.{Metrics, Tags}
 import fr.acinq.eclair.payment._
 import fr.acinq.eclair.router.Router
-import fr.acinq.eclair.router.Router.ChannelHop
+import fr.acinq.eclair.router.Router.{ChannelHop, ChannelRelayParams}
 import fr.acinq.eclair.wire.protocol.OfferTypes.{InvoiceRequest, Offer, PaymentInfo}
 import fr.acinq.eclair.wire.protocol.PaymentOnion.FinalPayload
 import fr.acinq.eclair.wire.protocol._
-import fr.acinq.eclair.{CltvExpiryDelta, FeatureSupport, Features, InvoiceFeature, Logs, MilliSatoshi, MilliSatoshiLong, NodeParams, TimestampMilli, randomBytes32, randomKey}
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -251,22 +250,42 @@ object MultiPartHandler {
                                     paymentPreimage_opt: Option[ByteVector32] = None,
                                     paymentType: String = PaymentType.Standard) extends ReceivePayment
 
+  /**
+   * A dummy blinded hop that will be added at the end of a blinded route.
+   * The fees and expiry delta should match those of real channels, otherwise it will be obvious that dummy hops are used.
+   */
+  case class DummyBlindedHop(feeBase: MilliSatoshi, feeProportionalMillionths: Long, cltvExpiryDelta: CltvExpiryDelta)
```
thomash-acinq, line 257: These fees and expiry delta won't be visible, only the aggregated value with the other hops. It seems weird to allow the user to choose where to put the fees when the results will be indistinguishable.
You can have a dummy hop with fees followed by a dummy hop without fees, the same in reverse order, or just two hops with half the fees each and the resulting blinded routes will look exactly the same to a third party.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/OfferTypes.scala
```diff
@@ -69,6 +70,22 @@ object OfferTypes {
     def fee(amount: MilliSatoshi): MilliSatoshi = nodeFee(feeBase, feeProportionalMillionths, amount)
   }
 
+  object PaymentInfo {
+    /** Compute aggregated fees and expiry for a blinded route. */
+    def apply(amount: MilliSatoshi, hops: Seq[Router.ChannelHop]): PaymentInfo = {
```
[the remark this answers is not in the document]
t-bast: I'm not totally convinced this is the perfect place, I agree...but I didn't want to expose the `CreateInvoiceActor` to callers that need to create these things. Maybe we should find a place inside the `router` package for these three utility functions? I'll think about this a bit more.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: > These fees and expiry delta won't be visible, only the aggregated value with the other hops.

But if they're set to 0, the sender can trivially figure out that we're using only dummy hops since the aggregated values will be very low. It feels to me that a smart(ish) way of generating dummy hops is by looking at all credible hops going out and selecting fees/expiry that match most of these real hops, and iterating like that, that's why I made it that way. It also forces the caller to think about the fact that _each_ dummy hop should add real fees/expiry to be useful to avoid reducing the anonymity set.

PR #2505, opened 2022-11-24
Move route blinding construction to RouteCalculation

t-bast:
We'd like to avoid having the `wire` package depend on data types defined in the `router` package, the other way around is better as suggested by @thomash-acinq.


t-bast:
I'll give a try to add a new object to contain those utilities, hold on for a few minutes before reviewing.

t-bast:
This is ready to review, a dedicated object inside the `Router` makes more sense here than what I previously did IMO

PR #2507, opened 2022-11-24
Create routes to self using remote alias

t-bast:
The introduction of `scid-alias` broke the ability to create a valid route to our own node using `FinalizeRoute` for private or unconfirmed channels because we use our local alias as the shortChannelId of the graph edge in both directions.

Using the same identifier for both directions makes sense, because it's a stable identifier whereas the remote alias could be updated by our peer. It's generally not an issue, except when we are building a route, because our peer will not know how to route if we give them our local alias in a payment onion (they expect their own local alias, which from our point of view is the remote alias).

The only way to build routes to ourselves is by using `FinalizeRoute`, so we fix the issue only in this handler by looking at our private channels when we notice that we are the destination.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -35,6 +35,19 @@ import scala.util.{Failure, Random, Success, Try}
 
 object RouteCalculation {
 
+  private def getEdgeRelayScid(d: Data, localNodeId: PublicKey, e: GraphEdge): ShortChannelId = {
+    if (e.desc.b == localNodeId) {
+      // We are the destination of that edge: local graph edges always use either the local alias or the real scid.
+      // We want to use the remote alias when available, because our peer won't understand our local alias.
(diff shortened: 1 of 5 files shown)


thomash-acinq:
Are you sure we need this?
In the failing tests, `e.desc.shortChannelId` was good, the problem was with `e.params.shortChannelId`.

thomash-acinq (approved):

t-bast:
> In the failing tests, e.desc.shortChannelId was good, the problem was with e.params.shortChannelId.

I think that's only because you didn't test the private channels case. For private channels, `e.desc.shortChannelId` would be the local alias (as the newly added unit tests show: they fail without the change to `RouteCalculation`), which our peer doesn't use for relay.

pm47 (approved):
This is really hacky, but should be fixed once we finally make the key of your graph the _logical_ `(node1,node2)`, as opposed to the current _physical_ `(scid, node1, node2)`, because we will then have to handle the scid selection explicitly.

t-bast:
Agreed, that will be a nice refactoring, we should aim for this. It won't be a trivial change though, it impacts many places in the codebase, but it's a necessary one.

PR #2510, opened 2022-11-25
Cancel previous `LiftChannelExclusion`

thomash-acinq:
Keep at most one scheduled `LiftChannelExclusion` by cancelling the previous one if the new ban lasts longer

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/RouteCalculation.scala
@@ -99,7 +99,7 @@ object RouteCalculation {
       implicit val sender: ActorRef = ctx.self // necessary to preserve origin when sending messages to other actors
 
       val extraEdges = r.extraEdges.map(GraphEdge(_)).filterNot(_.desc.a == r.source).toSet // we ignore routing hints for our own channels, we have more accurate information
-      val ignoredEdges = r.ignore.channels ++ d.excludedChannels
+      val ignoredEdges = r.ignore.channels ++ d.excludedChannels.keySet
       val params = r.routeParams
       val routesToFind = if (params.randomize) DEFAULT_ROUTES_COUNT else 1
 
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Router.scala
@@ -18,7 +18,7 @@ package fr.acinq.eclair.router
 
 import akka.Done
 import akka.actor.typed.scaladsl.adapter.actorRefAdapter
-import akka.actor.{Actor, ActorLogging, ActorRef, Props, Terminated, typed}
+import akka.actor.{Actor, ActorLogging, ActorRef, Cancellable, Props, Terminated, typed}
 import akka.event.DiagnosticLoggingAdapter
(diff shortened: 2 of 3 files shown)


thomash-acinq:
I'm not sure what to do with tests. Since all of this involves timer I can only write tests that call `Thread.sleep`. Last time we talked about this we didn't have a satisfying solution.

t-bast (commented):
The code looks good to me :+1: 

> I'm not sure what to do with tests. Since all of this involves timer I can only write tests that call Thread.sleep. Last time we talked about this we didn't have a satisfying solution.

Using `Thread.sleep` is definitely not a satisfying solution, but that's not at all your only option. A very simple way of testing most of this code is to add to the `Router` a `GetExcludedChannels` command that returns `d.excludedChannels`. You can then trivially add unit tests that look at the updated `d.excludedChannels` after different commands. It could also be a useful command for plugins, that lets them retrieve the latest list of excluded channels and when their exclusion ends.

thomash-acinq:
Adding a `GetExcludedChannels` won't help us test what made us make this change in the first place: that when we exclude a channel for 10 minutes it doesn't come back after 10 seconds. For that we need `Thread.sleep` or equivalent.

t-bast:
> Adding a GetExcludedChannels won't help us test what made us make this change in the first place: that when we exclude a channel for 10 minutes it doesn't come back after 10 seconds. For that we need Thread.sleep or equivalent.

I disagree, you can test that the `liftExclusionAt` value changed and the timer object changed as well, which is already quite good. Then the only untested code that you have to review is to verify that the duration passed to that timer matches the updated `liftExclusionAt`. This is much better than not testing anything or having to use `Thread.sleep`.

thomash-acinq:
This is a test that would already pass before this change.

t-bast:
> This is a test that would already pass before this change.

You're being quite dishonest here...in a fully black box test this would be true, but I'm asking you to test that the value of `liftExclusionAt` was correctly updated (which you just couldn't test with the previous code because that data wasn't even stored), so that the only code that is not tested (and requires careful examination) is 1 line where we create the timer. If you think this isn't better testing than what we currently have on `master`, I don't know what to tell you...

thomash-acinq:
I've added your test but I still think that the original problem is that we had ghost timers with `LiftChannelExclusion` and we could still have these ghost timers and all the tests would pass. That's what I meant by "the test would already pass before the changes".  And I don't see a way to test the absence of these ghosts other than waiting for the timers to expire.

t-bast:
That's something that we can potentially test for with access to the timer scheduler, but I think it's painful to do without akka typed (but maybe I'm wrong and we can test it right now if you want to explore that). I don't know if your test fixtures are properly designed to handle this today though, it may require a bigger refactoring before we can test that properly, I don't know if that's worth the effort right now.

t-bast (approved):

pm47:
For reference this is what @t-bast is referring to: https://doc.akka.io/docs/akka/current/typed/testing-async.html#controlling-the-scheduler. It allows things like:
```scala
val manualTime: ManualTime = ManualTime()
manualTime.timePasses(2.millis)
```
It's indeed only available on _akka typed_, altough we can probably instantiate a typed actor system and spawn our _classic_ `Router` on it. But the test as of 37b5c7a7d4a23bd145f99d213bc16a646671b682 seems good enough imo.

PR #2360, opened 2022-08-01
Broadcast commit tx when nothing at stake

t-bast:
When we have nothing at stake (channel was never used and we don't have funds to claim), we previously directly went to the CLOSED state without publishing our commitment. This can be an issue for our peer if they have lost data or had a hard time getting a funding tx confirmed.

We now publish our commitment once to help them get their funds back in all cases and avoid the CSV delays when getting their funds back.

Fixes #1730

PR #2512, opened 2022-11-29
Fix closing channel state when nothing at stake

t-bast:
This is a follow up to #2360 which was actually buggy: the channel actor doesn't stop itself immediately after going into the CLOSED state, so it received a `WatchFundingTxSpent` event that was handled in the global `whenUnhandled` block, logged a warning and went back to the CLOSING state.

We now go through the normal steps to close those channels, by waiting for the commit tx to be confirmed before actually going to CLOSED.


pm47 (approved):
I like it better that way, much simpler to reason about.

PR #2513, opened 2022-11-30
Test change address type generation

t-bast:
Starting with bitcoind 23.0, a new `changetype` parameter was introduced. If not specified, bitcoind will generate a change output with a type that matches the main output to make it harder for chain analysis to detect which output is the change.

The issue is that lightning requires segwit utxos: if an on-chain payment is sent to a non-segwit address, we still want our change output to use segwit, otherwise we won't be able to use it. We thus must set `addresstype` and `changetype` in `bitcoin.conf` to ensure we never generate legacy change addresses.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
@@ -21,7 +21,7 @@ import akka.pattern.pipe
 import akka.testkit.TestProbe
 import fr.acinq.bitcoin
 import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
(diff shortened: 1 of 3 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
```diff
@@ -961,4 +961,21 @@ class BitcoinCoreClientSpec extends TestKitBaseClass with BitcoindService with A
     assert(Btc(receivedAmount.values).toMilliBtc == amount)
   }
 
+  test("generate segwit change outputs") {
```
t-bast, line 964: Note that this test suite sets the `addresstype` and `changetype` to `bech32m` (see the fixture setup)

pm47 (approved):
LGTM

Xekyo:
Hi @t-bast, I think 24.0 will stop generating a P2PKH descriptor on new wallets, or you could create a wallet that only has descriptors for the types you want. If the P2PKH descriptor is not present, it will just create change of your default type.

t-bast:
Thanks for the comment! Migrating to descriptors in on our roadmap, that will indeed be a much cleaner way of handling this issue :wink:

PR #2514, opened 2022-12-01
Remove warning for pruned local channel update

t-bast:
When a channel is pruned, it is removed from our graph and thus isn't found by `d.resolve()`, but it shouldn't create a warning in the logs, this isn't an unknown channel.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
@@ -533,8 +533,11 @@ object Validation {
         }
       case None =>
         lcu.shortIds.real.toOption match {
-          case Some(realScid) if d.prunedChannels.contains(realScid) => log.debug("created local channel update for pruned channelId={} scid={}", lcu.channelId, realScid)
-          case _ => log.warning("unrecognized local channel update for channelId={} localAlias={}", lcu.channelId, lcu.shortIds.localAlias)
+          case Some(realScid) if d.prunedChannels.contains(realScid) =>
+            log.debug("this is a known pruned local channel, processing channel_update for channelId={} scid={}", lcu.channelId, realScid)
+          case _ =>
+            // should never happen
+            log.warning("unrecognized local channel update for channelId={} localAlias={}", lcu.channelId, lcu.shortIds.localAlias)
         }
         // handle the update: it will be rejected if there is no related channel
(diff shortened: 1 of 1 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -532,8 +532,11 @@ object Validation {
             handleChannelUpdate(d1, db, nodeParams.currentBlockHeight, Left(lcu))
         }
       case None =>
-        // should never happen, we log a warning and handle the update, it will be rejected since there is no related channel
-        log.warning("unrecognized local channel update for channelId={} localAlias={}", lcu.channelId, lcu.shortIds.localAlias)
+        lcu.shortIds.real.toOption match {
+          case Some(realScid) if d.prunedChannels.contains(realScid) => log.debug("created local channel update for pruned channelId={} scid={}", lcu.channelId, realScid)
```
pm47, line 536: For consistency
```suggestion
          case Some(realScid) if d.prunedChannels.contains(realScid) => log.debug("this is a known pruned local channel, processing channel_update for channelId={} scid={}", lcu.channelId, realScid)
```

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
```diff
@@ -532,8 +532,11 @@ object Validation {
...
+          case _ => log.warning("unrecognized local channel update for channelId={} localAlias={}", lcu.channelId, lcu.shortIds.localAlias)
```
pm47, line 537: I use the "should never happen" when an occurence is definitely a bug, but that's a personal convention.

t-bast: Sounds good, done in https://github.com/ACINQ/eclair/pull/2514/commits/c6e0cec473a6847d7891824794f5deecdc9aca26

eclair-core/src/main/scala/fr/acinq/eclair/router/Validation.scala
t-bast: Updated in https://github.com/ACINQ/eclair/pull/2514/commits/c6e0cec473a6847d7891824794f5deecdc9aca26

pm47 (approved):

PR #2516, opened 2022-12-01
Fix balance histogram

thomash-acinq:
The Kamon histogram was not created properly.


t-bast (approved):
Damn, sorry I missed that during the review :man_facepalming: 
We should probably deploy that fix on our node whenever possible to be able to start filling that data?

PR #2518, opened 2022-12-04
Switch to io.zonky.test.db.postgres.embedded.EmbeddedPostgres

rorp:
This PR improves compatibility of the Postgres tests across different OS's without Docker.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
@@ -1,7 +1,6 @@
 package fr.acinq.eclair
 
 import akka.actor.ActorSystem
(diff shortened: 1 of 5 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/TestDatabases.scala
```diff
@@ -128,6 +127,8 @@ object TestDatabases {
   }
 
   object TestPgDatabases {
+    import _root_.io.zonky.test.db.postgres.embedded.EmbeddedPostgres
```
t-bast, line 130: Why is that `_root_` necessary?

t-bast (approved):
LGTM, I checked the diff between this repo and the previous one and nothing looks suspicious.

PR #2493, opened 2022-11-17
Fix mempool flaky tests

pm47:
As usual, there was a race related to `EventStream` subscriptions.

There are still remaining issues, including:

- I can't figure out how this can possibly happen cc @t-bast 
```
- transaction confirmed *** FAILED ***
  java.lang.AssertionError: assertion failed: timeout (15 seconds) during expectMsg while waiting for TxInMempool(272f5128df6c1e9c4c233fc29906468c2b0616b0a43f5ded5eb8cc9ce2d5b218,151,true)
  at scala.Predef$.assert(Predef.scala:279)
  at akka.testkit.TestKitBase.expectMsg_internal(TestKit.scala:461)
  at akka.testkit.TestKitBase.expectMsg(TestKit.scala:438)
  at akka.testkit.TestKitBase.expectMsg$(TestKit.scala:438)
  at akka.testkit.TestKit.expectMsg(TestKit.scala:973)
  at fr.acinq.eclair.channel.publish.MempoolTxMonitorSpec.$anonfun$new$2(MempoolTxMonitorSpec.scala:99)
  at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
  at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
  at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
  at org.scalatest.Transformer.apply(Transformer.scala:22)
```
- not investigated
```
- watch for spent transactions *** FAILED ***
[...]

PR #2520, opened 2022-12-08
Fix flaky tests

t-bast:
This PR supersedes #2493: it cherry-picks the first two commits and fixes other flaky tests.
Best reviewed commit-by-commit.


t-bast:
> Here is an alternative proposal for https://github.com/ACINQ/eclair/commit/634c5dcb079b5c97c06892468515f16a1285077b: [autowatcher.patch](https://github.com/ACINQ/eclair/files/10246060/autowatcher.patch). We can simply define an secondary ZmqWatcher implementation.

Thanks, it's better! Applied in https://github.com/ACINQ/eclair/pull/2520/commits/5936b56ccb8fdad09b18f80ba7064b4115c88cf9 and rebased.

pm47 (approved):

Issue #2426, opened 2022-09-20
Adding a local channel to ignoreShortChannelIds sometimes leads to illegalArgumentException “unknown channel:…..”


DerEwige:
Eclair only checks public announced channels when looking for the ShortChannelID

If you try to ignore a local channel that is either not (yet) announced or private you get the error that this channel is not known, even though it is a local channel.

t-bast:
Fixed by #2523

PR #2523, opened 2022-12-09
Improve `findRoute` ignored channels behavior

### LINKED ISSUE #2346: Unkown ChannelIDs passed to FindRoute throws exception
Invalide ShortChannelIDs (format, etc.) should totally throw exeptions on this option, but not Unknown Channels imho.

Tested version:
Eclair v0.7.0

Description:
The API functions [FindRoute, FindRouteToNode, FindRouteBetweenNodes] have an option ignoreShortChannelIds. If you pass a no longer existing ShortChannelId to this option it will give an error.

    2022-07-16 09:27:48,760 ERROR f.a.eclair.Boot - API call failed with cause=unknown channel: 745034x723x1

In my opinion this should not throw an error. You try to find a route ignoring a channel, if eclair does not know this channel, this is fine as it is ignored implicitly and does not need to be ignored explicitly.
I would much rather prefer a warning instead for this 3 functions on unknown ShortChannelIds passed.

t-bast:
We use router data to resolve both private and public channels, even when scid-alias is used.
When a channel cannot be found, we simply skip it since adding it to the ignore list wouldn't have any impact.

@rorp @DerEwige can you verify that this is what you expect?

Supersedes #2483

Fixes #2346 and #2426

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -53,7 +53,6 @@ import scodec.{Attempt, DecodeResult, codecs}
 
 import java.nio.charset.StandardCharsets
 import java.util.UUID
-import scala.collection.immutable.SortedMap
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future, Promise}
 import scala.util.{Failure, Success}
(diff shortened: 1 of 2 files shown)


DerEwige:
looking good to me

rorp:
👍

thomash-acinq (approved):

PR #2527, opened 2022-12-12
Remove redundant log messages

t-bast:
Reduce the amount of logs generated by default by moving logs that are redundant or only used for targeted debugging to `debug` level or removing them entirely.

I focused on logs that generate some volume and that I never personnally use for day-to-day log browsing, but let me know if some of these logs are important for you and it would be annoying to have to turn on `debug` level for their respective component.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -698,7 +698,7 @@ object Helpers {
         val closingTx = Transactions.makeClosingTx(commitInput, actualLocalScript, actualRemoteScript, localParams.isInitiator, dustLimit, closingFees.preferred, localCommit.spec)
         val localClosingSig = keyManager.sign(closingTx, keyManager.fundingPublicKey(commitments.localParams.fundingKeyPath), TxOwner.Local, commitmentFormat)
         val closingSigned = ClosingSigned(channelId, closingFees.preferred, localClosingSig, TlvStream(ClosingSignedTlv.FeeRange(closingFees.min, closingFees.max)))
-        log.info(s"signed closing txid=${closingTx.tx.txid} with closing fee=${closingSigned.feeSatoshis}")
+        log.debug(s"signed closing txid=${closingTx.tx.txid} with closing fee=${closingSigned.feeSatoshis}")
         log.debug(s"closingTxid=${closingTx.tx.txid} closingTx=${closingTx.tx}}")
         (closingTx, closingSigned)
       }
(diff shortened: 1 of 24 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -698,7 +698,7 @@ object Helpers {
         val closingTx = Transactions.makeClosingTx(commitInput, actualLocalScript, actualRemoteScript, localParams.isInitiator, dustLimit, closingFees.preferred, localCommit.spec)
         val localClosingSig = keyManager.sign(closingTx, keyManager.fundingPublicKey(commitments.localParams.fundingKeyPath), TxOwner.Local, commitmentFormat)
         val closingSigned = ClosingSigned(channelId, closingFees.preferred, localClosingSig, TlvStream(ClosingSignedTlv.FeeRange(closingFees.min, closingFees.max)))
-        log.info(s"signed closing txid=${closingTx.tx.txid} with closing fee=${closingSigned.feeSatoshis}")
+        log.debug(s"signed closing txid=${closingTx.tx.txid} with closing fee=${closingSigned.feeSatoshis}")
```
pm47, line 701: It was nice to have the fee, although this does feel like debug-level info. Well...

pm47 (commented):
It does feel a bit like we only use `warning` or `debug` now 😅. Maybe we should just raise our default log level to `WARN`?

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: But we actually already have it logged at two places in `info` already:

- in the `closing_signed` message that's logged entirely
- in the handler for `closing_signed` in `Channel.scala` in the branches where we finalize the closing tx

That's why I moved it to `debug`, it's really redundant with other logs we're keeping.

t-bast:
> It does feel a bit like we only use warning or debug now sweat_smile. Maybe we should just raise our default log level to WARN?

There are still a lot of useful logs at the `info` level in the codebase! I don't think we should raise our log level to `WARN`, we'd be missing on a lot of useful logs to track that everything is right in several components.

The only exception may be the payment FSMs: if we're confident that we don't need to troubleshoot payments, then we can configure them to `WARN`, because they generate a lot of logs when relaying trampoline payments. I would keep them for now though, as it would be frustrating to have to debug payments and be missing the data, but if log volume really is an issue, maybe we can just activate `INFO` log when we want to troubleshoot, and use `WARN` the rest of the time?

pm47 (approved):

PR #2528, opened 2022-12-13
Allow non segwit outputs in dual funding

t-bast:
This PR contains two independent commits to improve dual funding support:

- removing the hard-coded minimum witness weight which will always be invalid on exotic inputs
- allowing non-segwit outputs


pm47 (approved):
Will the first commit clash with `lightning-kmp` implementation?

t-bast:
This is ready in lightning-kmp as well :sunglasses: 
https://github.com/ACINQ/lightning-kmp/pull/417

pm47:
⚡

PR #2529, opened 2022-12-13
Add mempool eviction tests

t-bast:
When wallet transactions are evicted from the mempool (because their fee is too low), we have a strong dependency on bitcoind's behavior: it must not double-spend the inputs of the evicted transaction, otherwise it could break 0-conf channels.

We verify that behavior in a unit test where we set the mempool to 5MB (the minimum authorized mempool size) and simulate this eviction scenario.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
@@ -21,7 +21,7 @@ import akka.pattern.pipe
 import akka.testkit.TestProbe
 import fr.acinq.bitcoin
 import fr.acinq.bitcoin.scalacompat.Crypto.PublicKey
-import fr.acinq.bitcoin.scalacompat.{Block, Btc, BtcDouble, ByteVector32, MilliBtcDouble, OutPoint, Satoshi, SatoshiLong, Script, ScriptWitness, Transaction, TxIn, TxOut, computeP2PkhAddress, computeP2WpkhAddress}
+import fr.acinq.bitcoin.scalacompat.{Block, Btc, BtcDouble, ByteVector32, MilliBtcDouble, OP_DROP, OP_PUSHDATA, OutPoint, Satoshi, SatoshiLong, Script, ScriptWitness, Transaction, TxIn, TxOut, computeP2PkhAddress, computeP2WpkhAddress}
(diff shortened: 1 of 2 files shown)


eclair-core/src/test/scala/fr/acinq/eclair/blockchain/bitcoind/BitcoinCoreClientSpec.scala
```diff
@@ -978,4 +978,59 @@ class BitcoinCoreClientSpec extends TestKitBaseClass with BitcoindService with A
...
+    }
+
+    // We create a large unconfirmed transaction with a low feerate.
+    val tx = publishLargeTx(parentTxs.head, 450_000 sat, wallet)
+    // Transactions with higher feerates are added to the mempool and evict the first transaction.
+    parentTxs.tail.foreach(parentTx => publishLargeTx(parentTx, 300_000 sat, miner))
+    // Even though the wallet transaction has been evicted, bitcoind doesn't double-spend its inputs.
+    wallet.getMempoolTx(tx.txid).pipeTo(sender.ref)
+    assert(sender.expectMsgType[Failure].cause.getMessage.contains("Transaction not in mempool"))
+    wallet.listUnspent().pipeTo(sender.ref)
```
sstone, line 1029: nit: you could use the low-level `gettransaction`  RPC call to check that tx is till in the wallet ?

t-bast: Done in https://github.com/ACINQ/eclair/pull/2529/commits/89dc986b66b7629bf542b45cf1a98441108e755f

t-bast: Actually this is using `getrawtransaction` and seems to be flaky, I'll try `gettransaction` and see if it's better.

sstone: Yes I think `gettransaction` is not exposed in our RPC client yet, you either need to add this method or call `invoke` directly and parse the returned JSON object

t-bast: Done in https://github.com/ACINQ/eclair/pull/2529/commits/373171a2131f224d8658079f671ab49dfb286092

sstone (approved):

PR #2535, opened 2022-12-19
Improve blinded payments e2e tests

t-bast:


eclair-core/src/test/resources/logback-test.xml
```diff
@@ -53,7 +53,7 @@
         <!--<appender-ref ref="CONSOLE_NOTIFICATIONS"/>-->
     </logger>
 
-    <logger name="fr.acinq.eclair.router" level="WARN"/>
+    <!--<logger name="fr.acinq.eclair.router" level="WARN"/>-->
```
pm47, line 56: Changes to this file should be reverted.

pm47 (commented):
I just realize now that you are using the new basic integration tests fgor blinded payments 🤘

eclair-core/src/test/resources/logback-test.xml
t-bast: Why? That change is actually useful to debug failing tests that use the `MyCapturingAppender`. Does it have a negative impact for your workflow? I find myself constantly commenting this line, so it feels like a better default.

pm47: No negative impact, it just looked like a left over. I'm fine with more granularity for router logs, but then I'd rather remove this line entirely.
```suggestion
```

The other modification is a no-op.

t-bast: Removed in https://github.com/ACINQ/eclair/pull/2535/commits/7ed656cb785ce139ed9ba3e4f3cd227b6af9fc3c

The other change is useful: the previous way of commenting the line was incompatible with IntelliJ's shortcuts for commenting/uncommenting lines, which I use very often when running tests with logs, so it was cumbersome to manually remove characters to uncomment the line. Doing `ctrl + /` is more convenient :)

pm47 (approved):

PR #2536, opened 2022-12-19
Dual funding latest changes

t-bast:
Implement the latest changes in the dual-funding specification:

- final tlv type value for `require_confirmed_inputs` tlv
- add `second_per_commitment_point` to `open_channel2` and `accept_channel2`
- echo remote `tx_abort` (and wait for remote to echo our `tx_abort`)

There is general consensus on the spec PR for these changes, so I think we should include them now.


eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
```diff
@@ -201,6 +201,7 @@ object InteractiveTxBuilder {
             params: Params,
             commitTxFeerate: FeeratePerKw,
             remoteFirstPerCommitmentPoint: PublicKey,
+            remoteSecondPerCommitmentPoint: PublicKey,
```
pm47, line 204: Providing a `remoteNextCommitInfo` would save a `get` by the caller, but it's less pure because we make the implicit assumption that there are no pending remote commit. I'll have the same kind of stuff on the splices PR.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -435,43 +439,50 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
+                d.commitments.remoteCommit.remotePerCommitmentPoint,
+                d.commitments.remoteNextCommitInfo.toOption.get,
+                wallet))
+              txBuilder ! InteractiveTxBuilder.Start(self, d.allFundingTxs)
+              stay() using d.copy(rbfStatus = RbfStatus.RbfInProgress(txBuilder)) sending TxAckRbf(d.channelId, fundingParams.localAmount)
+            }
+          case RbfStatus.RbfAborted =>
+            log.info("rejecting rbf attempt: our previous tx_abort was not acked")
+            stay() sending Warning(d.channelId, InvalidRbfTxAbortNotAcked(d.channelId).getMessage)
+          case _ =>
```
pm47, line 483:
```suggestion
          case _: RbfStatus.RbfRequested | _: RbfStatus.RbfInProgress =>
```

t-bast: Done in https://github.com/ACINQ/eclair/pull/2536/commits/bbd9610daba263d41fc16b1f4734c56506e228e5

pm47 (approved):

PR #2539, opened 2022-12-23
Move `Commitments` methods from companion to class

pm47:
This is just refactoring, there is no feature change. The goal is to later be able to add an upper layer that manages several commitments and acts as a "meta commitments".

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -550,37 +550,37 @@ case class Commitments(channelId: ByteVector32,
   }
 
   def sendFee(cmd: CMD_UPDATE_FEE, feeConf: OnChainFeeConf): Either[ChannelException, (Commitments, UpdateFee)] = {
-    if (!this.localParams.isInitiator) {
-      Left(NonInitiatorCannotSendUpdateFee(this.channelId))
+    if (!localParams.isInitiator) {
+      Left(NonInitiatorCannotSendUpdateFee(channelId))
     } else {
       // let's compute the current commitment *as seen by them* with this change taken into account
-      val fee = UpdateFee(this.channelId, cmd.feeratePerKw)
+      val fee = UpdateFee(channelId, cmd.feeratePerKw)
       // update_fee replace each other, so we can remove previous ones
-      val commitments1 = this.copy(localChanges = this.localChanges.copy(proposed = this.localChanges.proposed.filterNot(_.isInstanceOf[UpdateFee]) :+ fee))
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -477,145 +460,145 @@ object Commitments {
...
+        case None => Left(UnknownHtlcId(channelId, fail.id))
       }
     }
   }
 
-  def sendFee(commitments: Commitments, cmd: CMD_UPDATE_FEE, feeConf: OnChainFeeConf): Either[ChannelException, (Commitments, UpdateFee)] = {
-    if (!commitments.localParams.isInitiator) {
-      Left(NonInitiatorCannotSendUpdateFee(commitments.channelId))
+  def sendFee(cmd: CMD_UPDATE_FEE, feeConf: OnChainFeeConf): Either[ChannelException, (Commitments, UpdateFee)] = {
+    if (!this.localParams.isInitiator) {
```
t-bast, line 553: nit: why did you start using `this.` in this function and below, and above you directly accessed fields like `localParams` without the `this` prefix? I guess you simply replaced `commitments.` with `this.`?

t-bast (commented):
LGTM, just wondering whether we should keep `this.` or directly access fields

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
pm47: I flip-flopped between the two and intended to leave a comment asking for reviewer's opinion. `this` is a bit more fool-proof, but it's verbose and the compiler should warn us in case we accidentally shadow a variable's name.

pm47:
Rebased.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
t-bast: I don't have a strong preference either, but it would be nice to have consistency across the file.
It's a bit easier to make sure things are consistent when *not* using `this.`: we simply have to check its absence.
It's harder to verify that `this.` is used for every field, the compiler doesn't tell us and we have no automated way of checking it.
So I weakly favor not using `this.`, but it's not clear-cut...

pm47: Done in [9e2ee70](https://github.com/ACINQ/eclair/pull/2539/commits/9e2ee70d5c326a8c8fab929daeea6dcfcf397757).

t-bast (approved):

PR #2553, opened 2023-01-04
Refactor wallet input locking

t-bast:
Provide explicit methods for locking and unlocking inputs.
Rename the previous "rollback" method.
Rename related functions.


t-bast:
This is most likely not a good idea, as highlighted [here](https://github.com/ACINQ/eclair/pull/2554#issuecomment-1372183370).

PR #2554, opened 2023-01-04
Avoid locking inputs during interactive-tx

t-bast:
Unless it's necessary (e.g. when using 0-conf), we avoid locking inputs during interactive-tx sessions as it can be abused by malicious peers to lock some of our liquidity (griefing attack).

This PR is best reviewed commit-by-commit.

Builds on top of #2553


t-bast:
While seemingly simple, this mechanism creates a subtle but potentially deadly issue. We must *always* lock inputs selected for funding, otherwise locking wouldn't work at all, as the following scenario highlights:

- we fund a transaction for which we don't lock utxos
- we fund another unrelated transaction for which we lock utxos
- the second transaction ends up using the same utxos as the first one
- but the first transaction confirms, invalidating the second one

This would break the assumptions of the second transaction: its inputs are locked,  so it doesn't expect to potentially be double-spent.

It's particularly dangerous if the node supports zero-conf, as this could happen to a zero-conf transaction and completely break the channel.

PR #2556, opened 2023-01-06
Factor funding spent handlers

pm47:
⚠️  While the change is minor, this is a very sensitive part of the code. Commits should be reviewed separately.

We first factor all handlers for all states in commit 0905619f8eba4ae787c8d59dac01551ce84e4383. Critically, note that tests are left unchanged.

The second commit changes tests, but it is about a corner case.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -1049,11 +1049,12 @@ object Helpers {
           // txNumber must be lesser than 48 bits long
           None
         } else {
-          log.warning(s"a revoked commit has been published with txnumber=$txNumber")
           // now we know what commit number this tx is referring to, we can derive the commitment point from the shachain
           remotePerCommitmentSecrets.getHash(0xFFFFFFFFFFFFL - txNumber)
             .map(d => PrivateKey(d))
             .map(remotePerCommitmentSecret => {
+              log.warning(s"a revoked commit has been published with txnumber=$txNumber")
+
               val remotePerCommitmentPoint = remotePerCommitmentSecret.publicKey
               val remoteDelayedPaymentPubkey = Generators.derivePubKey(remoteParams.delayedPaymentBasepoint, remotePerCommitmentPoint)
               val remoteRevocationPubkey = Generators.revocationPubKey(keyManager.revocationPoint(channelKeyPath).publicKey, remotePerCommitmentPoint)
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
@@ -1040,8 +1040,10 @@ object Helpers {
        */
       def claimCommitTxOutputs(keyManager: ChannelKeyManager, commitments: Commitments, commitTx: Transaction, db: ChannelsDb, feeEstimator: FeeEstimator, feeTargets: FeeTargets)(implicit log: LoggingAdapter): Option[RevokedCommitPublished] = {
         import commitments._
+        // a valid tx will always have at least one input, but this ensures we don't throw in tests
+        val sequence = commitTx.txIn.headOption.map(_.sequence).getOrElse(0L)
+        val obscuredTxNumber = Transactions.decodeTxNumber(sequence, commitTx.lockTime)
         val channelKeyPath = keyManager.keyPath(localParams, channelConfig)
-        val obscuredTxNumber = Transactions.decodeTxNumber(commitTx.txIn.head.sequence, commitTx.lockTime)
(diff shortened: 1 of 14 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
```diff
@@ -287,7 +288,7 @@ trait ErrorHandlers extends CommonHandlers {
 
   def handleRemoteSpentOther(tx: Transaction, d: PersistentChannelData) = {
     log.warning(s"funding tx spent in txid=${tx.txid}")
-    Closing.RevokedClose.claimCommitTxOutputs(keyManager, d.commitments, tx, nodeParams.db.channels, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets) match {
+    Try(Closing.RevokedClose.claimCommitTxOutputs(keyManager, d.commitments, tx, nodeParams.db.channels, nodeParams.onChainFeeConf.feeEstimator, nodeParams.onChainFeeConf.feeTargets)).toOption.flatten match {
```
t-bast, line 291: This is a shortcut, I'd rather slightly rework `Closing.RevokedClose.claimCommitTxOutputs` to make sure it never throws: I believe we only need to remove the two `require` and replace them by returning `None` if their conditions aren't verified. That would be cleaner, wouldn't it?

t-bast (commented):
Nice refactoring, it looks safe to me!

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
pm47: > That would be cleaner, wouldn't it?

Taken in isolation, yes, but then it would be inconsistent with other `*Closed.claim*` methods which all assume that they are called for the correct tx and may throw.

I think it could make sense to have all those methods return a `Try`, but then it's a bigger change.

t-bast: What about first checking whether the tx is a revoked commit before calling `Closing.RevokedClose.claimCommitTxOutputs`? That would be easier to read and it would be more obvious that the two "other" cases are either a revoked commitment or something completely unknown.

pm47: I just realize now that `RevokedClose.claimCommitTxOutputs` returns an `Option`, so my earlier point about creating an inconsistency is moot . I reworked it as per your initial suggestion in 238f51f67.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -1040,88 +1040,91 @@ object Helpers {
...
-                case _: AnchorOutputsCommitmentFormat => withTxGenerationLog("remote-main-delayed") {
-                  Transactions.makeClaimRemoteDelayedOutputTx(commitTx, localParams.dustLimit, localPaymentPoint, localParams.defaultFinalScriptPubKey, feeratePerKwMain).map(claimMain => {
-                    val sig = keyManager.sign(claimMain, keyManager.paymentPoint(channelKeyPath), TxOwner.Local, commitmentFormat)
-                    Transactions.addSigs(claimMain, sig)
-                  })
+        if (txNumber > 0xffffffffffffL) {
+          // txNumber must be lesser than 48 bits long
+          None
+        } else {
+          log.warning(s"a revoked commit has been published with txnumber=$txNumber")
```
t-bast, line 1052: nit: this log line should go inside the last `map`, we only know that it's a revoked commit once we've been able to extract the per-commitment secret

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
t-bast: https://github.com/ACINQ/eclair/commit/238f51f671daed0ce7d353d341efd063a06cb0d5 looks good to me, we can now remove that added `Try`!

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
pm47: Done in 2cb3262db.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ErrorHandlers.scala
pm47: 🤦‍♂️ 

Done in 2cb3262db.

t-bast (approved):

t-bast (approved):

PR #2557, opened 2023-01-06
Factor funding tx acceptance

pm47:
The handling is mostly the same in connected/disconnected states. This does some refactoring to make them as similar as possible.

NB: we don't make a difference between previous/latest commitment being confirmed.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
@@ -570,19 +570,20 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
       val d1 = DATA_WAIT_FOR_DUAL_FUNDING_READY(d.commitments, shortIds, channelReady)
       goto(WAIT_FOR_DUAL_FUNDING_READY) using d1 storing() sending channelReady
 
-    case Event(WatchFundingConfirmedTriggered(blockHeight, txIndex, confirmedTx), d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
+    case Event(w: WatchFundingConfirmedTriggered, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
       // We find which funding transaction got confirmed.
       val allFundingTxs = DualFundingTx(d.fundingTx, d.commitments) +: d.previousFundingTxs
-      allFundingTxs.find(_.commitments.fundingTxId == confirmedTx.txid) match {
+      allFundingTxs.find(_.commitments.fundingTxId == w.tx.txid) match {
         case Some(DualFundingTx(_, commitments)) =>
-          log.info("channelId={} was confirmed at blockHeight={} txIndex={} with funding txid={}", d.channelId, blockHeight, txIndex, confirmedTx.txid)
+          log.info("channelId={} was confirmed at blockHeight={} txIndex={} with funding txid={}", d.channelId, w.blockHeight, w.txIndex, w.tx.txid)
           watchFundingTx(commitments)
-          context.system.eventStream.publish(TransactionConfirmed(commitments.channelId, remoteNodeId, confirmedTx))
-          val realScidStatus = RealScidStatus.Temporary(RealShortChannelId(blockHeight, txIndex, commitments.commitInput.outPoint.index.toInt))
+          context.system.eventStream.publish(TransactionConfirmed(commitments.channelId, remoteNodeId, w.tx))
+          // We can forget other funding attempts now that one of the funding txs is confirmed.
+          val otherFundingTxs = allFundingTxs.filter(_.commitments.fundingTxId != w.tx.txid).map(_.fundingTx)
+          rollbackDualFundingTxs(otherFundingTxs)
+          val realScidStatus = RealScidStatus.Temporary(RealShortChannelId(w.blockHeight, w.txIndex, commitments.commitInput.outPoint.index.toInt))
           val (shortIds, channelReady) = acceptFundingTx(commitments, realScidStatus = realScidStatus)
           d.deferred.foreach(self ! _)
-          val otherFundingTxs = allFundingTxs.filter(_.commitments.fundingTxId != confirmedTx.txid).map(_.fundingTx)
-          rollbackDualFundingTxs(otherFundingTxs)
           val toSend = d.rbfStatus match {
             case RbfStatus.RbfInProgress(txBuilder) =>
               txBuilder ! InteractiveTxBuilder.Abort
@@ -595,7 +596,7 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
           }
           goto(WAIT_FOR_DUAL_FUNDING_READY) using DATA_WAIT_FOR_DUAL_FUNDING_READY(commitments, shortIds, channelReady) storing() sending toSend
         case None =>
-          log.error(s"internal error: the funding tx that confirmed doesn't match any of our funding txs: ${confirmedTx.bin}")
+          log.error(s"internal error: the funding tx that confirmed doesn't match any of our funding txs: ${w.tx.bin}")
           rollbackDualFundingTxs(allFundingTxs.map(_.fundingTx))
           goto(CLOSED)
       }
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
@@ -29,8 +29,6 @@ import fr.acinq.eclair.transactions.Scripts
 import fr.acinq.eclair.wire.protocol._
 import fr.acinq.eclair.{Features, RealShortChannelId, ToMilliSatoshiConversion, UInt64}
 
-import scala.util.{Failure, Success}
-
 /**
  * Created by t-bast on 19/04/2022.
  */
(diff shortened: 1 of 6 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
```diff
@@ -59,25 +59,30 @@ trait DualFundingHandlers extends CommonFundingHandlers {
...
-      log.info("channelId={} was confirmed at blockHeight={} txIndex={} with a previous funding txid={}", d.channelId, w.blockHeight, w.txIndex, w.tx.txid)
-      val confirmed = d.previousFundingTxs.find(_.commitments.fundingTxId == w.tx.txid).get
-      watchFundingTx(confirmed.commitments)
-      context.system.eventStream.publish(TransactionConfirmed(d.channelId, remoteNodeId, w.tx))
-      // We can forget other funding attempts now that one of the funding txs is confirmed.
-      val otherFundingTxs = d.fundingTx +: d.previousFundingTxs.filter(_.commitments.fundingTxId != w.tx.txid).map(_.fundingTx)
-      rollbackDualFundingTxs(otherFundingTxs)
-      stay() using d.copy(commitments = confirmed.commitments, fundingTx = confirmed.fundingTx, previousFundingTxs = Nil) storing()
-    } else {
-      log.error(s"internal error: a funding tx confirmed that doesn't match any of our funding txs: ${w.tx.txid}")
```
pm47, line 79: Small bug fixed here: `rollbackDualFundingTxs` was missing.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -584,10 +579,8 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
               Seq(channelReady)
           }
           goto(WAIT_FOR_DUAL_FUNDING_READY) using DATA_WAIT_FOR_DUAL_FUNDING_READY(commitments, shortIds, channelReady) storing() sending toSend
-        case None =>
-          log.error(s"internal error: the funding tx that confirmed doesn't match any of our funding txs: ${confirmedTx.bin}")
-          rollbackDualFundingTxs(allFundingTxs.map(_.fundingTx))
-          goto(CLOSED)
+        case Failure(t) =>
+          handleLocalError(t, d, Some(w))
```
t-bast, line 583: If you do that the eclair process will stop when that happens, that's not what we'd like...
I don't think we want to spend our local commitment either, that doesn't make any sense.
I don't see what else we can do than going to either `CLOSED` or `ERR_INFORMATION_LEAK` in that case (which should never happen anyway).

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
```diff
@@ -59,25 +59,34 @@ trait DualFundingHandlers extends CommonFundingHandlers {
     }
   }
 
+  def pruneCommitments(w: WatchFundingConfirmedTriggered, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED): Try[DualFundingTx] = {
```
t-bast, line 62: Why return a `Try`? This can simply be an `Option[DualFundingTx]`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
```diff
@@ -59,25 +59,34 @@ trait DualFundingHandlers extends CommonFundingHandlers {
...
+   */
   def handleDualFundingConfirmedOffline(w: WatchFundingConfirmedTriggered, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) = {
-    if (w.tx.txid == d.commitments.fundingTxId) {
-      watchFundingTx(d.commitments)
-      context.system.eventStream.publish(TransactionConfirmed(d.channelId, remoteNodeId, w.tx))
-      // We can forget previous funding attempts now that the funding tx is confirmed.
-      rollbackDualFundingTxs(d.previousFundingTxs.map(_.fundingTx))
-      stay() using d.copy(previousFundingTxs = Nil) storing()
-    } else if (d.previousFundingTxs.exists(_.commitments.fundingTxId == w.tx.txid)) {
-      log.info("channelId={} was confirmed at blockHeight={} txIndex={} with a previous funding txid={}", d.channelId, w.blockHeight, w.txIndex, w.tx.txid)
```
t-bast, line 70: There was a reason why that function only added a log line when a previous transaction confirmed, this ensured the log line wouldn't be repeated every time you restart eclair if the peer doesn't reconnect. This can be annoying for Phoenix swaps for example, which could cause logs from many channels to pollute the logs after every restart...

It's easy to fix though:

- remove the `log.info` from `pruneCommitments`
- move it to the event handler of `ChannelOpenDualFunded`
- in `handleDualFundingConfirmedOffline`, in the `Success` case, we can compare the confirmed txId with what we previously had in `d`, and if they don't match add the log line that says a previous funding tx confirmed

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
pm47: I agree, it doesn't really make sense.

How about simply `stay()`? It's not really a case of `ERR_INFORMATION_LEAK` I think, more like a bug, maybe it's best to just wait.

For the sake of argument, we could almost call `.get` on the `find`, right?

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
pm47: I had to have a `Throwable` to feed to `handleLocalError`, but that was before https://github.com/ACINQ/eclair/pull/2557#discussion_r1063599710.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
pm47: You are referring to the watch set here:
https://github.com/ACINQ/eclair/blob/7b9a4a15539913c43e6f7c23fb01e09861612843/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala#L320-L322

Shouldn't we only set it when there are multiple competing funding txs instead?

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
t-bast: > For the sake of argument, we could almost call .get on the find, right?

That's true, but I'm always very wary of that...but just adding an error log and doing `stay()` works fine for me!

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
t-bast: We have to set those watches to make sure we detect offline force-closes and correctly get our main output back...

pm47: Right, this is where we put the `WatchFundingSpent`, gotcha.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
pm47: Done as part of [ff5aada](https://github.com/ACINQ/eclair/pull/2557/commits/ff5aada6abce62cb8081400bbf8666577c890fd3).

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
pm47: I went with a slight variation of your suggestion in [d92c894](https://github.com/ACINQ/eclair/pull/2557/commits/d92c894384a8378221966e72b9bac8a62145ad9d).

In `handleDualFundingConfirmedOffline`, I believe that we are really interested in the case where we are actuzlly pruning _any_ funding tx, rather than a previous vs latest.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/DualFundingHandlers.scala
```diff
@@ -59,25 +59,39 @@ trait DualFundingHandlers extends CommonFundingHandlers {
...
+        // ERR_INFORMATION_LEAK, because here we receive a response from the watcher from a WatchConfirmed that we put
+        // ourselves and somehow forgot.
+        log.error(s"internal error: the funding tx that confirmed doesn't match any of our funding txs: ${w.tx.bin}")
+        None
+    }
+  }
+
+  /**
+   * We could just ignore the event and handle it after the next reconnection, but this allows us to eagerly rollback
+   * and forget deprecated funding transactions.
```
t-bast, line 84: This comment should mention that it's more than that, it's a matter of funds safety: if we didn't handle this event while offline, our peer could force-close and never reconnect, and we wouldn't claim our funds back.

pm47: Fixed in [07e6465](https://github.com/ACINQ/eclair/pull/2557/commits/07e646591988dd28c52070ac65589e3ce6c86cc1).

Interestingly, in parallel I also added an additional factorization in commit [5c0c189](https://github.com/ACINQ/eclair/pull/2557/commits/5c0c189e9c92dc04c83a3a06ce57b430a1d7a25a) which made the `watchFundingTx` appear expliclty in `handleDualFundingConfirmedOffline`. I wasn't too sure about that commit, but this makes me think it's directionnally correct, because the critical operation is more obvious, while it was hidden before.

t-bast (approved):
LGTM!

PR #2558, opened 2023-01-06
Wait for tx to be published for zero-conf

t-bast:
When using zero-conf, we wait for the transaction to be published before using a channel. This guarantees that the bitcoind wallet has successfully recorded the transaction and won't double-spend it.

Otherwise if the transaction fails to publish and the utxos are unlocked, or if bitcoind is restarted, we may accidentally double-spend it which would break the channel and be very painful to recover.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenSingleFunded.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenSingleFunded.scala
@@ -292,13 +292,10 @@ trait ChannelOpenSingleFunded extends SingleFundingHandlers with ErrorHandlers {
               log.info(s"waiting for them to publish the funding tx for channelId=$channelId fundingTxid=${commitments.fundingTxId}")
               watchFundingTx(commitments)
               Funding.minDepthFundee(nodeParams.channelConf, commitments.localParams.initFeatures, fundingAmount) match {
-                case Some(fundingMinDepth) =>
-                  blockchain ! WatchFundingConfirmed(self, commitments.fundingTxId, fundingMinDepth)
-                  goto(WAIT_FOR_FUNDING_CONFIRMED) using DATA_WAIT_FOR_FUNDING_CONFIRMED(commitments, None, nodeParams.currentBlockHeight, None, Right(fundingSigned)) storing() sending fundingSigned
-                case None =>
-                  val (shortIds, channelReady) = acceptFundingTx(commitments, RealScidStatus.Unknown)
-                  goto(WAIT_FOR_CHANNEL_READY) using DATA_WAIT_FOR_CHANNEL_READY(commitments, shortIds, channelReady) storing() sending Seq(fundingSigned, channelReady)
+                case Some(fundingMinDepth) => blockchain ! WatchFundingConfirmed(self, commitments.fundingTxId, fundingMinDepth)
+                case None => blockchain ! WatchPublished(self, commitments.fundingTxId)
               }
+              goto(WAIT_FOR_FUNDING_CONFIRMED) using DATA_WAIT_FOR_FUNDING_CONFIRMED(commitments, None, nodeParams.currentBlockHeight, None, Right(fundingSigned)) storing() sending fundingSigned
           }
       }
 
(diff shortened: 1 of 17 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenSingleFunded.scala
```diff
@@ -292,13 +292,10 @@ trait ChannelOpenSingleFunded extends SingleFundingHandlers with ErrorHandlers {
...
               watchFundingTx(commitments)
               Funding.minDepthFundee(nodeParams.channelConf, commitments.localParams.initFeatures, fundingAmount) match {
-                case Some(fundingMinDepth) =>
-                  blockchain ! WatchFundingConfirmed(self, commitments.fundingTxId, fundingMinDepth)
-                  goto(WAIT_FOR_FUNDING_CONFIRMED) using DATA_WAIT_FOR_FUNDING_CONFIRMED(commitments, None, nodeParams.currentBlockHeight, None, Right(fundingSigned)) storing() sending fundingSigned
-                case None =>
-                  val (shortIds, channelReady) = acceptFundingTx(commitments, RealScidStatus.Unknown)
-                  goto(WAIT_FOR_CHANNEL_READY) using DATA_WAIT_FOR_CHANNEL_READY(commitments, shortIds, channelReady) storing() sending Seq(fundingSigned, channelReady)
+                case Some(fundingMinDepth) => blockchain ! WatchFundingConfirmed(self, commitments.fundingTxId, fundingMinDepth)
+                case None => blockchain ! WatchPublished(self, commitments.fundingTxId)
```
pm47, line 296: The fundee case in single-funding wasn't problematic, I supposed you did it for consistency?

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenSingleFunded.scala
```diff
@@ -377,13 +373,23 @@ trait ChannelOpenSingleFunded extends SingleFundingHandlers with ErrorHandlers {
...
       if (remoteChannelReady.alias_opt.isDefined && d.commitments.localParams.isInitiator) {
         log.info("this channel isn't zero-conf, but we are funder and they sent an early channel_ready with an alias: no need to wait for confirmations")
-        val (shortIds, localChannelReady) = acceptFundingTx(d.commitments, RealScidStatus.Unknown)
-        self ! remoteChannelReady
         // NB: we will receive a WatchFundingConfirmedTriggered later that will simply be ignored
-        goto(WAIT_FOR_CHANNEL_READY) using DATA_WAIT_FOR_CHANNEL_READY(d.commitments, shortIds, localChannelReady) storing() sending localChannelReady
-      } else {
-        log.debug("received their channel_ready, deferring message")
-        stay() using d.copy(deferred = Some(remoteChannelReady)) // no need to store, they will re-send if we get disconnected
+        blockchain ! WatchPublished(self, d.commitments.fundingTxId)
```
pm47, line 377: This is a bit weird: we rely on our peer to tell us whether the tx was published. But they could do what we were doing before and this would be inaccurate.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenSingleFunded.scala
t-bast: Yes, I thought it was more consistent that way, especially in the tests.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenSingleFunded.scala
t-bast: > This is a bit weird: we rely on our peer to tell us whether the tx was published.

I don't understand what you mean, we're not relying on them to tell us if the tx was published, we set a watch to see it published in our own mempool? They're simply signaling that they're ok with zero-conf by sending `channel_ready`, so we agree to do zero-conf once we've made sure the tx was published.

pm47: I'm wearing 👓 today, but I still misread that you were emitting the _TransactionPublished_ event 🤷‍♂️.

pm47 (approved):
I can't think of a regression that could be caused by the removal of `WAIT_FOR_DUAL_FUNDING_CREATED`->`WAIT_FOR_DUAL_FUNDING_READY` transition.

pm47 (approved):

PR #2559, opened 2023-01-10
Always lock utxos when funding transactions

t-bast:
We must *always* lock inputs selected for funding, otherwise locking wouldn't work at all, as the following scenario highlights:

- we fund a transaction for which we don't lock utxos
- we fund another unrelated transaction for which we lock utxos
 - the second transaction ends up using the same utxos as the first one
 - but the first transaction confirms, invalidating the second one

This would break the assumptions of the second transaction: its inputs are locked,  so it doesn't expect to potentially be double-spent.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/OnChainWallet.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/blockchain/OnChainWallet.scala
@@ -33,7 +33,7 @@ trait OnChainChannelFunder {
   import OnChainWallet._
 
   /** Fund the provided transaction by adding inputs (and a change output if necessary). */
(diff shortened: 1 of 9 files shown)


pm47 (commented):
I'm confused: we were always locking utxos already, you're just hardcoding it right?

t-bast:
> I'm confused: we were always locking utxos already, you're just hardcoding it right?

Yes exactly: we were already good on `master` since we always lock utxos (except in some tests).

But in some new stuff I was working on, I initially thought it was clever to *not* lock utxos, until I realized we can't, so I decided to make it very clear that we should always lock utxos: this was future implementers won't waste time building something on top of non-locked utxos.

pm47 (approved):

PR #2560, opened 2023-01-10
Skip waiting for empty `tx_signatures`

t-bast:
When our peer doesn't contribute to an `interactive-tx` session, we don't need to wait for their `tx_signatures`: they will be empty anyway, so we can create them locally and publish the transaction immediately.

Since Phoenix sends an encrypted channel backup in `tx_signatures`, we may need to explicitly update the stored backup when we receive what appears to be a redundant `tx_signatures`.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
@@ -763,8 +763,9 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
       // When both peers contribute the same amount, the peer with the lowest pubkey must transmit its `tx_signatures` first.
       LexicographicalOrdering.isLessThan(commitments.localParams.nodeId.value, commitments.remoteNodeId.value)
     } else {
-      // The peer with the lowest total of input amount must transmit its `tx_signatures` first.
-      completeTx.localAmountIn < completeTx.remoteAmountIn
+      // If our peer didn't contribute to the transaction, we don't need to wait for their `tx_signatures`.
+      // Otherwise, the peer with the lowest total of input amount must transmit its `tx_signatures` first.
+      completeTx.remoteAmountIn == 0.sat || completeTx.localAmountIn < completeTx.remoteAmountIn
     }
     if (shouldSignFirst) {
       signTx(completeTx, None)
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
```diff
@@ -763,8 +763,9 @@ private class InteractiveTxBuilder(replyTo: ActorRef[InteractiveTxBuilder.Respon
       // When both peers contribute the same amount, the peer with the lowest pubkey must transmit its `tx_signatures` first.
       LexicographicalOrdering.isLessThan(commitments.localParams.nodeId.value, commitments.remoteNodeId.value)
     } else {
-      // The peer with the lowest total of input amount must transmit its `tx_signatures` first.
-      completeTx.localAmountIn < completeTx.remoteAmountIn
+      // If our peer didn't contribute to the transaction, we don't need to wait for their `tx_signatures`.
+      // Otherwise, the peer with the lowest total of input amount must transmit its `tx_signatures` first.
+      completeTx.remoteAmountIn == 0.sat || completeTx.localAmountIn < completeTx.remoteAmountIn
```
pm47, line 768: Nit: I think clarity beats conciseness in this instance:

```scala
    val shouldSignFirst = if (completeTx.localAmountIn == completeTx.remoteAmountIn) {
      // When both peers contribute the same amount, the peer with the lowest pubkey must transmit its `tx_signatures` first.
      LexicographicalOrdering.isLessThan(commitments.localParams.nodeId.value, commitments.remoteNodeId.value)
    } else if (completeTx.remoteAmountIn == 0.sat) {
      // If our peer didn't contribute to the transaction, we don't need to wait for their `tx_signatures`, they will be
      // empty anyway.
      true
    } else {
      // Otherwise, the peer with the lowest total of input amount must transmit its `tx_signatures` first.
      completeTx.localAmountIn < completeTx.remoteAmountIn
    }
```

pm47 (approved):
LGTM, just a nit

eclair-core/src/main/scala/fr/acinq/eclair/channel/InteractiveTxBuilder.scala
t-bast: Heh, that was actually the first version of the code I wrote, done in https://github.com/ACINQ/eclair/pull/2560/commits/3d0c4297c6760db27f722b7d1f83cda483568d14

pm47 (approved):

PR #2562, opened 2023-01-10
Update final onchain address when possible

sstone:
When a channel gets closed, and when possible, we send all onchain funds to a recent onchain address and not to an address that was generated when the channel was created. This is possible in all cases except when mutual-closing a channel that sets `upfront-shutdown-script`.


sstone:
replaced by https://github.com/ACINQ/eclair/pull/2565

PR #2566, opened 2023-01-13
Add offer manager

thomash-acinq:
Because offers are a very generic mechanism, handling them can require interacting with an inventory system (do we actually have the quantity that the payer is requesting) or other such systems which do not have their place inside eclair. For this reason offer handlers must be implemented as plugins that communicate with the offer manager. On startup, the offer handlers must register their offers with the offer manager, the offer manager will then forward the invoice requests and blinded payments to the relevant offer handler for approval.

Because offer invoices can be requested by  any node, we do not store them in DB when they are created (which could be a DoS vector) but only when they are paid. To do that, we store some metadata about the invoice (offer id, preimage, payer id, created at, quantity, amount, features and extra data from the offer handler) in the pathId of the blinded route to use for the payment. This data is encrypted and signed by us so that it can't be forged by the payer. By default, this is not enough to fully reconstruct the invoice, so what is eventually stored in the DB once the payment succeeds does not exactly match the invoice that was sent to the payer, if that's a requirement for some use case, all of it can be put in the extra data passed by the offer handler.


eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -187,15 +188,24 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
...
 
-    case DoFulfill(preimage, MultiPartPaymentFSM.MultiPartPaymentSucceeded(paymentHash, parts)) if doHandle(paymentHash) =>
+    case DoFulfill(payment, MultiPartPaymentFSM.MultiPartPaymentSucceeded(paymentHash, parts)) if doHandle(paymentHash) =>
       Logs.withMdc(log)(Logs.mdc(paymentHash_opt = Some(paymentHash))) {
         log.debug("fulfilling payment for amount={}", parts.map(_.amount).sum)
         val received = PaymentReceived(paymentHash, parts.map {
           case p: MultiPartPaymentFSM.HtlcPart => PaymentReceived.PartialPayment(p.amount, p.htlc.channelId)
         })
-        if (db.receiveIncomingPayment(paymentHash, received.amount, received.timestamp)) {
+        val recordedinDb = payment match {
```
t-bast, line 197: nit:
```suggestion
        val recordedInDb = payment match {
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -274,16 +284,25 @@ object MultiPartHandler {
    * @param routes              routes that must be blinded and provided in the invoice.
    * @param router              router actor.
    * @param paymentPreimage_opt payment preimage.
+   * @param pathId_opt          path id for the payment paths. If None, a different one will be randomly generated for
+   *                            each path and the mapping blinding point -> path id will be store in the database.
```
t-bast, line 288:
```suggestion
   *                            each path and the mapping blinding point -> path id will be stored in the database.
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -274,16 +284,25 @@ object MultiPartHandler {
...
   case class ReceiveOfferPayment(nodeKey: PrivateKey,
                                  invoiceRequest: InvoiceRequest,
                                  routes: Seq[ReceivingRoute],
                                  router: ActorRef,
                                  paymentPreimage_opt: Option[ByteVector32] = None,
+                                 pathId_opt: Option[ByteVector] = None,
+                                 additionalTlvs: Set[InvoiceTlv] = Set.empty,
+                                 customTlvs: Set[GenericTlv] = Set.empty,
+                                 storeInDb: Boolean = true,
                                  paymentType: String = PaymentType.Blinded) extends ReceivePayment {
```
t-bast, line 301: We should probably remove that argument and set it ourselves inside this function, especially if we use a new payment type for the DB-less offer invoices.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -274,16 +284,25 @@ object MultiPartHandler {
...
     require(nodeKey.publicKey == invoiceRequest.offer.nodeId, "the node id of the invoice must be the same as the one from the offer")
     require(routes.forall(_.nodes.nonEmpty), "each route must have at least one node")
     require(invoiceRequest.offer.amount.nonEmpty || invoiceRequest.amount.nonEmpty, "an amount must be specified in the offer or in the invoice request")
+    require(storeInDb || paymentPreimage_opt.nonEmpty, "the preimage will be lost")
```
t-bast, line 305: I don't understand this requirement, the preimage shouldn't be lost, if will be included in the payment metadata in our blinded path?

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -362,29 +381,41 @@ object MultiPartHandler {
...
+    }
   }
 
   object GetIncomingPaymentActor {
 
     // @formatter:off
     sealed trait Command
     case class GetIncomingPayment(replyTo: ActorRef) extends Command
+    case class PaymentFound(payment: IncomingBlindedPayment) extends Command
+    object NoPayment extends Command
```
t-bast, line 412: This is rather `RejectPayment`, we also send that command for invalid incoming payment.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -395,7 +426,8 @@ object MultiPartHandler {
                     case Some(payment: IncomingStandardPayment) => replyTo ! ProcessPacket(packet.add, payload, Some(payment))
                     case None => replyTo ! ProcessPacket(packet.add, payload, None)
                   }
-                case payload: FinalPayload.Blinded =>
+                  Behaviors.stopped
+                case payload: FinalPayload.Blinded if payload.pathId.length == 32 =>
```
t-bast, line 430: This isn't very robust, can't we instead look for tlvs in the payload that are only added for DB-less invoices? That could be exposed as a boolean field directly defined in `FinalPayload.Blinded`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -405,7 +437,26 @@ object MultiPartHandler {
...
+                  waitForPayment(nodeParams, replyTo, packet.add, payload)
               }
+          }
+        }
+      }
+    }
+
+    def waitForPayment(nodeParams: NodeParams, replyTo: ActorRef, add: UpdateAddHtlc, payload: FinalPayload.Blinded): Behavior[Command] = {
+      Behaviors.setup { context =>
+        Behaviors.withMdc(Logs.mdc(category_opt = Some(LogCategory.PAYMENT), paymentHash_opt = Some(add.paymentHash))) {
```
t-bast, line 452: No need to call `setup` / `withMdc` again, you should pass the `context` as argument.

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
```diff
@@ -0,0 +1,250 @@
...
+import fr.acinq.eclair.wire.protocol.{GenericTlv, MessageOnion, OfferTypes, OnionMessagePayloadTlv, TlvStream}
+import fr.acinq.eclair.{Feature, Features, MilliSatoshi, NodeParams, TimestampMilli, TimestampSecond, randomBytes32}
+import scodec.bits.ByteVector
+import scodec.codecs._
+import scodec.{Attempt, Codec, DecodeResult}
+
+import scala.concurrent.duration.DurationInt
+import scala.util.{Success, Try}
+
+object OfferManager {
```
t-bast, line 40: Can you move the `offer` folder inside the `payment` folder?

Also, this file needs some documentation. DB-less invoice schemes should even eventually be documented in BLIPs, to allow implementations to learn from each other and find optimal strategies for what data to include in blinded paths. You should explain what you choose to include in the payment path, what trade-offs were made and how this guarantees security. Think about reviewers and future maintainers of this component, and try to make their life as easy as possible by adding comments around subtle parts of the code and on important fields/classes. You should also explain that you're deferring parts of the flow to plugins, and what those plugins' responsibilities are.

I'm particularly interested in detailing how you ensure that the blinded paths used for payments are the ones that were generated with the invoice.

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
```diff
@@ -0,0 +1,250 @@
...
+  // @formatter:off
+  sealed trait Command
+  case class RegisterOffer(offer: Offer, nodeKey: PrivateKey, pathId_opt: Option[ByteVector32], handler: ActorRef[HandlerCommand]) extends Command
```
t-bast, line 43: It would be useful to document that the `pathId_opt` here is what has been included in the blinded paths associated with the offer and must be verified when receiving `invoice_request`s, otherwise it's easy to get confused between which `path_id`s are associated to what (this is different from the `path_id` that will be used for the payments, right?).

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
```diff
@@ -0,0 +1,250 @@
...
+  case class DisableOffer(offer: Offer) extends Command
+  case class RequestInvoice(messagePayload: MessageOnion.FinalPayload, postman: ActorRef[Postman.SendMessage]) extends Command
+  case class Payment(replyTo: ActorRef[MultiPartHandler.GetIncomingPaymentActor.Command], paymentHash: ByteVector32, payload: FinalPayload.Blinded) extends Command
```
t-bast, line 46: nit: this should be called `ReceivePayment`, all the other commands are actions, not nouns.

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
```diff
@@ -0,0 +1,250 @@
...
+
+  sealed trait HandlerCommand
+  case class HandleInvoiceRequest(replyTo: ActorRef[InvoiceRequestActor.Command], invoiceRequest: InvoiceRequest) extends HandlerCommand
+  case class HandlePayment(replyTo: ActorRef[PaymentActor.Command], offerId: ByteVector32, data: ByteVector) extends HandlerCommand
```
t-bast, line 50: It's very unclear to just pass a `data` blob, why don't you pass the `PaymentMetadata` class?

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
```diff
@@ -0,0 +1,250 @@
...
+  // @formatter:on
+
+  case class RegisteredOffer(offer: Offer, nodeKey: PrivateKey, pathId_opt: Option[ByteVector32], handler: ActorRef[HandlerCommand])
+
+  case class PaymentMetadata(preimage: ByteVector32,
```
t-bast, line 55: Can you share some numbers on the space used in the onion when paying such invoices?

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
```diff
@@ -0,0 +1,250 @@
...
+      ("createdAt" | timestampSecond) ::
+      ("quantity" | uint64overflow) ::
+      ("amount" | millisatoshi) ::
+      ("features" | lengthPrefixedFeaturesCodec) ::
+      ("extraData" | bytes)).as[PaymentMetadata]
+
+  case class SignedMetadata(signature: ByteVector64,
+                            offerId: ByteVector32,
+                            metadata: ByteVector) {
+    def verify(publicKey: PublicKey): Try[PaymentMetadata] = Try {
```
t-bast, line 75: Please remove all usages of `Try` in this file, all the error cases should be easy to handle cleanly without throwing exceptions.

t-bast (commented):
Concept ACK on the architecture, this is going in the right direction.

eclair-core/src/test/scala/fr/acinq/eclair/integration/OfferIntegrationSpec.scala
```diff
@@ -0,0 +1,168 @@
...
+import fr.acinq.eclair.router.Router
+import fr.acinq.eclair.wire.protocol.OfferTypes._
+import fr.acinq.eclair.{CltvExpiryDelta, EclairImpl, Features, Kit, MilliSatoshiLong, randomBytes32, randomKey}
+import scodec.bits.ByteVector
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.jdk.CollectionConverters._
+
+class OfferIntegrationSpec extends IntegrationSpec {
```
t-bast, line 44: You don't need a real `bitcoind` for those tests, it makes these tests much slower than they need to be. You should use the `MinimalNodeFixture` framework instead of spinning up complete eclair nodes.

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
thomash-acinq: Moved and added more documentation.

The preimage is in the blinded route, so if the blinded path used for payment does not match the invoice we have no way to fulfill the payment.

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
thomash-acinq: It adds 181 bytes plus the data added by the plugin.

eclair-core/src/main/scala/fr/acinq/eclair/offer/OfferManager.scala
thomash-acinq: I've removed most of it. I've kept the `Try[Bolt12Invoice]` because that comes from`MultiPartHandler` and removing it would require a major refactoring.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -187,15 +188,24 @@ class MultiPartHandler(nodeParams: NodeParams, register: ActorRef, db: IncomingP
...
+    case DoFulfill(payment, MultiPartPaymentFSM.MultiPartPaymentSucceeded(paymentHash, parts)) if doHandle(paymentHash) =>
       Logs.withMdc(log)(Logs.mdc(paymentHash_opt = Some(paymentHash))) {
         log.debug("fulfilling payment for amount={}", parts.map(_.amount).sum)
         val received = PaymentReceived(paymentHash, parts.map {
           case p: MultiPartPaymentFSM.HtlcPart => PaymentReceived.PartialPayment(p.amount, p.htlc.channelId)
         })
-        if (db.receiveIncomingPayment(paymentHash, received.amount, received.timestamp)) {
+        val recordedinDb = payment match {
+          // Incoming blinded payments corresponding to invoice requests are not stored in the database until they have been paid.
+          case IncomingBlindedPayment(invoice, preimage, paymentType, None, _, _) =>
```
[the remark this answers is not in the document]
thomash-acinq: Blinded payments are now only used for offers.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
thomash-acinq: Removed

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
thomash-acinq: The payment metadata has already been created at this point. If the preimage is not generated yet, then it can't be in the payment metadata.
Anyways, `paymentPreimage` and `pathId` are no longer optional.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
thomash-acinq: We try to get an `IncomingPayment` matching the current HTLC and we receive either `PaymentFound` or `NoPayment`. I don't think `RejectPayment` would be a better name here. `RejectPacket` is what this actor returns if it can't find a matching `IncomingPayment`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
thomash-acinq: Blinded payments are now always DB-less.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
thomash-acinq: Done

eclair-core/src/test/scala/fr/acinq/eclair/integration/OfferIntegrationSpec.scala
thomash-acinq: I've moved all the tests to `BlindedPaymentSpec` and `PaymentIntegrationSpec`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
```diff
@@ -292,18 +311,19 @@ object MultiPartHandler {
...
     def apply(nodeParams: NodeParams): Behavior[Command] = {
       Behaviors.setup { context =>
-        Behaviors.receiveMessage {
+        Behaviors.receiveMessagePartial {
           case CreateInvoice(replyTo, receivePayment) =>
             val paymentPreimage = receivePayment.paymentPreimage_opt.getOrElse(randomBytes32())
             val paymentHash = Crypto.sha256(paymentPreimage)
             receivePayment match {
               case r: ReceiveStandardPayment =>
-                Try {
```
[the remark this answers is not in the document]
t-bast: But these `require` are here to catch _developer_ bugs, not _runtime_ issues: none of them can trigger outside of unit tests, since we don't let node operators disable `Features.PaymentSecret` and we make sure that we don't pass in features that contain `Features.RouteBlinding`. None of these `require` can be influenced by outside data, so we shouldn't treat this as a `Try` (otherwise we would have made that `Bolt11Invoice` return a `Try[Bolt11Invoice]`).

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
t-bast: What I meant is that the `OfferManager` sends this `NoPayment` command in a lot of cases, most of which are for a different reason than the fact there is no matching payment.

Most of the cases where the `OfferManager` sends this command are when the payment we receive is malformed and want to reject it, which indicates that the name is misleading...

eclair-core/src/main/scala/fr/acinq/eclair/payment/offer/OfferManager.scala
```diff
@@ -0,0 +1,331 @@
...
+   * - the amount to pay
+   * - optional data from the handler
+   *
+   * This data is what we need to recreate an invoice similar to the one that was actually sent to the payer. It will
+   * not be exactly the same (notably the blinding route will be different) but it will contain what we need to fulfill
+   * the payment HTLC.
+   *
+   * It takes 181 bytes plus what the handler adds.
+   */
+  case class SignedMetadata(signature: ByteVector64,
```
t-bast, line 121: Using a signature is quite wasteful (64 bytes is 4.6% of the total onion size...), I'm pretty sure we can do much better (for example by encrypting the metadata with a key derived from the `offerId` and our `nodeKey` instead of signing it). This is an area that other implementations have given a lot of thought about, can you send a mail to the lightning-dev mailing list detailing the current DB-less scheme you're proposing and asking others to share the schemes they plan on implementing?

This way we can gather knowledge from many people and find a better scheme. What's nice is that all of this is transient and short-lived anyway, so we can change our DB-less scheme by simply updating eclair without caring about backwards-compatibility with the old scheme.

eclair-core/src/main/scala/fr/acinq/eclair/payment/offer/OfferManager.scala
```diff
@@ -0,0 +1,331 @@
...
+    case class ApproveRequest(amount: MilliSatoshi,
+                              routes: Seq[ReceivingRoute],
+                              data: ByteVector,
+                              additionalTlvs: Set[InvoiceTlv] = Set.empty,
+                              customTlvs: Set[GenericTlv] = Set.empty) extends Command
+
+    /**
+     * Sent by the offer handler to reject the request. For instance because stock has been exhausted.
+     */
+    object RejectRequest extends Command
```
t-bast, line 221: In that case, the sender of `invoice_request` isn't getting any response...shouldn't we at least respond with `invoice_error` to let them know that they shouldn't retry? Otherwise they'll probably just retry sending the `invoice_request`.

eclair-core/src/main/scala/fr/acinq/eclair/payment/offer/OfferManager.scala
```diff
@@ -0,0 +1,331 @@
...
+    private case class WrappedInvoice(invoice: Try[Bolt12Invoice]) extends Command
+
+    private case class WrappedOnionMessageResponse(response: Postman.OnionMessageResponse) extends Command
+
+    def apply(nodeParams: NodeParams,
+              invoiceRequest: InvoiceRequest,
+              offerHandler: ActorRef[HandleInvoiceRequest],
+              nodeKey: PrivateKey,
+              router: akka.actor.ActorRef,
+              replyPath: OnionMessages.Destination,
```
t-bast, line 232: Naming that field `replyPath` is a bit confusing here, at first I thought this was _our_ reply path but we're not expecting a reply...what about naming this `senderPath` or something similar to clear that confusion?

eclair-core/src/main/scala/fr/acinq/eclair/payment/offer/OfferManager.scala
```diff
@@ -0,0 +1,331 @@
...
+
+    private def waitForSent(): Behavior[Command] = {
+      Behaviors.receiveMessagePartial {
+        case WrappedOnionMessageResponse(_) =>
+          Behaviors.stopped
+      }
+    }
+  }
+
+  object PaymentActor {
```
t-bast, line 287: This actor needs to use a timeout after which we automatically reject the payment. We cannot fully trust plugins to be correctly written, and if a plugin doesn't answer and the HTLC times out, this will lead to a channel force-close which isn't acceptable. If the plugin doesn't answer in a timely manner, the safe choice is to fail the HTLC and stop this actor.

eclair-core/src/main/scala/fr/acinq/eclair/payment/offer/OfferManager.scala
```diff
@@ -0,0 +1,331 @@
...
+
+    /**
+     * Sent by the offer handler to reject the payment. For instance because stock has been exhausted.
+     */
+    object RejectPayment extends Command
+
+    def apply(nodeParams: NodeParams, replyTo: ActorRef[MultiPartHandler.GetIncomingPaymentActor.Command], offer: Offer, metadata: PaymentMetadata): Behavior[Command] = {
+      Behaviors.receiveMessage {
+        case AcceptPayment(additionalTlvs, customTlvs) =>
+          val dummyKey = PublicKey(hex"020000000000000000000000000000000000000000000000000000000000000001")
```
t-bast, line 305: nit:

```suggestion
          val dummyKey = Transactions.PlaceHolderPubKey
```

eclair-core/src/main/scala/fr/acinq/eclair/payment/offer/OfferManager.scala
```diff
@@ -0,0 +1,331 @@
...
+          // This invoice is not the one we've sent (we don't store the real one as it would be a DoS vector) but it shares all the important bits with the real one.
+          val dummyInvoice = Bolt12Invoice(TlvStream(offer.records.records ++ Seq[InvoiceTlv](
```
t-bast, line 307: This isn't very satisfying, we're forced to include ~200 bytes of dummy fields (because they are required fields for a valid Bolt12 invoice) and the resulting invoice will look weird when displayed.

I'm not 100% sure it's a good idea, but have you looked into defining a `case class PaidBolt12Invoice(records: TlvStream[InvoiceTlv]) extends Invoice` that wouldn't have the requirements of `Bolt12Invoice` and would let us include only the fields that we really need to store? We can define a separate `val hrp = "lnpi"` for it to make the serialization more explicit (which would be future-proof if at some point we start storing "real" Bolt12 invoices in the DB)?

eclair-core/src/main/scala/fr/acinq/eclair/payment/receive/MultiPartHandler.scala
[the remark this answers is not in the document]
t-bast: I gave it a try in https://github.com/ACINQ/eclair/commit/07c1a84663ebf88aea3057fa82112b367408c95a and it's looking better IMO

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -165,9 +166,9 @@ trait Eclair {
 
   def sendOnionMessage(intermediateNodes: Seq[PublicKey], destination: Either[PublicKey, Sphinx.RouteBlinding.BlindedRoute], replyPath: Option[Seq[PublicKey]], userCustomContent: ByteVector)(implicit timeout: Timeout): Future[SendOnionMessageResponse]
 
-  def payOffer(offer: Offer, amount: MilliSatoshi, quantity: Long, externalId_opt: Option[String] = None, maxAttempts_opt: Option[Int] = None, maxFeeFlat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None, pathFindingExperimentName_opt: Option[String] = None)(implicit timeout: Timeout): Future[UUID]
+  def payOffer(offer: Offer, amount: MilliSatoshi, quantity: Long, path: Seq[PublicKey] = Nil, externalId_opt: Option[String] = None, maxAttempts_opt: Option[Int] = None, maxFeeFlat_opt: Option[Satoshi] = None, maxFeePct_opt: Option[Double] = None, pathFindingExperimentName_opt: Option[String] = None)(implicit timeout: Timeout): Future[UUID]
```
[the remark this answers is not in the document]
thomash-acinq: Removed

t-bast (commented):
This is looking mostly good, I've done some refactoring to improve readability and tests in #2618 and then we should be good to go.

thomash-acinq:
Thank you for the refactoring. It looks very nice now.

PR #2568, opened 2023-01-13
Add `GetNode` router API

t-bast:
Add a new API to return information about a specific node: its announcement, number of active channels and total capacity. This has been requested several times by node operators.

### DIFF
--- a/eclair-core/src/test/scala/fr/acinq/eclair/router/RouterSpec.scala
+++ b/eclair-core/src/test/scala/fr/acinq/eclair/router/RouterSpec.scala
@@ -48,6 +48,50 @@ import scala.concurrent.duration._
 
 class RouterSpec extends BaseRouterSpec {
 
+  test("properly announce valid new nodes announcements and ignore invalid ones") { fixture =>
+    import fixture._
+    val eventListener = TestProbe()
+    system.eventStream.subscribe(eventListener.ref, classOf[NetworkEvent])
+    system.eventStream.subscribe(eventListener.ref, classOf[Rebroadcast])
+    val peerConnection = TestProbe()
+
+    {
+      // continue to rebroadcast node updates with deprecated Torv2 addresses
+      val torv2Address = List(NodeAddress.fromParts("hsmithsxurybd7uh.onion", 9735).get)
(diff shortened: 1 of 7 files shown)


eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/PathFinding.scala
```diff
@@ -56,12 +56,18 @@ trait PathFinding {
     }
   }
 
+  val node: Route = postRequest("node") { implicit t =>
```
t-bast, line 59: I called this `node` for consistency with the `nodes` API, but it's a bit weird...should I rename this?

eclair-core/src/test/scala/fr/acinq/eclair/router/RouterSpec.scala
```diff
@@ -1040,49 +1095,4 @@ class RouterSpec extends BaseRouterSpec {
     }
   }
 
-  test("properly announce valid new nodes announcements and ignore invalid ones") { fixture =>
```
t-bast, line 1043: I moved this test (without changes) because it made sense to keep it closer to a very similar test for channels.

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/PathFinding.scala
remyers: It looks like LND uses [`GetNodeInfo`](https://lightning.engineering/api-docs/api/lnd/lightning/get-node-info) or just `node` for REST and I couldn't find an equivalent in [CLN](https://lightning.readthedocs.io/lightning-help.7.html).

Using `node` seems fine for our API.

remyers (commented):
This will be very useful for what I'm doing - so big concept ACK. Changest LGTM too.

eclair-node/src/main/scala/fr/acinq/eclair/api/handlers/PathFinding.scala
t-bast: Great, if `lnd` calls this API `node` we should do the same, consistency will help cross-implementation devs. Thanks for checking that!

pm47 (approved):

PR #2570, opened 2023-01-17
Build message inside Postman

thomash-acinq:
The postman now has a higher level API to send messages: the user provides the content and the paths to use and the postman takes care of building the message packet and sending it to the next node.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
@@ -559,6 +559,9 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
                                 recipient: Either[PublicKey, Sphinx.RouteBlinding.BlindedRoute],
                                 replyPath: Option[Seq[PublicKey]],
                                 userCustomContent: ByteVector)(implicit timeout: Timeout): Future[SendOnionMessageResponse] = {
+    if (replyPath.nonEmpty && (replyPath.get.isEmpty || replyPath.get.last != appKit.nodeParams.nodeId)) {
+      return Future.failed(new Exception("Reply path must end at our node."))
+    }
     TlvCodecs.tlvStream(MessageOnionCodecs.onionTlvCodec).decode(userCustomContent.bits) match {
       case Attempt.Successful(DecodeResult(userTlvs, _)) =>
         val destination = recipient match {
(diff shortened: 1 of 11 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
```diff
@@ -557,30 +556,22 @@ class EclairImpl(appKit: Kit) extends Eclair with Logging {
...
-              case Postman.SendingStatus(failure: MessageRelay.Failure) => SendOnionMessageResponse(sent = false, Some(failure.toString), None)
-            }
-          case Failure(error) => Future.successful(SendOnionMessageResponse(sent = false, failureMessage = Some(error.toString), response = None))
+    TlvCodecs.tlvStream(MessageOnionCodecs.onionTlvCodec).decode(userCustomContent.bits) match {
+      case Attempt.Successful(DecodeResult(userTlvs, _)) =>
+        val destination = recipient match {
+          case Left(key) => OnionMessages.Recipient(key, None)
+          case Right(route) => OnionMessages.BlindedPath(route)
+        }
+        appKit.postman.ask(ref => Postman.SendMessage(intermediateNodes, destination, replyPath.map(_ :+ appKit.nodeParams.nodeId), userTlvs, ref, appKit.nodeParams.onionMessageConfig.timeout))(timeout, appKit.system.scheduler.toTyped).mapTo[Postman.OnionMessageResponse].map {
```
t-bast, line 568: nit: I would expect the user to include themselves in the `replyPath`: if they do, we will end up duplicating the final `nodeId` by adding it ourselves here. Our current [API documentation](https://acinq.github.io/eclair/#messages) doesn't clarify that, we should update it with the next release. I find it slightly more intuitive to ask the user to include their `nodeId` at the end of the `replyPath` when they provide one, but it's your call!

t-bast (commented):
LGTM :+1:

eclair-core/src/main/scala/fr/acinq/eclair/Eclair.scala
t-bast: Thanks! But you're now duplicating it in this line, it should be:
```suggestion
        appKit.postman.ask(ref => Postman.SendMessage(intermediateNodes, destination, replyPath, userTlvs, ref, appKit.nodeParams.onionMessageConfig.timeout))(timeout, appKit.system.scheduler.toTyped).mapTo[Postman.OnionMessageResponse].map {
```

thomash-acinq: How did I miss that! It should be fixed now.

t-bast (approved):

PR #2571, opened 2023-01-18
More changes to splices data model

t-bast:
This is a collection of small independent commits on top of #2540

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -826,7 +826,7 @@ case class Commitments(channelId: ByteVector32,
   /**
    * When reconnecting, we drop all unsigned changes.
    */
-  def discardUnsignedUpdates(implicit log: LoggingAdapter): Commitments = {
+  def discardUnsignedUpdates()(implicit log: LoggingAdapter): Commitments = {
     log.debug("discarding proposed OUT: {}", localChanges.proposed.map(msg2String(_)).mkString(","))
     log.debug("discarding proposed IN: {}", remoteChanges.proposed.map(msg2String(_)).mkString(","))
     val commitments1 = copy(
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
@@ -882,7 +882,7 @@ case class Commitments(channelId: ByteVector32,
 
   def params: Params = Params(channelId, channelConfig, channelFeatures, localParams, remoteParams, channelFlags)
 
-  def common: Common = Common(localChanges, remoteChanges, localNextHtlcId, remoteNextHtlcId, localCommit.index, remoteCommit.index, originChannels, remoteNextCommitInfo.swap.map(waitingForRevocation => WaitForRev(waitingForRevocation.nextRemoteCommit.index, waitingForRevocation.sent, waitingForRevocation.sentAfterLocalCommitIndex)).swap, remotePerCommitmentSecrets)
+  def common: Common = Common(localChanges, remoteChanges, localNextHtlcId, remoteNextHtlcId, localCommit.index, remoteCommit.index, originChannels, remoteNextCommitInfo.swap.map(waitingForRevocation => WaitForRev(waitingForRevocation.sent, waitingForRevocation.sentAfterLocalCommitIndex)).swap, remotePerCommitmentSecrets)
 
   def commitment: Commitment = Commitment(localFundingStatus, remoteFundingStatus, localCommit, remoteCommit, remoteNextCommitInfo.swap.map(_.nextRemoteCommit).toOption)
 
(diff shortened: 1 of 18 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala
```diff
@@ -826,7 +826,7 @@ case class Commitments(channelId: ByteVector32,
   /**
    * When reconnecting, we drop all unsigned changes.
    */
-  def discardUnsignedUpdates(implicit log: LoggingAdapter): Commitments = {
+  def discardUnsignedUpdates()(implicit log: LoggingAdapter): Commitments = {
```
pm47, line 829: Actually this method can be removed entirely, we only need it at the `MetaCommitments` level

eclair-core/src/main/scala/fr/acinq/eclair/channel/MetaCommitments.scala
```diff
@@ -38,19 +38,21 @@ case class Common(localChanges: LocalChanges, remoteChanges: RemoteChanges,
                   originChannels: Map[Long, Origin], // for outgoing htlcs relayed through us, details about the corresponding incoming htlcs
                   remoteNextCommitInfo: Either[WaitForRev, PublicKey], // this one is tricky, it must be kept in sync with Commitment.nextRemoteCommit_opt
                   remotePerCommitmentSecrets: ShaChain) {
+  val nextRemoteCommitIndex = remoteCommitIndex + 1
+
   /**
    * When reconnecting, we drop all unsigned changes.
    */
-  def discardUnsignedUpdates(implicit log: LoggingAdapter): Common = {
+  def discardUnsignedUpdates()(implicit log: LoggingAdapter): Common = {
```
pm47, line 46: I don't think we should add parentheses, as there is no side effect?

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -381,8 +379,8 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
 
   when(WAIT_FOR_DUAL_FUNDING_CONFIRMED)(handleExceptions {
     case Event(txSigs: TxSignatures, d: DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED) =>
-      d.metaCommitments.main.localFundingStatus match {
-        case DualFundedUnconfirmedFundingTx(fundingTx: PartiallySignedSharedTransaction) => InteractiveTxBuilder.addRemoteSigs(d.fundingParams, fundingTx, txSigs) match {
+      d.metaCommitments.main.localFundingStatus.asInstanceOf[DualFundedUnconfirmedFundingTx].sharedTx match {
```
pm47, line 382:
```suggestion
      d.mainFundingTx match {
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -309,17 +309,20 @@ private[channel] object ChannelCodecs3 {
...
 
+    private val dualFundedUnconfirmedFundingTxCodec: Codec[DualFundedUnconfirmedFundingTx] = (
+      ("sharedTx" | signedSharedTransactionCodec) ::
+        ("createdAt" | blockHeight)).as[DualFundedUnconfirmedFundingTx]
+
     val fundingTxStatusCodec: Codec[LocalFundingStatus] = discriminated[LocalFundingStatus].by(uint8)
       .typecase(0x00, optional(bool8, txCodec).as[SingleFundedUnconfirmedFundingTx])
       .typecase(0x01, txCodec.map(tx => SingleFundedUnconfirmedFundingTx(Some(tx))).decodeOnly) // for backward compat
-      .typecase(0x02, signedSharedTransactionCodec.as[DualFundedUnconfirmedFundingTx])
+      .typecase(0x02, dualFundedUnconfirmedFundingTxCodec)
```
pm47, line 319: Did you forget to handle backward compatibility? Would a blockheight of zero make sense?

pm47 (commented):
Changes have been cherry-picked ton #2540, with the following amendment: fcf5698f0.

eclair-core/src/main/scala/fr/acinq/eclair/channel/MetaCommitments.scala
t-bast: It's very weird without parenthesis, `discardXXX` really feels like it's _doing_ something, not just _accessing_ something...

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
t-bast: Nice!

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
t-bast: I didn't realize you had been clever to make it work with backwards-compat! Yes a blockheight of zero totally makes sense, it doesn't really matter if current data isn't migrated to meaningful values.

eclair-core/src/main/scala/fr/acinq/eclair/channel/MetaCommitments.scala
pm47: I put them back in 612b327a8.

PR #2572, opened 2023-01-19
Never lock utxos when funding

t-bast:
We previously locked utxos when funding transactions to avoid accidentally double-spending ourselves if an unrelated transaction was funded concurrently before the first one was published.

It didn't fully protect us against double-spending ourselves though. If bitcoind was restarted while eclair was running, it would clear existing locks and remove those guarantees. But that was an acceptable edge case.

However, dual funding and liquidity ads introduce griefing attacks if we lock our utxos: an attacker can waste our liquidity in funding attempts that will never be published, as explained here:
https://github.com/lightning/bolts/pull/851#discussion_r997537630

The easiest way to get rid of that griefing attack is to never lock our inputs: this ensures that honest users will double-spend the inputs used for sessions with malicious users, at no cost to us. It also greatly simplifies our code because we don't need to worry about unlocking utxos.

Note that once a transaction has been successfully published once, we are then guaranteed that bitcoind will never accidentally double-spend it: we have to call `abandontransaction` to tell bitcoind to forget that transaction and free its inputs.

There is one drawback with this approach: if we have concurrent sessions with honest peers, there is a risk that we will double-spend ourselves and only one of the sessions will succeed. This trade-off is acceptable because:

- the time window where such race condition may happen is small: honest peers will quickly complete the funding flow and publish the signed transaction, which protects against accidental double-spends
- funding on-chain transactions is an infrequent operation: block size is limited, so we have to rate-limit our funding attempts anyway
- to be perfectly safe, all funding code must handle the possibility of being double-spent before the transaction is published: the main danger here was zero-conf channels, but it has been fixed in #2558.
- all existing funding scenarios already handle the case where the transaction cannot be published and detect double-spends
- when a funding attempt fails, it can simply be restarted (e.g. LSP operations that open channels on-the-fly)

It's also important to note that bitcoind funding locks have to either be used all the time or never. We cannot use them in some scenarios and not others, because it wouldn't provide any meaningful guarantees for the scenarios that lock inputs, as shown by the following sample flow:

- fund a transaction for which we don't lock utxos
- fund another unrelated transaction for which we lock utxos
- the second transaction ends up using the same utxos as the first one, since they were unlocked and thus ready to be used
- but the first transaction confirms, invalidating the second one, which incorrectly thought it was safe from double-spends because it used locks

### DIFF
--- a/docs/release-notes/eclair-vnext.md
+++ b/docs/release-notes/eclair-vnext.md
@@ -17,16 +17,6 @@
 
 ### Miscellaneous improvements and bug fixes
 
-#### Strategies to handle locked utxos at start-up (#2278)
-
-If some utxos are locked when eclair starts, it is likely because eclair was previously stopped in the middle of funding a transaction.
-While this doesn't create any risk of loss of funds, these utxos will stay locked for no good reason and won't be used to fund future transactions.
-Eclair offers three strategies to handle that scenario, that node operators can configure by setting `eclair.bitcoind.startup-locked-utxos-behavior` in their `eclair.conf`:
-
-- `stop`: eclair won't start until the corresponding utxos are unlocked by the node operator
-- `unlock`: eclair will automatically unlock the corresponding utxos
-- `ignore`: eclair will leave these utxos locked and start
-
 ## Verifying signatures
 
 You will need `gpg` and our release signing key 7A73FE77DE2C4027. Note that you can get it:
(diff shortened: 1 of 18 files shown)


pm47:
Two more arguments in favor of this trade-off:
- It favors fast peers over slow peers, which is fair.
- We can always add a best effort ordering/backpressure mechanism if we see too much concurrency issues, with the ease of mind in any case we handle double spending gracefully

t-bast:
Another, potentially more reasonable alternative, would be to use a separate wallet for liquidity ads. The main wallet would be used by default and would lock utxos when funding. When we would respond to a liquidity ads from an untrusted node, then we would use the liquidity ads wallet and would never lock utxos.

PR #2574, opened 2023-01-19
Don't send payment secret when using keysend

t-bast:
Apparently lnd rejects keysend payments that contain a payment secret, instead of simply ignoring that field.

The blip doesn't mention this requirement:
https://github.com/lightning/blips/blob/master/blip-0003.md

But we don't have a strong reason to send it either, so let's remove that.

Waiting for @kiwiidb to confirm that this fixes the lnd compatibility issues.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/payment/send/Recipient.scala
@@ -110,7 +110,7 @@ case class SpontaneousRecipient(nodeId: PublicKey,
 
   override def buildPayloads(paymentHash: ByteVector32, route: Route): Either[OutgoingPaymentError, PaymentPayloads] = {
     ClearRecipient.validateRoute(nodeId, route).map(_ => {
-      val finalPayload = NodePayload(nodeId, FinalPayload.Standard.createKeySendPayload(route.amount, totalAmount, expiry, preimage, customTlvs))
+      val finalPayload = NodePayload(nodeId, FinalPayload.Standard.createKeySendPayload(route.amount, expiry, preimage, customTlvs))
       Recipient.buildPayloads(totalAmount, expiry, Seq(finalPayload), route.hops)
     })
   }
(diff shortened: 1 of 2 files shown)


kiwiidb:
I've built and tested this branch on regtest and can confirm that sending keysend payments now works to both LND and CLN nodes.

kiwiidb:
Keysend from LND -> Eclair works as well on this branch. However, CLN -> Eclair fails with `WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS`.

thomash-acinq (approved):

t-bast:
> However, CLN -> Eclair fails with WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS.

That's surprising, can you share logs for that failure?

kiwiidb:
Found them:
```
2023-01-20 09:32:05,984 WARN  f.a.e.p.r.PaymentHandler PAY 
h:9deb8d9b868a8a7e9ffd0aba03ab64892e538673072c7510490dc83fbae632aa -
received payment with expiry too small for amount=1000000 msat totalAmount=1000000 msat
```
This if for the CLN command:
`lightning-cli --network regtest keysend 02ca4adae421945a328735519f6113e9c8d95979a9c573aba6bc26ca72af187564 1000000`.
There is a direct channel between these 2 nodes.

t-bast:
Got it, this is completely independent of keysend. By default eclair uses very safe values of expiry deltas, but when using keysend we can't communicate that to the payer so they may use smaller values. This is generally an issue because accepting that payment creates a risk that if your peer force-closes, you won't be able to get the funds back in time...

You can change that default value in your `eclair.conf`, by setting for example (the default is 30 blocks):

```
eclair.channel.min-final-expiry-delta-blocks = 3
```

We could relax that for keysend, but this is a bit painful to do, I'd rather have senders behave well and send a bigger expiry delta here...

Abdulkbk:
I noticed when one sets:
```
eclair.channel.min-final-expiry-delta-blocks = 6
```

eclair fails with

```
fatal error: requirement failed: channel.min-final-expiry-delta-blocks must be strictly greater than channel.fulfill-safety-before-timeout-blocks; otherwise it may lead to undesired channel closure
```

One needs to explicitly set `eclair.channel.fulfill-safety-before-timeout-blocks` to a value less than the above to fix this. 

I am considering creating a PR that adds doc for keysend because I spent many hours getting CLN to work with Eclair. Is that okay? @t-bast

PR #2576, opened 2023-01-20
Improve `InteractiveTxBuilder` model

pm47:
We introduce a `Purpose` trait, which indicates what the operation is about. Currently there are two possibilities:
- `InitialCommitment`
- `RbfInitialCommitment`

This allows for some nice factoring, but more importantly it will be complemented by a third possibility for splices.

Note that this change preserves the original separation between two main concerns:
- creating the funding tx (in `InteractiveTxParams`)
- creating the commitment (in `InteractiveTxBuilder`).

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
@@ -225,7 +225,7 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
               open.secondPerCommitmentPoint),
             localPushAmount = accept.pushAmount, remotePushAmount = open.pushAmount,
             wallet))
-          txBuilder ! InteractiveTxBuilder.Start(self, Nil)
+          txBuilder ! InteractiveTxBuilder.Start(self)
           goto(WAIT_FOR_DUAL_FUNDING_CREATED) using DATA_WAIT_FOR_DUAL_FUNDING_CREATED(channelId, accept.pushAmount, open.pushAmount, txBuilder, None) sending accept
       }
 
@@ -288,7 +288,7 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
               accept.secondPerCommitmentPoint),
             localPushAmount = d.lastSent.pushAmount, remotePushAmount = accept.pushAmount,
             wallet))
-          txBuilder ! InteractiveTxBuilder.Start(self, Nil)
+          txBuilder ! InteractiveTxBuilder.Start(self)
           goto(WAIT_FOR_DUAL_FUNDING_CREATED) using DATA_WAIT_FOR_DUAL_FUNDING_CREATED(channelId, d.lastSent.pushAmount, accept.pushAmount, txBuilder, None)
       }
 
@@ -464,10 +464,10 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
               val txBuilder = context.spawnAnonymous(InteractiveTxBuilder(
                 nodeParams, fundingParams,
                 commitmentParams = d.metaCommitments.params,
-                purpose = InteractiveTxBuilder.FundingTxRbf(d.metaCommitments.common, d.metaCommitments.commitments.head),
+                purpose = InteractiveTxBuilder.FundingTxRbf(d.metaCommitments.common, d.metaCommitments.commitments.head, previousTransactions = d.allFundingTxs),
                 localPushAmount = d.localPushAmount, remotePushAmount = d.remotePushAmount,
                 wallet))
-              txBuilder ! InteractiveTxBuilder.Start(self, d.allFundingTxs)
+              txBuilder ! InteractiveTxBuilder.Start(self)
               stay() using d.copy(rbfStatus = RbfStatus.RbfInProgress(txBuilder)) sending TxAckRbf(d.channelId, fundingParams.localAmount)
             }
           case RbfStatus.RbfAborted =>
@@ -498,10 +498,10 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
           val txBuilder = context.spawnAnonymous(InteractiveTxBuilder(
             nodeParams, fundingParams,
             commitmentParams = d.metaCommitments.params,
-            purpose = InteractiveTxBuilder.FundingTxRbf(d.metaCommitments.common, d.metaCommitments.commitments.head),
+            purpose = InteractiveTxBuilder.FundingTxRbf(d.metaCommitments.common, d.metaCommitments.commitments.head, previousTransactions = d.allFundingTxs),
             localPushAmount = d.localPushAmount, remotePushAmount = d.remotePushAmount,
             wallet))
-          txBuilder ! InteractiveTxBuilder.Start(self, d.allFundingTxs)
+          txBuilder ! InteractiveTxBuilder.Start(self)
           stay() using d.copy(rbfStatus = RbfStatus.RbfInProgress(txBuilder))
         case _ =>
           log.info("ignoring unexpected tx_ack_rbf")
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
@@ -79,7 +79,7 @@ object InteractiveTxBuilder {
 
   // @formatter:off
   sealed trait Command
-  case class Start(replyTo: ActorRef[Response], previousTransactions: Seq[SignedSharedTransaction]) extends Command
+  case class Start(replyTo: ActorRef[Response]) extends Command
   sealed trait ReceiveMessage extends Command
   case class ReceiveTxMessage(msg: InteractiveTxConstructionMessage) extends ReceiveMessage
   case class ReceiveCommitSig(msg: CommitSig) extends ReceiveMessage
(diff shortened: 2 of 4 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -455,24 +455,17 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
-                d.fundingParams.fundingPubkeyScript,
-                msg.lockTime,
-                d.fundingParams.dustLimit,
-                msg.feerate,
-                d.fundingParams.requireConfirmedInputs,
+              val fundingParams = d.fundingParams.copy(
+                // we don't change our funding contribution
+                remoteAmount = msg.fundingContribution,
+                lockTime = msg.lockTime,
+                targetFeerate = msg.feerate
```
pm47, line 462: This makes it clearer what parameters are actually modified.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -455,24 +455,17 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
               val txBuilder = context.spawnAnonymous(InteractiveTxBuilder(
-                remoteNodeId, nodeParams, fundingParams,
-                d.localPushAmount, d.remotePushAmount,
-                d.metaCommitments.params,
-                d.commitments.localCommit.spec.commitTxFeerate,
-                d.commitments.remoteCommit.remotePerCommitmentPoint,
-                d.commitments.remoteNextCommitInfo.toOption.get,
+                nodeParams, fundingParams,
+                commitmentParams = d.metaCommitments.params,
+                purpose = InteractiveTxBuilder.RbfInitialCommitment(d.commitments.common, d.commitments.commitment),
```
pm47, line 467: No more repetition (and no more weird `remoteNextCommitInfo.toOption.get`), we explicitly reuse the `common` object that we already have.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
```diff
@@ -114,6 +115,31 @@ object InteractiveTxBuilder {
...
 
+  // @formatter:off
+  sealed trait Purpose {
+    def previousLocalBalance: MilliSatoshi = 0.msat
+    def previousRemoteBalance: MilliSatoshi = 0.msat
+    def commitmentIndex: Long = 0
+    def perCommitmentPoint: PublicKey
+    def commitTxFeerate: FeeratePerKw
+    def common: Common
+  }
```
pm47, line 126: I'm not sure about the name. The real distinction is whether:
- we create the first commitment
- we replace the first commitment
- we spend the existing commitment (splice)

We can make it conceptually independent from the notion of splice. Maybe I'm going too deep.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
```diff
@@ -114,6 +115,31 @@ object InteractiveTxBuilder {
...
+    override val common: Common = Common(
+      localChanges = LocalChanges(Nil, Nil, Nil), remoteChanges = RemoteChanges(Nil, Nil, Nil),
+      localNextHtlcId = 0L, remoteNextHtlcId = 0L,
+      localCommitIndex = 0L, remoteCommitIndex = 0L,
+      originChannels = Map.empty,
+      remoteNextCommitInfo = Right(nextPerCommitmentPoint),
+      remotePerCommitmentSecrets = ShaChain.init
+    )
+  }
+  case class RbfInitialCommitment(common: Common, commitment: Commitment) extends Purpose {
```
pm47, line 137: Initially I went with:

```suggestion
  case class RbfInitialCommitment(common: Common, commitment: Commitment, previousTransactions: Seq[InteractiveTxBuilder.SignedSharedTransaction]) extends Purpose {
```

It is more pure (there is no ambiguity when rbfing) and works well, but would force to heavily refactor the tests, so I backtracked to get some feedback first.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
t-bast: It feels like you're trying hard to create an abstraction, but there is a 1:1 mapping between the abstraction and the concrete use-case, so it feels like the abstraction is just making things less clear...What about being more explicit (at least initially) about the scenarios we support?

- `Purpose` can probably stay `Purpose`
- `InitialCommitment` -> `FundingTx`
- `RbfInitialCommitment` -> `FundingTxRbf`
- (future) `ExistingCommitment` -> `SpliceTx`
- (future) `RbfExistingCommitment` -> `SpliceTxRbf`

Those are the only use-cases we can think of right now, and they all need different parameters. We will probably later use `InteractiveTx` for mutual close, but this will also require a slightly different handling, so it will use a different `Purpose`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
```diff
@@ -114,6 +115,31 @@ object InteractiveTxBuilder {
     val minNextFeerate: FeeratePerKw = targetFeerate * 25 / 24
   }
 
+  // @formatter:off
+  sealed trait Purpose {
+    def previousLocalBalance: MilliSatoshi = 0.msat
+    def previousRemoteBalance: MilliSatoshi = 0.msat
+    def commitmentIndex: Long = 0
```
t-bast, line 122: I'm not a big fan of having default values in traits, it's easier to check for correctness when each purpose explicitly sets those values, otherwise a future purpose may incorrectly inherit invalid default values.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
```diff
@@ -114,6 +115,31 @@ object InteractiveTxBuilder {
...
+    def perCommitmentPoint: PublicKey
```
t-bast, line 123: Is this local or remote? I believe it's remote, it's worth being explicit with the name, isn't it?
```suggestion
    def remotePerCommitmentPoint: PublicKey
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
t-bast: I see what you mean, the `aliceRbf` and `bobRbf` actors that are currently created directly in the fixture are a bit limiting here...it was a bit lazy to implement it this way, it could make sense to refactor this which would unblock more testing scenarios (e.g. more than 1 RBF attempt), and it would make sense to have the `previousTransactions` in this class. I think it would be a nice change, I'll let you decide depending on the amount of work necessary to rework the tests!

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
t-bast: But it's also a bit dangerous because we may forget to update some of the parameters...having them all listed ensures we think about each one and whether it can still use the previous value or not. I don't have a strong opinion, both approaches have their pros and cons.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
t-bast, line 467: I think we should be explicit about using the `metaCommitments` and its first `Commitment` here:

```suggestion
                purpose = InteractiveTxBuilder.RbfInitialCommitment(d.metaCommitments.common, d.metaCommitments.commitments.head),
```

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -496,24 +489,17 @@ trait ChannelOpenDualFunded extends DualFundingHandlers with ErrorHandlers {
...
           val txBuilder = context.spawnAnonymous(InteractiveTxBuilder(
-            remoteNodeId, nodeParams, fundingParams,
-            d.localPushAmount, d.remotePushAmount,
-            d.metaCommitments.params,
-            d.commitments.localCommit.spec.commitTxFeerate,
-            d.commitments.remoteCommit.remotePerCommitmentPoint,
-            d.commitments.remoteNextCommitInfo.toOption.get,
+            nodeParams, fundingParams,
+            commitmentParams = d.metaCommitments.params,
+            purpose = InteractiveTxBuilder.RbfInitialCommitment(d.commitments.common, d.commitments.commitment),
```
t-bast, line 501:
```suggestion
            purpose = InteractiveTxBuilder.RbfInitialCommitment(d.metaCommitments.common, d.metaCommitments.commitments.head),
```

t-bast (commented):
Looks mostly good to me :+1:

eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
pm47: Went with your suggestion in d47aec354.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
pm47: Renamed in d47aec354.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
pm47: Done in d47aec354.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
pm47: Done in d47aec354.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
pm47: I prefer the `copy`, as it is more friendly for casual readers. Indeed special care will be needed for future parameters, but I trust our review process!

eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxBuilder.scala
pm47: Let me take a 2nd shot at it 🤓

pm47: Refactored tests in 49c763809, and did the change in dbc8b6015. Pretty proud of that one!

t-bast (approved):
LGTM!

PR #2579, opened 2023-01-23
Refactor commitments and meta-commitments

t-bast:
Move all the logic that was previously inside `Commitments` to `MetaCommitments`.
This is split in many small commits that should be reviewed independently.

I recommend looking at the diff between the code that is removed from `Commitments` and the code added to `MetaCommitments`. You'll see that it is quite minimal and mechanical, the only subtlety is that whenever the `localCommit` or `remoteCommit` is updated, we must not forget updating the `localCommitIndex` or `remoteCommitIndex` in the `common` field.

Some of these changes show where the current model doesn't work well:

- `remotePerCommitmentPoint` is duplicated across all remote commits: it's not obvious, but it has to be the same value
- htlcs should be shared between commitments
- the coupling between remoteNextCommitInfo and `nextRemoteCommit_opt` is really weird to work with (especially in `receiveRevocation`): this can be fixed by slightly re-working the model (future PR)
- I'm not sure funding public keys should be in `LocalParams` and `RemoteParams`: I believe people wanted to use different keys for each splice operation, so it should be a property of each `Commitment`?

I made e2e tests with eclair v0.8.0 and everything seems to be working fine :+1:


t-bast:
Rebased to fix conflicts with `master`.

eclair-core/src/main/scala/fr/acinq/eclair/channel/MetaCommitments.scala
```diff
@@ -74,125 +392,577 @@ case class MetaCommitments(params: Params,
...
 
-  def receiveFulfill(fulfill: UpdateFulfillHtlc): Either[ChannelException, (MetaCommitments, Origin, UpdateAddHtlc)] = {
-    sequence(all.map(_.receiveFulfill(fulfill)))
-      .map { res: List[(Commitments, Origin, UpdateAddHtlc)] => (res.head._1.common, res.map(_._1.commitment), res.head._2, res.head._3) }
-      .map { case (common, commitments, origin, add) => (this.copy(common = common, commitments = commitments), origin, add) }
-  }
+    val common1 = common.addRemoteProposal(add).copy(remoteNextHtlcId = common.remoteNextHtlcId + 1)
+
+    // let's compute the current commitment *as seen by us* including this change
+    commitments.foreach(commitment => {
```
pm47, line 531: I believe the body of this `foreach` should be a member of `Commitment`.

pm47 (commented):
I'm unconvinced by this change, because, even if the previous `MetaCommitments` was sometimes cumbersome in some ways, it was also very straightforward and made it clear how it simply dispatched methods to multiple commitments.

This new version is still cumbersome (due the limitations in the model that you mentioned), but  now the `MetaCommitments` is a monolithic piece of code that appears like it manages all commitments by itself.

Put it another way: I think this refactoring would made a lot of sense if we were able to put all the methods in either `Params`/`Common`/`Commitment`, and then have `MetaCommitments` methods either:
- be shortcuts to `params.x` or `common.y`
- do `commitments.foreach(_.xyz)`

This way we would keep the clarity about what `MetaCommitments` itself is about.`

Am I missing something?

pm47 (approved):
Discussed offline, this is a necessary first step before reworking the data model. We can't yet separate `Commitment`-specific methods because things may be impacted later.

PR #2585, opened 2023-01-24
Allow sending message to route that starts with us

thomash-acinq:
In case we are the introduction node of a message blinded route, we would not be able to send a message to that route. We now unwrap the first hop in case the route starts with us.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
@@ -62,8 +62,7 @@ object OnionMessages {
         val (nextNodeId, nextBlinding) = RouteBlindingEncryptedDataCodecs.decode(originKey.get, route.blindingKey, route.blindedNodes.head.encryptedPayload) match {
           case Left(invalidData) => throw new Exception(invalidData.toString)
           case Right(decoded) =>
-            (decoded.tlvs.get[RouteBlindingEncryptedDataTlv.OutgoingNodeId].get.nodeId,
-              decoded.tlvs.get[RouteBlindingEncryptedDataTlv.NextBlinding].map(_.blinding).getOrElse(decoded.nextBlinding))
+            (decoded.tlvs.get[RouteBlindingEncryptedDataTlv.OutgoingNodeId].get.nodeId, decoded.nextBlinding)
         }
         Sphinx.RouteBlinding.BlindedRoute(nextNodeId, nextBlinding, route.blindedNodes.tail)
       case Recipient(nodeId, pathId, padding) =>
--- a/eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
@@ -44,10 +44,11 @@ object OnionMessages {
 
   def buildRoute(blindingSecret: PrivateKey,
                  intermediateNodes: Seq[IntermediateNode],
-                 destination: Destination): Sphinx.RouteBlinding.BlindedRoute = {
+                 destination: Destination,
+                 originKey: Option[PrivateKey] = None): Try[Sphinx.RouteBlinding.BlindedRoute] = Try{
     val last = destination match {
-      case Recipient(nodeId, _, _) => OutgoingNodeId(nodeId) :: Nil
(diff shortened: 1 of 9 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/message/Postman.scala
```diff
@@ -102,13 +102,40 @@ object Postman {
             case Failure(f) =>
               replyTo ! MessageFailed(f.getMessage)
             case Success((nextNodeId, message)) =>
-              if (replyPath.isEmpty) { // not expecting reply
-                sendStatusTo += (messageId -> replyTo)
-              } else { // expecting reply
-                subscribed += (messageId -> replyTo)
-                context.scheduleOnce(timeout, context.self, Unsubscribe(messageId))
+              if (nextNodeId == nodeParams.nodeId) {
+                OnionMessages.process(nodeParams.privateKey, message) match {
```
t-bast, line 106: Shouldn't this be done in `OnionMessages.buildMessage` instead? It looks like it would be more natural there, we would start by checking if we are the introduction node of the `destination`'s blinded path:

- if yes, we decrypt the first hop of that route
- otherwise, we create a new blinded path towards the `destination` as currently done

Also, I'm not sure we correctly handle the case where we're asked to send a message to ourselves? Is this something we actually need to handle explicitly, or will it never happen?

thomash-acinq: We don't handle the sending to ourselves scenario indeed. I just don't think there is a reason to do that.

t-bast: No need to explicitly handle it then if it's not an issue. What do you think of doing the blinded path unwrapping in `buildMessage`?

thomash-acinq: Yes, it could be done in `OnionMessages.buildMessage` too, I just thought it was easier to it like that. I can refactor it inside `OnionMessages.buildMessage` if you prefer.

t-bast: I think it would make more sense there, because any caller of `buildMessage` wants the first blinded hop to be unwrapped if we're the introduction node, right?

thomash-acinq: Done

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -44,10 +44,11 @@ object OnionMessages {
 
   def buildRoute(blindingSecret: PrivateKey,
                  intermediateNodes: Seq[IntermediateNode],
-                 destination: Destination): Sphinx.RouteBlinding.BlindedRoute = {
+                 destination: Destination,
+                 originKey: Option[PrivateKey] = None): Try[Sphinx.RouteBlinding.BlindedRoute] = Try{
```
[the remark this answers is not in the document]
thomash-acinq: The reason it is optional is that we don't always know who is the origin node. It could be us if we're building a route to use it to send a message but it could be anyone if we're building a reply path.

Regarding exceptions, your alternative is just more verbose. I think `Try { myOption.get }` is much cleaner than `myOption match { case Some(x) => Success(x) case None => Failure(new IllegalArgumentException("empty option")) }`.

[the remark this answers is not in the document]
t-bast: > The reason it is optional is that we don't always know who is the origin node.

But we don't care about the origin node, we only care about the case where _we_ are the origin node, right? The only thing we want to check here is that if we (ie `nodeParams.nodeId`) are the introduction node of the destination's blinded route, we should decrypt one hop.

> I think Try { myOption.get } is much cleaner

`myOption.get` throws an exception, which captures a stack trace, which is expensive and should simply always be avoided, this is one of the most repeated coding best practice: never throw exceptions when you can easily avoid it.

thomash-acinq:
Now using `Either` instead of exceptions and the `buildRoute` function now only works with a node id as the recipient (the version that deals with a blinded route recipient is now private)

t-bast (approved):
LGTM  :+1:

PR #2586, opened 2023-01-24
Equality for TlvStream

thomash-acinq:
The order of the elements in a TLV stream is an implementation detail that will disappear with serialization. Equality between TlvStream shouldn't depend on this order.


pm47 (commented):
I think the proper way to do that is to change the collection types to `Set`s:

```scala
case class TlvStream[T <: Tlv](records: Set[T], unknown: Set[GenericTlv] = Nil)
```

And handle the ordering in the serialization, because, as you mention, it's a serialization concern.

Playing with equality of case classes is footgunny.

t-bast:
Why do you want to implement equality on tlv streams in the first place?

thomash-acinq:
> Why do you want to implement equality on tlv streams in the first place?

I need it for bolt12 to check that an invoice matches an invoice request (I extract a `TlvStream[InvoiceRequestTlv]` from the invoice and check that it equals the invoice request).

t-bast:
Ok that makes sense, but I wouldn't go with changing the equality implementation of the case class. Since you can only have a single instance of a given tlv in a stream, it makes more sense to do as @pm47 mentioned and use `Set`s. This may complicate slightly the codec (which needs to read/write and ordered `Seq`) but I don't think it should be too hard?

thomash-acinq:
Now using sets.

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
```diff
@@ -497,7 +497,7 @@ object ReplyChannelRange {
             checksums: Option[ReplyChannelRangeTlv.EncodedChecksums]): ReplyChannelRange = {
     timestamps.foreach(ts => require(ts.timestamps.length == shortChannelIds.array.length))
     checksums.foreach(cs => require(cs.checksums.length == shortChannelIds.array.length))
-    new ReplyChannelRange(chainHash, firstBlock, numberOfBlocks, syncComplete, shortChannelIds, TlvStream(timestamps.toList ::: checksums.toList))
+    new ReplyChannelRange(chainHash, firstBlock, numberOfBlocks, syncComplete, shortChannelIds, TlvStream(Set(timestamps, checksums).flatten: Set[fr.acinq.eclair.wire.protocol.ReplyChannelRangeTlv]))
```
t-bast, line 500: nit:
```suggestion
new ReplyChannelRange(chainHash, firstBlock, numberOfBlocks, syncComplete, shortChannelIds, TlvStream(Set(timestamps, checksums).flatten[ReplyChannelRangeTlv]))
```

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
```diff
@@ -655,10 +655,10 @@ object ChannelRelayerSpec {
   )
 
   def createBlindedPayload(update: ChannelUpdate, isIntroduction: Boolean): ChannelRelay.Blinded = {
-    val tlvs = TlvStream[OnionPaymentPayloadTlv](Seq(
+    val tlvs = TlvStream[OnionPaymentPayloadTlv](Set(
       Some(OnionPaymentPayloadTlv.EncryptedRecipientData(hex"2a")),
       if (isIntroduction) Some(OnionPaymentPayloadTlv.BlindingPoint(randomKey().publicKey)) else None,
-    ).flatten)
+    ).flatten: Set[OnionPaymentPayloadTlv])
```
t-bast, line 661:
```suggestion
    ).flatten[OnionPaymentPayloadTlv])
```

t-bast (commented):
This is definitely cleaner, ACK!

eclair-core/src/main/scala/fr/acinq/eclair/wire/protocol/LightningMessageTypes.scala
thomash-acinq: fixed

eclair-core/src/test/scala/fr/acinq/eclair/payment/relay/ChannelRelayerSpec.scala
thomash-acinq: fixed

t-bast (approved):
LGTM :100:

PR #2587, opened 2023-01-25
Use `MetaCommitments` as much as possible

t-bast:
This PR contains many independent commits that refactor usage of `Commitments` in favor of `Params`, `Common`, `Commitment` and `MetaCommitments` where it makes sense. This is a first step before doing more work on the data model of `MetaCommitments`: this PR touches a lot of lines of code, but makes only trivial changes, while data model changes will be more subtle, so we want them to have to update less lines of code to help future reviews.

One commit is particularly large (https://github.com/ACINQ/eclair/commit/4c9bbf2dccc46def7b1a2a83db8a21771aa8a7cd). But it only consists of several search/replace, and doesn't contain any logical change. One big advantage of this change in my opinion is that we can now simply look at all places where `metaCommitments.latest` is used to see where we actually need a commitment, and may need to be smart about which one we'll use (not a concern yet, but will be in the future when we implement the final version of splicing and need to handle all error cases). Before that commit, looking at all the places that depended on `Commitments` would return way too many results to help the development process.


pm47 (commented):
Indeed the changeset is huge, but it's probably safe, because all non-splice channels (which is, really, all channels) have a single `Commitment` so the change is purely mechanical.

I was expecting the `Commitments` class to disappear entirely though, what was the blocker?

If we manage to nuke `Commitments`, the next step should be to rename `MetaCommitments` to `Commitments` (if not the type, at least the arg names).

t-bast:
> it's probably safe, because all non-splice channels (which is, really, all channels) have a single Commitment so the change is purely mechanical.

Yes exactly, for the current code without splices it's perfectly safe.

> I was expecting the Commitments class to disappear entirely though, what was the blocker?
> If we manage to nuke Commitments, the next step should be to rename MetaCommitments to Commitments (if not the type, at least the arg names).

I'm keeping that for the data model changes PR ;)

pm47 (dismissed):

t-bast:
Rebased.

pm47 (approved):

PR #2588, opened 2023-01-25
Absolute priority to local channels

thomash-acinq:
Even when there is a path that is a lot cheaper, we still want to use the local channel.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
@@ -305,7 +305,7 @@ object Graph {
     val totalFees = prev.fees + fee
     val cltv = if (edge.desc.a == sender && !includeLocalChannelCost) CltvExpiryDelta(0) else edge.params.cltvExpiryDelta
     val totalCltv = prev.cltv + cltv
-    weightRatios match {
+    val richWeight = weightRatios match {
       case Left(weightRatios) =>
         val hopCost = if (edge.desc.a == sender) 0 msat else nodeFee(weightRatios.hopCost, prev.amount)
         import RoutingHeuristics._
@@ -351,6 +351,12 @@ object Graph {
           val weight = totalFees.toLong + totalHopsCost.toLong + totalRiskCost + failureCost.toLong / totalSuccessProbability
           RichWeight(totalAmount, prev.length + 1, totalCltv, totalSuccessProbability, totalFees, totalHopsCost, weight)
         }
+    }
+    if (edge.desc.a == sender) {
+      // If this is a local channel it shouldn't add any weight. We always prefer local channels.
+      richWeight.copy(weight = prev.weight)
+    } else {
+      richWeight
     }
   }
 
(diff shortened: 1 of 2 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/router/Graph.scala
```diff
@@ -352,6 +352,12 @@ object Graph {
           RichWeight(totalAmount, prev.length + 1, totalCltv, totalSuccessProbability, totalFees, totalHopsCost, weight)
         }
     }
+    if (edge.desc.a == sender) {
+      // If this is a local channel it shouldn't add any weight. We always prefer local channels.
+      richWeight.copy(weight = prev.weight)
+    } else {
+      richWeight
+    }
```
t-bast, line 360: I'm not sure about that fix, it's effectively the same thing as always setting `includeLocalChannelCost` to `false`, isn't it?
If it makes us just ignore the value of `includeLocalChannelCost`, shouldn't we just remove that parameter? 

IIRC, we added that parameter because when we relay trampoline payments, we want to verify that the fees of or local channel is paid (otherwise we'd be relaying for free). But couldn't we re-work that to do the following:

- we remove `includeLocalChannelCost` from this file entirely and act as if it were always `false`
- this way we'll always favor local channels when relaying trampoline payments in the routes we find
- in `RouteCalculation.scala`, once we got the best routes from `yenKShortestPaths`, check the fees and apply `includeLocalChannelCost`: if the routes are then invalid, return a `RouteNotFound` error, this way we will send back a `TrampolineFeeInsufficient`, and when the sender retries with more fees, this time it will work and use local channels?
- even better than returning `RouteNotFound`, we could return an error that says we found routes but they cost `X sats` in fees: in the future when we rework the trampoline error messages in the spec, we can then include this information for the sender

What do you think?

thomash-acinq: > I'm not sure about that fix, it's effectively the same thing as always setting `includeLocalChannelCost` to `false`, isn't it? If it makes us just ignore the value of `includeLocalChannelCost`, shouldn't we just remove that parameter?

No, it's different. `RichWeight` has many fields, in particular it has a field `fees` which is used to check to the path satisfies the boundaries and a field `weight` which is used to decide which path is better. `includeLocalChannelCost` affects both, what I did here is to touch only `weight` to influence the "best" path decision while still checking that the fees are below the maximum we can afford.

We could also do as you suggest but it would require more changes, I tried to touch as little code as I could.

thomash-acinq: >  * even better than returning `RouteNotFound`, we could return an error that says we found routes but they cost `X sats` in fees: in the future when we rework the trampoline error messages in the spec, we can then include this information for the sender

That would only work in the very narrow case where the fees of the route including the local channel are too high but removing our fee move it below the limit.

t-bast (approved):
LGTM, that makes more sense than my suggestion :+1:

PR #2589, opened 2023-01-27
Clean up onion messages code

t-bast:
- stop throwing unnecessary exceptions
- unwrap route first hop in `buildMessage` instead of `buildRoute`

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
@@ -28,7 +28,7 @@ import scodec.{Attempt, DecodeResult}
 
 import scala.annotation.tailrec
 import scala.concurrent.duration.FiniteDuration
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 object OnionMessages {
 
@@ -42,10 +42,7 @@ object OnionMessages {
   case class Recipient(nodeId: PublicKey, pathId: Option[ByteVector], padding: Option[ByteVector] = None) extends Destination
   // @formatter:on
 
-  def buildRoute(blindingSecret: PrivateKey,
-                 intermediateNodes: Seq[IntermediateNode],
-                 destination: Destination,
-                 originKey: Option[PrivateKey] = None): Try[Sphinx.RouteBlinding.BlindedRoute] = Try{
+  def buildRoute(blindingSecret: PrivateKey, intermediateNodes: Seq[IntermediateNode], destination: Destination): Sphinx.RouteBlinding.BlindedRoute = {
     val last = destination match {
       case Recipient(nodeId, _, _) => Set(OutgoingNodeId(nodeId))
       case BlindedPath(Sphinx.RouteBlinding.BlindedRoute(nodeId, blindingKey, _)) => Set(OutgoingNodeId(nodeId), NextBlinding(blindingKey))
@@ -58,13 +55,6 @@ object OnionMessages {
         .map(tlvs => RouteBlindingEncryptedDataCodecs.blindedRouteDataCodec.encode(TlvStream(tlvs)).require.bytes)
     }
     destination match {
-      case BlindedPath(route) if originKey.exists(route.introductionNodeId == _.publicKey) =>
-        val (nextNodeId, nextBlinding) = RouteBlindingEncryptedDataCodecs.decode(originKey.get, route.blindingKey, route.blindedNodes.head.encryptedPayload) match {
-          case Left(invalidData) => throw new Exception(invalidData.toString)
-          case Right(decoded) =>
-            (decoded.tlvs.get[RouteBlindingEncryptedDataTlv.OutgoingNodeId].get.nodeId, decoded.nextBlinding)
-        }
-        Sphinx.RouteBlinding.BlindedRoute(nextNodeId, nextBlinding, route.blindedNodes.tail)
       case Recipient(nodeId, pathId, padding) =>
         val tlvs: Set[RouteBlindingEncryptedDataTlv] = Set(padding.map(Padding), pathId.map(PathId)).flatten
         val lastPayload = RouteBlindingEncryptedDataCodecs.blindedRouteDataCodec.encode(TlvStream(tlvs)).require.bytes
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -58,24 +58,24 @@ object OnionMessages {
...
-        val (nextNodeId, nextBlinding) = RouteBlindingEncryptedDataCodecs.decode(originKey.get, route.blindingKey, route.blindedNodes.head.encryptedPayload) match {
-          case Left(invalidData) => throw new Exception(invalidData.toString)
-          case Right(decoded) =>
-            (decoded.tlvs.get[RouteBlindingEncryptedDataTlv.OutgoingNodeId].get.nodeId,
-              decoded.tlvs.get[RouteBlindingEncryptedDataTlv.NextBlinding].map(_.blinding).getOrElse(decoded.nextBlinding))
+      case BlindedPath(route) if nodeKey.publicKey == route.introductionNodeId =>
+        RouteBlindingEncryptedDataCodecs.decode(nodeKey, route.blindingKey, route.blindedNodes.head.encryptedPayload) match {
+          case Left(f) => Failure(new IllegalArgumentException(f.message))
+          case Right(decoded) => decoded.tlvs.get[RouteBlindingEncryptedDataTlv.OutgoingNodeId].map(_.nodeId) match {
+            case Some(nextNodeId) => Success(Sphinx.RouteBlinding.BlindedRoute(nextNodeId, decoded.nextBlinding, route.blindedNodes.tail))
```
thomash-acinq, line 65: What happened to `decoded.tlvs.get[RouteBlindingEncryptedDataTlv.NextBlinding]`? If it is present, it should overwrite `decoded.nextBlinding`.

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
```diff
@@ -58,24 +58,24 @@ object OnionMessages {
...
+            case None => Failure(new IllegalArgumentException("outgoing_node_id is missing from route blinding data"))
+          }
         }
-        Sphinx.RouteBlinding.BlindedRoute(nextNodeId, nextBlinding, route.blindedNodes.tail)
       case Recipient(nodeId, pathId, padding) =>
         val tlvs: Set[RouteBlindingEncryptedDataTlv] = Set(padding.map(Padding), pathId.map(PathId)).flatten
         val lastPayload = RouteBlindingEncryptedDataCodecs.blindedRouteDataCodec.encode(TlvStream(tlvs)).require.bytes
-        Sphinx.RouteBlinding.create(blindingSecret, intermediateNodes.map(_.nodeId) :+ nodeId, intermediatePayloads :+ lastPayload).route
```
thomash-acinq, line 72: I don't agree that adding `Success` everywhere is cleaner.

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
t-bast: It was just duplicating the code that is inside `RouteBlindingEncryptedDataCodecs.decode`.

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
t-bast: Removing code that throws exception is a no-brainer: we simply should never throw exceptions if we can avoid it.

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
thomash-acinq: good catch, thanks

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
thomash-acinq: But why? The code is much nicer.

eclair-core/src/test/scala/fr/acinq/eclair/integration/MessageIntegrationSpec.scala
```diff
@@ -81,7 +81,7 @@ class MessageIntegrationSpec extends IntegrationSpec {
     val eventListener = TestProbe()
     nodes("B").system.eventStream.subscribe(eventListener.ref, classOf[OnionMessages.ReceiveMessage])
 
-    val blindedRoute = buildRoute(randomKey(), Seq(IntermediateNode(nodes("A").nodeParams.nodeId), IntermediateNode(nodes("B").nodeParams.nodeId), IntermediateNode(nodes("B").nodeParams.nodeId)), Recipient(nodes("B").nodeParams.nodeId, None)).get
+    val blindedRoute = buildRoute(nodes("A").nodeParams.privateKey, randomKey(), Seq(IntermediateNode(nodes("A").nodeParams.nodeId), IntermediateNode(nodes("B").nodeParams.nodeId), IntermediateNode(nodes("B").nodeParams.nodeId)), Recipient(nodes("B").nodeParams.nodeId, None)).get
```
thomash-acinq, line 84: This doesn't test anything interesting since we are no longer the introduction point for this route.

eclair-core/src/main/scala/fr/acinq/eclair/message/Postman.scala
```diff
@@ -88,10 +88,10 @@ object Postman {
           Behaviors.same
         case SendMessage(intermediateNodes, destination, replyPath, messageContent, replyTo, timeout) =>
           val messageId = randomBytes32()
-          val replyRoute = replyPath.map(replyHops => {
+          val replyRoute = replyPath.flatMap(replyHops => {
```
thomash-acinq, line 91: We want `replyRoute` to be `None` if and only if `replyPath` is `None` too. You're breaking the API by doing that.

eclair-core/src/main/scala/fr/acinq/eclair/message/OnionMessages.scala
t-bast: On top of having bad performance because it captures a stack trace that we don't need, and needing to catch that exception just to actually convert it to a `Failure()` (which is what the `Try` does), it's a shortcut to avoid properly handling error cases: the error message we'll get for a `option.get` won't be helpful at all compared to an error message that says that the tlv stream doesn't contain a necessary outgoing node id. The code that throws exceptions is more concise, but it's definitely not nicer IMO...

eclair-core/src/main/scala/fr/acinq/eclair/message/Postman.scala
```diff
@@ -88,10 +88,10 @@ object Postman {
...
             val intermediateHops = replyHops.dropRight(1).map(OnionMessages.IntermediateNode(_))
             val lastHop = OnionMessages.Recipient(replyHops.last, Some(messageId))
-            OnionMessages.buildRoute(randomKey(), intermediateHops, lastHop).get
+            OnionMessages.buildRoute(nodeParams.privateKey, randomKey(), intermediateHops, lastHop).toOption
```
thomash-acinq, line 94: What if we want the reply path to be zero hop? This breaks it.

eclair-core/src/main/scala/fr/acinq/eclair/message/Postman.scala
t-bast: You're talking about the case where `buildRoute` would return a failure, right? That case was previously simply unhandled and would have crashed the actor without sending a response to the `replyTo` (because the `.get` would have failed), which is strictly worse, isn't it?

But you're raising a good point, I assumed that the reply path would never actually _start_ at our node. It would feel dumb, because even if you want the receiver to directly send a response to you, the reply path you'd use would be a blinded path `Receiver -> You` so that they're never 100% sure you sent the message, right? Should we check that though and reject reply paths that start at our node?

eclair-core/src/main/scala/fr/acinq/eclair/message/Postman.scala
t-bast: But...that's completely dumb, isn't it? Even if you want the receiver to send the response directly to you, it's always preferable to use a blinded path from them to you, which gives deniability?

eclair-core/src/test/scala/fr/acinq/eclair/integration/MessageIntegrationSpec.scala
t-bast: Woups, definitely! This shows that something is missing in this test and the other one: we should explicitly assert that the resulting route's introduction node is what we expect. I'll fix that.

t-bast:
@thomash-acinq I've reworked that to follow my [initial comment](https://github.com/ACINQ/eclair/pull/2585#discussion_r1088062902) and decrypt the first hop if necessary in `buildMessage` instead of inside `buildRoute`. Is that better?

I think that most of the weirdness comes from the fact that `buildRoute` isn't private but is used in two different contexts: creating a reply path to ourselves and creating a route to the destination of a message. A hint that this is weird is that the code inside `buildRoute` is patter matching twice on `destination` to do different things, the only common part is creating the intermediate payloads. I'm wondering if it wouldn't be clearer if we refactored that further to separate building routes towards a `Recipient` and routes towards a `BlindedPath`, since only the `BlindedPath` case needs to handle the case where we're the first hop, and cannot be used to build a reply path to us?

eclair-core/src/main/scala/fr/acinq/eclair/message/Postman.scala
t-bast: This change disappeared in https://github.com/ACINQ/eclair/pull/2589/commits/9dd1f2e822a4563cefd66618326da537cdf2b612

eclair-core/src/main/scala/fr/acinq/eclair/message/Postman.scala
t-bast: This change disappeared in https://github.com/ACINQ/eclair/pull/2589/commits/9dd1f2e822a4563cefd66618326da537cdf2b612

PR #2592, opened 2023-01-31
Store `fundingParams` with the `sharedTx`

pm47:
It makes sense to store the params with each `rbf` attempt (as opposed to the last one currently), and allows us to remove `DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED.fundingParams`.

Note that I used error codecs to (not) deal with migration, I think it's appropriate in this particular case.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
@@ -511,9 +511,9 @@ final case class DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED(metaCommitments: MetaCommi
                                                       lastChecked: BlockHeight, // last time we checked if the channel was double-spent
                                                       rbfStatus: RbfStatus,
                                                       deferred: Option[ChannelReady]) extends PersistentChannelData {
-  def latestFundingTx: DualFundedUnconfirmedFundingTx = allFundingTxs.head
-  def previousFundingTxs: Seq[DualFundedUnconfirmedFundingTx] = allFundingTxs.tail
   def allFundingTxs: Seq[DualFundedUnconfirmedFundingTx] = metaCommitments.commitments.map(_.localFundingStatus).collect { case fundingTx: DualFundedUnconfirmedFundingTx => fundingTx }
+  def latestFundingTx: DualFundedUnconfirmedFundingTx = metaCommitments.latest.localFundingStatus.asInstanceOf[DualFundedUnconfirmedFundingTx]
+  def previousFundingTxs: Seq[DualFundedUnconfirmedFundingTx] = allFundingTxs diff Seq(latestFundingTx)
 }
 final case class DATA_WAIT_FOR_DUAL_FUNDING_READY(metaCommitments: MetaCommitments,
                                                   shortIds: ShortIds) extends PersistentChannelData
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
@@ -416,7 +416,7 @@ object LocalFundingStatus {
   sealed trait UnconfirmedFundingTx extends LocalFundingStatus
   /** In single-funding, fundees only know the funding txid */
   case class SingleFundedUnconfirmedFundingTx(signedTx_opt: Option[Transaction]) extends UnconfirmedFundingTx
-  case class DualFundedUnconfirmedFundingTx(sharedTx: SignedSharedTransaction, createdAt: BlockHeight) extends UnconfirmedFundingTx {
+  case class DualFundedUnconfirmedFundingTx(sharedTx: SignedSharedTransaction, createdAt: BlockHeight, fundingParams: InteractiveTxParams) extends UnconfirmedFundingTx {
     override def signedTx_opt: Option[Transaction] = sharedTx.signedTx_opt
   }
   case class ConfirmedFundingTx(tx: Transaction) extends LocalFundingStatus {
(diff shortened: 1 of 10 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -397,11 +395,9 @@ private[channel] object ChannelCodecs3 {
         ("lastChecked" | blockHeight) ::
         ("rbfStatus" | provide[RbfStatus](RbfStatus.NoRbf)) ::
         ("deferred" | optional(bool8, lengthDelimited(channelReadyCodec)))).map {
-      case metaCommitments :: fundingTx :: fundingParams :: localPushAmount :: remotePushAmount :: previousFundingTxs :: waitingSince :: lastChecked :: rbfStatus :: deferred :: HNil =>
-        val metaCommitments1 = metaCommitments
-          .modify(_.commitments.at(0).localFundingStatus).setTo(DualFundedUnconfirmedFundingTx(fundingTx, waitingSince))
-          .modify(_.commitments).using(_ ++ previousFundingTxs.map { case DualFundingTx(f, c) => c.commitment.copy(localFundingStatus = DualFundedUnconfirmedFundingTx(f, waitingSince)) })
-        DATA_WAIT_FOR_DUAL_FUNDING_CONFIRMED(metaCommitments1, fundingParams, localPushAmount, remotePushAmount, waitingSince, lastChecked, rbfStatus, deferred)
+      case metaCommitments :: fundingTx :: fundingParams :: localPushAmount :: remotePushAmount :: _ :: waitingSince :: lastChecked :: rbfStatus :: deferred :: HNil =>
```
pm47, line 398: Notice how we are not using `previousFundingTxs`: the codec would have failed earlier.

Let me know if you find this approach a bit too _cavalier_ 🐴.

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
```diff
@@ -505,16 +505,15 @@ final case class DATA_WAIT_FOR_DUAL_FUNDING_CREATED(channelId: ByteVector32,
...
                                                       localPushAmount: MilliSatoshi,
                                                       remotePushAmount: MilliSatoshi,
                                                       waitingSince: BlockHeight, // how long have we been waiting for a funding tx to confirm
                                                       lastChecked: BlockHeight, // last time we checked if the channel was double-spent
                                                       rbfStatus: RbfStatus,
                                                       deferred: Option[ChannelReady]) extends PersistentChannelData {
-  def latestFundingTx: SignedSharedTransaction = metaCommitments.latest.localFundingStatus.asInstanceOf[DualFundedUnconfirmedFundingTx].sharedTx
-  def previousFundingTxs: Seq[SignedSharedTransaction] = metaCommitments.commitments.map(_.localFundingStatus).collect { case DualFundedUnconfirmedFundingTx(sharedTx, _) if sharedTx.txId != latestFundingTx.txId => sharedTx }
-  def allFundingTxs: Seq[SignedSharedTransaction] = latestFundingTx +: previousFundingTxs
+  def latestFundingTx: DualFundedUnconfirmedFundingTx = allFundingTxs.head
```
t-bast, line 514: Using `.head` makes an assumption on the ordering used inside `metaCommitments.commitments`, that's why I'd rather access it more explicitly via `metaCommitments.latest` which removes that assumption, but if you think it's acceptable we can keep it that way.

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
```diff
@@ -18,7 +18,7 @@ package fr.acinq.eclair.channel.fsm
 
 import akka.actor.typed.scaladsl.adapter.{ClassicActorContextOps, actorRefAdapter}
 import akka.actor.{ActorRef, Status}
-import com.softwaremill.quicklens.{ModifyPimp, QuicklensAt}
+import com.softwaremill.quicklens.{ModifyPimp, QuicklensAt, QuicklensWhen}
```
t-bast, line 21: nit: this is unused

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
```diff
@@ -312,9 +312,7 @@ private[channel] object ChannelCodecs3 {
     /** Once a dual funding tx has been signed, we must remember the associated commitments. */
     case class DualFundingTx(fundingTx: SignedSharedTransaction, commitments: Commitments)
 
-    private val dualFundingTxCodec: Codec[DualFundingTx] = (
-      ("fundingTx" | signedSharedTransactionCodec) ::
-        ("commitments" | commitmentsCodec)).as[DualFundingTx]
+    private val dualFundingTxCodec: Codec[DualFundingTx] = fail[DualFundingTx](Err("you have been running dual funding before it was officially released! contact developers"))
```
t-bast, line 315: I think you don't even need to bother with this error or backwards-compat' here, nobody should be using dual-funding in the wild yet since eclair and cln aren't even compatible at that point.

t-bast (commented):
Looks mostly good, I almost did that when I initially implemented dual funding, so let's go for it!

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
pm47: I used `.head` because I was also using `.tail`, and it allowed me to remove the cast to `DualFundedUnconfirmedFundingTx`.

Alternatively I could do:
```scala
  def allFundingTxs: Seq[DualFundedUnconfirmedFundingTx] = metaCommitments.commitments.map(_.localFundingStatus).collect { case fundingTx: DualFundedUnconfirmedFundingTx => fundingTx }
  def latestFundingTx: DualFundedUnconfirmedFundingTx = metaCommitments.latest.localFundingStatus.asInstanceOf[DualFundedUnconfirmedFundingTx]
  def previousFundingTxs: Seq[DualFundedUnconfirmedFundingTx] = allFundingTxs diff Seq(latestFundingTx)
```

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
t-bast: Why? Can't you just modify the `dualFundingTxCodec` without caring about backwards-compat?

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
t-bast: I think that's better since it decouples us from the internals of `MetaCommitments`

eclair-core/src/main/scala/fr/acinq/eclair/wire/internal/channel/version3/ChannelCodecs3.scala
t-bast: Oh I see we can't because we've previously been clever with the `dualFundingTxCodec` and made it work with the previous codec for funding txs

eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/ChannelOpenDualFunded.scala
pm47: Done in [04cd2d9](https://github.com/ACINQ/eclair/pull/2592/commits/04cd2d988e59c712ad535c3ddd5287ed0ad9f688).

eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelData.scala
pm47: I see your point: it's ugly, but morally right ^^. Done in 6ee161f69.

t-bast (approved):

PR #2552, opened 2023-01-04
Add support for plugins that intercept open channel messages 

remyers:
This PR introduces a new actor called `OpenChannelInterceptor` to intercept `OpenChannel` requests sent to a `Peer`actor from  the API (initiated) and remote peers (non-initiated).  

If a request is accepted, a `SpawnChannelInitiator` or  `SpawnChannelNonInitiator` message is sent back to the peer to continue the open channel flow. Otherwise an error message is sent back to the API call or remote peer. If a new request comes in while one is already being processed, it will return an error.

A new singleton actor called `PendingChannelsRateLimiter` is introduced to track pending channels and rate limit the creation of unconfirmed open channel request from remote peers. 

If the request is not rate limited and if an `InterceptOpenChannelPlugin` is defined, the plugin will be queried. The plugin must respond before a fixed timeout with either `AcceptOpenChannel` or `RejectOpenChannel`.  A subset of local channel parameters may also be modified when accepting channels. 

(Edit: updated scope of PR)

PR #2593, opened 2023-01-31
Add `ChannelAborted` event

t-bast:
That additional event lets subcribers know when a channel was closed without ever being used.

The possible flows for a channel lifecycle are now:

- ChannelCreated -> ChannelOpened -> ChannelAborted (when we don't have funds in the channel)
- ChannelCreated -> ChannelOpened -> ChannelClosed (when we have funds in the channel)
- ChannelCreated -> ChannelAborted (when we don't have funds in the channel)
- ChannelCreated -> ChannelClosed (when we have funds in the channel)
- ChannelAborted (when we don't have funds in the channel)
- ChannelAborted -> ChannelClosed (when we have funds in the channel)

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/ChannelEvents.scala
@@ -44,10 +44,11 @@ case class ChannelRestored(channel: ActorRef, channelId: ByteVector32, peer: Act
 
 case class ChannelIdAssigned(channel: ActorRef, remoteNodeId: PublicKey, temporaryChannelId: ByteVector32, channelId: ByteVector32) extends ChannelEvent
 
-/**
- * This event will be sent whenever a new scid is assigned to the channel, be it a real, local alias or remote alias.
- */
+/** This event will be sent whenever a new scid is assigned to the channel, be it a real, local alias or remote alias. */
 case class ShortChannelIdAssigned(channel: ActorRef, channelId: ByteVector32, shortIds: ShortIds, remoteNodeId: PublicKey) extends ChannelEvent
+
+/** This event will be sent if a channel was aborted before completing the opening flow. */
+case class ChannelAborted(channel: ActorRef, remoteNodeId: PublicKey, channelId: ByteVector32) extends ChannelEvent
 
 /** This event will be sent once a channel has been successfully opened and is ready to process payments. */
 case class ChannelOpened(channel: ActorRef, remoteNodeId: PublicKey, channelId: ByteVector32) extends ChannelEvent
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fsm/Channel.scala
@@ -241,6 +241,7 @@ class Channel(val nodeParams: NodeParams, val wallet: OnChainChannelFunder with 
         // NB: order matters!
         case closing: DATA_CLOSING if Closing.nothingAtStake(closing) =>
           log.info("we have nothing at stake, going straight to CLOSED")
+          context.system.eventStream.publish(ChannelAborted(self, remoteNodeId, closing.channelId))
           goto(CLOSED) using closing
         case closing: DATA_CLOSING =>
           val isInitiator = closing.metaCommitments.params.localParams.isInitiator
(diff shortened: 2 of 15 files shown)


remyers:
I can see how this will greatly simplify PR #2552! In that context, are these changes meant to ensure that we can rely on the fact that any time a channel actor is spawned, it will end up emitting either `ChannelOpened`, `ChannelClosed` or `ChannelAborted` at some point? 

For example, I can see that in the case of a channel receiving `BITCOIN_FUNDING_PUBLISH_FAILED` then the `ChannelErrorOccurred` event will be emitted but not `ChannelClosed` or `ChannelAborted`. So perhaps these changes only ensure `ChannelOpened`, `ChannelClosed` or `ChannelAborted` for a channel where we are not the initiator. What is the set of `ChannelEvents` we can expect before a channel actor is finally finished? is that set different if we initiate the channel open? 

I want to make sure I understand how the changes to the Channel FSM work. My understanding is that the new `ChannelAborted` event adds three new lifecycle sequences to the two that already produced `ChannelClosed` :

 1) when a peer closes without our node having any value at stake in the channel (`nothingAtStake(...)` is true):
- ChannelCreated -> ChannelOpened -> ChannelAborted
 2) the FSM is transitioning to the `CLOSED` state and is in a transient state prior to `ChannelCreated` being emitted:
- ChannelAborted
 3) the FSM transitioning to the `CLOSED` state and is in a state waiting for the funding tx to confirm:
- ChannelCreated -> ChannelAborted

t-bast:
> are these changes meant to ensure that we can rely on the fact that any time a channel actor is spawned, it will end up emitting either ChannelOpened, ChannelClosed or ChannelAborted at some point?

Yes exactly!

> For example, I can see that in the case of a channel receiving BITCOIN_FUNDING_PUBLISH_FAILED then the ChannelErrorOccurred event will be emitted but not ChannelClosed or ChannelAborted. 

In that case, the `ChannelAborted` event should be emitted in the `onTransition` block. If that's not the case, that is a bug! Can you point me to a specific line of code where this may happen?

> I want to make sure I understand how the changes to the Channel FSM work. My understanding is that the new ChannelAborted event adds three new lifecycle sequences to the two that already produced ChannelClosed

That looks correct!

t-bast:
I think you're right, very good catch, we can go from `WAIT_FOR_FUNDING_CONFIRMED` or `WAIT_FOR_CHANNEL_READY` to `CLOSING` by receiving an error, and then go to `CLOSED` if we realize that the funding transaction actually cannot be published. That transition from `CLOSING` to `CLOSED` doesn't generate any of the two events.

I think that a transition from `WAIT_FOR_FUNDING_CONFIRMED` or `WAIT_FOR_CHANNEL_READY` to `CLOSING` should trigger a `ChannelAborted`, what do you think?

remyers (commented):
Looks good! I rifted on your tests for the ready and transient states in [5713779](https://github.com/ACINQ/eclair/pull/2593/commits/5713779374d12769ed6aad6df69e5d04470a4d75)

(I meant to push them to my remote, but I guess it ended up in your branch)

t-bast:
> Looks good! I rifted on your tests for the ready and transient states in [5713779](https://github.com/ACINQ/eclair/pull/2593/commits/5713779374d12769ed6aad6df69e5d04470a4d75)
> (I meant to push them to my remote, but I guess it ended up in your branch)

I reviewed your commit, and it only adds tests which looks good, so I'm keeping it, thanks!
I fixed the failing tests in https://github.com/ACINQ/eclair/pull/2593/commits/3f295c8e1698f18e31ea6e93a221ca3928256499, should be ready to review now.

remyers:
> I reviewed your commit, and it only adds tests which looks good, so I'm keeping it, thanks!
> I fixed the failing tests in [3f295c8](https://github.com/ACINQ/eclair/commit/3f295c8e1698f18e31ea6e93a221ca3928256499), should be ready to review now.

Looks good; I was also looking at those failures and noticed a few more places we could check for `ChannelAborted` in those files. If you think the changes in [be6ab3c](https://github.com/remyers/eclair/commit/be6ab3c3ca53bc58c43de16c69e5a55c2948bb9d) are not too pendantic, then they might be worth adding. At some point when the `Channel` FSM is refactored they could be useful to have.

remyers (approved):
Looks good! 

Although the changes were simple enough, it was hard to figure out the best way to confirm `ChannelAborted` was emitted when expected. Looking at the FSM was difficult (though educational). Looking at and instrumenting the tests gave me more confidence that the state machine was emitting them when expected.

t-bast:
I added more tests based on your commit in https://github.com/ACINQ/eclair/pull/2593/commits/82221966933c43d383e8145da6f4e8f2fa879364

remyers (approved):
Looks good! Nice to have all the tests instrumented for `ChannelAborted` now.

PR #2594, opened 2023-02-01
Remove unnecessary dust limit hack in interactive-tx

t-bast:
We previously inflated the funding amount to always exceed the dust limit, otherwise bitcoind would reject the funding attempt. It complicated the code for a scenario that doesn't make sense in practice, so it's better to just let the funding attempt fail in that case.

We also randomize the order of inputs and outputs, which is a common privacy best practice.

### DIFF
--- a/eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxFunder.scala
+++ b/eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxFunder.scala
@@ -95,7 +95,6 @@ private class InteractiveTxFunder(replyTo: ActorRef[InteractiveTxFunder.Response
   }
 
   def start(): Behavior[Command] = {
-    require(fundingParams.localAmount >= 0.sat, "funding amount cannot be negative")
     log.debug("contributing {} to interactive-tx construction", fundingParams.localAmount)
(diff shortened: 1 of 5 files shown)


eclair-core/src/main/scala/fr/acinq/eclair/channel/fund/InteractiveTxFunder.scala
```diff
@@ -95,17 +95,9 @@ private class InteractiveTxFunder(replyTo: ActorRef[InteractiveTxFunder.Response
...
-      // to contribute to the shared output. We create a non-zero amount here to ensure that bitcoind will fund the
-      // fees for the shared output (because it would otherwise reject a txOut with an amount of zero).
-      fundingParams.localAmount.max(fundingParams.dustLimit)
-    } else {
-      fundingParams.localAmount
-    }
-    require(toFund >= 0.sat, "funding amount cannot be negative")
-    log.debug("contributing {} to interactive-tx construction", toFund)
-    if (toFund == 0.sat) {
+    require(fundingParams.localAmount >= 0.sat, "funding amount cannot be negative")
```
pm47, line 98: How about putting that `require` in the `InteractiveTxParams` class?

pm47 (commented):
Great stuff! A reasonable simplification in this tricky part of the code is welcome.

pm47 (approved):

PR #2596, opened 2023-02-03
Limit number of RBF attempts during dual funding

t-bast:
Each RBF attempt adds more data that we need to store and process, so we want to limit our peers to a reasonable use of RBF.
We sent a warning to let them know that they are close to reaching the limits.


pm47 (approved):

harding:
I'm curious about the motivation for this change (but it's just curiosity, so please feel free to ignore).  Given Bitcoin Core's default incrementalRelayFee (1 sat/vbyte), it would seem to me that each RBF of the funding transaction will cost the funder an extra ~0.04 USD at current prices (150 vbytes * 1e-8 * 25000 USD).  That's not much, but I would've guessed that it was plenty sufficient to prevent abuse.  Am I missing something?

t-bast:
Hey @harding! The issue here is that with the `interactive-tx` protocol, we cannot be sure that the resulting transaction is actually valid: our peer could be creating seemingly valid RBF transactions, which cost us resources (because we're watching for those transactions to potentially confirm or be spent by a force-close, and store data for the corresponding commitment transaction), but don't cost anything to the attacker because they're actually invalid transactions that cannot even be broadcast.

One case where that happens is when our peer uses an unconfirmed input that we don't have in our mempool: we can't know whether this input is fake or if we will receive it later, or discover it in a future block. We don't want to reject unconfirmed inputs that we don't know of, because it would harm some honest use-cases since mempool mismatches are expected. We figured that the simplest solution is to just limit the amount of DoS that can be generated that way by limiting the number of allowed RBF attempts, since honest peers shouldn't need a high number of RBF attempts unless they're doing something wrong, right?

harding:
@t-bast that makes sense!  I hadn't thought about the case of orphan txes, or other txes that an LN node couldn't validate.  Thank you!

PR #2597, opened 2023-02-03
Remove reserve requirement on first commitment

t-bast:
This requirement was always met for single-funded channels, but with dual funding it may not be met if the non-initiator contributes a very large amount (more than 100x what the initiator contributes). That scenario could happen, and the funding attempt shouldn't fail as long as the initiator's balance contains enough funds to pay the commit tx fees.


eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
```diff
@@ -437,16 +437,14 @@ object Helpers {
...
-          (fundingAmount / 100).max(localParams.dustLimit)
-        } else {
-          localParams.requestedChannelReserve_opt.get
-        }
-        val missing = toRemote.truncateToSatoshi - reserve - fees
-        if (missing < 0.sat) {
-          return Left(CannotAffordFees(channelId, missing = -missing, reserve = reserve, fees = fees))
+        val missing = fees - toRemote.truncateToSatoshi
+        if (missing > 0.sat) {
+          return Left(CannotAffordFees(channelId, missing = missing, reserve = 0 sat, fees = fees))
```
pm47, line 447: For clarity, we should define a `CannotAffordFees` that doesn't include a `reserve`.

pm47 (commented):
LGTM, just a nit.

eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala
t-bast: Good idea, done in https://github.com/ACINQ/eclair/pull/2597/commits/c80bd514848f3d024df78b156594f3d1d00cb2c6

pm47 (approved):

